PR #557, opened 2022-06-05
Add upper bounds to dep versions

adamjstewart:
Another attempt to make TorchGeo more stable. Assuming that our dependencies follow [semver](https://semver.org/), this PR assures that each TorchGeo release will continue to build for the indefinite future. 

For a concrete example of why this PR is necessary, our 0.2.1 release currently doesn't work with the latest version of torchmetrics since 0.9 introduced a backwards-incompatible change (fixed in #555). Our 0.2.0 and older releases also doesn't work with torchmetrics 0.8 due to another backwards-incompatible change (fixed in #361, #382). Since neither release specified an upper bound on the torchmetrics dependency, the simple `pip install torchgeo` no longer works, you have to manually specify a compatible version of torchmetrics too. In Spack and Conda, we can specify this restriction after the fact, but this isn't possible with pip without a new release.

Caveats:

* This PR assumes our dependencies know about and care enough to follow semver properly. It remains to be seen how closely our dependencies follow semver. Luckily, none of our dependencies use non-semver versioning schemes like dates.
* These upper bounds need to be updated any time a new major release (or minor release for 0.X versions) comes out. We'll see how this interacts with dependabot. For now, I'm following the GitHub release announcements for all of our dependencies, so that should be sufficient to catch these kinds of things.

Closes #544

PR #551, opened 2022-06-03
Set up dependabot configuration for pip

weiji14:
Enable dependabot to manage python dependencies pins for a more stable CI.

References:
- https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates

Closes #488

Issue #32, opened 2021-07-04
Determine minimum supported dependency versions


adamjstewart:
Before releasing, we should determine the minimum supported version of each dependency. We should also consider a test with this version just to make sure it doesn't change.

adamjstewart:
Now that we're using `requirements.txt` and dependabot, I think it wouldn't be unreasonable to add a CI test that uses a separate `requirements-min.txt` with the minimum known working versions of our deps. We can then make sure the full range works as expected.

adamjstewart:
Just discovered that matplotlib does the same thing: https://github.com/matplotlib/matplotlib/blob/main/requirements/testing/minver.txt

PR #567, opened 2022-06-08
Resolve DeprecationWarnings

adamjstewart:
This is the third PR in a series of attempts to improve the stability and robustness of TorchGeo's build system (#557, #551).

If you look at the logs for our unit tests, you'll see a slew of warning messages, desperately trying to warn us that TorchGeo is using deprecated features that will be removed in future versions of our dependencies. Up until now, these were ignored until someone noticed/reported an issue. With this PR, these kind of warnings will become fatal. We weren't able to do something this extreme before, but now that we use dependabot, we can restrict these kind of failures to version update PRs, not user PRs.

### Deprecation Fixes

- [x] Rtree 1.0.0 introduced a new `len(index)` method to replace `index.get_size()` (see https://github.com/Toblerity/rtree/pull/207). The latter is now deprecated and may be removed someday.
- [x] Pillow 9.1.0 relocated all resampling methods from `Image.*` to `Image.Resampling.*`. The former is now deprecated and will be removed in Pillow 10

### Considerations

* When a dependency changes its API, should we use the new API and bump the required version of that dependency, or should we maintain compatibility with historic versions? 

I did the former for rtree/torchmetrics/other instances because I didn't see a good reason not to, but I did the latter for pillow because pillow-simd doesn't yet have a 9.1.0 release and normal pillow is much slower. There's also the question of how to maintain coverage of both branches. This may require a test using the minimum supported versions: #32.

* What's the best way to handle warnings in our dependencies, not in our own code?

So far I'm just ignoring these using `filterwarnings`, but that may not be sufficient. What I would really love is an option to only issue warnings for our own code, not for dependencies, but I don't believe that's an option. One potential issue here is if a dep of a dep has a new release that causes the tests to fail, but it isn't managed by dependabot because it's not a direct dep.

* How long should we keep around warning ignores? Even after they no longer exist in newer versions?

Ideally, our tests will pass not only on CI which uses the latest versions of everything, but also locally where users may have any version of these dependencies installed. I'm not sure if it's worth keeping a running list of all possible warning messages for all versions of our dependencies. One option would be to remove `"error"` from `filterwarnings` in `pyproject.toml` and instead use `pytest -Werror` in CI so that warnings are only considered errors in CI, not locally.

### DIFF
--- a/torchgeo/datasets/cyclone.py
+++ b/torchgeo/datasets/cyclone.py
@@ -146,7 +146,7 @@ class TropicalCycloneWindEstimation(VisionDataset):
             if img.height != self.size or img.width != self.size:
                 if tuple(int(v) for v in PIL.__version__.split(".")) >= (9, 1):
                     resample = Image.Resampling.BILINEAR
-                else:
+                else:  # pragma: no cover
                     resample = Image.BILINEAR
                 img = img.resize(size=(self.size, self.size), resample=resample)
             array: "np.typing.NDArray[np.int_]" = np.array(img)
--- a/torchgeo/datasets/cyclone.py
+++ b/torchgeo/datasets/cyclone.py
@@ -144,9 +144,10 @@ class TropicalCycloneWindEstimation(VisionDataset):
         filename = os.path.join(directory.format("source"), "image.jpg")
         with Image.open(filename) as img:
             if img.height != self.size or img.width != self.size:
-                if tuple(int(v) for v in PIL.__version__.split(".")) >= (9, 1):
+                # Added in PIL 9.1.0
+                try:
                     resample = Image.Resampling.BILINEAR
-                else:  # pragma: no cover
+                except AttributeError:  # pragma: no cover
                     resample = Image.BILINEAR
                 img = img.resize(size=(self.size, self.size), resample=resample)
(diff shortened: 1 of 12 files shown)


torchgeo/datasets/cyclone.py
```diff
@@ -143,7 +144,11 @@ def _load_image(self, directory: str) -> Tensor:
         filename = os.path.join(directory.format("source"), "image.jpg")
         with Image.open(filename) as img:
             if img.height != self.size or img.width != self.size:
-                img = img.resize(size=(self.size, self.size), resample=Image.BILINEAR)
+                if tuple(int(v) for v in PIL.__version__.split(".")) >= (9, 1):
```
adamjstewart, line 147: This is a bad idea: https://github.com/pytorch/vision/issues/6153

Will fix.

calebrob6 (approved):

calebrob6:
re Pillow being really slow, we do already have an opencv dependency, so could use that

adamjstewart:
According to https://github.com/pytorch/vision/issues/6153, Pillow-SIMD is actually faster than OpenCV and Tensor operations on the GPU.

PR #570, opened 2022-06-11
Datasets: download from HTTPS

adamjstewart:
We should always use HTTPS when available. UC Merced doesn't support HTTPS unfortunately.


calebrob6 (approved):

calebrob6:
I confirmed all new links work

PR #571, opened 2022-06-12
Docs: shorter dataset/datamodule/model names

### LINKED ISSUE #569: Shorter dataset names in docs
The section names in https://torchgeo.readthedocs.io/en/stable/api/datasets.html are quite long, we should use abbreviated names when possible. This is especially important for the navigation bar on the right-hand side.

adamjstewart:
Closes #569

### Before

Cluttered, can't easily count the number of datasets:

Screen Shot 2022-06-11 at 9 24 29 PM

### After

Simple, easy to find the dataset you're looking for:

Screen Shot 2022-06-11 at 9 24 42 PM

### DIFF
--- a/docs/api/datamodules.rst
+++ b/docs/api/datamodules.rst
@@ -6,13 +6,13 @@ torchgeo.datamodules
 Geospatial DataModules
 ----------------------
 
-Chesapeake Bay High-Resolution Land Cover Project
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


calebrob6:
My only complaint here is that we loose what all the acronyms are supposed to mean in some cases, e.g. ADVANCE:

image

calebrob6 (approved):

calebrob6:
(this isn't a big problem or anything and I agree the list is _much_ more clean)

adamjstewart:
Yes, we can put those acronyms in the package class docstring. I'll do that here.

PR #574, opened 2022-06-12
Test minimum version of deps

### LINKED ISSUE #32: Determine minimum supported dependency versions

adamjstewart:
This is the fourth PR in a series of attempts to improve the stability and robustness of TorchGeo's CI (#567, #557, #551).

With this PR, we now have a minimum supported version listed for all of our dependencies, and we actively test to ensure that this stays up-to-date with each commit. 

In the past, I tried not to overly restrict our dependencies if I didn't know the oldest supported version. With this PR, I'm taking the policy of "the oldest supported version is the oldest we can test". For example, many of our dependencies are pinned to versions due to Python 3.7 support or because another dependency requires a newer version, even if an older version would work with our code.

Closes #32

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -8,7 +8,7 @@ dependencies:
   - fiona>=1.8
   - h5py>=2.6
   - numpy>=1.17.2
-  - open3d>=0.11.2
+  - open3d>=0.13
   - pip
   - pycocotools>=2
   - pyproj>=2.2
@@ -39,7 +39,7 @@ dependencies:
     - git+https://github.com/pytorch/pytorch_sphinx_theme
     - radiant-mlhub>=0.2.1
     - rtree>=1
-    - scikit-learn>=0.19.2
+    - scikit-learn>=0.21
     - scipy>=1.1
     - segmentation-models-pytorch>=0.2
     - setuptools>=42
--- a/environment.yml
+++ b/environment.yml
@@ -10,7 +10,6 @@ dependencies:
   - numpy>=1.17.2
   - open3d>=0.11.2
   - pip
-  - protobuf<4
   - pycocotools>=2
   - pyproj>=2.2
   - python>=3.7
--- a/environment.yml
+++ b/environment.yml
@@ -30,7 +30,7 @@ dependencies:
     - nbsphinx>=0.8.5
     - omegaconf>=2.1
     - opencv-python>=3.4.2.17
-    - pandas>=0.23.2
+    - pandas>=1
     - pillow>=6.2
     - pydocstyle[toml]>=6.1
     - pytest>=6.1.2
--- a/environment.yml
+++ b/environment.yml
@@ -31,7 +31,7 @@ dependencies:
     - omegaconf>=2.1
     - opencv-python>=3.4.2.17
     - pandas>=1
-    - pillow>=6.2
+    - pillow>=8.2
     - pydocstyle[toml]>=6.1
     - pytest>=6.1.2
     - pytest-cov>=2.4
--- a/environment.yml
+++ b/environment.yml
@@ -7,7 +7,7 @@ dependencies:
   - einops>=0.3
   - fiona>=1.8
   - h5py>=2.6
-  - numpy>=1.17.2
+  - numpy>=1.18
   - open3d>=0.13
   - pip
   - pycocotools>=2
--- a/environment.yml
+++ b/environment.yml
@@ -39,8 +39,8 @@ dependencies:
     - git+https://github.com/pytorch/pytorch_sphinx_theme
     - radiant-mlhub>=0.2.1
     - rtree>=1
-    - scikit-learn>=0.19.2
-    - scipy>=1.1
+    - scikit-learn>=0.21
+    - scipy>=1.2
     - segmentation-models-pytorch>=0.2
     - setuptools>=42
     - sphinx>=4
--- a/environment.yml
+++ b/environment.yml
@@ -4,48 +4,47 @@ channels:
   - conda-forge
   - open3d-admin  # for open3d
 dependencies:
-  - einops
-  - fiona>=1.5
-  - h5py
-  - numpy
+  - einops>=0.3
+  - fiona>=1.8
+  - h5py>=2.6
+  - numpy>=1.17.2
   - open3d>=0.11.2
   - pip
   - protobuf<4
-  - pycocotools
+  - pycocotools>=2
   - pyproj>=2.2
   - python>=3.7
-  - pytorch>=1.7
+  - pytorch>=1.9
   - rarfile>=3
-  - rasterio>=1.0.16
+  - rasterio>=1.0.20
   - shapely>=1.3
   - torchvision>=0.10
   - pip:
-    - black>=21.4b0
+    - black[jupyter]>=21.8
     - flake8>=3.8
     - ipywidgets>=7
     - isort[colors]>=5.8
-    - jupyterlab
     - kornia>=0.6.4
-    - laspy>=2.0.0
+    - laspy>=2
     - mypy>=0.900
     - nbmake>=0.1
     - nbsphinx>=0.8.5
     - omegaconf>=2.1
-    - opencv-python
+    - opencv-python>=3.4.2.17
     - pandas>=0.23.2
-    - pillow>=2.9
+    - pillow>=6.2
     - pydocstyle[toml]>=6.1
-    - pytest>=6
+    - pytest>=6.1.2
     - pytest-cov>=2.4
-    - pytorch-lightning>=1.3
+    - pytorch-lightning>=1.5.1
     - git+https://github.com/pytorch/pytorch_sphinx_theme
     - radiant-mlhub>=0.2.1
     - rtree>=1
-    - scikit-learn>=0.18
-    - scipy>=0.9
+    - scikit-learn>=0.19.2
+    - scipy>=1.1
     - segmentation-models-pytorch>=0.2
     - setuptools>=42
     - sphinx>=4
-    - timm>=0.2.1
+    - timm>=0.4.12
     - torchmetrics>=0.7
     - zipfile-deflate64>=0.2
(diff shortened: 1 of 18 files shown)


setup.cfg
```diff
@@ -25,37 +25,40 @@ keywords = pytorch, deep learning, machine learning, remote sensing, satellite i
 
 [options]
 install_requires =
-    einops<0.5
-    # fiona 1.5+ required for fiona.transform module
-    fiona~=1.5
+    # einops 0.3+ required for einops.repeat
+    einops>=0.3,<0.5
+    # fiona 1.8+ required for reading empty files
+    fiona>=1.8,<2
```
adamjstewart, line 31: Our fake data for SpaceNet 4 contains empty `labels.geojson` files. Older versions of fiona raise a different error message than the one we test for.

TODO: can we just add real data to these files to support older fiona?

setup.cfg
```diff
@@ -25,37 +25,40 @@ keywords = pytorch, deep learning, machine learning, remote sensing, satellite i
...
     # kornia 0.6.4+ required for kornia.contrib.compute_padding
     kornia>=0.6.4,<0.7
-    matplotlib<4
-    numpy<2
+    # matplotlib 3.3+ required for (H, W, 1) image support in plt.imshow
+    matplotlib>=3.3,<4
```
adamjstewart, line 35: Some of our datasets (idtrees?) are plotting (H, W, 1) images, can we just squash these so we can support older matplotlib?

setup.cfg
```diff
@@ -25,37 +25,40 @@ keywords = pytorch, deep learning, machine learning, remote sensing, satellite i
...
     protobuf<4
     # pyproj 2.2+ required for CRS object
     pyproj>=2.2,<4
-    # pytorch-lightning 1.3+ required for gradient_clip_algorithm argument to Trainer
-    pytorch-lightning~=1.3
-    # rasterio 1.0.16+ required for CRS support
-    rasterio>=1.0.16,<2
+    # pytorch-lightning 1.5.1+ required for apply_to_collection bugfix
+    pytorch-lightning>=1.5.1,<2
+    # rasterio 1.0.20+ required for out_dtype parameter of DatasetReaderBase.read
```
adamjstewart, line 48: Unclear if this is really needed, could just cast to different dtype after reading

setup.cfg
calebrob6: Should we keep track of these types of questions somewhere? Or resolve them here?

setup.cfg
adamjstewart: Apparently the actual dataset occasionally contains empty files so this is on purpose.

setup.cfg
adamjstewart: I'm fine with either. If I get bored I'll try to fix these and support older versions. If I get busy I'll just merge this as is and deal with it another day.

calebrob6 (dismissed):

calebrob6:
Kk I'm happy with this regardless

adamjstewart:
TODO: update `environment.yml`

pyproject.toml
```diff
@@ -54,9 +56,6 @@ match_dir = "(datamodules|datasets|losses|models|samplers|torchgeo|trainers|tran
 addopts = "-m 'not slow'"
 # https://docs.pytest.org/en/latest/how-to/capture-warnings.html
 filterwarnings = [
-    # Treat all warnings as errors
-    "error",
```
adamjstewart, line 58: Took me 2 whole days to figure out that this is why the tests started failing all of a sudden after the rebase. I think this is too strict. Users running `pytest` from the command line will still see a list of warnings at the bottom. We only need to convert that to an error during CI.

environment.yml
```diff
@@ -5,48 +5,46 @@ channels:
...
-  - fiona>=1.5
-  - h5py
-  - numpy
+  - einops>=0.3
+  - fiona>=1.8
+  - h5py>=2.6
+  - numpy>=1.17.2
   - open3d>=0.11.2
   - pip
-  - protobuf<4
```
adamjstewart, line 14: No need to pin this here, it isn't used by CI so it won't cause us issues. If someone encounters an issue they can manually install an older protobuf.

setup.cfg
```diff
@@ -25,37 +25,38 @@ keywords = pytorch, deep learning, machine learning, remote sensing, satellite i
...
+    # matplotlib 3.3+ required for (H, W, 1) image support in plt.imshow
+    matplotlib>=3.3,<4
+    # numpy 1.17.2+ required by pytorch-lightning
+    numpy>=1.17.2,<2
     # omegaconf 2.1+ required for to_object method
-    omegaconf~=2.1
-    # pillow 2.9+ required for height attribute
-    pillow>=2.9,<10
-    # Required to address protocolbuffers/protobuf#10051
-    protobuf<4
```
adamjstewart, line 40: No need to pin this here, it isn't used by CI so it won't cause us issues. If someone encounters an issue they can manually install an older protobuf.

requirements/requirements.txt
```diff
@@ -47,7 +47,6 @@ scipy==1.7.3;python_version=='3.7'
 zipfile-deflate64==0.2.0
 
 # docs
-ipywidgets==7.7.0
```
adamjstewart, line 50: I added this once upon a time because I ran into issues locally, but the doc tests seem to pass without it and it isn't a direct dep, so I removed it

torchgeo/datasets/advance.py
```diff
@@ -199,7 +199,7 @@ def _load_target(self, path: str) -> Tensor:
                 "scipy is not installed and is required to use this dataset"
             )
 
-        array = wavfile.read(path)[1]
+        array = wavfile.read(path, mmap=True)[1]
```
adamjstewart, line 202: Without this I was seeing an issue with older scipy/torch:
```
tests/datasets/test_advance.py::TestADVANCE::test_getitem
  /home/adam/torchgeo/torchgeo/datasets/advance.py:203: UserWarning: The given NumPy array is not writeable, and PyTorch does not support non-writeable tensors. This means you can write to the underlying (supposedly non-writeable) NumPy array using the tensor. You may want to copy the array to protect its data or make it writeable before converting it to a tensor. This type of warning will be suppressed for the rest of this program. (Triggered internally at  /pytorch/torch/csrc/utils/tensor_numpy.cpp:180.)
```
It doesn't seem to be needed for newer scipy/torch, but as long as the tests pass I don't see any harm in always using this.

torchgeo/datasets/cyclone.py
```diff
@@ -146,7 +146,7 @@ def _load_image(self, directory: str) -> Tensor:
                 # Moved in PIL 9.1.0
                 try:
                     resample = Image.Resampling.BILINEAR
-                except AttributeError:  # pragma: no cover
+                except AttributeError:
```
adamjstewart, line 149: We can finally properly test these lines!

pyproject.toml
```diff
@@ -58,18 +60,33 @@ filterwarnings = [
     "error",
 
     # Warnings raised by dependencies of dependencies, out of our control
+    # https://github.com/Cadene/pretrained-models.pytorch/issues/221
+    "ignore:.* is deprecated and will be removed in Pillow 10:DeprecationWarning:pretrainedmodels.datasets.utils",
     # https://github.com/pytorch/vision/pull/5898
+    "ignore:.* is deprecated and will be removed in Pillow 10:DeprecationWarning:torchvision.transforms.functional_pil",
     # https://github.com/rwightman/pytorch-image-models/pull/1256
-    "ignore:.*is deprecated and will be removed in Pillow 10:DeprecationWarning",
```
adamjstewart, line 63: If we don't specify a module, this will ignore the test for all files, including our own. By specifying a model, we can ignore things with more fine-grained precision. We should probably do this for all warnings raised by deps of deps.

pyproject.toml
```diff
@@ -58,18 +60,33 @@ filterwarnings = [
...
+    "ignore:torch.distributed._sharded_tensor will be deprecated:DeprecationWarning:torch.distributed._sharded_tensor",
+    # https://github.com/rasterio/rasterio/issues/1742
+    # https://github.com/rasterio/rasterio/pull/1753
+    "ignore:Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated:DeprecationWarning:rasterio.crs",
+    # https://github.com/pytorch/pytorch/issues/60053
+    # https://github.com/pytorch/pytorch/pull/60059
+    "ignore:Named tensors and all their associated APIs are an experimental feature and subject to change.:UserWarning:torch.nn.functional",
+    # https://github.com/numpy/numpy/issues/14920
+    # https://github.com/numpy/numpy/issues/15748
+    "ignore:numpy.ufunc size changed, may indicate binary incompatibility:RuntimeWarning",
```
adamjstewart, line 83: Only occurs for the minimum version tests, but basically, at least one of our deps was built with a different version of numpy than we are using to run the tests. Doesn't seem to cause any harm other than the warning.

adamjstewart:
@calebrob6 this should be ready to merge now assuming the tests pass!

calebrob6:
Sick, nice work!!

torchgeo/datasets/cyclone.py
calebrob6: Finallllly!!!!!!

calebrob6:
Okay awesome, thanks!

pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [build-system]
 requires = [
+    # setuptools 42+ required for metadata.license_files support in setup.cfg
     "setuptools>=42,<63",
-    "wheel<0.38",
```
adamjstewart, line 4: PyPA no longer recommends listing wheel as a build system requirement: https://github.com/pypa/packaging.python.org/pull/1050

adamjstewart:
Omg, tests are finally passing. Too good to be true. Didn't even need to make it into triple digit commits.

adamjstewart:
TODO: bump version checks in test files.

calebrob6 (approved):

calebrob6 (approved):

PR #578, opened 2022-06-14
Add DeepGlobe dataset for land cover 

saumyasinha:
Here is a sample plot for the dataset. These are with alpha 0, 0.5, and 1 respectively.
sample_alpha0
sample_alpha0 5
sample_alpha1


calebrob6:
We should cite https://openaccess.thecvf.com/content_cvpr_2018_workshops/papers/w4/Demir_DeepGlobe_2018_A_CVPR_2018_paper.pdf in the docstring in addition to the kaggle/codalab links

torchgeo/datasets/deepglobelandcover.py
```diff
@@ -0,0 +1,264 @@
...
+    draw_semantic_segmentation_masks,
+    extract_archive,
+    rgb_to_mask,
+)
+
+
+class DeepGlobeLandCover(VisionDataset):
+    """DeepGlobe Land Cover Classification Challenge dataset.
+
+    The `DeepGlobe Land Cover Classification Challenge <https://competitions.codalab.org/competitions/18468/>` dataset
```
adamjstewart, line 28:
```suggestion
    The `DeepGlobe Land Cover Classification Challenge <https://competitions.codalab.org/competitions/18468>`_ dataset 
```
Two things here, the URL doesn't work with a trailing slash, and rST requires a trailing underscore to turn this into a URL (confusing, I know).

torchgeo/datasets/deepglobelandcover.py
```diff
@@ -0,0 +1,264 @@
...
+
+class DeepGlobeLandCover(VisionDataset):
+    """DeepGlobe Land Cover Classification Challenge dataset.
+
+    The `DeepGlobe Land Cover Classification Challenge <https://competitions.codalab.org/competitions/18468/>` dataset 
+    offers high-resolution sub-meter satellite imagery focusing for the task of semantic segmentation to detect areas of 
+    urban, agriculture, rangeland, forest, water, barren, and unknown.
+    It contains 1,146 satellite images of size 2448 x 2448 pixels in total, 
+    split into training/validation/test sets, the original dataset can be downloaded from 
+    `Kaggle <https://www.kaggle.com/datasets/balraj98/deepglobe-land-cover-classification-dataset>`.
```
adamjstewart, line 33:
```suggestion
    `Kaggle <https://www.kaggle.com/datasets/balraj98/deepglobe-land-cover-classification-dataset>`_.
```

torchgeo/datasets/deepglobelandcover.py
```diff
@@ -0,0 +1,264 @@
...
+    4. Water
+    5. Barren land
+    6. Unknown
+
+    File names for satellite images and the corresponding mask image are id_sat.jpg and id_mask.png, 
+    where id is an integer assigned to every image. 
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * DeepGlobe 2018: A Challenge to Parse the Earth Through Satellite Images
```
adamjstewart, line 59: Let's use a URL, that way people can quickly find and download the bibtex

torchgeo/datasets/deepglobelandcover.py
```diff
@@ -0,0 +1,264 @@
...
+        (255, 255, 255),
+        (0, 0, 0)
+    ]
+
+    def __init__(
+        self,
+        root: str = "data",
+        split: str = "train",
+        transforms: Optional[Callable[[Dict[str, Tensor]], Dict[str, Tensor]]] = None,
+        checksum: bool = False,
```
adamjstewart, line 91: Can this dataset be automatically downloaded?

torchgeo/datasets/deepglobelandcover.py
```diff
@@ -0,0 +1,264 @@
...
+    
+        if os.path.isfile(filepath):
+            if self.checksum and not check_integrity(filepath, self.md5):
+                raise RuntimeError("Dataset found, but corrupted.")
+            extract_archive(filepath)
+            return
+
+        # Check if the user requested to download the dataset
+        raise RuntimeError(
+            "Dataset not found in `root` directory, either specify a different"
```
adamjstewart, line 208: Let's print the root directory here so it's easier to tell if someone passes the wrong directory, that's what we do in other datasets

torchgeo/datasets/deepglobelandcover.py
```diff
@@ -0,0 +1,264 @@
...
+            + " `root` directory or manually download the dataset to this directory."
+        )
+
+    def plot(
```
adamjstewart, line 212: Can you add a saved figure to the PR description so we can make sure the plot looks correct?

torchgeo/datasets/deepglobelandcover.py
saumyasinha: This helps, thanks!

torchgeo/datasets/deepglobelandcover.py
```diff
@@ -0,0 +1,264 @@
...
+class DeepGlobeLandCover(VisionDataset):
+    """DeepGlobe Land Cover Classification Challenge dataset.
+
+    The `DeepGlobe Land Cover Classification Challenge <https://competitions.codalab.org/competitions/18468/>` dataset 
+    offers high-resolution sub-meter satellite imagery focusing for the task of semantic segmentation to detect areas of 
+    urban, agriculture, rangeland, forest, water, barren, and unknown.
+    It contains 1,146 satellite images of size 2448 x 2448 pixels in total, 
+    split into training/validation/test sets, the original dataset can be downloaded from 
+    `Kaggle <https://www.kaggle.com/datasets/balraj98/deepglobe-land-cover-classification-dataset>`.
+    However, we only use the training dataset with 803 images since the original test and valid dataset
```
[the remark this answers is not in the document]
saumyasinha: I see! I believe this competition ended in 2018 though and a couple of other papers that use this dataset only use the training data. This other dataset that I am using now is very well organized and is available on Kaggle just like the original one. Let me know what you think.

torchgeo/datasets/deepglobelandcover.py
saumyasinha: We need to be signed into Kaggle for this.

adamjstewart: Could add an optional dep on https://pypi.org/project/kaggle/ like we do radiant-mlhub.

torchgeo/datasets/deepglobelandcover.py
saumyasinha: Sure!

torchgeo/datasets/deepglobelandcover.py
saumyasinha: Makes sense!

torchgeo/datamodules/potsdam.py
```diff
@@ -37,7 +37,7 @@ def __init__(
             num_workers: The number of workers to use in all created DataLoaders
             val_split_pct: What percentage of the dataset to use as a validation set
         """
-        super().__init__()  # type: ignore[no-untyped-call]
```
calebrob6, line 40: This shouldn't have been removed -- this is one of the reasons that mypy is complaining

saumyasinha: Yes, I fixed this..thanks!

adamjstewart:
Can you add a `tests/data/deepglobelandcover/data.py` script that creates the fake data and zip file? See other `tests/data/*/data.py` scripts for examples.

torchgeo/datasets/deepglobelandcover.py
saumyasinha: I've shared it with Caleb! I can add it here too.

saumyasinha:
I made a small change in the ` draw_semantic_segmentation_masks` function inside `torchgeo/torchgeo/datasets/utils.py`.  The `classes` variable was earlier just using the unique values from the `mask`, but for segmentation scenarios where the image's mask doesn't include all the given classes such as an image with only "rivers" or "forest land" but not "urban land"...this function doesn't plot the mask right. I think this is because `classes`  should be set to all possible class labels so that `class_masks` can be appropriately created and colors can be correctly assigned. 

I may not have done a great job in describing the issue, but I am happy to talk more about it. This fixed the plots for me, but there might be more elegant ways of doing this. I do have to fix the current mypy issue, but this commit should show the changes I've made.

calebrob6:
Can you check to see what other `plot` methods use `draw_semantic_segmentation_masks` then test one of them with this change?

calebrob6:
Can you post them here?

adamjstewart:
Can you rebase your branch so that CI runs the new minimum version tests? There's a button on GitHub or you can do it manually from the command line. Also, can you add this dataset/datamodule to the documentation? You'll need to add a section to `docs/api/datasets.rst` and `docs/api/datamodules.rst` under non-geospatial datasets. You'll also need to add a line to the table in `docs/api/non_geo_datasets.csv`.

saumyasinha:
> Can you post them here?

Yeah! 
The following plots are a sample from Vaihingen dataset, having 5 out 6 classes in their segmented outputs. Attaching them here: 
The first one is plotted with the existing `draw_semantic_segmentation_masks`, followed by the one plotted with my suggested change and the third one is the manually downloaded maks image.
 
vahingen_sample10_with_existing_visualization
vahingen_sample10_with_modified_visualization
image

calebrob6 (dismissed):

calebrob6:
@adamjstewart, did you want to review this?

adamjstewart:
Yes, will review tomorrow

docs/api/non_geo_datasets.csv
```diff
@@ -1,33 +1,34 @@
-Dataset,Task,Source,# Samples,# Classes,Size (px),Resolution (m),Bands
```
adamjstewart, line 1: What changed in this file? I would expect only a 1-line change

tests/data/deepglobelandcover/data.py
```diff
@@ -0,0 +1,74 @@
...
+
+def generate_test_data(root: str, n_samples: int = 3) -> str:
+    """Creates test data archive for DeepGlobeLandCover dataset and
+    returns its md5 hash.
+    Args:
+        root (str): Path to store test data
+        n_samples (int, optional): Number of samples.
+    Returns:
+        str: md5 hash of created archive
+    """
```
adamjstewart, line 22:
```suggestion
    """Create test data archive for DeepGlobeLandCover dataset.
    
    Args:
        root: path to store test data
        n_samples: number of samples.
        
    Returns:
        md5 hash of created archive
    """
```
We don't run `pydocstyle` on these files but if we did it would require these changes

tests/data/deepglobelandcover/data.py
```diff
@@ -0,0 +1,74 @@
...
+
+    folder_path = os.path.join(root, "data")
+
+    train_img_dir = os.path.join(folder_path, "data", "training_data", "images")
+    train_mask_dir = os.path.join(folder_path, "data", "training_data", "masks")
+    test_img_dir = os.path.join(folder_path, "data", "test_data", "images")
+    test_mask_dir = os.path.join(folder_path, "data", "test_data", "masks")
+
+    if not os.path.exists(train_img_dir):
+        os.makedirs(train_img_dir)
```
adamjstewart, line 34:
```suggestion
    os.makedirs(train_img_dir, exist_ok=True)
```
Can do this for other lines too

tests/datamodules/test_deepglobelandcover.py
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
adamjstewart, line 1: I'm trying to get rid of all of the files in `tests/datamodules` and instead test them directly with a trainer. Can you delete this file and add a new line in `tests/trainers/test_segmentation.py` instead? These `tests/datamodules` tests don't actually ensure that the datamodule works properly.

tests/datasets/test_deepglobelandcover.py
```diff
@@ -0,0 +1,76 @@
...
+
+class TestDeepGlobeLandCover:
+    @pytest.fixture(params=["train", "test"])
+    def dataset(
+        self, monkeypatch: MonkeyPatch, request: SubRequest
+    ) -> DeepGlobeLandCover:
+        md5 = "2cbd68d36b1485f09f32d874dde7c5c5"
+        splits = ["train", "test"]
+        monkeypatch.setattr(DeepGlobeLandCover, "md5", md5)
+        monkeypatch.setattr(DeepGlobeLandCover, "splits", splits)
```
adamjstewart, line 26: I don't think it's necessary to monkeypatch `splits` right? These are the same values as the default

tests/datasets/test_deepglobelandcover.py
```diff
@@ -0,0 +1,76 @@
...
+        root = os.path.join("tests", "data", "deepglobelandcover")
+        split = request.param
+        transforms = nn.Identity()  # type: ignore[no-untyped-call]
```
adamjstewart, line 29:
```suggestion
        transforms = nn.Identity() 
```
Latest version of PyTorch adds type hints for this

torchgeo/datasets/deepglobelandcover.py
```diff
@@ -0,0 +1,267 @@
...
+    extract_archive,
+    rgb_to_mask,
+)
+
+
+class DeepGlobeLandCover(VisionDataset):
+    """DeepGlobe Land Cover Classification Challenge dataset.
+
+    The `DeepGlobe Land Cover Classification Challenge
+    <https://competitions.codalab.org/competitions/18468/>`__ dataset
```
adamjstewart, line 29:
```suggestion
    <https://competitions.codalab.org/competitions/18468>`__ dataset
```
URL with slash at the end doesn't work

torchgeo/datasets/deepglobelandcover.py
```diff
@@ -0,0 +1,267 @@
...
+    offers high-resolution sub-meter satellite imagery focusing for the task of
+    semantic segmentation to detect areas of urban, agriculture, rangeland, forest,
+    water, barren, and unknown. It contains 1,146 satellite images of size
+    2448 x 2448 pixels in total, split into training/validation/test sets, the original
```
adamjstewart, line 33: Dataset length and image size might be better to put in the bulleted lists below so it's easier to locate. Either way, it will be in the bigger docs table.

torchgeo/datasets/deepglobelandcover.py
```diff
@@ -0,0 +1,267 @@
...
+    dataset can be downloaded from `Kaggle <https://www.kaggle.com/datasets/balraj98/
+    deepglobe-land-cover-classification-dataset>`__.
+    However, we only use the training dataset with 803 images since the original test
+    and valid dataset are not accompanied by labels. The dataset that we use with a
+    custom train/test split can be downloaded from `Kaggle <https://www.kaggle.com/
+    datasets/geoap96/deepglobe2018-landcover-segmentation-traindataset>`__ (created as a
```
adamjstewart, line 39: Do URLs work when split onto multiple lines like this? Need to check the generated docs...

torchgeo/datasets/deepglobelandcover.py
```diff
@@ -0,0 +1,267 @@
...
+        self.checksum = checksum
+
+        self._verify()
+        if split == "train":
+            split_folder = "training_data"
+        else:
+            split_folder = "test_data"
+
+        self.files = []
+        for image in os.listdir(
```
adamjstewart, line 118: The order of files returned by `os.listdir` is non-deterministic. It might be wise to sort `self.files` by image key at the end to make things more reproducible.

torchgeo/datasets/deepglobelandcover.py
```diff
@@ -0,0 +1,267 @@
...
+            os.path.join(root, self.data_root, split_folder, "images")
+        ):
+            if image.endswith(".jpg"):
+                id = image[:-8]
+                image_path = os.path.join(
+                    root, self.data_root, split_folder, "images", image
+                )
+                mask_path = os.path.join(
+                    root, self.data_root, split_folder, "masks", str(id) + "_mask.png"
+                )
```
adamjstewart, line 128: Since there is always a 1-to-1 mapping between image_path and mask_path, why not only store one in `self.files` (use a list instead of a dict) and convert the name on the fly? Then we store less data in memory and sorting becomes easier.

adamjstewart (changes requested):

torchgeo/datasets/deepglobelandcover.py
adamjstewart: Looks like they do, cool.

torchgeo/datasets/deepglobelandcover.py
```diff
@@ -0,0 +1,267 @@
...
+    5. Barren land
+    6. Unknown
+
+    File names for satellite images and the corresponding mask image are id_sat.jpg and
+    id_mask.png, where id is an integer assigned to every image.
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://arxiv.org/pdf/1805.06561.pdf
+
```
adamjstewart, line 63: Also needs:
```rst
.. versionadded:: 0.3 
```
to document which version of TorchGeo this was added in.

docs/api/non_geo_datasets.csv
calebrob6: Line endings were Windows style, fixed

torchgeo/datasets/deepglobelandcover.py
calebrob6: Added `sorted()` around `os.listdir()` instead

torchgeo/datasets/deepglobelandcover.py
calebrob6: I got rid of the dictionaries which I think is a small improvement. Doing string operations on the fly (i.e. every index into the dataset) will be a little slower than simple lookup and two lists of size "number of files" is tiny, so I don't think this is a problem.

torchgeo/datasets/deepglobelandcover.py
calebrob6: Just going to leave this alone

tests/datamodules/test_deepglobelandcover.py
calebrob6: This is needed to get test coverage on the val_pct argument to the datamodule (similar to other tests in the datamodules directory).

calebrob6: (Although I added it to the segmentation trainer test list too)

calebrob6:
This should be good to go now!

docs/api/non_geo_datasets.csv
```diff
@@ -4,6 +4,7 @@ Dataset,Task,Source,# Samples,# Classes,Size (px),Resolution (m),Bands
 `BigEarthNet`_,C,Sentinel-1/2,"590,326",19--43,120x120,10,"SAR, MSI"
 `COWC`_,"C, R","CSUAV AFRL, ISPRS, LINZ, AGRC","388,435",2,256x256,0.15,RGB
 `Kenya Crop Type`_,S,Sentinel-2,"4,688",7,"3,035x2,016",10,MSI
+`Deep Globe Land Cover Challenge`_,S,DigitalGlobe +Vivid,803,7,"2,448x2,448",0.5,RGB
```
adamjstewart, line 7: Should this be "+ Vivid" with a space? I'm not very familiar with DigitalGlobe.

tests/datamodules/test_deepglobelandcover.py
adamjstewart: Can we add two `tests/conf/deepglobelandcover_*.yaml` files, one for val_pct = 0 and one for 0.5? That's what I did for other datasets.

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -0,0 +1,122 @@
...
+
+        Args:
+            stage: stage to set up
+        """
+        transforms = Compose([self.preprocess])
+
+        dataset = DeepGlobeLandCover(self.root_dir, "train", transforms=transforms)
+
+        self.train_dataset: Dataset[Any]
+        self.val_dataset: Dataset[Any]
```
adamjstewart, line 71: Would it be more specific to type these as `VisionDataset`?

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -0,0 +1,122 @@
...
+            )
+        else:
+            self.train_dataset = dataset
+            self.val_dataset = dataset
+
+        self.test_dataset = DeepGlobeLandCover(
+            self.root_dir, "test", transforms=transforms
+        )
+
+    def train_dataloader(self) -> DataLoader[Any]:
```
adamjstewart, line 85: Would it be more specific to type these as `DataLoader[Dict[str, Any]]`?

adamjstewart (commented):
Almost there! Just a few remaining questions

docs/api/non_geo_datasets.csv
calebrob6: No, +Vivid is the name of a basemap provided by DigitalGlobe

torchgeo/datamodules/deepglobelandcover.py
calebrob6: probably, by `dataset_split(...)` returns a `Subset[Any]` which isn't compatible. Do you know how to get around that?

torchgeo/datamodules/deepglobelandcover.py
calebrob6: It isn't clear to me how I can find what "DataLoader" expects as a type argument though

torchgeo/datamodules/deepglobelandcover.py
adamjstewart: Oh true, nevermind then, this is fine for now.

torchgeo/datamodules/deepglobelandcover.py
adamjstewart: I believe `Dataset` and `DataLoader` type args are the return type of the dataset. But since we're using `Dataset[Any]` I don't think there's an easy way for mypy to figure out that this is correct. Might want to just ignore this.

adamjstewart (approved):

torchgeo/datamodules/deepglobelandcover.py
calebrob6: Trying it anyway :)

tests/conf/deepglobelandcover_0.yaml
```diff
@@ -0,0 +1,19 @@
...
+    segmentation_model: "unet"
+    encoder_name: "resnet18"
+    encoder_weights: null
+    learning_rate: 1e-3
+    learning_rate_schedule_patience: 6
+    verbose: false
+    in_channels: 3
+    num_classes: 7
+    num_filters: 1
+    ignore_zeros: False
```
adamjstewart, line 14:
```suggestion
    ignore_index: null
```
Should we also have a `conf/deepglobelandcover.yaml` file? I guess that could be added later after we find a good set of hyperparams.

tests/conf/deepglobelandcover_5.yaml
```diff
@@ -0,0 +1,19 @@
...
+    segmentation_model: "unet"
+    encoder_name: "resnet18"
+    encoder_weights: null
+    learning_rate: 1e-3
+    learning_rate_schedule_patience: 6
+    verbose: false
+    in_channels: 3
+    num_classes: 7
+    num_filters: 1
+    ignore_zeros: null
```
adamjstewart, line 14:
```suggestion
    ignore_index: null
```

adamjstewart (approved):

adamjstewart:
Running the tests results additional files being created during zip file extraction. We should add these to the repo so that running the tests doesn't result in new files in `git status`.

adamjstewart:
https://github.com/microsoft/torchgeo/pull/651

PR #590, opened 2022-06-18
Add issue templates

### LINKED ISSUE #584: Add issue template
We should add an issue template with a couple of options:

* Questions: open a discussion instead
* Bug reports: report an issue
* Feature requests: suggest a new dataset/model/whatever

For bug reports, the template should ask for TorchGeo version at the minimum, maybe other diagnostic info.

For feature requests, there can be sections for things like rationale, proposed implementation, and possible alternatives.

See https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

adamjstewart:
This PR adds issue templates to the repository. They are loosely based on [PyTorch's](https://github.com/pytorch/pytorch/issues/new/choose) and [Spack's](https://github.com/spack/spack/issues/new/choose) issue templates. See the [official docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests) for more details on possible values and syntax.

I don't really know how to view/test these without forking the repo and merging them to main in the fork? I guess we can do all the wordsmithing in this PR and fix any bugs we find after merging.

Closes #584

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -1,0 +1,58 @@
+name: 🐛 Bug report
+description: Create a report to help us reproduce and fix a bug
+labels: [bug]
+
+body:
+- type: textarea
+  attributes:
+    label: Description
+    description: >
+      Please provide a clear and concise description of the bug.
+  validations:
+    required: true
+- type: textarea
+  attributes:
+    label: Steps to reproduce
+    description: >
+      We can't solve your issue if we can't reproduce it. Please provide a
+      [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example)
+      that shows how to reproduce the bug. If the bug requires any additional files to
+      reproduce, please upload those or provide a download link. Your code should be
+      runnable and include all relevant imports.
+    placeholder: |
+      1. Download any Landsat 8 scene from https://earthexplorer.usgs.gov/
+      2. Run the following code to reproduce the error
+
+      ```python
+      from torchgeo.datasets import Landsat8
+
+      ds = Landsat8("/path/to/downloads/directory")
+      ...
+      ```
+  validations:
+    required: true
+- type: input
+  attributes:
+    label: Version
+    description: |
+      What version of TorchGeo are you using? This can be found using the following code.
+      ```console
+      $ python
+      >>> import torchgeo
+      >>> torchgeo.__version__
+      X.Y.Z
+      ```
+      If you are using a development release (ends with ".dev0") please also include the specific git commit.
+
+      Whenever possible, try to reproduce your issue with the latest commit from `main`. You never know, someone may have already fixed your bug!
+    placeholder: |
+      1.2.3.dev0 (e1285e6cc6b65080e82bdbf7de9dea3f647d8b3b)
+  validations:
+    required: true
+- type: markdown
+  attributes:
+    value: >
+      Thanks for taking the time to report this bug! TorchGeo is an open-source project
+      maintained by its users. If you're Python savvy and want to contribute a pull
+      request to fix this bug, we'll be happy to review it. If not, we'll try to fix it
+      as long as we can reproduce it.
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,5 @@
+blank_issues_enabled: true
+contact_links:
+  - name: ❓ Questions
+    url: https://github.com/microsoft/torchgeo/discussions
+    about: Ask questions or discuss ideas with other TorchGeo users
--- a/.github/ISSUE_TEMPLATE/feature-request.yml
+++ b/.github/ISSUE_TEMPLATE/feature-request.yml
@@ -1,0 +1,38 @@
+name: 🚀 Feature request
+description: Submit a proposal/request for a new TorchGeo feature
+labels: [feature]
+
+body:
+- type: textarea
+  attributes:
+    label: Summary
+    description: >
+      A clear and concise summary of your suggestion.
+- type: textarea
+  attributes:
+    label: Rationale
+    description: >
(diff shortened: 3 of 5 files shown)


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,58 @@
...
+      from torchgeo.datasets import Landsat8
+
+      ds = Landsat8("/path/to/downloads/directory")
+      ...
+      ```
+  validations:
+    required: true
+- type: input
+  attributes:
+    label: Version
```
adamjstewart, line 36: Instead of just the TorchGeo version, we could expand this to the full development environment, including:

* OS: e.g., macOS 12.4
* Python: e.g., 3.9.12
* TorchGeo: e.g., 1.2.3.dev0 (e1285e6cc6b65080e82bdbf7de9dea3f647d8b3b)
* Installation method: e.g., pip, conda, spack, git, etc.
* Dependency versions: e.g., `pip list`, `conda list`, `spack find`, etc.

I'm not sure how useful the rest of these are since TorchGeo is pure-Python and pretty heavily tested...

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,58 @@
...
+      1.2.3.dev0 (e1285e6cc6b65080e82bdbf7de9dea3f647d8b3b)
+  validations:
+    required: true
+- type: markdown
+  attributes:
+    value: >
+      Thanks for taking the time to report this bug! TorchGeo is an open-source project
+      maintained by its users. If you're Python savvy and want to contribute a pull
+      request to fix this bug, we'll be happy to review it. If not, we'll try to fix it
+      as long as we can reproduce it.
```
adamjstewart, line 58: Wanted to clarify that TorchGeo isn't a commercial product with a team of developers just waiting to fix your bugs lol

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: true
+contact_links:
+  - name: ❓ Questions
+    url: https://github.com/microsoft/torchgeo/discussions
```
adamjstewart, line 4: Could replace this with a Slack invite link if we ever decide to go down that route

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,58 @@
...
+- type: textarea
+  attributes:
+    label: Description
+    description: >
+      Please provide a clear and concise description of the bug.
+  validations:
+    required: true
+- type: textarea
+  attributes:
+    label: Steps to reproduce
```
adamjstewart, line 15: A lot of other repos have sections for "expected result" and "actual result" but I always find myself writing something like "I would expect this to work" and "it doesn't work" so I've never found those sections super useful. Hopefully users will clarify when things don't work as they would expect.

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,58 @@
...
+      2. Run the following code to reproduce the error
+
+      ```python
+      from torchgeo.datasets import Landsat8
+
+      ds = Landsat8("/path/to/downloads/directory")
+      ...
+      ```
+  validations:
+    required: true
```
calebrob6, line 33: My only concern is making the reproducible example required. I don't want someone to _not_ report something just because they _have_ to provide a reproducible example. Maybe we can explain that bugs are much easier to fix if there is reproducible code and we don't have a lot of free time to make our own examples?

.github/ISSUE_TEMPLATE/bug-report.yml
calebrob6: Yeah I think torchgeo version is fine for now (i.e. until we see something crazy that is not caught by tests or explained by dependency/Python version)

.github/ISSUE_TEMPLATE/bug-report.yml
calebrob6: ➕

.github/ISSUE_TEMPLATE/feature-request.yml
```diff
@@ -0,0 +1,38 @@
...
+- type: textarea
+  attributes:
+    label: Additional information
+    description: >
+      Any additional information that might be relevant to the proposed feature.
+- type: markdown
+  attributes:
+    value: >
+      Thanks for suggesting this awesome new feature! TorchGeo is an open-source project
+      maintained by its users. If you're Python savvy and want to contribute a pull
```
calebrob6, line 36: Kind of sounds like we are cool with taking on any new features. Maybe we can rephrase to include that we'll discuss the feature, then, if it is in scope: forget about it for a few months, discuss again, then implement :)

.github/ISSUE_TEMPLATE/bug-report.yml
adamjstewart: I think having a _reproducible_ example is a low bar. At the bare minimum, just upload the code you ran and the files you ran it on to produce the error message. Having a _minimal_ reproducible example is a bigger hurdle, but users rarely consider that to be a requirement unfortunately. If someone reported an issue with just an error message and didn't tell me how they got that error message, I would probably just close the issue. I would go so far as to say having a reproducible example is the _only_ thing that is absolutely required to report an issue.

adamjstewart:
Added a note explaining the scope of features we accept. I think having a reproducible example should absolutely be required.

.github/ISSUE_TEMPLATE/bug-report.yml
calebrob6: That's totally fine -- just a minor concern! The more I think about it the more I agree, we really don't have a ton of time to dig into things (esp. if they are obscure).

calebrob6 (approved):

adamjstewart:
Looks great! https://github.com/microsoft/torchgeo/issues/new/choose

PR #553, opened 2022-06-03
Install only the needed lint tool in style checks CI

weiji14:
Instead of installing all of  `torchgeo`'s required dependencies, just install the lint tool (`black`/`flake8`/`isort`/`pydocstyle`/`pyupgrade`) directly. Should make for a faster style checks CI.

As per https://github.com/microsoft/torchgeo/pull/552#discussion_r889339370

PR #605, opened 2022-06-18
Move requirements-min.txt, split deps in multiple files

adamjstewart:
Turns out dependabot actually does support multiple files! I noticed this in #603 when it accidentally updated `requirements-min.txt` too. This PR does two things:

- [x] Move `requirements-min.txt` to a subdirectory to avoid dependabot updates
- [x] Split `requirements.txt` into separate files for each CI step

If `requirements/min/all.txt` doesn't avoid dependabot, I'll try moving it to a different directory.

Splitting `requirements.txt` into separate files fulfills @weiji14's vision from #553! Style tests now take 15 s instead of 2 min.


.github/workflows/release.yaml
```diff
@@ -36,7 +36,8 @@ jobs:
         python-version: '3.10'
     - name: Install pip dependencies
       run: |
-        pip install -r requirements/requirements.txt
+        pip install --pre 'rasterio>=1.0.16'
+        pip install .[datasets,tests]
```
adamjstewart, line 40: It's probably better to go back to installing from `setup.cfg` during release testing. We want to make sure this works since it's how actual PyPI installs will work.

setup.cfg
```diff
@@ -96,6 +96,8 @@ datasets =
     # https://github.com/brianhelba/zipfile-deflate64/issues/19
     zipfile-deflate64>=0.2,<0.3
 docs =
+    # ipywidgets 7+ required by nbsphinx
+    ipywidgets>=7,<8
```
adamjstewart, line 100: I removed this in https://github.com/microsoft/torchgeo/pull/574#discussion_r897595816 because I couldn't remember why it was required and it didn't seem to be needed anymore. Turns out the addition of `black[jupyter]` installed ipywidgets behind the scenes, so that's why the tests didn't fail. Splitting things up into separate files uncovered https://github.com/spatialaudio/nbsphinx/issues/24#issuecomment-1159529577 which explains why I added this in the first place.

requirements/datasets.txt
```diff
@@ -0,0 +1,18 @@
...
+# to add or remove a dependency. It documents the version of our dependencies used in
+# CI, these are not hard requirements for using TorchGeo.
+
+# datasets
+h5py==3.6.0
+laspy==2.1.2
+open3d==0.14.1;python_version<'3.10'
+opencv-python==4.6.0.66
+pandas==1.4.2;python_version>='3.8'
+pandas==1.3.5;python_version=='3.7'
```
adamjstewart, line 11: It's still unclear if dependabot will actually update lines like this, or if it will try to update both. Ideally, it would only update the first occurrence, or somehow parse the environment marker. If that doesn't work, we can create separate files for Python 3.7 (and 3.8 once numpy and friends drop that too).

calebrob6 (approved):

calebrob6:
Awesome, thanks for digging so far into this @adamjstewart!

PR #627, opened 2022-06-25
Rename VisionDataset to NonGeoDataset

adamjstewart:
This PR renames the following classes:

* `VisionDataset` -> `NonGeoDataset`
* `VisionClassificationDataset` -> `NonGeoClassificationDataset`

TorchGeo was never intended to be a computer vision library, it is for _all_ geospatial data types. Originally, we mostly had computer vision (satellite/aerial/drone imagery) datasets, but we now have multiple non-vision datasets:

* Canadian Building Footprints, Open Buildings: vector labels
* EDDMapS, GBIF, iNaturalist: point labels
* Aster Global DEM, EU-DEM: topography
* ADVANCE: audio

Question: should I keep `VisionDataset` and `VisionClassificationDataset` aliases with deprecation warnings to help ease the transition? I'm not sure how widely used we are or how stable our users expect us to be. We're still in alpha release, but users don't necessarily realize the implications of that.


adamjstewart:
Super easy to do, like 1 line of code to make it work and maybe 4 lines of code to add a deprecation warning.

calebrob6:
Cool, then why not!

adamjstewart:
Think I finally got this working. The only files you need to review are `torchgeo/datasets/geo.py` and `tests/datasets/test_geo.py`, everything else is a simple sed replacement. We can merge it now or merge it later, either way we'll need to change the superclass of any datasets that get added in the meantime.

calebrob6 (approved):

PR #448, opened 2022-03-02
GridGeoSampler: Sample beyond roi limit if needed

remtav:
fix for #431 as per suggested by @adamjstewart

PR #376, opened 2022-01-28
Samplers: avoid bounding boxes smaller than patch size

adamjstewart:
Fixes #319 

@RitwikGupta @tritolol

PR #629, opened 2022-06-28
GridGeoSampler: Sample beyond roi limit if needed

### LINKED ISSUE #431: GridGeoSampler may never reach end of image
In ran into this issue after running multiple tests and double-checking why my inference script using GridGeoSampler never reached the end (right and bottom) of an image. 

I'm using GridGeoSampler with an image of size 1591 x 2187 pixels, a chip size of 512 and stride of 256. For the sake of clarity, let's just say my image is 600 x 600 (pixel units or crs units doesn't matter here -> just assume minimum bounds in x and y are 0 if you're thinking crs units):

I want GridGeoSampler to sample my 600 x 600 image in chips of 512x512, with a stride of 256. 
The total number of row and columns [as calculated in the \_\_init__ and \_\_iter_ of GridGeoSampler](https://github.com/microsoft/torchgeo/blob/main/torchgeo/samplers/single.py#L198) will be:
rows = columns = (600 - 512) // 256 +1 = 1

Therefore, the \_\_iter__ function will use two for loops [with range(rows)](https://github.com/microsoft/torchgeo/blob/main/torchgeo/samplers/single.py#L205) and range (columns). Since those two variables are 1, only 1 512x512 sample will be returned by the sampler, thus ignoring 88 (600-512) pixels of width and height where valid values existed in my image.

To adapt the current implementation, I believe rows and columns should be calculated this way:
( image width - chip size + stride size ) // stride size + 1
or in the code, for rows:
`rows = int((bounds.maxy - bounds.miny - self.size[0] + self.stride[0]) // self.stride[0]) + 1`

remtav:
See discussion on #448 
Fixes #431 
Also reverts #376

### DIFF
--- a/tests/samplers/test_single.py
+++ b/tests/samplers/test_single.py
@@ -171,9 +171,13 @@ class TestGridGeoSampler:
 
     def test_iter(self, sampler: GridGeoSampler) -> None:
         for query in sampler:
-            assert sampler.roi.minx <= query.minx <= query.maxx <= sampler.roi.maxx
-            assert sampler.roi.miny <= query.miny <= query.miny <= sampler.roi.maxy
-            assert sampler.roi.mint <= query.mint <= query.maxt <= sampler.roi.maxt
+            assert sampler.roi.minx <= query.minx
+            assert sampler.roi.miny <= query.miny
+            assert sampler.roi.mint <= query.mint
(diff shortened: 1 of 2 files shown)


remtav:
Obviously, I'm not the best with git.

PR #630, opened 2022-06-28
GridGeoSampler: change stride of last patch to sample entire ROI

### LINKED ISSUE #431: GridGeoSampler may never reach end of image

remtav:
This PR changes the way in which `GridGeoSampler` samples patches from each tile if the tile size is not a multiple of the stride. We want to cover the entire tile, requiring us to adjust the stride of the final row/column.

This also changes the number of patches returned from each tile. Let $i$ be the size of the input tile. Let $k$ be the requested size of the output patch. Let $s$ be the requested stride. Let $o$ be the number of output rows/columns sampled from each tile.

### Before

$$ o = \left\lfloor \frac{i - k}{s} \right\rfloor + 1 $$

### After                                                                                         
                                                                                         
$$ o = \left\lceil \frac{i - k}{s} \right\rceil + 1 $$

Reboot of #448
Fixes #431


remtav:
@adamjstewart this is ready for review. I added more tests than necessary. I can remove some if you'd prefer.
Also, I'm not sure how to reproduce the [problem you mentionned](https://github.com/microsoft/torchgeo/pull/448#issuecomment-1111197241) when two tiles that have a common edge face create a sample with empty area. Tips to implement this tests are welcome.
Currently, [see TODO](https://github.com/microsoft/torchgeo/pull/630/files#diff-4f36986ccd487e7dc063a65dbce34e74ba44cc1b21fc9b8e32720512fc5cddd0R230-R231)

remtav:
It seems like a test with the NAIP dataset is failing, but on Windows, the tests passes. Can you try on Ubuntu @adamjstewart ? I'll be back home thursday to test on Ubuntu locally if needed.

calebrob6:
@remtav, can you add me as well?

calebrob6:
Okay so I just played with this for awhile, the NAIPChesapeakeDataModule has stride hardcoded as 128 which I think makes the 2nd box generated by the sampler jump out of the dataset bounds. The test data is 218x128 and strides of < 64 seem to work.

huajie2020 (approved):

adamjstewart:
Wait a minute. Do we want to sample beyond the ROI?? What if I have a geospatial train/val/test split and the sampler goes beyond the ROI of train into val/test? Wouldn't that be bad?

calebrob6:
An alternative behavior to make sure we sample all space within the ROI without sampling beyond is to force an additional overlapping box along the right and bottom sides of the ROI.

adamjstewart:
@calebrob6 which of these possible solutions does that correspond to? https://github.com/microsoft/torchgeo/pull/448#issuecomment-1069371131

calebrob6:
Adjust the stride of the last patch only

remtav:
> Adjust the stride of the last patch only -- e.g. of the logic is here https://github.com/microsoft/poultry-cafos/blob/main/cafo/data/TileDatasets.py#L65

That's also the solution I had implemented in my first solution (WIP):
https://github.com/microsoft/torchgeo/pull/630/commits/cb554c67386ae252a26028cbe7903b94f58b1bd6

adamjstewart:
Trying to release 0.3.1 in approximately 1 week. @remtav can you revert back to your original solution and we can iterate from there? If you're busy, I can take over from here. I think we've agreed on what the correct behavior should be, which was the hardest part. Note that #736 may also affect the same lines of code.

remtav:
> Trying to release 0.3.1 in approximately 1 week. @remtav can you revert back to your original solution and we can iterate from there? If you're busy, I can take over from here. I think we've agreed on what the correct behavior should be, which was the hardest part. Note that #736 may also affect the same lines of code.

Sure, I can revert and build on top of initial solution with unit tests and formatting clean up. I'll push at the latest on monday.

torchgeo/samplers/single.py
```diff
@@ -233,11 +248,29 @@ def __iter__(self) -> Iterator[BoundingBox]:
             for i in range(rows):
                 miny = bounds.miny + i * self.stride[0]
                 maxy = miny + self.size[0]
+                if maxy > bounds.maxy:
+                    last_stride_y = self.stride[0] - (miny - (bounds.maxy - self.size[0]))
+                    maxy = bounds.maxy
+                    miny = bounds.maxy - self.size[0]
+                    warnings.warn(
```
adamjstewart, line 255: You can get rid of the warnings, this will almost always be the case

torchgeo/samplers/single.py
```diff
@@ -257,7 +290,7 @@ class PreChippedGeoSampler(GeoSampler):
     and subclass :class:`~torchgeo.datasets.GeoDataset` but have already been
     pre-processed into :term:`chips <chip>`.
 
-    This sampler should not be used with :class:`~torchgeo.datasets.NonGeoDataset`.
+    This sampler should not be used with :class:`~torchgeo.datasets.VisionDataset`.
```
adamjstewart, line 293: This should be undone, we renamed the base class

adamjstewart (commented):
Still need to check to make sure the rounding is right, will compare to #736 later.

torchgeo/samplers/single.py
remtav: Yep! I'm currently on this. Will also make sure unit tests are adapted to this new (or old) solution

remtav:
Ready for review @adamjstewart 
As for rounding issues #736 , I've favored using `/`, then `math.ceil`, rather than the original `//` and `+1` approach (which doesn't seem right when ROI and grid sampling are equally sized anyways):
`int((bounds.maxy - bounds.miny - self.size[0]) // self.stride[0]) + 1`
Hopefully that does it, but please double check.

tests/samplers/test_single.py
```diff
@@ -182,24 +186,74 @@ def test_iter(self, sampler: GridGeoSampler) -> None:
...
+        rows = math.ceil(
+            (100 - sampler.size[0] + sampler.stride[0]) / sampler.stride[0]
+        )
+        cols = math.ceil(
+            (100 - sampler.size[1] + sampler.stride[1]) / sampler.stride[1]
+        )
+        length = rows * cols * 2  # two items in dataset
         assert len(sampler) == length
 
+    def test_len_larger(self, sampler: GridGeoSampler) -> None:
```
adamjstewart, line 198: What is the purpose of this test?

tests/samplers/test_single.py
```diff
@@ -182,24 +186,74 @@ def test_iter(self, sampler: GridGeoSampler) -> None:
...
             assert query in roi
 
     def test_small_area(self) -> None:
+        ds = CustomGeoDataset()
+        ds.index.insert(0, (0, 1, 0, 1, 0, 1))
+        sampler = GridGeoSampler(ds, 2, 10)
+        assert len(sampler) == 0
+
+    # TODO: skip patches with area=0 when two tiles are
+    #  side-by-side with an overlapping edge face.
```
adamjstewart, line 227: Is this still relevant? We already skip patches with height/width < size

tests/samplers/test_single.py
```diff
@@ -182,24 +186,74 @@ def test_iter(self, sampler: GridGeoSampler) -> None:
...
+        ds = CustomGeoDataset()
+        ds.index.insert(0, (0, 6, 0, 5, 0, 10))
+        sampler = GridGeoSampler(ds, 5, 5, units=Units.CRS)
+        assert len(sampler) == 2
+        assert list(sampler)[0] == BoundingBox(
+            minx=0.0, maxx=5.0, miny=0.0, maxy=5.0, mint=0.0, maxt=10.0
+        )
+        assert list(sampler)[1] == BoundingBox(
+            minx=1.0, maxx=6.0, miny=0.0, maxy=5.0, mint=0.0, maxt=10.0
+        )
```
adamjstewart, line 256: The point of these tests is to hardcode what the expected patches should be and how long they should be, correct? I like that idea, but still need to triple check the logic.

P.S. You can use ints instead of floats here.

adamjstewart (commented):
Code looks correct for all situations I can think of. Just a few remaining questions/suggestions before this is ready to merge.

torchgeo/samplers/single.py
```diff
@@ -200,17 +201,24 @@ def __init__(
...
         for hit in self.hits:
             bounds = BoundingBox(*hit.bounds)
 
-            rows = int((bounds.maxy - bounds.miny - self.size[0]) // self.stride[0]) + 1
-            cols = int((bounds.maxx - bounds.minx - self.size[1]) // self.stride[1]) + 1
+            # last patch samples outside the bounds
+            rows = math.ceil(
+                (bounds.maxy - bounds.miny - self.size[0] + self.stride[0])
+                / self.stride[0]
+            )
```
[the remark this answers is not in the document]
adamjstewart: Going to switch to +1 to match https://theano-pymc.readthedocs.io/en/latest/tutorial/conv_arithmetic.html#no-zero-padding-non-unit-strides and other derivations I've seen. The only difference between this and our implementation is that we replace floor with ceil since we want to sample the last patch.

adamjstewart (approved):

adamjstewart:
Thanks for the fix @remtav! Sorry this took so long to merge.

torchgeo/samplers/single.py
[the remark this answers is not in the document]
adamjstewart: Yep, glad we finally got it merged, sorry it took so long!

PR #637, opened 2022-06-29
Docs: don't override __module__

adamjstewart:
In Sphinx 3, import aliases would throw off intersphinx:

* https://github.com/pytorch/pytorch/issues/60979
* https://stackoverflow.com/questions/40018681

This was supposedly fixed in Sphinx 4:

* https://github.com/sphinx-doc/sphinx/issues/9395

I was finally able to upgrade PyTorch and torchvision to use Sphinx 5:

* https://github.com/pytorch/pytorch/pull/70309
* https://github.com/pytorch/vision/pull/5121

So in theory, we should no longer need to override `__module__`. Fingers crossed that the doc tests pass and the docs look okay!

### DIFF
--- a/tests/models/test_changestar.py
+++ b/tests/models/test_changestar.py
@@ -10,10 +10,6 @@ from torch.nn.modules import Module
 
 from torchgeo.models import ChangeMixin, ChangeStar, ChangeStarFarSeg
 
-# https://github.com/pytorch/pytorch/issues/60979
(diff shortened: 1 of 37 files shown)


adamjstewart:
Well, this doesn't work at all. Will open an issue with the Sphinx devs...

huajie2020 (approved):

PR #643, opened 2022-07-02
LandCover.ai: add data.py script

### LINKED ISSUE #633: Better LandCover.ai testing
### Summary

We should actually use OpenCV to create small image patches in our LandCover.ai tests.

### Rationale

LandCover.ai relies on OpenCV to convert large scenes to smaller patches. However, our tests don't actually use OpenCV, and the minimum version of OpenCV we claim to support may not actually work.

### Implementation

Convert the blank `split.py` we ship into a script that uses OpenCV in the same way as the file downloaded from the dataset.

### Alternatives

_No response_

### Additional information

_No response_

adamjstewart:
This was fun. I added a `data.py` to write a `split.py` to read fake GeoTIFFs and convert them to fake JPEGs and PNGs. Other than wanting a `data.py` for every dataset, the point of this exercise was that our tests actually use OpenCV now. This will test to see if our minimum supported version of OpenCV actually works.

Closes #633


tests/data/landcoverai/data.py
```diff
@@ -0,0 +1,120 @@
...
+
+# Create train/val/test splits
+files = ["M-33-20-D-c-4-2_0", "M-33-20-D-c-4-2_1"]
+for split in ["train", "val", "test"]:
+    with open(f"{split}.txt", "w") as f:
+        for file in files:
+            f.write(f"{file}\n")
+
+# Create split.py
+code = f"""\
```
calebrob6, line 88: yesss

calebrob6 (approved):

PR #644, opened 2022-07-02
SemanticSegmentationTrainer: ignore_zeros -> ignore_index

### LINKED ISSUE #444: Bug: missing "not" operator to set ignore_zeros in segmentation trainer
I believe there's a bug in the translation from ignore_zeros to ignore_index in the [\_\_init__ of SemanticSegmentationTask](https://github.com/microsoft/torchgeo/blob/c933694ddc9a121af9415389949aefcc84760472/torchgeo/trainers/segmentation.py#L89).
`self.ignore_zeros = None if kwargs["ignore_zeros"] else 0` should be:
`self.ignore_zeros = None if kwargs["ignore_zeros"] is None else 0` as is done [when define loss function](https://github.com/microsoft/torchgeo/blob/c933694ddc9a121af9415389949aefcc84760472/torchgeo/trainers/segmentation.py#L57)

`self.ignore_zeros` is later used to define the "self.ignore_index" of the self.train_metrics. For sake of clarity, I'd suggest renaming it to  `self.ignore_index` directly in the \_\_init__. 

Therefore, this
```
...
        self.ignore_zeros = None if kwargs["ignore_zeros"] else 0

        self.config_task()

        self.train_metrics = MetricCollection(
            [
                Accuracy(
                    num_classes=self.hparams["num_classes"],
                    ignore_index=self.ignore_zeros,
...
```
would become
```
...
        self.ignore_index = None if kwargs["ignore_zeros"] is None else 0

        self.config_task()

        self.train_metrics = MetricCollection(
            [
                Accuracy(
                    num_classes=self.hparams["num_classes"],
                    ignore_index=self.ignore_index,
...
```
If you'd like, I can definitely push a PR for to fix this bug if the solution seems ok to all. Let me know.

calebrob6:
Fixes #444

### DIFF
--- a/conf/chesapeake_cvpr.yaml
+++ b/conf/chesapeake_cvpr.yaml
@@ -17,7 +17,7 @@ experiment:
     in_channels: 4
     num_classes: 7
     num_filters: 256
-    ignore_zeros: False
+    ignore_zeros: null
     imagenet_pretraining: True
   datamodule:
     root_dir: "data/chesapeake/cvpr"
--- a/tests/conf/oscd_all.yaml
+++ b/tests/conf/oscd_all.yaml
@@ -11,7 +11,7 @@ experiment:
     in_channels: 26
     num_classes: 2
     num_filters: 1
-    ignore_zeros: False
+    ignore_index: null
   datamodule:
     root_dir: "tests/data/oscd"
     batch_size: 1
--- a/tests/conf/oscd_all.yaml
+++ b/tests/conf/oscd_all.yaml
@@ -11,7 +11,7 @@ experiment:
     in_channels: 26
     num_classes: 2
     num_filters: 1
-    ignore_zeros: True
+    ignore_zeros: False
   datamodule:
     root_dir: "tests/data/oscd"
     batch_size: 1
--- a/tests/trainers/test_segmentation.py
+++ b/tests/trainers/test_segmentation.py
@@ -123,6 +123,6 @@ class TestSemanticSegmentationTask:
     def test_ignoreindex_with_jaccard(self, model_kwargs: Dict[Any, Any]) -> None:
         model_kwargs["loss"] = "jaccard"
         model_kwargs["ignore_index"] = 0
-        match = "Warning ignore_index has no effect on training when loss='jaccard'"
+        match = "ignore_index has no effect on training when loss='jaccard'"
         with pytest.warns(UserWarning, match=match):
             SemanticSegmentationTask(**model_kwargs)
--- a/tests/trainers/test_segmentation.py
+++ b/tests/trainers/test_segmentation.py
@@ -97,7 +97,7 @@ class TestSemanticSegmentationTask:
             "encoder_name": "resnet18",
             "encoder_weights": None,
             "in_channels": 1,
-            "num_classes": 1,
+            "num_classes": 2,
             "loss": "ce",
             "ignore_zeros": True,
         }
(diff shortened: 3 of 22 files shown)


adamjstewart:
Is there any reason not to rename `ignore_zeros` to `ignore_index` in the `__init__` parameter list?

calebrob6:
We don't need to ignore other values and old segmentation trainer checkpoints will not be easily loadable

tests/conf/oscd_all.yaml
```diff
@@ -11,7 +11,7 @@ experiment:
     in_channels: 26
     num_classes: 2
     num_filters: 1
-    ignore_zeros: True
+    ignore_zeros: False
```
calebrob6, line 14: We shouldn't ignore_zeros with two classes, this causes the Accuracy metric to break

tests/trainers/test_segmentation.py
```diff
@@ -97,7 +97,7 @@ def model_kwargs(self) -> Dict[Any, Any]:
             "encoder_name": "resnet18",
             "encoder_weights": None,
             "in_channels": 1,
-            "num_classes": 1,
+            "num_classes": 2,
```
calebrob6, line 100: Instantiating the torchmetrics with a single class doesn't make sense

torchgeo/trainers/segmentation.py
```diff
@@ -89,20 +90,30 @@ def __init__(self, **kwargs: Any) -> None:
         self.save_hyperparameters()  # type: ignore[operator]
         self.hyperparams = cast(Dict[str, Any], self.hparams)
 
-        self.ignore_zeros = None if kwargs["ignore_zeros"] else 0
+        if not isinstance(kwargs["ignore_zeros"], bool):
+            raise ValueError("ignore_zeros must be a bool")
+        if kwargs["ignore_zeros"] and kwargs["loss"] == "jaccard":
```
calebrob6, line 95: I specifically warn about this because a user might expect training with jaccard to behave similarly to training with ce or focal loss, however it doesn't

adamjstewart:
> We don't need to ignore other values

We don't need to ignore other values _yet_. Presumably this feature exists in torchmetrics because some datasets use a class other than 0 for background. This will break as soon as someone adds a dataset with a different index that needs to be ignored.

> old segmentation trainer checkpoints will not be easily loadable

I'm concerned about reproducibility of experiments, but not necessarily reproducibility with older versions of TorchGeo. We are still in alpha mode and backwards incompatible changes like this would be better sooner rather than later. I still say we should switch to `ignore_index` to match torchmetrics.

calebrob6:
Okay fair enough! Any thoughts on the warning?

conf/chesapeake_cvpr.yaml
```diff
@@ -17,7 +17,7 @@ experiment:
     in_channels: 4
     num_classes: 7
     num_filters: 256
-    ignore_zeros: False
+    ignore_zeros: null
```
adamjstewart, line 20:
```suggestion
    ignore_index: null
```

torchgeo/trainers/segmentation.py
```diff
@@ -89,20 +89,29 @@ def __init__(self, **kwargs: Any) -> None:
         self.save_hyperparameters()  # type: ignore[operator]
         self.hyperparams = cast(Dict[str, Any], self.hparams)
 
-        self.ignore_zeros = None if kwargs["ignore_zeros"] else 0
-
+        if not (
+            isinstance(kwargs["ignore_index"], int) or (kwargs["ignore_index"] is None)
+        ):
```
adamjstewart, line 94:
```suggestion
        if not isinstance(kwargs["ignore_index"], (int, type(None))):
```
Another possible option, no preference on which you choose.

torchgeo/trainers/segmentation.py
```diff
@@ -89,20 +89,29 @@ def __init__(self, **kwargs: Any) -> None:
...
+            raise ValueError("ignore_index must be an int or None")
+        if (kwargs["ignore_index"] is not None) and (kwargs["loss"] == "jaccard"):
+            warnings.warn(
+                "Warning ignore_index has no effect on training when"
+                + " loss='jaccard'",
```
adamjstewart, line 99:
```suggestion
                "ignore_index has no effect on training when loss='jaccard'",
```
The warning will already start with "UserWarning: ..." so no need to warn twice.

adamjstewart (changes requested):

adamjstewart (approved):

PR #648, opened 2022-07-02
protobuf updated

prasadjivane:
protobuf and timm updated

### DIFF
--- a/requirements/required.txt
+++ b/requirements/required.txt
@@ -15,7 +15,7 @@ numpy==1.21.6;python_version=='3.7'
 omegaconf==2.2.2
 packaging==21.3
 pillow==9.2.0
-protobuf==3.20.1
+protobuf==4.21.2
 pyproj==3.3.1;python_version>='3.8'
 pyproj==3.2.0;python_version=='3.7'
 pytorch-lightning==1.6.4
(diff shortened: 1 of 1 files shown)


adamjstewart:
PyTorch does not yet support protobuf 4 (see https://github.com/protocolbuffers/protobuf/issues/10051, https://github.com/pytorch/pytorch/issues/78362, https://github.com/pytorch/pytorch/commit/6b4ffa14df70e3bd93e3e77ed2690443c0da660d). 

segmentation_models_pytorch requires timm 0.4.12 exactly (see https://github.com/qubvel/segmentation_models.pytorch/blob/master/requirements.txt). So we can't upgrade to a newer version yet.

In general, `requirements/*.txt` are automatically updated by dependabot, you shouldn't need to manually update them. See the comment at the top of the file. The reason that these deps aren't being automatically updated is that we explicitly disable this in `.github/dependabot.yml`.

adamjstewart:
Correct, TorchGeo doesn't directly depend on protobuf so it isn't listed. And segmentation_models_pytorch isn't compatible with timm 0.5, so we can't update. If you need timm 0.5 support you'll have to talk to the segmentatoin_models_pytorch developers about updating their dependency.

kushan-gunasekera:
For anyone who is looking for a proper solution,

answer:- https://stackoverflow.com/a/72493690/6194097

PR #657, opened 2022-07-07
Datamodule cleanup

calebrob6:
Cleaning up how preprocessing in the datamodules works in preparation for writing a prediction utility. 

Now, each DataModule has a `preprocess(self, sample)` method that performs the image preprocessing and doesn't break if there isn't a "mask" or "label" key in the sample. This is important as `predict.py` needs to be able to apply the same image pre-processing that a trained model checkpoint used while training, however the new data that the model is being run over will definitely not have any "mask" or "label" parts.

I also took this opportunity to clean up a few things throughout the DataModules -- I'll go through and explain these changes inline with comments.

### DIFF
--- a/conf/etci2021.yaml
+++ b/conf/etci2021.yaml
@@ -7,7 +7,7 @@ experiment:
     encoder_weights: "imagenet"
     learning_rate: 1e-3
     learning_rate_schedule_patience: 6
-    in_channels: 7
+    in_channels: 6
     num_classes: 2
     ignore_index: 0
   datamodule:
--- a/torchgeo/datamodules/chesapeake.py
+++ b/torchgeo/datamodules/chesapeake.py
@@ -187,6 +187,18 @@ class ChesapeakeCVPRDataModule(LightningDataModule):
 
         return sample
 
+    def remove_bbox(self, sample: Dict[str, Any]) -> Dict[str, Any]:
+        """Removes the bounding box property from a sample.
+
+        Args:
+            sample: dictionary with geographic metadata
+
+        Returns
+            sample without the bbox property
+        """
+        del sample["bbox"]
+        return sample
+
     def nodata_check(
         self, size: int = 512
     ) -> Callable[[Dict[str, Tensor]], Dict[str, Tensor]]:
@@ -238,6 +250,7 @@ class ChesapeakeCVPRDataModule(LightningDataModule):
                 self.center_crop(self.patch_size),
                 self.nodata_check(self.patch_size),
                 self.preprocess,
+                self.remove_bbox,
             ]
         )
         val_transforms = Compose(
@@ -245,12 +258,14 @@ class ChesapeakeCVPRDataModule(LightningDataModule):
                 self.center_crop(self.patch_size),
                 self.nodata_check(self.patch_size),
                 self.preprocess,
+                self.remove_bbox,
             ]
         )
         test_transforms = Compose(
             [
                 self.pad_to(self.original_patch_size, image_value=0, mask_value=0),
                 self.preprocess,
+                self.remove_bbox,
             ]
         )
 
--- a/torchgeo/datamodules/chesapeake.py
+++ b/torchgeo/datamodules/chesapeake.py
@@ -169,23 +169,21 @@ class ChesapeakeCVPRDataModule(LightningDataModule):
         Returns:
             preprocessed sample
         """
-        sample["image"] = sample["image"] / 255.0
-        sample["mask"] = sample["mask"].squeeze()
-
-        if self.use_prior_labels:
-            sample["mask"] = F.normalize(sample["mask"].float(), p=1, dim=0)
-            sample["mask"] = F.normalize(
-                sample["mask"] + self.prior_smoothing_constant, p=1, dim=0
-            )
-        else:
-            if self.class_set == 5:
-                sample["mask"][sample["mask"] == 5] = 4
-                sample["mask"][sample["mask"] == 6] = 4
-            sample["mask"] = sample["mask"].long()
-
         sample["image"] = sample["image"].float()
+        sample["image"] /= 255.0
 
-        del sample["bbox"]
+        if "mask" in sample:
+            sample["mask"] = sample["mask"].squeeze()
+            if self.use_prior_labels:
+                sample["mask"] = F.normalize(sample["mask"].float(), p=1, dim=0)
+                sample["mask"] = F.normalize(
+                    sample["mask"] + self.prior_smoothing_constant, p=1, dim=0
+                )
+            else:
+                if self.class_set == 5:
+                    sample["mask"][sample["mask"] == 5] = 4
+                    sample["mask"][sample["mask"] == 6] = 4
+                sample["mask"] = sample["mask"].long()
 
         return sample
 
--- a/torchgeo/datamodules/cowc.py
+++ b/torchgeo/datamodules/cowc.py
@@ -42,7 +42,7 @@ class COWCCountingDataModule(pl.LightningDataModule):
         self.batch_size = batch_size
         self.num_workers = num_workers
 
-    def custom_transform(self, sample: Dict[str, Any]) -> Dict[str, Any]:
+    def preprocess(self, sample: Dict[str, Any]) -> Dict[str, Any]:
         """Transform a single sample from the Dataset.
 
         Args:
@@ -51,8 +51,10 @@ class COWCCountingDataModule(pl.LightningDataModule):
         Returns:
             preprocessed sample
         """
-        sample["image"] = sample["image"] / 255.0  # scale to [0, 1]
-        sample["label"] = sample["label"].float()
+        sample["image"] = sample["image"].float()
+        sample["image"] /= 255.0  # scale to [0, 1]
+        if "label" in sample:
+            sample["label"] = sample["label"].float()
         return sample
 
     def prepare_data(self) -> None:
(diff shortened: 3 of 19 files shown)


conf/etci2021.yaml
```diff
@@ -7,7 +7,7 @@ experiment:
     encoder_weights: "imagenet"
     learning_rate: 1e-3
     learning_rate_schedule_patience: 6
-    in_channels: 7
+    in_channels: 6
```
calebrob6, line 10: Previously we had 7 channel inputs because we concatenated a given water mask with the image inputs. Generating predictions with a water mask would be tricky at best / outside of the scope of torchgeo so I removed this.

torchgeo/datamodules/chesapeake.py
```diff
@@ -169,24 +169,34 @@ def preprocess(self, sample: Dict[str, Any]) -> Dict[str, Any]:
         Returns:
             preprocessed sample
         """
-        sample["image"] = sample["image"] / 255.0
```
calebrob6, line 172: I changed all `sample["image"] = sample["image"] / 255.0` to `sample["image"] /= 255.0` as the inline operation shouldn't make any copies.

torchgeo/datamodules/chesapeake.py
```diff
@@ -169,24 +169,34 @@ def preprocess(self, sample: Dict[str, Any]) -> Dict[str, Any]:
...
-        sample["mask"] = sample["mask"].squeeze()
+        sample["image"] = sample["image"].float()
+        sample["image"] /= 255.0
+
+        if "mask" in sample:
```
calebrob6, line 175: Added this type of check in order to be able to re-use the `preprocess(...)` logic in cases where "masks"/"labels" aren't available.

torchgeo/datamodules/chesapeake.py
```diff
@@ -169,24 +169,34 @@ def preprocess(self, sample: Dict[str, Any]) -> Dict[str, Any]:
...
-            )
-        else:
-            if self.class_set == 5:
-                sample["mask"][sample["mask"] == 5] = 4
-                sample["mask"][sample["mask"] == 6] = 4
-            sample["mask"] = sample["mask"].long()
+        return sample
 
-        sample["image"] = sample["image"].float()
+    def remove_bbox(self, sample: Dict[str, Any]) -> Dict[str, Any]:
```
calebrob6, line 190: Removing the "bbox" key is required for using the samples in the lightning trainer, however we need the "bbox" for predictions so I decoupled this logic from preprocess.

torchgeo/datamodules/cowc.py
```diff
@@ -42,7 +42,7 @@ def __init__(
         self.batch_size = batch_size
         self.num_workers = num_workers
 
-    def custom_transform(self, sample: Dict[str, Any]) -> Dict[str, Any]:
```
calebrob6, line 45: Renamed all image preprocessing to `preprocess(...)`

torchgeo/datamodules/eurosat.py
```diff
@@ -9,7 +9,7 @@
 import pytorch_lightning as pl
 import torch
 from torch.utils.data import DataLoader
-from torchvision.transforms import Compose, Normalize
```
calebrob6, line 12: Removed Normalize as it is more explicit to subtract band means then divide by band stdevs.

torchgeo/datamodules/so2sat.py
```diff
@@ -25,14 +25,6 @@ class So2SatDataModule(pl.LightningDataModule):
 
     band_means = torch.tensor(
         [
-            -3.591224256609313e-05,
```
calebrob6, line 28: Our previous implementation here didn't really make sense. We were loading all band data, doing channel-wise standardization over all bands, then throwing away the S1 bands so I removed loading of the S1 bands.

torchgeo/datamodules/ucmerced.py
```diff
@@ -25,10 +24,6 @@ class UCMercedDataModule(pl.LightningDataModule):
     Uses random train/val/test splits.
     """
 
-    band_means = torch.tensor([0, 0, 0])
```
calebrob6, line 28: Why

torchgeo/datamodules/usavars.py
```diff
@@ -24,7 +23,6 @@ def __init__(
         self,
         root_dir: str,
         labels: Sequence[str] = USAVars.ALL_LABELS,
-        transforms: Optional[Callable[[Dict[str, Tensor]], Dict[str, Tensor]]] = None,
```
calebrob6, line 27: We don't allow transforms to be passed in to any other DataModules

adamjstewart: This is what we call a slippery slope: https://github.com/microsoft/torchgeo/pull/498#discussion_r843194575

Also see https://github.com/microsoft/torchgeo/discussions/596

torchgeo/datamodules/eurosat.py
adamjstewart: Not sure if I agree with this. I think the `Normalize` transform is more explicit since it tells you why you're doing it, not just what you're doing.

torchgeo/datamodules/usavars.py
```diff
@@ -33,18 +31,28 @@ def __init__(
...
+        """Transform a single sample from the Dataset.
+
+        Args:
+            sample: dictionary containing image
+
+        Returns:
+            preprocessed sample
+        """
+        sample["image"] = sample["image"].float()
+        sample["image"] /= 255.0
```
adamjstewart, line 53: It seems like we're somewhat arbitrarily deciding whether to normalize to [0, 1] or [-1, 1]. Not sure if there is any advantage to one or the other.

conf/etci2021.yaml
adamjstewart: Wouldn't predictions be out_channels, not in_channels?

torchgeo/datamodules/chesapeake.py
adamjstewart: Do you remember why we had to remove bbox again? I feel like there was an option to keep it if we did something else differently...

conf/etci2021.yaml
calebrob6: Yes, but the number of in_channels changed from 7 to 6 because we were previously using a pre-computed water mask as a channel.

torchgeo/datamodules/chesapeake.py
calebrob6: I'm suspicious that it has something to do with collation, but the trainers/test definitely break if you don't remove it.

torchgeo/datamodules/eurosat.py
calebrob6: True, I was thinking of it as, "there are many ways to 'normalize' something where `Normalize` happens to apply channel wise standardization". I don't have a strong opinion here.

torchgeo/datamodules/usavars.py
calebrob6: Yep!

torchgeo/datamodules/usavars.py
calebrob6: Haha yeah I know; I can't make up my mind here. If we have this `Optional[Callable]` as an argument, then we can't (easily) initialize a class instance from a configuration file. This will make general-purpose scripts like `predict.py` more difficult to write. On the other hand, our datamodules aren't very useful for experimentation as-is, because good train-time augmentations are a very important.

torchgeo/datamodules/usavars.py
calebrob6: I wouldn't mind arbitrarily being consistent and do [0,1] normalization for everything.

torchgeo/datamodules/eurosat.py
adamjstewart: I don't have a strong preference here either. I tend to use builtin library functions where possible since I trust them more than I trust my own coding, but this one is obviously very simple. I think the biggest advantage of using `Normalize` is that it does the `.reshape(-1, 1, 1)` here for you, so that's one less thing to worry about. It also uses `.view(-1, 1, 1)` which I think is more efficient. It also accepts a list, so you don't need to convert it to a tensor yourself. Personally I'm gonna vote for using `Normalize` to keep the code as simple as possible.

torchgeo/datamodules/usavars.py
adamjstewart: Can't we have a set of default transforms associated with each datamodule? Then users can override them in Python if they need to via this optional callable (experimentation) but we'll still have config files without customization that default to the optimal known transforms (reproduction).

torchgeo/datamodules/usavars.py
adamjstewart: I'm not sure if this needs to be solved in this PR, just something to think about. We should probably try both and see if it makes any difference. It may be different on a case-by-case basis.

torchgeo/datamodules/usavars.py
calebrob6: Yeah I think that's the best of both worlds. I don't think an "optimal" known transform exists across these datasets -- but a sane default implemented consistently would be a good step. Further, as long as these are optional train transforms, instantiating the DataModule without them shouldn't have any effect on using the DataModule for pre-processing purposes.

calebrob6:
Okay, so I'm going to switch to using Normalize consistently then leave further changes for another PR

adamjstewart (dismissed):
Technically, a lot of these changes (transforms -> preprocess) are API changes that should be documented with `.. versionchanged:: 0.3` but I doubt too many people are calling these directly or overriding them so I won't ask for documentation of these changes.

Inria adds an `augmentations` parameter. Do we want to remove this now for consistency or keep it and think about how to handle this later? Inria was added in this release so its API hasn't been public so we can remove it without documenting the change if we do it now.

calebrob6:
> Inria adds an augmentations parameter. Do we want to remove this now for consistency or keep it and think about how to handle this later? Inria was added in this release so its API hasn't been public so we can remove it without documenting the change if we do it now.

I think remove it

adamjstewart (approved):

PR #661, opened 2022-07-10
Landsat: add plot method

adamjstewart:
We removed the `RasterDataset.plot` method in #476. Landsat was the only remaining `RasterDataset` that didn't have a replacement `plot` method. Added it so we don't have to explain why Landsat used to have a `plot` method but no longer does.

Also cleaned up some typing and documentation mistakes. `RasterDataset.__getitem__` returns `Dict[str, Any]`, not `Dict[str, Tensor]`. Plot methods weren't _added_ in 0.3 since they existed (albeit in the base class) in 0.2. They were _changed_ because they now have completely different inputs.

Example image from Novaya Zemlya, Russia:
landsat

### DIFF
--- a/docs/api/geo_datasets.csv
+++ b/docs/api/geo_datasets.csv
@@ -12,7 +12,7 @@ Dataset,Type,Source,Size (px),Resolution (m)
 `GBIF`_,Points,Citizen Scientists,-,-
 `GlobBiomass`_,Masks,Landsat,"45,000x45,000",100
 `iNaturalist`_,Points,Citizen Scientists,-,-
-`Landsat`_,Imagery,Landsat,-,30
+`Landsat`_,Imagery,Landsat,"8,900x8,900",30
 `NAIP`_,Imagery,Aerial,"6,100x7,600",1
 `Open Buildings`_,Geometries,"Maxar, CNES/Airbus",-,-
 `Sentinel`_,Imagery,Sentinel,"10,000x10,000",10
--- a/tests/datasets/test_landsat.py
+++ b/tests/datasets/test_landsat.py
@@ -4,6 +4,7 @@
 import os
 from pathlib import Path
 
+import matplotlib.pyplot as plt
 import pytest
 import torch
 import torch.nn as nn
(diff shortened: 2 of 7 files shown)


calebrob6:
How are you generating that actually? I'm getting problems:

```
from torchgeo.datasets import Landsat8
import numpy as np

ds = Landsat8(root="tmp/", bands=Landsat8.all_bands)
sampler = RandomGeoSampler(ds, size=50, length=1)
bb = next(iter(sampler))
ds[bb]
```

Is giving me the following error:
```
IndexError                                Traceback (most recent call last)
/tmp/ipykernel_32350/2392223127.py in <module>
----> 1 ds[bb]

~/code/torchgeo/torchgeo/datasets/geo.py in __getitem__(self, query)
    417                             end = match.end("resolution")
    418                             filename = filename[:start] + "*" + filename[end:]
--> 419                     filepath = glob.glob(os.path.join(directory, filename))[0]
    420                     band_filepaths.append(filepath)
    421                 data_list.append(self._merge_files(band_filepaths, query))

IndexError: list index out of range
```

Where:
```
> ls tmp
LC08_L2SP_013033_20190830_20200826_02_T1_SR_B1.TIF
LC08_L2SP_013033_20190830_20200826_02_T1_SR_B2.TIF
LC08_L2SP_013033_20190830_20200826_02_T1_SR_B3.TIF
LC08_L2SP_013033_20190830_20200826_02_T1_SR_B4.TIF
LC08_L2SP_013033_20190830_20200826_02_T1_SR_B5.TIF
LC08_L2SP_013033_20190830_20200826_02_T1_SR_B6.TIF
LC08_L2SP_013033_20190830_20200826_02_T1_SR_B7.TIF
```

calebrob6:
Ahhh nvrmnd I didn't have all the bands downloaded but it was defaulting to all bands.

calebrob6 (approved):

calebrob6:
Stretching to full range is okay. When the image patch doesn't have a wide range it looks overexposed, but for most cases it will be fine!

E.g. water only patch
image

What happens when you divide by constant and clip:
image

PR #663, opened 2022-07-10
Replace open3d with pyvista

isaaccorley:
Depends on https://github.com/pyvista/pyvista/discussions/3692

Refactors point cloud plotting in the `torchgeo.datasets.IDTReeS` dataset to use `pyvista` instead of `open3d`

Note: Currently figuring out the min version which passes tests.

See #662 

Example:

https://user-images.githubusercontent.com/22203655/178155777-dac2e462-d978-4005-9412-3ffd91a10bcb.mov

Closes #932


torchgeo/datasets/idtrees.py
```diff
@@ -549,10 +550,10 @@ def plot_las(self, index: int, colormap: Optional[str] = None) -> Any:
...
-        vis = open3d.visualization.Visualizer()
-        vis.create_window()
-        vis.add_geometry(pcd)
-        return vis
+        point_cloud = pyvista.PolyData(points)  # type: ignore[attr-defined]
+        point_cloud["color"] = colors
+        plotter = pyvista.Plotter()  # type: ignore[attr-defined]
+        _ = plotter.add_mesh(point_cloud, scalars="color", rgb=True)
+        plotter.camera_position = "yz"
+        plotter.background_color = "white"
```
isaaccorley, line 558: Users can still modify the plotter object. This simply plots the interactive colored point cloud on a white background.

isaaccorley:
Need to figure out how to get the plotting to run headless within the CI container. That seems to be what's causing tests to fail.

adamjstewart:
Did you ever open an issue with the developers? Can you link to it here?

.github/workflows/tests.yaml
```diff
@@ -45,7 +45,7 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install apt dependencies (Linux)
-      run: sudo apt-get install unrar
+      run: sudo apt-get install unrar libgl1-mesa-glx xvfb
```
banesullivan, line 48: Let's remove this apt install and leverage PyVista's headless display action to set up everything appropriately on each OS

.github/workflows/tests.yaml
```diff
@@ -45,7 +45,7 @@ jobs:
...
       if: ${{ runner.os == 'Linux' }}
```
banesullivan, line 49:
```suggestion
      if: ${{ runner.os == 'Linux' }}
    - name: Setup headless display
      uses: pyvista/setup-headless-display-action@v1
```

tests/datasets/test_idtrees.py
```diff
@@ -141,14 +141,19 @@ def test_plot(self, dataset: IDTReeS) -> None:
             plt.close()
 
     @pytest.mark.skipif(
-        sys.platform in ["darwin", "win32"],
-        reason="segmentation fault on macOS and windows",
+        sys.platform != "linux",
+        reason="Pyvista tests require a virtual frame buffer only supported on linux",
```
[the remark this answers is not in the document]
banesullivan: The added action in my suggestion above should work on all platforms.

torchgeo/datasets/idtrees.py
```diff
@@ -567,25 +567,26 @@ def normalize(x: Tensor) -> Tensor:
 
         return fig
 
-    def plot_las(self, index: int, colormap: Optional[str] = None) -> Any:
+    def plot_las(
+        self, index: int, colormap: Optional[str] = None
+    ) -> "pyvista.Plotter":  # type: ignore[name-defined] # noqa: F821
```
banesullivan, line 572: I would add some `**kwargs` here so that users can control the `Plotter` init params and the `add_mesh` params.

torchgeo/datasets/idtrees.py
banesullivan: I recommend not setting this background and letting PyVista's theme determine the background. Users can easily control this with `pv.set_plot_theme('document')` (before creating a plotter). We may change the default PyVista theme soon here as grey backgrounds aren't the best IMO

torchgeo/datasets/idtrees.py
```diff
@@ -612,10 +613,10 @@ def plot_las(self, index: int, colormap: Optional[str] = None) -> Any:
...
-        pcd.points = open3d.utility.Vector3dVector(points)
-        pcd.colors = open3d.utility.Vector3dVector(colors)
-        vis = open3d.visualization.Visualizer()
-        vis.create_window()
-        vis.add_geometry(pcd)
-        return vis
+        point_cloud = pyvista.PolyData(points)  # type: ignore[attr-defined]
+        point_cloud["color"] = colors
+        plotter = pyvista.Plotter()  # type: ignore[attr-defined]
+        _ = plotter.add_mesh(point_cloud, scalars="color", rgb=True)
```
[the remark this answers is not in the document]
banesullivan: I recommend changing this section to PyVista's `plot()` API instead. That way, you can easily capture user-passed `kwargs` and pass them to PyVista to let users easily change the display parameters (e.g., `point_size`) and/or change Jupyter plotting backends, rendering techniques like eye-dome lighting, etc.

This would look like:

```py
return point_cloud.plot(scalars="color", rgb=True, **kwargs)
```

torchgeo/datasets/idtrees.py
```diff
@@ -612,10 +613,10 @@ def plot_las(self, index: int, colormap: Optional[str] = None) -> Any:
...
 
-        pcd = open3d.geometry.PointCloud()
-        pcd.points = open3d.utility.Vector3dVector(points)
-        pcd.colors = open3d.utility.Vector3dVector(colors)
-        vis = open3d.visualization.Visualizer()
-        vis.create_window()
-        vis.add_geometry(pcd)
-        return vis
+        point_cloud = pyvista.PolyData(points)  # type: ignore[attr-defined]
+        point_cloud["color"] = colors
```
banesullivan, line 617: I'd recommend a `to_pyvista()` method to let users easily get a PyVista PolyData object without plotting for them

torchgeo/datasets/idtrees.py
banesullivan: If switching to the `.plot()` API, these can all be captured and `**kwargs` and passed along

tests/datasets/test_idtrees.py
```diff
@@ -141,14 +141,19 @@ def test_plot(self, dataset: IDTReeS) -> None:
...
-        vis.close()
-        vis = dataset.plot_las(index=0, colormap=None)
-        vis.close()
-        vis = dataset.plot_las(index=1, colormap=None)
-        vis.close()
+        pytest.importorskip("pyvista", minversion="0.35.1")
+        import pyvista
+
+        pyvista.OFF_SCREEN = True
+        pyvista.start_xvfb(wait=0)
```
banesullivan, line 152: If using the GitHub action `start_xvfb` shouldn't be needed

banesullivan (commented):
Stoked to see you all switching to PyVista! Here is some feedback and hopefully my pointers will get the tests working/passing for each platform

.github/workflows/tests.yaml
adamjstewart: Let's actually put it before the "Install apt dependencies" so all actions are at the beginning and all custom steps are below.

isaaccorley:
A couple things:
- I decided to setup and return the point_cloud PolyData object. Users will be able to run `pyvista.plot(point_cloud, ...)` themselves to render everything how they see fit. I think this is better because it doesn't assume the other parameters a user wants to define. Let me know if anyone has any objections to this.
- I was able to remove the `colormap` arg entirely because `pyvista.plot` supports a `cmap` arg which effectively does the same thing. So pyvista is already helping clean up the code 👍

tests/datasets/test_idtrees.py
```diff
@@ -140,15 +139,12 @@ def test_plot(self, dataset: IDTReeS) -> None:
...
     def test_plot_las(self, dataset: IDTReeS) -> None:
-        pytest.importorskip("open3d", minversion="0.11.2")
-        vis = dataset.plot_las(index=0, colormap="BrBG")
-        vis.close()
-        vis = dataset.plot_las(index=0, colormap=None)
-        vis.close()
-        vis = dataset.plot_las(index=1, colormap=None)
-        vis.close()
+        pytest.importorskip("pyvista", minversion="0.35.1")
+        import pyvista
```
adamjstewart, line 144:
```suggestion
        pyvista = pytest.importorskip("pyvista", minversion="0.35.1")
```
The version should be updated to the minimum supported version

tests/datasets/test_idtrees.py
```diff
@@ -140,15 +139,12 @@ def test_plot(self, dataset: IDTReeS) -> None:
...
+
+        # Test point cloud without colors
+        point_cloud = dataset.plot_las(index=0)
+        pyvista.plot(  # type: ignore[attr-defined]
```
adamjstewart, line 148: @banesullivan I assume type hints for pyvista are a work in progress? If you point me to where this function is defined I would be happy to add some.

torchgeo/datasets/idtrees.py
```diff
@@ -567,25 +566,25 @@ def normalize(x: Tensor) -> Tensor:
 
         return fig
 
-    def plot_las(self, index: int, colormap: Optional[str] = None) -> Any:
+    def plot_las(
+        self, index: int
+    ) -> "pyvista.Plotter":  # type: ignore[name-defined] # noqa: F821
```
adamjstewart, line 571: What was the deal with these ignores?

torchgeo/datasets/idtrees.py
```diff
@@ -595,27 +594,12 @@ def plot_las(self, index: int, colormap: Optional[str] = None) -> Any:
         points: "np.typing.NDArray[np.int_]" = np.stack(
             [las.x, las.y, las.z], axis=0
         ).transpose((1, 0))
+        point_cloud = pyvista.PolyData(points)  # type: ignore[attr-defined]
```
adamjstewart, line 597: Same question here, would be happy to add type hints for this if needed

setup.cfg
```diff
@@ -78,15 +78,14 @@ datasets =
...
-    # https://github.com/isl-org/Open3D/issues/1550
-    open3d>=0.11.2,<0.15;python_version<'3.10'
     # opencv-python 3.4.2.17 is oldest buildable version on PyPI
     opencv-python>=3.4.2.17,<5
     # pandas 0.23.2+ required for Python 3.7 wheels
     pandas>=0.23.2,<2
     # pycocotools 2.0.0 is oldest version on PyPI
     pycocotools>=2,<3
+    # pyvista required for point cloud plotting
+    pyvista
```
adamjstewart, line 88: Last step is to determine the minimum supported version of pyvista. @banesullivan any guess as to the minimum version of pyvista that provides the features we use? If not we'll just do a binary search on all versions until we find where the tests start to fail.

torchgeo/datasets/idtrees.py
```diff
@@ -595,27 +594,12 @@ def plot_las(self, index: int, colormap: Optional[str] = None) -> Any:
...
-        pcd.points = open3d.utility.Vector3dVector(points)
-        pcd.colors = open3d.utility.Vector3dVector(colors)
-        vis = open3d.visualization.Visualizer()
-        vis.create_window()
-        vis.add_geometry(pcd)
-        return vis
+        # Some point cloud files have no color->points mapping
+        if hasattr(las, "red"):
+            colors = np.stack([las.red, las.green, las.blue], axis=0)
+            colors = colors.transpose((1, 0)) / 65535
```
adamjstewart, line 602: I know this code was already here, but where does 65535 come from?

adamjstewart (changes requested):

torchgeo/datasets/idtrees.py
isaaccorley: the RGB values per point for whatever reason are uint16 so they need to be normalized to the range [0, 1] so they can be used as a colormap. (I didn't find any documentation why they are between [0, 65535]

torchgeo/datasets/idtrees.py
isaaccorley: The ignores are because we define pyvista inside the `plot_las` method so mypy and flake8 are complaining that pyvista is not defined

torchgeo/datasets/idtrees.py
isaaccorley: Even if I import `from pyvista import PolyData` and change the line to `point_cloud: PolyData = PolyData(points)` mypy still complains `Module "pyvista" has no attribute "PolyData"  [attr-defined]`

setup.cfg
isaaccorley: I get segmentation faults even when using xvfb when these tests run so I'll need to move over to my mac and run them. This should be fun

banesullivan: > any guess as to the minimum version of pyvista that provides the features we use?

Hm... do you *need* to provide a min version? The features used here have been around for quite some time. I'd recommend not setting the min version as far back as you can, but arbitrarily choose a more recent version for stability. Again, IMO, I would leave the version unconstrained in the `setup.cfg` but if you must, I'd target v0.33 or v0.34 (we dropped Python3.6 support in v0.34)

tests/datasets/test_idtrees.py
banesullivan: Indeed a work in progress though we do have partial: https://github.com/pyvista/pyvista/blob/main/pyvista/py.typed

setup.cfg
adamjstewart: I would like to provide a minimum version just to be explicit. For some packages like OpenCV it's the minimum version released on PyPI because all versions worked. For others it's the minimum version that supports Python 3.7 (the oldest we support). The goal is that the widest possible range is supported. We have unit tests that run our unit tests with the oldest supported version so we can ensure that it is correct and users are unlikely to encounter an invalid environment.

banesullivan: The real versioning issues might come from VTK... You'll need to be careful if installing an older version of PyVista with newer versions of VTK. More recent versions of PyVista do a better job of targeting all versions of VTK whereas older versions will have conflicts on import with a recent version of VTK. 

This can get quite complicated quite quickly.

I'd recommend choosing a recent min version of PyVista (e.g., 0.34) and limiting VTK to 9.x.x (though it is a really, really bad idea to set a version constraint on VTK in the setup.cfg)

isaaccorley: > > I get segmentation faults even when using xvfb when these tests run so I'll need to move over to my mac and run them. This should be fun
> 
> Are you doing this locally? You only need xvfb in headless environments

I'm working out of an ec2 instance so I was trying to it out

torchgeo/datasets/idtrees.py
adamjstewart: In that case I would suggest:
```suggestion
            colors = colors.transpose((1, 0)) / np.iinfo(np.uint16).max
```
so that there aren't any magic numbers.

setup.cfg
adamjstewart: Probably quicker to do a binary search on versions using GitHub Actions than it is to test locally or on ec2.

We probably won't be able to control which version of VTK gets installed since it's a non-Python dependency. I say we go with the oldest version of PyVista that works in CI without messing with any system packages. That's what we do for all other dependencies.

banesullivan: >We probably won't be able to control which version of VTK gets installed since it's a non-Python dependency

Nope, it's a Python dep! Its wheels are on PyPI, and you can easily control what version get's installed with pip. Pinning in the setup.cfg can be a bad idea because a lot of folks use VTK from different sources (e.g. ParaView, built from source, etc and the version may not resolve)

adamjstewart: Gotcha. We probably won't pin vtk in `setup.cfg`, although we may pin in `requirements/min.old` since that's what's used in CI.

adamjstewart (approved):

tests/datasets/test_idtrees.py
```diff
@@ -140,15 +139,9 @@ def test_plot(self, dataset: IDTReeS) -> None:
...
-    )
     def test_plot_las(self, dataset: IDTReeS) -> None:
-        pytest.importorskip("open3d", minversion="0.11.2")
-        vis = dataset.plot_las(index=0, colormap="BrBG")
-        vis.close()
-        vis = dataset.plot_las(index=0, colormap=None)
-        vis.close()
-        vis = dataset.plot_las(index=1, colormap=None)
-        vis.close()
+        pyvista = pytest.importorskip("pyvista", minversion="0.35.1")
```
adamjstewart, line 143: @isaaccorley do you remember why our testing minversion is different from our setup.cfg min version? I tried correcting this in #1276 but it's failing. Is 0.35.1 the real min version that works? I don't want to claim to support versions we can't test.

isaaccorley: Might be an oversight on my part. I don't remember the reason exactly. I recall there was something to do with certain wheels for different OS not being available maybe?

adamjstewart: Wheels are pure Python, but could be limited by VTK. I'm going to try to find the oldest version for which our tests actually pass. We haven't been testing pyvista at all in our minimum version CI.

banesullivan: >  but could be limited by VTK

Most likely the case

> I'm going to try to find the oldest version for which our tests actually pass.

I don't recommend this (see above comment https://github.com/microsoft/torchgeo/pull/663#discussion_r1043542639). Once you go back to older versions of PyVista (or let users do this), you'll need to limit VTK carefully. We work tirelessly to make sure the latest PyVista is compatible with all of VTK 9+, but older versions of PyVista may outright fail on import with newer versions of VTK. I recommend choosing a fairly recent version of PyVista as the minimum for this reason.

adamjstewart: Managed to get PyVista 0.29 working fine without having to mess with VTK: #1276

PR #675, opened 2022-07-14
Fixes VectorDataset rounding bug causing sample mask size mismatch

### LINKED ISSUE #674: VectorDataset incorrect sample shape
### Description

Querying a sample from a VectorDataset object returns (sometimes?) a mask with incorrect shape.

This is caused by not rounding the calculated width and height pixel values to the nearest integer, and just taking the `int(width)` and `int(height)` values instead for `masks = rasterio.features.rasterize` in [torchgeo.datasets.geo.py:614](https://github.com/microsoft/torchgeo/blob/b41b6d68f99edfbe4d82b31f6f9b16cbd33f485e/torchgeo/datasets/geo.py#L614).

### Steps to reproduce

Run the following code to reproduce the error:
```
from torchgeo.datasets import CanadianBuildingFootprints

cbf_dataset = CanadianBuildingFootprints(download=True)  # or any other VectorDataset-based dataset

size = 512

sampler = RandomGeoSampler(cbf_dataset, size=size, length=100)  # or any other sampler
query = next(iter(sampler))

sample = cbf_dataset[query]

print(sample["mask"].shape)
assert sample["mask"].shape[0] == size, sample["mask"].shape
assert sample["mask"].shape[1] == size, sample["mask"].shape
```

### Version

0.3.0

TCherici:
I've had to adjust the test to also account for the extra pixel generated, which seems right to me (`1.9-1.1 = 0.8` which would be 8 pixels with a `res=0.1`).

Fixes #674

### DIFF
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -611,12 +611,12 @@ class VectorDataset(GeoDataset):
         )
         if shapes:
             masks = rasterio.features.rasterize(
-                shapes, out_shape=(int(height), int(width)), transform=transform
+                shapes, out_shape=(int(round(height)), int(round(width))), transform=transform
             )
         else:
             # If no features are found in this query, return an empty mask
             # with the default fill value and dtype used by rasterize
-            masks = np.zeros((int(height), int(width)), dtype=np.uint8)
+            masks = np.zeros((int(round(height)), int(round(width))), dtype=np.uint8)
 
         sample = {"mask": torch.tensor(masks), "crs": self.crs, "bbox": query}
 
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -611,12 +611,12 @@ class VectorDataset(GeoDataset):
         )
         if shapes:
             masks = rasterio.features.rasterize(
-                shapes, out_shape=(int(round(height)), int(round(width))), transform=transform
+                shapes, out_shape=(round(height), round(width)), transform=transform
             )
         else:
             # If no features are found in this query, return an empty mask
             # with the default fill value and dtype used by rasterize
-            masks = np.zeros((int(round(height)), int(round(width))), dtype=np.uint8)
+            masks = np.zeros((round(height), round(width)), dtype=np.uint8)
 
         sample = {"mask": torch.tensor(masks), "crs": self.crs, "bbox": query}
 
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -223,7 +223,7 @@ class TestVectorDataset:
     def test_empty_shapes(self, dataset: CustomVectorDataset) -> None:
         query = BoundingBox(1.1, 1.9, 1.1, 1.9, 0, 0)
         x = dataset[query]
-        assert torch.equal(x["mask"], torch.zeros(7, 7, dtype=torch.uint8))
+        assert torch.equal(x["mask"], torch.zeros(8, 8, dtype=torch.uint8))
 
     def test_invalid_query(self, dataset: CustomVectorDataset) -> None:
         query = BoundingBox(3, 3, 3, 3, 0, 0)


torchgeo/datasets/geo.py
```diff
@@ -611,12 +611,12 @@ def __getitem__(self, query: BoundingBox) -> Dict[str, Any]:
         )
         if shapes:
             masks = rasterio.features.rasterize(
-                shapes, out_shape=(int(height), int(width)), transform=transform
+                shapes, out_shape=(int(round(height)), int(round(width))), transform=transform
```
adamjstewart, line 614: This line is too long now. Can you run `black torchgeo/datasets/geo.py` to fix this?

adamjstewart (approved):
Can confirm that this fixes the issue, thanks for the PR!

torchgeo/datasets/geo.py
ashnair1: There is need to cast it to `int` again if you're not specifying the number of decimals to round to.

torchgeo/datasets/geo.py
```diff
@@ -611,12 +611,12 @@ def __getitem__(self, query: BoundingBox) -> Dict[str, Any]:
...
             )
         else:
             # If no features are found in this query, return an empty mask
             # with the default fill value and dtype used by rasterize
-            masks = np.zeros((int(height), int(width)), dtype=np.uint8)
+            masks = np.zeros((int(round(height)), int(round(width))), dtype=np.uint8)
```
ashnair1, line 619: Same comment as above. Int casting not required.

ashnair1:
> There are several other places where we aren't rounding before casting to an int. I'm not worried about the str -> int conversions, but some of these float -> int conversions could be problematic. Not sure if integer division includes rounding or not. I wonder if any of these should also be fixed:


I had encountered a related issue with `dataset_split` in `torchgeo/datamodules/utils.py`. Direct int casting basically strips out the decimal part so it's a problem if `len(dataset) * val_pct` < 0. I had encountered this issue while testing coverage for a datamodule. The test dataset in question had `len(dataset)` = 3 and `val_pct` = 0.3. 

Fix using `round`:


```diff
    if test_pct is None:
-        val_length = int(len(dataset) * val_pct)
+        val_length = round(len(dataset) * val_pct)
        train_length = len(dataset) - val_length
        return random_split(dataset, [train_length, val_length])
    else:
-       val_length = int(len(dataset) * val_pct)
-       test_length = int(len(dataset) * test_pct)
+       val_length = round(len(dataset) * val_pct)
+       test_length = round(len(dataset) * test_pct)
```

torchgeo/datasets/geo.py
ashnair1: Unless this was needed to satisfy mypy.

TCherici:
> I had encountered a related issue with dataset_split in torchgeo/datamodules/utils.py. Direct int casting basically strips out the decimal part so it's a problem if len(dataset) * val_pct < 0. I had encountered this issue while testing coverage for a datamodule. The test dataset in question had len(dataset) = 3 and val_pct = 0.3.

I don't understand how `len(dataset) * val_pct` could be less than 0 for the values you gave, and I'm not sure if this is related to the bugfix in this PR. @ashnair1 If you think it is I will add it

ashnair1:
> > I had encountered a related issue with dataset_split in torchgeo/datamodules/utils.py. Direct int casting basically strips out the decimal part so it's a problem if len(dataset) * val_pct < 0. I had encountered this issue while testing coverage for a datamodule. The test dataset in question had len(dataset) = 3 and val_pct = 0.3.
> 
> I don't understand how `len(dataset) * val_pct` could be less than 0 for the values you gave, and I'm not sure if this is related to the bugfix in this PR. @ashnair1 If you think it is I will add the fix you gave as a commit to this PR.

My bad. Meant less than 1. The comment is unrelated to this PR. I was just replying to @adamjstewart's comment on other places this round vs int errors can occur.

adamjstewart (approved):
We can fix other instances of this bug in a separate PR.

PR #677, opened 2022-07-15
Update required.txt

prasadjivane:
VR changes added as per latest changes.


calebrob6:
As far as I can tell, this is spam similar to #648

adamjstewart:
Yep, duplicate of #648

PR #687, opened 2022-07-26
Allow band indexing in RasterDataset

ashnair1:
`RasterDataset` by default loads all bands. This PR allows us to be more selective.

### DIFF
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -156,20 +156,29 @@ class TestGeoDataset:
 
 
 class TestRasterDataset:
-    @pytest.fixture(params=[True, False])
+    @pytest.fixture(params=zip([True, False], [["R", "G", "B"], None]))
     def naip(self, request: SubRequest) -> NAIP:
         root = os.path.join("tests", "data", "naip")
+        bands = request.param[1]
         crs = CRS.from_epsg(3005)
         transforms = nn.Identity()
-        cache = request.param
-        return NAIP(root, crs=crs, transforms=transforms, cache=cache)
+        cache = request.param[0]
+        return NAIP(root, crs=crs, bands=bands, transforms=transforms, cache=cache)
 
-    @pytest.fixture(params=[True, False])
+    @pytest.fixture(
+        params=zip(
+            [True, False],
+            [
+                ["B04", "B03", "B02"],
+                ["B01", "B02", "B03", "B04", "B05", "B06", "B07", "B08", "B09", "B11"],
+            ],
+        )
+    )
     def sentinel(self, request: SubRequest) -> Sentinel2:
         root = os.path.join("tests", "data", "sentinel2")
-        bands = ["B01", "B02", "B03", "B04", "B05", "B06", "B07", "B08", "B09", "B11"]
+        bands = request.param[1]
         transforms = nn.Identity()
-        cache = request.param
+        cache = request.param[0]
         return Sentinel2(root, bands=bands, transforms=transforms, cache=cache)
 
     @pytest.fixture()
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -160,29 +160,29 @@ class TestGeoDataset:
 
 
 class TestRasterDataset:
-    @pytest.fixture(params=zip([True, False], [["R", "G", "B"], None]))
+    @pytest.fixture(params=zip([["R", "G", "B"], None], [True, False]))
     def naip(self, request: SubRequest) -> NAIP:
         root = os.path.join("tests", "data", "naip")
-        bands = request.param[1]
+        bands = request.param[0]
         crs = CRS.from_epsg(3005)
         transforms = nn.Identity()
-        cache = request.param[0]
+        cache = request.param[1]
         return NAIP(root, crs=crs, bands=bands, transforms=transforms, cache=cache)
 
     @pytest.fixture(
         params=zip(
-            [True, False],
             [
                 ["B04", "B03", "B02"],
                 ["B01", "B02", "B03", "B04", "B05", "B06", "B07", "B08", "B09", "B11"],
             ],
+            [True, False],
         )
     )
     def sentinel(self, request: SubRequest) -> Sentinel2:
         root = os.path.join("tests", "data", "sentinel2")
-        bands = request.param[1]
+        bands = request.param[0]
         transforms = nn.Identity()
-        cache = request.param[0]
+        cache = request.param[1]
         return Sentinel2(root, bands=bands, transforms=transforms, cache=cache)
 
     @pytest.fixture()
@@ -195,12 +195,14 @@ class TestRasterDataset:
         assert isinstance(x, dict)
         assert isinstance(x["crs"], CRS)
         assert isinstance(x["image"], torch.Tensor)
+        assert len(naip.bands) == x["image"].shape[0]
 
     def test_getitem_separate_files(self, sentinel: Sentinel2) -> None:
         x = sentinel[sentinel.bounds]
         assert isinstance(x, dict)
         assert isinstance(x["crs"], CRS)
         assert isinstance(x["image"], torch.Tensor)
+        assert len(sentinel.bands) == x["image"].shape[0]
 
     def test_getitem_uint_dtype(self, custom_dtype_ds: RasterDataset) -> None:
         x = custom_dtype_ds[custom_dtype_ds.bounds]
@@ -220,7 +222,11 @@ class TestRasterDataset:
             RasterDataset(str(tmp_path))
 
     def test_no_allbands(self) -> None:
-        with pytest.raises(AssertionError, match="all_bands must be specified"):
+        msg = (
+            "CustomSentinelDataset is missing an `all_bands` attribute,"
+            " so `bands` cannot be specified."
+        )
+        with pytest.raises(AssertionError, match=msg):
             root = os.path.join("tests", "data", "sentinel2")
             bands = ["B04", "B03", "B02"]
             transforms = nn.Identity()
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -222,15 +222,16 @@ class TestRasterDataset:
             RasterDataset(str(tmp_path))
 
     def test_no_allbands(self) -> None:
+        root = os.path.join("tests", "data", "sentinel2")
+        bands = ["B04", "B03", "B02"]
+        transforms = nn.Identity()
+        cache = True
         msg = (
             "CustomSentinelDataset is missing an `all_bands` attribute,"
             " so `bands` cannot be specified."
         )
+
         with pytest.raises(AssertionError, match=msg):
-            root = os.path.join("tests", "data", "sentinel2")
-            bands = ["B04", "B03", "B02"]
-            transforms = nn.Identity()
-            cache = True
             CustomSentinelDataset(root, bands=bands, transforms=transforms, cache=cache)
 
 
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -4,7 +4,7 @@
 import os
 import pickle
 from pathlib import Path
-from typing import Dict
+from typing import Dict, List
 
 import pytest
 import torch
@@ -47,6 +47,10 @@ class CustomGeoDataset(GeoDataset):
 
 class CustomVectorDataset(VectorDataset):
     filename_glob = "*.geojson"
+
+
+class CustomSentinelDataset(Sentinel2):
+    all_bands: List[str] = []
 
 
 class CustomNonGeoDataset(NonGeoDataset):
@@ -214,6 +218,14 @@ class TestRasterDataset:
     def test_no_data(self, tmp_path: Path) -> None:
         with pytest.raises(FileNotFoundError, match="No RasterDataset data was found"):
             RasterDataset(str(tmp_path))
+
+    def test_no_allbands(self) -> None:
+        with pytest.raises(AssertionError, match="all_bands must be specified"):
+            root = os.path.join("tests", "data", "sentinel2")
+            bands = ["B04", "B03", "B02"]
+            transforms = nn.Identity()
+            cache = True
+            CustomSentinelDataset(root, bands=bands, transforms=transforms, cache=cache)
 
 
 class TestVectorDataset:
--- a/torchgeo/datasets/agb_live_woody_density.py
+++ b/torchgeo/datasets/agb_live_woody_density.py
@@ -6,7 +6,7 @@
 import glob
 import json
 import os
-from typing import Any, Callable, Dict, Optional, Sequence
+from typing import Any, Callable, Dict, Optional
 
 import matplotlib.pyplot as plt
 from rasterio.crs import CRS
@@ -83,11 +83,10 @@ class AbovegroundLiveWoodyBiomassDensity(RasterDataset):
         """
         self.root = root
         self.download = download
-        bands: Sequence[str] = []
 
         self._verify()
 
-        super().__init__(root, crs, res, bands, transforms, cache)
+        super().__init__(root, crs, res, transforms=transforms, cache=cache)
 
     def _verify(self) -> None:
         """Verify the integrity of the dataset.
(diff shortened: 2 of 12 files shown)


adamjstewart (changes requested):
There is already a `self.bands` attribute for this. Right now, it's only used when `self.separate_files`, but we should be able to reuse this for the purpose you're using it for here. The only problem is that `self.bands` is the string names, not the integer indices. Maybe we can translate from one to the other using `self.all_bands`?

ashnair1:
I did notice those attributes and don't think they can be reused.

`bands`, `all_bands` and `rgb_bands` are a list of band filenames when bands are stored separately.  Suppose I have a directory of 8 band geotiffs and I want to create a `RasterDataset` only using bands 2, 3 and 4. How could I do that? There are no band names to go off of.

Conceptually I see it as two different things. The current `bands`/`all_bands` attribute helps in assembling a multi band tensor from disparate band files. The `band_indexes` attribute I'm proposing is to provide users a provision to select the bands they require from the assembled multi band tensor.

adamjstewart:
I think it's possible to reuse these. There's no reason that `bands`/`all_bands`/`rgb_bands` should only apply to `separate_files = True`, we also use them in datasets like `NAIP` where `separate_files = False`. If we enforce that the bands in `all_bands` are in the same order as the actual band indices in a single file, then we can 1) check to make sure that `bands` are real bands listed in `all_bands`, and 2) use the ordering of `all_bands` to compute `band_indices` from `bands`. The only thing that would need to change is to add a `bands` parameter to the affected datasets. I actually think it would be worth adding this directly to `RasterDataset` since most of our datasets are multi-band and the ones that aren't can just define a single band.

ashnair1:
I think I understand. Do you mean something like this in the `__init__` method?

```
if self.all_bands and self.bands:
    band_indexes =  [self.all_bands.index(i) + 1 for i in self.bands]
    assert len(band_indexes) == len(self.bands)
else:
    band_indexes = None

self.band_indexes = band_indexes
```

adamjstewart:
Yes, something like that would be half of it. The other half is users need a way of specifying this when they instantiate a dataset (`bands` parameter). Some datasets have this, others don't. I think it's worth adding this to `RasterDataset`, then we can actually get rid of the `__init__` in some subclasses.

ashnair1:
So we have a `bands` parameter that determines our bands of interest and `all_bands` which are all the bands in the dataset.

If we go back to my previous example of creating a `RasterDataset` from a directory of 8 band tiffs using only bands 2, 3, and 4 it should look like this right?


```
class Raster8BandFolder(RasterDataset):

all_bands = [f"B{i}" for i in range(1,9)]

def __init__(
        self,
        root: str = "data",
        crs: Optional[CRS] = None,
        res: Optional[float] = None,
        bands: Sequence[str] = [],
        transforms: Optional[Callable[[Dict[str, Any]], Dict[str, Any]]] = None,
        cache: bool = True,
    ) -> None:

    super().__init__(root, crs, res, bands, transforms, cache)


raster8 = Raster8BandFolder(root="/path/to/folder", bands=["B2", "B3", "B4"])
```

ashnair1:
Modification to `RasterDataset`

```diff
class RasterDataset(GeoDataset):
    """Abstract base class for :class:`GeoDataset` stored as raster files."""

    #: Glob expression used to search for files.
    #:
    #: This expression should be specific enough that it will not pick up files from
    #: other datasets. It should not include a file extension, as the dataset may be in
    #: a different file format than what it was originally downloaded as.
    filename_glob = "*"

    #: Regular expression used to extract date from filename.
    #:
    #: The expression should use named groups. The expression may contain any number of
    #: groups. The following groups are specifically searched for by the base class:
    #:
    #: * ``date``: used to calculate ``mint`` and ``maxt`` for ``index`` insertion
    #:
    #: When :attr:`separate_files`` is True, the following additional groups are
    #: searched for to find other files:
    #:
    #: * ``band``: replaced with requested band name
    #: * ``resolution``: replaced with a glob character
    filename_regex = ".*"

    #: Date format string used to parse date from filename.
    #:
    #: Not used if :attr:`filename_regex` does not contain a ``date`` group.
    date_format = "%Y%m%d"

    #: True if dataset contains imagery, False if dataset contains mask
    is_image = True

    #: True if data is stored in a separate file for each band, else False.
    separate_files = False

    #: Names of all available bands in the dataset
    all_bands: List[str] = []

    #: Names of RGB bands in the dataset, used for plotting
    rgb_bands: List[str] = []

    #: Color map for the dataset, used for plotting
    cmap: Dict[int, Tuple[int, int, int, int]] = {}

    def __init__(
        self,
        root: str,
        crs: Optional[CRS] = None,
        res: Optional[float] = None,
+       bands: List[str] = [],
        transforms: Optional[Callable[[Dict[str, Any]], Dict[str, Any]]] = None,
        cache: bool = True,
    ) -> None:
        """Initialize a new Dataset instance.

        Args:
            root: root directory where dataset can be found
            crs: :term:`coordinate reference system (CRS)` to warp to
                (defaults to the CRS of the first file found)
            res: resolution of the dataset in units of CRS
                (defaults to the resolution of the first file found)
+           bands: list of band names to be used
            transforms: a function/transform that takes an input sample
                and returns a transformed version
            cache: if True, cache file handle to speed up repeated sampling

        Raises:
            FileNotFoundError: if no files are found in ``root``
        """
        super().__init__(transforms)

        self.root = root
        self.cache = cache

        # Populate the dataset index
        i = 0
        pathname = os.path.join(root, "**", self.filename_glob)
        filename_regex = re.compile(self.filename_regex, re.VERBOSE)
        for filepath in glob.iglob(pathname, recursive=True):
            match = re.match(filename_regex, os.path.basename(filepath))
            if match is not None:
                try:
                    with rasterio.open(filepath) as src:
                        # See if file has a color map
                        if len(self.cmap) == 0:
                            try:
                                self.cmap = src.colormap(1)
                            except ValueError:
                                pass

                        if crs is None:
                            crs = src.crs
                        if res is None:
                            res = src.res[0]

                        with WarpedVRT(src, crs=crs) as vrt:
                            minx, miny, maxx, maxy = vrt.bounds
                except rasterio.errors.RasterioIOError:
                    # Skip files that rasterio is unable to read
                    continue
                else:
                    mint: float = 0
                    maxt: float = sys.maxsize
                    if "date" in match.groupdict():
                        date = match.group("date")
                        mint, maxt = disambiguate_timestamp(date, self.date_format)

                    coords = (minx, maxx, miny, maxy, mint, maxt)
                    self.index.insert(i, coords, filepath)
                    i += 1

        if i == 0:
            raise FileNotFoundError(
                f"No {self.__class__.__name__} data was found in '{root}'"
            )

+        if not self.all_bands:
+            band_indexes = None
+        else:
+            if self.bands:
+                band_indexes = [self.all_bands.index(i) + 1 for i in self.bands]
+               assert len(band_indexes) == len(self.bands)
+            else:
+                band_indexes = None
+
+            if self.rgb_bands:
+                rgb_band_indexes = [self.all_bands.index(i) + 1 for i in self.rgb_bands]
+                assert len(rgb_band_indexes) == len(self.rgb_bands)
+
+       self.band_indexes = band_indexes
        self._crs = cast(CRS, crs)
        self.res = cast(float, res)
```

adamjstewart:
Yes, that looks correct to me. You wouldn't even need to override `__init__` in your `Raster8BandFolder` example dataset since you aren't changing anything.

ashnair1:
Great. I'll push a PR tomorrow (once mypy is satisfied) and we can continue from there.

torchgeo/datasets/geo.py
```diff
@@ -418,10 +435,10 @@ def __getitem__(self, query: BoundingBox) -> Dict[str, Any]:
                             filename = filename[:start] + "*" + filename[end:]
                     filepath = glob.glob(os.path.join(directory, filename))[0]
                     band_filepaths.append(filepath)
-                data_list.append(self._merge_files(band_filepaths, query))
+                data_list.append(self._merge_files(band_filepaths, query, [1]))
```
adamjstewart, line 438: Why use `[1]` instead of the default?

torchgeo/datasets/geo.py
```diff
@@ -375,6 +378,20 @@ def __init__(
...
+        else:
+            if self.bands:
+                band_indexes = [self.all_bands.index(i) + 1 for i in self.bands]
+                assert len(band_indexes) == len(self.bands)
+            else:
+                band_indexes = None
+
+            if self.rgb_bands:
+                rgb_band_indexes = [self.all_bands.index(i) + 1 for i in self.rgb_bands]
+                assert len(rgb_band_indexes) == len(self.rgb_bands)
```
adamjstewart, line 392: This variable doesn't seem to be used anywhere. I think it will have the side-effect that you aren't allowed to not include all RGB bands during model training, which we don't want. I would remove this if-statement.

torchgeo/datasets/sentinel.py
```diff
@@ -97,9 +97,9 @@ def __init__(
         Raises:
             FileNotFoundError: if no files are found in ``root``
         """
-        self.bands = bands if bands else self.all_bands
+        bands = bands or self.all_bands
```
adamjstewart, line 100: So `band_indices` is `Optional[Sequence[str]]` and `bands` is `Sequence[str]`. And `band_indices = None` and `bands = []` mean the same thing? I wonder if it would be more clear to match these such that `bands = None` means use all bands.

adamjstewart (commented):
Apologies for taking so long to review this!

torchgeo/datasets/geo.py
ashnair1: Hmm must be a remainder from when I was debugging this. Removed

torchgeo/datasets/geo.py
ashnair1: Removed

torchgeo/datasets/sentinel.py
ashnair1: Updated `bands` to be `Optional[Sequence[str]]` where setting `bands=None` defaults to all_bands

ashnair1:
Looks like we can remove Sentinel2's `__init__`

adamjstewart:
Yep, let's remove Sentinel-2's `__init__`

torchgeo/datasets/geo.py
```diff
@@ -375,6 +377,15 @@ def __init__(
                 f"No {self.__class__.__name__} data was found in '{root}'"
             )
 
+        if bands and self.all_bands:
+            band_indexes = [self.all_bands.index(i) + 1 for i in bands]
+            self.bands = bands
+            assert len(band_indexes) == len(self.bands)
+        else:
```
adamjstewart, line 384: Should we raise an error `elif bands`? If someone asks for specific bands but doesn't define `all_bands` then that could lead to unexpected behavior.

adamjstewart (commented):
Could also add a dummy `_verify()` method to the base class and get rid of the `__init__` for most methods, but let's save that for another PR. There's also a bug where classes with no `__init__` don't get any docs, but there's a sphinx setting to fix that I need to play around with.

Can you add unit tests for this? We'll want to test selecting a subset of bands for at least one `separate = True` and one `separate = False` dataset. Test just needs to make sure that the total number of bands returned actually changes.

torchgeo/datasets/geo.py
ashnair1: What about making `all_bands` an abstract method? Wouldn't it be sensible to ask users to define the bands of the datasets they create?

adamjstewart: The problem with that is datasets that only have a single band. Especially for label datasets I don't think that makes sense.

adamjstewart:
Remind me to review this this weekend.

ashnair1:
Pinging @adamjstewart

tests/datasets/test_geo.py
```diff
@@ -156,20 +160,29 @@ def test_and_nongeo(self, dataset: GeoDataset) -> None:
...
     def naip(self, request: SubRequest) -> NAIP:
         root = os.path.join("tests", "data", "naip")
+        bands = request.param[1]
         crs = CRS.from_epsg(3005)
         transforms = nn.Identity()
-        cache = request.param
-        return NAIP(root, crs=crs, transforms=transforms, cache=cache)
-
-    @pytest.fixture(params=[True, False])
+        cache = request.param[0]
```
adamjstewart, line 169: Minor: I would change the order of these request params and the order in which these variables are defined to match the order of arguments in the next line

tests/datasets/test_geo.py
```diff
@@ -156,20 +160,29 @@ def test_and_nongeo(self, dataset: GeoDataset) -> None:
...
+            [
+                ["B04", "B03", "B02"],
+                ["B01", "B02", "B03", "B04", "B05", "B06", "B07", "B08", "B09", "B11"],
+            ],
+        )
+    )
     def sentinel(self, request: SubRequest) -> Sentinel2:
         root = os.path.join("tests", "data", "sentinel2")
-        bands = ["B01", "B02", "B03", "B04", "B05", "B06", "B07", "B08", "B09", "B11"]
+        bands = request.param[1]
```
adamjstewart, line 183: None of our tests actually test to see if the image returned by `__getitem__` has the same number of bands as we ask for, right? That's the main thing I want to test here.

torchgeo/datasets/agb_live_woody_density.py
```diff
@@ -83,10 +83,11 @@ def __init__(
         """
         self.root = root
         self.download = download
+        bands: Sequence[str] = []
```
adamjstewart, line 86: Should these be `[]` or `None`? To me, `[]` implies no bands, whereas `None` implies all bands. In fact, we could just not pass `bands` to `__init__` at all. You would need to change `transforms, cache` to `transform=transforms, cache=cache` since the inputs are now out of order. This also raises the question of what we should do when a user uses `bands=[]`. Should we raise an error saying that `bands` can't be empty? What do you think?

torchgeo/datasets/geo.py
```diff
@@ -375,6 +377,17 @@ def __init__(
                 f"No {self.__class__.__name__} data was found in '{root}'"
             )
 
+        if bands and self.all_bands:
+            band_indexes = [self.all_bands.index(i) + 1 for i in bands]
+            self.bands = bands
+            assert len(band_indexes) == len(self.bands)
+        elif bands:
+            raise AssertionError("all_bands must be specified")
```
adamjstewart, line 385: This error message isn't really clear about what the problem or solution are. How about something like:
```python
msg = f"{self.__class__.__name__} is missing an `all_bands` attribute, so `bands` cannot be specified."
raise AssertionError(msg)
```
May need to split onto two lines due to line length.

torchgeo/datasets/landsat.py
```diff
@@ -74,10 +74,10 @@ def __init__(
         Raises:
             FileNotFoundError: if no files are found in ``root``
         """
-        self.bands = bands if bands else self.all_bands
-        self.filename_glob = self.filename_glob.format(self.bands[0])
+        if bands:
+            self.filename_glob = self.filename_glob.format(bands[0])
```
adamjstewart, line 78: I don't think this will work, `self.filename_glob` always needs to be expanded or else we won't find any files during globbing. If `bands` is `[]` or `None`, could we just use `self.all_bands[0]`?

torchgeo/datasets/agb_live_woody_density.py
ashnair1: Since a `RasterDataset` is guaranteed to have at least one band I don't think there is a distinction between `[]` and `None`. If `bands` evaluates to False (`None` or `[]`) we use `all_bands` else we use `bands`.

> In fact, we could just not pass bands to `__init__` at all. You would need to change transforms, cache to `transform=transforms`, `cache=cache` since the inputs are now out of order

Agreed. I'll push the change.

torchgeo/datasets/sentinel.py
```diff
@@ -95,11 +95,11 @@ def __init__(
         Raises:
             FileNotFoundError: if no files are found in ``root``
         """
-        self.bands = bands if bands else self.all_bands
-        self.filename_glob = self.filename_glob.format(self.bands[0])
+        bands = bands or self.all_bands
```
adamjstewart, line 98: I like what you did here, can we do the same thing for landsat too? I think the situation is the same, but this logic is much more clear than the if-statement.

adamjstewart (approved):
This looks great! Couple minor formatting requests, but otherwise this looks ready to me.

PR #689, opened 2022-07-27
Adds custom typehinting for size (pixel or CRS) arguments

TCherici:
My IDE was complaining about me using an integer for the `size` parameter, despite that being not only supported, but also the default expected type.

I'm not 100% certain about the naming of the custom type (`PixelsOrCRS`), or if its addition even hinders readability, but explicitly writing `Union[int, float]` would make the type hinting very verbose and difficult to read imo.

### DIFF
--- a/torchgeo/samplers/batch.py
+++ b/torchgeo/samplers/batch.py
@@ -12,7 +12,7 @@ from torch.utils.data import Sampler
 
 from ..datasets import BoundingBox, GeoDataset
 from .constants import Units
-from .utils import _to_tuple, get_random_bounding_box
+from .utils import PixelsOrCRS, _to_tuple, get_random_bounding_box
 
 # https://github.com/pytorch/pytorch/issues/60979
 # https://github.com/pytorch/pytorch/pull/61045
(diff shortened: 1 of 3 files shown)


adamjstewart:
I think `Number` would be a better name for `Union[int, float]`.

But more importantly, this seems like a bug in your IDE, not a bug in our type hints. According to [PEP 484](https://peps.python.org/pep-0484/#the-numeric-tower):

> Rather than requiring that users write `import numbers` and then use `numbers.Float` etc., this PEP proposes a straightforward shortcut that is almost as effective: when an argument is annotated as having type `float`, an argument of type `int` is acceptable

What IDE do you use? Maybe it would be more productive to open an issue with them instead. This doesn't just affect `torchgeo.samplers`, it affects all Python libraries that use type hints.

adamjstewart:
> What IDE do you use? Maybe it would be more productive to open an issue with them instead. This doesn't just affect `torchgeo.samplers`, it affects all Python libraries that use type hints.

@TCherici?

adamjstewart:
Closing as this is a bug in the IDE, not a bug in TorchGeo or mypy.

adamjstewart:
Note that a slight change was made to these type hints in #876. Float is more general than int. If ints are input, ints are also returned, not floats.

PR #701, opened 2022-08-05
Test rarfile on Windows

adamjstewart:
https://github.com/markokr/rarfile/issues/72


adamjstewart:
No idea why this works for all versions of Python except 3.7. I'm pretty stumped on this one. If we can't figure it out, we may have to wait until we drop Python 3.7 support.

adamjstewart:
My best guess is that this has something to do with the fact that we are using Conda with Python 3.7 on Windows. It may be time to stop testing this.

isaaccorley (approved):
lgtm

PR #707, opened 2022-08-08
Tutorials: use IntersectionDataset in sampler

adamjstewart:
I believe this was a holdover from 0.1.0 back when `IntersectionDataset` didn't have its own index. Current best practice is to use `IntersectionDataset` instead of one of the individual datasets so that the sampler doesn't try to sample outside of the intersection.


isaaccorley (approved):
Lgtm

Issue #660, opened 2022-07-09
Add DatasetNotFoundError for consistent error messages


adamjstewart:
### Summary

We should add a `torchgeo.datasets.utils.DatasetNotFoundError` class for consistent error messages when datasets aren't found.

### Rationale

There is currently very little consistency between the errors reported by different datasets. Some raise `RuntimeError`, others raise `FileNotFoundError`. There are dozens of different messages raised by each dataset with little commonality between the text. Some mention using `download=True`, others mention changing the root directory. Some print the root directory, others leave users guessing.

### Implementation

We should subclass `FileNotFoundError` and except the dataset instance as input. With this, we can check for `download` and `checksum` attributes to determine the appropriate message. Some possible formats include:
```python
# Has a `self.download` attribute
f"Dataset not found in `root={self.root}` and `download=False`, either specify a different `root` directory or use `download=True` to automatically download the dataset."

# Doesn't have a `self.download` attribute
f"Dataset not found in `root={self.root}`, either specify a different `root` directory or manually download the dataset."
```
This will also require consistent documentation in docstrings. Maybe something like:
```rst
# Has a ``self.download`` attribute
Raises:                                                                          
    FileNotFoundError: if ``download=False`` but dataset is not found            
        or checksums don't match          

# Doesn't have a ``self.download`` attribute
Raises:                                                                          
    FileNotFoundError: if dataset is not found   
```
Of course, we'll also need to update the tests as well.

### Alternatives

We could simply update everything to have common text, but things would rapidly become out of sync without a centralized class to standardize the error message. This also avoids code duplication.

PR #730, opened 2022-08-19
DataModules: pass kwargs directly to datasets

### LINKED ISSUE #666: Pass datamodule kwargs to datasets
### Summary

We should remove dataset-specific arguments from datamodules and instead pass them directly to the dataset through kwargs.

### Rationale

As an example, we'll look at `TropicalCycloneWindEstimation` and `CycloneDataModule`.

`TropicalCycloneWindEstimation` has a number of options that users may want to configure:

* root: root directory containing data
* split: train or test
* transforms: data augmentations to apply
* download: whether or not to download the dataset
* api_key: MLHUB_API_KEY
* checksum: whether or not to checksum the download

However, `CycloneDataModule` only exposes a subset of these:

* root_dir: different name for some reason
* api_key: MLHUB_API_KEY

If a user wants to, for example, automatically download the dataset, they have to modify the source code of `CycloneDataModule`. If we instead pass the `**kwargs` from `CycloneDataModule` directly to `TropicalCycloneWindEstimation`, we end up with less code duplication, more features, and greater consistency.

### Implementation

Here is what the change would look like for `CycloneDataModule`:
```diff
diff --git a/torchgeo/datamodules/cyclone.py b/torchgeo/datamodules/cyclone.py
index 9a97242..32db799 100644
--- a/torchgeo/datamodules/cyclone.py
+++ b/torchgeo/datamodules/cyclone.py
@@ -26,11 +26,9 @@ class CycloneDataModule(pl.LightningDataModule):
 
     def __init__(
         self,
-        root_dir: str,
         seed: int,
         batch_size: int = 64,
         num_workers: int = 0,
-        api_key: Optional[str] = None,
         **kwargs: Any,
     ) -> None:
         """Initialize a LightningDataModule for NASA Cyclone based DataLoaders.
@@ -45,11 +43,10 @@ class CycloneDataModule(pl.LightningDataModule):
                 downloaded
         """
         super().__init__()  # type: ignore[no-untyped-call]
-        self.root_dir = root_dir
         self.seed = seed
         self.batch_size = batch_size
         self.num_workers = num_workers
-        self.api_key = api_key
+        self.kwargs = kwargs
 
     def preprocess(self, sample: Dict[str, Any]) -> Dict[str, Any]:
         """Transform a single sample from the Dataset.
@@ -76,12 +73,7 @@ class CycloneDataModule(pl.LightningDataModule):
         This includes optionally downloading the dataset. This is done once per node,
         while :func:`setup` is done once per GPU.
         """
-        TropicalCycloneWindEstimation(
-            self.root_dir,
-            split="train",
-            download=self.api_key is not None,
-            api_key=self.api_key,
-        )
+        TropicalCycloneWindEstimation(split="train", **self.kwargs)
 
     def setup(self, stage: Optional[str] = None) -> None:
         """Create the train/val/test splits based on the original Dataset objects.
```
This should be done for all other datamodules.

### Alternatives

The alternative is to add new parameters to each datamodule as we need them. This results in an inconsistent API.

### Additional information

_No response_

adamjstewart:
Closes #666

### Pros

* Consistency: every datamodule now has the same kwargs as its corresponding dataset
* New features: this means that every datamodule now supports automatic downloads if its corresponding dataset does too
* No unused kwargs: previously, a typo in an argument or non-existent argument would go unnoticed, now it will raise an error
* Reduced code duplication: -140 lines of code

### Cons

* Backwards incompatibility, `root_dir` has been renamed to `root` for consistency
* Can no longer pass `root_dir` as the first positional arg, must be a keyword arg

To elaborate, the following are no longer valid:
```python
BigEarthNetDataModule('data/bigearthnet')
BigEarthNetDataModule(root_dir='data/bigearthnet')
```
Instead, users will need to use:
```python
BigEarthNetDataModule(root='data/bigearthnet')
```

### DIFF
--- a/conf/oscd.yaml
+++ b/conf/oscd.yaml
@@ -19,7 +19,7 @@ experiment:
     ignore_index: 0
   datamodule:
     root: "data/oscd"
-    batch_size: 32
+    train_batch_size: 32
     num_workers: 4
     val_split_pct: 0.1
     bands: "all"
--- a/conf/oscd.yaml
+++ b/conf/oscd.yaml
@@ -18,7 +18,7 @@ experiment:
     num_filters: 256
     ignore_index: 0
   datamodule:
-    root_dir: "data/oscd"
+    root: "data/oscd"
     batch_size: 32
     num_workers: 4
     val_split_pct: 0.1
--- a/torchgeo/datamodules/so2sat.py
+++ b/torchgeo/datamodules/so2sat.py
@@ -72,6 +72,8 @@ class So2SatDataModule(pl.LightningDataModule):
             band_set: Collection of So2Sat bands to use
             unsupervised_mode: Makes the train dataloader return imagery from the train,
                 val, and test sets
+            **kwargs: Additional keyword arguments passed to
+                :class:`~torchgeo.datasets.So2Sat`
         """
         super().__init__()
         self.batch_size = batch_size
--- a/torchgeo/datamodules/so2sat.py
+++ b/torchgeo/datamodules/so2sat.py
@@ -58,29 +58,24 @@ class So2SatDataModule(pl.LightningDataModule):
 
     def __init__(
         self,
-        root_dir: str,
         batch_size: int = 64,
         num_workers: int = 0,
-        bands: str = "rgb",
         unsupervised_mode: bool = False,
         **kwargs: Any,
     ) -> None:
         """Initialize a LightningDataModule for So2Sat based DataLoaders.
 
         Args:
-            root_dir: The ``root`` arugment to pass to the So2Sat Dataset classes
             batch_size: The batch size to use in all created DataLoaders
             num_workers: The number of workers to use in all created DataLoaders
-            bands: Either "rgb" or "s2"
             unsupervised_mode: Makes the train dataloader return imagery from the train,
                 val, and test sets
         """
         super().__init__()
-        self.root_dir = root_dir
         self.batch_size = batch_size
         self.num_workers = num_workers
-        self.bands = bands
         self.unsupervised_mode = unsupervised_mode
+        self.kwargs = kwargs
 
         self.norm = Normalize(self.band_means, self.band_stds)
 
@@ -97,7 +92,8 @@ class So2SatDataModule(pl.LightningDataModule):
         sample["image"] = self.norm(sample["image"])
         sample["image"] = sample["image"][self.reindex_to_rgb_first, :, :]
 
-        if self.bands == "rgb":
+        bands = self.kwargs.get("bands", "all")
+        if bands == "rgb":
             sample["image"] = sample["image"][:3, :, :]
 
         return sample
@@ -107,7 +103,7 @@ class So2SatDataModule(pl.LightningDataModule):
 
         This method is only called once per run.
         """
-        So2Sat(self.root_dir, checksum=False)
+        So2Sat(**self.kwargs)
 
     def setup(self, stage: Optional[str] = None) -> None:
         """Initialize the main ``Dataset`` objects.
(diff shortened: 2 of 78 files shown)


conf/oscd.yaml
```diff
@@ -18,8 +18,8 @@ experiment:
     num_filters: 256
     ignore_index: 0
   datamodule:
-    root_dir: "data/oscd"
-    batch_size: 32
```
adamjstewart, line 22: The OSCD data module has never had a `batch_size` parameter. The correct name is `train_batch_size`. This parameter was being ignored since we never actually used `kwargs` before.

torchgeo/datamodules/so2sat.py
```diff
@@ -58,29 +58,27 @@ class So2SatDataModule(pl.LightningDataModule):
 
     def __init__(
         self,
-        root_dir: str,
         batch_size: int = 64,
         num_workers: int = 0,
-        bands: str = "rgb",
+        band_set: str = "rgb",
```
adamjstewart, line 63: Renamed from `bands` to `band_set` to match `SEN12MSDataModule` and to remove ambiguity between `So2SatDataModule` and `So2Sat` where `bands` is a list of band names.

isaaccorley (dismissed):
I couldn't find any issues. lgtm this one was much needed.

adamjstewart:
Will give @calebrob6 a chance to review, no rush on this PR since it's going in 0.4.0.

calebrob6:
Agree that this was much needed -- I just don't like how it isn't obvious from the DataModule constructors that you need to pass a root parameter. Since every dataset will need a root parameter can't we just keep it in?

adamjstewart:
> Since every dataset will need a root parameter

In theory, we could someday have datasets without a root parameter (streaming data from Azure/AWS for example). But yes, all of our current datasets have one.

> I just don't like how it isn't obvious from the DataModule constructors that you need to pass a root parameter... can't we just keep it in?

We could keep it in and forward it like we used to. Or we could properly document the available kwargs, or the fact that kwargs will be forwarded. Or both.

calebrob6:
Properly document the available kwargs seems best to me

adamjstewart:
@calebrob6 check out the docs for the latest version and see if that's better. The alternative would be to make a formal doc section like:
```rst
Initialize a LightningDataModule for BigEarthNet based DataLoaders.           
                                                                                         
Args:                                                                            
    batch_size: The batch size to use in all created DataLoaders                 

Keyword Args:
    split: train/val/test split to load                                          
    bands: load Sentinel-1 bands, Sentinel-2, or both. one of {s1, s2, all}      
    num_classes: number of classes to load in target. one of {19, 43}            
    download: if True, download dataset and store it in the root directory       
    checksum: if True, check the MD5 of the downloaded files (may be slow)
```
but that ends up with a lot of duplication and becomes harder to maintain.

adamjstewart:
Closing/reopening to try to kick off CLA bot.

adamjstewart:

calebrob6:
This looks like a good compromise on documentation. What if we specially check for "root" in kwargs and raise a helpful error message if it isn't included? My worry is that if a user doesn't add `root` they'll get an error message from the Dataset object which might be confusing.

calebrob6:
(or warning)

adamjstewart:
The stack trace will tell them the full path the code took to raise that error (not necessarily clear to a non-programmer though). `root` isn't required, all datasets have a default `root`. If anyone ever implements #660 the error message will be a bit more helpful for datasets that don't already explain what `root` is set to.

calebrob6:
It only needs to be a message on the datasets that do have root, right?

adamjstewart:
At the moment this is every dataset.

calebrob6 (approved):

PR #734, opened 2022-08-21
dependabot: fewer setuptools updates

adamjstewart:
Setuptools has a new release almost daily:

* https://github.com/pypa/setuptools/releases
* https://pypi.org/project/setuptools/#history

We don't need to test every single patch release, let's focus on testing major and minor version updates only.


isaaccorley (approved):
Had to do this for some other projects too. lgtm

PR #394, opened 2022-02-11
Add band selection to So2Sat Dataset and adopt plot method

nilsleh:
This PR addresses the last comment by Adam in issue #289 about adding band selection to the So2Sat Dataset. Changes I made:
- add band selection as a sequence and select desired bands in __getitem__
- adopt the plot method to look for RGB indices and only plot then
- since the dataset takes a specific split as an argument, I thought it would make sense that the integrity check also only checks the specific selected split, so that not all splits need to have been downloaded

The remaining question I have is whether there is a standard naming convention for the S1 bands, described [in the dataset repo](https://github.com/zhu-xlab/So2Sat-LCZ42). I couldn't find one and therefore just prepended "S1".

Plot examples:
s021
s022
s023

PR #735, opened 2022-08-21
So2Sat: rename bands

adamjstewart:
Follow-up to #394 

I think it's better to ensure that all band names in So2Sat match the band names used in our Sentinel datasets. Since some bands come from Sentinel-1 and some bands come from Sentinel-2, the best way to handle this is to prefix `S1_` and `S2_` in front of all bands.

Alternative: do Sentinel-1 bands really go by B1, B2, ...? I see a lot of VV/VH/HV/HH online. Someone should implement a `Sentinel1` dataset so we can match that. If the bands no longer use B# names, we can drop the prefix from all bands.

@nilsleh


isaaccorley (approved):

calebrob6:
An FYI: what the S1_* bands correspond to is specified here https://github.com/zhu-xlab/So2Sat-LCZ42#description-of-the-content-of-sen1

adamjstewart:
Opened #821 to start on this. It seems like Sentinel-1 bands are named 001, 002, ..., 005, 006. Do we want to use S1_001 and S2_B1 for consistency with S1/S2, or S1_B1 and S2_B01 for internal consistency?

nilsleh:
I don't have an opinion on this but also haven't been long enough in this area to know how much people would care. I think as long as the band correspondence is clearly documented either way is fine.

adamjstewart:
Not sure if this should influence the decision made in this PR, but in #821 we ultimately went with HH, HV, VV, VH instead of 001, 002, ... for the band names.

PR #740, opened 2022-08-25
Skip Python 3.7 Windows tests

adamjstewart:
Many of our dependencies, including rasterio, do not provide Windows wheels for Python 3.7. To get around this, we were using Conda to install them. However, this seems to interfere with #701, and increases test setup time. I don't think there's a significant benefit to testing Python 3.7 on Windows since we test 3.7 on other platforms and other versions on Windows. Python 3.7 will still be supported on Windows, it just won't be tested.

Alternative is to keep using Conda and skip rarfile tests, or to drop Python 3.7 support entirely: #656


isaaccorley (approved):
lgtm

PR #754, opened 2022-09-03
Sentinel-2: add support for files downloaded from USGS EarthExplorer

### LINKED ISSUE #505: Sentinel2 data set not able to locate data file downloaded from https://scihub.copernicus.eu
I see an error when I try to access the data downloaded from https://scihub.copernicus.eu/dhus/#/home for sentinel-2 using torchgeo Sentinel2 dataset. After investigating I found that the function does not uses the standard naming convention for sentinel-2 data products.

files downloaded from USGS Earth Explorer seem to have a different filename format.

Naming Conventions
https://sentinels.copernicus.eu/web/sentinel/user-guides/sentinel-2-msi/naming-convention
https://www.usgs.gov/centers/eros/science/sentinel-2-data-dictionary#tile_number

If there is a way to change file name please suggest.

----ERROR LOG----
```console
$ python train.py
  File "train.py", line 12, in <module>
    dataset = Sentinel2(root='./../data/sentinel', crs=None, transforms=None, bands=[], cache=True)
  File "/home/manish/projects/pytorch/torchgeo_lcc/torchgeo_env/lib/python3.8/site-packages/torchgeo/datasets/sentinel.py", line 103, in __init__
    super().__init__(root, crs, res, transforms, cache)
  File "/home/manish/projects/pytorch/torchgeo_lcc/torchgeo_env/lib/python3.8/site-packages/torchgeo/datasets/geo.py", line 374, in __init__
    raise FileNotFoundError(
FileNotFoundError: No Sentinel2 data was found in './../data/sentinel'
```

adamjstewart:
Closes #505 @tinkueg

Depending on whether you download Sentinel-2 imagery from USGS EarthExplorer or Copernicus Open Access Hub, the filenames are slightly different. This PR fixes several issues:

- [x] Adds support for Sentinel-2 imagery downloaded from USGS EarthExplorer
- [x] Prevent dataset from loading bands at different resolutions
- [x] Support loading even when band B02 is not present

Note that this approach does not allow one to load bands unless they are present at the specified resolution. For example, in the files that I downloaded, band B01 is only available at 20m and 60m resolutions, not 10m.


tests/data/sentinel2/data.py
```diff
@@ -12,19 +13,84 @@
...
+                        "T26EMU_20220414T110751_WVP_60m.jp2",
+                    ],
+                }
+            }
+        }
+    },
+}
 
 
 def create_file(path: str, dtype: str, num_channels: int) -> None:
```
adamjstewart, line 96: Driver is wrong, will fix once I rebuild GDAL with OpenJPEG support

tests/data/sentinel2/data.py
```diff
@@ -12,24 +13,89 @@
...
     profile = {}
-    profile["driver"] = "GTiff"
+    profile["driver"] = "JP2OpenJPEG"
```
adamjstewart, line 98: Why doesn't this work? With `GTiff`, rasterio happily writes GeoTIFF files with the `.jp2` extension. When I change to `JP2OpenJPEG`, rasterio creates all of the directories, but doesn't write any files to disk. There is no error message. I've reproduced this with both Spack- and Conda-installed GDAL/rasterio. GDAL itself seems fine. `gdalinfo --formats` lists this driver under both read and write. `gdal_create -of JP2OpenJPEG` works fine. Quite confused...

tests/data/sentinel2/data.py
```diff
@@ -4,40 +4,106 @@
...
     profile["dtype"] = dtype
     profile["count"] = num_channels
-    profile["crs"] = "epsg:4326"
-    profile["transform"] = rasterio.transform.from_bounds(0, 0, 1, 1, 1, 1)
+    profile["crs"] = CRS.from_epsg(32616)
+    profile["transform"] = Affine(10.0, 0.0, 399960.0, 0.0, -10.0, 4500000.0)
```
calebrob6, line 104: This will make every file have 10m/px resolution. Is this desired?

adamjstewart: Fixed

calebrob6:
Is it the case that the Copernicus Open Access Hub 10m data also comes in 20m and 60m resolutions? (e.g. I see B02 in the 10m, 20m, and 60m folders)

calebrob6 (approved):

adamjstewart:
Yes, at least when I downloaded it a few months ago: https://github.com/microsoft/torchgeo/issues/505#issuecomment-1099646913

PR #755, opened 2022-09-03
Random GeoSamplers: add default length

### LINKED ISSUE #380: Improvements to RandomBatchGeoSampler
I propose two improvements to `RandomBatchGeoSampler`:

1. The `size` parameter should be defined in pixels, not in CRS units. The output of a DataLoader using this sampler will be fed directly into a CNN or similar in most cases, which expect a certain defined input size in pixels.
To accomplish this, the user has to perform the conversion of CRS to pixels manually like this
```
tile_size_pix = 256
sampler_size = tile_size_pix * input_ds.res
sampler = RandomBatchGeoSampler(input_ds, size=sampler_size, ...)
```
which is inconvenient.

2. The `length` parameter should have a meaningful default value. Currently, it is described as "length (int) – number of samples per epoch". Since an epoch is defined as processing an entire dataset exactly once, a meaningful default value could be calculated by using the dataset area and the sampler tile area
```
sampler_len = int(get_bbox_area(input_ds.bounds)/(sampler_size*sampler_size))
sampler = RandomBatchGeoSampler(input_ds, size=sampler_size, batch_size=8, length=sampler_len)
```
with
```
# This function would be nice to have in the BoundingBox class
def get_bbox_area(box):
    return (box.maxx - box.minx) * (box.maxy - box.miny)
```

Let me know if this makes sense.

adamjstewart:
Closes #380 @tritolol 

Uses the same logic we use to compute `GridGeoSampler` length, updated by @remtav in #630

Has to wait until the 0.4.0 release since we add a new public function. We could make it private, but I wanted to render the math in the docs 😃

### DIFF
--- a/torchgeo/samplers/batch.py
+++ b/torchgeo/samplers/batch.py
@@ -12,7 +12,7 @@ from torch.utils.data import Sampler
 
 from ..datasets import BoundingBox, GeoDataset
 from .constants import Units
-from .utils import _to_tuple, get_random_bounding_box
+from .utils import _to_tuple, get_random_bounding_box, tile_to_chips
 
 # https://github.com/pytorch/pytorch/issues/60979
 # https://github.com/pytorch/pytorch/pull/61045
@@ -70,7 +70,7 @@ class RandomBatchGeoSampler(BatchGeoSampler):
         dataset: GeoDataset,
         size: Union[Tuple[float, float], float],
         batch_size: int,
-        length: int,
+        length: Optional[int] = None,
         roi: Optional[BoundingBox] = None,
         units: Units = Units.PIXELS,
     ) -> None:
@@ -94,6 +94,9 @@ class RandomBatchGeoSampler(BatchGeoSampler):
 
         .. versionchanged:: 0.3
            Added ``units`` parameter, changed default to pixel units
+
+        .. versionchanged:: 0.4
+           ``length`` parameter is now optional, a reasonable default will be used
         """
         super().__init__(dataset, roi)
         self.size = _to_tuple(size)
@@ -102,7 +105,7 @@ class RandomBatchGeoSampler(BatchGeoSampler):
             self.size = (self.size[0] * self.res, self.size[1] * self.res)
 
         self.batch_size = batch_size
-        self.length = length
+        self.length = 0
         self.hits = []
         areas = []
         for hit in self.index.intersection(tuple(self.roi), objects=True):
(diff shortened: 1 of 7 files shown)


torchgeo/samplers/batch.py
```diff
@@ -70,7 +70,7 @@ def __init__(
         dataset: GeoDataset,
         size: Union[Tuple[float, float], float],
         batch_size: int,
-        length: int,
+        length: Optional[int] = None,
```
calebrob6, line 73: Should we specify in the docstring what this defaults to?

adamjstewart: It's a bit hard to describe exactly what this defaults to. Any suggestions for wording?

adamjstewart: Ping @calebrob6

torchgeo/samplers/batch.py
```diff
@@ -111,8 +114,15 @@ def __init__(
                 bounds.maxx - bounds.minx >= self.size[1]
                 and bounds.maxy - bounds.miny >= self.size[0]
             ):
+                if bounds.area > 0:
+                    rows, cols = tile_to_chips(bounds, self.size)
+                    self.length += rows * cols
+                else:
+                    self.length += 1
```
calebrob6, line 121: also, why is length incremented for a hit that has area <= 0?

adamjstewart: This is for point datasets like GBIF/iNaturalist/EDDMapS

torchgeo/samplers/batch.py
calebrob6: It appears to default to the maximal number of non-overlapping patches of size `size` that can be fit into the area of the dataset.

torchgeo/samplers/batch.py
```diff
@@ -88,12 +88,17 @@ def __init__(
             size: dimensions of each :term:`patch`
             batch_size: number of samples per batch
             length: number of samples per epoch
+                (defaults to the maximal number of non-overlapping :term:`chips <chip>`
+                of size ``size`` that can be sampled from the dataset)
```
adamjstewart, line 92: My only problems with this are:

1. It's technically not true, the last row/col will overlap
2. It may imply that the chips returned will not overlap, which is not the case

calebrob6: 1. Or something like, "the number of chips that would be sampled following the "tile_to_chips" strategy from ...."
2. If you want, you can also add a note that randomly sampled chips might overlap

adamjstewart: 1. I feel like this is going to end up being a whole paragraph lol, that's why I originally said "reasonable default" and left it at that. Maybe just throw in an "approximately" somewhere

calebrob6 (approved):

PR #226, opened 2021-11-07
0.1.0 release

adamjstewart:
See https://github.com/microsoft/torchgeo/wiki/Releasing-Instructions

Closes #55

PR #756, opened 2022-09-03
Tutorials: replace cell magic with shell command

adamjstewart:
I'm hoping this is what's causing https://github.com/microsoft/torchgeo/runs/8172122269?check_suite_focus=true to fail. Also see https://github.com/treebeardtech/nbmake/issues/80.

When I run the previous code in Google Colab, I see:
```
UsageError: Line magic function `%nvidia-smi` not found.
```
From my understanding, both `%` and `!` will let you run shell commands, but `%` is a cell magic command for which there are only a finite number of mappings (pip and apt are allowed, but not nvidia-smi). The other difference is that `!` runs in a subprocess, while `%` runs in the current process. This is required for something like `pip install`, but not for something like `nvidia-smi`.


adamjstewart:
Second commit is something I missed in #714

adamjstewart:
Bleh, this fixes Jupyter but breaks nbsphinx. See https://github.com/spatialaudio/nbsphinx/issues/670.

adamjstewart:
I knew this seemed familiar. I actually broke this (`!` -> `%`) back in #226, our first release. This PR attempts to revert that change.

adamjstewart:
This fixes some of the issues. I'm still seeing unexplainable errors in https://github.com/microsoft/torchgeo/runs/8172778528?check_suite_focus=true. Basically, `torch.cuda.is_available()` is returning True, but CUDA doesn't work. Our transforms tutorial explicitly passes things to the GPU if it's available, and our trainers tutorial uses Lightning which does the same. Not sure why `torch.cuda.is_available()` is lying, and we can't easily reproduce this locally.

adamjstewart:
This looks promising! https://github.com/github/roadmap/issues/505

calebrob6:
> This fixes some of the issues. I'm still seeing unexplainable errors in https://github.com/microsoft/torchgeo/runs/8172778528?check_suite_focus=true. Basically, `torch.cuda.is_available()` is returning True, but CUDA doesn't work. Our transforms tutorial explicitly passes things to the GPU if it's available, and our trainers tutorial uses Lightning which does the same. Not sure why `torch.cuda.is_available()` is lying, and we can't easily reproduce this locally.

It seems that you have that line backwards `device = "cpu" if torch.cuda.is_available() else "cuda"`, should be `"cuda" if torch.cuda.is_available() ...`

calebrob6 (approved):

PR #758, opened 2022-09-05
Add support for ObjectDetection

### LINKED ISSUE #442: [Feature Request] Object Detection Trainer
Hi, I'm curious if there are any plans to add an object detection trainer?

Also, I wonder how feasible it may be to extend this library to work with video data and multi-object tracking tasks? Or is torchvideo more appropriate for this task?

ashnair1:
Adds a new `ObjectDetectionTask` which uses torchvision's `Faster-RCNN` under the hood.

Closes #442

### DIFF
--- a/evaluate.py
+++ b/evaluate.py
@@ -14,7 +14,11 @@ import pytorch_lightning as pl
 import torch
 from torchmetrics import Accuracy, JaccardIndex, MetricCollection
 
-from torchgeo.trainers import ClassificationTask, SemanticSegmentationTask
+from torchgeo.trainers import (
+    ClassificationTask,
+    ObjectDetectionTask,
+    SemanticSegmentationTask,
+)
 from train import TASK_TO_MODULES_MAPPING
 
 
@@ -106,6 +110,14 @@ def run_eval_loop(
             y = batch["mask"].to(device)
         elif "label" in batch:
             y = batch["label"].to(device)
+        elif "boxes" in batch:
+            y = [
+                {
+                    "boxes": batch["boxes"][i].to(device),
+                    "labels": batch["labels"][i].to(device),
+                }
+                for i in range(len(batch["image"]))
+            ]
         with torch.inference_mode():
             y_pred = model(x)
         metrics(y_pred, y)
@@ -176,6 +188,20 @@ def main(args: argparse.Namespace) -> None:
             "learning_rate": model.hparams["learning_rate"],
             "loss": model.hparams["loss"],
         }
+    elif issubclass(TASK, ObjectDetectionTask):
+        val_row = {
+            "split": "val",
+            "detection_model": model.hparams["detection_model"],
+            "backbone": model.hparams["backbone"],
+            "learning_rate": model.hparams["learning_rate"],
+        }
+
+        test_row = {
+            "split": "test",
+            "detection_model": model.hparams["detection_model"],
+            "backbone": model.hparams["backbone"],
+            "learning_rate": model.hparams["learning_rate"],
+        }
     else:
         raise ValueError(f"{TASK} is not supported")
 
@@ -240,6 +266,9 @@ def main(args: argparse.Namespace) -> None:
                     "jaccard_index": test_results["test_JaccardIndex"].item(),
                 }
             )
+        elif issubclass(TASK, ObjectDetectionTask):
+            val_row.update({"map": val_results["map"].item()})
+            test_row.update({"map": test_results["map"].item()})
 
     assert set(val_row.keys()) == set(test_row.keys())
     fieldnames = list(test_row.keys())
--- a/tests/trainers/test_detection.py
+++ b/tests/trainers/test_detection.py
@@ -1,0 +1,56 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import os
+from typing import Any, Dict, Type, cast
+
+import pytest
+from omegaconf import OmegaConf
+from pytorch_lightning import LightningDataModule, Trainer
+
+from torchgeo.datamodules import NASAMarineDebrisDataModule
+from torchgeo.trainers import ObjectDetectionTask
+
+
+class TestObjectDetectionTask:
+    @pytest.mark.parametrize(
+        "name,classname", [("nasa_marine_debris", NASAMarineDebrisDataModule)]
+    )
+    def test_trainer(self, name: str, classname: Type[LightningDataModule]) -> None:
+
+        conf = OmegaConf.load(os.path.join("tests", "conf", name + ".yaml"))
+        conf_dict = OmegaConf.to_object(conf.experiment)
+        conf_dict = cast(Dict[Any, Dict[Any, Any]], conf_dict)
+
+        # Instantiate datamodule
+        datamodule_kwargs = conf_dict["datamodule"]
+        datamodule = classname(**datamodule_kwargs)
+
+        # Instantiate model
+        model_kwargs = conf_dict["module"]
+        model = ObjectDetectionTask(**model_kwargs)
+
+        # Instantiate trainer
+        trainer = Trainer(fast_dev_run=True, log_every_n_steps=1, max_epochs=1)
+        trainer.fit(model=model, datamodule=datamodule)
+        trainer.test(model=model, datamodule=datamodule)
+
+    @pytest.fixture
+    def model_kwargs(self) -> Dict[Any, Any]:
+        return {
+            "detection_model": "faster-rcnn",
+            "backbone": "resnet18",
+            "num_classes": 2,
+        }
+
+    def test_invalid_model(self, model_kwargs: Dict[Any, Any]) -> None:
+        model_kwargs["detection_model"] = "invalid_model"
+        match = "Model type 'invalid_model' is not valid."
+        with pytest.raises(ValueError, match=match):
+            ObjectDetectionTask(**model_kwargs)
+
+    def test_invalid_backbone(self, model_kwargs: Dict[Any, Any]) -> None:
+        model_kwargs["backbone"] = "invalid_backbone"
+        match = "Backbone type 'invalid_backbone' is not valid."
+        with pytest.raises(ValueError, match=match):
+            ObjectDetectionTask(**model_kwargs)
--- a/tests/trainers/test_detection.py
+++ b/tests/trainers/test_detection.py
@@ -34,6 +34,7 @@ class TestObjectDetectionTask:
         trainer = Trainer(fast_dev_run=True, log_every_n_steps=1, max_epochs=1)
         trainer.fit(model=model, datamodule=datamodule)
         trainer.test(model=model, datamodule=datamodule)
+        trainer.predict(model=model, dataloaders=datamodule.val_dataloader())
 
     @pytest.fixture
     def model_kwargs(self) -> Dict[Any, Any]:
--- a/tests/trainers/test_detection.py
+++ b/tests/trainers/test_detection.py
@@ -2,9 +2,10 @@
 # Licensed under the MIT License.
 
 import os
-from typing import Any, Dict, Type, cast
+from typing import Any, Callable, Dict, Type, cast
 
 import pytest
+from _pytest.monkeypatch import MonkeyPatch
 from omegaconf import OmegaConf
 from pytorch_lightning import LightningDataModule, Trainer
 
@@ -16,9 +17,10 @@ class TestObjectDetectionTask:
     @pytest.mark.parametrize(
         "name,classname", [("nasa_marine_debris", NASAMarineDebrisDataModule)]
     )
-    def test_trainer(self, name: str, classname: Type[LightningDataModule]) -> None:
-
-        conf = OmegaConf.load(os.path.join("tests", "conf", name + ".yaml"))
+    def test_trainer(
+        self, name: str, classname: Type[LightningDataModule], monkeypatch: MonkeyPatch
+    ) -> None:
+        conf = OmegaConf.load(os.path.join("tests", "conf", f"{name}.yaml"))
         conf_dict = OmegaConf.to_object(conf.experiment)
         conf_dict = cast(Dict[Any, Dict[Any, Any]], conf_dict)
 
@@ -34,6 +36,11 @@ class TestObjectDetectionTask:
         trainer = Trainer(fast_dev_run=True, log_every_n_steps=1, max_epochs=1)
         trainer.fit(model=model, datamodule=datamodule)
         trainer.test(model=model, datamodule=datamodule)
+
+        if isinstance(getattr(datamodule, "plot"), Callable):  # type: ignore[arg-type]
+            monkeypatch.delattr(classname, "plot")
+            datamodule2 = classname(**datamodule_kwargs)
+            trainer.validate(model=model, datamodule=datamodule2)
 
     @pytest.fixture
     def model_kwargs(self) -> Dict[Any, Any]:
--- a/tests/trainers/test_detection.py
+++ b/tests/trainers/test_detection.py
@@ -54,3 +54,7 @@ class TestObjectDetectionTask:
         match = "Backbone type 'invalid_backbone' is not valid."
         with pytest.raises(ValueError, match=match):
             ObjectDetectionTask(**model_kwargs)
+
+    def test_non_pretrained_backbone(self, model_kwargs: Dict[Any, Any]) -> None:
+        model_kwargs["pretrained"] = False
+        ObjectDetectionTask(**model_kwargs)
--- a/tests/trainers/test_detection.py
+++ b/tests/trainers/test_detection.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT License.
 
 import os
-from typing import Any, Callable, Dict, Type, cast
+from typing import Any, Dict, Type, cast
 
 import pytest
 from _pytest.monkeypatch import MonkeyPatch
@@ -17,9 +17,7 @@ class TestObjectDetectionTask:
     @pytest.mark.parametrize(
         "name,classname", [("nasa_marine_debris", NASAMarineDebrisDataModule)]
     )
-    def test_trainer(
-        self, name: str, classname: Type[LightningDataModule], monkeypatch: MonkeyPatch
-    ) -> None:
+    def test_trainer(self, name: str, classname: Type[LightningDataModule]) -> None:
         conf = OmegaConf.load(os.path.join("tests", "conf", f"{name}.yaml"))
         conf_dict = OmegaConf.to_object(conf.experiment)
         conf_dict = cast(Dict[Any, Dict[Any, Any]], conf_dict)
@@ -36,11 +34,6 @@ class TestObjectDetectionTask:
         trainer = Trainer(fast_dev_run=True, log_every_n_steps=1, max_epochs=1)
         trainer.fit(model=model, datamodule=datamodule)
         trainer.test(model=model, datamodule=datamodule)
-
-        if isinstance(getattr(datamodule, "plot"), Callable):  # type: ignore[arg-type]
-            monkeypatch.delattr(classname, "plot")
-            datamodule2 = classname(**datamodule_kwargs)
-            trainer.validate(model=model, datamodule=datamodule2)
 
     @pytest.fixture
     def model_kwargs(self) -> Dict[Any, Any]:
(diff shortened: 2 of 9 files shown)


ashnair1:
> For torchvision dependency version, see what I did in https://github.com/microsoft/torchgeo/pull/631 for the other trainers. It's possible to check the version of torchvision and use different syntax for each version. It's kinda ugly and I don't like it, but I'm not sure if it's wise to just drop support for all older torchvision versions. Open to feedback on this though.

Saw this snippet. 
```
if parse(torchvision.__version__) >= parse("0.12"):
    if backbone_pretrained:
        kwargs = {
            "weights": getattr(
                torchvision.models, f"ResNet{backbone[6:]}_Weights"
            ).DEFAULT
        }
    else:
        kwargs = {"weights": None}
  else:
    kwargs = {"pretrained": backbone_pretrained}
```

This can work. Since the number of backbones supported also include the `ResNext` and `Wide_Resnet` variants, I opted to create  a mapping between backbones and weights. Perhaps this could be moved to a common location in a later PR.


Agree with not dropping support. We only need to drop a version if it breaks something or if it requires really ugly code to handle it. Think we're good for now. 

> For the torchmetrics dependency version, what was the reason for the version bump? It seems like MeanAveragePrecision was introduced in 0.7.0 (renamed from MAP) according to the changelog.

You're right. Think it was just the filename that was changed (`map.py` -> `mean_ap.py`). Reverted that commit.

adamjstewart:
Another option would be to change the `pretrained` kwarg on our end so that instead of taking a boolean it takes whatever torchvision wants. Then it's up to the user to decide which pretrained model to use. Thoughts?

ashnair1:
> Another option would be to change the pretrained kwarg on our end so that instead of taking a boolean it takes whatever torchvision wants. Then it's up to the user to decide which pretrained model to use. Thoughts?

Sure, that could work. That way the user would be able to specify the weights to be used and if they leave it unspecified, we can just use the default weights.

adamjstewart:
Hmm, but then it becomes difficult to store this configuration in a YAML file...

ashnair1:
Right. The YAML would only be informative if you're using torchvision > 0.12

adamjstewart:
But even if the YAML is torchvision > 0.12 specific, can you store objects in YAML or only strings/floats/bools? We could store a string and exec it but that feels like a security risk...

ashnair1:
The weights in the YAML can be string. We could just specify it like this.

``` diff
experiment:
  task: "nasa_marine_debris"
  name: "nasa_marine_debris_test"
  module:
    detection_model: "faster-rcnn"
    backbone: "resnet50"
+   pretrained: "ResNet50_Weights.IMAGENET1K_V1"
    num_classes: 2
    learning_rate: 1.2e-4
    verbose: false
  datamodule:
    root_dir: "data/nasamr/nasa_marine_debris"
    batch_size: 4
    num_workers: 56
    val_split_pct: 0.2
```

The pretrained arg can then be passed into the `weights` arg

adamjstewart:
Ah, we could use `torchvision.prototype.models.get_weight` to convert the string to the enum. But yeah, it would be a bit frustrating for the YAML to only work properly with torchvision 0.12.

Let's keep it like this for now but strongly consider dropping support for older torchvision at some point.

ashnair1:
The `pretrained` arg will officially be dropped in version `0.15`. Perhaps that would be a good time to drop old versions and peg minimum to `0.13`.

ashnair1:
Coverage drop is mainly caused by the code required for supporting minimum versions. 

On that note, has anybody looked into combining the coverage reports for normal tests and minimal tests via [`coverage combine`](https://coverage.readthedocs.io/en/6.4.3/cmd.html#combining-data-files-coverage-combine) before uploading to codecov?

adamjstewart:
We run tests on both the latest version of dependencies and the minimum version of dependencies. Both tests generate coverage reports, and should automatically be merged when uploaded to codecov. This works on all of our other PRs, not sure why coverage is dropping for you.

ashnair1:
This is weird. The coverage drop seems to be caused by unrelated files. Also a warning is shown stating there are unmerged base commits. Any ideas?

image

adamjstewart:
> The coverage drop seems to be caused by unrelated files.

This usually means that coverage has increased in main since this branch was created. Usually, rebase or merge commits solve this. Wish codecov was smarter about this...

ashnair1:
Since I've opened this PR, there have been a total of 4 commits merged to main. The two tutorial commits, one related to Sentinel2 and the torchvision constraint fix so it's unlikely a coverage increase is the reason.

I wonder if this has to do with me rebasing and squashing commits in this branch.

adamjstewart:
Yep, I have no idea what's going on here. It's just this PR, right? Probably best to reach out to codecov and see if this is a bug in codecov.

I'm focusing on 0.3.1 stuff at the moment, but if this is still broken after 0.3.1 is finished I can try to report it to codecov myself.

isaaccorley:
I went down the torchvision.detection rabbit hole awhile back. I think my only complaint with it was that they don't support negative sample images that don't have any objects in them during training. Do you know if this has changed? I need to catch up on all the releases.

isaaccorley:
Also I believe they added support for constructing a resnet fpn backbone from a pretrained torchvision resnet model. @calebrob6 and @anthonymlortiz, this might be of interest because we could add support for pretrained SSL resnet backbones to this task.

ashnair1:
@isaaccorley

> I went down the torchvision.detection rabbit hole awhile back. I think my only complaint with it was that they don't support negative sample images that don't have any objects in them during training. Do you know if this has changed? I need to catch up on all the releases.

They support negative samples. Support was added in [`0.6.0`](https://github.com/pytorch/vision/releases/tag/v0.6.0) via pytorch/vision@e75b497


> Also I believe they added support for constructing a resnet fpn backbone from a pretrained torchvision resnet model. @calebrob6 and @anthonymlortiz, this might be of interest because we could add support for pretrained SSL resnet backbones to this task.

This is supported as well. Just need to pass the backbone name and the `pretrained` flag. 

They recently added (in `v0.13.0`) the Multi Weights API which allows you to select which weights your backbone should load. Adam opened an issue to track this - #762 


I really wish `torchvision` would add support for oriented/rotated bounding boxes i.e. `(x, y, w, h, a)`. Orientation is a significant factor when you're dealing with object detection in RS imagery. Especially if your objects are small and clustered together. `detectron2` supports [RotatedBoxes](https://detectron2.readthedocs.io/en/latest/_modules/detectron2/structures/rotated_boxes.html) and the necessary ops like [rotated nms](https://detectron2.readthedocs.io/en/latest/_modules/detectron2/layers/nms.html), [rotated ROIAlign](https://detectron2.readthedocs.io/en/latest/_modules/detectron2/layers/roi_align_rotated.html), [Rotated RPN](https://github.com/facebookresearch/detectron2/blob/main/detectron2/modeling/proposal_generator/rrpn.py) etc and it works really well. Was able to get some nice results on the DOTA dataset using it. 

There is an open issue regarding porting over the necessary ops to `torchvision` (pytorch/vision#2761) but there doesn't seem to be much interest :disappointed:.

isaaccorley:
> I really wish `torchvision` would add support for oriented/rotated bounding boxes i.e. `(x, y, w, h, a)`. Orientation is a significant factor when you're dealing with object detection in RS imagery. Especially if your objects are small and clustered together. `detectron2` supports [RotatedBoxes](https://detectron2.readthedocs.io/en/latest/_modules/detectron2/structures/rotated_boxes.html) and the necessary ops like [rotated nms](https://detectron2.readthedocs.io/en/latest/_modules/detectron2/layers/nms.html), [rotated ROIAlign](https://detectron2.readthedocs.io/en/latest/_modules/detectron2/layers/roi_align_rotated.html), [Rotated RPN](https://github.com/facebookresearch/detectron2/blob/main/detectron2/modeling/proposal_generator/rrpn.py) etc and it works really well. Was able to get some nice results on the DOTA dataset using it.
> 
> There is an open issue regarding porting over the necessary ops to `torchvision` ([pytorch/vision#2761](https://github.com/pytorch/vision/issues/2761)) but there doesn't seem to be much interest 😞.

@conor-wallace was also able to get good benchmark results on DOTA with detectron2 and we use it a ton for aerial imagery now. Though we would love to never touch detectron2 again if torchvision added rotated oriented support.

evaluate.py
```diff
@@ -106,6 +110,14 @@ def run_eval_loop(
             y = batch["mask"].to(device)
         elif "label" in batch:
             y = batch["label"].to(device)
+        elif "boxes" in batch:
+            y = [
+                {
+                    "boxes": batch["boxes"][i].to(device),
+                    "labels": batch["labels"][i].to(device),
+                }
+                for i in range(len(batch["image"]))
```
adamjstewart, line 119: I wonder if `torchgeo.datasets.unbind_samples` could help here

evaluate.py
```diff
@@ -176,6 +188,21 @@ def main(args: argparse.Namespace) -> None:
...
+            "learning_rate": model.hparams["learning_rate"],
+        }
+
+        test_row = {
+            "split": "test",
+            "detection_model": model.hparams["detection_model"],
+            "backbone": model.hparams["backbone"],
+            "learning_rate": model.hparams["learning_rate"],
+        }
+
```
adamjstewart, line 205: Can you get rid of this empty line so it's clear that these are all clauses of the same if-statement?

evaluate.py
```diff
@@ -241,6 +268,10 @@ def main(args: argparse.Namespace) -> None:
                 }
             )
```
adamjstewart, line 270: Same here

torchgeo/datamodules/nasa_marine_debris.py
```diff
@@ -138,3 +140,11 @@ def test_dataloader(self) -> DataLoader[Any]:
             shuffle=False,
             collate_fn=collate_fn,
         )
+
+    def plot(self, *args: Any, **kwargs: Any) -> plt.Figure:
+        """Run :meth:`torchgeo.datasets.NASAMarineDebris.plot`."""
+        try:
```
adamjstewart, line 146: Any chance we can simplify this? Maybe `s/dataset/self.dataset` right before we perform the split, so we can always access `self.dataset` at any time?

torchgeo/datasets/nasa_marine_debris.py
```diff
@@ -99,6 +99,12 @@ def __getitem__(self, index: int) -> Dict[str, Tensor]:
         boxes = self._load_target(self.files[index]["target"])
         sample = {"image": image, "boxes": boxes}
 
+        # Filter invalid boxes
+        w_check = (sample["boxes"][:, 2] - sample["boxes"][:, 0]) > 0
+        h_check = (sample["boxes"][:, 3] - sample["boxes"][:, 1]) > 0
+        indices = w_check & h_check
+        sample["boxes"] = sample["boxes"][indices]
```
adamjstewart, line 106: How many of the boxes were invalid?

@isaaccorley recently had to do the same thing for IDTReeS, not sure if we should try to reuse code for that.

torchgeo/trainers/detection.py
```diff
@@ -0,0 +1,249 @@
...
+from torch.utils.data import DataLoader
+from torchvision.models.detection import FasterRCNN
+from torchvision.models.detection.backbone_utils import resnet_fpn_backbone
+from torchvision.models.detection.rpn import AnchorGenerator
+from torchvision.ops import MultiScaleRoIAlign
+
+if parse(torchmetrics.__version__) >= parse("0.8.0"):
+    from torchmetrics.detection.mean_ap import MeanAveragePrecision
+else:
+    from torchmetrics.detection.map import MeanAveragePrecision  # type: ignore
```
adamjstewart, line 24: What is the type ignore here for?

torchgeo/trainers/detection.py
```diff
@@ -0,0 +1,249 @@
...
+
+if parse(torchvision.__version__) > parse("0.12"):
```
adamjstewart, line 26:
```suggestion
if parse(torchvision.__version__) >= parse("0.13"):
```

torchgeo/trainers/detection.py
```diff
@@ -0,0 +1,249 @@
...
+        Args:
+            outputs: list of items returned by validation_step
+        """
+        metrics = self.val_metrics.compute()
+        renamed_metrics = {f"val_{i}": metrics[i] for i in metrics.keys()}
+        self.log_dict(renamed_metrics)
+        self.val_metrics.reset()
+
+    def test_step(self, *args: Any, **kwargs: Any) -> None:
+        """Compute test map.
```
adamjstewart, line 202:
```suggestion
        """Compute test MAP.
```
?

adamjstewart (commented):
This looks great, just a few minor comments to see if we can simplify things a bit.

You can ignore the coverage decrease for now. Before merging I'll run the tests locally just to make sure that they're all false negatives and force merge if I have to. Still quite curious why it isn't working though.

torchgeo/trainers/detection.py
ashnair1: The following mypy error gets triggered without it 

```
torchgeo/trainers/detection.py:24: error: Name "MeanAveragePrecision" already defined (possibly by an import)  [no-redef]
```

evaluate.py
ashnair1: Think it could. I'll take a look.

ashnair1: It can work but we will still need to iterate over them and move them to specified device.

torchgeo/trainers/detection.py
adamjstewart: Can we use `# type: ignore[no-redef]`? I try to be as specific as possible with type ignores. Doesn't matter if it causes the line to take up multiple lines.

torchgeo/trainers/detection.py
```diff
@@ -0,0 +1,251 @@
...
+
+from ..datasets.utils import unbind_samples
+
+# https://github.com/pytorch/pytorch/issues/60979
+# https://github.com/pytorch/pytorch/pull/61045
+DataLoader.__module__ = "torch.utils.data"
+
+
+class ObjectDetectionTask(pl.LightningModule):
+    """LightningModule for object detection of images."""
```
adamjstewart, line 51: Needs `versionadded`

torchgeo/trainers/detection.py
```diff
@@ -0,0 +1,251 @@
...
+    def config_task(self) -> None:
+        """Configures the task based on kwargs parameters passed to the constructor."""
+        backbone_pretrained = self.hyperparams.get("pretrained", True)
+        if self.hyperparams["detection_model"] == "faster-rcnn":
+            if "resnet" in self.hyperparams["backbone"]:
+                kwargs = {
+                    "backbone_name": self.hyperparams["backbone"],
+                    "trainable_layers": self.hyperparams.get("trainable_layers", 3),
+                }
+                if parse(torchvision.__version__) > parse("0.12"):
```
adamjstewart, line 62:
```suggestion
                if parse(torchvision.__version__) >= parse("0.13"):
```

adamjstewart (dismissed):
LGTM, will merge after #780 is merged

torchgeo/datasets/nasa_marine_debris.py
isaaccorley: We can reuse the IDTReeS._filter_boxes method and make it a function so it can be used for both datasets. Doesn't have to be in this PR but it seems as though this may be a common issue with object detection datasets.

adamjstewart:
Can you rebase to resolve the merge conflict?

Still no idea why codecov has been buggy lately. I'll force merge if I need to.

adamjstewart:
The drop in coverage here is legit, I was able to reproduce it locally:
```console
$ pytest --cov=torchgeo/trainers --cov-report=term-missing tests/trainers/
=============================================================================================== test session starts ================================================================================================
platform darwin -- Python 3.9.13, pytest-7.1.3, pluggy-1.0.0
rootdir: /Users/ajstewart/torchgeo, configfile: pyproject.toml
plugins: anyio-3.6.1, nbmake-1.3.3, cov-3.0.0
collected 57 items                                                                                                                                                                                                 

tests/trainers/test_byol.py ....                                                                                                                                                                             [  7%]
tests/trainers/test_classification.py ..................                                                                                                                                                     [ 38%]
tests/trainers/test_detection.py ...                                                                                                                                                                         [ 43%]
tests/trainers/test_regression.py ....                                                                                                                                                                       [ 50%]
tests/trainers/test_segmentation.py ..................                                                                                                                                                       [ 82%]
tests/trainers/test_utils.py ..........                                                                                                                                                                      [100%]

---------- coverage: platform darwin, python 3.9.13-final-0 ----------
Name                                  Stmts   Miss  Cover   Missing
-------------------------------------------------------------------
torchgeo/trainers/__init__.py             9      0   100%
torchgeo/trainers/byol.py               141      1    99%   326
torchgeo/trainers/classification.py     164      0   100%
torchgeo/trainers/detection.py          101      5    95%   24, 66-68, 187-188
torchgeo/trainers/regression.py          87      1    99%   45
torchgeo/trainers/segmentation.py        99      0   100%
torchgeo/trainers/utils.py               52      0   100%
-------------------------------------------------------------------
TOTAL                                   653      7    99%


=============================================================================================== 57 passed in 52.34s ================================================================================================
```
See https://app.codecov.io/gh/microsoft/torchgeo/pull/758 for the pictorial representation.

calebrob6:
image

the try/except in the validation_step

adamjstewart:
There's another line too.

ashnair1:
The try except block in the val_step is the only one line left. Not sure how to test this. I'm guessing `test_segmentation.py` passes because it includes datamodules that don't have `plot` implemented so it would trigger the exception.

isaaccorley:
Maybe instead of try/except we should check if the datamodule has a plot method. Something like: if isinstance(getattr(datamodule, "plot"), Callable). Then in the test we can override the plot method to be None or something so it will cover both cases.

adamjstewart:
There's a reason for the try-except. Lightning is very much dynamically typed, and makes no guarantees that any of these attributes will exist:

* `self.trainer.datamodule`
* `datamodule.plot`
* `self.logger.experiment`

That's also the reason that we have to ignore mypy warnings on so many lines. The correct solution would be to use hasattr/getattr on all of these like @isaaccorley suggested. I was just lazy and wrapped everything in a single try-except. 

If you want to rework the code to use hasattr/getattr, I would create a new PR and do the same for existing trainers. We should probably add plot methods to all datamodules like we have to all datasets, so that will cause this same issue in other tests too.

If you want to keep using the try-except, my suggestion would be to add a test that monkeypatches the datamodule to remove the plot method. This will give you full coverage and ensure that the trainer works even for datamodules without a plot method.

tests/trainers/test_detection.py
```diff
@@ -0,0 +1,67 @@
...
+        # Instantiate model
+        model_kwargs = conf_dict["module"]
+        model = ObjectDetectionTask(**model_kwargs)
+
+        # Instantiate trainer
+        trainer = Trainer(fast_dev_run=True, log_every_n_steps=1, max_epochs=1)
+        trainer.fit(model=model, datamodule=datamodule)
+        trainer.test(model=model, datamodule=datamodule)
+
+        if isinstance(getattr(datamodule, "plot"), Callable):  # type: ignore[arg-type]
```
adamjstewart, line 40: Let's put this in a separate test function, there's no reason we need to test this for every datamodule.

ashnair1: How do you mean? Add this separately as `test_missing_attribute` because we don't want to mix this with `test_trainer`?

Currently `test_detection.py` only tests the `NasaMarineDebrisDatamodule`.

adamjstewart (approved):

adamjstewart:
@isaaccorley does this need a predict step?

torchgeo/trainers/detection.py
```diff
@@ -0,0 +1,255 @@
...
+            training loss
+        """
+        batch = args[0]
+        x = batch["image"]
+        batch_size = x.shape[0]
+        y = [
+            {"boxes": batch["boxes"][i], "labels": batch["labels"][i]}
+            for i in range(batch_size)
+        ]
+        loss_dict = self.forward(x, y)
```
adamjstewart, line 150: According to https://github.com/microsoft/torchgeo/pull/790#discussion_r984943775 we should use `self()` instead of `self.forward()`. I'll submit a PR to change existing trainers, but can you make that change here?

adamjstewart (dismissed):

isaaccorley:
> @isaaccorley does this need a predict step?

Ideally yes, we can have a predict step that returns the predicted boxes and class labels per sample.

adamjstewart:
Let's add that here before we forget, I want to make sure all of our trainers support the same things.

isaaccorley:
@ashnair1 Let me know if you want some help on it. I can walk you through the predict_steps I made for the classifiers.

ashnair1:
Added `predict_step` and corresponding test.

adamjstewart (approved):

PR #759, opened 2022-09-05
Tutorials: trainers require optional datasets

adamjstewart:
Separated this into a separate PR.

### DIFF
--- a/docs/tutorials/trainers.ipynb
+++ b/docs/tutorials/trainers.ipynb
@@ -43,7 +43,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


adamjstewart:
Tested this tutorial locally. It still fails, but the reason it fails is because we currently run `pip install torchgeo`, which installs 0.3.0, which installs radiant-mlhub 0.5.2, which doesn't work. Once we release, `pip install torchgeo` will install 0.3.1, which will install radiant-mlhub 0.4.1, which will work. So this mystery has been solved.

calebrob6 (approved):

Issue #684, opened 2022-07-21
IDTReeS bugs


ashnair1:
### Description

While using IDTReeS to train an object detector, I noticed quite a few bugs within this dataset.

### 1. Coordinates are swapped
           
Fixed by #683

![OSBS_27_before](https://user-images.githubusercontent.com/32295036/180204371-2a1b19b8-9f47-421f-a5b9-9c82a2367f64.png)

The above image is a comparison of the actual ground truth of **OSBS_27** and the plot produced by torchgeo.

### 2. Missing polygon 

Shapefile contains 5 polygons whereas the plot only shows 4. This issue seems to be at the dataset level.

![OSBS_27_missingpoly](https://user-images.githubusercontent.com/32295036/180206989-3de7b45d-ddac-4cb1-abc0-505c9152b971.png)

### 3. Negative box coordinates in pixel space 

This is a weird one. I noticed that some boxes had negative pixel coordinates. Observed this in **MLBS_5** but its possible this could be present in others as well.

Comparing QGIS and the *corrected* torchgeo plot for **MLBS_5** didn't reveal much

![MLBS_5_initial](https://user-images.githubusercontent.com/32295036/180208170-17ef0487-a45b-46ef-8e3d-5f04acb1fcb1.png)

As you can see, the top 4 polygons are missing. But unlike the previous issue, these polygons do exist. However they seem to be outside the bounds of the image.

![MLBS_5_second](https://user-images.githubusercontent.com/32295036/180208315-718be8df-9182-4237-b9d7-7b4c7f35c7b9.png)

This could explain the negative coordinates but I'm not sure as to why there is a difference between the plots.

### Steps to reproduce

Most of these issues can be seen by downloading the IDTReeS dataset and comparing the torchgeo plots vs the shapefiles.

### Version

0.4.0.dev0

adamjstewart:
@ashnair1 @isaaccorley any thoughts on 2 and 3? I would like to fix these issues in the 0.3.1 (coming in approximately 1 week).

Could 3 explain 2? In example 2, the missing patch seems to be right on the edge of the image.

isaaccorley:
I think fixing the coordinates would likely fix the issue. I would bet that our bbox plotting method doesn't support boxes that are out of image bounds

adamjstewart:
May also be related to #679, we use `int()` instead of `round()` after converting geometries. Let me fix that and we can see if it fixes this too.

ashnair1:
[...]

PR #683, opened 2022-07-21
Fix coordinate swap bug in IDTReeS

ashnair1:
Image used: **OSBS_27**

Before:
OSBS_27_before

After:
OSBS_27_after

PR #760, opened 2022-09-05
IDTReeS - Clip boxes outside of image bounds

isaaccorley:
This PR
- clips the bounding box coordinates to the image size (essentially clips boxes that go outside of the image bounds) and filters any boxes that have zero area after clipping.
- fixed bug where id of some geometries were being overriden resulting in some images have no objects.

Fixes part of #684

For `MLBS_5` the boxes before

```python
tensor([[107, -57, 165, -10],
        [157, -48, 190, -10],
        [127,  24, 198,  92],
        [149, 122, 199, 174],
        [ 30,  70,  70, 109],
        [ -6, -80,  48, -10],
        [ 47, 121, 114, 197],
        [119,  84, 144, 129],
        [108,  74, 129, 101],
        [ 36,  21,  83,  73],
        [137,  73, 187, 128],
        [ 38, -63,  81, -10],
        [ 19,  93,  62, 152],
        [ 71,  81, 117, 133]])
```

and after filtering

```python
tensor([[127,  24, 198,  92],
        [149, 122, 199, 174],
        [ 30,  70,  70, 109],
        [ 47, 121, 114, 197],
        [119,  84, 144, 129],
        [108,  74, 129, 101],
        [ 36,  21,  83,  73],
        [137,  73, 187, 128],
        [ 19,  93,  62, 152],
        [ 71,  81, 117, 133]])
```

Visual result:
image

### DIFF
--- a/torchgeo/datasets/idtrees.py
+++ b/torchgeo/datasets/idtrees.py
@@ -289,10 +289,10 @@ class IDTReeS(NonGeoDataset):
         with rasterio.open(path) as f:
             for geom in geoms:
                 coords = [f.index(x, y) for x, y in geom]
-                xmin = min(coord[0] for coord in coords)
-                xmax = max(coord[0] for coord in coords)
-                ymin = min(coord[1] for coord in coords)
-                ymax = max(coord[1] for coord in coords)
+                xmin = min(coord[1] for coord in coords)
+                xmax = max(coord[1] for coord in coords)
+                ymin = min(coord[0] for coord in coords)
+                ymax = max(coord[0] for coord in coords)
                 boxes.append([xmin, ymin, xmax, ymax])
 
         tensor = torch.tensor(boxes)
--- a/torchgeo/datasets/idtrees.py
+++ b/torchgeo/datasets/idtrees.py
@@ -284,11 +284,15 @@ class IDTReeS(NonGeoDataset):
         geometries = cast(Dict[int, Dict[str, Any]], self.geometries)
 
         # Find object ids and geometries
+        # The train set geometry->image mapping is contained
+        # in the train/Field/itc_rsFile.csv file
         if self.split == "train":
             indices = self.labels["rsFile"] == base_path
             ids = self.labels[indices]["id"].tolist()
             geoms = [geometries[i]["geometry"]["coordinates"][0][:4] for i in ids]
-        # Test set - Task 2 has no mapping csv. Mapping is inside of geometry
+        # The test set has no mapping csv. The mapping is inside of the geometry
+        # properties i.e. geom["property"]["plotID"] contains the RGB image filename
+        # Return all geometries with the matching RGB image filename of the sample
         else:
             ids = [
                 k
@@ -397,9 +401,10 @@ class IDTReeS(NonGeoDataset):
         for path in filepaths:
             with fiona.open(path) as src:
                 for feature in src:
+                    # The train set has a unique id for each geometry in the properties
                     if self.split == "train":
                         features[feature["properties"]["id"]] = feature
-                    # Test set task 2 has no id
+                    # The test set has no unique id so create a dummy id
                     else:
                         id = len(features) + 1
                         features[id] = feature
--- a/torchgeo/datasets/idtrees.py
+++ b/torchgeo/datasets/idtrees.py
@@ -5,7 +5,7 @@
 
 import glob
 import os
-from typing import Any, Callable, Dict, List, Optional, Tuple, cast
+from typing import Any, Callable, Dict, List, Optional, Tuple, cast, overload
 
 import fiona
 import matplotlib.pyplot as plt
@@ -214,20 +214,20 @@ class IDTReeS(NonGeoDataset):
                 sample["boxes"] = self._load_boxes(path)
                 h, w = sample["image"].shape[1:]
                 sample["boxes"], _ = self._filter_boxes(
-                    image_size=(h, w), boxes=sample["boxes"]
+                    image_size=(h, w), min_size=1, boxes=sample["boxes"], labels=None
                 )
         else:
-            boxes = self._load_boxes(path)
-            labels = self._load_target(path)
+            sample["boxes"] = self._load_boxes(path)
+            sample["label"] = self._load_target(path)
 
             h, w = sample["image"].shape[1:]
-            boxes, labels = self._filter_boxes(  # type: ignore[assignment]
-                image_size=(h, w), boxes=boxes, labels=labels
+            sample["boxes"], sample["label"] = self._filter_boxes(
+                image_size=(h, w),
+                min_size=1,
+                boxes=sample["boxes"],
+                labels=sample["label"],
             )
-            sample["boxes"] = boxes
-            sample["label"] = labels
 
-        # Filter boxes
         if self.transforms is not None:
             sample = self.transforms(sample)
 
@@ -405,12 +405,24 @@ class IDTReeS(NonGeoDataset):
                         features[id] = feature
         return features
 
+    @overload
+    def _filter_boxes(
+        self, image_size: Tuple[int, int], min_size: int, boxes: Tensor, labels: Tensor
+    ) -> Tuple[Tensor, Tensor]:
+        ...
+
+    @overload
+    def _filter_boxes(
+        self, image_size: Tuple[int, int], min_size: int, boxes: Tensor, labels: None
+    ) -> Tuple[Tensor, None]:
+        ...
+
     def _filter_boxes(
         self,
         image_size: Tuple[int, int],
+        min_size: int,
         boxes: Tensor,
-        min_size: int = 1,
-        labels: Optional[Tensor] = None,
+        labels: Optional[Tensor],
     ) -> Tuple[Tensor, Optional[Tensor]]:
         """Clip boxes to image size and filter boxes with sides less than ``min_size``.
 
--- a/torchgeo/datasets/idtrees.py
+++ b/torchgeo/datasets/idtrees.py
@@ -396,12 +396,13 @@ class IDTReeS(NonGeoDataset):
         features: Dict[int, Dict[str, Any]] = {}
         for path in filepaths:
             with fiona.open(path) as src:
-                for i, feature in enumerate(src):
+                for feature in src:
                     if self.split == "train":
                         features[feature["properties"]["id"]] = feature
                     # Test set task 2 has no id
                     else:
-                        features[i] = feature
+                        id = len(features) + 1
+                        features[id] = feature
         return features
 
     def _filter_boxes(
--- a/torchgeo/datasets/idtrees.py
+++ b/torchgeo/datasets/idtrees.py
@@ -14,6 +14,7 @@ import rasterio
 import torch
 from rasterio.enums import Resampling
 from torch import Tensor
+from torchvision.ops import clip_boxes_to_image, remove_small_boxes
 from torchvision.utils import draw_bounding_boxes
 
 from .geo import NonGeoDataset
@@ -212,15 +213,15 @@ class IDTReeS(NonGeoDataset):
             if self.task == "task2":
                 sample["boxes"] = self._load_boxes(path)
                 w, h = sample["image"].shape[1:]
-                sample["boxes"], _ = filter_boxes(
+                sample["boxes"], _ = self._filter_boxes(
                     image_size=(h, w), boxes=sample["boxes"]
                 )
         else:
             sample["boxes"] = self._load_boxes(path)
             sample["label"] = self._load_target(path)
 
-            w, h = sample["image"].shape[1:]
-            sample["boxes"], sample["label"] = filter_boxes(  # type:ignore[assignment]
+            h, w = sample["image"].shape[1:]
+            sample["boxes"], sample["label"] = self._filter_boxes(
                 image_size=(h, w), boxes=sample["boxes"], labels=sample["label"]
             )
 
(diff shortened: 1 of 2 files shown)


ashnair1:
The coordinate swap was already done in #683

isaaccorley:
@ashnair1 Thanks for pointing that out. Totally missed it.

torchgeo/datasets/idtrees.py
```diff
@@ -296,6 +296,7 @@ def _load_boxes(self, path: str) -> Tensor:
                 boxes.append([xmin, ymin, xmax, ymax])
 
         tensor = torch.tensor(boxes)
+        tensor = torch.clamp(tensor, min=0, max=self.image_size[0])
```
weiji14, line 299: This assumes that images are always square (i.e. `self.image_size[0] == self.image_size[1]`). Is this always the case for this IDTReeS dataset?

Thinking about whether [`torchvision.ops.clip_boxes_to_image`](https://pytorch.org/vision/0.13/generated/torchvision.ops.clip_boxes_to_image.html#torchvision.ops.clip_boxes_to_image) would work here, or if it's better to keep things simple.

isaaccorley: Yes, if you check load_image we read them in as 200x200.

I'm indifferent on using clip_boxes_to_image since it appears to perform the same functionality. I would just make sure it's not a recent addition to torchvision so it makes our dependencies strict.

weiji14: Cool, if the images are always square, then this addition is ok. Just hope the next person adding an object detection dataset is careful when copying this code :laughing:

weiji14 (approved):

torchgeo/datasets/idtrees.py
adamjstewart: I also had the same question about square vs rectangular, can you add a comment saying the image will always be square?

isaaccorley:
Created generic `torchgeo.datasets.utils.filter_boxes` function in case this occurs in other object detection datasets. Also refactored to not hardcode the image size but to use the current sample image (h, w).

torchgeo/datasets/idtrees.py
```diff
@@ -211,10 +211,20 @@ def __getitem__(self, index: int) -> Dict[str, Tensor]:
         if self.split == "test":
             if self.task == "task2":
                 sample["boxes"] = self._load_boxes(path)
+                w, h = sample["image"].shape[1:]
+                sample["boxes"], _ = filter_boxes(
+                    image_size=(h, w), boxes=sample["boxes"]
```
weiji14, line 216: Doesn't matter if it's square, but should it both be `h, w`?

```suggestion
                h, w = sample["image"].shape[1:]
                sample["boxes"], _ = filter_boxes(
                    image_size=(h, w), boxes=sample["boxes"]
```

torchgeo/datasets/idtrees.py
```diff
@@ -211,10 +211,20 @@ def __getitem__(self, index: int) -> Dict[str, Tensor]:
...
+                )
         else:
             sample["boxes"] = self._load_boxes(path)
             sample["label"] = self._load_target(path)
 
+            w, h = sample["image"].shape[1:]
+            sample["boxes"], sample["label"] = filter_boxes(  # type:ignore[assignment]
+                image_size=(h, w), boxes=sample["boxes"], labels=sample["label"]
```
weiji14, line 224:
```suggestion
            h, w = sample["image"].shape[1:]
            sample["boxes"], sample["label"] = filter_boxes(  # type:ignore[assignment]
                image_size=(h, w), boxes=sample["boxes"], labels=sample["label"]
```

torchgeo/datasets/utils.py
```diff
@@ -707,3 +709,32 @@ def percentile_normalization(
...
+    Args:
+        image_size: tuple of (height, width) of image
+        min_size: filter boxes that have any side less than min_size
+        boxes: [N, 4] shape tensor of xyxy bounding box coordinates
+        labels: (Optional) [N,] shape tensor of bounding box labels
+
+    Returns:
+        a tuple of filtered boxes and labels
+
+    .. versionadded:: 0.3.1
```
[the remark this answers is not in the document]
isaaccorley: I could put it directly in the dataset as a method with a note to move to it's own function in 0.4.0. Does that work?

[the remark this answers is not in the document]
adamjstewart: That's fine by me. I wouldn't move it to `utils` unless multiple datasets need it.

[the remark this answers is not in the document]
isaaccorley: Okay sounds good

torchgeo/datasets/idtrees.py
isaaccorley: Fixed

torchgeo/datasets/idtrees.py
isaaccorley: Fixed

torchgeo/datasets/idtrees.py
```diff
@@ -391,6 +404,35 @@ def _load_geometries(self, directory: str) -> Dict[int, Dict[str, Any]]:
...
+        Args:
+            image_size: tuple of (height, width) of image
+            min_size: filter boxes that have any side less than min_size
+            boxes: [N, 4] shape tensor of xyxy bounding box coordinates
+            labels: (Optional) [N,] shape tensor of bounding box labels
+
+        Returns:
+            a tuple of filtered boxes and labels
+
+        .. versionadded:: 0.3.1
```
adamjstewart, line 425: You can remove this line

torchgeo/datasets/idtrees.py
```diff
@@ -383,14 +396,44 @@ def _load_geometries(self, directory: str) -> Dict[int, Dict[str, Any]]:
...
         for path in filepaths:
             with fiona.open(path) as src:
-                for i, feature in enumerate(src):
+                for feature in src:
                     if self.split == "train":
                         features[feature["properties"]["id"]] = feature
                     # Test set task 2 has no id
                     else:
-                        features[i] = feature
+                        id = len(features) + 1
```
isaaccorley, line 404: Found a phantom bug here making some of the boxes be empty due to being overriden and not existing in `self.geometries`

torchgeo/datasets/idtrees.py
```diff
@@ -383,14 +396,42 @@ def _load_geometries(self, directory: str) -> Dict[int, Dict[str, Any]]:
...
+                        features[id] = feature
         return features
 
+    def _filter_boxes(
```
adamjstewart, line 408: I think the correct way to do this is the unfortunately verbose:
```python
from typing import override

@override
def _filter_boxes(self, image_size: Tuple[int, int], boxes: Tensor, min_size: int, labels: Tensor) -> Tuple[Tensor, Tensor]:
    ...

@override 
def _filter_boxes(self, image_size: Tuple[int, int], boxes: Tensor, min_size: int, labels: None) -> Tuple[Tensor, None]:
    ...
```
followed by the same code below. Basically, if the input labels is Tensor, the output labels is also Tensor, and if it's None it's None. Then you can remove the `type: ignore` above.

torchgeo/datasets/idtrees.py
adamjstewart: I'm gonna be honest, I don't really understand what `_load_geometries` does and specifically what `features` is supposed to contain. It seems like a nested dict, but what are the keys and what are the values? Also, I see that we're using `Any` and `cast` in several places, which is usually bad. Can you elaborate on this and maybe add more comments documenting it? I want to make sure you're not the only one who understands how this Dataset works.

weiji14 (dismissed):
Code logic and docstrings seems to be sound. Did you check the image and box alignment? I'm not quite sure how to interpret the sample image you posted at https://github.com/microsoft/torchgeo/pull/760#issue-1362426866, the colors look a bit weird :sweat_smile:

isaaccorley:
> #684

Yes. I don't select the colors. torchvision's `draw_bounding_boxes` will randomly select box colors based on the number of classes in the image since I don't provide any colors.

torchgeo/datasets/idtrees.py
isaaccorley: It's a dictionary of geojson like features (basically a dict with properties and a geometry). For the train set these geometries have unique ids inside of the shapely file. However, for the test set they have no ids, so I have to make one up. In this case I just make one that increments every time a new feature is added.

isaaccorley: Added more comments for clarity

isaaccorley:
Just wanted to add some notes for everyone's clarity on the dataset.

- The test set - task 1 has no boxes or labels only RGB/HSI/CHM images and LAS point clouds
- The test set - task 2 shapefiles (e.g. task2/ITC/test_MLBS.shp) and contains the image file the boxes belong to in the properties, e.g. below in `properties["plotID"]` the image file is `MLBS_1.tif`.

```python
{
  'geometry': {'coordinates': [[(541886.6194198467, 4136580.9452913683),
                                 (541886.6194198467, 4136588.0890556537),
                                 (541891.3025542134, 4136588.0890556537),
                                 (541891.3025542134, 4136580.9452913683),
                                 (541886.6194198467, 4136580.9452913683)]],
                'type': 'Polygon'},
   'id': '0',
   'properties': OrderedDict([('indvdID', 'MLBSE00048'),
                              ('plotID', 'MLBS_1.tif')]),
   'type': 'Feature'
}
```

- The train set shapefiles (e.g. train/ITC/train_MLBS.shp) contain the boxes but **do not** contain the image file. However it does contain a unique geometry id `properties["id"]`.

```python
{
  'geometry': {'coordinates': [[(542071.1868343, 4134996.0734410095),
                                 (542071.1868343, 4134999.0),
                                 (542074.96994003, 4134999.0),
                                 (542074.96994003, 4134996.0734410095),
                                 (542071.1868343, 4134996.0734410095)]],
                'type': 'Polygon'},
   'id': '0',
   'properties': OrderedDict([('id', 57), ('indvdID', 'MLBS01681')]),
   'type': 'Feature'
}
```

The image <-> geometry id mapping is inside `train/Field/itc_rsFile.csv`.


```python
"id","indvdID","rsFile"
57,"MLBS01681","MLBS_1.tif"
45,"MLBS01679","MLBS_1.tif"
49,"MLBS01674","MLBS_1.tif"
63,"MLBS01678","MLBS_1.tif"
...
```

Note that the test set **does not** have any csv files only the train set. Because of this, the implementation for parsing the dataset can become quite verbose and requires different logic for the both train set/test set - task1/ test set - task2. 

Hopefully this clarifies some of the implementation decisions.

torchgeo/datasets/idtrees.py
isaaccorley: - `self.images` is a globbed list of image paths
- `self.geometries` is a dict where each entry maps a unique id to a geometry and it's properties. (test set - task 1 does not have any geometries so it's `None` for this case.
- `self.labels` is the `itc_rsFile.csv` file read into a pandas dataframe (this is only valid for the train set, so for all other cases it is `None`)

That's correct, since there are multiple shapefiles, the test set geometries were being overriden.

My initial plan was to use the `properties["indvdID"]` however I found that there were duplicates so I couldn't use this as the id so the next option was to just make a dummy id.

We could use further optimize by using a counter `i = 0` but `self._load_geometries` is only called once on instantiation of the dataset so the loop only occurs once. If you prefer I add this then that is fine with me.

adamjstewart: Do we need a dict for geometries? Could we use a list instead?

isaaccorley: The dict makes it easy for me to grab the ids and then just index the dict for the geometries needed for the image. I could revamp the dataset and just have a list of [image_path, list_of_geometries_per_image, and list_of_labels_per_image] instead?

adamjstewart: I think I still don't understand. If `self.images`, `self.geometries`, and `self.labels` don't share the same keys/order, how do you map from one to the other? I'm trying to follow the path from `__getitem__` index to file loading but there are a lot of moving parts.

isaaccorley: getitem[idx] -> image = self.images[idx]  -> self.labels[image] -> ids -> self.labels AND self.geometries

isaaccorley: I agree this likely needs a refactor to simplify things however, this PR is for bug fixes. The dataset currently works fine as is. Would you like to refactor for 0.4.0?

adamjstewart: Okay, but if `self.geometries[image]`, then why does it feel like the id in `features[id]` is completely arbitrary and not the name of the image file?

isaaccorley: The id is only arbitrary for the test set not the train set. You need the id to index the csv file to find the label and image filename.

adamjstewart: Wait, so `self._geometries()` is only called on the test set? Because it looks like the id is arbitrary any time this function is called...

adamjstewart: Nvm, I just noticed the if statement.

adamjstewart (changes requested):
Okay, I think I finally understand the code well enough to review this now!

So if I understand correctly, this PR only solves issue (3) in #684. Once this PR is merged, A) all bounding boxes will be clipped to the image bounds which will fix plotting, and B) bounding boxes won't be accidentally deleted from the testing set due to integer overlap.

My only change request is to replace the current `id = len(features) + 1` logic with a counter `i = 0, i += 1` so we aren't computing the length of a list in each iteration.

torchgeo/datasets/idtrees.py
isaaccorley: I can follow up with a refactor in another PR. I think what I will do is on instantiation do all the ugly parsing and create a list of tuples of (image_path, boxes, box_labels) and that should be less convoluted when getitem is actually called.

isaaccorley:
> Okay, I think I finally understand the code well enough to review this now!
> 
> So if I understand correctly, this PR only solves issue (3) in #684. Once this PR is merged, A) all bounding boxes will be clipped to the image bounds which will fix plotting, and B) bounding boxes won't be accidentally deleted from the testing set due to integer overlap.
> 
> My only change request is to replace the current `id = len(features) + 1` logic with a counter `i = 0, i += 1` so we aren't computing the length of a list in each iteration.

I've made the requested change. Also note that this PR doesn't just fix plotting but also allows a user to actually train on the dataset whereas before you would have issues dealing with boxes with negative values.

adamjstewart (approved):

PR #761, opened 2022-09-06
Fix torchvision constraint

ashnair1:
Multi-Weights API was introduced in `0.13`. This PR adjusts the constraint accordingly.

### DIFF
--- a/tests/trainers/conftest.py
+++ b/tests/trainers/conftest.py
@@ -18,7 +18,7 @@ from torch.nn.modules import Module
 @pytest.fixture(scope="package")
 def model() -> Module:
     kwargs: Dict[str, Optional[bool]] = {}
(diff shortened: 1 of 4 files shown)


adamjstewart (changes requested):
Don't know how I missed this, thanks for catching this!

Can you use `>= parse("0.13")` instead? `> parse("0.12")` also includes 0.12.1 which would not have this feature.

ashnair1:
No problem :+1: . `torchvision` doesn't have a `0.12.1` though. Refer [pypi](https://pypi.org/project/torchvision/#history)

adamjstewart:
Not _yet_. And it probably won't, but they _could_ release a patch release to fix an important bug.

adamjstewart (approved):

PR #767, opened 2022-09-07
Fix transforms notebook

calebrob6:
Closes https://github.com/microsoft/torchgeo/issues/763

### DIFF
--- a/docs/tutorials/transforms.ipynb
+++ b/docs/tutorials/transforms.ipynb
@@ -606,10 +606,15 @@
     ")\n",
     "\n",
     "device = \"cuda\" if torch.cuda.is_available() else \"cpu\"\n",
+    "\n",
+    "\n",
     "def get_batch_cpu():\n",
     "    return dict(image=torch.randn(64, 13, 512, 512).to(\"cpu\"))\n",
+    "\n",
+    "\n",
     "def get_batch_gpu():\n",
     "    return dict(image=torch.randn(64, 13, 512, 512).to(device))\n",
+    "\n",
     "\n",
     "transforms_gpu = deepcopy(transforms).to(device)"
    ]
--- a/docs/tutorials/transforms.ipynb
+++ b/docs/tutorials/transforms.ipynb
@@ -1,32 +1,4 @@
 {
- "nbformat": 4,
- "nbformat_minor": 0,
- "metadata": {
-  "colab": {
-   "name": "indices.ipynb",
-   "provenance": []
-  },
-  "interpreter": {
-   "hash": "6e850ee5f92358dcfdbb90dda05d686956eb0825584ddd5eff31b34875ddfee0"
-  },
-  "kernelspec": {
-   "display_name": "Python 3",
-   "language": "python",
-   "name": "python3"
-  },
-  "language_info": {
-   "codemirror_mode": {
-    "name": "ipython",
-    "version": 3
-   },
-   "file_extension": ".py",
-   "mimetype": "text/x-python",
-   "name": "python",
-   "nbconvert_exporter": "python",
-   "version": "3.8.11"
-  },
-  "accelerator": "GPU"
- },
  "cells": [
   {
    "cell_type": "markdown",
@@ -79,14 +51,14 @@
   },
   {
    "cell_type": "code",
+   "execution_count": null,
    "metadata": {
     "id": "wOwsb8KT_uXR"
    },
+   "outputs": [],
    "source": [
     "%pip install torchgeo"
-   ],
-   "execution_count": null,
-   "outputs": []
+   ]
   },
   {
    "cell_type": "markdown",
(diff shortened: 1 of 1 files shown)


calebrob6:
Explanation of what was wrong:
- the `MinMaxNormalize` class needed to have `nn.Parameters` so that it would go to the GPU on `.to(device)` correctly
- If you have `transforms = nn.Sequential(...)` then do `transforms_gpu = transforms.to(device)`, the stuff in `transforms` will also be on `device`. We want those to be separate so have to deepcopy first.

calebrob6:
If you run it on colab you can see a more drastic difference between CPU and GPU:

image

isaaccorley:
Thanks for catching this one! Totally forgot that the min/max tensors needed to be parameters to be put on the GPU...

docs/tutorials/transforms.ipynb
```diff
@@ -111,12 +85,11 @@
     "import torchvision.transforms as T\n",
     "from PIL import Image\n",
     "from torch.utils.data import DataLoader\n",
+    "from copy import deepcopy\n",
```
adamjstewart, line 88: I guess isort doesn't currently support jupyter notebooks like black does. I wouldn't bother doing this manually until it does unless you really care. Trying not to be too anal about stuff like that.

docs/tutorials/transforms.ipynb
```diff
@@ -260,30 +233,30 @@
...
-   ],
-   "execution_count": 4,
    "outputs": [
     {
-     "output_type": "stream",
      "name": "stdout",
+     "output_type": "stream",
      "text": [
-      "Number of images in dataset: 27000\n",
+      "Number of images in dataset: 16200\n",
```
adamjstewart, line 249: I'm guessing this changed because the original dataset had no train/val/test splits and these were added after the tutorial was written?

docs/tutorials/transforms.ipynb
```diff
@@ -619,60 +606,65 @@
...
+     "output_type": "stream",
      "text": [
-      "1 loop, best of 1: 12.3 s per loop\n"
+      "2.89 s ± 65.8 ms per loop (mean ± std. dev. of 5 runs, 1 loop each)\n"
      ]
     }
+   ],
+   "source": [
+    "%%timeit -n 1 -r 5\n",
+    "_ = transforms(get_batch_cpu())"
```
adamjstewart, line 643:
```suggestion
    "transforms(get_batch_cpu())"
```

docs/tutorials/transforms.ipynb
```diff
@@ -619,60 +606,65 @@
...
+     "output_type": "stream",
      "text": [
-      "1 loop, best of 1: 8.58 s per loop\n"
+      "1.91 s ± 395 ms per loop (mean ± std. dev. of 5 runs, 1 loop each)\n"
      ]
     }
+   ],
+   "source": [
+    "%%timeit -n 1 -r 5\n",
+    "_ = transforms_gpu(get_batch_gpu())"
```
adamjstewart, line 667:
```suggestion
    "transforms_gpu(get_batch_gpu())"
```

docs/tutorials/transforms.ipynb
```diff
@@ -79,14 +51,14 @@
   },
   {
    "cell_type": "code",
+   "execution_count": null,
```
adamjstewart, line 54:
```suggestion
   "execution_count": 1,
```

I think that will convert the current:
```
[
]:
```
to:
```
[1]:
```
which just looks nicer.

docs/tutorials/transforms.ipynb
```diff
@@ -703,39 +696,67 @@
...
+    "print(f\"Class Label: {dataset.classes[sample['label']]}\")\n",
+    "image.resize((256, 256), resample=Image.BILINEAR)"
    ]
   }
- ]
+ ],
+ "metadata": {
+  "accelerator": "GPU",
+  "colab": {
+   "name": "indices.ipynb",
```
adamjstewart, line 736: So THIS is why all of the tutorials have the wrong filename when I open them in Colab! I'll fix the other tutorials in a separate PR.
```suggestion
   "name": "transforms.ipynb",
```

adamjstewart (changes requested):
These timing comparisons look much better now, thanks! It's great to update tutorial outputs like this once in a while. nbsphinx has an option to do this automatically when you build the docs, but then build times are super slow and we have no control over things like removing pip install output.

docs/tutorials/transforms.ipynb
isaaccorley: I think this will end up printing the transformed tensor which is not what we want right?

adamjstewart: Oh true

adamjstewart (approved):
Remaining change requests also apply to other tutorials, I just fix all of them at the same time in a follow up PR.

PR #786, opened 2022-09-21
UnionDataset: fix __getitem__ bug

adamjstewart:
### What was the bug?

TL;DR: `UnionDataset` doesn't work and has never worked.

In `UnionDataset.__getitem__`, we were using the following check:
```python
if ds.index.intersection(tuple(query)):
    samples.append(ds[query])
```
The idea was that we shouldn't be trying to query from a dataset unless the dataset contains files that overlap with the query. 

However, this doesn't work. `rtree.index.Index.intersection` returns a generator, which always evaluates to True. This resulted in bugs like #769.

### What was the fix?

The fix is simple: just convert the generator to a list. The list will evaluate to False if it's empty, and the dataset won't be sampled from because it has no overlap.

### Why didn't the tests catch this?

Our existing tests were basically useless. Our `CustomGeoDatset.__getitem__` just returned the query, so even if the query didn't overlap at all with the dataset, it didn't raise an error. And our `UnionDataset` tests only tested the union of overlapping datasets, they never tested disparate geospatial locations.

I changed the tests to actually check the rtree index for the query, and to test the union of two disjoint datasets. The tests fail without the fix, and pass with the fix. Hopefully the new and improved tests will be more useful.

### DIFF
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -42,7 +42,10 @@ class CustomGeoDataset(GeoDataset):
         self.res = res
 
     def __getitem__(self, query: BoundingBox) -> Dict[str, BoundingBox]:
-        return {"index": query}
+        hits = self.index.intersection(tuple(query), objects=True)
+        hit = next(iter(hits))
+        bounds = BoundingBox(*hit.bounds)
+        return {"index": bounds}
 
 
 class CustomVectorDataset(VectorDataset):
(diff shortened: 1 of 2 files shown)


calebrob6:
I confirmed this fix works today by patching it into my code that uses a UnionDataset with a semantic segmentation trainer.

calebrob6 (approved):

PR #787, opened 2022-09-22
Remove "labels" folder from verify for xView2

RitwikGupta:
The `labels` folder exists for the `tier3` imagery that was released after the `tier1` imagery for xView2. It should really be called `targets` but we messed up. When you're merging `tier1` with `tier3`, simply copy all the files from `tier3/labels` to `tier1/targets`.

### DIFF
--- a/torchgeo/datasets/xview.py
+++ b/torchgeo/datasets/xview.py
@@ -188,7 +188,7 @@ class XView2(NonGeoDataset):
         # Check if the files already exist
         exists = []
         for split_info in self.metadata.values():
-            for directory in ["images", "labels", "targets"]:
+            for directory in ["images", "targets"]:
                 exists.append(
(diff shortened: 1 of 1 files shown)


isaaccorley:
We should also update the docs to explain to users that they need to manually move tier3 labels into the tier1 folder.

adamjstewart:
I would prefer that the TorchGeo dataset "just works" without the user needing to change anything on their part. If the download still has that directory in the latest version, we should support it.

RitwikGupta:
The download has that directory, but the dataset does not use it whatsoever. The user has to actually manually convert Tier3 JSON labels to masks, but the dataset doesn't do anything with the labels folder. Therefore, there's no reason to check for its existence.

adamjstewart (approved):

PR #790, opened 2022-09-23
ClassificationTask predict step

isaaccorley:
Running `trainer.predict()` with the `ClassificationTask` fails because the default pytorch_lightning `predict_step` expects a batch tuple and not a dict. This PR overrides `predict_step` to work with our classification datasets.

E.g.

```python

```

### DIFF
--- a/torchgeo/trainers/classification.py
+++ b/torchgeo/trainers/classification.py
@@ -249,7 +249,7 @@ class ClassificationTask(pl.LightningModule):
         """
         batch = args[0]
         x = batch["image"]
-        y_hat = self(x).softmax(dim=-1)
+        y_hat: Tensor = self(x).softmax(dim=-1)
         return y_hat
 
     def configure_optimizers(self) -> Dict[str, Any]:
--- a/torchgeo/trainers/classification.py
+++ b/torchgeo/trainers/classification.py
@@ -238,6 +238,20 @@ class ClassificationTask(pl.LightningModule):
         self.log_dict(self.test_metrics.compute())
         self.test_metrics.reset()
 
+    def predict_step(self, *args: Any, **kwargs: Any) -> Tensor:
+        """Compute and return the predictions.
+
+        Args:
+            batch: the output of your DataLoader
+
+        Returns:
+            predicted softmax probabilities
+        """
+        batch = args[0]
+        x = batch["image"]
+        y_hat = self(x).softmax(dim=-1)
+        return y_hat
+
     def configure_optimizers(self) -> Dict[str, Any]:
         """Initialize the optimizer and learning rate scheduler.
 
--- a/tests/trainers/test_classification.py
+++ b/tests/trainers/test_classification.py
@@ -61,6 +61,7 @@ class TestClassificationTask:
         trainer = Trainer(fast_dev_run=True, log_every_n_steps=1, max_epochs=1)
         trainer.fit(model=model, datamodule=datamodule)
         trainer.test(model=model, datamodule=datamodule)
+        trainer.predict(model=model, dataloaders=datamodule.val_dataloader())
 
     def test_no_logger(self) -> None:
         conf = OmegaConf.load(os.path.join("tests", "conf", "ucmerced.yaml"))


calebrob6:
I don't understand the failing test here.

adamjstewart (commented):
Should we add this to all trainers?

adamjstewart:
No idea what's going on with the failing test, the latest passing commit on master has the same versions of mypy and numpy, so I don't know why it would be failing all of a sudden.

adamjstewart:
Technically this could go in 0.3.2 since it doesn't actually add a new feature, it overrides an existing feature.

isaaccorley:
Looks like it's a bug introduced with the latest mypy that breaks with Python 3.10.7. The fix will come out in the next mypy release. https://github.com/python/mypy/issues/13627

isaaccorley:
The plan is to add this to all the trainers but each one will have it's own specific needs, e.g you may not want to accumulate predicted segmentation masks into a list in memory and would rather save them in some other way.

adamjstewart:
That is unfortunate, I guess dependabot can't control exactly which Python version we test. We could pin to 3.10.6 if we want to get the tests working.

isaaccorley:
@adamjstewart mypy v0.981 is out which includes a fix for this failing test. Should I update `requirements/tests.txt` in this PR or a separate one?

adamjstewart:
Dependabot will update it for you if you wait < 24 hrs (I think it checks once per day).

isaaccorley:
Not sure why codecov is complaining. When I go to the link it only shows coverage dips on scripts I didn't modify.

torchgeo/trainers/classification.py
```diff
@@ -238,6 +238,20 @@ def test_epoch_end(self, outputs: Any) -> None:
...
+        Args:
+            batch: the output of your DataLoader
+
+        Returns:
+            predicted softmax probabilities
+        """
+        batch = args[0]
+        x = batch["image"]
+        y_hat: Tensor = self(x).softmax(dim=-1)
+        return y_hat
```
adamjstewart, line 253: Why is the predict step so different from the train/val/test steps? Shouldn't it look something like:
```suggestion
        y_hat = self.forward(x)
        y_hat_hard = y_hat.argmax(dim=1)
        return y_hat_hard
```

calebrob6: Because you will frequently want the per-class probabilities instead of just the top class

calebrob6 (approved):

torchgeo/trainers/classification.py
adamjstewart: And that's not true for train/val/test?

isaaccorley: The train/val/test steps aren't meant to return data to a user. The operations that happen are only meant to perform the forward pass and compute metrics (which is why the argmax op is needed). The predict step needs to return the raw softmax outputs because we don't want to assume what a user wants to do with the output.

adamjstewart: Do we need `yhat: Tensor` or can we just use `yhat`?

isaaccorley: I had to add that to avoid mypy complaining.

adamjstewart: Why was it complaining? If you use `self.forward(x)` instead does it stop complaining? What if you break it into multiple lines like we do in other steps?

adamjstewart: (type hint warnings are not something to be ignored, they're something to be fixed)

calebrob6: - The recommended way to run inference is with `self(x)` not `self.forward(x)` as `self(x)` calls hooks while `self.forward(x)` silently doesn't.
- `y_hat` is a Tensor so I don't see the problem. Breaking something into multiple lines to get `mypy` to stop complaining is ridiculous overhead.

adamjstewart: Should we change other methods to replace `self.forward(x)` with `self(x)`? Just want to make sure we're consistent.

isaaccorley: Yes we should change. We shouldn't be calling forward anywhere.

adamjstewart: P.S. I feel like the `yhat: Tensor` thing is due to missing type hints for softmax but I couldn't figure out exactly where I need to add those. PyTorch's source code is a maze.

isaaccorley: I think I was able to fix it by calling it as a function e.g. torch.softmax(tensor) instead of calling it as a method e.g. tensor.softmax()

PR #792, opened 2022-09-26
MultiLabelClassificationTask predict step

isaaccorley:
This PR is similar to #790 and adds a `predict_step` method to `MultiLabelClassificationTask` so that users can utilize PyTorch Lightning's `trainer.predict()` function which will automatically loop and predict over a PyTorch DataLoader e.g.:

```python


isaaccorley:
Not sure why codecov is complaining. When I go to the link it only shows coverage dips on scripts I didn't modify.

calebrob6 (approved):

PR #798, opened 2022-09-28
Add VHR10DataModule

ashnair1:
- Added plot function for VHR10 
- Added test

Since this is a `CocoDataset` it might be better to subclass from torchvision's [CocoDataset](https://pytorch.org/vision/main/_modules/torchvision/datasets/coco.html) instead of `NonGeoDataset`. Thoughts?

Plot sample (split="positive"):

VHR10_small

Had to bump minimum version of `pyococotools` to `2.0.1` to support `draw_bbox` arg.


torchgeo/datasets/nwpu.py
```diff
@@ -244,3 +245,33 @@ def _download(self) -> None:
...
+    ) -> plt.Figure:
+        """Plot a sample from the dataset.
+
+        Args:
+            sample: a sample returned by :meth:`__getitem__`
+            suptitle: optional string to use as a suptitle
+            draw_boxes: draw only segmentation masks if False else draw boxes as well
+
+        Returns:
+            a matplotlib Figure with the rendered sample
```
adamjstewart, line 263: Needs versionadded

ashnair1:
That being said, I get where you're coming from. Plots from pycocotools are not as visually impressive, customizable or as informative as the second plot. 

I've written out a new plot function that does this. Still ironing out some details but here is a sample.

final_img

Left (Masks and Boxes)
Center (Only Masks with mask alpha = 0.2)
Right (Only boxes)

ashnair1:
Training a object detector on VHR10 now and plots look encouraging. Might need bump up the figure size. 

image

calebrob6:
@ashnair1 maybe try throwing in a `plt.tightlayout()` before returning `plt.gcf()`

ashnair1:
This PR will be split into smaller PRs once I make sure everything works as expected.

adamjstewart:
Note that the datamodule will need to be refactored once #992 is merged. But that should hopefully make it significantly simpler.

adamjstewart:
Closed by #847 and #1082

PR #802, opened 2022-09-30
Datasets: add default 'root' argument

adamjstewart:
Apparently not all of our datasets have a default `root` argument.

### DIFF
--- a/torchgeo/datasets/spacenet.py
+++ b/torchgeo/datasets/spacenet.py
@@ -69,8 +69,8 @@ class SpaceNet(NonGeoDataset, abc.ABC):
 
     def __init__(
         self,
-        root: str = "data",
-        image: str = "",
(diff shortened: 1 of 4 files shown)


torchgeo/datasets/spacenet.py
```diff
@@ -93,7 +93,10 @@ def __init__(
             RuntimeError: if ``download=False`` but dataset is missing
         """
         self.root = root
-        self.image = image  # For testing
+        if image:
+            self.image = image
+        else:
+            self.image = list(self.imagery.keys())[0]
```
adamjstewart, line 99: @ashnair1 had to hack your SpaceNet datasets a bit. `root` can't be optional unless all things after it are optional, including `image`.

calebrob6 (approved):

PR #808, opened 2022-10-03
Dependabot: manage GitHub Actions

adamjstewart:
We've been seeing a lot of codecov issues. Apparently the bash uploader is deprecated and there's a new version that works completely differently. Hoping this fixes our issues.

Also updated other actions. We could drop the version requirement, but this helps with stability. If only dependabot could handle these files for us too.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,9 @@
 version: 2
 updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
   - package-ecosystem: "pip"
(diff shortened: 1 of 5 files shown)


adamjstewart:
> If only dependabot could handle these files for us too.

Dependabot can handle these files for us too! See the latest commit.

adamjstewart:
Even this PR is having issues with codecov. I'm going to try to get this to pass, then merge, then if the issue still persists it will be easier to complain to someone.

calebrob6:
Seems good!

calebrob6 (approved):

Issue #253, opened 2021-11-19
Add/fix plot methods to Datasets


calebrob6:
The following benchmark datasets need `plot(...)` methods:

- [x] ADVANCE (#264)
- [x] BeninSmallHolderCashews (#264)
- [x] BigEarthNet (#264)
- [x] COWC (#300)
- [x] CV4AKenyaCropType (#312)
- [x] TropicalCycloneWindEstimation (#298)
- [x] GID15 (#288)
- [x] LEVIRCDPlus (#335)
- [x] VHR10 (#798)
- [x] PatternNet (#314)
- [x] SEN12MS (#320)
- [x] SpaceNet (#252, #311)
- [x] ZueriCrop (#334)

Further, the `plot(...)` methods of RasterDataset and VectorDataset need to be removed, and added individually to:

- [x] CanadianBuildingFootprints (#410)
- [x] CDL (#415)
- [x] NAIP (#407)
- [x] Landsat (#661)
- [x] ChesapeakeCVPR (#820)
- [x] Sentinel (#416)

The `ChesapeakeCVPR` dataset is a bit of an outlier as it is a `GeoDataset`. I need to think more about that.

adamjstewart:
> Further, the plot(...) methods of RasterDataset and VectorDataset need to be reworked or removed.

I would like to move those `plot` methods from `RasterDataset`/`VectorDataset` to the individual dataset classes since there is so much diversity. When that happens, we'll need to change them to accept a sample instead of an image.

ashnair1:
@calebrob6 I've added the plot method to the SpaceNet baseclass so all currently implemented SpaceNet datasets (1,2,4,7) can be plotted.

adamjstewart:
How should we handle plotting for IntersectionDataset/UnionDataset?

calebrob6:
> How should we handle plotting for IntersectionDataset/UnionDataset?

If you aren't doing anything too crazy, you should be able to send the samples that you get from the joined dataset to the individual `dataset.plot(...)` methods. For example, with the `Landsat8 & CDL` dataset, you will be able to call both the Landast8 and CDL plot methods with the resulting samples. You _could_ get crazy and make an `IntersectionDataset.plot(...)` that calls each sub-datasets's plot method and stitches together the resulting plt.Figures into some Frankenstein monster.

adamjstewart:
@calebrob6 `ChesapeakeCVPR` is the only remaining dataset without a `plot` method if you want to take a stab at this.

[...]

PR #814, opened 2022-10-04
Add plot method to all DataModules

adamjstewart:
Now that (almost) all of our Datasets have a plot method, our DataModules should too. This PR adds a plot method to all DataModules except ChesapeakeCVPR, which is still missing a plot method. See #253.

### DIFF
--- a/torchgeo/datamodules/oscd.py
+++ b/torchgeo/datamodules/oscd.py
@@ -196,3 +196,10 @@ class OSCDDataModule(pl.LightningDataModule):
         return DataLoader(
             self.test_dataset, batch_size=1, num_workers=self.num_workers, shuffle=False
         )
+
+    def plot(self, *args: Any, **kwargs: Any) -> plt.Figure:
+        """Run :meth:`torchgeo.datasets.OSCD.plot`.
+
+        .. versionadded:: 0.4
+        """
+        return self.test_dataset.plot(*args, **kwargs)
--- a/torchgeo/datamodules/cyclone.py
+++ b/torchgeo/datamodules/cyclone.py
@@ -148,3 +148,10 @@ class CycloneDataModule(pl.LightningDataModule):
             num_workers=self.num_workers,
             shuffle=False,
         )
+
+    def plot(self, *args: Any, **kwargs: Any) -> plt.Figure:
+        """Run :meth:`torchgeo.datasets.TropicalCycloneWindEstimation.plot`.
+
+        .. versionadded:: 0.4
+        """
+        return self.all_train_dataset.plot(*args, **kwargs)
--- a/torchgeo/datamodules/deepglobelandcover.py
+++ b/torchgeo/datamodules/deepglobelandcover.py
@@ -122,3 +122,10 @@ class DeepGlobeLandCoverDataModule(pl.LightningDataModule):
             num_workers=self.num_workers,
             shuffle=False,
         )
+
+    def plot(self, *args: Any, **kwargs: Any) -> plt.Figure:
+        """Run :meth:`torchgeo.datasets.DeepGlobeLandCover.plot`.
+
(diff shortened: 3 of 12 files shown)


torchgeo/datamodules/oscd.py
```diff
@@ -196,3 +197,10 @@ def test_dataloader(self) -> DataLoader[Any]:
         return DataLoader(
             self.test_dataset, batch_size=1, num_workers=self.num_workers, shuffle=False
         )
+
+    def plot(self, *args: Any, **kwargs: Any) -> plt.Figure:
+        """Run :meth:`torchgeo.datasets.OSCD.plot`.
+
+        .. versionadded:: 0.4
+        """
+        return self.test_dataset.plot(*args, **kwargs)
```
calebrob6, line 206: Heads up that this won't work because the sample that comes in here will be a flattened image

adamjstewart: Yep, I noticed a few failures like this. Not sure what the best approach is other than to try to make things more uniform (dataset.__getitem__ should always return C x H x W, plot method should always convert to H x W x C). Prob going to take me a while to get these working uniformly. If anyone else wants to hack on this PR go ahead.

calebrob6: Well the issue is that Dataset .plot() methods expects a sample as returned by __getitem__, whereas the samples that are used within the train/val methods in our trainers have been modified by the DataModule preprocessing functions

calebrob6: E.g. in OSCD, `sample["image"]` as returned by `__getitem__` will be something like (2 x C x H x W) whereas in the trainer it will be converted to (2C x H x W). If we do channel-wise standardization, this will also be applied during the train/val steps.

adamjstewart: In that case, we should either change the dataset to return 2C x H x W, or create a new ObjectDetectionTask that can handle 2 x C x H x W samples without any changes.

calebrob6: Do you mean `ChangeDetectionTask`?

Generally, what happens if the DataModule subtracts the mean and divides by the stdev?

adamjstewart: Yes.

Not really sure how to handle this. Kornia transforms have a reverse operation, right? Maybe we could apply that if it exists.

isaaccorley: Kornia has a denormalize but unfortunately it doesn't work per channel it only takes in a single value for mean and std.

Here's a torchvision denormalize I've been using lately:

```

class Denormalize(nn.Sequential):
    def __init__(
        self, mean: Sequence[float] = IMAGENET_MEAN, std: Sequence[float] = IMAGENET_STD
    ) -> None:
        self.mean = mean
        self.std = std
        self.inv_mean = torch.tensor([-x for x in mean])
        self.inv_std = torch.tensor([1 / x for x in std])
        super().__init__(
            T.Normalize(mean=[0.0, 0.0, 0.0], std=self.inv_std),
            T.Normalize(mean=self.inv_mean, std=[1.0, 1.0, 1.0]),
        )
```

calebrob6: So I think we should factor out augmentations and preprocessing from the datamodules (where we can easily support denormalizing in a consistent way)

adamjstewart:
Okay, I backed out a few datamodule plot methods that were broken. Those can be added one at a time in future PRs by anyone who cares. I'm still hoping to add plot methods to most datamodules, even if normalization may make visualization a bit broken. I'm not really sure what to do about those other than undo datamodule-specific normalizations inside datamodule plot methods before passing to dataset plot methods.

ashnair1 (approved):

calebrob6 (approved):

Issue #813, opened 2022-10-04
Trainers: predict step


adamjstewart:
### Summary

This issue is to track progress on adding a `predict` step to all Trainers.

- [x] BYOLTask (#819)
- [x] ClassificationTask (#790)
- [x] MultiLabelClassificationTask (#792)
- [x] ObjectDetectionTask (#758)
- [x] RegressionTask (#818)
- [x] SemanticSegmentationTask (#939)

### Rationale

The default `predict` step does not know how to handle our batch dicts.

### Implementation

See implementations that have already been finished.

isaaccorley:
BYOL could just return embeddings for each sample. RegressionTask should also be straightforward similar to ClassificationTask. I'll tackle these today. I think SemanticSegmentation may need some thought since it can be a large amount of data per sample and we don't want users to run OOM by accumulating masks for each sample.

Geethen:
If you have not already considered a similar option for the semantic segmentation regression task: https://rasterio.readthedocs.io/en/latest/topics/concurrency.html

uses rasterio + windows and concurrent to batch write data to a single large image.

adamjstewart:
What's left to do for the SemanticSegmentationTask? Would like to finish this before the next release.

isaaccorley:
Adding a predict step similar to the other tasks will just accumulate the output masks into a list. For large images and datasets this may not be desirable for memory usage reasons. We may want to consider saving the outputs somewhere instead or just leaving it to the user to override how they want the model to perform predictions.

adamjstewart:
Could we go with a simpler implementation in which we return the predictions without saving? Or assume the user will want to predict on a small enough area that it can fit in memory? The current state where it crashes is unideal.

isaaccorley:
Yes I'll make a PR for this. I'll add a note to the docstring to override by making a custom task if a user wants to do something specific with the masks

PR #818, opened 2022-10-04
add predict_step to RegressionTask and tests

isaaccorley:
Related to #813. This PR adds a `predict_step` method to `RegressionTask` so that users can utilize PyTorch Lightning's `trainer.predict()` function which will automatically loop and predict over a PyTorch DataLoader e.g.:

```python


isaaccorley:
I also found the reason why we need the `Tensor` type hint is due to `RegressionTask.forward` returning `Any` and not `Tensor` so mypy assumes that calling `self(x)` returns `Any`.

calebrob6 (approved):

adamjstewart (approved):

PR #819, opened 2022-10-04
add predict_step to BYOLTask

isaaccorley:
Related to #813. This PR adds a `predict_step` method to `BYOLTask` so that users can utilize PyTorch Lightning's `trainer.predict()` function which will automatically loop and predict over a PyTorch DataLoader e.g.:

```python

```

Note the way the EncoderWrapper is setup is that calling `self(x)` on a batch of images will return the projector output embeddings. However, we only need this during training. For linear evaluation or other downstream tasks we want the output of the image encoder (the output of the avg pool layer of the ResNet). To work around this I had to add another hook to store the embeddings on a forward pass in the `self._embeddings` attribute.

### DIFF
--- a/torchgeo/trainers/byol.py
+++ b/torchgeo/trainers/byol.py
@@ -458,10 +458,11 @@ class BYOLTask(pl.LightningModule):
 
         Args:
             batch: the output of your DataLoader
+
         Returns:
             image embeddings
         """
         batch = args[0]
         x = batch["image"]
-        _ = self(x)
+        self(x)
         return self.model.encoder._embedding
--- a/torchgeo/trainers/byol.py
+++ b/torchgeo/trainers/byol.py
@@ -211,6 +211,9 @@ class EncoderWrapper(Module):
         # time this is called
         self._encoded = self.projector(output)
 
+        # Store the image embeddings
+        self._embedding = output
+
     def _register_hook(self) -> None:
         """Register a hook for layer that we will extract features from."""
         layer = list(self.model.children())[self.layer]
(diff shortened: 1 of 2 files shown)


torchgeo/trainers/byol.py
```diff
@@ -449,3 +452,16 @@ def validation_step(self, *args: Any, **kwargs: Any) -> None:
...
+        """Compute and return the output embeddings of the image encoder.
+
+        Args:
+            batch: the output of your DataLoader
+        Returns:
+            image embeddings
+        """
+        batch = args[0]
+        x = batch["image"]
+        _ = self(x)
```
adamjstewart, line 466:
```suggestion
        self(x)
```

torchgeo/trainers/byol.py
```diff
@@ -449,3 +452,16 @@ def validation_step(self, *args: Any, **kwargs: Any) -> None:
 
     def test_step(self, *args: Any, **kwargs: Any) -> Any:
         """No-op, does nothing."""
+
+    def predict_step(self, *args: Any, **kwargs: Any) -> Tensor:
+        """Compute and return the output embeddings of the image encoder.
+
+        Args:
+            batch: the output of your DataLoader
```
adamjstewart, line 460: I would add a newline here

torchgeo/trainers/byol.py
isaaccorley: I believe this would print the outputs which is not what we want.

isaaccorley: I was wrong, it doesn't seem to print the outputs.

adamjstewart: It isn't a Jupyter Notebook...

isaaccorley: Fixed

torchgeo/trainers/byol.py
isaaccorley: Fixed

adamjstewart (approved):

PR #821, opened 2022-10-06
Add Sentinel-1 dataset

adamjstewart:
I've never worked with Sentinel-1 or SAR data before, so I have a ton of questions. Maybe @isaaccorley can help?

1. The files have no CRS/transform, what's up with that?
2. How do I plot SAR data?


isaaccorley:
1. That doesn't seem right. Where are you getting the data from? PC?

2. Normally the bands are plotted as a false color image. So you would make some combination of 3 bands like (VV, VH, ratio of VV/VH) and then plot it as if it's RGB. Of course this will likely be noisy since the data is typically heavily preprocessed.

adamjstewart:
1. Data is from Copernicus, haven't looked at PC data yet

adamjstewart:
2. There are 3 VV and 3 VH. What's the difference? Should I plot all 3 combos?

adamjstewart:
I'm not sure if I'm the right person for this PR, I don't even know the difference between:

* S1A/S1B
* SLC/GRD/OCN/RAW
* HH/VV/VH/HV/HH+HV/VV+VH
* SM/IW/EW/WV

adamjstewart:
According to https://docs.terradue.com/geohazards-tep/tutorials/s1-grd-rgb-composite.html, we want GRD, not SLC, and we want VV (red), HH (green), and VV / VH (blue)

adamjstewart:
According to https://sentinels.copernicus.eu/web/sentinel/user-guides/sentinel-1-sar/product-overview/polarimetry, we want VV (red), VH (green), and |VV| / |VH| (blue)

RitwikGupta:
I'd recommend making two different classes: one for Sentinel-1 SLCs, and one for Sentinel-1 GRDs. Punt on the first one for now since that involves dealing with non-georeferenced (kind of), complex-valued data. GRD should be easy to integrate. Plot the VV/VH bands separately as 2x1 subplot.

RitwikGupta:
> S1A/S1B: I guess there's just 2 satellites
Yes. S1B is broken and will not be fixed. You'll only see S1A after Dec 2021.

> [SLC/GRD](https://sentinels.copernicus.eu/web/sentinel/user-guides/sentinel-1-sar/product-types-processing-levels/level-1)/[OCN](https://sentinels.copernicus.eu/web/sentinel/user-guides/sentinel-1-sar/product-types-processing-levels/level-2)/[RAW](https://sentinels.copernicus.eu/web/sentinel/user-guides/sentinel-1-sar/product-types-processing-levels/level-0)
SLC =  single look, complex
GRD = ground range detected
OCN = ocean
RAW = literally raw radar data

We only care about GRD and SLC, in that order.

> HH/VV/VH/HV/HH+HV/VV+VH
Just like how EO can be captured at different wavelengths, SAR is captured at various polarizations. The first letter tells us which polarization was transmitted, the second one tells us which polarization was received. HH = H transmit, H receive. HH+HV is just a literal sum.

> [SM/IW/EW/WV](https://sentinels.copernicus.eu/web/sentinel/user-guides/sentinel-1-sar/acquisition-modes)
These are different acquisition modes for the Sentinel-1 satellites. They trade-off capture area and resolution based on different desired characteristics. Regardless of mode, you'll still have SLC/GRD products.

adamjstewart:
I downloaded `S1A_IW_GRDH_1SDV_20221204T210847_20221204T210912_046187_0587AD_F19F.SAFE` and `S1A_IW_SLC__1SDV_20221006T133300_20221006T133327_045322_056B2B_84A7.SAFE` but neither have a `crs` value when opened with rasterio. Is there something I'm missing? I'm just downloading random images from Copernicus in hopes of finding one that I can visualize.

adamjstewart:
This seems relevant: https://gis.stackexchange.com/questions/233375/sentinel-1-data-opened-with-rasterio-has-no-affine-transform-crs

adamjstewart:
Yep, these files use ground control points, not CRS. This is a fun hiccup.

adamjstewart:
If anyone wants to follow along, the exact files I'm trying to use are https://scihub.copernicus.eu/dhus/odata/v1/Products('d3b5c561-ac0a-4617-a040-75708c0d28b8')/$value

Not sure if that URL will work for others though.

adamjstewart:
Alright, found some decent image parameters! These are scenes over the big island of Hawai'i:

HH_HV
VV_VH

Happy to tweak these parameters if anyone has better suggestions.

torchgeo/datasets/sentinel.py
```diff
@@ -25,6 +25,185 @@ class Sentinel(RasterDataset):
...
+    * `Interferometric Wide (IW) swath
+      <https://sentinels.copernicus.eu/web/sentinel/user-guides/sentinel-1-sar/acquisition-modes/interferometric-wide-swath>`_
+    * `Extra Wide (EW) swatch
+      <https://sentinels.copernicus.eu/web/sentinel/user-guides/sentinel-1-sar/acquisition-modes/extra-wide-swath>`_
+    * `Wave (WV)
+      <https://sentinels.copernicus.eu/web/sentinel/user-guides/sentinel-1-sar/acquisition-modes/wave>`_
+
+    At the moment, this dataset only supports the GRD product type. Data must be
+    radiometrically terrain corrected (RTC). This can be done manually using a DEM, or
+    you can download an On Demand RTC product from ASF DAAC.
```
adamjstewart, line 81: If we do add support for SLC/OCN/RAW, there's a good chance we'll need separate classes to support these. I'll leave this for future work and stick with `Sentinel1` for now.

torchgeo/datasets/sentinel.py
```diff
@@ -25,6 +25,185 @@ class Sentinel(RasterDataset):
...
+                and returns a transformed version
+            cache: if True, cache file handle to speed up repeated sampling
+
+        Raises:
+            AssertionError: if ``bands`` is invalid
+            FileNotFoundError: if no files are found in ``root``
+        """
+        # Sentinel-1 uses dual polarization, it can only transmit either
+        # horizontal or vertical at a single time
+        assert list(bands) == ["HH", "HV"] or list(bands) == ["VV", "VH"]
```
adamjstewart, line 159: This is overly restrictive, but it's done for two reasons:

1. `RasterDataset` requires `bands` to be present for all scenes, you can't have some scenes with HH/HV and other scenes with VV/VH. The easiest way to support that would be:
    ```python
    Sentinel1(root, bands=["HH", "HV"]) | Sentinel1(root, bands=["VV", "VH"])
    ```
2. It greatly simplifies plotting if we assume we always have both bands and they are in a specific order

Will anyone be upset if they can't use only a single band or can't specify them in reverse order?

torchgeo/datasets/sentinel.py
```diff
@@ -25,6 +25,185 @@ class Sentinel(RasterDataset):
...
+            a matplotlib Figure with the rendered sample
+        """
+        # Either HH/HV or VV/VH, doesn't really matter which
+        hh = sample["image"][0]
+        hv = sample["image"][1]
+
+        # https://gis.stackexchange.com/a/400780/123758
+        hh = torch.clamp(hh / 0.3, min=0, max=1)
+        hv = torch.clamp(hv / 0.05, min=0, max=1)
+        hh_hv = torch.clamp(hh / hv / 25, min=0, max=1)
```
adamjstewart, line 188: No idea if I'm doing this right, but it looks much better than all the random hacks I tried.

adamjstewart:
@RitwikGupta it would be good to get your review of this since you're the SAR expert (at least among us). We can always expand this in future PRs, especially if you're focusing on better ML for SAR!

torchgeo/datasets/sentinel.py
```diff
@@ -25,6 +25,189 @@ class Sentinel(RasterDataset):
...
+    ) -> None:
+        """Initialize a new Dataset instance.
+
+        Args:
+            root: root directory where dataset can be found
+            crs: :term:`coordinate reference system (CRS)` to warp to
+                (defaults to the CRS of the first file found)
+            res: resolution of the dataset in units of CRS
+                (defaults to the resolution of the first file found)
+            bands: bands to return (defaults to all bands)
```
calebrob6, line 151: Says it defeaults to all bands but it seems like it defaults to VV and VH

torchgeo/datasets/sentinel.py
```diff
@@ -25,6 +25,189 @@ class Sentinel(RasterDataset):
...
+    * `Interferometric Wide (IW) swath
+      <https://sentinels.copernicus.eu/web/sentinel/user-guides/sentinel-1-sar/acquisition-modes/interferometric-wide-swath>`_
+    * `Extra Wide (EW) swatch
+      <https://sentinels.copernicus.eu/web/sentinel/user-guides/sentinel-1-sar/acquisition-modes/extra-wide-swath>`_
+    * `Wave (WV)
+      <https://sentinels.copernicus.eu/web/sentinel/user-guides/sentinel-1-sar/acquisition-modes/wave>`_
+
+    .. note::
+       At the moment, this dataset only supports the GRD product type. Data must be
+       radiometrically terrain corrected (RTC). This can be done manually using a DEM,
```
calebrob6, line 81: Have you tested this with GRD and RTC data downloaded from the Planetary Computer? https://planetarycomputer.microsoft.com/dataset/group/sentinel-1

torchgeo/datasets/sentinel.py
calebrob6: Bands shouldn't have to be in a specific order as you are given the order in `self.bands`

torchgeo/datasets/sentinel.py
```diff
@@ -25,6 +25,189 @@ class Sentinel(RasterDataset):
...
+        Returns:
+            a matplotlib Figure with the rendered sample
+        """
+        # Either HH/HV or VV/VH, doesn't really matter which
+        hh = sample["image"][0]
+        hv = sample["image"][1]
+        hh_hv = hh / hv
+
+        # https://gis.stackexchange.com/a/400780/123758
+        hh = torch.clamp(hh / 0.3, min=0, max=1)
```
calebrob6, line 190: these variable names are confusing as "hh" could mean the "HH" band or the "VH" band -- maybe using variables like `horizontal_receive` and `vertical_receive` would be better?

torchgeo/datasets/sentinel.py
adamjstewart: Nope, please do! I can never remember the syntax to download files from PC. If the filename structure or file format is different, let me know and I can add more tests.

torchgeo/datasets/sentinel.py
adamjstewart: hh means HH or VV, not HH or VH.

* hh: the band where transmit == receive
* hv: the band where transmite != receive

Wasn't sure what a better name would be. It doesn't matter too much since it's a local variable that users can't access. But I would at least like to make this clear. Will clarify this comment.

torchgeo/datasets/sentinel.py
adamjstewart: True, but it adds a bit of extra code:
```python
if self.bands[0] == "HH" or self.bands[0] == "VV":
    hh = self.bands[0]
    hv = self.bands[1]
else:
    hh = self.bands[1]
    hv = self.bands[0]
```
and it isn't clear to me if it's worth it.

torchgeo/datasets/sentinel.py
calebrob6: You just go to the explorer, select either S1 dataset, click a scene, and click the code snippet:

image

```
import pystac
import planetary_computer
import rioxarray

item_url = "https://planetarycomputer.microsoft.com/api/stac/v1/collections/sentinel-1-grd/items/S1A_IW_GRDH_1SDV_20221218T210639_20221218T210704_046391_058EB1"

# Load the individual item metadata and sign the assets
item = pystac.Item.from_file(item_url)
signed_item = planetary_computer.sign(item)

# Open one of the data assets (other asset keys to use: 'vv')
asset_href = signed_item.assets["vh"].href
```

From -- https://planetarycomputer.microsoft.com/explore?c=124.0274%2C-16.4940&z=9.00&v=2&d=sentinel-1-grd&m=Most+recent+-+VV%2C+VH&r=VV%2C+VH+False-color+composite&s=false%3A%3A100%3A%3Atrue&sr=desc&ae=0

torchgeo/datasets/sentinel.py
calebrob6: These 6 lines aren't some complicated hack / keeping them out isn't a great simplification. I would keep plotting working regardless of band order or make some enum that specifies which subset of options we accept for bands

adamjstewart: Okay, check the latest commit. I relaxed pretty much all constraints so that the following are supported:

* single band: [HH], [HV], [VV], [VH]
* dual band: [HH, HV], [VV, VH]
* dual band (reverse order): [HV, HH], [VH, VV]

but the following are still invalid:

* no bands: [] - doesn't make sense, impossible to support
* duplicate bands: [HH, HH] - doesn't make sense, but possible to support
* quad bands: [HH, HV, VV, VH] - not possible to support with current RasterDataset implementation

These are supported both for data loading and plotting. I also added a ton of new tests to check for these.

Only remaining thing we could potentially add support for is a plotting option to choose between RGB or dual grayscale for dual band plotting. But I didn't want to make things any more complicated than they are, and we don't support this for any other dataset at the moment.

torchgeo/datasets/sentinel.py
adamjstewart: I might add this to our Wiki so I don't keep forgetting it...

calebrob6 (dismissed):

torchgeo/datasets/sentinel.py
calebrob6: THis seems completely reasonable!

torchgeo/datasets/sentinel.py
adamjstewart: https://github.com/microsoft/torchgeo/wiki/Downloading-Data

Will try to update that with other sites I've used over the years

adamjstewart: Files on the Planetary Computer have the following URL:

> https://sentinel1euwestrtc.blob.core.windows.net/sentinel1-grd-rtc/GRD/2016/11/9/IW/DH/S1B_IW_GRDH_1SDH_20161109T042206_20161109T042231_002877_004E0A_7013/measurement/iw-hh.rtc.tiff

When downloaded, the file is named `iw-hh.rtc.tiff`. Our current RasterDataset implementation only supports parsing the filename, not the directory. And if you download the data locally, you don't have access to the rest of the URL anyway. So this data loader likely works with Planetary Computer data, but requires you to manually rename all files to match the format used by ASF DAAC.

adamjstewart: Also, this dataset likely doesn't work for the GRD files, they have to first be RTC'ed to convert the ground control points to a CRS.

adamjstewart: Okay, I confirmed that it works with Planetary Computer RTC data as long as you rename the files. Also confirmed that both the HH/HV and HH plots work correctly.

tests/datasets/test_sentinel.py
```diff
@@ -3,14 +3,98 @@
...
+            # Only horizontal or vertical receive
+            ["HH"],
+            ["HV"],
+            ["VV"],
+            ["VH"],
+            # Both horizontal and vertical receive
+            ["HH", "HV"],
+            ["HV", "HH"],
+            ["VV", "VH"],
+            ["VH", "VV"],
```
RitwikGupta, line 36: Why are these not just `set(["HH", "HV"])` and `set(["VV", "VH"])`?

torchgeo/datasets/sentinel.py
```diff
@@ -25,6 +25,221 @@ class Sentinel(RasterDataset):
...
+        assert len(bands) > 0, "'bands' cannot be an empty list"
+        assert len(bands) == len(set(bands)), "'bands' contains duplicate bands"
+
+        for band in bands:
+            assert band in self.all_bands, f"invalid band '{band}'"
+
+        # Sentinel-1 uses dual polarization, it can only transmit either
+        # horizontal or vertical at a single time
+        msg = """
+'bands' cannot contain both horizontal and vertical transmit at the same time.
```
RitwikGupta, line 169: Why not? Is this limitation solely for plotting purposes? It is totally valid for multiple passes over the same area to be of different polarizations. It's also valid for consecutive captures from the satellite to be of different transmit/receive pairings. Additionally, the intensities for VH and HV are equal due to Helmholtz reciprocity!

torchgeo/datasets/sentinel.py
```diff
@@ -25,6 +25,221 @@ class Sentinel(RasterDataset):
...
+
+        if len(bands) == 1:
+            # Only horizontal or vertical receive, plot as grayscale
+
+            image = sample["image"][0]
+            image = torch.clamp(image, min=0, max=1)
+
+            title = f"({bands[0]})"
+        else:
+            # Both horizontal and vertical receive, plot as RGB
```
RitwikGupta, line 208: Is there any reason to plot these images as RGB? The case to plot RGB images is rare as the third band ratio choice is extremely dependent on application. Why not just present an Nx1 subplot with all bands plotted in grayscale?

torchgeo/datasets/sentinel.py
```diff
@@ -25,6 +25,221 @@ class Sentinel(RasterDataset):
...
+    # x:          Sentinel-1 Mission (A or B)
+    # yy:         Beam Mode
+    # aaaaaaaa:   Start Date of Acquisition (YYYYMMDD)
+    # bbbbbb:     Start Time of Acquisition (HHMMSS)
+    # pp:         Polarization: Dual-pol (D) vs. Single-pol (S),
+    #                 primary polarization (H vs. V)
+    # o:          Orbit Type: Precise (P), Restituted (R), or Original Predicted (O)
+    # zz:         Terrain Correction Pixel Spacing
+    # u:          Software Package Used: GAMMA (G)
+    # d:          Gamma-0 (g) or Sigma-0 (s) Output
```
RitwikGupta, line 104: If you must put conditions on what values one dataset can have, then ensure that one instantiation can only have either $\gamma_0$ or $\sigma_0$. Otherwise differing images will mean vastly different things!

torchgeo/datasets/sentinel.py
```diff
@@ -25,6 +25,221 @@ class Sentinel(RasterDataset):
...
+    # e:          Power (p) or Decibel (d) or Amplitude (a) Output
```
RitwikGupta, line 105: The user should specify a scale they want their dataset to be in, and the dataset should normalize all files to that scale. Otherwise, differing images will mean vastly different things.
For reference:
```
amplitude = sqrt(power)
decibel = 10 * log10(power)
decibel = 10 * log10(amplitude ** 2)
```

RitwikGupta (changes requested):
The limitations on what kinds of files can be in one Dataset instance are currently wrong.

I don't think there's any reason why we need to limit the kinds of bands available within the dataset at a time. Also, we need to be mindful that cross polarization is reciprocal.

We need to limit that a dataset can either be $\gamma_0$ or $\beta_0$, and image scales need to be normalized to one choice, power by default.

torchgeo/datasets/sentinel.py
RitwikGupta: Amplitude data is better for viz, so when plotting you can always plot the sqrt of the power data. I'd recommend normalizing to power by default unless the user wants otherwise.

tests/data/sentinel1/data.py
```diff
@@ -0,0 +1,65 @@
...
+
+SIZE = 36
+
+np.random.seed(0)
+
+FILENAME_HIERARCHY = Union[Dict[str, "FILENAME_HIERARCHY"], List[str]]
+
+filenames: FILENAME_HIERARCHY = {
+    # Copernicus Open Access Hub
+    "S1A_IW_SLC__1SDV_20221006T133300_20221006T133327_045322_056B2B_84A7.SAFE": {
```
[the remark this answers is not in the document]
adamjstewart: This commit has not existed for a very long time. I would suggest reviewing the latest version only.

tests/datasets/test_sentinel.py
adamjstewart: For our MSI datasets, we want to be able to optionally select a subset of bands (RGB-only, RGB+NIR, etc.). I was following the same approach here, but I'm not actually sure if that makes sense for SAR. Is there any reason why someone might want to only use a single band and not both? If not, I can simplify the code quite a bit.

torchgeo/datasets/sentinel.py
adamjstewart: This limitation is inherent in our current `RasterDataset` implementation. Basically, every `GeoDataset` has an R-tree index containing a **single** filename for each spatiotemporal location. We populate this R-tree during `__init__` using a filename glob, which has to contain a **single** band name (can't be both HH and VV). Then, during `__getitem__`, we loop over **all** bands and swap the band in the filename to load the data. If all bands aren't present for every spatiotemporal location, then loading will fail. This was designed for MSI and HSI imagery without any thought towards (or knowledge of) SAR data.

There are a few options to deal with this:

### 1. Leave it up to the user

This is our current approach. The user can create two datasets and compute the union themselves.

Pro: simple
Con: users may have to do extra work

### 2. Compute the union for them

We could subclass `UnionDataset` instead of `RasterDataset` and create two `RasterDatasets` for the user. This is actually a very elegant solution and shouldn't be that difficult.

Pro: users don't have to do any extra work
Con: it would not be possible to change bands. You could create a dataset containing all 4 bands, but it would not be possible to create a dataset with only 2 bands

I assume horizontal transmit and vertical transmit are fundamentally different and some users may want to work with only one or the other but not both? This would not be possible with this idea. A workaround to support this would be to create 3 datasets:

* Sentinel1Horizontal: [HH, HV]: RasterDataset
* Sentinel1Vertical: [VV, VH]: RasterData
* Sentinel1: [HH, HV, VV, VH]: UnionDataset

### 3. Override `__init__` and `__getitem__` in Sentinel1

We could replace the default logic for populating the R-tree and loading the data to search for both HH and VV and to load [HH, HV] and [VV, VH].

Pro: would support 4 bands or 2 bands
Con: mostly duplicate code, more difficult to maintain, need to do this for every bi-pole SAR instrument

### 4. Modify RasterDataset to support optional bands

Instead of populating the R-tree with a single band for each spatiotemporal location, we could load all bands we can find. Then, at loading time, we only load the bands that were found.

Pro: supports both SAR as well as MSI/HSI where the user might not have downloaded all bands but forgot to change `bands`
Con: significantly slower since R-tree lookup will go from $\mathcal{O}(\log N)$ to $\mathcal{O}(\log CN)$ and memory will go from $\mathcal{O}(N)$ to $\mathcal{O}(CN)$ for all datasets. Even slower would be loading and grabbing the CRS and bounding box of every file from disk. Also, if some downloads have different numbers of bands, this will break when loading data. Easy fix for this would be to specify a subset of bands when instantiating the dataset, but the error message would not be clear and this may be difficult to figure out for an inexperienced user

Let me know how you feel about 1–4. I'm leaning towards 1 or 2. I'm hesitant to go with 3 because duplicate code == evil, and 4 would require a separate PR and longer discussion.

torchgeo/datasets/sentinel.py
adamjstewart: Because RGB looks cooler 😄 

Seriously though, I only went with RGB cuz that's how we plot MSI datasets. When I view thumbnails of these images on Copernicus, they also choose to display an RGB image. What other third band ratio choices are common? We could add an option to select one or the other when plotting. Of course, users can always implement their own plotting code, this is just supposed to be a reasonable default.

torchgeo/datasets/sentinel.py
adamjstewart: We currently don't enforce this, but we could. But that would mean creating separate `Sentinel1Gamma` and `Sentinel1Sigma` classes. How common is it for a user to have a directory containing a mix of both gamma and sigma files?

adamjstewart: Or it could be a parameter of `__init__` and we would only need a single `Sentinel1` class. Would users be confused if they don't know what this thing means and TorchGeo can't find any data for them because they didn't specify the right option?

torchgeo/datasets/sentinel.py
adamjstewart: Again, how common is it for a user to have multiple of these formats mixed in a single directory? Right now we just load the data as it's stored in the file. We could add a parameter to `__init__` so that only one of these can be loaded at a single time, or we could modify the scale during loading, but that would require overriding `__getitem__`.

torchgeo/datasets/sentinel.py
RitwikGupta: I'm not tied to displaying an image one way or another, but if it makes the code easier, then displaying N grayscale bands would be infinitely easier than dealing with all these combinations/ratios for RGB imagery.

torchgeo/datasets/sentinel.py
RitwikGupta: It's not common to mix gamma and sigma. Let's resolve this comment as we can leave this up to the user. Can we add a comment noting that, though?

torchgeo/datasets/sentinel.py
calebrob6: I prefer 1 -- until we have someone who is actively using SAR imagery with TorchGeo let's not spend too much time making this class perfect

calebrob6 (approved):

PR #822, opened 2022-10-06
Labeler: use comments to avoid aliases

adamjstewart:
Apparently things starting with an asterisk are aliases in YAML. Explicitly encode as strings to avoid this.


adamjstewart:
Bleh, GitHub Actions will only use the version in main for security reasons, so we'll have to force merge this. That also explains why we didn't notice in the PR.

calebrob6:
nbd

calebrob6 (approved):

PR #832, opened 2022-10-11
Update torchmetrics dep in other places

adamjstewart:
Follow-up to #816 

I would love to delete these files someday so we don't have to track our deps in so many places...

@remtav

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,5 +34,5 @@ repos:
         hooks:
         -   id: mypy
             args: [--strict, --ignore-missing-imports, --show-error-codes]
-            additional_dependencies: [torch>=1.11, torchmetrics>=0.7, pytorch-lightning>=1.7, pytest>=6, omegaconf>=2.1, kornia>=0.6, numpy>=1.22.0]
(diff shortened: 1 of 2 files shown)


remtav:
> Follow-up to #816
> 
> I would love to delete these files someday so we don't have to track our deps in so many places...
> 
> @remtav

Yeah I understand it's a bit cumbersome on maintenance to support conda on top of the rest. For our group's (Natural Resources Canada) production's need, conda has proven a functional and satisfying solution for dealing with multiple python environment's on our HPC infrastructure. Using [mamba](https://mamba.readthedocs.io/en/latest/installation.html) instead of conda has made things even easier and much faster. There may be a question of comfort here, virtualenv hasn't been tested really and other solutions could be suitable. It's simpler to stay with conda for now so we're glad torchgeo supports it :)

adamjstewart:
Completely understand, conda is super popular. I use Spack for everything which is yet another thing I have to support. But I ended up removing the `spack.yaml` from this repo and only maintaining a recipe in Spack itself because I'm the only one who uses Spack. We'll likely continue to support conda for the foreseeable future, even if it isn't my personal choice of PM.

calebrob6 (approved):

PR #846, opened 2022-10-14
Datamodule naming TropicalCyclone

### LINKED ISSUE #815: Dataset/DataModule consistency
### Summary

We should ensure that all Dataset and DataModule file and class names match.

### Rationale

The user should not have to guess what the DataModule name is, it should be easy to determine based on the Dataset name. Consistency is important.

### Implementation

At the moment, we have datasets like `TropicalCycloneWindEstimation` and data modules like `CycloneDataModule`. We should pick one name and use it consistently. Not sure if there are other instances of this.

### Alternatives

_No response_

### Additional information

_No response_

nilsleh:
This PR closes #815 by renaming the `TropicalCycloneWindEstimation` to `TropicalCyclone` and the `CycloneDataModule` to `TropicalCycloneDataModule`. The naming scheme of data modules is now consistent again where the name of the data module is the dataset name concatenated with `DataModule`.


torchgeo/datamodules/cyclone.py
```diff
@@ -10,18 +10,22 @@
...
-class CycloneDataModule(pl.LightningDataModule):
+class TropicalCycloneDataModule(pl.LightningDataModule):
     """LightningDataModule implementation for the NASA Cyclone dataset.
 
     Implements 80/20 train/val splits based on hurricane storm ids.
     See :func:`setup` for more details.
+
+    .. versionchanged:: 0.4.0
+            Class name changed from CycloneDataModule to TropicalCycloneDataModule to be
+            consistent with TropicalCyclone dataclass.
```
adamjstewart, line 28: Indented too far, dataclass -> dataset

torchgeo/datasets/cyclone.py
```diff
@@ -38,6 +38,10 @@ class TropicalCycloneWindEstimation(NonGeoDataset):
 
        * `radiant-mlhub <https://pypi.org/project/radiant-mlhub/>`_ to download the
          imagery and labels from the Radiant Earth MLHub
+
+    ..versionchanged:: 0.4.0
```
adamjstewart, line 42:
```suggestion
    .. versionchanged:: 0.4.0
```

adamjstewart (changes requested):
New name looks good, I confirmed that there are no other datasets/datamodules with this kind of inconsistency.

adamjstewart (approved):

PR #847, opened 2022-10-14
Add plot method for VHR10 dataset

### LINKED ISSUE #894: Using VHR10 dataset with AugmentationSequential
https://github.com/microsoft/torchgeo/blob/0c7d46be10c47fdb370364fddb7328d1b0661ac2/torchgeo/datasets/nwpu.py#L143

sample (dict) has the key: `'label'`, but should have the key `'mask'`

The reason:

If using torchgeo augmentations
https://github.com/microsoft/torchgeo/blob/main/torchgeo/transforms/transforms.py#L40

The parameter `data_keys` in `kornia.augmentation.AugmentationSequential` accepts only `[“input”, “mask”, “bbox”, “bbox_xyxy”, “bbox_xywh”, “keypoints”]` as stated in [kornia doc](https://kornia.readthedocs.io/en/latest/augmentation.container.html#augmentation-sequential)

ashnair1:
Part 1 of splitting #798 into separate PRs.

Summary of changes:
- Added tests
- Added data.py for VHR10
- Renamed scripts from `nwpu.py` to `vhr10.py`. 
- Added `scikit-image` dependency
- Updated minimum supported version of `scipy`

Example plots:

Ground truth and predicition (object detection training)
planes

Controlling features to be shown and their respective opacity
final_img

Closes #894


setup.cfg
```diff
@@ -49,6 +49,8 @@ install_requires =
     rasterio>=1.0.20,<2
     # rtree 1+ required for len(index), index & index, index | index
     rtree>=1,<2
+    # scikit-image required for find_contours
+    scikit-image>=0.15.0,<0.20
```
adamjstewart, line 53: This definitely shouldn't be required if it's optionally needed by a single dataset. Let's move it to the datasets extra, just like pycocotools and rarfile.

torchgeo/datasets/vhr10.py
```diff
@@ -0,0 +1,537 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
adamjstewart, line 1: What's the motivation to rename from nwpu to vhr10?

adamjstewart (commented):
Can you add example plots?

torchgeo/datasets/vhr10.py
ashnair1: NWPU is the name of the institute that created the dataset - Northwestern Polytechnical University. VHR10 is the name of the dataset.

They're also the one that created RESISC45 dataset

setup.cfg
```diff
@@ -94,6 +94,8 @@ datasets =
     radiant-mlhub>=0.2.1,<0.5
     # rarfile 3+ required for correct Rar file detection
     rarfile>=3,<5
+    # scikit-image required for find_contours
+    scikit-image>=0.19.0,<1
```
adamjstewart, line 98: `find_contours` existed long before scikit-image 0.19, what's the minimum version that works? I try not to require newer versions unless absolutely necessary. The easiest way to test this would be to pip install successively older versions of scikit-image and see what the oldest working version is.

ashnair1: `find_contours` was added in `v0.4`. 

I tried installing previous versions of `scikit-image` (while keeping the other dependencies constant) and they failed for one reason or the other. Downgrading numpy and scipy might allow those versions to work but I don't think that's the way to go.

For reference, I tested the versions via the following:
```
# Create test environment and install minimal dependencies
conda create -n min-test python=3.7
pip install -r requirements/min.old

# Install version of scikit-image 
pip install scikit-image=A.B.C

# Run test
pytest tests/datasets/test_vhr10.py

```

adamjstewart: I'll try to play around with this when I get a chance and see if I can get older versions working without any code changes.

tests/data/vhr10/data.py
```diff
@@ -0,0 +1,99 @@
...
+            path,
+            "w",
+            driver="JP2OpenJPEG",
+            height=img.shape[0],
+            width=img.shape[1],
+            count=3,
+            dtype=img.dtype,
+        ) as dst:
+            for i in range(1, dst.count + 1):
+                dst.write(img, i)
```
adamjstewart, line 29: If the data files aren't georeferenced, there's no reason to use rasterio. Let's just use pillow. See https://github.com/microsoft/torchgeo/tree/main/tests/data#rgb-images.

tests/datasets/test_vhr10.py
```diff
@@ -88,11 +94,36 @@ def test_not_downloaded(self, tmp_path: Path) -> None:
...
                 match="pycocotools is not installed and is required to use this datase",
             ):
                 VHR10(dataset.root, dataset.split)
+
+            with pytest.raises(
+                ImportError,
+                match="scikit-image is not installed and is required to plot masks",
+            ):
+                x = dataset[0]
+                _ = dataset.plot(x)
```
adamjstewart, line 111:
```suggestion
                dataset.plot(x)
```

tests/datasets/test_vhr10.py
```diff
@@ -88,11 +94,36 @@ def test_not_downloaded(self, tmp_path: Path) -> None:
...
+            scores = [0.7, 0.3, 0.7]
+            for i in range(3):
+                x = dataset[i]
+                x["prediction_labels"] = x["labels"]
+                x["prediction_boxes"] = x["boxes"]
+                x["prediction_scores"] = torch.Tensor([scores[i]])
+                if "masks" in x:
+                    x["prediction_masks"] = x["masks"]
+                    dataset.plot(x, show_feats="masks")
+                plt.close()
```
adamjstewart, line 129: Should this be indented once more?

torchgeo/datasets/vhr10.py
```diff
@@ -0,0 +1,544 @@
...
+        Tensor: Mask tensor
+    """
+    from pycocotools import mask as coco_mask  # noqa: F401
+
+    masks = []
+    for polygons in segmentations:
+        rles = coco_mask.frPyObjects(polygons, height, width)
+        mask = coco_mask.decode(rles)
+        # if len(mask.shape) < 3:
+        #     mask = mask[..., None]
```
adamjstewart, line 40: Can remove this if it isn't needed

torchgeo/datasets/vhr10.py
```diff
@@ -0,0 +1,544 @@
...
+    else:
+        masks_tensor = torch.zeros((0, height, width), dtype=torch.uint8)
+    return masks_tensor
+
+
+class ConvertCocoAnnotations:
+    """Callable for converting the boxes, masks and labels into tensors.
+
+    This is a modified version of ConvertCocoPolysToMask() from torchvision found in
+    https://github.com/pytorch/vision/blob/main/references/detection/coco_utils.py
```
adamjstewart, line 55:
```suggestion
    https://github.com/pytorch/vision/blob/v0.14.0/references/detection/coco_utils.py
```
This URL will be more stable in case the feature is ever removed

torchgeo/datasets/vhr10.py
```diff
@@ -0,0 +1,544 @@
...
+    """
+
+    def __call__(self, sample: Dict[str, Any]) -> Dict[str, Any]:
+        """Converts MS COCO fields (boxes, masks & labels) from list of ints to tensors.
+
+        Args:
+            sample (Dict[str, Any]): Sample
+
+        Returns:
+            Dict[str, Any]: Processed sample
```
adamjstewart, line 65:
```suggestion
            sample: Sample

        Returns:
            Processed sample
```
Types are already taken from function signature, no reason to duplicate

torchgeo/datasets/vhr10.py
```diff
@@ -0,0 +1,544 @@
...
+        mask = mask.any(dim=2)
+        masks.append(mask)
+    if masks:
+        masks_tensor = torch.stack(masks, dim=0)
+    else:
+        masks_tensor = torch.zeros((0, height, width), dtype=torch.uint8)
+    return masks_tensor
+
+
+class ConvertCocoAnnotations:
```
adamjstewart, line 51: This is all quite a bit more complicated than it used to be. What was the reason why this was needed?

torchgeo/datasets/vhr10.py
```diff
@@ -0,0 +1,544 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
adamjstewart, line 1: Can we temporarily undo the filename change and move it to a separate PR (either before or after this PR is merged)? That way it will be easier to see in this PR which things actually changed and which didn't. Or did the majority of things change so the diff isn't actually useful?

torchgeo/datasets/vhr10.py
```diff
@@ -0,0 +1,544 @@
...
+            prediction_labels = sample["prediction_labels"].numpy()
+            prediction_scores = sample["prediction_scores"].numpy()
+            if "prediction_boxes" in sample:
+                prediction_boxes = sample["prediction_boxes"].numpy()
+                show_pred_boxes = True
+            if "prediction_masks" in sample:
+                prediction_masks = sample["prediction_masks"].numpy()
+                show_pred_masks = True
+
+            N_PRED = len(prediction_labels)
```
adamjstewart, line 436: Let's use lowercase for local variables

adamjstewart (commented):
More minor comments. My main concerns so far are:

1. Handling of COCO polygons seems to be much more complex than it used to be
2. Lots of very recent dependency versions

It would be good to have a justification for these things. Complexity is fine if it's necessary but I always try to avoid it if it isn't. I haven't gotten a chance to play around with finding the minimum versions, but you can always drop the version repeatedly in the file until CI starts to fail to find a lower bound.

setup.cfg
ashnair1: Tried this again. `0.19.0` is the oldest version that works consistently. `0.18.0` works on Python 3.7 but fails on others. Had to update scipy as well.

torchgeo/datasets/vhr10.py
ashnair1: This is mainly following torchvision's implementation. I assume this is more complex because they wanted to handle some edge cases.

adamjstewart: If those edge cases don't exist in VHR-10 and we don't have any plans of using this elsewhere I would skip the complexity until it's actually needed

ashnair1: So I went over the implementation and while it is verbose I don't think there is any unnecessary complexity. Here we're mainly parsing the COCO dataset and returning the labels as a dict of tensors as we do for all other datasets. 

There are roughly 10 extra lines for handling cases where segmentation masks are not provided, which was a personal addition. Since MS COCO is a common spec for handling both object detection and instance segmentation annotations; I added this so that we would be covered if we ever encountered a MS COCO style remote sensing object detection dataset (only boxes). If required, I can remove these 10 lines.

torchgeo/datasets/vhr10.py
ashnair1: Rename reverted

tests/datasets/test_vhr10.py
```diff
@@ -7,6 +7,7 @@
 from pathlib import Path
```
adamjstewart, line 7: Can you undo this rename too? We can do all renames in unison later.

requirements/min.old
```diff
@@ -31,7 +31,8 @@ pandas==0.23.2
 pycocotools==2.0.0
 radiant-mlhub==0.2.1
 rarfile==3.0
-scipy==1.2.0
+scikit-image==0.18.0
+scipy==1.7.0
```
adamjstewart, line 35: Why the scipy version bump? You'll also need to bump this lower bound in other places

tests/data/vhr10/data.py
```diff
@@ -0,0 +1,87 @@
+import json
```
adamjstewart, line 1: Let's put everything in a nwpu directory for now and rename later

torchgeo/datasets/nwpu.py
```diff
@@ -192,7 +312,7 @@ def _load_target(self, id_: int) -> Dict[str, Any]:
         # Images in the "negative" image set have no annotations
         annot = []
         if self.split == "positive":
-            annot = self.coco.loadAnns(self.coco.getAnnIds(id_))
+            annot = self.coco.loadAnns(self.coco.getAnnIds(id_ - 1))
```
adamjstewart, line 315: I wonder how I got this working in the past if the images and labels were off by one...

requirements/min.old
ashnair1: The older versions kept throwing a  `AttributeError:'scipy.signal._max_len_seq_inner.array' has no attribute '__reduce_cython__'`.

torchgeo/datasets/nwpu.py
ashnair1: The reason for this is because of the way `id_` was computed (`id_ = index % len(self) + 1`) which caused it to start from 1. The computation is fine for `load_image` because the images are labelled `001.jpg`, `002.jpg` so `f"{id_:03d.jpg}"` worked. But for `load_target` we needed the offset because we're passing in the image id.

I'm not sure it was ever working, to be honest. The tests only asserted that images and labels were loaded as expected and ensured the length of the dataset was either 650 or 150 (depending on the split).

requirements/min.old
adamjstewart: Locally or in CI?

ashnair1: CI

adamjstewart (approved):

adamjstewart:
Will move `nwpu` to `vhr10` in a follow-up PR.

adamjstewart (approved):

adamjstewart (approved):

calebrob6:
image

LGTM!

PR #852, opened 2022-10-15
Docs/trainers for Classification and Segmentation Task

nilsleh:
- I thought it would be useful to add some documentation to the trainers, just to clarify how flexible they are and the wide choices of models they support without someone having to dig.
- The documentation for `ClassificationTask` states that possible weights are `Either “random”, “imagenet_only”, “imagenet_and_random”, or “random_rgb”` but only "imagenet" or "random" work otherwise you get an error, so I corrected the documentation for that
- In the documentation there were keyword args missing that were actually necessary to use the task with a trainer, so I added documentation for the required keywords

### DIFF
--- a/torchgeo/trainers/classification.py
+++ b/torchgeo/trainers/classification.py
@@ -72,7 +72,7 @@ class ClassificationTask(pl.LightningModule):
             )
         else:
             raise ValueError(
-                f"Model type '{classification_model}' is not a valid timm model."
+                f"Model type '{classification_model}' is not a valid timm model. "
                 f"Valid model types are {valid_models}."
             )
 
--- a/torchgeo/trainers/classification.py
+++ b/torchgeo/trainers/classification.py
@@ -34,10 +34,23 @@ Linear.__module__ = "nn.Linear"
 
 
 class ClassificationTask(pl.LightningModule):
-    """LightningModule for image classification."""
+    """LightningModule for image classification.
+
+    Supports any available `Timm model
+    <https://rwightman.github.io/pytorch-image-models/>`_
+    as an architecture choice.
+
+    .. versionchanged:: 0.4.0
+        Add documentation about available classification models.
+    """
 
     def config_model(self) -> None:
-        """Configures the model based on kwargs parameters passed to the constructor."""
+        """Configures the model based on kwargs parameters passed to the constructor.
+
+        .. versionchanged:: 0.4.0
+            Add documentation about available pretrained weights and available
+            Timm models.
+        """
         in_channels = self.hyperparams["in_channels"]
         classification_model = self.hyperparams["classification_model"]
 
@@ -68,6 +81,7 @@ class ClassificationTask(pl.LightningModule):
         else:
             raise ValueError(
                 f"Model type '{classification_model}' is not a valid timm model."
+                f"Valid model types are {valid_models}."
             )
 
         if custom_pretrained:
@@ -99,8 +113,16 @@ class ClassificationTask(pl.LightningModule):
         Keyword Args:
             classification_model: Name of the classification model use
             loss: Name of the loss function
-            weights: Either "random", "imagenet_only", "imagenet_and_random", or
-                "random_rgb"
+            weights: Either "random" or "imagenet"
+            num_classes: Number of prediction classes
+            in_channels: Number of input channels to model
+            learning_rate: Learning rate for optimizer
+            learning_rate_schedule_patience: Patience for learning rate scheduler
+
+        :: versionchanged:: 0.4.0
+            Add keyword arguments 'num_classes', 'in_channels', 'learing_rate' and
+            'learning_rate_schedule_patience' and change supported
+            weights name.
         """
         super().__init__()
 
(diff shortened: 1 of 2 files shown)


torchgeo/trainers/classification.py
```diff
@@ -34,10 +34,23 @@
...
 class ClassificationTask(pl.LightningModule):
-    """LightningModule for image classification."""
+    """LightningModule for image classification.
+
+    Supports any available `Timm model
+    <https://rwightman.github.io/pytorch-image-models/>`_
+    as an architecture choice.
+
+    .. versionchanged:: 0.3.2
+        Add documentation about available classification models.
```
adamjstewart, line 44: No need to document documentation improvements

adamjstewart (commented):
Thanks, this was much needed!

torchgeo/trainers/segmentation.py
```diff
@@ -29,10 +29,21 @@
...
+    `TIMM encoders <https://smp.readthedocs.io/en/latest/encoders_timm.html>`_.
+    """
 
     def config_task(self) -> None:
-        """Configures the task based on kwargs parameters passed to the constructor."""
+        """Configures the task based on kwargs parameters passed to the constructor.
+
+        .. versionchanged:: 0.3.2
+            Improve error message of supported segmentation architectures
+            and loss function.
```
adamjstewart, line 45: Same here. versionchanged/added should only ever be 2 digits because anything worth documenting is either a new feature or a change in backwards compatibility.

nilsleh: ah okay, I wasn't aware of the exact definition, I thought since I changed the code within the function, I should track it with versionchanged

adamjstewart: Nah, only API or behavior changes

torchgeo/trainers/segmentation.py
```diff
@@ -56,6 +62,7 @@ def config_task(self) -> None:
         else:
             raise ValueError(
                 f"Model type '{self.hyperparams['segmentation_model']}' is not valid."
+                f"Currently, only support 'unet', 'deeplabv3+' and 'fcn'."
```
adamjstewart, line 65:
```suggestion
                f"Currently, only supports 'unet', 'deeplabv3+' and 'fcn'."
```

torchgeo/trainers/segmentation.py
```diff
@@ -70,7 +77,10 @@ def config_task(self) -> None:
                 "multiclass", ignore_index=self.ignore_index, normalized=True
             )
         else:
-            raise ValueError(f"Loss type '{self.hyperparams['loss']}' is not valid.")
+            raise ValueError(
+                f"Loss type '{self.hyperparams['loss']}' is not valid."
+                f"Currently, support 'ce', 'jaccard' or 'focal' loss."
```
adamjstewart, line 82:
```suggestion
                f"Currently, supports 'ce', 'jaccard' or 'focal' loss."
```

torchgeo/trainers/classification.py
```diff
@@ -68,6 +73,7 @@ def config_model(self) -> None:
         else:
             raise ValueError(
                 f"Model type '{classification_model}' is not a valid timm model."
```
adamjstewart, line 75:
```suggestion
                f"Model type '{classification_model}' is not a valid timm model. "
```

torchgeo/trainers/segmentation.py
```diff
@@ -56,6 +62,7 @@ def config_task(self) -> None:
         else:
             raise ValueError(
                 f"Model type '{self.hyperparams['segmentation_model']}' is not valid."
```
adamjstewart, line 64:
```suggestion
                f"Model type '{self.hyperparams['segmentation_model']}' is not valid. "
```

torchgeo/trainers/segmentation.py
```diff
@@ -70,7 +77,10 @@ def config_task(self) -> None:
                 "multiclass", ignore_index=self.ignore_index, normalized=True
             )
         else:
-            raise ValueError(f"Loss type '{self.hyperparams['loss']}' is not valid.")
+            raise ValueError(
+                f"Loss type '{self.hyperparams['loss']}' is not valid."
```
adamjstewart, line 81:
```suggestion
                f"Loss type '{self.hyperparams['loss']}' is not valid. "
```

adamjstewart (dismissed):

torchgeo/trainers/classification.py
```diff
@@ -67,7 +72,8 @@ def config_model(self) -> None:
             )
         else:
             raise ValueError(
-                f"Model type '{classification_model}' is not a valid timm model."
+                f"Model type '{classification_model}' is not a valid timm model. "
+                f"Valid model types are {valid_models}."
```
isaaccorley, line 76: This is a very long list. I don't think we want to print this.

nilsleh: Fair point, I added a code block in the docstring instead to see available models. Is that okay?

PR #853, opened 2022-10-16
Fix Vaihingen datamodule

### LINKED ISSUE #851: Vaihingen datamodule
### Description

I would expect that with a existing Vaihingen datamodule, I only need to define a segmentation task and a `pl.Trainer` to train a model on this dataset (but maybe this expectation is wrong). However, the Vaihingen dataset has variable sized image dimensions, and therefore one cannot specify a `batch_size>1` because otherwise tensors cannot be stacked. So either there should be a collate function for the dataloaders in the datamodule or some comment in the documentation because the default batch_size of the datamodule is 64.

### Steps to reproduce

```
from torchgeo.datamodules import Vaihingen2DDataModule

datamodule = Vaihingen2DDataModule(root="./data/Vaihingen")

    segmentation_model="unet",
    encoder_name="resnet18",
    encoder_weights="imagenet",
    in_channels=3,
    num_classes=6,
    loss="jaccard",
    ignore_index=None,
    learning_rate=0.001,
    learning_rate_schedule_patience=5
)

trainer = pl.Trainer(
    fast_dev_run=True,
    enable_progress_bar=False
)

trainer.fit(
    model=task,
    datamodule=datamodule
)
```

### Version

0.4.0.dev0

nilsleh:
This PR closes #851. As suggested, based on the `OSCDDatamodule`, random patch crops are taken during training. For validation and testing the batch size is fixed to 1 like in `OSCDDatamodule` but instead of a fixed padding size, each sample is padded to the next larger multiple of 32 for the encoder-decoder architectures.


torchgeo/datamodules/vaihingen.py
```diff
@@ -3,10 +3,14 @@
...
 
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
+import kornia.augmentation as K
 import pytorch_lightning as pl
+import torch
+from einops import repeat
 from torch.utils.data import DataLoader, Dataset
+from torch.utils.data._utils.collate import default_collate
```
adamjstewart, line 13:
```suggestion
from torch.utils.data import DataLoader, Dataset, default_collate
```
We shouldn't import from underscore files, those aren't public

torchgeo/datamodules/vaihingen.py
```diff
@@ -23,26 +27,44 @@ class Vaihingen2DDataModule(pl.LightningDataModule):
...
         val_split_pct: float = 0.2,
         **kwargs: Any,
     ) -> None:
         """Initialize a LightningDataModule for Vaihingen2D based DataLoaders.
 
         Args:
-            batch_size: The batch size to use in all created DataLoaders
+            train_batch_size: The batch size used in the train DataLoader
+                (val_batch_size == test_batch_size == 1). The effective batch size
+                will be 'train_batch_size' * 'num_patches_per_tile'
```
adamjstewart, line 42: Can you also clarify this in `OSCDDataModule`?

torchgeo/datamodules/vaihingen.py
```diff
@@ -23,26 +27,44 @@ class Vaihingen2DDataModule(pl.LightningDataModule):
 
     def __init__(
         self,
-        batch_size: int = 64,
+        train_batch_size: int = 32,
         num_workers: int = 0,
+        patch_size: Tuple[int, int] = (64, 64),
```
adamjstewart, line 32: I know this is the same in `OSCDDataModule`, but personally I think it would be better to support both ints and tuples of ints. This also matches how our samplers work. Doesn't necessarily have to change in this PR, could be a follow-up PR.

torchgeo/datamodules/vaihingen.py
```diff
@@ -23,26 +27,44 @@ class Vaihingen2DDataModule(pl.LightningDataModule):
...
 
         Args:
-            batch_size: The batch size to use in all created DataLoaders
+            train_batch_size: The batch size used in the train DataLoader
+                (val_batch_size == test_batch_size == 1). The effective batch size
+                will be 'train_batch_size' * 'num_patches_per_tile'
             num_workers: The number of workers to use in all created DataLoaders
             val_split_pct: What percentage of the dataset to use as a validation set
+            patch_size: Size of random patch from image and mask (height, width), should
+                be a multiple of 32 for most segmentation architectures
```
adamjstewart, line 46: Can you clarify this in other data modules as well?

torchgeo/datamodules/vaihingen.py
```diff
@@ -23,26 +27,44 @@ class Vaihingen2DDataModule(pl.LightningDataModule):
...
+            num_patches_per_tile: number of random patches per sample
             **kwargs: Additional keyword arguments passed to
                 :class:`~torchgeo.datasets.Vaihingen2D`
+
+        :: versionchanged:: 0.4
```
adamjstewart, line 51:
```suggestion
        .. versionchanged:: 0.4
```

torchgeo/datamodules/vaihingen.py
```diff
@@ -63,61 +85,91 @@ def setup(self, stage: Optional[str] = None) -> None:
 
         Args:
             stage: stage to set up
+
+        :: versionchanged:: 0.4
```
adamjstewart, line 89:
```suggestion
        .. versionchanged:: 0.4
```

torchgeo/datamodules/vaihingen.py
```diff
@@ -63,61 +85,91 @@ def setup(self, stage: Optional[str] = None) -> None:
...
+            Add functionality to randomly crop patches from a tile during
+            training and pad validation and test samples to next multiple of 32
         """
-        transforms = Compose([self.preprocess])
 
-        dataset = Vaihingen2D(split="train", transforms=transforms, **self.kwargs)
+        def n_random_crop(sample: Dict[str, Any]) -> Dict[str, Any]:
```
adamjstewart, line 94: Needs docstring

torchgeo/datamodules/vaihingen.py
```diff
@@ -63,61 +85,91 @@ def setup(self, stage: Optional[str] = None) -> None:
...
+                mask = repeat(sample["mask"], "h w -> t h w", t=2).float()
+                image, mask = self.rcrop(sample["image"], mask)
+                mask = mask.squeeze()[0]
+                images.append(image.squeeze())
+                masks.append(mask.long())
+            sample["image"] = torch.stack(images).squeeze(0)
+            sample["mask"] = torch.stack(masks).squeeze(0)
+            return sample
+
+        def pad_to(sample: Dict[str, Any]) -> Dict[str, Any]:
```
adamjstewart, line 106: Docstring missing args and returns sections

torchgeo/datamodules/vaihingen.py
```diff
@@ -63,61 +85,91 @@ def setup(self, stage: Optional[str] = None) -> None:
...
         self.test_dataset = Vaihingen2D(
-            split="test", transforms=transforms, **self.kwargs
+            split="test", transforms=test_transforms, **self.kwargs
         )
 
     def train_dataloader(self) -> DataLoader[Any]:
-        """Return a DataLoader for training.
+        """Return a DataLoader for training."""
+
+        def collate_wrapper(batch: List[Dict[str, Any]]) -> Dict[str, Any]:
```
adamjstewart, line 149: Missing docstring

torchgeo/datamodules/vaihingen.py
```diff
@@ -63,61 +85,91 @@ def setup(self, stage: Optional[str] = None) -> None:
...
+            self.val_dataset = train_dataset
 
         self.test_dataset = Vaihingen2D(
-            split="test", transforms=transforms, **self.kwargs
+            split="test", transforms=test_transforms, **self.kwargs
         )
 
     def train_dataloader(self) -> DataLoader[Any]:
-        """Return a DataLoader for training.
+        """Return a DataLoader for training."""
```
adamjstewart, line 147: Docstring missing returns section

adamjstewart (changes requested):
I'm sure some of these comments apply to other existing data modules too. Someday I would like to have all public and private methods properly documented. We're using `pydocstyle` to try to enforce this, but it's broken, so it doesn't complain like it should when methods aren't fully documented.

torchgeo/datamodules/vaihingen.py
```diff
@@ -63,24 +87,77 @@ def setup(self, stage: Optional[str] = None) -> None:
...
+                sample: contains image and mask tile from dataset
+
+            Returns:
+                stacked randomly cropped patches from input tile
+            """
+            images, masks = [], []
+            for i in range(self.num_patches_per_tile):
+                mask = repeat(sample["mask"], "h w -> t h w", t=2).float()
+                image, mask = self.rcrop(sample["image"], mask)
+                mask = mask.squeeze()[0]
```
isaaccorley, line 109: There's a lot of squeezing going on here. It's better to specify which singleton dimension your are removing. I think if you don't squeeze then you can replace torch.stack with torch.cat.

torchgeo/datamodules/vaihingen.py
```diff
@@ -89,10 +166,27 @@ def train_dataloader(self) -> DataLoader[Any]:
...
+
+            Args:
+                batch: sample batch from dataloader containing image and mask
+
+            Returns:
+                sample batch where the batch dimension is
+                'train_batch_size' * 'num_patches_per_tile'
+            """
+            r_batch: Dict[str, Any] = default_collate(batch)
+            r_batch["image"] = torch.flatten(r_batch["image"], 0, 1)
```
isaaccorley, line 181: Nit: einops.rearrange would be more readable here

isaaccorley (dismissed):

torchgeo/datamodules/vaihingen.py
adamjstewart: Yes, this is really important so that `batch_size=1` works correctly (often used in our unit tests)

nilsleh: I tested the last commit with `batch_size=1` and `num_patches_per_tile=1`

nilsleh:
I don't really understand why the minimum test is failing.

torchgeo/datamodules/vaihingen.py
nilsleh: Not doing that is raising an error during minimum pytests: https://github.com/microsoft/torchgeo/actions/runs/3339526468/jobs/5528583740. `ImportError: cannot import name 'default_collate' from 'torch.utils.data' (/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/site-packages/torch/utils/data/__init__.py)`

calebrob6:
The best I can tell is that torch is throwing a UserWarning about the `align_corners` parameter to an interpolate method. This is called for some reason when we use kornia's RandomCrop. I tried setting it explicitly...

calebrob6:
Okay no idea. @adamjstewart any suggestions? Can we bump kornia's minimum version or ignore this warning?

adamjstewart:
Since this is a warning (not an error) and the warning has been removed in newer versions of PyTorch, I went with silencing the warning over bumping the minimum version of PyTorch. Latest commit should hopefully get the tests to pass.

torchgeo/datamodules/vaihingen.py
adamjstewart: Alright, we can change this next time we bump the minimum version of PyTorch. It looks like it isn't public until 1.11.

torchgeo/datamodules/vaihingen.py
```diff
@@ -3,13 +3,19 @@
...
+import kornia.augmentation as K
 import matplotlib.pyplot as plt
 import pytorch_lightning as pl
+import torch
+from einops import rearrange
 from torch.utils.data import DataLoader, Dataset
+from torch.utils.data._utils.collate import default_collate
 from torchvision.transforms import Compose
 
+from torchgeo.samplers.utils import _to_tuple
```
adamjstewart, line 17:
```suggestion
from ..samplers.utils import _to_tuple
```
I prefer to use relative imports within torchgeo

torchgeo/datamodules/vaihingen.py
nilsleh: Caleb suggested in #876 a different approach and naming of this batch_size parameter. Should I change this PR accordingly?

adamjstewart: Added a comment there, we should decide on a final naming scheme in #876 and then use the same scheme here.

torchgeo/datamodules/vaihingen.py
```diff
@@ -24,26 +29,46 @@ class Vaihingen2DDataModule(pl.LightningDataModule):
...
+        num_patches_per_tile: int = 32,
         val_split_pct: float = 0.2,
         **kwargs: Any,
     ) -> None:
         """Initialize a LightningDataModule for Vaihingen2D based DataLoaders.
 
         Args:
-            batch_size: The batch size to use in all created DataLoaders
+            train_batch_size: The batch size used in the train DataLoader
+                (val_batch_size == test_batch_size == 1). The effective batch size
```
adamjstewart, line 43: Why does val/test have to have a batch size of 1?

nilsleh: That's how it is implemented in the OSCD module. I guess the idea was to have random crops during training but then evaluate on the entire uncropped image during val/test?

calebrob6: yes

calebrob6:
@adamjstewart do you know why rtd is saying "/home/docs/checkouts/readthedocs.org/user_builds/torchgeo/checkouts/853/docs/tutorials/transforms.ipynb:: WARNING: Pygments lexer name 'ipython3' is not known"? Is that why the build failed?

adamjstewart:
Should be fixed by #922, just need to rebase.

adamjstewart (approved):

PR #854, opened 2022-10-16
Change regression task to timm support

nilsleh:
This PR changes the `RegressionTask` to support Timm models instead of torchvision.models for greater model variety and consistency, because Timm is used in the other tasks as well.

- new keyword argument `num_outputs` which allows the regression task to have multiple regression outputs, either if there are multiple regression targets or some form of probabilistic neural networks
- other keyword arguments will be the same as `ClassificationTask` after #852 is merged with the exception of the loss argument because regression only has mse loss at the moment


torchgeo/trainers/regression.py
```diff
@@ -26,36 +26,70 @@
...
-
-        if parse(torchvision.__version__) >= parse("0.13"):
-            if pretrained:
-                kwargs = {
-                    "weights": getattr(
-                        torchvision.models, f"ResNet{model[6:]}_Weights"
-                    ).DEFAULT
-                }
+        in_channels = self.hyperparams["in_channels"]
+        regression_model = self.hyperparams["regression_model"]
```
adamjstewart, line 39: Currently there's a mix of `<task>_model` and `model` used throughout TorchGeo. I think it would be better to use `model` and `backbone`. We may also want to divide up the docs so that datamodules are listed by task and not by geo/nongeo. Maybe same for datasets too, haven't decided. But the docs can wait. Just wanted to mention the kwarg thing now since you're changing it.

torchgeo/trainers/regression.py
```diff
@@ -26,36 +26,70 @@
...
+            self.model = timm.create_model(
+                regression_model,
+                num_classes=self.hyperparams["num_outputs"],
+                in_chans=in_channels,
+                pretrained=imagenet_pretrained,
+            )
         else:
-            kwargs = {"pretrained": pretrained}
+            raise ValueError(
+                f"Model type '{regression_model}' is not a valid timm model."
```
adamjstewart, line 67: Should we list valid options like in your other PR?

torchgeo/trainers/regression.py
nilsleh: do you mean the argument `<task_model>` should be named just `model` for classification and regression and then instead of `self.model` and everywhere else we mention `model` like in the documentation, it should actually say `backbone`?

adamjstewart: What I'm suggesting is that `model` should refer to the name of the overall model, while `backbone` refers to the model backbone (if one exists). So for classification and regression, there would only be a `model` parameter (e.g., ResNet50), while for segmentation there would be a `model` (e.g., UNet) and a `backbone` (e.g., ResNet50).

Although SMP seems to call them encoder/decoder, not backbone. So maybe it would depend on the model?

calebrob6:
Pulling out discussion from the comments, I think it makes most sense to call the deep learning model, "model", in the trainer code. In semantic segmentation tasks this will be a "unet", "deeplabv3+", or whatever else smp accepts, as that is the "model" that is being used, elsewhere this will be any timm architecture. In the segmentation trainer, it makes sense to call the encoder, "encoder".

I.e. what Adam said above but "encoder" instead of "backbone" (it makes sense to use "backbone" for BYOL and other SSL trainers).

nilsleh:
Let

> Pulling out discussion from the comments, I think it makes most sense to call the deep learning model, "model", in the trainer code. In semantic segmentation tasks this will be a "unet", "deeplabv3+", or whatever else smp accepts, as that is the "model" that is being used, elsewhere this will be any timm architecture. In the segmentation trainer, it makes sense to call the encoder, "encoder".
> 
> I.e. what Adam said above but "encoder" instead of "backbone" (it makes sense to use "backbone" for BYOL and other SSL trainers).

Let me know if this is what you had in mind and I understood you correctly. I will then open another PR to change the segmentation task names.

calebrob6:
Yep, exactly that!

torchgeo/trainers/regression.py
```diff
@@ -26,36 +26,74 @@
...
+                )
             else:
-                kwargs = {"weights": None}
+                imagenet_pretrained = self.hyperparams["weights"] == "imagenet"
+            custom_pretrained = False
+        else:
+            custom_pretrained = True
+
+        # Create the model
+        valid_models = timm.list_models(pretrained=True)
```
adamjstewart, line 63: Shouldn't this be:
```suggestion
        valid_models = timm.list_models(pretrained=False)
```
That way you can still use a model without pre-trained weights, or with custom pre-trained weights.

adamjstewart (commented):
One minor change and I think this is ready to go.

adamjstewart (approved):
Would love to get one more review from someone before merging

nilsleh:
> I.e. what Adam said above but "encoder" instead of "backbone" (it makes sense to use "backbone" for BYOL and other SSL trainers).

@adamjstewart looking back through the comments, I interpret the following:
- Segmentation task: segmentation_model -> model and encoder_name -> encoder
- Regression task: regression_model -> model
- Classifcation task: classification_model -> model
- BYOL task: encoder_name -> encoder (although Caleb seems to suggest encoder_name -> backbone?)
- Object Detection: detection_model -> model and backbone -> ?

adamjstewart:
Hmm, I would personally use "model" for the high-level model in all tasks and "encoder" iff that model has the ability to replace some kind of encoder/backbone with a pre-trained one. I guess I don't see a big distinction between SL encoders and SSL backbones, but I'll let @calebrob6 clarify.

calebrob6:
Encoder or backbone is fine for BYOL

nilsleh:
Okay so,
- Segmentation task: segmentation_model -> model and encoder_name -> encoder
- Regression task: regression_model -> model
- Classifcation task: classification_model -> model
- BYOL task: encoder_name -> encoder
- Object Detection: detection_model -> model 
is the final verdict?

calebrob6:
LGTM!

torchgeo/trainers/regression.py
```diff
@@ -26,36 +26,74 @@
...
+                )
             else:
-                kwargs = {"weights": None}
+                imagenet_pretrained = self.hyperparams["weights"] == "imagenet"
+            custom_pretrained = False
+        else:
+            custom_pretrained = True
+
+        # Create the model
+        valid_models = timm.list_models(pretrained=False)
```
adamjstewart, line 63:
```suggestion
        valid_models = timm.list_models(pretrained=imagenet_pretrained)
```

adamjstewart (approved):

ashnair1 (approved):

PR #861, opened 2022-10-20
Fix a bug for ETCI2021 loader

daiki-kimura:
### Summary

Fix the ETCI 2021 Data loader by removing glob for each folder

### Details

I tried ETCI2021DataModule (on Mac), then found a bug for loading the images.
The loaded images are sometimes not correct by each sorted glob

You can find the image file paths are not same (of vv and water_mask) in attached result.
Screen Shot 2022-10-20 at 18 16 52

This PR changes the loader will do sorted glob once in vv folder, and then use it for others.

### DIFF
--- a/torchgeo/datasets/etci2021.py
+++ b/torchgeo/datasets/etci2021.py
@@ -171,17 +171,15 @@ class ETCI2021(NonGeoDataset):
         folders = [os.path.join(folder, "tiles") for folder in folders]
         for folder in folders:
             vvs = sorted(glob.glob(os.path.join(folder, "vv", "*.png")))
-            vhs = [
-                vv.replace("/vv/", "/vh/").replace("_vv.png", "_vh.png") for vv in vvs
-            ]
+            vhs = [vv.replace("vv", "vh") for vv in vvs]
             water_masks = [
-                vv.replace("/vv/", "/water_body_label/").replace("_vv.png", ".png")
+                vv.replace("_vv.png", ".png").replace("vv", "water_body_label")
                 for vv in vvs
             ]
 
             if split != "test":
                 flood_masks = [
-                    vv.replace("/vv/", "/flood_label/").replace("_vv.png", ".png")
+                    vv.replace("_vv.png", ".png").replace("vv", "flood_label")
                     for vv in vvs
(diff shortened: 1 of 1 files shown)


daiki-kimura:
@microsoft-github-policy-service agree [company="IBM"]

daiki-kimura:
@microsoft-github-policy-service agree company="IBM"

torchgeo/datasets/etci2021.py
```diff
@@ -171,25 +171,56 @@ def _load_files(self, root: str, split: str) -> List[Dict[str, str]]:
...
+            ]
 
             if split != "test":
-                flood_masks = sorted(
-                    glob.glob(os.path.join(folder, "flood_label", "*.png"))
-                )
+                flood_masks = [
+                    vv.replace("/vv/", "/flood_label/").replace("_vv.png", ".png")
+                    for vv in vvs
+                ]
```
ashnair1, line 186: Forward slashes seems to be causing issues in Windows.

daiki-kimura: Yeah, thank you. I add the fix.
(there is no "vv" in region name for ETCI 2021)

torchgeo/datasets/etci2021.py
```diff
@@ -171,25 +171,54 @@ def _load_files(self, root: str, split: str) -> List[Dict[str, str]]:
...
+                        print(
+                            "Not found: %s"
+                            % (
+                                vh
+                                if not os.path.exists(vh)
+                                else water_mask
+                                if not os.path.exists(water_mask)
+                                else flood_mask
+                            )
+                        )
```
adamjstewart, line 212: Do these situations actually occur? Are there any examples in the dataset that are missing certain layers?

daiki-kimura: Actually it doesn't use. I put it just in case. 
I remove it

adamjstewart (approved):
Any idea why the sorted glob didn't work? Are there additional files that were causing things to be shifted?

daiki-kimura:
I think this is the reason:
Screen Shot 2022-10-21 at 0 23 56

PR #862, opened 2022-10-20
Allow multilabels in VectorDataset

pmandiola:
This PR modifies VectorDataset to handle multiple labels. Without this, masks built from shapefiles are always coded into binary masks.

An optional parameter `label_name` is added to VectorDataset to indicate the name of the property where the label is stored in the shapefile. This label is then rasterized to the mask, instead of just a 1.

Here is an example of a mask from a VectorDataset using this option:

image

I added an additional test to VectorDataset, but I'm not sure its the right way to do it. Let me know if it's ok or I can change it if needed.


tests/datasets/test_geo.py
```diff
@@ -52,6 +52,10 @@ class CustomVectorDataset(VectorDataset):
     filename_glob = "*.geojson"
 
 
+class CustomVectorDatasetMultilabel(VectorDataset):
+    filename_glob = "*.geojson"
```
adamjstewart, line 56: Probably don't need a separate class for this since it's identical

adamjstewart (commented):
This looks great! One small comment on how the tests could be simplified.

torchgeo/datasets/geo.py
```diff
@@ -545,6 +546,8 @@ def __init__(
             res: resolution of the dataset in units of CRS
             transforms: a function/transform that takes input sample and its target as
                 entry and returns a transformed version
+            label_name: name of the dataset property that has the label to be
+                rasterized into the mask
 
         Raises:
             FileNotFoundError: if no files are found in ``root``
```
adamjstewart, line 553: I would add something like:
```rst
.. versionadded:: 0.4
   The *label_name* parameter.
```

adamjstewart (approved):

adamjstewart:
LGTM, I think the only remaining task is to read and accept the CLA: https://github.com/microsoft/torchgeo/pull/862#issuecomment-1285997103

Issue #30, opened 2021-07-02
Add stitching utilities


adamjstewart:
We should add a `torchgeo.utils` package that provides utilities for common operations, including stitching together patches to create a prediction on an entire tile. This is equivalent to [`torchvision.utils`](https://pytorch.org/vision/stable/utils.html). See https://arxiv.org/pdf/1805.12219.pdf for a survey of common stitching techniques that we should support. This includes clipping (best default) and averaging. May also want to add weighted averaging. Let's see what libraries exist for this. If rasterio can do this for us for free, that would be awesome.

Z-Zheng:
See here https://github.com/BloodAxe/pytorch-toolbelt

robmarkcole:
Some nice work from @wrignj08 -> https://github.com/DPIRD-DMA/MultiClean
Also relevant are https://github.com/zsylvester/segmenteverygrain and https://github.com/Multihuntr/seamless-seg

PR #866, opened 2022-10-24
Add splitting utilities for GeoDatasets

pmandiola:
This PR adds the ability to create train/val/test RasterDatasets from the same underlying raster files, inspired from a functionality in the rastervision package.

An `extent_crop` param is added to RasterDataset to indicate the portion of the raster files to be skipped. This way, one could create a train RasterDataset using only the top 80% of the files by specifying `extent_crop=(0.2, 0, 0, 0)`, and a test RasterDataset using the remaining 20% with `extent_crop=(0, 0, 0.8, 0)`.

I got a bit stuck with how to implement a test for this, but before figuring it out I wanted to check if this looks ok to you.


adamjstewart:
This feature serves the same purpose as `GeoSampler`'s `roi` parameter. The only difference is that `roi` has to be in geospatial coordinates.

The reason I originally chose to put this parameter in the sampler instead of the dataset is in order to save time. Instantiating a dataset is kinda slow (you need to recursively search the filesystem for files), so if you only need to instantiate 1 dataset and 3 samplers, that's much faster than 3 datasets and 1 sampler.

That doesn't mean our current design is ideal, and I agree we at least need an easier way to easily split a dataset without knowledge of geospatial metadata. In addition to the split strategy you propose here, I would also love a way to overlay a grid over a dataset and randomly distribute cells to different splits (see Figure 2 in https://arxiv.org/abs/1805.02855 for an example of what I mean).

I feel like we should add a `torchgeo.utils` directory with utilities for splitting and joining datasets. This somewhat relates to #30. I think we'll find that there are too many ways to split a dataset and a single parameter in either the dataset or sampler won't cut it. What do you think?

pmandiola:
One big difference with the GeoSampler's roi is when the dataset has non-contiguos rasters. In our specific problem, we have one raster per city so all the raster files in the RasterDataset are non-contiguos. Using one roi for the full dataset through the GeoSampler doesn't make sense in this case. Or at least I wouldn't know how to make it work.

With this method, when the RasterDataset is loaded, each file is cropped by reducing the bounds that get added to the index. So for our specific case with cities, we can easily divide each raster/city into train/val/test datasets. Also, not needing the CRS beforehand is a big advantage.

I don't know where it would be best to put all these splitting utilities but I *think*  doing it on the dataset level gives you more flexibility. The GeoSampler's roi parameter can be added to the dataset, but I don't think it's possible to do the same we are doing here.

pmandiola:
@adamjstewart I used the same original logic for splitting and turned it into a function that could go into utils. What do you think of this approach? It might be possible to do the same with a GeoSampler by splitting its underlying dataset

```
def train_test_split(
    dataset: GeoDataset,
    test_size: float = 0.25,
    random_seed: int = None,
) -> List[GeoDataset]:

    assert 0 < test_size < 1
    
    np.random.seed(random_seed)

    index_train = Index(interleaved=False, properties=Property(dimension=3))
    index_test = Index(interleaved=False, properties=Property(dimension=3))

    for i, hit in enumerate(dataset.index.intersection(dataset.index.bounds, objects=True)):
        box = BoundingBox(*hit.bounds)
        horizontal, flip = np.random.randint(2, size=2)
        if flip:
            box_train, box_test = box.split(1 - test_size, horizontal)
        else:
            box_test, box_train = box.split(test_size, horizontal)
        index_train.insert(i, tuple(box_train))
        index_test.insert(i, tuple(box_test))
    
    dataset_train = deepcopy(dataset)
    dataset_train.index = index_train
    dataset_test = deepcopy(dataset)
    dataset_test.idnex = index_test

    return dataset_train, dataset_test
```

Also added a split() method to BoundingBox:

```
  def split(
      self,
      proportion: float,
      horizontal: bool = True,
  ) -> "BoundingBox":
      """Split BoundingBox in two.

      Args:
          proportion: split proportion
          horizontal: whether the split is horizontal (True) or
              vertical

      Returns:
          The cropped BoundingBox
      """
      
      if horizontal:
          w = self.maxx - self.minx
          splitx = self.minx + int(w * proportion)
          bbox1 = BoundingBox(self.minx, splitx, self.miny, self.maxy, self.mint, self.maxt)
          bbox2 = BoundingBox(splitx, self.maxx, self.miny, self.maxy, self.mint, self.maxt)
      else:
          h = self.maxy - self.miny
          splity = self.miny + int(h * proportion)
          bbox1 = BoundingBox(self.minx, self.maxx, self.miny, splity, self.mint, self.maxt)
          bbox2 = BoundingBox(self.minx, self.maxx, splity, self.maxy, self.mint, self.maxt)

      return bbox1, bbox2
```

adamjstewart:
I'm currently preparing for a conference, but I'll try to look at this in more detail next week. I'll have to decide whether splitting should happen at the dataset or sampler level.

pmandiola:
Hi @adamjstewart, any thoughts on this dataset/sampler splitting strategy? Let me know what you think so I can finalize it or adapt it to work for samplers instead.

One pro of splitting at the dataset level is that later one could use different sampling strategies for train/val/test, like using a RandomSampler for training and then a GridSampler for validation and testing.

adamjstewart:
> What do you think about merging both into one `random_split` function and use the dataset type to decide how to do the split?

I'm worried about this because in the case of GeoDataset there are several different ways you might want to split the data. So I don't think having a single function will work. That's why we need to enumerate all possible splitting strategies so we can decide on function names.

pmandiola:
Ok, I see your point. Let's list them then and see how we can move forward. I'll start with some we have discussed here:

1. **Random split for NonGeoDatasets.** This would be the usual way of splitting a torch dataset, currently implemented in the `dataset_split` function in `torchgeo.datamodules.utils`

2. **Randomly split each bounding box in a GeoDataset's index** This is one way of randomly split a GeoDataset, as implemented in this PR.

3. **Randomly assign each bounding box in a GeoDataset's index to new GeoDatasets.** A very simple way of randomly splitting a GeoDataset, not implemented.

4. **Overlay a grid over each bounding box in a GeoDataset's index and randomly distribute cells to different splits.** As described in https://arxiv.org/abs/1805.02855, not implemented.

5. **Split a GeoDataset by defining the ROI for each new dataset.** Similar to what is now implemented with the GeoSampler's `roi` parameter, but in a separate function allowing multiple roi's to be defined for multiple resulting GeoDatasets.

6. **Split a GeoDataset by it's time dimension.** Not sure if time series forecasting is already implemented in torchgeo, but this would be useful for doing it.

pmandiola:
I went forward and re-wrote the definitions into a more functional form and included a proposal for the function names. I think it would make a lot of sense to have this in a new `torchgeo.datasets.splits` file.

What are your thoughts @adamjstewart? Any other strategy you would add at this stage?

**Regular splits:**

`random_nongeo_split`: randomly splits a NonGeoDataset into non-overlapping new NonGeoDatasets.

**Spatial splits:**

`random_bbox_assignment`: randomly assigns each `BoundingBox` in a GeoDataset's index to new GeoDatasets.

`random_bbox_splitting`: splits each `BoundingBox` in a GeoDataset's index and randomly assigns each part to new GeoDatasets.

`random_grid_cell_assignment`: overlays a grid over each `BoundingBox` in a GeoDataset's index and randomly assigns each cell to new GeoDatasets.

`roi_split`: intersects a ROI with a GeoDataset for each desired new GeoDataset.

**Time splits:**

`time_series_split`: splits a GeoDataset along the time dimension into new GeoDatasets.

adamjstewart:
Those names sound good to me. How do you propose these splitters work? Would they accept a dataset as input and create multiple datasets? Or would they be arguments to a sampler?

pmandiola:
I'm thinking on a function that takes a dataset as an argument and returns multiple datasets, like the `train_test_split` function I wrote in this PR.

I'm not sure whether to ask for `val_pct` and `train_pct` parameters like in the current `torchgeo.datamodules.utils.dataset_split` implementation, or move to use a list of `lengths` like in torch's `random_split`.

adamjstewart:
The latest version of [random_split](https://pytorch.org/docs/stable/data.html#torch.utils.data.random_split) accepts both a list of lengths and a list of percentages. I would like to do the same, both for consistency with PyTorch and because it's more useful this way.

pmandiola:
@adamjstewart I implemented `random_bbox_assignment`, `random_bbox_splitting `, and `roi_split`. Also `random_nongeo_split`, but it just calls torch's `random_split` so I'm not sure it needs to be there (and tests are failing because I added a test with fractions, which are supported only in newer versions of torch).

I also had to add a small change to IntersectionDataset to avoid adding empty BoundingBoxes to the index.

I think I know how to implement `time_series_split`. But `random_grid_cell_assignment` I'm not so sure. If we want it to work exactly as in the paper (where each cell is a sample that is randomly distributed), I think it has to be implemented with a GridSampler. But we could modify it a little to work with datasets: each cell could be a subset of the BoundingBox, from which samples can be taken. However this would make the index much bigger, having more than one entry for the same underlying file.

Let me know what you think.

pmandiola:
Only `random_grid_cell_assignment` is missing now. I'm still not entirely sure how to implement it so I'm skipping it for now.

@adamjstewart if everything looks ok, probably something should be done with `torchgeo.datamodules.utils.dataset_split` and the GeoSampler's `roi` parameter which are now redundant. Not so sure if you'd like to deprecate them or not, and how to properly do it.

pmandiola:
All splitting strategies discussed above are now implemented. I think it's ready for review @adamjstewart.

Still unsure what to do with `torchgeo.datamodules.utils.dataset_split` and the GeoSampler's `roi` parameter. We could keep `torchgeo.datamodules.utils.dataset_split` for backward compatibility if needed but make it use `random_nongeo_split` under the hood. The GeoSampler's `roi` parameter can be useful for slightly different use cases so it might be fine to keep it.

adamjstewart:
Thanks! I have a few other big PRs I'm working on and need to review for the next release, but I'll try to get to your PR soon. I'm also on vacation next week so it may have to wait until the end of the month. I'll try to find other developers to review this in the meantime.

pmandiola:
@adamjstewart just wanted to check if this was still something you wanted to add to torchgeo. Let me know if there's anything I can do to help moving this forward

adamjstewart:
Thanks for the ping! This completely fell off my radar, let me review this now...

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,375 @@
...
+        A new GeoDataset.
+
+    .. versionadded:: 0.4
+    """
+    new_dataset = deepcopy(dataset)
+    new_dataset.index = index
+    return new_dataset
+
+
+def random_nongeo_split(
```
adamjstewart, line 66: We might actually be able to get rid of this. `torch.utils.data.random_split` now supports fractions as well, so we no longer need to do anything special, we can use the builtin method.

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,375 @@
...
+
+from rtree.index import Index, Property
+from torch import Generator, default_generator, randint, randperm
+from torch._utils import _accumulate
+from torch.utils.data import Subset, TensorDataset, random_split
+
+from ..datasets import GeoDataset, NonGeoDataset
+from .utils import BoundingBox
+
+__all__ = (
```
adamjstewart, line 18: For anything we want to be public, we should add it to `torchgeo/datasets/__init__.py`, everything else is considered to be private.

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,375 @@
...
+
+"""Dataset splitting utilities."""
+
+from copy import deepcopy
+from math import floor
+from typing import Any, List, Optional, Sequence, Tuple, Union
+
+from rtree.index import Index, Property
+from torch import Generator, default_generator, randint, randperm
+from torch._utils import _accumulate
```
adamjstewart, line 12: Is there another option we could use here that is public?

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,375 @@
...
+    lengths = [floor(frac * total) for frac in fractions]
+    remainder = int(total - sum(lengths))
+    # add 1 to all the lengths in round-robin fashion until the remainder is 0
+    for i in range(remainder):
+        idx_to_add_at = i % len(lengths)
+        lengths[idx_to_add_at] += 1
+    return lengths
+
+
+def _create_geodataset_like(dataset: GeoDataset, index: Index) -> GeoDataset:
```
adamjstewart, line 49: Not sure if it's worth it to make a separate function to replace 2 lines of code with 1 line of code

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,375 @@
...
+
+    Args:
+        dataset: dataset to be split
+        lengths: lengths or fractions of splits to be produced
+        generator: (optional) generator used for the random permutation
+
+    Returns
+        A list of the subset datasets.
+
+    .. versionadded:: 0.4
```
adamjstewart, line 103: These should be updated to 0.5 now, sorry we weren't able to get this into 0.4

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,375 @@
...
+) -> List[GeoDataset]:
+    """Overlays a grid over a GeoDataset and randomly assigns cells to new GeoDatasets.
+
+    This function will go through each BoundingBox in the GeoDataset's index, overlay
+    a grid over it, and randomly assign each cell to new GeoDatasets.
+
+    Args:
+        dataset: dataset to be split
+        fractions: fractions of splits to be produced
+        grid_size: (optional) size of the grid
```
adamjstewart, line 204: We should clarify whether size here means "number of grid cells per image" or "number of pixels per grid cell" or "size in meters of each grid cell". We had a similar confusion with our samplers when we first rolled them out. I'm guessing it's the first one, but we should clarify in the documentation which we mean. I don't _think_ we will need to support the other options, but if we do we can figure that out in a different PR.

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,375 @@
...
+                new_indexes[i].insert(j, tuple(new_box), hit.object)
+                j += 1
+
+    return [_create_geodataset_like(dataset, index) for index in new_indexes]
+
+
+def time_series_split(
+    dataset: GeoDataset, lengths: Sequence[Union[int, float, Tuple[int, int]]]
+) -> List[GeoDataset]:
+    """Split a GeoDataset on it's time dimension to create non-overlapping GeoDatasets.
```
adamjstewart, line 299:
```suggestion
    """Split a GeoDataset on its time dimension to create non-overlapping GeoDatasets.
```

torchgeo/datasets/splits.py
adamjstewart: These should also be added to `docs/api/datasets.rst` under the "Utilities" section.

adamjstewart (changes requested):
This looks great! A few minor comments on documentation and usage, but otherwise I like the API and organization. Once these comments are addressed, I'll give it a second pass and look at the tests and documentation more closely. I'll also see if anyone else wants to review.

We'll have to think about whether or not we want to deprecate the old splitting style (pass an ROI to a sampler) but that can happen in a separate PR. The solution you have here is clearly more flexible, and I don't think we suffer from the time complexity I complained about before because we're only querying the filesystem once, we don't have to do it every time we create a dataset.

torchgeo/datasets/splits.py
pmandiola: hmm... true it is only 2 lines, but this function is called at the end of every splitting function with a generator. I thought it looked cleaner this way instead of having the same for loop at the end of each function. I can get rid of it if you want.

adamjstewart: Let's get rid of it, the function definition plus docstring plus everything else is probably more lines of code than the code repetition. And it's not a particularly convoluted 2 lines of code either.

torchgeo/datasets/splits.py
pmandiola: Removed!

torchgeo/datasets/splits.py
pmandiola: Added the new functions to both

torchgeo/datasets/splits.py
pmandiola: changed for itertools' accumulate

tests/datasets/test_splits.py
```diff
@@ -0,0 +1,314 @@
...
+
+from torchgeo.datasets import GeoDataset
+from torchgeo.datasets.splits import (
+    random_bbox_assignment,
+    random_bbox_splitting,
+    random_grid_cell_assignment,
+    roi_split,
+    time_series_split,
+)
+from torchgeo.datasets.utils import BoundingBox
```
adamjstewart, line 18: Could import all of these from `torchgeo.datasets` if you want to test that the import alias works

tests/datasets/test_splits.py
```diff
@@ -0,0 +1,314 @@
...
+
+    # No overlap
+    assert len(train_ds & val_ds) == 0
+    assert len(val_ds & test_ds) == 0
+    assert len(test_ds & train_ds) == 0
+
+    # Union equals original
+    assert (train_ds | val_ds | test_ds).bounds == ds.bounds
+
+    # Test __get_item__
```
adamjstewart, line 76:
```suggestion
    # Test __getitem__
```

tests/datasets/test_splits.py
```diff
@@ -0,0 +1,314 @@
...
+    assert len(train_ds) == expected_lengths[0]
+    assert len(val_ds) == expected_lengths[1]
+    assert len(test_ds) == expected_lengths[2]
+
+    # No overlap
+    assert len(train_ds & val_ds) == 0
+    assert len(val_ds & test_ds) == 0
+    assert len(test_ds & train_ds) == 0
+
+    # Union equals original
```
adamjstewart, line 73: These tests are fantastically thorough!

torchgeo/datasets/geo.py
```diff
@@ -852,8 +852,10 @@ def _merge_dataset_indices(self) -> None:
             for hit2 in ds2.index.intersection(hit1.bounds, objects=True):
                 box1 = BoundingBox(*hit1.bounds)
                 box2 = BoundingBox(*hit2.bounds)
-                self.index.insert(i, tuple(box1 & box2))
-                i += 1
+                new_box = box1 & box2
+                if new_box.area > 0:
```
adamjstewart, line 856: I'm on the fence about this change. We also support point datasets like GBIF/iNaturalist/EDDMapS where each bbox is a single point. The idea is that you can load up a database of known locations for a tree species and find them in satellite imagery. I think this change may break that use case. 

Why was it necessary? I'm not opposed to considering it, but controversial changes tend to slow down PRs.

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,359 @@
...
+    "random_bbox_splitting",
+    "random_grid_cell_assignment",
+    "roi_split",
+    "time_series_split",
+)
+
+
+def _fractions_to_lengths(fractions: Sequence[float], total: int) -> Sequence[int]:
+    """Utility to divide a number into a list of integers according to fractions.
+
```
adamjstewart, line 28: May be useful to add something like:
```rst
Implementation based on :meth:`torch.utils.data.random_split`.
```
both for legal reasons but also for updating someday if we ever find a bug.

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,359 @@
...
+        lengths[idx_to_add_at] += 1
+    return lengths
+
+
+def random_bbox_assignment(
+    dataset: GeoDataset,
+    lengths: Sequence[Union[int, float]],
+    generator: Optional[Generator] = default_generator,
+) -> List[GeoDataset]:
+    """Split a GeoDataset randomly assigning its index's BoundingBoxes.
```
adamjstewart, line 52: Can you add another sentence after this clarifying how it's different from random_bbox_splitting?

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,359 @@
...
+    # add 1 to all the lengths in round-robin fashion until the remainder is 0
+    for i in range(remainder):
+        idx_to_add_at = i % len(lengths)
+        lengths[idx_to_add_at] += 1
+    return lengths
+
+
+def random_bbox_assignment(
+    dataset: GeoDataset,
+    lengths: Sequence[Union[int, float]],
```
adamjstewart, line 49: In typing, int is a subset of float, so you don't need to union:
```suggestion
    lengths: Sequence[float],
```

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,359 @@
...
+    generator: Optional[Generator] = default_generator,
+) -> List[GeoDataset]:
+    """Split a GeoDataset randomly assigning its index's BoundingBoxes.
+
+    Args:
+        dataset: dataset to be split
+        lengths: lengths or fractions of splits to be produced
+        generator: (optional) generator used for the random permutation
```
adamjstewart, line 57: I don't usually add (optional) since it's implied by the fact that the parameter has a default value (and shows up in the docs)

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,359 @@
...
+    if not (sum(lengths) == 1 or sum(lengths) == len(dataset)):
+        raise ValueError(
+            "Sum of input lengths must equal 1 or the length of dataset's index."
+        )
+
+    if any(n <= 0 for n in lengths):
+        raise ValueError("All items in input lengths must be greater than 0.")
+
+    if sum(lengths) == 1:
+        lengths = _fractions_to_lengths(lengths, len(dataset))
```
adamjstewart, line 73: I'm surprised typing doesn't complain about this since you're converting `Sequence[float]` to `Sequence[int]`.

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,359 @@
...
+    ]
+
+    new_indexes = [
+        Index(interleaved=False, properties=Property(dimension=3)) for _ in lengths
+    ]
+
+    for i, length in enumerate(lengths):
+        for j in range(length):  # type: ignore[arg-type]
+            hit = hits.pop()
+            new_indexes[i].insert(j, hit.bounds, hit.object)
```
adamjstewart, line 89: This feels inefficient but I'm still thinking of a more efficient way to do it.

torchgeo/datasets/splits.py
adamjstewart: Oh wait, that's why you have `ignore[arg-type]` everywhere. Can we create a new variable or explicitly cast to `Sequence[int]`?

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,359 @@
...
+    for i in range(remainder):
+        idx_to_add_at = i % len(lengths)
+        lengths[idx_to_add_at] += 1
+    return lengths
+
+
+def random_bbox_assignment(
+    dataset: GeoDataset,
+    lengths: Sequence[Union[int, float]],
+    generator: Optional[Generator] = default_generator,
```
adamjstewart, line 50: torch.randperm defaults to None instead of default_generator. Should we do the same?

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,359 @@
...
+        Index(interleaved=False, properties=Property(dimension=3)) for _ in fractions
+    ]
+
+    for i, hit in enumerate(
+        dataset.index.intersection(dataset.index.bounds, objects=True)
+    ):
+        box = BoundingBox(*hit.bounds)
+        fraction_left = 1.0
+
+        horizontal, flip = randint(0, 2, (2,), generator=generator)
```
adamjstewart, line 136: So the idea is we randomly choose whether to split horizontally or vertically? I wonder if users might want to explicitly choose one or the other. We could leave that for a future PR if someone wants it, don't want to make this PR any more complex than it already is.

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,359 @@
...
+        if not (sum(lengths) == 1 or sum(lengths) == totalt):  # type: ignore[arg-type]
+            raise ValueError(
+                "Sum of input lengths must equal 1 or the dataset's time length."
+            )
+
+        if any(n <= 0 for n in lengths):  # type: ignore[operator]
+            raise ValueError("All items in input lengths must be greater than 0.")
+
+        if sum(lengths) == 1:  # type: ignore[arg-type]
+            lengths = [totalt * f for f in lengths]  # type: ignore[operator]
```
adamjstewart, line 307: Same comment here, we should find a way to avoid silencing mypy

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,359 @@
...
+            raise ValueError(
+                "Pairs of timestamps in lengths can't be out of dataset's time bounds."
+            )
+
+        if any(  # type: ignore[misc]
+            start < x < end or start < y < end for x, y in lengths[i + 1 :]
+        ):
+            raise ValueError("Pairs of timestamps in lengths can't overlap.")
+
+        # remove one second from each BoundingBox's maxt to avoid overlapping
```
adamjstewart, line 335: In `torchgeo.datasets.utils.disambiguate_timestamp`, we remove 1 microsecond. Not sure if this really matters though, as long as overlap is avoided. I doubt anyone will be doing anything with satellite imagery at the millisecond resolution.

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,359 @@
...
+    ]
+
+    for i, hit in enumerate(
+        dataset.index.intersection(dataset.index.bounds, objects=True)
+    ):
+        box = BoundingBox(*hit.bounds)
+        fraction_left = 1.0
+
+        horizontal, flip = randint(0, 2, (2,), generator=generator)
+        for j, frac in enumerate(fractions):
```
adamjstewart, line 137: How does this work when len(fractions) > 2? Do we just choose a random direction (horizontal, vertical) and then split a square into rectangles with a certain fractional width?

adamjstewart (commented):
There's a ton of tedious error-prone code required here (a couple releases back we fixed a ton of floating point rounding issues for stuff like this) but your unit tests are so thorough it gives me confidence that things are likely bug-free, nice job!

@calebrob6 is planning on drawing up some pictures to illustrate what each splitter does in a follow-up PR, should help with the docstrings.

Most of my review comments are clarification comments or type hint suggestions, I couldn't actually find any bugs or missing features.

torchgeo/datasets/splits.py
```diff
@@ -0,0 +1,359 @@
...
+    new_datasets = []
+    for index in new_indexes:
+        ds = deepcopy(dataset)
+        ds.index = index
+        new_datasets.append(ds)
+
+    return new_datasets
+
+
+def time_series_split(
```
adamjstewart, line 278: @nilsleh can you review this one? I know you do a lot of time-series analysis work. Basically wondering if this seems like a good all-purpose technique for time-series splitting, or if there are so many different ways someone might want to perform a time-series split that we need more specialized functions.

One that thing that's never been clear to me (in general, not talking about this PR) is how to integrate time-series analysis (where you might want to do rolling window predictions) into something like PyTorch Lightning (where everything is designed for train-val-test splits).

nilsleh (changes requested):
I think this is a super convenient utility, thanks a lot @pmandiola! Looking ahead with the WIP of #877 there are likely to be some other changes/adaptions since we have not settled on an implementation for Time Series Support. However, I think this is a great starting point and convenient to generally create time series splits from more difficult to inspect raster data as supposed to a dataframe where you can just slice your data. Another case important to spatio-temporal data looking ahead would be to create splits both across time and geospatial location.

With respect to your question about time-series analysis and rolling windows @adamjstewart , I have made some comments in the time-series support PR. So I think we can discuss that there.

torchgeo/datasets/geo.py
pmandiola: Because GeoDataset intersections are done at the file level and not feature level, this shouldn't be a problem for point GeoDatasets. Unless there's only one point in a file.

I made this change because I wanted to check that resulting splits didn't intersect, and when joining them I should get the original bbox. At least in my mind two contiguous bboxes aren't intersecting. But we could also solve this by adding a small gap like we did with the time split.

torchgeo/datasets/splits.py
pmandiola: True, I don't love it either but couldn't find another way.

torchgeo/datasets/geo.py
adamjstewart: GeoDataset intersections are done at the bbox level. Our raster and vector datasets may be at the file level, but these point datasets are at the point level. The entire dataset consists of a single CSV file containing lat/long coords, and each coord is added as a separate element to the R-tree.

I agree that contiguous bboxes shouldn't be intersecting, but that's how rtree defines intersection. In the past, the way we've solved this is for the sampler to ignore patches with 0 area. Does that not work here?

pmandiola: Oh I get it, so it wouldn't work in that case. I was removing the change and modifying the non-overlapping test to check for total area == 0 instead of len == 0, but I'm getting a weird R-Tree error. Probably this happened in the first place and that's why I made this change but I forgot.

When attempting this in the tests:
```
# No overlap
assert isclose(_get_total_area(train_ds & val_ds), 0)
assert isclose(_get_total_area(val_ds & test_ds), 0)
assert isclose(_get_total_area(test_ds & train_ds), 0)
```

I get this exception from RTree, only in the third assert:
```
rtree.exceptions.RTreeError: Coordinates must not have minimums more than maximums
```
The first 2 asserts are contiguous bboxes, so the index is not empty but the area is cero. In the third case the index is empty but for some reason the bounds of the index are invalid. Or that is what is seems to me at least.

torchgeo/datasets/splits.py
pmandiola: I followed what is done in torch's random_split.

tests/datasets/test_splits.py
pmandiola: Thanks!

torchgeo/datasets/geo.py
adamjstewart: > the index is empty but for some reason the bounds of the index are invalid

This is a known bug in rtree: https://github.com/Toblerity/rtree/issues/204

Does it help to first check if len > 0 and then check if area > 0? Or does that have the same issue?

pmandiola: That did it, thanks! I'll rollback this change then.

torchgeo/datasets/splits.py
pmandiola: For len(fractions) > 2 it first splits the bbox using the first fraction, where one resulting bbox's area will correspond to that fraction and the second will have to be splitted again. So we now split that second bbox using the second fraction, adjusted to the remaining area. And we keep repeating until done.

Not sure if it's clear enough.

adamjstewart: Might be useful to add more comments to the code. Some of the logic is difficult to follow, not because it's overly complicated but because there's no easier way to do it.

torchgeo/datasets/splits.py
pmandiola: I'll change it to microsecond for consistency

torchgeo/datasets/splits.py
pmandiola: Yeah all the `ignore[arg-type]` are because we are supporting different types in lengths, which are used in different ways. I'm not sure how to handle it properly for mypy, specially in time_series_split were there could be numbers or tuples.

I prefer having only one input though.

torchgeo/datasets/splits.py
pmandiola: Yes it's done randomly. Didn't add it as a parameter but it could be done.

torchgeo/datasets/splits.py
pmandiola: I managed to get rid of most ignores using mypy's cast function. Only one remained in line 316 as mypy is still complaining about types there for some reason.

pmandiola:
Thanks @adamjstewart and @nilsleh for your reviews! I think I addressed most of your comments but happy to change anything else if needed.

adamjstewart (approved):

nilsleh (approved):

adamjstewart:
Thanks again for the hard work on this! Sorry it took so long to review!

pmandiola:
No worries, happy to help!

adamjstewart:
Some of our users have commented that the documentation isn't clear enough about the difference between some of these splitting utilities. Also that we don't provide recommendations of which splitting utility to use for which situation.

@calebrob6 I know you're passionate about this. Any interest/time in making diagrams to explain the differences between these functions and adding some text to explain how to choose which one is right for you?

PR #870, opened 2022-10-26
Add transforms to UnionDataset and IntersectionDataset

### LINKED ISSUE #867: Can't add transforms to UnionDataset or IntersectionDataset
### Summary

Both UnionDataset and IntersectionDataset don't have a way to add transforms neither call `self.transforms` in their `__get_item__` method. The transforms from the merged or intersected datasets are called in their own `__get_item__` methods, but there are situations where having it on the UnionDataset or IntersectionDataset is more convenient and useful.

### Rationale

In some situations having transforms on each merged or intersected dataset might be enough, but here's a couple of cases where this is needed:

* Using the IntersectionDataset to add a mask to an image. For the transforms where information information from both mask and image are needed (i.e. filter mask where image has no data).
* Using UnionDataset to merge multiple RasterDatasets (or IntersectionDatasets) and add only one transforms to the whole dataset instead of on each one.

### Implementation

One possible way would be to just add the call to self.transforms in UnionDataset and IntersectionDataset:

```
```

And let people assign a transform after doing the intersection/union by doing `my_dataset.transforms = my_transforms`. Or adding an `add_transforms` method to IntersectionDataset and UnionDataset to make it more explicit.

### Alternatives

One alternative is to add a transforms parameter to IntersectionDataset and UnionDataset, and also call it on `__get_item__`. But this forces people to use the constructor directly instead of doing & or |.

Another option is to create a TransformableDataset that takes a dataset and transforms as parameters. Calls the underlying datasets `__get_item__` and then transforms in its `__get_item__`.

### Additional information

_No response_

pmandiola:
This PR adds the feature described in #867

Closes #867

### DIFF
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -1020,12 +1020,14 @@ class UnionDataset(GeoDataset):
         samples = []
         for ds in self.datasets:
             if list(ds.index.intersection(tuple(query))):
-                sample = ds[query]
-                if self.transforms is not None:
-                    sample = self.transforms(sample)
-                samples.append(sample)
+                samples.append(ds[query])
 
-        return self.collate_fn(samples)
+        sample = self.collate_fn(samples)
+
+        if self.transforms is not None:
+            sample = self.transforms(sample)
+
+        return sample
 
     def __str__(self) -> str:
         """Return the informal string representation of the object.
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -822,6 +822,7 @@ class IntersectionDataset(GeoDataset):
         collate_fn: Callable[
             [Sequence[Dict[str, Any]]], Dict[str, Any]
         ] = concat_samples,
+        transforms: Optional[Callable[[Dict[str, Any]], Dict[str, Any]]] = None,
     ) -> None:
         """Initialize a new Dataset instance.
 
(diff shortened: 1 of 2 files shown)


torchgeo/datasets/geo.py
```diff
@@ -1003,7 +1020,10 @@ def __getitem__(self, query: BoundingBox) -> Dict[str, Any]:
         samples = []
         for ds in self.datasets:
             if list(ds.index.intersection(tuple(query))):
-                samples.append(ds[query])
+                sample = ds[query]
+                if self.transforms is not None:
+                    sample = self.transforms(sample)
+                samples.append(sample)
```
adamjstewart, line 1026: Why does the transform apply to the collated sample in IntersectionDataset but to the uncollated sample in UnionDataset?

adamjstewart (commented):
Tests are failing because we need to add a unit test to hit this new code.

torchgeo/datasets/geo.py
pmandiola: I had both applied to the collated sample first but mypy gave me a typing error so I changed it to this. I tried again to show you the error and it worked so I was probably doing something wrong in the first place. I'll push the change.

adamjstewart (approved):

calebrob6:
Thanks @pmandiola!

Issue #855, opened 2022-10-16
Random sized patches support for other non-geospatial datamodules


nilsleh:
### Summary

In #851 it was suggested that datamodules with variable sized input images should implement a random size crop  from tile to patches during training and have a keyword argument to specify the desired `patch_size`.

When looking through our other Non-geospatial datamodules, there are other cases that might be good to add the same support, just because the default images are very large and not suitable for batches even though all images have the same dimensions:

- [x] GID-15, defaullt image size 6800x7200 (#928)
- [x] Inria Aerial Image Labeling, default image size 5000x5000
- [x] DeepGlobeLandCoverDataModule, default image size is 2,448x2,448 (#876)
- [x] Potsdam, default image size is 6000x6000 (#929)

Maybe, there can also be a threshold for future datamodule additions: for example if image size is greater than X, implement the random patch crop logic for that datamodule.

### Rationale

I think it would lower the threshold of using these datasets. Since datamodules are supposed to make it very quick and easy for users to conduct segmentation experiments, the patch_size logic would make it a bit more easier to practically use the datamodules right away.

### Implementation

Implement a logic like in the `OSCDDatamodule.`

ashnair1:
The `InriaAerialImageLabelling` datamodule already supports extraction of random crops.

Issue #516, opened 2022-04-26
Glossary: "tile" term is prone to confusion


remtav:
The [glossary](https://github.com/microsoft/torchgeo/blob/main/docs/user/glossary.rst) you've put up is very useful to try and set a standard for geospatial deep-learning related terms. However in the current version, the term "tile" is prone to confusion.

In torchgeo's glossary:
[tile](https://torchgeo.readthedocs.io/en/latest/user/glossary.html#term-tile) = A single image file taken by a remote sensor like a satellite.
[tiling](https://torchgeo.readthedocs.io/en/latest/user/glossary.html#term-tiling) = Splitting a [tile](https://torchgeo.readthedocs.io/en/latest/user/glossary.html#term-tile) into [patches](https://torchgeo.readthedocs.io/en/latest/user/glossary.html#term-patch). This is the reverse operation of [stitching](https://torchgeo.readthedocs.io/en/latest/user/glossary.html#term-stitching).
Therefore, "tiling" isn't "the process of creating tiles from imagery" but, rather "the process of creating chips/patches from tiles", which is confusing. For example, you'd think that a "clipping" process *creates* "clips", rather than cut up a "clip" (entire image) into smaller chips/patches/pieces.

In the [Cloud Optimized Geotiff spec](https://github.com/cogeotiff/cog-spec/blob/master/spec.md), the term "tile" refers to the image internal blocks (e.g. 512 x 512 px at various scales) which allow to compartmentalize the image to allow range requests.  Consequently, a COG "tile" seems to be more similar to the term "patch" or "chip" from Torchgeo's standpoint.

I'd simply suggest using "scene" or "image" instead of "tile". "tile" could become a synonym of "chip" or "patch".,

The term "tiling", however, seems to be used by other libraries such as [solaris](https://solaris.readthedocs.io/en/latest/tutorials/notebooks/api_tiling_tutorial.html). This term could remain in my opinion.

As a final note, there seems to be a need for a EO-DL synset as a way to reference all these specific terms and synonyms. Your glossary is a good starting point :)

adamjstewart:
Yes, completely agree, the current terminology is quite confusing. I also see "tile" used to refer to WRS row/path "tiles". I'm a bit afraid of trying to standardize our own terminology if the remote sensing community hasn't yet standardized their own terminology. 

[...]

PR #876, opened 2022-10-29
Add random crop logic to DeepGlobeLandCover Datamodule

nilsleh:
Adressing #855 by adding random crop logic to datamodule. Additionally, add a test file for the datamodule.


torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -23,26 +29,45 @@ class DeepGlobeLandCoverDataModule(pl.LightningDataModule):
...
+        """Initialize a LightningDataModule for DeepGlobeLandCover based DataLoaders.
 
         Args:
-            batch_size: The batch size to use in all created DataLoaders
+            train_batch_size: The batch size used in the train DataLoader
+                (val_batch_size == test_batch_size == 1). The effective batch size
+                will be 'train_batch_size' * 'num_patches_per_tile'
             num_workers: The number of workers to use in all created DataLoaders
             val_split_pct: What percentage of the dataset to use as a validation set
+            patch_size: Size of random patch from image and mask (height, width), should
```
calebrob6, line 47: "a multiple of 32" is a much better way of stating that "the numbers should be able to be evenly divisible by 2 5 times in a row" (i.e. what I usually say)

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -23,26 +29,45 @@ class DeepGlobeLandCoverDataModule(pl.LightningDataModule):
...
+        num_patches_per_tile: int = 32,
         val_split_pct: float = 0.2,
         **kwargs: Any,
     ) -> None:
-        """Initialize a LightningDataModule for DeepGlobe Land Cover based DataLoaders.
+        """Initialize a LightningDataModule for DeepGlobeLandCover based DataLoaders.
 
         Args:
-            batch_size: The batch size to use in all created DataLoaders
+            train_batch_size: The batch size used in the train DataLoader
```
calebrob6, line 42: I think this should be named "number of tiles per batch", if I set "train_batch_size" to be 32, I will expect the batch size to be 32, not 32*`num_patches_per_tile`

calebrob6 (dismissed):

calebrob6:
I confirmed this all works as expected with the actual dataset!

torchgeo/datamodules/deepglobelandcover.py
calebrob6: @nilsleh, thoughts on this?

nilsleh: If you rename it to `num_tiles_per_batch` , do you intend that the effective batch_size would be `num_patches_per_tile` * `num_tiles_per_batch`? Then we can reverse the doc string a bit to explain that.

calebrob6: Actually, I think setting `batch_size` and `num_tiles_per_batch` explicitly (as users might expect to be able to set a batch size) then calculating `num_patches_per_tiles = batch_size // num_tiles_per_batch` makes the most sense.

nilsleh: That makes sense. However, then the `batch_size` you are setting is also not guaranteed to end up the same as what you specify, if you have `batch_size=16` and `num_tiles_per_batch=3` for example, then you would end up with a batch_size of just 15 at the end, so maybe suggest that it should be evenly divisible in the docstring?

calebrob6: Very good points (thank you for taking the time to think through this)!

I think suggesting that `num_tiles_per_batch` should evenly divide `batch_size` (and warning if not) is a reasonable way to go so that we can keep `batch_size` as an argument.

> Also you would have to check or warn that a tile actually has enough patches. For example, if batch_size=32 and num_tiles_per_batch=1 ...

I'm not too worried about this as we can set reasonable defaults. Anyone who changes them should be looking at tile sizes anyway. (e.g. as a user, I would only imagine setting `num_tiles_per_batch=1` if I was trying something particularly crazy, because generally you would want batches with diverse sets of patches in them).

nilsleh: Yes, that is reasonable, I made a commit accordingly with a "draft" for a warning message.

nilsleh:
I noticed something after the last commit, namely the case when `val_split_pct=0`.  Because then the `val_dataset` is assigned the training dataset with its transforms (importantly the random crop tile logic), however, the `val_dataloader` does not receive the collate_fn like the `train_dataloader` to combine the patches into a batch. I suppose there are two cases:
1. Also pass the collate_fn to `val_dataloader` if `val_split_pct=0` (included in the last commit)
2. Or when `val_split_pct=0` then `train_dataset` is assigned to `val_dataset` but with the validation transforms that do not contain the crop logic. 

Just wanted to ask, which logic we should go for and whether we can do that consistently across all the datamodules.

torchgeo/datamodules/deepglobelandcover.py
adamjstewart: Good discussion, my only comment is that we should be careful of the use of the word "tile" since it can mean different things to different people: #516. Maybe "patch" or "chip" instead?

nilsleh: in this case the name `num_tiles_per_batch` is quiet accurate, right? Because it specifies from how many tiles/scenes in the dataset to take random cropped patches from to overall yield an overall `batch_size`. Or based on #516, would `num_scenes_per_batch` be better?

adamjstewart: What I'm saying is that "tile" and "scene" mean different things to different people, so we should avoid using them. I think `num_patches_per_batch` or `num_chips_per_batch` may be better.

adamjstewart: Oh wait, it's not patches/chips, it's files. We may want to make a decision on #516 before we decide on a name for this one.

calebrob6: I think "tiles" is fine -- we can change later if needed but I don't want to hold up this

calebrob6:
What is the minimum test complaining about?

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -23,26 +29,62 @@ class DeepGlobeLandCoverDataModule(pl.LightningDataModule):
...
+            self.batch_size >= num_tiles_per_batch
+        ), "num_tiles_per_batch should be less than or equal to batch_size."
+
+        self.num_patches_per_tile = self.batch_size // num_tiles_per_batch
+        self.num_tiles_per_batch = num_tiles_per_batch
+
+        if (self.num_patches_per_tile % 2) != 0 and (
+            self.num_patches_per_tile != num_tiles_per_batch
+        ):
+            warnings.warn(
```
calebrob6, line 76: Makes sense to me!

adamjstewart:
> What is the minimum test complaining about?

Same thing I fixed [here](https://github.com/microsoft/torchgeo/pull/853/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711). Either merge that PR first or make the same change to this PR.

torchgeo/datasets/utils.py
```diff
@@ -216,6 +218,22 @@ def download_radiant_mlhub_collection(
...
 
+def collate_patches_per_tile(batch: List[Dict[str, Any]]) -> Dict[str, Any]:
+    """Define collate function to combine patches per tile and batch size.
+
+    Args:
+        batch: sample batch from dataloader containing image and mask
+
+    Returns:
+        sample batch where the batch dimension is
+        'train_batch_size' * 'num_patches_per_tile'
```
adamjstewart, line 229: Needs a `.. versionadded`

torchgeo/datasets/utils.py
```diff
@@ -216,6 +218,22 @@ def download_radiant_mlhub_collection(
     collection.download(output_dir=download_root, api_key=api_key)
 
 
+def collate_patches_per_tile(batch: List[Dict[str, Any]]) -> Dict[str, Any]:
```
adamjstewart, line 221: Can we move this below and name it to match our other collation functions (stack/merge/unbind samples)? Should also export it in `__init__.py` like we do for the other ones.

tests/datamodules/test_deepglobelandcover.py
```diff
@@ -0,0 +1,56 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
adamjstewart, line 1: Is this necessary? I've been trying to remove all of `tests/datamodules` since they aren't actually useful tests. Instead, I want to test all datamodules with real trainers in `tests/trainers`.

nilsleh: If I check codecov with pytest and remove this datamodule test then the warning we introduced about the batch size is not checked at the moment, otherwise coverage is 100% with only trainer test. So should I add a test for this warning in the trainer and remove this test file?

adamjstewart: To get coverage of this line, you can either

1. use a batch size that is not a multiple so that the warning is hit, or
2. create two config files (one multiple, one not) and test both.

I would personally prefer the second solution, although the first is faster to test.

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -23,26 +28,62 @@ class DeepGlobeLandCoverDataModule(pl.LightningDataModule):
...
             num_workers: The number of workers to use in all created DataLoaders
             val_split_pct: What percentage of the dataset to use as a validation set
+            patch_size: Size of random patch from image and mask (height, width), should
+                be a multiple of 32 for most segmentation architectures
+            num_tiles_per_batch: number of random tiles to consider sampling patches
+                from per sample, should evenly divide batch_size and be less than
+                or equal to batch_size
             **kwargs: Additional keyword arguments passed to
                 :class:`~torchgeo.datasets.DeepGlobeLandCover`
+
```
adamjstewart, line 52: Can you add:
```rst
Raises:
    AssertionError: if num_tiles_per_batch > batch_size
```

torchgeo/datasets/__init__.py
```diff
@@ -101,6 +101,7 @@
     BoundingBox,
     concat_samples,
     merge_samples,
+    rearrange_patches_to_sample,
     stack_samples,
     unbind_samples,
```
adamjstewart, line 106: One of these is not like the other...

Trying to think of a more consistent naming scheme here. What about `rearrange_samples`? Would that be too generic?

torchgeo/datasets/utils.py
```diff
@@ -598,6 +600,25 @@ def unbind_samples(sample: Dict[Any, Sequence[Any]]) -> List[Dict[Any, Any]]:
...
+    Returns:
+        sample batch where the batch dimension is
+        'batch_size' * 'num_patches_per_tile'
+
+    .. versionadded: 0.4
+    """
+    r_batch: Dict[str, Any] = default_collate(batch)  # type: ignore[no-untyped-call]
+    if "mask" in r_batch:
+        r_batch["mask"] = rearrange(r_batch["mask"], "b t h w -> (b t) h w")
+    r_batch["image"] = rearrange(r_batch["image"], "b t c h w -> (b t) c h w")
```
adamjstewart, line 618: Should also check if "image" is in dict. Actually, could we just loop over all tensors in the dict and merge the first two dims? Then we don't have to duplicate code or worry about names.

adamjstewart: Another option (in terms of both code and function name) would be `torch.flatten(value, end_dim=1)`

torchgeo/datasets/__init__.py
ashnair1: Just wanted to mention that in other related PRs (Potsdam and GID-15), this function was named as `collate_patches_per_tile`. Perhaps that would be a better fit?

torchgeo/datasets/utils.py
ashnair1: > Should also check if "image" is in dict. Actually, could we just loop over all tensors in the dict and merge the first two dims? Then we don't have to duplicate code or worry about names.

Is there a scenario where we use this particular function and "image" is not present in the dict? 


Regarding the code. I like the current einops implementation as it's more clear to the user the what the operation does to the image (5D -> 4D) and mask tensors (4D -> 3D). Even if they're undergoing the same op i.e. merging first two dims.

torchgeo/datasets/__init__.py
adamjstewart: The "one of these is not like the other" statement would still apply. We should either 1) change all of our existing collation functions to match this new naming style, or 2) change this function to match our existing collation function naming style.

tests/trainers/test_segmentation.py
```diff
@@ -38,6 +38,7 @@ class TestSemanticSegmentationTask:
             ("chesapeake_cvpr_5", ChesapeakeCVPRDataModule),
             ("deepglobelandcover_0", DeepGlobeLandCoverDataModule),
             ("deepglobelandcover_5", DeepGlobeLandCoverDataModule),
+            # ("deepglobelandcover_batch_size", DeepGlobeLandCoverDataModule),
```
adamjstewart, line 41: Should this be removed?

tests/trainers/test_segmentation.py
```diff
@@ -146,3 +147,26 @@ def test_missing_attributes(
         model = SemanticSegmentationTask(**model_kwargs)
         trainer = Trainer(fast_dev_run=True, log_every_n_steps=1, max_epochs=1)
         trainer.validate(model=model, datamodule=datamodule)
+
+    @pytest.mark.parametrize(
+        "name,classname",
+        [("deepglobelandcover_batch_size", DeepGlobeLandCoverDataModule)],
+    )
```
adamjstewart, line 154: Don't need to parametrize this if there's only one, but I think your other PRs will do the same thing so you'll just keep appending to this list, right?

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -63,26 +107,76 @@ def setup(self, stage: Optional[str] = None) -> None:
...
+            Args:
+                sample: contains image and mask sample from dataset
+
+            Returns:
+                padded image and mask
+            """
+            h, w = sample["image"].shape[1], sample["image"].shape[2]
+            new_h = int(32 * ((h // 32) + 1))
+            new_w = int(32 * ((w // 32) + 1))
```
adamjstewart, line 146: Need to review this in more detail to make sure this doesn't introduce any rounding issues like we saw in previous releases.

torchgeo/datasets/utils.py
nilsleh: I think I also find the more explicit einops operation better because it would be more clearer if someone looks at what `rearrange_samples` does in the source code.

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -63,26 +107,76 @@ def setup(self, stage: Optional[str] = None) -> None:
...
+            for i in range(self.num_patches_per_tile):
+                image, mask = self.rcrop(sample["image"], sample["mask"].float())
+                images.append(image.squeeze(0))
+                masks.append(mask.squeeze().long())
+
+            sample["image"] = torch.stack(images)
+            sample["mask"] = torch.stack(masks)
+            return sample
+
+        def pad_to(sample: Dict[str, Any]) -> Dict[str, Any]:
```
adamjstewart, line 135: Presumably these are going to be used in multiple datamodules? Can we move these to `torchgeo/datamodules/utils.py`? Or maybe they should go in `torchgeo/transforms/*`? I'm trying to avoid adding transforms that aren't specific to geospatial data, but I'm not sure if there's anything like this in other libraries to avoid the duplication.

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -63,26 +107,76 @@ def setup(self, stage: Optional[str] = None) -> None:
...
+                padded image and mask
+            """
+            h, w = sample["image"].shape[1], sample["image"].shape[2]
+            new_h = int(32 * ((h // 32) + 1))
+            new_w = int(32 * ((w // 32) + 1))
+
+            padto = K.PadTo((new_h, new_w))
+
+            sample["image"] = padto(sample["image"])[0]
+            sample["mask"] = padto(sample["mask"].float()).long()[0, 0]
```
adamjstewart, line 151: Why cast to float then long? What's the original type?

tests/trainers/test_segmentation.py
nilsleh: yes, I will append new configs for the other PRs

torchgeo/datasets/utils.py
adamjstewart: > Is there a scenario where we use this particular function and "image" is not present in the dict?

Most GeoDatasets contain either an image or a mask but not both. So any time someone used this collation function for a mask-only dataset this would break.

adamjstewart: The problem with this is that it makes assumptions about the keys of the dict. None of our other collation functions do this, and I don't think it's necessary or wise to do this. For example, what if someone wants to use this with 6D images or 5D masks?

torchgeo/datasets/utils.py
```diff
@@ -598,6 +600,25 @@ def unbind_samples(sample: Dict[Any, Sequence[Any]]) -> List[Dict[Any, Any]]:
...
 
+def rearrange_samples(batch: List[Dict[str, Any]]) -> Dict[str, Any]:
+    """Define collate function to combine patches per tile and batch size.
+
+    Args:
+        batch: sample batch from dataloader containing image and mask
+
+    Returns:
+        sample batch where the batch dimension is
+        'batch_size' * 'num_patches_per_tile'
```
adamjstewart, line 611: This function does not "define a collate function", it _is_ a collate function. Also, rearrange is a bit ambiguous. Maybe something like:
```suggestion
def flatten_samples(samples: Iterable[Dict[str, Any]]) -> Dict[str, Any]:
    """Flatten the first two dimensions of a list of samples.
    
    Useful for ...
    
    Args:
        samples: list of samples
        
    Returns:
        a single sample
```

torchgeo/datamodules/deepglobelandcover.py
nilsleh: Yes, they will be used in all datamodules, same with the `pad_to` function. It feels like it is more of an util than a transform, because `self.rcop` is a Kornia sequential augmentation it is just that we want to apply it `n`times to a single tile.

torchgeo/datasets/utils.py
```diff
@@ -598,6 +600,25 @@ def unbind_samples(sample: Dict[Any, Sequence[Any]]) -> List[Dict[Any, Any]]:
...
+
+    .. versionadded: 0.4
+    """
+    r_batch: Dict[str, Any] = default_collate(batch)  # type: ignore[no-untyped-call]
```
adamjstewart, line 615: I would call this variable collated to match the other functions

torchgeo/datasets/utils.py
adamjstewart: For example, additional axis could be time for time-series data, or z for x,y,z climate/weather data.

torchgeo/datamodules/deepglobelandcover.py
adamjstewart: Alright, let's move it to utils then

torchgeo/datamodules/deepglobelandcover.py
nilsleh: It is 'long()' before but `Kornia.PadTo` expects a float. And we need `long` again because it is the mask.

torchgeo/datasets/utils.py
nilsleh: Sounds good. Does that mean you also want to do `torch.flatten()` instead of the einops operation?

torchgeo/datamodules/deepglobelandcover.py
adamjstewart: Gotcha. Can you add a comment explaining this?

torchgeo/datasets/utils.py
adamjstewart: Yes, for the reasons mentioned above. Explicit is good, but making assumptions about the names of keys and dimensions of keys is not.

torchgeo/datamodules/deepglobelandcover.py
nilsleh: Do you have a preferred name, for the `pad_to` function when it is in utils. Currently, it is explicitly used for segmentation samples where architectures rely on the dimensions being a multiple of 32. Would you want it to be a bit more universal where it accepts a  `multiple_of` argument to specify 32 or any other integer?

Similar question about the naming the `n_random_crop` in utils.

adamjstewart: No preference. I'm fine with keeping it specific to 32 until we find a case where we need some other multiple. For now, let's name it starting with an underscore and not export it in `__init__.py`. That way we reserve the right to rename or change it later without changing backwards compatibility.

nilsleh: For the `n_random_crop` function I am a bit stuck. We use this function in `torch.Compose` which only accepts a single argument, but we need to specify both the number of patches to take and the Kornia augmentation operation. Would I have to write a small class that takes `kornia_augmen` and `num_patches` as init argument and then during the forward just accepts a sample and transforms it?

adamjstewart: Yes, it would be a class just like all transforms

nilsleh: actually, I just found that the `AugmentationSequential` module in transforms.py is what we need, except that we want to apply it to generate `n` patches, so should I adapt that class to accept a 'num_patches_per_tile' argument?

adamjstewart: `AugmentationSequential` looks identical to `Compose` other than handling of dicts, I think you still need to pass a list of classes to it just like `Compose`.

adamjstewart: @isaaccorley wrote that, maybe he can comment.

nilsleh: my idea was to define `AugmentationSequential` and then pass that into the Compose, but it doesn't seem to work, so I will define a new torch transform class.

nilsleh: I think I found a way now, not sure if you will like it (will push). I wanted to test it on the actual DeepGlobe dataset and found that the link and directory structure is outdated. Should I do a separate PR for that?

nilsleh:
Not sure what you think of the transform implementation I did. It is very similar to the existing `AugmentationSequential` Transform and I think in the future we could consolidate that in a new PR. Trying to stick the existing `AugmentationSequential` into a `Compose` doesn't seem to work, but in a sense this new version is more general.

torchgeo/datamodules/deepglobelandcover.py
adamjstewart: Yeah, prob separate PR

adamjstewart (changes requested):
I don't think this approach makes sense. Or rather, it's a perfectly valid approach, but it's not the way torchvision/torchgeo/kornia do things, and I don't see any reason why we need to introduce a completely different way of handling this. The way I think about it is:

1. `_pad_segmentation_sample` is a transform/data augmentation
2. All transforms/data augmentations in torchvision/torchgeo/kornia are classes that implement `__call__`, not functions
3. Kornia's AugmentationSequential/AugmentationPatches and torchvision's Compose are containers, not transforms/data augmentations

There are also a lot of inconsistencies in the API that this introduces:

1. All torchgeo transforms are in `torchgeo/transforms`, not `torchgeo/datasets`
2. Kornia: AugmentationSequential/AugmentationPatches, TorchGeo: AugmentationSequential/PatchesAugmentation (names are not consistent)
3. Kornia: AugmentationPatches splits into patches, performs augmentation, reassembles. TorchGeo: PatchesAugmentation applies augmentations to all patches (behavior is not consistent)

I still think we should use a transform instead of a function. Then you can use it with Compose or AugmentationSequential or whatever else. I would subclass `nn.Module` like all of our other existing transforms.

I also think this probably belongs in `torchgeo/transforms/`, although we can still keep it a hidden method until we have a better idea of where to fit it in the library. We can put it in `transforms.py` for now.

nilsleh:
> 

Okay so I should 
1. create one transform class for `_pad_segmentation_sample` and put it in `transforms.py` instead of `utils.py`?
2. regarding what I currently named `PatchesAugmentation` in `transforms.py`, it is intended as a torchvision transform that uses Kornia because it handles dicts but is used again in torch.Compose because for the dataset processing that also includes the `preprocess` function. I am not sure what you are recommending here? I am not aiming to copy the Kornia `AugmentationPatches` here, rather just one transformation for the segmentation datamodules.

adamjstewart:
I think we want to move

FROM: function `_pad_segmentation_sequential` in `torchgeo/datasets/utils.py`
TO: class `PadSegmentationSequential` in `torchgeo/transforms/transforms.py`

We can bikeshed on the name later, but I'll prob want to rename it to match whatever torchvision/kornia calls it.

I also think we want to remove `PatchesAugmentation` entirely and use either Compose or AugmentationSequential like we do with all of our other transforms.

I'm starting to suspect that there's something I'm not understanding about why `PatchesAugmentation` is even required, or why we can't use torchvision's `Pad` or Kornia's `PadTo` transforms. Might be useful to clear up that misunderstanding first, maybe none of this is required.

nilsleh:
Okay I am onboard with everything you said about `_pad_segmentation_sequential`. 

From what I can tell trying things out locally here, the current torchgeo `AugmentationSequential` cannot be used because
a) it does not support generating a specified number of patches from a single tile
b) it cannot be used within torch.Compose(), which the datamodule currently uses to concatenate the steps of the `preprocess` class function and the "random cropping patches from tile" operation we desire

In contrast the Kornia Augmentation possibilities are useful because they can operate on dictionaries and simultaneously on "image" and "mask", but would still have to concatenate it with the dataset preprocess steps. So I thought this is why it is useful to have a "PatchAugmentation" Transformation that is implemented like a torchvision transformation and can hence be used in torch.compose() to follow the `preprocess` function but internally uses the Kornia augmentations to operate on the dictionaries we use in torchgeo.

adamjstewart:
Okay, so I think the way `Compose` and `AugmentationSequential` are intended to be used is as follows:
```python
Compose(  # container
    transform1,  # transform
    transform2,  # transform
    ...
)
AugmentationSequential(  # container
    transform1,  # transform
    transform2,  # transform
    ...
)
```
Right now, you're trying to use them as:
```python
train_transforms = Compose(  # container
    preprocess,  # function
    PatchesAugmentation(  # container
        AugmentationSequential(  # container
            RandomCrop  # transform
        )
    )
)
test_transforms = Compose(  # container
    preprocess,  # function
    _pad_segmentation_sample,  # function
)
```
What if we turn preprocess and _pad_segmentation_sample into transforms and then use:
```python
train_transforms = Compose(  # container
    Normalize,  # transform
    RandomCrop,  # transform
)
test_transforms = Compose(  # container
    Normalize,  # transform
    PadSegmentationSample,  # transform
)
```
You might need to replace `Compose` with `AugmentationSequential` here, I'm not completely sure what the difference is. But I think simplifying these would help greatly.

@isaaccorley maybe you can comment, you have a lot more experience with Kornia than I do.

adamjstewart:
#965 also has a good example of using `AugmentationSequential` to string together multiple transforms, I'm not sure why that approach wouldn't work here.

nilsleh:
> #965 also has a good example of using `AugmentationSequential` to string together multiple transforms, I'm not sure why that approach wouldn't work here.

I think the difference here is that we need to create multiple random crops from a single tile, while in that PR it can just operate on tiles directly, so you don't have the changes in dimensions. 

Making a Normalize transform makes sense and the "RandomCrop" transform you write would do the "cropping multiple patches from a tile" operation? My thought with the double container:
```
train_transforms = Compose(  # container
    preprocess,  # function
    PatchesAugmentation(  # torchgeo container
        K.AugmentationSequential(  # kornia container
            RandomCrop  # transform
        )
    )
)
```
was that if you are creating those n random patches from a single tile, you might want to do additional augmentations like flipping, noise etc on the single patches independently, so each `K.AugmentationSequential` would operate on a patch independently and not the same augmentation chain on a batch.

isaaccorley:
I'm still trying to catch up on the discussions here, but to be clear, the **only** reason for creating a custom `AugmentationSequential` class was to simplify kornia's `AugmentationSequential` class to work with our sample dicts. It's not meant as a template.

We should use the code similarly to what's in `torchgeo.datamodules.OSCDDataModule`

nilsleh:
Isaacs comment and message clarified a lot and I see why I overcomplicated things, apologies for that. Working on a new version now that is simpler.

nilsleh:
the old "transforms" are not "preprocessing" steps to format the data, and then Issac suggested that any additional augmentations like flips etc can be done with a separate `K.AugmentationSequential` in an `on_after_batch_transfer` like the [RESISC45 datamodule](https://github.com/microsoft/torchgeo/blob/43356ee2012a29669cf2f8db6bc3ad920f6f0997/torchgeo/datamodules/resisc45.py#L49). 

I hope what I pushed now is better, but probably not done, so will incorporate any feedback again tomorrow.

torchgeo/datasets/utils.py
```diff
@@ -598,6 +599,29 @@ def unbind_samples(sample: Dict[Any, Sequence[Any]]) -> List[Dict[Any, Any]]:
...
+    .. versionadded: 0.4
+    """
+    collated: Dict[str, Any] = default_collate(samples)  # type: ignore[no-untyped-call]
+    flattened_sample: Dict[str, Any] = {}
+    for key, val in collated.items():
+        if isinstance(val, Tensor):
+            flattened_sample[key] = torch.flatten(val, end_dim=1)
+        else:
+            flattened_sample[key] = val
+    return flattened_sample
```
adamjstewart, line 622:
```suggestion
    for key, val in collated.items():
        if isinstance(val, Tensor):
            collated[key] = torch.flatten(val, end_dim=1)
    return collated
```

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -23,38 +27,74 @@ class DeepGlobeLandCoverDataModule(pl.LightningDataModule):
...
+                " num_tiles_per_batch evenly divides batch_size"
+            )
 
-        Returns:
-            preprocessed sample
-        """
-        sample["image"] = sample["image"].float()
-        sample["image"] /= 255.0
-        return sample
+        self.train_preprocess = Compose(
```
adamjstewart, line 85:
```suggestion
        self.train_transforms = Compose(
```

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -23,38 +27,74 @@ class DeepGlobeLandCoverDataModule(pl.LightningDataModule):
...
+            [
+                ConstantNormalize(255.0),
+                RepeatedRandomCrop(
+                    self.patch_size,
+                    self.num_patches_per_tile,
+                    data_keys=["input", "mask"],
+                ),
+            ]
+        )
+        self.test_preprocess = Compose(
```
adamjstewart, line 95:
```suggestion
        self.test_transforms = Compose(
```

torchgeo/datasets/utils.py
```diff
@@ -598,6 +599,29 @@ def unbind_samples(sample: Dict[Any, Sequence[Any]]) -> List[Dict[Any, Any]]:
...
+
+    Useful for combining several patches per sample into a batch of samples.
+
+    Args:
+        samples: list of samples
+
+    Returns:
+        a single sample
+
+    .. versionadded: 0.4
```
adamjstewart, line 613:
```suggestion
    .. versionadded:: 0.4
```

adamjstewart (commented):
This is looking much closer now! I'm still a bit unsure if we really need custom transforms or not.

torchgeo/transforms/transforms.py
```diff
@@ -73,3 +73,129 @@ def forward(self, sample: Dict[str, Tensor]) -> Dict[str, Tensor]:
             sample["boxes"] = sample["boxes"].to(boxes_dtype)
 
         return sample
+
+
+class ConstantNormalize(Module):
```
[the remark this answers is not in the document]
nilsleh: - K.Normalize does (x-mean)/std normalization. I saw that we have a lot of datamodules for non-geodatasets with rgb imagery where we just do scaling to 0-1 by dividing by 255. I guess you could also use K.Normalize(mean=0, std=255) to do the same but I thought a ConstantNormalize is a bit less hacky?
- K.RandomCrop will only crop a single image once, but we want to create n patches from a single image so `RepeatedRandomCrop` is a sort of wrapper around `K.RandomCrop` that does this patchifying 
- from what I understand K.PadTo would have to be initialized with a desired `h,w` once before passing it to the transform, however, we have several datasets where the images are not of constant size but actually vary quiet a bit, so for each sample we have to compute the next available multiple of 32 in this case and pad according to that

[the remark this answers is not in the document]
adamjstewart: * I think `K.Normalize` would be better since we don't need to maintain any additional code in TorchGeo
* torchvision has a 5-crop and 10-crop, I think we could argue for having an N-crop. Would prob be easiest to implement this in TorchGeo for now and try to contribute it to Kornia so that we can remove it someday
* Can we not just choose a consistent size and always pad to that? How widely do these sizes vary?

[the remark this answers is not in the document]
nilsleh: - okay, will change that
- okay, will change the name to N-crop? or would NCrop be different from the current `RepeatedRandomCrop`
- looking at the datasets doc page: vaihingen (1281-3816), SpaceNet (102-900), OSCD (40-1180) are the ones with quiet a bit of variation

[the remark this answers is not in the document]
adamjstewart: * :+1:
* I like `NCrop` but we'll see what the Kornia devs think. Can open a PR simultaneous to this one to start getting feedback
* But we're cropping all of those first, right? Won't the crops have approximately consistent dimensions?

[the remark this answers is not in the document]
nilsleh: - last point: no when you are cropping you are choosing a patch_size that should be a multiple of 32, the dynamic padding is done for eval/test samples that come in the full tile size and are adjusted to fit the model architecture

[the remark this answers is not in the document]
adamjstewart: Maybe we can upstream a PadToMultiple? Not sure if that form of padding would be useful outside of satellite imagery. Curious what @isaaccorley thinks (I think @calebrob6 is officially on vacation now)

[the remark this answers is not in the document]
nilsleh: It seems like a `PadToMultiple` should be useful for segmentation tasks, however, I think most computer vision segmentation datasets are curated so that its often already taken care of.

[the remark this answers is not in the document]
nilsleh: I think there might be another issue:

When using `K.Normalize` I need to wrap it in order for the this single augmentation to work with our dictionaries. If I use the torchgeo,AugmentationSequential:
```
self.test_transform = AugmentationSequential(
            [K.Normalize(mean=0.0, std=255.0), PadToMultiple(32)], data_keys=["image", "mask"]
        )
```
It complains because even though, `PadToMultiple()` is a `nn.Module` but contains a `Kornia.PadTo` inside:
`NotImplementedError: Only nn.Module are supported at this moment. Got [Normalize(Normalize(p=1.0, p_batch=1.0, same_on_batch=True, mean=tensor([0.]), std=tensor([255.]))), PadToMultiple()].`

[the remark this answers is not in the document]
nilsleh: I have also tried the following:
```
self.test_transform = Compose(
            [
                K.AugmentationSequential(
                    K.Normalize(mean=0.0, std=255.0),
                    data_keys=["input", "mask"]
                ), 
                PadToMultiple(32)], 
        )
```
even though, you already mentioned that this is a bad idea. It also doesn't work because Composes passes a dict,
while Kornia internally expects a tuple of the dict values, but instead gets `({"image":.., "key":,,}, )`and then there is an error because `len(args) != len(data_keys)` since `len(args)=1`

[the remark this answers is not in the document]
isaaccorley: For training transforms I don't believe we need to pad since we are random cropping. For test and val sets we do need to do though.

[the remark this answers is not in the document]
nilsleh: new approach to use the `K.Normalize` function with our custom `NCrop` or `PatToMultiple`:
```
self.train_transform = nn.Sequential(
    torchgeo.transforms.AugmentationSequential(
            K.Normalize(mean=0., std=255.), data_keys=["image", "mask"]
    ),
    NCrop(patch_size, self.num_patches_per_tile)
 )
```
This works for setting up the augmentation, however, Kornia always expects floats and currently the `__getitem__` from the deepGlobe dataset returns a `torch.uint8` so should it be changed there?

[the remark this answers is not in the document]
nilsleh: Currently, the last commit includes a version, where the transforms are chained with `torch.nn.Sequential`. The wrapper `torchgeo.transforms.AugmentationSequential` uses `K.AugmentationSequential` internally, and I get errors when trying to use that, becuase internally, the sample dict is not passed from module to module I think but processes the dict items separately as tuples. I am not really familiar with Kornia, so maybe @isaaccorley can comment on that?


```
from torchgeo.transforms import AugmentationSequential, NCrop
import torch

image = torch.rand((3, 256, 256))
mask = torch.randint(high=10, size=(256, 256))
sample = {"image": image, "mask": mask}

# works  (saw this being done in tests/transforms/test_transforms.py
transforms = torch.nn.Sequential(
       AugmentationSequential(
            K.Normalize(mean=0.0, std=255.0), data_keys=["image", "mask"]
        ),
        NCrop(patch_size, self.num_patches_per_tile),
)

# does not work
transforms = AugmentationSequential(
    K.Normalize(mean=0., std=255.), NCrop((32, 32), 3), data_keys=["image", "mask"]
)
```

torchgeo/transforms/transforms.py
```diff
@@ -73,3 +73,100 @@ def forward(self, sample: Dict[str, Tensor]) -> Dict[str, Tensor]:
             sample["boxes"] = sample["boxes"].to(boxes_dtype)
 
         return sample
+
+
+class NCrop(Module):
```
adamjstewart, line 78: Let's keep these private for now (prefix with an underscore, don't include in `__init__`) and try to upstream them to Kornia. Do you want to submit the PR to Kornia or do you want me to?

torchgeo/transforms/transforms.py
```diff
@@ -73,3 +73,100 @@ def forward(self, sample: Dict[str, Tensor]) -> Dict[str, Tensor]:
...
+
+        Args:
+            sample: the input
+
+        Returns:
+            stacked random patches
+        """
+        images, masks = [], []
+        for i in range(self.num_patches_per_tile):
+            crop = K.RandomCrop(self.patch_size, p=1.0)
```
adamjstewart, line 112: This can result in possibly overlapping tiles. Not necessarily a deal breaker, but something that Kornia is likely to point out.

torchgeo/transforms/transforms.py
adamjstewart: How is this different from https://kornia.readthedocs.io/en/latest/contrib.html#kornia.contrib.extract_tensor_patches? We use this in the inria datamodule.

torchgeo/datasets/utils.py
```diff
@@ -598,6 +599,26 @@ def unbind_samples(sample: Dict[Any, Sequence[Any]]) -> List[Dict[Any, Any]]:
     return _dict_list_to_list_dict(sample)
 
 
+def flatten_samples(samples: Iterable[Dict[str, Any]]) -> Dict[str, Any]:
```
adamjstewart, line 602: This looks a lot like `collate_wrapper` in the inria and oscd datamodules. Can you replace those with imports of this?

torchgeo/datamodules/deepglobelandcover.py
adamjstewart: I think I actually want to reverse the decision made here. Allowing users to specify `batch_size` and `num_tiles_per_batch` introduces a lot of error checking (`batch_size` must be > `num_tiles_per_batch`) and adjustments/warning (when `batch_size` is not a multiple of `num_tiles_per_batch`). I prefer @nilsleh's original suggestion of `num_tiles_per_batch` and `num_patches_per_tile`.

torchgeo/transforms/transforms.py
adamjstewart: Answering my own question: NCrop is N random crops (better for testing), ExtractTensorPatches is a regular grid (better for evaluation).

setup.cfg
```diff
@@ -29,8 +29,8 @@ install_requires =
     einops>=0.3,<0.7
     # fiona 1.8+ required for reading empty files
     fiona>=1.8,<2
-    # kornia 0.6.4+ required for kornia.contrib.compute_padding
-    kornia>=0.6.4,<0.7
+    # kornia 0.6.5+ required due to change in kornia.augmentation API
+    kornia>=0.6.5,<0.7
```
adamjstewart, line 33: In Kornia 0.6.5+, all augmentation instance methods have a new `flags` parameter. So the transforms I added won't work with Kornia 0.6.4 and older. Once we upstream these transforms to Kornia, we'll need to depend on an even newer version anyway.

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -3,87 +3,92 @@
...
 """DeepGlobe Land Cover Classification Challenge datamodule."""
 
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Tuple, Union
 
 import matplotlib.pyplot as plt
 import pytorch_lightning as pl
-from torch.utils.data import DataLoader, Dataset
-from torchvision.transforms import Compose
+from kornia.augmentation import Normalize
```
adamjstewart, line 10: I'm planning on removing all torchvision transforms. Torchvision relies on PIL for many of its transforms, which doesn't support MSI. Kornia has all of the same transforms, but they are in pure PyTorch, so they can run on the GPU and support MSI. I don't see a good reason not to only use Kornia transforms.

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -3,87 +3,92 @@
...
+            patch_size: The size of each patch, either ``size`` or ``(height, width)``.
+                Should be a multiple of 32 for most segmentation architectures
+            val_split_pct: The percentage of the dataset to use as a validation set
+            num_workers: The number of workers to use for parallel data loading
             **kwargs: Additional keyword arguments passed to
                 :class:`~torchgeo.datasets.DeepGlobeLandCover`
+
+        .. versionchanged:: 0.4
+           *batch_size* was replaced by *num_tile_per_batch*, *num_patches_per_tile*,
+           and *patch_size*.
```
adamjstewart, line 57: I tend to only document API changes, not internal changes. So the fact that we're now using random cropping isn't documented, only that the parameters changed.

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -3,87 +3,92 @@
...
-        self.num_workers = num_workers
+
+        self.num_tiles_per_batch = num_tiles_per_batch
+        self.num_patches_per_tile = num_patches_per_tile
+        self.patch_size = _to_tuple(patch_size)
         self.val_split_pct = val_split_pct
+        self.num_workers = num_workers
         self.kwargs = kwargs
 
-    def preprocess(self, sample: Dict[str, Any]) -> Dict[str, Any]:
```
adamjstewart, line 46: We cannot use instance methods as transforms, see #886 for what happens when we do.

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -3,87 +3,92 @@
...
-        self.train_dataset: Dataset[Any]
-        self.val_dataset: Dataset[Any]
-
-        if self.val_split_pct > 0.0:
-            self.train_dataset, self.val_dataset, _ = dataset_split(
-                dataset, val_pct=self.val_split_pct, test_pct=0.0
-            )
-        else:
-            self.train_dataset = dataset
-            self.val_dataset = dataset
```
adamjstewart, line 82: I have no idea why our previous logic was so complicated, but I don't think it needs to be.

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -118,12 +120,32 @@ def test_dataloader(self) -> DataLoader[Dict[str, Any]]:
...
+
+        Args:
+            batch: A batch of data that needs to be altered or augmented
+            dataloader_idx: The index of the dataloader to which the batch belongs
+
+        Returns:
+            A batch of data
+        """
+        if self.trainer:
+            if self.trainer.training:
```
adamjstewart, line 139: So much cleaner than our [previous logic](https://github.com/microsoft/torchgeo/blob/v0.3.1/torchgeo/datamodules/inria.py#L229,L235)!

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -118,12 +120,32 @@ def test_dataloader(self) -> DataLoader[Dict[str, Any]]:
...
+                batch = self.train_transform(batch)
+            elif self.trainer.validating or self.trainer.testing:
+                batch = self.test_transform(batch)
+
+        # Kornia adds a channel dimension to the mask
+        batch["mask"] = batch["mask"].squeeze(1)
```
adamjstewart, line 145: Kornia does a lot of weird stuff with transforms that I don't like. Masks are required to be floats (why? slower, more storage). If the mask you input doesn't have a channel dimension, it will add one. Some of the transforms actually break if the mask doesn't have a channel dimension when you input it, so we may need to add an unsqueeze above.

torchgeo/samplers/utils.py
```diff
@@ -4,13 +4,23 @@
...
 
 
+@overload
+def _to_tuple(value: Union[Tuple[int, int], int]) -> Tuple[int, int]:
+    ...
+
+
+@overload
+def _to_tuple(value: Union[Tuple[float, float], float]) -> Tuple[float, float]:
+    ...
```
adamjstewart, line 21: In Python typing, all ints are floats, but not all floats are ints. This meant that if I pass an int as input, mypy would consider its output type to be float. These overloads ensure that int maps to int and float maps to float as expected.

torchgeo/transforms/transforms.py
```diff
@@ -69,3 +74,147 @@ def forward(self, sample: Dict[str, Tensor]) -> Dict[str, Tensor]:
...
+
+        Args:
+            input: the input tensor
+            params: generated parameters
+            flags: static parameters
+
+        Returns:
+            the transformation
+        """
+        out: Tensor = self.identity_matrix(input)
```
adamjstewart, line 104: This isn't correct but something is required and we don't actually use it. I'll iron out the details when we upstream this.

torchgeo/transforms/transforms.py
```diff
@@ -69,3 +74,147 @@ def forward(self, sample: Dict[str, Tensor]) -> Dict[str, Tensor]:
...
+        """
+        size = flags["window_size"]
+        h, w = input.shape[-2:]
+        padding = compute_padding((h, w), size)
+        input = extract_tensor_patches(input, size, size, padding)
+        input = torch.flatten(input, 0, 1)  # [B, N, C?, H, W] -> [B*N, C?, H, W]
+        return input
+
+
+class _RandomNCrop(GeometricAugmentationBase2D):
```
adamjstewart, line 133: I'm concerned that there's a bug in this somewhere. In my tests, if the mask does not have a channel dimension, the crop is incorrect. Can anyone run the trainer and visualize the input/mask for several mini-batches?

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -3,87 +3,92 @@
...
-            self.train_dataset, self.val_dataset, _ = dataset_split(
-                dataset, val_pct=self.val_split_pct, test_pct=0.0
-            )
-        else:
-            self.train_dataset = dataset
-            self.val_dataset = dataset
-
-        self.test_dataset = DeepGlobeLandCover(
-            split="test", transforms=transforms, **self.kwargs
+        train_dataset = DeepGlobeLandCover(split="train", **self.kwargs)
```
nilsleh, line 87: I think the `self.train_transforms` are missing here.

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -3,87 +3,92 @@
...
+        self.train_dataset, self.val_dataset = dataset_split(
+            train_dataset, self.val_split_pct
         )
+        self.test_dataset = DeepGlobeLandCover(split="test", **self.kwargs)
```
nilsleh, line 91: I think self.test_transform are missing here as argument.

torchgeo/datamodules/deepglobelandcover.py
adamjstewart: Transforms are now run on the GPU for each batch, not run on the CPU for each sample. So all data augmentation is moved to `on_after_batch_transfer`.

nilsleh:
```
dm = DeepGlobeLandCoverDataModule(
    root="./data/archive/", num_tiles_per_batch=2, num_patches_per_tile=4, patch_size=64
)

model = SemanticSegmentationTask(
    model="unet",
    backbone="resnet18",
    weights="imagenet",
    in_channels=3,
    num_classes=2,
    loss="jaccard",
    ignore_index=0,
    learning_rate=0.01,
    learning_rate_schedule_patience=0,
)
trainer = Trainer(fast_dev_run=True, log_every_n_steps=1, max_epochs=1)
trainer.fit(model, dm)
```

But in `model.training_step` the input has shape `[8, 3, 64, 64]` but the mask has shape `[4, 2, 64, 64]`.

nilsleh:
Here is a visualization, looks correct.

Input image:
image

Sampled patches:
patches

adamjstewart (approved):

PR #903, opened 2022-11-16
Change name of arguments in Segmentation and Byol task and add timm support for Byol task

nilsleh:
Change the argument name for the tasks according to comment in #854. 

segmentation_model -> model
encoder_name -> encoder

Additionally, add timm model support for byol task.


adamjstewart:
How hard would it be to split BYOL stuff and Segmentation stuff into separate PRs?

nilsleh:
> How hard would it be to split BYOL stuff and Segmentation stuff into separate PRs?

Don't know but I can do that!

torchgeo/trainers/byol.py
```diff
@@ -308,66 +309,74 @@ def update_target(self) -> None:
...
+
+    Supports any available `Timm model
+    <https://rwightman.github.io/pytorch-image-models/>`_
+    as an architecture choice. To see a list of available pretrained
+    models, you can do:
+
+    .. code-block:: python
+
+        import timm
+        print(timm.list_models(pretrained=True)
```
adamjstewart, line 322: Same here, I would suggest listing all available models:
```suggestion
        print(timm.list_models())
```
Also, you were missing a parenthesis

PR #905, opened 2022-11-17
Add Modis dataset

nilsleh:
This PR adds support for the Modis data product from USGS. 

The files are `.hdf` format and I could not figure out a different way than using the rioxarray library with a source installed GDAL version. I have also looked at Planetary Computer data and when you download data from there the bands come as individual `.tif` files, meaning that in a sense torchgeo already has Modis data support when writing a `RasterDataset` class for it.

Issues I am currently having:

- [ ] when running pytest, there occurs `rasterio.errors.NotGeoreferencedWarning: Dataset has no geotransform, gcps, or rpcs. The identity matrix will be returned`  when trying to open the `.hdf`file, when using `ds = Modis(root=test_data_root)` no errors occur
- [ ] generating dummy `.hdf` dummy data is not working yet, and raises `rioxarray.exceptions.TooManyDimensions: Only 2D and 3D data arrays supported` when trying to save the data even though I feel like I only create a 3D array

Plot example:

### DIFF
--- a/docs/api/datasets.rst
+++ b/docs/api/datasets.rst
@@ -108,6 +108,11 @@ Landsat
 .. autoclass:: Landsat2
 .. autoclass:: Landsat1
 
+Modis
+^^^^^
+
+.. autoclass:: Modis
+
 NAIP
 ^^^^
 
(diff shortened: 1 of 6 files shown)


nilsleh:
Closing, as I am not going to have time to work on this.

PR #917, opened 2022-11-27
Add Multi-Weight Support API

### LINKED ISSUE #693: Missing information about sentinel-2 bands used in pretrained ResNet model
### Issue

Hello,

In torchgeo.model the resnet50 model pretrained on Sentinel-2 data uses 10 channels , but I cannot find anywhere which of the 13 channels are used and in what order ? 

Can you clarify that in the documentation ?

### Fix

_No response_

### LINKED ISSUE #762: Multi-Weight Support API
### Summary

We should consider moving towards a multi-weight support API for pre-trained model weights.

### Rationale

Our pre-trained weight support was modeled after how torchvision used to handle weights:
```python
model = resnet50(pretrained=True)
```
However, model weights are not boolean, there are many different potential weights for:

* Different satellites (e.g., Sentinel-2, Landsat 8)
* Band sets (e.g., all bands, RGB-only, false color)
* Downstream tasks (e.g., land cover mapping, object detection)
* Training methods (e.g., supervised, self-supervised)

Currently, we use the following API:
```python
model = resnet50(sensor="sentinel2", bands="all", pretrained=True)
```
However, this has several drawbacks:

1. It's not extensible if we want to add a different downstream task or training strategy
2. It's unclear and undocumented what all available pre-trained models are

### Implementation

Torchvision recently added a multi-weight support API: https://pytorch.org/blog/introducing-torchvision-new-multi-weight-support-api/

We should consider emulating torchvision using enums for all available pre-trained model weights.

### Alternatives

Alternative is "business as usual".

### Additional information

I don't _think_ this would require us to drop support for torchvision 0.12 and older, although it may make things simpler.

### LINKED ISSUE #799: Resnet50 pretrained model documentation
### Issue

- The documentation suggests that the `sensor` flag can take on several sources, but one has to go into the source code to find out which values are possible
- The documentation states that the `bands` flag can accept 'all' or 'rgb', however, currently only 'all' is supported and specifying 'rgb' throws an error
- Add documentation that Sentinel2 imagery is pretrained on 10 channels and which of the overall available 13 channels they are so that downstream tasks can be configured accordingly

I can open an PR if there are suggestions to the above issues and change the documentation accordingly.

### Fix

Be more clear about which values are currently supported and an explanation about how the pretrained weights came about.

nilsleh:
Adds support for loading pretrained weights from various sources, specifically for Earth Observational Data following torchvision's new API.

Example:
```
from torchgeo.models import ResNet50_Weights
from torchgeo.trainers import ClassificationTask

task = ClassificationTask(
    model="resnet50",
    loss="ce",
    num_classes=10,
    in_channels=13,
    weights=ResNet50_Weights.SENTINEL2_ALL_MOCO,
)
```
Some things I have encountered:
- I think torchvision is able to make a cleaner API because all the available weights they have are exclusively for models they themselves implement in torchvision. In the Earth Observational Data case, the pretrained weights I have found come from a variety of sources but mostly use `timm` which aligns with `torchgeo`. However, there are sometimes some hoops to jump through before being able to actually load the state dict into the `timm` model. This is currently handled by `get_state_dict()`
- With the few weights I have already run into issues with naming schemes discussed in #804 and it seems like with trying to collect pretrained weights from a variety of sources for specific use cases will make the proper naming scheme unclear, but maybe there is a more clever way to handle it with the accompanying meta data?

Questions:
- where should the weights be integrated within the torchgeo file hierarchy
- I am not sure what the best way of unit testing this is

### References

* [Introducing TorchVision's New Multi-Weight Support API](https://pytorch.org/blog/introducing-torchvision-new-multi-weight-support-api/)
* [Easily list and initialize models with new APIs in TorchVision](https://pytorch.org/blog/easily-list-and-initialize-models-with-new-apis-in-torchvision/)
* [torchvision.models documentation](https://pytorch.org/vision/stable/models.html)

Closes #693
Closes #762
Closes #799

See #804 for design discussion


torchgeo/models/weights.py
```diff
@@ -0,0 +1,154 @@
+"""Pretrained weights for models."""
```
adamjstewart, line 1: These should be in `torchgeo/models/resnet.py`, this is where they live in [torchvision](https://github.com/pytorch/vision/blob/main/torchvision/models/resnet.py)

torchgeo/models/weights.py
```diff
@@ -0,0 +1,154 @@
...
+import os
+from typing import Any, Dict
+
+import torch
+import torch.nn as nn
+from torchvision.models._api import Weights, WeightsEnum
```
adamjstewart, line 7: I'm guessing this will require us to bump the minimum version of torchvision (and therefore torch)? I'm fine with that.

torchgeo/models/weights.py
```diff
@@ -0,0 +1,154 @@
...
+        url="https://drive.google.com/file/d/1iSHHp_cudPjZlshqWXVZj5TK74P32a2q",
+        transforms=nn.Identity(),
+        meta={
+            "ssl_method": "dino",
+            "publication": "https://arxiv.org/abs/2211.07044",
+            "dataset": "SSL4EO-S12",
+            "repo": "https://github.com/zhu-xlab/SSL4EO-S12",
+        },
+    )
+    DEFAULT = SENTINEL2_ALL_MOCO
```
adamjstewart, line 83: I don't think it makes sense for there to be a DEFAULT set of weights since it depends on the number of input channels.

torchgeo/models/weights.py
```diff
@@ -0,0 +1,154 @@
...
+        },
+    )
+    SENTINEL2_ALL_DINO = Weights(
+        url="https://drive.google.com/file/d/1iSHHp_cudPjZlshqWXVZj5TK74P32a2q",
+        transforms=nn.Identity(),
+        meta={
+            "ssl_method": "dino",
+            "publication": "https://arxiv.org/abs/2211.07044",
+            "dataset": "SSL4EO-S12",
+            "repo": "https://github.com/zhu-xlab/SSL4EO-S12",
```
adamjstewart, line 80: I'm hoping to add all pre-trained models from this site

torchgeo/models/weights.py
```diff
@@ -0,0 +1,154 @@
...
+        transforms=nn.Identity(),
+        meta={
+            "ssl_method": "seco",
+            "dataset": "seco",
+            "publication": "https://arxiv.org/abs/2103.16607",
+            "repo": "https://github.com/ServiceNow/seasonal-contrast",
+        },
+    )
+
+    AERIAL_MILLIONAID_RGB = Weights(
```
adamjstewart, line 47: What does AERIAL mean here? It would be good to have some unique identifier that describes the exact sensor since different planes may have different # channels or resolution

torchgeo/models/weights.py
```diff
@@ -0,0 +1,154 @@
...
+            "publication": "https://arxiv.org/abs/2211.07044",
+            "dataset": "SSL4EO-S12",
+            "repo": "https://github.com/zhu-xlab/SSL4EO-S12",
+        },
+    )
+    DEFAULT = SENTINEL2_ALL_MOCO
+
+    def get_state_dict(self) -> Dict[str, Any]:
+        """Retrieve pretrained weights state_dict."""
+        root = os.path.join(os.getcwd(), "weights")
```
adamjstewart, line 87: Probably makes more sense to use the same default as [torchvision](https://pytorch.org/docs/stable/hub.html#torch.hub.load_state_dict_from_url).

torchgeo/models/weights.py
```diff
@@ -0,0 +1,154 @@
...
+        map_location = torch.device("cpu")
+
+        # need to define identifiers for each case
+        if self.meta["ssl_method"] == "dino":
+            filename = "resnet50_dino_sentinel_weights.pth"
```
adamjstewart, line 92: This implies that every dino model we ever add is for sentinel and resnet50, which certainly won't be true forever

torchgeo/models/weights.py
```diff
@@ -0,0 +1,154 @@
...
+
+        elif self.meta["dataset"] == "imagenet":
+            filename = "resnet50_imagenet_rgb_weights.pth"
+            state_dict = load_checkpoint_from_url(
+                root, filename, self.url, map_location
+            )
+
+        elif self.meta["dataset"] == "millionaid":
+            try:
+                import yacs  # noqa: F401
```
adamjstewart, line 125: What is this and why is it necessary?

adamjstewart (commented):
> where should the weights be integrated within the torchgeo file hierarchy

Let's do what torchvision does. Take a look at their source code to get an idea of how things are structured. So weights for each class of model will go in a model-specific file (resnet.py, etc.). Note that we won't store any of these weights in the source code of TorchGeo, just URLs from which they can be downloaded.

> I am not sure what the best way of unit testing this is

Same way we test our current weights. See `tests/models/test_resnet.py` for details. Unfortunately, these aren't tested very well. In order to improve test times and stability, none of our tests actually access the internet. Instead, we create fake weights and test that we can load those. It's slightly better than nothing, but not by much. If you can figure out a better way to test things (maybe an optional "test with internet access" option that is only used during release testing?) let me know.

torchgeo/models/weights.py
nilsleh: it's necessary for this set of weights, otherwise 'load_checkpoint' will thrown an error saying that this library is missing. so it seems the authors of those weights used [yacs](https://github.com/rbgirshick/yacs).

torchgeo/models/weights.py
nilsleh: Okay, so if we have weights available for a new model, for example some transformer, we would create `torchgeo/models/some_transformer.py` and add the WeightsEnum there?

torchgeo/models/weights.py
nilsleh: The weights are pretrained on  [MillionAid dataset](https://captain-whu.github.io/DiRS/) and come from GoogleEarth so would `GoogleEarth` be better?

torchgeo/models/weights.py
nilsleh: yes, this was just a hack for the moment, still thinking about the best `unique identifier` for the different weights.

torchgeo/models/weights.py
nilsleh: on the note of resnet.py, do you still want to keep the torchgeo resnet implementation in there? It feels not that useful, since we have timm models for the different tasks and the pretrained weights we add now are also based on timm.

adamjstewart: Doesn't matter to me whether we use torchvision or timm, as long as a user can easily load a ResNet model with our pre-trained weights.

adamjstewart:
More potential weights: https://github.com/sustainlab-group/SatMAE

torchgeo/models/weights.py
nilsleh: do you mean naming the default root `checkpoints` instead, that is the default name in torchvision from what I can tell

adamjstewart: See the docs I linked to above. There is a `model_dir` that lets you override the default, which defaults to `<hub_dir>/checkpoints`, where `<hub_dir>` [depends on OS](https://pytorch.org/docs/stable/hub.html#torch.hub.get_dir)

nilsleh:
There is some code duplication at the moment with `load_checkpoint_from_url` and the `adjust_` function because they are used across files for the resnet and vit weights, but don't know in which utils.py I should put them best?

adamjstewart:
Prob in `torchgeo/models/utils.py`

torchgeo/models/weights.py
nilsleh: If I use `torch.hub.load_state_dict_from_url()` I get `_pickle.UnpicklingError: invalid load key, '<'.` so it somehow doesn't work with these weights. It does work with the `load_checkpoint_from_url()` function I wrote. So is it okay to use that and then use `torch.hub.get_dir()` as the root where the weights will be placed?

adamjstewart: Would be worth investigating why the builtin option doesn't work or if we can modify it in some way.

nilsleh: ah I think you can only use torch.hub.load_state_dict_from_url() fresh, so not if you have downloaded it before, so deleting the directory did it.

nilsleh: the suggestions I see for this is to deleted the directory and redownload it but that doesn't solve it. Maybe it's the way the .pth file was created?

nilsleh: maybe this is the issue:
`"https://drive.google.com/file/d/1Tx07L6OilkfcgE2HWiSXHRmRepCPdn6V/view?usp=sharing/B13_vits16_moco_0099_ckpt.pth"` is the link for the download. In the torch source code for `load_checkpoint_from_url()` they use a `urlparse` function which jumps on the `?` after `view` in the url and splits it in such a way that the followed `os.path.basename` is garbage. So will check if I can somehow get different download links?

nilsleh: `torchvision.datasets.utils` has a `download_url` function that torchgeo also uses for datasets downloads and it checks for files being from google drive, but I think for `torch.hub.load_checkpoint_from_url()` torch assumes just one source of checkpoint files because they have all their own torchvision models and checkpoints. So I think that is why it does not work with `torch.hub.load_checkpoint_from_url()` but ckpt download does work when using the more generic torchvision ` download_url` function.

nilsleh:
```
def load_checkpoint_from_url(
    root: str, filename: str, url: str, map_location: torch.device
) -> Dict[str, Any]:
    """Download and load a checkpoint.

    Args:
        root: root directory where checkpoint file should be saved
        filename: filename the downloaded checkpoint file should have
        url: url to checkpoint file to download
        map_location: torch device to load checkpoint

    Returns:
        loaded checkpoint
    """
    if not os.path.exists(os.path.join(root, filename)):
        download_url(url, root=root, filename=filename)
    return torch.load(os.path.join(root, filename), map_location=map_location)
```
This is the function I am using at the moment to download the checkpoints, so it just relies on the `torchvision.datasets` `download_url` function

nilsleh:
> Same way we test our current weights. See `tests/models/test_resnet.py` for details.

You mentioned that these are "fake weights", so should there be a script like `data.py` for each model weight we implement? Could this also be done by downloading the weights we implement myself, then just extracting the state_dict of interest for each weight and zip that one up? Or should the fake weights be created from scratch?

adamjstewart:
> You mentioned that these are "fake weights", so should there be a script like `data.py` for each model weight we implement?

The `data.py` stuff is a temporary measure. We want to be be able to test that we can load data from disk, so we need to somehow get data on disk. At first, we just manually created fake data in `tests/data`, but it was hard to update (e.g., change image size) without a lot of work. We later added `data.py` scripts to automate this. Another benefit of this is that we know where the fake data came from, and can ensure that real data didn't slip into the repo. In the long run, I would actually like to get rid of `tests/data` entirely and generate all fake data on the fly.

Since model weights are effectively new (we're going to remove our current resnet weights and replace them entirely) I think it's best to jump straight to the final goal. I think it makes the most sense to create a [pytest fixture](https://docs.pytest.org/en/6.2.x/fixture.html) for each model weight. This fixture would create a fake model with random weights. We'll either need to write this model to disk and return a filepath, then monkeypatch `load_state_dict_from_file` like we do in `tests/models/test_resnet.py`, OR find a way to keep the model in memory to remove I/O costs.

> Could this also be done by downloading the weights we implement myself, then just extracting the state_dict of interest for each weight and zip that one up? Or should the fake weights be created from scratch?

Either downloading the weights during testing or storing a copy of the downloaded weights would both be a much better test of our code, but we don't want to do this for 3 reasons (the same reasons we don't do this for datasets):

1. **Tests should not rely on internet access**: if they do, we see intermittent CI issues more often than I would like, and downloading can be very slow for large files
3. **Git is not designed to store large files**: it's generally frowned upon to store large binary files in git, as diffing and compression are ineffective
4. **We run into legal restrictions**: depending on the license, we legally can't store files from other repos in our repo, especially if they don't have any license at all

Since these weights are relatively small (compared to datasets), I think a good compromise would be as follows. Let's design the unit tests as I described above with fake models generated on the fly. Our goal here is to get 100% coverage, not to actually test that the real weights work. Then, we create integration tests that test the real weights. These tests will only be run when creating a new release, so we don't have to worry about CI issues or time. The way to do this is to mark the tests as slow (see `tests/samplers/test_single.py::TestGeoSampler::test_dataloader` for an example of this). Our slow tests are only run on release branches, not PR branches.

nilsleh:
Thanks for the elaborate explanation, that's good to know. In the above commit, I made a draft for the unit tests, however, struggling a bit. 

1. CI tests: Idea was as you said to create a fixture for each pretrained weight we include in models. Then to run the tests, I though to make a parameterization to run over the fixtures (don't know if that is good practice). However, I am not sure how I can dynamically set the url of a specific weight, because I need to set the attribute, i.e.:
```
monkeypatch.setattr(
        torchgeo.models.ResNet50_Weights.<dynamically model weight name>, "url", ckpt_path
    )
```
2. slow tests: atm commented out, where I am getting the error `test_resnet50_weights_download: function already takes an argument 'weight_name' with a default value` which I don't really get.

adamjstewart:
> Then to run the tests, I though to make a parameterization to run over the fixtures (don't know if that is good practice).

This is good practice.

> However, I am not sure how I can dynamically set the url of a specific weight, because I need to set the attribute

You can either parametrize the [function](https://docs.pytest.org/en/7.2.x/how-to/parametrize.html) or the [fixture](https://docs.pytest.org/en/7.2.x/how-to/fixtures.html#fixture-parametrize). Either way, instead of parametrizing with a list of objects, you can use a list of tuples (name, object, url, etc.) so that you have access to everything you need.

> test_resnet50_weights_download: function already takes an argument 'weight_name' with a default value

Replace `weight_name=str` with `weight_name: str`.

nilsleh:
> Replace `weight_name=str` with `weight_name: str`.

:facepalm:

nilsleh:
I am not sure but are the `minimum` and `readthedocs` test both failing due to pytorch version differences?

nilsleh:
I also have a question about the proper documentation for the multi-weight api. I feel like it would make sense to have an overview table like the dataset tables that describe the features of the available pretrained weights. Thoughts on this?

adamjstewart:
> I am not sure but are the minimum and readthedocs test both failing due to pytorch version differences?

The minimum tests are failing because `WeightsEnum` did not exist in torchvision 0.10.0, you'll need to bump the minimum supported version (in several files). This will also cause the minimum supported version of torch to change as well.

The readthedocs tests are failing because `WeightsEnum` isn't documented. You'll need to add this to the `nitpick_ignore` list in `docs/conf.py`.

> I also have a question about the proper documentation for the multi-weight api. I feel like it would make sense to have an overview table like the dataset tables that describe the features of the available pretrained weights. Thoughts on this?

At the very least it would be good to have a table with benchmark accuracy of each model on a few datasets. You don't have to calculate this yourself, you can just borrow what the authors reported. I believe torchvision has something similar for all of their models. I also want to document where each model weight comes from so people can cite them appropriately, similar to how we do this for all datasets. If I'm getting greedy, it would also be good to have a tutorial on how to choose a model and how to use it, both with and without PyTorch Lightning. If documentation is taking too long and I don't have time to write it myself (I know you're leaving on a trip soon) then I'll just merge this and wait until after the release to worry about proper documentation.

nilsleh:
I changed the unittests so that they also check that the pretrained weights work with our trainers. This means that the `weights` parameter in addition to `random`, `imagenet`, a str `ckpt_path` must now also support a `state_dict: Dict[str, Tensor]` which is what the `WeightsEnum` returns. For the `ClassificationTask` and `RegressionTask` it is fairly straightforward. Although they implement essentially the same logic, so potentially make a utils function that implements the loading functionality?

Regarding the BYOLTask, it should be the same but additionally, we would have to introduce an `input_size` parameter to the task, because at the moment the `input_size` to `BYOL` class is hardcoded as (256, 256). However, the vit small architecture requires an input size of (224, 224). Such an `input_size` parameter would generally make it more flexible so seems useful. However, since this would mean a lot more file changes, I abstained from doing those changes in this PR.

Regarding the SegmentationTask, Pytorch Segmentation Models only accepts a str argument to specify the backbone and available weights ("random", "imagenet", in a few cases more). So I am guessing that it is a bit more work to replace the timm encoder with a pretrained one after initialization.

nilsleh:
> Could we accept a string as input and add a function that looks up the correct enum?

The WeightEnum can be called like a dict: ` ResNet18_Weights["<weight_name"]`, so `<weight_name>` could be specified as a string in a YAML file. Should I add a function in `torchgeo.models.utils` that looks up the the `weight_name` corresponding to an available backbone and return the enum?

adamjstewart:
Wouldn't it have to be specified as "<model>_<weight_name>" in the YAML file so you also know that it's ResNet18 and not ViT? Anything you can do to make the trainer inputs and the yaml inputs similar would be appreciated.

nilsleh:
My current idea is something like in utils.py:

```
BACKBONE_NAMES = ["resnet18", "resnet50", "vit_small_patch16_224"]

def lookup_pretrained_weights(backbone: str, weight_name: str) -> Weights:
    """Lookup correct pretrained weights.
    
    This is useful when running experiments from a YAML file and
    looking for the desired pretrained weight based on *weight_name*.

    Args:
        backbone: backbone name
        weight_name: pretrained weight name

    Returns:
        Pretrained Weight from WeightEnum

    Raises:
        ValueError if *backbone* does not have pretrained weights available
    """
    if backbone == "resnet18":
        return resnet_weights.ResNet18_Weights[weight_name]
    elif backbone == "resnet50":
        return resnet_weights.ResNet50_Weights[weight_name]
    elif backbone == "vit_small_patch16_224":
        return vit_weights.VITSmall16_Weights[weight_name]
    else:
        raise ValueError(f"Currently, only pretrained weights for {BACKBONE_NAMES} are available.")
```

With a config file that looks like:
```
experiment:
  task: "eurosat"
  module:
    loss: "ce"
    model: "resnet18"
    learning_rate: 1e-3
    weights: "SENTINEL2_ALL_MOCO"
    in_channels: 13
    num_classes: 2
  datamodule:
    root: "tests/data/eurosat"
    batch_size: 1
    num_workers: 0
```

and then do something like this,
```
model_kwargs_without_weights_key = {key: val for key, val in config["module"] if key not in ["weights"]}
task = ClassificationTask(
    weights=lookup_pretrained_weights(config["module"]["model"], config["module"]["weights"].get_state_dict(),
    **model_kwargs_without_weights_key
)
```

Currently running into circular imports between `models/utils.py` and `models/resnet.py`.

adamjstewart:
To resolve the circular imports you can just create a third `torchgeo/models/weights.py` file.

torchgeo/models/weights.py
adamjstewart: I'm probably going to remove this specific set of weights. We can add it later and see if there's an easier way to load it. Maybe we can repackage it ourselves and upload it somewhere so yacs isn't required.

adamjstewart:
Currently working on adopting the API from https://pytorch.org/blog/easily-list-and-initialize-models-with-new-apis-in-torchvision/. This will allow us to store weight names as strings in YAML files and load them directly into trainers.

adamjstewart:
Next steps to keep myself organized:

- [ ] Download, modify, and re-save all weights in a format that can be loaded by timm
- [ ] Upload as assets to the most recent release on GitHub
- [ ] Add entrypoints that load a model given its weights
- [ ] Add hubconf.py file that points to these entrypoints
- [ ] Test that `torch.hub` works as expected

adamjstewart:
I'm probably going to organize performance metric tables by satellite instead of by model. Otherwise the table would be super sparse since most datasets aren't applicable to most satellites. This begs the question of whether we should also organize our weight enums like this as well. Or just have a single enum for all weights. Or drop torchvision's weights enum logic entirely and roll our own. Will experiment and see what I like best.

adamjstewart:
We now support [PyTorch Hub](https://pytorch.org/hub/):
```
$ python
Python 3.10.8 (main, Dec 19 2022, 16:10:16) [Clang 12.0.0 (clang-1200.0.32.29)] on darwin
>>> import torch
>>> from torchgeo.models import ResNet18_Weights
>>> torch.hub.list('nilsleh/torchgeo:models/weights', trust_repo=True)
Downloading: "https://github.com/nilsleh/torchgeo/zipball/models/weights" to /Users/Adam/.cache/torch/hub/models_weights.zip
['resnet18', 'resnet50', 'vit_small_patch16_224']
>>> model = torch.hub.load('nilsleh/torchgeo:models/weights', 'resnet18')
Using cache found in /Users/Adam/.cache/torch/hub/nilsleh_torchgeo_models_weights
>>> model = torch.hub.load('nilsleh/torchgeo:models/weights', 'resnet18', weights=ResNet18_Weights.SENTINEL2_RGB_MOCO)
Using cache found in /Users/Adam/.cache/torch/hub/nilsleh_torchgeo_models_weights
```
See [torch.hub](https://pytorch.org/docs/stable/hub.html) docs for more details.

adamjstewart:
Tests pass locally, no idea why they're failing in CI...

nilsleh:
> Tests pass locally, no idea why they're failing in CI...

for me the tests also fail locally, with the same errors as CI, but don't know why yet.

adamjstewart:
I think the failing model tests are because it's trying to load the resnet-18 weights in a resnet-50 (and vice versa). But I have no idea why it would be trying to do that, or why it doesn't affect me on macOS. It's interesting that the ViT tests don't fail, only ResNet.

adamjstewart:
Next TODO: all trainers should except a `weights` argument that is either an enum or a string (or None). We'll need some kind of lookup mechanism that translates from the string to the enum. String support is important for YAML config files.

nilsleh:
I think *what* was happening is that the loading url from the monkey patch was always the first created mocked weights. For example,the first listed resnet18 weight is an RGB weight and the subsequent weight a 13 band weight giving an error because still trying to load the RGB weight. The tests for vit were passing because they are both 13 band weights but I think under the hood it was also loading the first mocked weight for the second test. 

Making the path where the mocked weights are saved unique by adding the weight name made it work. However, not sure *why* the above issue existed in the first place because the `tmp_path` is new for every test.

nilsleh:
Don't know why the windows tests are failing now though...

adamjstewart:
Windows tests are prob because of \ instead of /. This is fine for a path but not fine for a URL. I can fix this.

adamjstewart:
Not sure what's up with `tests/data/models/resnet50-sentinel2-2.pt.zip`, it shows up in Files Changed but not locally so I can't delete it.

calebrob6 (approved):

calebrob6:
I love it! My only question/concern is whether we have validated that the pre-trained weight files produce something reasonably close to what was claimed where the weight files were released?

adamjstewart:
We have not. What I would really like is a script to loop over all weights and datasets and populate the CSV table so that we can 1) validate them, and 2) fill in the missing metrics. But I don't currently have the compute to do this.

PR #918, opened 2022-11-27
Change BYOL task argument name and switch to timm models

nilsleh:
Change the argument name for the tasks according to comment in #854. And switch from torchvision models to timm models.


torchgeo/trainers/byol.py
```diff
@@ -308,66 +309,76 @@ def update_target(self) -> None:
...
+
+    Supports any available `Timm model
+    <https://rwightman.github.io/pytorch-image-models/>`_
+    as an architecture choice. To see a list of available pretrained
+    models, you can do:
+
+    .. code-block:: python
+
+        import timm
+        print(timm.list_models(pretrained=True)
```
adamjstewart, line 322: Should we remove the `pretrained=True` arg here? What's the default?

nilsleh: The default is False. I think if you are starting of with a model, pretrained imagenet is most likely better than from scratch so more helpful to see those models.

adamjstewart: I would personally leave it off since all models are valid, not just pretrained models.

adamjstewart (dismissed):

nilsleh:
Sorry, I think I messed this up by also changing something in utils.py to get the tests to pass, but probably that should be its own PR change.

nilsleh:
Mhm now I am actually not sure what the best way to proceed is, because the utils.py function extract_encoder (extract_backbone) needs to look for backbone because that is the argument name for the trainer now, so I think I also need to change utils.py?

adamjstewart:
I think this is the last torchvision -> timm/argument rename PR? Should be easier to convert everything in unison now.

tests/trainers/test_utils.py
```diff
@@ -52,26 +52,26 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         return cast(torch.Tensor, self.conv1(x))
 
 
-def test_extract_encoder_unsupported_model(tmp_path: Path) -> None:
+def test_extract_backbone_unsupported_model(tmp_path: Path) -> None:
     checkpoint = {"hyper_parameters": {"some_unsupported_model": "resnet18"}}
     path = os.path.join(str(tmp_path), "dummy.ckpt")
     torch.save(checkpoint, path)
-    err = "Unknown checkpoint task. Only encoder or model extraction is supported"
+    err = "Unknown checkpoint task. Only backbone or model" " extraction is supported"
```
adamjstewart, line 59:
```suggestion
    err = "Unknown checkpoint task. Only backbone or model extraction is supported"
```

torchgeo/trainers/utils.py
```diff
@@ -39,18 +42,18 @@ def extract_encoder(path: str) -> Tuple[str, "OrderedDict[str, Tensor]"]:
...
+            {k: v for k, v in state_dict.items() if "model.backbone.model" in k}
         )
         state_dict = OrderedDict(
-            {k.replace("model.encoder.model.", ""): v for k, v in state_dict.items()}
+            {k.replace("model.backbone.model.", ""): v for k, v in state_dict.items()}
         )
     else:
         raise ValueError(
-            "Unknown checkpoint task. Only encoder or model extraction is supported"
+            "Unknown checkpoint task. Only backbone or model" " extraction is supported"
```
adamjstewart, line 56:
```suggestion
            "Unknown checkpoint task. Only backbone or model extraction is supported"
```

conf/byol.yaml
```diff
@@ -8,7 +8,7 @@ experiment:
   name: "test_byol"
   module:
     model: "byol"
-    encoder: "resnet18"
+    backbone: "resnet18"
     input_channels: 4
     imagenet_pretraining: True
```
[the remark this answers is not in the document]
nilsleh: do you mean adding a weights parameter to specify pretrained weights for resnet18?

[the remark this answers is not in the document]
adamjstewart: I mean should we rename `imagenet_pretraining` to `weights`?

adamjstewart (approved):

PR #919, opened 2022-11-27
Change segmentation model argument names

nilsleh:
Change the argument name for the tasks according to comment in #854, and add doc strings.


evaluate.py
```diff
@@ -179,7 +179,7 @@ def main(args: argparse.Namespace) -> None:
     elif issubclass(TASK, SemanticSegmentationTask):
```
adamjstewart, line 179: Same problem with chmod here

nilsleh: I think that is from resolving the merge conflict inside github and committing from there.

adamjstewart (approved):

calebrob6 (dismissed):

evaluate.py
```diff
@@ -179,7 +179,7 @@ def main(args: argparse.Namespace) -> None:
...
         val_row = {
             "split": "val",
-            "segmentation_model": hparams["segmentation_model"],
+            "model": hparams["model"],
             "encoder_name": hparams["encoder_name"],
             "encoder_weights": hparams["encoder_weights"],
```
adamjstewart, line 184: These should be `encoder` and `weights` right?

tests/conf/chesapeake_cvpr_7.yaml
```diff
@@ -2,7 +2,7 @@ experiment:
   task: "chesapeake_cvpr"
   module:
     loss: "ce"
-    segmentation_model: "unet"
+    model: "unet"
     encoder_name: "resnet18"
     encoder_weights: null
     encoder_output_stride: 16
```
adamjstewart, line 8:
```suggestion
    backbone: "resnet18"
    weights: null
    backbone_output_stride: 16
```

tests/conf/chesapeake_cvpr_prior.yaml
```diff
@@ -2,7 +2,7 @@ experiment:
   task: "chesapeake_cvpr"
   module:
     loss: "ce"
-    segmentation_model: "unet"
+    model: "unet"
     encoder_name: "resnet50"
     encoder_weights: null
     encoder_output_stride: 16
```
adamjstewart, line 8:
```suggestion
    backbone: "resnet50"
    weights: null
    backbone_output_stride: 16
```

tests/conf/sen12ms_s1.yaml
```diff
@@ -2,11 +2,11 @@ experiment:
...
     loss: "focal"
-    segmentation_model: "fcn"
+    model: "fcn"
     num_filters: 1
-    encoder_name: "resnet18"
-    encoder_weights: null
-    encoder_output_stride: 16
+    backbone: "resnet18"
+    weights: null
+    backbone_output_stride: 16
```
adamjstewart, line 9: Wait a minute, is the `backbone_output_stride` parameter actually used? If not, let's just remove it from all conf files.

adamjstewart (changes requested):

tests/conf/chesapeake_cvpr_7.yaml
nilsleh: I did not change the "encoder" naming here, because this file is not used by the `test_segmentation` trainer pytest. It is used by the Byol task so I thought I would change them separate, because otherwise byol task tests would fail.

tests/conf/chesapeake_cvpr_prior.yaml
nilsleh: same reasoning as above.

tests/conf/sen12ms_s1.yaml
nilsleh: I think it is not used, "encoder_output_stride" is a valid argument for the pytorch-segmentation-models model input, but we do not mention it elsewhere explicitly.

adamjstewart: Gotcha. If we don't use it, I would just remove it.

tests/conf/chesapeake_cvpr_7.yaml
adamjstewart: Thanks for the clarification. I'll let you update these in the BYOL PR then.

adamjstewart (approved):

PR #920, opened 2022-11-27
Change argument name Object Detection Task

nilsleh:
And also change the model argument name for the object detection task:
detection_model -> model

### DIFF
--- a/tests/conf/nasa_marine_debris.yaml
+++ b/tests/conf/nasa_marine_debris.yaml
@@ -1,7 +1,7 @@
 experiment:
   task: "nasa_marine_debris"
   module:
-    detection_model: "faster-rcnn"
+    model: "faster-rcnn"
     backbone: "resnet18"
     num_classes: 2
     learning_rate: 1.2e-4
--- a/torchgeo/trainers/detection.py
+++ b/torchgeo/trainers/detection.py
@@ -50,7 +50,7 @@ class ObjectDetectionTask(pl.LightningModule):
     def config_task(self) -> None:
         """Configures the task based on kwargs parameters passed to the constructor."""
         backbone_pretrained = self.hyperparams.get("pretrained", True)
-        if self.hyperparams["detection_model"] == "faster-rcnn":
+        if self.hyperparams["model"] == "faster-rcnn":
             if "resnet" in self.hyperparams["backbone"]:
                 kwargs = {
                     "backbone_name": self.hyperparams["backbone"],
@@ -88,21 +88,24 @@ class ObjectDetectionTask(pl.LightningModule):
             )
 
         else:
-            raise ValueError(
-                f"Model type '{self.hyperparams['detection_model']}' is not valid."
-            )
+            raise ValueError(f"Model type '{self.hyperparams['model']}' is not valid.")
 
     def __init__(self, **kwargs: Any) -> None:
         """Initialize the LightningModule with a model and loss function.
 
         Keyword Args:
-            detection_model: Name of the detection model type to use
+            model: Name of the detection model type to use
             backbone: Name of the model backbone to use
             in_channels: Number of channels in input image
             num_classes: Number of semantic classes to predict
+            learning_rate: Learning rate for optimizer
+            learning_rate_schedule_patience: Patience for learning rate scheduler
 
         Raises:
             ValueError: if kwargs arguments are invalid
+
+        .. versionchanged:: 0.4
+           The *detection_model* parameter was renamed to *model*.
         """
         super().__init__()
         # Creates `self.hparams` from kwargs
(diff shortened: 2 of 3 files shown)


tests/conf/nasa_marine_debris.yaml
```diff
@@ -1,7 +1,7 @@
 experiment:
   task: "nasa_marine_debris"
   module:
-    detection_model: "faster-rcnn"
+    model: "faster-rcnn"
     backbone: "resnet18"
```
adamjstewart, line 5: I think we decided to call this "encoder"?

nilsleh: Oh I understood, that like in segmentation where the Unet is the model, for Object detection the faster-rcnn would be the model. So you want to go detection_model -> encoder? It kind of feels together with "backbone".

adamjstewart: No, what I meant was "detection_model" -> "model" is correct, but shouldn't it be "backbone" -> "encoder"?

nilsleh: https://github.com/microsoft/torchgeo/pull/854#issuecomment-1328656114 maybe to summarize it all together.

ashnair1: Doesn't "encoder" imply there is a "decoder"? Segmentation models follow an encoder decoder architecture where the pre-trained backbone (ResNet, VGG etc) acts as the encoder. I believe that's why `segmentation.models` use that term. But in other tasks I've seen them usually referred to as backbone.

adamjstewart: Maybe we should use "model" and "backbone" everywhere instead of "model" and "encoder"? Are there any models that let you replace both the "encoder" and "decoder" separately?

calebrob6: > Are there any models that let you replace both the "encoder" and "decoder" separately?

Not that I know of -- the decoder part of models in `smp` is fixed.

"backbone" is a good choice

adamjstewart: I think there's two questions here:

1. Can we use the same name for all sub-models?
2. Should we call them encoders or backbones?

I agree that encoder kind of implies decoder, which won't be the case for all models. Would backbone be a better more generic name for what we're describing?

calebrob6: no, they are both good choices -- I lean towards encoder because it is more specific to me, but I don't have a strong opinion as long as we are consistent. An encoder to me is any architecture that produces a latent representation

torchgeo/trainers/detection.py
```diff
@@ -50,7 +50,7 @@ class ObjectDetectionTask(pl.LightningModule):
     def config_task(self) -> None:
```
adamjstewart, line 50: Did you want to add the same blurb above saying where to find a list of available models and backbones?

nilsleh: yes, that would be good a good documentation.  don't know what the best way is in this case, but I made a proposal

torchgeo/trainers/detection.py
```diff
@@ -44,13 +44,24 @@
...
+    Currently, supports a Faster R-CNN model from
+    `torchvision
+    <https://pytorch.org/vision/stable/models/faster_rcnn.html>`_ with
+    one of the following *backbone* arguments:
+
+    .. code-block:: python
+
+        ['resnet18', 'resnet34', 'resnet50', 'resnet101', 'resnet152',
+        'resnext50_32x4d','resnext101_32x8d', 'wide_resnet50_2',
+        'wide_resnet101_2']
```
adamjstewart, line 56: Did we want to use timm here too or is that not easy to do? It would allow us to ignore all the weights craziness above.

nilsleh: Agreed, but not sure if it works to have any timm backbone with a torchvision detection architecture, maybe @ashnair1 knows or has an opinion? Otherwise, I have time to test it tomorrow.

adamjstewart: segmentation-models-pytorch doesn't support object detection, does it?

ashnair1: Plugging in timm backbones into torchvision detectors would be difficult.  https://github.com/pytorch/vision/issues/6172 links to an implementation of Faster RCNN with a timm backbone but it looks like a lot of work was needed for it to function. 

> segmentation-models-pytorch doesn't support object detection, does it?

No

adamjstewart: Let's stick with what we have for now and work on improvements if we find a better option in the future.

adamjstewart (approved):

PR #928, opened 2022-12-03
Add datamodule for GID-15 dataset

### LINKED ISSUE #855: Random sized patches support for other non-geospatial datamodules

nilsleh:
This PR adds a datamodule for the GID-15 based on #855 with the discussed random crop logic in #876.

Closes #855


adamjstewart:
This will need the same treatment as #974

adamjstewart (approved):

adamjstewart (approved):

adamjstewart (approved):

pyproject.toml
```diff
@@ -88,6 +88,8 @@ filterwarnings = [
     # https://github.com/lanpa/tensorboardX/issues/653
     # https://github.com/lanpa/tensorboardX/pull/654
     "ignore:Call to deprecated create function:DeprecationWarning:tensorboardX",
+    # https://github.com/kornia/kornia/issues/777
+    "ignore:Default upsampling behavior when mode=bilinear is changed to align_corners=False since 0.4.0:UserWarning:torch.nn.functional",
```
adamjstewart, line 92: This is only needed because our tests are trying to crop a 2x2 patch from a 1x1 image, which obviously requires resizing. If someone ever implements a `data.py` for GID-15 and increases the image size, this can be removed.

PR #932, opened 2022-12-05
Fix failing minimum tests

ashnair1:
This PR fixes the cause for the minimum tests failing in #854 (and presumably any PR created after Dec 1st). 

`scikit-learn` has officially begun deprecating `sklearn` to prevent malicious actors from using it. So any dependency that attempts to install `sklearn` would install the new `sklearn` package which throws an exception  at certain intervals - Refer https://pypi.org/project/sklearn/ for more details.

`open3d=0.11.2` was our problem package. The open3d team had switched from `sklearn` to `scikit-learn` in version `0.13.0`. This PR bumps the minimum version to the same.

It seems there was a PR prepped to replace `open3d` with `pyvista` - #663. If this lands for the `0.4.0` release, this PR can be ignored.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,8 +39,8 @@ install_requires =
     omegaconf>=2.1,<3
     # packaging 17+ required by pytorch-lightning
     packaging>=17,<22
-    # pillow 6.2+ required by matplotlib
-    pillow>=6.2,<10
+    # pillow 8.2+ required by open3d>=0.13
+    pillow>=8.2,<10
     # pyproj 2.2+ required for CRS object
     pyproj>=2.2,<4
     # pytorch-lightning 1.5.1+ required for apply_to_collection bugfix
--- a/setup.cfg
+++ b/setup.cfg
@@ -78,9 +78,8 @@ datasets =
     h5py>=2.6,<4
     # laspy 2+ required for laspy.read
     laspy>=2,<3
-    # open3d 0.11.2+ required to avoid GLFW error:
-    # https://github.com/isl-org/Open3D/issues/1550
-    open3d>=0.11.2,<0.15;python_version<'3.10'
+    # open3d 0.13.0+ required to avoid using deprecated sklearn dependency
+    open3d>=0.13.0,<0.15;python_version<'3.10'
     # opencv-python 3.4.2.17 is oldest buildable version on PyPI
     opencv-python>=3.4.2.17,<5
     # pandas 0.23.2+ required for Python 3.7 wheels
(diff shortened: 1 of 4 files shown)


adamjstewart:
Good catch! Looks like we'll also need to bump our pandas min version as a result. @isaaccorley any updates on the pyvista PR?

setup.cfg
```diff
@@ -78,9 +78,8 @@ datasets =
     h5py>=2.6,<4
     # laspy 2+ required for laspy.read
     laspy>=2,<3
-    # open3d 0.11.2+ required to avoid GLFW error:
-    # https://github.com/isl-org/Open3D/issues/1550
-    open3d>=0.11.2,<0.15;python_version<'3.10'
+    # open3d 0.13.0+ required to avoid using deprecated sklearn dependency
+    open3d>=0.13.0,<0.15;python_version<'3.10'
```
adamjstewart, line 82:
```suggestion
    # open3d 0.13+ required to avoid using deprecated sklearn dependency
    open3d>=0.13,<0.15;python_version<'3.10'
```

ashnair1:
> Good catch! Looks like we'll also need to bump our pandas min version as a result. @isaaccorley any updates on the pyvista PR?

Indeed. Do you recommend that be done in a separate PR? Or shall I just include it here?

ashnair1:
Really kicked the hornet's nest with this one. Will need some time to find a stable configuration.

Current constraint: 

open3d 0.13.0 depends on jupyter-packaging~=0.10  and jupyter-packaging 0.10+ depends on setuptools>=46.4.0

ashnair1:
Got stung a few times but things look stable now.

**Summary of changes and reasons:**

open3d (`0.13.0`) required the following packages to be updated
- setuptools (indirectly)
- pandas  
- numpy
- pillow
- ipywidgets

`rasterio` was updated to avoid a deprecation warning caused by importing Iterable from collections.
`nbconvert` was pinned to reduce the resolve time. Without this, time required for running the minimum test jumped to 10 min.

adamjstewart:
> rasterio was updated to avoid a deprecation warning caused by importing Iterable from collections.

Let's silence that warning in `pyproject.toml` instead.

ashnair1:
> > rasterio was updated to avoid a deprecation warning caused by importing Iterable from collections.
> 
> Let's silence that warning in `pyproject.toml` instead.

Seems like this was already added in `pyproject.toml` line 80

isaaccorley:
I stay up late at night regretting ever adding open3d as a dependency.

adamjstewart:
Do you think we should push forward and merge this PR and then revert it if we get pyvista working, or is pyvista close? We could skip pyvista tests on macOS/Windows if those are too hard to test.

isaaccorley:
If we can skip windows and mac tests then we could get pyvista in. It's just a matter of installing xvfb via apt in the github action and then running pyvista.start_xvfb() prior to the plot test.

PR #965, opened 2022-12-20
Datamodule for SpaceNet1

calebrob6:
Some things to note:
- SpaceNet1 patches are (406, 438) so I zero-pad to 448 x 448
- I add one to the mask layer in preprocessing so the classes are {nodata, background, building}
- I ignore the 0 class (the value that gets added due to padding) in the train config
- I changed SpaceNet's `data.py` to generate 3 samples by default so that we can test train/val/test (@ashnair1)

I can confirm that this is working with `train.py`, here's a shot from TensorBoard

image


docs/api/datamodules.rst
```diff
@@ -94,6 +94,11 @@ So2Sat
 
 .. autoclass:: So2SatDataModule
 
+SpaceNet
+^^^^^^
+
+.. autoclass:: SpaceNet
```
adamjstewart, line 100:
```suggestion
SpaceNet
^^^^^^^^

.. autoclass:: SpaceNet1
```

tests/datamodules/test_spacenet.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
adamjstewart, line 1: Can we instead test this with a trainer and delete this file?

train.py
```diff
@@ -51,12 +53,14 @@
     "etci2021": (SemanticSegmentationTask, ETCI2021DataModule),
     "inria": (SemanticSegmentationTask, InriaAerialImageLabelingDataModule),
     "landcoverai": (SemanticSegmentationTask, LandCoverAIDataModule),
+    "loveda": (SemanticSegmentationTask, LoveDADataModule),
```
adamjstewart, line 56: I don't believe SemanticSegmentationTask supports LoveDADataModule. I think I tried adding it to our tests but it doesn't work.

adamjstewart:
> SpaceNet1 patches are (406, 438) so I zero-pad to 448 x 448

I'm guessing U-Nets require 1) square patches, and 2) patch sizes that are multiples of 32?

torchgeo/datamodules/spacenet.py
```diff
@@ -0,0 +1,178 @@
...
+
+# https://github.com/pytorch/pytorch/issues/60979
+# https://github.com/pytorch/pytorch/pull/61045
+DataLoader.__module__ = "torch.utils.data"
+
+
+class SpaceNet1DataModule(pl.LightningDataModule):
+    """LightningDataModule implementation for the SpaceNet1 dataset.
+
+    Randomly splits into train/val/test.
```
adamjstewart, line 24: Needs:
```rst
.. versionadded:: 0.4
```

train.py
calebrob6: I trained models with it and it worked IRL -- I think this is probably a test issue

adamjstewart: Hmm, can you remove this change from this PR and add it to a separate PR where you also try to get the tests working? I'm hesitant to put things in `train.py` if we're unable to test them properly, and it seems unrelated to this PR anyway.

calebrob6: sure

calebrob6:
> I'm guessing U-Nets require 1) square patches, and 2) patch sizes that are multiples of 32?

They just need patch sizes that are multiples of 32. There are 5 downsampling operations by default (going down the left side of the "U") which reduces the spatial resolution of the input by a factor of 32. If the input isn't evenly divisible by 32, then there will  be some rounding, and in the upsample part of the "U", the spatial dimensions of the features in the skip connections won't match.

tests/datamodules/test_spacenet.py
calebrob6: This brought up an issue in plotting. In SpaceNet's plot method we do percentile normalization of the input image (https://github.com/microsoft/torchgeo/blob/main/torchgeo/datasets/spacenet.py#L327). Now, because SpaceNet1 test data is 2 x 2 and I zero-pad to 448 x 448, we get a divide by zero error in the percent normalization method.

Options:
- Make SpaceNet test data larger (this is what I did)
- Add an epsilon value in the denominator of the percentile normalization method
- Switch to a subsistence farming lifestyle and never touch a computer again

calebrob6:
This is good to go pending the question in the above comment about larger test images vs. epsilon in the percentile normalization.

torchgeo/datamodules/spacenet.py
```diff
@@ -0,0 +1,183 @@
...
+            )
+            x, y = train_augmentations(x, y)
+
+            # torchmetrics expects masks to be longs without a channel dimension
+            batch["image"] = x
+            batch["mask"] = y.squeeze(1).long()
+
+        return batch
+
+    def preprocess(self, sample: Dict[str, Any]) -> Dict[str, Any]:
```
adamjstewart, line 101: I clearly haven't written a datamodule in a while. Is there a reason to do this here instead of in `on_after_batch_transfer`?

tests/data/spacenet/data.py
```diff
@@ -75,7 +75,7 @@ def create_test_image(img_dir: str, imgs: List[str]) -> List[List[float]]:
     """
     for img in imgs:
         imgpath = os.path.join(img_dir, img)
-        Z = np.arange(4, dtype="uint16").reshape(2, 2)
+        Z = np.random.randint(0, 255, size=(256, 256), dtype="uint16")
```
adamjstewart, line 78: What's the minimum size that will work? In most other datasets, we use 32x32 or 64x64.

tests/datamodules/test_spacenet.py
adamjstewart: As long as it isn't possible to hit this zero division error when using the real data, I'm fine with either 1 or 3. 2 might be a good idea though regardless. Up to you.

torchgeo/datamodules/spacenet.py
calebrob6: No worries!

`on_after_batch_transfer` only happens when transferring to the GPU for the training set while `preprocess` happens to every sample

tests/data/spacenet/data.py
calebrob6: 2x2 would work with the epsilon change due to the way we mock the model

torchgeo/datamodules/spacenet.py
adamjstewart: Only the training set is run on the GPU? Or on_after_batch_transfer is only run for the training set?

calebrob6: `on_after_batch_transfer` is run when any batch is transferred from the CPU to the GPU (see https://pytorch-lightning.readthedocs.io/en/latest/data/datamodule.html#on-after-batch-transfer), we currently have checks in place to only do the augmentation when we are training (not val, test, or inference). This is cool because it lets us do augmentation logic directly on the GPU.

I like to separate the image preprocessing to make it clear what you need to do to the images to use the models in inference mode. I.e. if we release trained models that have used this SpaceNet1DataModule, it is quite clear what inputs the model expects (SpaceNet1 dataset output fed through `preprocess`)

tests/datamodules/test_spacenet.py
calebrob6: I chose 2

torchgeo/datamodules/spacenet.py
adamjstewart: Personally I would run everything on the GPU so that `on_after_batch_transfer` runs normalization and optionally runs augmentation when training. But this is the same as all of our existing datamodules, so we can consider changing this at a later date.

torchgeo/datamodules/spacenet.py
```diff
@@ -0,0 +1,183 @@
...
+
+        Args:
+            sample: dictionary containing image and mask
+
+        Returns:
+            preprocessed sample
+        """
+        sample["image"] = sample["image"].float()
+        sample["image"] = self.padto(sample["image"]).squeeze()
+        sample["image"] /= 255.0
```
adamjstewart, line 112:
```suggestion
        sample["image"] = sample["image"].float() / 255
        sample["image"] = self.padto(sample["image"]).squeeze()
```
Division before padding is faster.

adamjstewart (approved):

PR #974, opened 2022-12-24
DataModules: only instantiate when download requested

### LINKED ISSUE #967: DataModules: skip prepare_data
### Summary

We should skip the `prepare_data` step in our data modules unless downloading is actually required.

### Rationale

Instantiating the dataset is slow. Currently, all of our datamodules instantiate the dataset at least twice on the off chance that someone is requesting the dataset to be downloaded. This isn't needed most of the time.

### Implementation

A better solution would be to replace:
```python
def prepare_data(self) -> None:
    Dataset(**self.kwargs)
```
with this:
```python
def prepare_data(self) -> None:
    if self.kwargs.get("download", False):
        Dataset(**self.kwargs)
```
in all of our data modules.

### Alternatives

_No response_

### Additional information

_No response_

adamjstewart:
Closes #967

### DIFF
--- a/tests/datamodules/test_loveda.py
+++ b/tests/datamodules/test_loveda.py
@@ -19,7 +19,11 @@ class TestLoveDADataModule:
         scene = ["rural", "urban"]
 
         dm = LoveDADataModule(
-            root=root, scene=scene, batch_size=batch_size, num_workers=num_workers
+            root=root,
+            scene=scene,
+            batch_size=batch_size,
+            num_workers=num_workers,
+            download=True,
         )
 
         dm.prepare_data()
--- a/torchgeo/datamodules/naip.py
+++ b/torchgeo/datamodules/naip.py
@@ -31,8 +31,6 @@ class NAIPChesapeakeDataModule(pl.LightningDataModule):
 
     def __init__(
         self,
-        naip_root: str,
-        chesapeake_root: str,
         batch_size: int = 64,
         num_workers: int = 0,
         patch_size: int = 256,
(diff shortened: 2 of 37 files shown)


tests/datamodules/test_loveda.py
```diff
@@ -19,7 +19,11 @@ def datamodule(self) -> LoveDADataModule:
         scene = ["rural", "urban"]
 
         dm = LoveDADataModule(
```
adamjstewart, line 21: This file will be deleted in #966, this is only temporary

torchgeo/datamodules/naip.py
```diff
@@ -41,22 +39,25 @@ def __init__(
...
-            naip_root: directory containing NAIP data
-            chesapeake_root: directory containing Chesapeake data
             batch_size: The batch size to use in all created DataLoaders
             num_workers: The number of workers to use in all created DataLoaders
             patch_size: size of patches to sample
             **kwargs: Additional keyword arguments passed to
-                :class:`~torchgeo.datasets.NAIP` and
-                :class:`~torchgeo.datasets.Chesapeake13`
+                :class:`~torchgeo.datasets.NAIP` (prefix keys with naip_) and
+                :class:`~torchgeo.datasets.Chesapeake13` (prefix keys with chesapeake_)
```
adamjstewart, line 47: When I first added this I was lazy and didn't provide a way to separate these. This seems like a good solution to me.

torchgeo/datamodules/so2sat.py
```diff
@@ -103,13 +103,6 @@ def preprocess(self, sample: Dict[str, Any]) -> Dict[str, Any]:
 
         return sample
 
-    def prepare_data(self) -> None:
-        """Make sure that the dataset is downloaded.
-
-        This method is only called once per run.
-        """
-        So2Sat(**self.kwargs)
```
adamjstewart, line 111: This dataset cannot be automatically downloaded, no reason to prepare_data

isaaccorley (approved):

PR #560, opened 2022-06-07
Add prediction utilities

ashnair1:
The main aim of this PR is to add utilities that will allow users to use a `torchgeo` segmentation model to create predictions on ~GeoTIFFs~ images of arbitrary size. 

Example using a model trained on the `InriaAerialImageLabelling` dataset shown below. Images in this dataset are of size 5000 x 5000 and the model was trained on 512 x 512 crops. 

image

PR #975, opened 2022-12-24
InriaAerialImageLabelingDataModule: fix predict dimensions

adamjstewart:
Previously, the following code would fail due to an extra dimension in predict samples:
```python
datamodule = InriaAerialImageLabelingDataModule(...)
model = SemanticSegmentationTask(...)
trainer = Trainer(...)

trainer.predict(model=model, datamodule=datamodule)
```
This is now fixed and properly tested with a real trainer. I also removed the inria datamodule tests since they aren't useful.


tests/trainers/test_segmentation.py
```diff
@@ -80,7 +82,9 @@ def test_trainer(
         trainer = Trainer(fast_dev_run=True, log_every_n_steps=1, max_epochs=1)
         trainer.fit(model=model, datamodule=datamodule)
         trainer.test(model=model, datamodule=datamodule)
-        trainer.predict(model=model, dataloaders=datamodule.val_dataloader())
+
+        if hasattr(datamodule, "predict_dataset"):
+            trainer.predict(model=model, datamodule=datamodule)
```
adamjstewart, line 87: The only reason we run predict on the val set instead of the predict set is because not all datamodules have a predict set. In order to get 100% coverage, we ran everything on val instead. However, at least for segmentation, we do have a predict set, so we should use it. This is how I discovered the bug to begin with.

torchgeo/datamodules/inria.py
```diff
@@ -98,7 +98,7 @@ def patch_sample(self, sample: Dict[str, Any]) -> Dict[str, Any]:
             self.patch_size,
             padding=padding,
         )
-        sample["image"] = rearrange(sample["image"], "() t c h w -> t () c h w")
+        sample["image"] = rearrange(sample["image"], "b n c h w -> (b n) c h w")
```
adamjstewart, line 101: `extract_tensor_patches` takes in "b c h w" and returns "b n c h w" where n is the number of patches extracted per image. This additional axis causes our code to fail. I think we should merge this into the batch dimension to fix this issue.

See https://kornia.readthedocs.io/en/latest/contrib.html#kornia.contrib.ExtractTensorPatches for details.

adamjstewart:
@ashnair1 this likely breaks your code in #560. I can't think of an easy way to support both "b c h w" _and_ "b n c h w" in `forward`, but if you can think of one let me know. Otherwise, you'll need to "b c h w -> b n c h w" manually before running `CombineTensorPatches`.

torchgeo/datamodules/inria.py
ashnair1: Could you add this line just after `extract_tensor_patches`?
 
```
sample["num_patches"] = sample["image"].shape[1]
```

This could be then used to reshape it back to "n b c h w" as is required in #560

ashnair1:
> @ashnair1 this likely breaks your code in #560. I can't think of an easy way to support both "b c h w" _and_ "b n c h w" in `forward`, but if you can think of one let me know. Otherwise, you'll need to "b c h w -> b n c h w" manually before running `CombineTensorPatches`.

 Can't think of a way to support both. Adding the number of patches to the sample dict as suggested in my comment will allow reshaping prior to `CombineTensorPatches`. That should work for now.

ashnair1 (approved):

PR #980, opened 2022-12-26
OSCDDataModule: fix dimensions when val_pct == 0

adamjstewart:
Haven't yet figured out how to fix this bug, but at least we're now aware of it.


adamjstewart:
@iejMac any ideas how to fix these failing tests? There seems to be a bug in the implementation when `val_pct` == 0.

isaaccorley:
Just took a quick glance and seems like if val_pct=0 then val_dataset = train_dataset, however the val dataloader doesn't get the same custom collate_fn passed to it. That might be a quick fix.

adamjstewart:
Yeah, it seems like `val_pct > 0` means use test transforms for all datasets and `val_pct == 0` means use train transforms for all datasets, neither of which makes sense. The train transforms do not seem to produce dimensions that are compatible, but we keep squeeze and indexing in random places and I have no idea what the right approach is.

isaaccorley:
I think the issue is that the train collate fn flattens the 2 images into the channel dimension so this can work with our segmentation trainer. But this doesn't happen when val_pct=0 for the val dataloader

adamjstewart:
Bigger issue is that this is a change detection dataset and we don't have a ChangeDetectionTask trainer. I think we should just remove this from the SemanticSegmentationTask trainer tests entirely.

PR #988, opened 2022-12-29
Fix for warning raised if model outputs no box predictions

ashnair1:
`draw_bounding_boxes` raises a warning if no boxes are passed. PR fixes this by calling the function only if `sample["prediction_boxes"]` has non zero elements.


torchgeo/datasets/nasa_marine_debris.py
```diff
@@ -236,7 +236,7 @@ def plot(
         image = draw_bounding_boxes(image=sample["image"], boxes=sample["boxes"])
```
adamjstewart, line 236: Can you do the same thing for "boxes"? Are there any images in the original dataset without positive examples?

adamjstewart (commented):
This is the same thing we do for IDTReeS so should be correct.

adamjstewart (approved):

PR #991, opened 2022-12-29
Added CSVLogger

### LINKED ISSUE #705: Add CSVLogger to train.py
### Summary

Our current `train.py` script only logs to TensorBoard. We should add a CSV logger.

### Rationale

In order to determine the optimal hyperparams in a hyperparam sweep, we need an easy way of finding the final val accuracy of each run. Currently, the only way to get this is to convert the TensorBoard logger output to a CSV. If we log directly to CSV this becomes much easier, especially for those who don't use TensorBoard.

### Implementation

_No response_

### Alternatives

_No response_

### Additional information

_No response_

bugraaldal:
Resolves #705 by adding CSVLogger to train.

### DIFF
--- a/train.py
+++ b/train.py
@@ -170,6 +170,7 @@ def main(conf: DictConfig) -> None:
     # Setup trainer
     ######################################
     tb_logger = pl_loggers.TensorBoardLogger(conf.program.log_dir, name=experiment_name)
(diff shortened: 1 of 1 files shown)


adamjstewart (approved):
Thanks, didn't realize this was so easy!

PR #992, opened 2022-12-30
DataModules: run all data augmentation on the GPU

### LINKED ISSUE #336: OSCDDataModule not compatible with SemanticSegmentationTask
OSCD is a change detection dataset. `OSCDDataset.__getitem__` returns an "image" of shape [2 x C x H x W]. Most of our other datasets return images of shape [C x H x W], and `SemanticSegmentationTask` is designed to work with these kind of images. It looks like there are some hacks in `OSCDDataModule` to work around this, but these hacks are very unreliable. For example, in:
```python
mask = repeat(sample["mask"], "h w -> t h w", t=2).float()               
image, mask = self.rcrop(sample["image"], mask)                          
mask = mask.squeeze()[0]                                                 
images.append(image.squeeze())                                           
masks.append(mask.long())         
```
we use `squeeze()` without specifying which dimension to remove. This has different behavior when batch size is 1 vs >1, so we should avoid using this.

To fix this, I think we could either change `__getitem__` to return an "image" of shape [2C x H x W] or create a new `ChangeDetectionTask` for these kinds of datasets.

See the following comments for additional discussion/issues:

* https://github.com/microsoft/torchgeo/pull/329#pullrequestreview-840627512
* https://github.com/microsoft/torchgeo/pull/329#issuecomment-1003144109

### LINKED ISSUE #337: Discrepancies between Chesapeake datamodules/trainers
During #329, I noticed a lot of issues with our Chesapeake datamodules and how they integrate with BYOLTask and SemanticSegmentationTask. For example:

* https://github.com/microsoft/torchgeo/pull/329#pullrequestreview-840627429
* https://github.com/microsoft/torchgeo/pull/329#pullrequestreview-841013077
* https://github.com/microsoft/torchgeo/pull/329#pullrequestreview-841102399

We should specify which dimension to squeeze over to fix the "batch size of 1 doesn't work" issue, and then fix the other issues that pop up when we do that. I believe there are some inconsistencies between:

* the data returned by ChesapeakeCVPRDataModule when using prior/no priors
* the data expected by SemanticSegmentationTask/BYOLTask

### LINKED ISSUE #619: Cheasapeake test set image and label size are 2x bigger
### Description

the dataloader from the test set of Cheasapeak returns images 2x bigger.

Most likely due to test_transforms from the setup function which has a different configuration than e.g. the validation set. It uses `pad_to(self.original_patch_size...` instead of `center_crop(self.patch_size)`

### Steps to reproduce

```python

   def convert_test():
    states = ["de", "md", "va", "wv", "pa", "ny"]

    dm = ChesapeakeCVPRDataModule(
        root_dir=SRC_DATASET_DIR,
        train_splits=[f"{state}-train" for state in states],
        val_splits=[f"{state}-val" for state in states],
        test_splits=[f"{state}-test" for state in states],
        patches_per_tile=500,
        patch_size=256,
        batch_size=1,
        num_workers=0,
        class_set=7,
    )
    dm.prepare_data()
    dm.setup()

    test_dl = dm.test_dataloader()

    for sample in test_dl:
        image = np.array(sample["image"])[0]
        print(image.shape)  # this value is 512,512 and should be 256,256

```

### Version

0.2.0

### LINKED ISSUE #886: Trainers: num_workers > 0 results in pickling error on macOS/Windows
i have a problem running torchgeo on eurosat MS tif images i wrote a simple code :
```python
eurosat = EuroSAT(euro_root, split="train", download=False)
# sampler = RandomGeoSampler(eurosat, size=64, length=10000)
# dataloader = DataLoader(eurosat, batch_size=128, sampler=sampler, collate_fn=stack_samples)
dataloader = DataLoader(eurosat, batch_size=128, collate_fn=stack_samples)
num_classes = 10
channels = 13
num_workers = 4
batch_size = 4
backbone = "resnet50"
weights = "imagenet"
lr = 0.01
lr_schedule_patience = 5
epochs = 50
datamodule = EuroSATDataModule(
     root_dir=euro_root,
     batch_size=batch_size,
     num_workers=num_workers,
)
task = ClassificationTask(
    classification_model=backbone,
    weights=weights,
    num_classes=num_classes,
    in_channels=channels,
    loss="ce",
    learning_rate=lr,
    learning_rate_schedule_patience=lr_schedule_patience
)
checkpoint_callback = pl.callbacks.ModelCheckpoint(
    monitor="val_loss",
    save_top_k=1,
    save_last=True,
)
early_stopping_callback = pl.callbacks.EarlyStopping(
    monitor="val_loss",
    min_delta=0.00,
    patience=10,
)

# Train
trainer = pl.Trainer(
    callbacks=[checkpoint_callback, early_stopping_callback],
    max_epochs=epochs
)
````
but i get this error on fit line :
````
.........
_pickle.PicklingError: Can't pickle <class 'nn.BatchNorm2d.BatchNorm2d'>: import of module 'nn.BatchNorm2d' failed
..........
OSError: [Errno 22] Invalid argument: 'C:\\Users\\mohsc\\PycharmProjects\\pythonProject\\<input>'
````
am i missing a preparing step or something ?
help would be appreciated, by the way tutorials are still lacking essential sample codes and documentations like this issue on MS images.
thanks

### PR DESCRIPTION
**Pull request by adamjstewart**
This PR overhauls all of our data modules to improve uniformity. This includes the following changes:

- [x] Add GeoDataModule and NonGeoDataModule base classes to reduce code duplication
- [x] Only instantiate the datasets that are needed for a particular stage
- [x] Replace torchvision with kornia (better support for MSI, GPU, inverse)
- [x] Replace dataset transforms with on_after_batch_transfer (CPU -> GPU, sample -> batch, faster)
- [x] Remove instance methods for preprocessing (fixes #886)
- [x] Fix bug where train/val/test split would differ for each stage and images would leak between sets
- [x] Deprecate `torchgeo.transforms.AugmentationSequential` (use `kornia.augmentation.AugmentationSequential` instead)
- [x] Deprecate `torchgeo.datamodules.dataset_split` (use `torch.utils.data.random_split` instead)

In a future PR, I'm planning on extending this to the rest of our transforms:

* Rewrite all index transforms to be compatible with Kornia (#999)
* Update tutorials to use Kornia with our transforms
* Upstream and remove our custom transforms and AugmentationSequential hacks

Fixes #619
Fixes #337
Fixes #336

### CODE REVIEW

**Comment by adamjstewart** [comment]
I'm having trouble understanding the failing BYOL tests. The same datamodules work fine with SemanticSegmentationTask. If anyone can figure out how to fix these, we can see how much coverage we lack and fix that.

File: torchgeo/datamodules/geo.py [hunk 1]
```diff
@@ -0,0 +1,544 @@
...
+    def __init__(
+        self,
+        dataset_class: Type[GeoDataset],
+        batch_size: int = 1,
+        patch_size: Union[int, Tuple[int, int]] = 64,
+        length: int = 1000,
+        num_workers: int = 0,
+        **kwargs: Any,
+    ) -> None:
+        """Initialize a new NonGeoDataModule instance.
```
ashnair1, line 45: NonGeoDataModule -> GeoDataModule

torchgeo/datamodules/geo.py
```diff
@@ -0,0 +1,544 @@
...
+            dataloader_idx: The index of the dataloader to which the batch belongs.
+
+        Returns:
+            A batch of data.
+        """
+        if self.trainer:
+            if self.trainer.training:
+                aug = self.train_aug or self.aug
+            elif self.trainer.validating:
+                aug = self.val_aug or self.aug
```
ashnair1, line 307: When we train, we also run some sanity checking epochs so that should be considered as well

```diff
- if self.trainer.training:
+ if self.trainer.training or self.trainer.sanity_checking
```

adamjstewart: Opened https://github.com/Lightning-AI/lightning/discussions/16266 to get to the bottom of this since the source code isn't clear.

ashnair1: Correct. My bad. Those are validation epochs

ashnair1: Btw for context, I tried training the Inria model again and that's how I found this issue.

adamjstewart: Why doesn't the inria test fail? Would like to improve the tests if we need to.

torchgeo/datamodules/inria.py
```diff
@@ -39,214 +24,73 @@ class InriaAerialImageLabelingDataModule(pl.LightningDataModule):
...
-            self.train_dataset = self.dataset
-            self.val_dataset = self.dataset
-            self.test_dataset = self.dataset
-
-        self.predict_dataset = InriaAerialImageLabeling(
-            split="test", transforms=test_transforms, **self.kwargs
+        self.val_aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
+            _ExtractTensorPatches(self.patch_size),
+            data_keys=["image", "mask"],
```
ashnair1, line 75: Think `val_aug` should be the same as `train_aug` just without the random rotations and flips.

adamjstewart: So you want to replace _ExtractTensorPatches with _RandomNCrop during validation? What about for the other datamodules that use this?

torchgeo/datamodules/geo.py
adamjstewart: Oh interesting, [`fast_dev_run`](https://pytorch-lightning.readthedocs.io/en/stable/common/trainer.html#fast-dev-run) implies `num_sanity_val_steps=0`.

torchgeo/datamodules/inria.py
ashnair1: It should be the same treatment. `_ExtractTensorPatches` is when we want all the patches to be considered i.e. during prediction. Otherwise the batch_size won't be honored and you would trigger OOM during validation in between training.

adamjstewart: Hmm, when evaluating (validation, test), don't we want to evaluate performance on the entire file, not just a few random patches?

ashnair1: Hmm not sure. The validation/test scores are an indicator of how well the model has generalized the patterns seen in the training set on the validation/test sets. So more patches, the more representative the score is. I guess the issue is more practical. 

Assuming a patch_size of 512, each image (5000 x 5000) in the inria dataset can be considered as a (100, 3, 512, 512) tensor

During evaluation, 
If we use RandomNCrops, we're looking at (N, 3, 512, 512) images per iteration whereas with ExtractTensorPatches we're looking at (100, 3, 512, 512) images per iteration.

adamjstewart: Is your point that N != 100 or that 100 is too large? This is the behavior we decided on in #876. All data modules with large images do this:

* DeepGlobe
* GID-15
* Inria
* OSCD
* Potsdam
* Vaihingen

The data module docs do make things clear at least:
```
The Inria Aerial Image Labeling dataset contains images that are too large to    
pass directly through a model. Instead, we randomly sample patches from image    
tiles during training and chop up image tiles into patch grids during            
evaluation. During training, the effective batch size is equal to                
``num_tiles_per_batch`` x ``num_patches_per_tile``.        
```

ashnair1: 100 being too large. But I guess this might be hardware dependent

adamjstewart: We could define NonGeoSamplers so that we can emulate GridGeoSampler for NonGeoDatasets, but that feels like overkill. I would like to target the lowest possible hardware requirements though, so if it's a bottleneck we can change it. Are you able to run `trainer.predict` with `ExtractTensorPatches` though?

adamjstewart: But `predict_aug` also uses `ExtractTensorPatches`, so shouldn't it have the same problem? Are you saying you don't have this problem because you don't use `trainer.predict()`, you do things manually? We should get the former working if it isn't working.

ashnair1: Yeah I didn't use `trainer.predict` in my prediction PR. I was manually running inference on each patch.

I'll take a look at creating predictions for inria using `trainer.predict` and see if it works.

torchgeo/datamodules/potsdam.py
```diff
@@ -45,117 +41,48 @@ def __init__(
...
+        self.train_aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
             _RandomNCrop(self.patch_size, self.num_patches_per_tile),
             data_keys=["image", "mask"],
         )
-        self.test_transform = AugmentationSequential(
-            Normalize(mean=0.0, std=255.0),
+        self.aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
             _ExtractTensorPatches(self.patch_size),
```
nilsleh, line 72: I am currently using this with the real dataset and observing the following: `_ExtractTensorPatches` extracts all possible patches from the scene, which if I choose a `patch_size` of 128 and given that the scene has a dimension of `4, 6000, 6000` yields a model input batch of `2209, 4, 128, 128` which is heavy on GPU memory.

torchgeo/datamodules/vaihingen.py
```diff
@@ -45,117 +41,48 @@ def __init__(
...
-        self.train_transform = AugmentationSequential(
-            Normalize(mean=0.0, std=255.0),
+        self.train_aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
             _RandomNCrop(self.patch_size, self.num_patches_per_tile),
             data_keys=["image", "mask"],
         )
-        self.test_transform = AugmentationSequential(
-            Normalize(mean=0.0, std=255.0),
+        self.aug = AugmentationSequential(
```
nilsleh, line 70: The real Vaihingen2D dataset consists of images that are not of the same size. When lightning tries tries to collate `num_tiles_per_batch` scenes before the augmentations that make all inputs the same size are applied, the error occurs.

torchgeo/datamodules/potsdam.py
```diff
@@ -45,117 +41,48 @@ def __init__(
...
-        self.train_transform = AugmentationSequential(
-            Normalize(mean=0.0, std=255.0),
+        self.train_aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
             _RandomNCrop(self.patch_size, self.num_patches_per_tile),
             data_keys=["image", "mask"],
         )
-        self.test_transform = AugmentationSequential(
-            Normalize(mean=0.0, std=255.0),
+        self.aug = AugmentationSequential(
```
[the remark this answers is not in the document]
adamjstewart: Is this with or without https://github.com/microsoft/torchgeo/pull/992/commits/2ea65a10b7a875d10f2d125b524a4f5c0ffe11dd?

torchgeo/datamodules/vaihingen.py
adamjstewart: Did we have this same problem for any other datamodules? Trying to remember if I've fixed this before...

adamjstewart: Oh, it was ChesapeakeCVPR where I had to do something like this. I'll try that approach.

adamjstewart:
@ashnair1 @nilsleh how would you propose to solve the ExtractTensorPatches OOM issue?

Easiest fix is to use RandomNCrop instead (this is what we do during train). I'm happy to do this. However, it doesn't really make sense during val/test, and isn't useful at all during predict.

Alternative fix would be to introduce a `GridNonGeoSampler` that works similarly to `GridGeoSampler` and allows a single image to span multiple batches. However, this would require all `NonGeoDatasets` to change from `__getitem__(idx: int)` to `__getitem__(idx: int, rows: slice, cols: slice)` which would be a ton of work.

Any other ideas for how to handle this?

torchgeo/datamodules/vaihingen.py
adamjstewart: A few options I can use:

1. Remove `num_tiles_per_batch`, hard code `batch_size=1`
2. Apply `RandomNCrop` to sample instead of batch, dimensions may get tricky
3. Pad each sample to 3,816

Preferences?

ashnair1:
> @ashnair1 @nilsleh how would you propose to solve the ExtractTensorPatches OOM issue?
> 
> Easiest fix is to use RandomNCrop instead (this is what we do during train). I'm happy to do this. However, it doesn't really make sense during val/test, and isn't useful at all during predict.
> 
> Alternative fix would be to introduce a `GridNonGeoSampler` that works similarly to `GridGeoSampler` and allows a single image to span multiple batches. However, this would require all `NonGeoDatasets` to change from `__getitem__(idx: int)` to `__getitem__(idx: int, rows: slice, cols: slice)` which would be a ton of work.
> 
> Any other ideas for how to handle this?

A trivial solution to OOM would be to run inference in an iterative manner whenever we use `_ExtractTensorPatches`. 

For example, in the `validation_step` of `torchgeo/trainers/segmentation.py` we iterate over the batch instead of passing in the entire batch. 

```diff
def validation_step(self, *args: Any, **kwargs: Any) -> None:
    """Compute validation loss and log example predictions.

    Args:
        batch: the output of your DataLoader
        batch_idx: the index of this batch
    """
    batch = args[0]
    batch_idx = args[1]
    x = batch["image"]
    y = batch["mask"]
-   y_hat = self(x)
+   from einops import rearrange
+   y_hat_list = []
+   for i in range(x.shape[0]):
+       out = rearrange(x[i], 'c h w -> 1 c h w') 
+       out = self(out)
+       y_hat_list.append(out)    
+   y_hat = rearrange(y_hat_list, 'b 1 c h w -> b c h w')
    y_hat_hard = y_hat.argmax(dim=1)

    loss = self.loss(y_hat, y)
```

adamjstewart:
> run inference in an iterative manner whenever we use `_ExtractTensorPatches`

I don't think there's an easy way to tell whether or not we are using `_ExtractTensorPatches`. We could say "if batch size > X run iteratively" but I'm not sure what X should be.

ashnair1:
How about making this the default behaviour for val, test and predict steps?

X is dependent on the hardware RAM which can only determined by the user. So I don't think we can go down that route.

adamjstewart:
I think replacing all val/test/predict steps from batch processing to a for-loop over the batch size would be extremely bad for speed.

nilsleh:
> We could say "if batch size > X run iteratively" but I'm not sure what X should be.
Don't know if it valid approach, but could you wrap `y_hat = self(x)` in a try, except block checking for OOM, and in the except block do the looping?

torchgeo/datamodules/potsdam.py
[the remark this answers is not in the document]
nilsleh: this was without that fix, now it is working without setting `num_sanity_val_steps`

torchgeo/datamodules/vaihingen.py
nilsleh: 1. the random image sizes occur for all different data splits, so setting `batch_size=1` during training does not seem great. 
2. that could be a possibility
3. I don't think I like this approach, because in some cases the padding would be so extreme, also there might be other new datasets coming in that have a similar problem

Not sure what you think of this approach, but I tried the following where I define a collate_fn for the dataloader:
```
def custom_collate(batch):
    images = [item["image"] for item in batch]
    masks = [item["mask"] for item in batch]
    return {"image": images, "mask": masks}
```

And then in `on_after_batch_transfer`:
```
if self.trainer:
        if self.trainer.training:
            # Kornia requires masks to have a channel dimension
            batch["mask"] = [rearrange(item, "h w -> () h w") for item in batch["mask"]]
            augmented = []
            for img, mask in zip(batch["image"], batch["mask"]):
                augmented.append(self.train_transform({"image": img, "mask": mask}))
           batch = {
                "image": torch.cat([item["image"] for item in augmented], dim=0),
                "mask": torch.cat([item["mask"] for item in augmented], dim=0)
           }
```
I guess it defeats the purpose a bit of applying augmentations on GPU to a batch, but I thought it keeps the general new workflow of applying augmentations in torchgeo datamodules.

adamjstewart:
If I recall, it isn't possible to check for OOM, the program crashes without raising an error. But correct me if I'm wrong. Also, it would be difficult to get test coverage of that branch.

nilsleh:
On the pytorch [docs](https://pytorch.org/docs/stable/notes/faq.html#my-out-of-memory-exception-handler-can-t-allocate-memory), they seem to suggest a possible way as a RuntimeError and as you said difficult to get test coverage.

torchgeo/datamodules/geo.py
```diff
@@ -0,0 +1,540 @@
...
+        """Run the plot method of the dataset if one exists.
+
+        Args:
+            *args: Arguments passed to plot method.
+            **kwargs: Keyword arguments passed to plot method.
+
+        Returns:
+            A matplotlib Figure with the image, ground truth, and predictions.
+        """
+        dataset = self.val_dataset or self.dataset
```
ashnair1, line 537:
```diff
- dataset = self.val_dataset or self.dataset
+ dataset = self.dataset or self.val_dataset
```
self.val_dataset can be of type `torch.utils.data.Subset` which doesn't have a plot method and `or` will choose the first option.

ashnair1 (commented):
Currently training fails during the sanity checking (validation) step `fig = datamodule.plot(sample)`  as `fig` will be of type None because `self.val_dataset`  (of type `torch.utils.data.Subset`) does not have a plot method.

adamjstewart:
Bleh. `fast_dev_run` disables sanity checks and there's no way to re-enable them. The bug you just reported is easy to fix, but I'm concerned about being able to test this in CI. If CI isn't catching these kinds of bugs, then they're bound to happen again. We could try adding slower tests that run only on release that remove `fast_dev_run` and train on the full (test) dataset.

torchgeo/datamodules/geo.py
```diff
@@ -0,0 +1,538 @@
...
+        """Apply batch augmentations to the batch after it is transferred to the device.
+
+        Args:
+            batch: A batch of data that needs to be altered or augmented.
+            dataloader_idx: The index of the dataloader to which the batch belongs.
+
+        Returns:
+            A batch of data.
+        """
+        if self.trainer:
```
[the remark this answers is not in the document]
adamjstewart: Feel free to add docs

adamjstewart:
Before I forget: one bug this PR likely introduces is removing seeds from random split methods. Since `setup` is run during each phase, a different split will be chosen during each stage, meaning images will leak between sets. Need to fix this before merging.

torchgeo/datamodules/vaihingen.py
adamjstewart: 1. I could do this for all stages, not just train

adamjstewart: Other benefit of 1 is that we can remove all this complicated "effective batch_size = num_tiles_per_batch x num_patches_per_tile" stuff and just have a single `batch_size` parameter that is the number of patches in a single tile. Each batch may be less diverse, but data loading should be faster, and the API becomes simpler.

calebrob6:
PR name should be something like "DataModules: Adamified" -- "run all data augmentation on the GPU" doesn't quiteeeee describe it :)

torchgeo/datamodules/chesapeake.py
```diff
@@ -3,21 +3,53 @@
...
-from ..samplers.batch import RandomBatchGeoSampler
-from ..samplers.single import GridGeoSampler
+from ..datasets import ChesapeakeCVPR
+from ..samplers import GridGeoSampler, RandomBatchGeoSampler
+from ..transforms import AugmentationSequential
+from .geo import GeoDataModule
 
 
-class ChesapeakeCVPRDataModule(LightningDataModule):
+class _Transform(nn.Module):
```
calebrob6, line 20: Should this have a TODO remove?

adamjstewart: Not sure how to remove. Kornia transforms turn all [B, C, H, W] and [C, H, W] into [B, C, H, W]. Personally I think that's a bug in Kornia (samples should not have a batch dimension added) but I'm not sure if I can convince the Kornia devs of this.

torchgeo/datamodules/geo.py
```diff
@@ -0,0 +1,540 @@
...
+                aug = self.test_aug or self.aug
+            elif self.trainer.predicting:
+                aug = self.predict_aug or self.aug
+
+            batch = aug(batch)
+
+        return batch
+
+    def plot(self, *args: Any, **kwargs: Any) -> plt.Figure:
+        """Run the plot method of the dataset if one exists.
```
calebrob6, line 311: Maybe we should comment that this method really only expects to be called from the trainer (if you are in "test" phase then nothing will happen silently because the self.dataset / self.val_dataset are not defined)

torchgeo/datamodules/landcoverai.py
```diff
@@ -22,154 +21,32 @@ class LandCoverAIDataModule(pl.LightningDataModule):
...
         )
-
-        self.val_dataset = LandCoverAI(
-            split="val", transforms=val_test_transforms, **self.kwargs
-        )
-
-        self.test_dataset = LandCoverAI(
-            split="test", transforms=val_test_transforms, **self.kwargs
+        self.aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std), data_keys=["image", "mask"]
```
calebrob6, line 51: Why specify data_keys here and not above?

calebrob6 (commented):
I'm not a huge fan of having `NonGeoDataModule` in "geo.py", do you think something like "base.py" would be appropriate?

adamjstewart:
> I'm not a huge fan of having `NonGeoDataModule` in "geo.py", do you think something like "base.py" would be appropriate?

We're doing the same thing with `NonGeoDataset`, so if we change one we should change both. The current name comes from `torchvision/models/vision.py`, but they don't have a `NonVisionDatset` obviously. I would be fine with changing to `base.py`, but let's consider doing that in a different PR since it also affects the datasets.

calebrob6:
When training on NASAMarineDebris I'm getting the following traceback:
```
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py", line 1103, in _run
    results = self._run_stage()
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py", line 1182, in _run_stage
    self._run_train()
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py", line 1195, in _run_train
    self._run_sanity_check()
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py", line 1267, in _run_sanity_check
    val_loop.run()
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/pytorch_lightning/loops/loop.py", line 199, in run
    self.advance(*args, **kwargs)
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/pytorch_lightning/loops/dataloader/evaluation_loop.py", line 152, in advance
    dl_outputs = self.epoch_loop.run(self._data_fetcher, dl_max_batches, kwargs)
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/pytorch_lightning/loops/loop.py", line 199, in run
    self.advance(*args, **kwargs)
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/pytorch_lightning/loops/epoch/evaluation_epoch_loop.py", line 121, in advance
    batch = next(data_fetcher)
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/pytorch_lightning/utilities/fetching.py", line 184, in __next__
    return self.fetching_function()
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/pytorch_lightning/utilities/fetching.py", line 265, in fetching_function
    self._fetch_next_batch(self.dataloader_iter)
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/pytorch_lightning/utilities/fetching.py", line 280, in _fetch_next_batch
    batch = next(iterator)
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 628, in __next__
    data = self._next_data()
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 1333, in _next_data
    return self._process_data(data)
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 1359, in _process_data
    data.reraise()
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/torch/_utils.py", line 543, in reraise
    raise exception
RuntimeError: Caught RuntimeError in DataLoader worker process 0.
Original Traceback (most recent call last):
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/torch/utils/data/_utils/worker.py", line 302, in _worker_loop
    data = fetcher.fetch(index)
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/torch/utils/data/_utils/fetch.py", line 61, in fetch
    return self.collate_fn(data)
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/torch/utils/data/_utils/collate.py", line 265, in default_collate
    return collate(batch, collate_fn_map=default_collate_fn_map)
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/torch/utils/data/_utils/collate.py", line 128, in collate
    return elem_type({key: collate([d[key] for d in batch], collate_fn_map=collate_fn_map) for key in elem})
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/torch/utils/data/_utils/collate.py", line 128, in <dictcomp>
    return elem_type({key: collate([d[key] for d in batch], collate_fn_map=collate_fn_map) for key in elem})
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/torch/utils/data/_utils/collate.py", line 120, in collate
    return collate_fn_map[elem_type](batch, collate_fn_map=collate_fn_map)
  File "/home/calebrobinson/.conda/envs/torchgeo/lib/python3.8/site-packages/torch/utils/data/_utils/collate.py", line 163, in collate_tensor_fn
    return torch.stack(batch, 0, out=out)
RuntimeError: stack expects each tensor to be equal size, but got [1, 4] at entry 0 and [2, 4] at entry 1
```

This is likely because the dataset getitem is returning a "boxes" key with different shapes.

calebrob6 (approved):

calebrob6:
Just confirmed that BigEarthNet trainer/datamodule works with `train.py`

torchgeo/datamodules/nasa_marine_debris.py
```diff
@@ -3,35 +3,14 @@
...
-
-    Args:
-        batch: list of sample dicts return by dataset
-
-    Returns:
-        batch dict output
-    """
-    output: Dict[str, Any] = {}
-    output["image"] = torch.stack([sample["image"] for sample in batch])
-    output["boxes"] = [sample["boxes"] for sample in batch]
```
adamjstewart, line 29: @calebrob6 starting a thread here so we can discuss how to handle NASA Marine Debris.

I think the solution in this `collate_fn` is actually correct, we want a list of tensors, not a single tensor: https://github.com/kornia/kornia/pull/1497

Note that the name is wrong, we should be used bbox_xyxy or bbox_xywh instead of bbox (boxes is translated to bbox internally in our `AugmentationSequential` wrapper) but we can fix that when working on https://github.com/microsoft/torchgeo/discussions/985.

adamjstewart: Okay, see if the last commit fixes this.

calebrob6 (approved):

PR #999, opened 2023-01-04
Convert all index transforms to Kornia

adamjstewart:
This PR converts all of our index transforms to be valid Kornia augmentations. These transforms can now be used with AugmentationSequential just like all of the other transforms we use.


adamjstewart:
Can't get the dimensions to work in the transforms tutorial, please help.

nilsleh:
> Can't get the dimensions to work in the transforms tutorial, please help.

The only thing I can get to work is to match the shape of the input. 

```
class MinMaxNormalize(K.IntensityAugmentationBase2D):
    """Normalize channels to the range [0, 1] using min/max values."""

    def __init__(self, mins: Tensor, maxs: Tensor) -> None:
        super().__init__(p=1)
        self.flags = {
            "mins": mins.view(1, mins.shape[0], 1, 1), 
            "maxs": maxs.view(1, maxs.shape[0], 1, 1)
        }

    def apply_transform(
        self,
        input: Tensor,
        params: Dict[str, Tensor],
        flags: Dict[str, int],
        transform: Optional[Tensor] = None,
) -> Tensor:
        return (input - flags["mins"]) / (flags["maxs"] - flags["mins"] + 1e-10)
```

torchgeo/datasets/eurosat.py
```diff
@@ -170,7 +170,7 @@ def __getitem__(self, index: int) -> Dict[str, Tensor]:
         """
         image, label = self._load_image(index)
 
-        image = torch.index_select(image, dim=0, index=self.band_indices)
+        image = torch.index_select(image, dim=0, index=self.band_indices).float()
```
calebrob6, line 173: already done in the datamodules/gpu branch

torchgeo/transforms/indices.py
```diff
@@ -8,16 +8,16 @@
...
 import torch
+from kornia.augmentation import IntensityAugmentationBase2D
 from torch import Tensor
-from torch.nn.modules import Module
 
 _EPSILON = 1e-10
 
 
-class AppendNormalizedDifferenceIndex(Module):
+class AppendNormalizedDifferenceIndex(IntensityAugmentationBase2D):
```
calebrob6, line 20: What is this super class?

calebrob6: Nevermind https://kornia.readthedocs.io/en/latest/augmentation.base.html#kornia.augmentation.IntensityAugmentationBase2D

calebrob6:
Heads up that running `transforms.ipynb` breaks because the SSL cert for the EuroSat dataset expired.

calebrob6 (approved):

calebrob6 (approved):

PR #493, opened 2022-04-03
Indices tutorial: better normalization for true color plot

adamjstewart:
This PR changes the normalization strategy for a Sentinel-2 true color image from "0.1th to 99.9th percentile" to "divide by 2000". This is in line with how our `Sentinel2.plot(...)` function works, although the scale factor is lower.

### Before

before

### After

after

It also fixes a couple of other bugs:

* the normalize parameter can be any float, not just an int
* URLs should be split by "/", not by os.sep

PR #1025, opened 2023-01-20
Correction regarding scaling factor for Sentinel-2 image

osgeokr:
My suggestion is to use 10,000 instead of 6,000 as scaling factor for Sentinel-2 image.
DN value of Sentinel-2 can be converted to Reflectance dividing by 10,000(scaling factor).
DN = 10000 * REFLECTANCE
https://docs.sentinel-hub.com/api/latest/data/sentinel-2-l2a/

Sorry in advance if the number 6,000 was used in a different way. :)


adamjstewart:
Thanks, can you also change the`plot` method in `torchgeo/datasets/sentinel.py`? Can you also add a link to https://docs.sentinel-hub.com/api/latest/data/sentinel-2-l2a/ as a comment in `torchgeo/datasets/sentinel.py`?

The last time we changed this was in #493. We lowered the number because the image was too dark and I couldn't find any official recommendation.

> For Sentinel-2 optical data, the relation between `DN` and `REFLECTANCE` (default unit) is: `DN = 10000 * REFLECTANCE`

Is `REFLECTANCE` on a 0–1 scale? I wonder if that's the best scale to use, since the maximum theoretical reflectance is quite high. I want images without clouds to look good, I'm not worried about oversaturation of clouds.

osgeokr:
Surface Reflectance values lies usually between 0.0 and 1.0. However, Specular effects on surface or clouds could lead to values higher than 1.0.
Pls refer to this document(p.40.): https://sentinel.esa.int/documents/247904/685211/Sentinel-2-MSI-L2A-Product-Format-Specifications.pdf

In my opinion, it would be better to set a scale factor(10,000) to convert the DN to reflectance, and a separate brightness factor to adjust the brightness value of the image(e.g. bright * image / 10,000).

osgeokr:
I will close this pull request for now. After fixing torchgeo/datasets/sentinel.py, I'll re-upload a pull request for this issue. Thank you.

PR #1029, opened 2023-01-23
Typo fix in pretrained weights tutorial

calebrob6:


adamjstewart (approved):

calebrob6:
Found another problem in the tutorial that will require more changes

torchgeo/trainers/utils.py
```diff
@@ -72,7 +72,9 @@ def load_state_dict(model: Module, state_dict: "OrderedDict[str, Tensor]") -> Mo
     expected_in_channels = state_dict["conv1.weight"].shape[1]
 
     num_classes = cast(nn.Module, model.fc).out_features
-    expected_num_classes = state_dict["fc.weight"].shape[0]
+    expected_num_classes = None
+    if "fc.weight" in state_dict:
```
calebrob6, line 76: "fc.weight" is not in (at least) the `ResNet50_Weights.SENTINEL2_ALL_MOCO` set of weights

docs/tutorials/pretrained_weights.ipynb
```diff
@@ -168,7 +168,7 @@
    "source": [
     "in_chans = weights.meta[\"in_chans\"]\n",
     "model = timm.create_model(\"resnet50\", in_chans=in_chans, num_classes=10)\n",
-    "model.load_state_dict(weights.get_state_dict(), strict=False)"
+    "model.load_state_dict(weights.get_state_dict(True), strict=False)"
```
adamjstewart, line 171:
```suggestion
    "model.load_state_dict(weights.get_state_dict(progress=True), strict=False)"
```

torchgeo/trainers/utils.py
```diff
@@ -81,15 +83,16 @@ def load_state_dict(model: Module, state_dict: "OrderedDict[str, Tensor]") -> Mo
         )
         del state_dict["conv1.weight"]
 
-    if num_classes != expected_num_classes:
+    if expected_num_classes is None:
+        warnings.warn("pretrained model does not include class weights")
```
adamjstewart, line 87: Is this worth warning about? Most (all?) of our pretrained models lack the last layer since it will need to be replaced anyway.

calebrob6: That's fine

torchgeo/trainers/utils.py
```diff
@@ -89,7 +91,6 @@ def load_state_dict(model: Module, state_dict: "OrderedDict[str, Tensor]") -> Mo
         del state_dict["fc.weight"], state_dict["fc.bias"]
```
adamjstewart, line 91: Should we say `if expected_num_classes and num_classes != expected_num_classes` to avoid warning when there is no layer?

calebrob6: yep

adamjstewart (approved):

PR #1031, opened 2023-01-23
Fix ViT weight names in docs

adamjstewart:
Wish there was a way to link to the enum itself so we can enforce that they are named correctly, but it doesn't seem to show up in the docs.


calebrob6 (approved):
good catch

PR #1033, opened 2023-01-23
Update image plots in custom raster dataset tutorial

adamjstewart:
As usual, the diff is useless, but you can view the new docs from the CI link. Wish we didn't have to use Jupyter for this.


ashnair1:
[kornia/tutorials](https://github.com/kornia/tutorials) tutorials uses [Review-Notebook](https://github.com/apps/review-notebook-app) to view notebook diffs. Might be worth checking out.

adamjstewart:
My bigger problem is that I'm a vim user and I like to be able to edit my code in vim. And manually editing a notebook with vim is hell. There are tools that automatically convert Python scripts to notebooks with correct formatting, but we would need to actually run the notebooks during CI on each commit and that would be super slow. There also aren't GPUs in CI so some of the results of transform timing would not work.

calebrob6 (approved):

PR #1036, opened 2023-01-23
0.4.0 release

adamjstewart:
#931

### DIFF
--- a/docs/tutorials/trainers.ipynb
+++ b/docs/tutorials/trainers.ipynb
@@ -132,7 +132,7 @@
     "data_dir = os.path.join(tempfile.gettempdir(), \"cyclone_data\")\n",
     "\n",
     "datamodule = TropicalCycloneDataModule(\n",
-    "    root=data_dir, batch_size=64, num_workers=6, api_key=MLHUB_API_KEY\n",
+    "    root=data_dir, batch_size=64, num_workers=6, download=True, api_key=MLHUB_API_KEY\n",
     ")"
    ]
   },
--- a/docs/tutorials/trainers.ipynb
+++ b/docs/tutorials/trainers.ipynb
@@ -161,7 +161,10 @@
    "source": [
     "task = RegressionTask(\n",
     "    model=\"resnet18\",\n",
+    "    weights=None,\n",
     "    pretrained=True,\n",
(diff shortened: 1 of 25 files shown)


setup.cfg
```diff
@@ -42,7 +42,7 @@ install_requires =
     # pyproj 2.2+ required for CRS object
     pyproj>=2.2,<4
     # pytorch-lightning 1.5.1+ required for apply_to_collection bugfix
-    pytorch-lightning>=1.5.1,<2
+    pytorch-lightning[extra]>=1.5.1,<2
```
adamjstewart, line 45: This is painful. Lightning dropped tensorboardX as a required dep. TorchGeo technically doesn't require tensorboardX, but train.py does. Some of our trainers may or may not be broken when plotting if tensorboard isn't installed. So technically we could just add a dep on tensorboard, even though we don't import it. The alternative is to depend on `pytorch-lightning[extra]` which adds tensorboardX and a bunch of other deps we don't need. I don't love either solution.

docs/tutorials/trainers.ipynb
```diff
@@ -161,7 +161,10 @@
    "source": [
     "task = RegressionTask(\n",
     "    model=\"resnet18\",\n",
+    "    weights=None,\n",
     "    pretrained=True,\n",
```
adamjstewart, line 165: Should we also drop pretrained=True?

nilsleh: I think so yes. There was a discussion somewhere of weights being one of `"imagenet", None or WeightEnum`? So is that what you had in mind?

adamjstewart: See https://torchgeo.readthedocs.io/en/latest/api/trainers.html#torchgeo.trainers.RegressionTask:

> **weights** – Either a weight enum, the string representation of a weight enum, True for ImageNet weights, False or None for random weights, or the path to a saved model state dict.

adamjstewart:
Indices tutorial works fine locally, remaining failures are all due to https://github.com/phelber/EuroSAT/issues/10. Not sure whether we should try to change the dataset or merge as is.

calebrob6 (approved):

PR #1037, opened 2023-01-24
Augs not applied during testing

ashnair1:
Missing test_aug causes failure during training.


ashnair1:
Alternatively since `self.val_aug` , `self.test_aug` and `self.predict_aug` are the sam, we could replace them with a singular `self.aug`  like in other datamodules.

adamjstewart (changes requested):
Can you push this to the releases/v0.4 branch so this makes it into the release?

adamjstewart:
Added this to the release branch: https://github.com/microsoft/torchgeo/pull/1036/commits/671737fdbc3c24557019fde4973ed3bd2794d380

PR #1039, opened 2023-01-24
Augs not applied during testing

ashnair1:
See #1037


adamjstewart (approved):

adamjstewart:
The tests prob won't pass until we fix everything else on the release branch. You can just add a new commit to the branch directly without a PR if that's quicker. The releases branch is merged into main when it's ready so no need for a separate PR to main.

adamjstewart:
Added this to the release branch: https://github.com/microsoft/torchgeo/pull/1036/commits/671737fdbc3c24557019fde4973ed3bd2794d380

Issue #1044, opened 2023-01-25
Semantic Segmentation/Object Detection: Add support for pretrained `weights`


adamjstewart:
### Summary

We should add support for using pretrained backbones to our semantic segmentation and object detection trainers.

### Rationale

Our classification/regression/BYOL trainers support pretrained weights, we should be uniform.

### Implementation

Need to figure out if any of these models support the use of timm for backbones or whether or not we can manually replace the backbone.

### Alternatives

We currently support ImageNet pre-trained backbones for a limited subset, but that isn't the same.

adamjstewart:
Update: segmentation weights are now supported but not detection weights.

PR #1046, opened 2023-01-25
Segmentation Pretrained Weights

isaaccorley:
This PR addresses part of #1044 adds the ability to load pretrained weights from a backbone model e.g. ResNet into a semantic segmentation encoder. This works for the segmentation-models-pytorch Unet and DeepLabv3 implementations but not the FCN because we aren't using a backbone for that.


adamjstewart:
This only supports ImageNet pretrained weights, which is pretty useless for us. We really want support for our weight enums like we have for regression/classification/byol. Could we manually load weights from state dict ourselves like we do with our timm-based models?

adamjstewart:
Oh wait, I didn't read the full PR. Maybe this does cover all the features we want. You'll definitely want to update the docstring though.

adamjstewart:
Also needs tests

nilsleh:
Maybe also put a note in the docs that not all our pretrained models (i.e. the ViTs) will be compatible according to the [smp docs](https://github.com/qubvel/segmentation_models.pytorch#timm-encoders-) or maybe there is a way around, not sure.

isaaccorley:
The other thing I realized is that by default smp will use torchvision for resnet backbones e.g. `resnet18`. If you want to force using timm implementations you have to prefix the encoder names with `tu-` e.g. `tu-resnet18`.

calebrob6:
We could just only allow timm pretrained backbones here

isaaccorley:
This actually needs #1049 to be solved otherwise this only supports resnet backbones.

isaaccorley:
> We could just only allow timm pretrained backbones here

So do we want to prefix `tu-` to all the backbones or do we want the user to do it?

adamjstewart:
That is a tough question. I would be fine with only supporting timm backbones for now, which would suggest prepending `tu-` for the user so that it matches other trainers. But there may come a time in which we want to support non-timm backbones because pretrained models are not always made with timm.

calebrob6 (approved):

calebrob6:
We could prefix "tu-" if it results in a valid timm backbone to give ourselves room to grow.

torchgeo/trainers/segmentation.py
```diff
@@ -80,13 +85,23 @@ def config_task(self) -> None:
...
+                self.model.encoder = utils.load_state_dict(self.model, state_dict)
+
     def __init__(self, **kwargs: Any) -> None:
         """Initialize the LightningModule with a model and loss function.
 
         Keyword Args:
             model: Name of the segmentation model type to use
             backbone: Name of the timm backbone to use
-            weights: None or "imagenet" to use imagenet pretrained weights in
+            weights: None or True to use imagenet pretrained weights in
```
calebrob6, line 104: Probably need to say this can be a string and stuff

calebrob6:
Ugh, also the horrific model checkpoint tests

calebrob6:
We need this :)

calebrob6:
@isaaccorley does this work with our custom ResNet weights?

isaaccorley:
I haven't tried using the weights enums but in theory it should since the encoders are just resnet models.

adamjstewart:
It should be the same loading code and docstring description as all of the other trainers.

isaaccorley:
I just made it work with the ResNet weights

tests/trainers/test_segmentation.py
```diff
@@ -111,6 +122,50 @@ def model_kwargs(self) -> dict[Any, Any]:
             "ignore_index": 0,
         }
 
+    @pytest.fixture(
+        params=[
+            weights
+            for model in list_models()
+            for weights in get_model_weights(model)
+            if "resnet" in weights.meta["model"]
```
isaaccorley, line 130: SMP doesn't support ViT weights for the encoder yet (only MiT)

torchgeo/trainers/segmentation.py
```diff
@@ -100,8 +119,9 @@ def __init__(self, **kwargs: Any) -> None:
         Keyword Args:
             model: Name of the segmentation model type to use
             backbone: Name of the timm backbone to use
-            weights: None or "imagenet" to use imagenet pretrained weights in
-                the backbone
+            weights: Either a weight enum, the string representation of a weight enum,
+                True for ImageNet weights, False or None for random weights,
+                or the path to a saved model state dict.
```
adamjstewart, line 124: Need versionchanged and need to document that FCN does not support pretrained models.

calebrob6 (dismissed):

tests/conf/inria.yaml
```diff
@@ -3,7 +3,7 @@ module:
   loss: "ce"
   model: "unet"
   backbone: "resnet18"
-  weights: "imagenet"
+  weights: true
```
adamjstewart, line 6: Does this download from the internet? We want to avoid that

tests/trainers/test_segmentation.py
```diff
@@ -111,6 +122,53 @@ def model_kwargs(self) -> dict[Any, Any]:
...
+        model_kwargs["weights"] = mocked_weights
+        SemanticSegmentationTask(**model_kwargs)
+
+    def test_weight_str(
+        self, model_kwargs: dict[str, Any], mocked_weights: WeightsEnum
+    ) -> None:
+        model_kwargs["backbone"] = mocked_weights.meta["model"]
+        model_kwargs["in_channels"] = mocked_weights.meta["in_chans"]
+        model_kwargs["weights"] = str(mocked_weights)
+        SemanticSegmentationTask(**model_kwargs)
```
adamjstewart, line 170: Can you add tests that test the real weights? See all other trainer tests for examples. It's the things marked as slow that only run during the release and are allowed to access the internet to test real world integration.

torchgeo/trainers/segmentation.py
```diff
@@ -127,6 +144,9 @@ class and used with 'ce' loss
             The *class_weights*, *freeze_backbone*,
             and *freeze_decoder* parameters.
 
+        .. versionchanged:: 0.5
+           The *weights* parameter supports WeightEnums and checkpoint paths.
```
adamjstewart, line 148:
```suggestion
           The *weights* parameter now supports WeightEnums and checkpoint paths.
```

adamjstewart (changes requested):

adamjstewart (approved):

PR #1047, opened 2023-01-26
Adding SAR utilities and functions

RitwikGupta:
The goal of this PR is to add a set of comprehensive set of SAR utilities to TorchGeo.

- [x] Add the ability to convert In-phase/Quadrature representation files to Phase/Magnitude representation.
- [x] Add multi-look features
- [ ] Debursting, sub-swatch merging, etc.
- [ ] Coherence and interferogram creation
- [ ] Quad and dual-pol decompositions
- [ ] The ability to parse STANAG 4607?

### DIFF
--- a/torchgeo/transforms/sar.py
+++ b/torchgeo/transforms/sar.py
@@ -1,15 +1,28 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-from typing import Any, Dict, List, Optional, Tuple, Union
+"""SAR utilities."""
+
+import math
 
 import torch
+import torch.nn.functional as F
 from torch import Tensor
-
-_EPSILON = 1e-10
 
 
 def inphase_quadrature_to_phase_magnitude(src: Tensor) -> Tensor:
+    """Convert a SAR image in in-phase/quadrature format to phase/magnitude.
+
+    Magnitude is in power scale.
+
+    .. versionadded:: 0.5
+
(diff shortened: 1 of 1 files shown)


adamjstewart:
I agree that these transforms should be stored in `torchgeo/transforms/sar.py`. Some other comments:

* We should make these compatible with Kornia. Unfortunately, documentation on how to write Kornia-style transforms is pretty sparse. For transforms that should only affect the image (as opposed to masks, bboxes, labels, etc.), you can follow the same style as I used for the transforms in `torchgeo/transforms/indices.py`. If you also need to modify masks/bboxes/etc., there's a different base class and some additional functions that need to be overridden.
* These should be added to `torchgeo/transforms/__init__.py`
* We may want to override the default docs in `docs/api/transforms.rst` to group index and SAR transforms separately, we can do this later

RitwikGupta:
I'd want to store these transforms in `torchgeo/transforms/sar/representation.py`, `torchgeo/transforms/sar/multilook.py`, etc. Thoughts?

adamjstewart:
Depends on how many transforms there are and how dissimilar they are. I don't really want to have 4 levels of nesting if we don't have to. Of course, everything will be imported in `torchgeo.transforms` so this won't make a difference for users.

adamjstewart:
@RitwikGupta is this PR still a work in progress? It's fine to merge partial SAR support, we don't need everything to be complete all at once.

PR #1058, opened 2023-01-28
Drop Python 3.7 support

adamjstewart:
### Why?

* Python 3.7 will reach [EOL](https://devguide.python.org/versions/) in June, likely before our next release.
* [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) dropped Python 3.7 support a long time ago
* [PyTorch](https://github.com/pytorch/pytorch/pull/93155) dropped Python 3.7 support
* Greatly simplifies our CI config files
* Fewer CPU cycles wasted testing ancient Python versions, better for the planet
* Google Colab and Planetary Computer both upgraded to Python 3.8
* Only new feature of interest to us in [Python 3.8](https://docs.python.org/3/whatsnew/3.8.html) is TypedDict (see #985)

### What?

This PR drops Python 3.7 support and updates all of our dependencies to the oldest version with Python 3.8 wheels. We could theoretically support older versions without wheels by compiling from source, but this slows down CI and I don't see a big benefit to supporting ancient versions of our dependencies. If anyone needs support for an older dependency version, let me know and I can change this.

See #656 for further discussion.


calebrob6 (approved):
🔥

PR #1062, opened 2023-01-29
Constrain dependencies to avoid CI hang

adamjstewart:
See https://github.com/pypa/pip/issues/11760


calebrob6:
Good grief that's ridiculous

calebrob6 (approved):

adamjstewart:
> Good grief that's ridiculous -- any chance we can just get rid of some dependencies?

I could have sworn that I wrote a discussion listing all of our deps, all of the places where we use them, and recommendations for which to make required and which to drop, but now I can't find it.

PR #1063, opened 2023-01-29
Remove scikit-learn

calebrob6:
We only use sklearn for GroupShuffleSplit. Reimplementing our own version to remove this dependency.

### DIFF
--- a/torchgeo/datamodules/utils.py
+++ b/torchgeo/datamodules/utils.py
@@ -4,7 +4,7 @@
 """Common datamodule utilities."""
 
 import math
-from typing import Any, Optional, Union
+from typing import Any, Optional, Union, Iterable
 
 import numpy as np
 from torch import Generator
@@ -57,11 +57,11 @@ def dataset_split(
 
 
 def group_shuffle_split(
-    groups: Any,
+    groups: Iterable[Any],
     train_size: Optional[float] = None,
     test_size: Optional[float] = None,
     random_state: Optional[int] = None,
-) -> tuple[Any, Any]:
+) -> tuple[np.ndarray, np.ndarray]:
     """Method for performing a single group-wise shuffle split of data.
 
     Loosely based off of `sklearn.model_selection.GroupShuffleSplit`.
--- a/torchgeo/datamodules/utils.py
+++ b/torchgeo/datamodules/utils.py
@@ -61,7 +61,7 @@ def group_shuffle_split(
     train_size: Optional[float] = None,
     test_size: Optional[float] = None,
     random_state: Optional[int] = None,
-) -> Tuple[Any, Any]:
+) -> tuple[Any, Any]:
     """Method for performing a single group-wise shuffle split of data.
 
     Loosely based off of `sklearn.model_selection.GroupShuffleSplit`.
--- a/torchgeo/datamodules/utils.py
+++ b/torchgeo/datamodules/utils.py
@@ -4,7 +4,8 @@
 """Common datamodule utilities."""
 
 import math
-from typing import Any, Optional, Union, Iterable
+from typing import Any, Optional, Union
+from collections.abc import Iterable
 
 import numpy as np
 from torch import Generator
--- a/torchgeo/datamodules/utils.py
+++ b/torchgeo/datamodules/utils.py
@@ -103,7 +103,7 @@ def group_shuffle_split(
 
     group_vals = set(groups)
     n_groups = len(group_vals)
-    n_test_groups = math.ceil(n_groups * test_size)
+    n_test_groups = round(n_groups * test_size)
     n_train_groups = n_groups - n_test_groups
 
     if n_train_groups == 0 or n_test_groups == 0:
@@ -113,9 +113,9 @@ def group_shuffle_split(
         )
 
     generator = np.random.default_rng(seed=random_state)
-    train_group_vals = generator.choice(
+    train_group_vals = set(generator.choice(
         list(group_vals), size=n_train_groups, replace=False
-    )
+    ))
 
     train_idxs = []
     test_idxs = []
--- a/torchgeo/datamodules/utils.py
+++ b/torchgeo/datamodules/utils.py
@@ -62,7 +62,7 @@ def group_shuffle_split(
     train_size: Optional[float] = None,
     test_size: Optional[float] = None,
     random_state: Optional[int] = None,
-) -> tuple["np.typing.NDArray[np.int_]", "np.typing.NDArray[np.int_]"]:
+) -> tuple[list[int], list[int]]:
     """Method for performing a single group-wise shuffle split of data.
 
     Loosely based off of `sklearn.model_selection.GroupShuffleSplit`.
@@ -126,4 +126,4 @@ def group_shuffle_split(
         else:
             test_idxs.append(i)
 
-    return np.array(train_idxs), np.array(test_idxs)
+    return train_idxs, test_idxs
--- a/torchgeo/datamodules/utils.py
+++ b/torchgeo/datamodules/utils.py
@@ -65,7 +65,7 @@ def group_shuffle_split(
 ) -> tuple[list[int], list[int]]:
     """Method for performing a single group-wise shuffle split of data.
 
-    Loosely based off of `sklearn.model_selection.GroupShuffleSplit`.
+    Loosely based off of :class:`sklearn.model_selection.GroupShuffleSplit`.
 
     Args:
         groups: a sequence of group values used to split. Should be in the same order as
--- a/torchgeo/datamodules/utils.py
+++ b/torchgeo/datamodules/utils.py
@@ -62,7 +62,7 @@ def group_shuffle_split(
     train_size: Optional[float] = None,
     test_size: Optional[float] = None,
     random_state: Optional[int] = None,
-) -> tuple[np.ndarray, np.ndarray]:
+) -> tuple[np.ndarray[np.int32], np.ndarray[np.int32]]:
     """Method for performing a single group-wise shuffle split of data.
 
     Loosely based off of `sklearn.model_selection.GroupShuffleSplit`.
--- a/torchgeo/datamodules/utils.py
+++ b/torchgeo/datamodules/utils.py
@@ -58,7 +58,7 @@ def dataset_split(
 
 
 def group_shuffle_split(
-    groups: Iterable[object],
+    groups: Iterable[Any],
     train_size: Optional[float] = None,
     test_size: Optional[float] = None,
     random_state: Optional[int] = None,
--- a/torchgeo/datamodules/utils.py
+++ b/torchgeo/datamodules/utils.py
@@ -58,7 +58,7 @@ def dataset_split(
 
 
 def group_shuffle_split(
-    groups: Iterable[Any],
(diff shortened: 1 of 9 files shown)


torchgeo/datamodules/utils.py
```diff
@@ -52,3 +54,75 @@ def dataset_split(
             [train_length, val_length, test_length],
             generator=Generator().manual_seed(0),
         )
+
+
+def group_shuffle_split(
+    groups: Any,
```
calebrob6, line 60: Not sure what the typing should be on this or the return values

adamjstewart:
Trying to decide if the +100 lines is worth it to remove one dependency...

calebrob6:
Which is more likely to cause issues? (I'm thinking sklearn obviously -- more generally, I'm  tired of dependencies breaking everything, it seems we spend as much time fiddling with CI and dependencies as we do on actual features)

Another relevant question is "do we see any other features of sklearn that we'll be using in the future?" (I'm thinking maybe if we implement fine tuning and want to use sklearn, but that isn't necessary)

calebrob6:
Can you give some arguments for/against?

adamjstewart:
### Pros

* -1 required _direct_ dependency
* -11 required _indirect_ dependencies (including scipy which was listed as optional)
* Fewer dependabot PRs
* Faster installs during tests

+/- sklearn isn't a _huge_ deal just because most ML people will already have it installed anyway. But it's always nice to decrease the number of our deps, both for install times and for simpler solves. For example, [sklearn requires setuptools < 60](https://github.com/scikit-learn/scikit-learn/blob/1.2.1/pyproject.toml#L4), while [fiona requires setuptools 61+](https://github.com/Toblerity/Fiona/blob/1.9.1/pyproject.toml#L5). Pip can handle this, but Spack can't, meaning you would have to choose between the latest version of sklearn or fiona, you couldn't have both. This was [fixed the other day](https://github.com/scikit-learn/scikit-learn/pull/25651), but still.

### Cons

* +100 lines of code we have to maintain
* Possibility of sklearn implementation getting bug fixes we don't
* Possibility of sklearn implementation changing API someday
* Possibility we may decide to re-add sklearn for another feature someday (

Maintenance burden is my biggest fear. We aren't explicitly making this function public (it doesn't get an import alias), but it's still something people could potentially try to use. If we do keep it, we should prob prefix with an underscore to avoid people relying on it.

### Alternatives

* Could move sklearn to `[datasets]` and make it optional
* Could simplify the function to the absolute bare minimum feature with no error checks, but that might lose compatibility with the sklearn implementation, making it difficult to revert someday if we decide to use sklearn for something else

I'm really on the fence with this one, not sure how to decide. Curious how you feel about alternative 1 (moving to `[datasets]`). Most of the pros go away, but at least it's optional. Technically it's for datamodules, not datasets, but close enough.

calebrob6:
I don't really like the idea of moving sklearn to `[datasets]` as the whole point of this was to reduce dependency related maintenance burden.

If we divide current lines of code by current number dependencies we can get an idea of how many lines of code are worth one dependency. This number will be assuredly by larger than 100 (if it isn't I'm sure I can code golf the current implementation of group splitting...). Put differently, would you take on a new dependency _just_ to get rid of 100 lines of code? (I think no way in hell 🙂)

For the "Possibility we may decide to re-add sklearn for another feature someday" -- that doesn't seem like a con at all. If we do, then we can drop these 100 lines of code, else it is a non-issue.

The function that we are maintaining has pretty clear logic:
- Takes of list of something (hashables maybe?), `groups`
- Splits the input list into two (according to a given percentage) while ensuring that the values in `groups` assigned to both splits are distinct.

We don't really care if sklearn changes their API for doing this because we just care about the functionality.

adamjstewart:
I was 50/50 on this but now I'm more like 60/40 in favor. Still approaching critical mass...

calebrob6:
> I was 50/50 on this but now I'm more like 60/40 in favor. Still approaching critical mass...

Anything actionable?

adamjstewart:
Not yet, still mulling...

adamjstewart:
Yeah let's do this. We're adding a bunch more new deps now, so it would be good to reduce. Can you rebase?

torchgeo/datamodules/utils.py
adamjstewart: Seems like input can be any Sequence, and output is currently np.ndarray

torchgeo/datamodules/utils.py
```diff
@@ -52,3 +54,75 @@ def dataset_split(
...
+
+    Loosely based off of `sklearn.model_selection.GroupShuffleSplit`.
+
+    Args:
+        groups: a sequence of group values used to split. Should be in the same order as
+            the data you want to split.
+        train_size: the proportion of groups to include in the train split. If None,
+            then it is set to complement `test_size`.
+        test_size: the proportion of groups to include in the test split (rounded up).
+            If None, then it is set to complement `train_size`.
```
adamjstewart, line 75: Is there any reason to have both params if train_size + test_size = 1?

torchgeo/datamodules/utils.py
```diff
@@ -52,3 +54,75 @@ def dataset_split(
...
+        random_state: controls the random splits (passed a seed to a
+            numpy.random.Generator), set for reproducible splits.
```
adamjstewart, line 77: Should we call this `generator` and use a `torch.Generator` object as input to match `torch.utils.data.random_split`?

torchgeo/datamodules/utils.py
```diff
@@ -52,3 +54,75 @@ def dataset_split(
...
+
+    if train_size is None and test_size is not None:
+        train_size = 1 - test_size
+    if test_size is None and train_size is not None:
+        test_size = 1 - train_size
+
+    assert train_size is not None and test_size is not None
+
+    if train_size <= 0 or train_size >= 1 or test_size <= 0 or test_size >= 1:
+        raise ValueError("`train_size` and `test_size` must be in the range (0,1).")
```
adamjstewart, line 102: `torch.utils.data.random_split` and `torchgeo.datasets.splits` support both fractions and lengths

torchgeo/datamodules/utils.py
calebrob6: so you can specify either train size or test size

adamjstewart: Could instead support a list of lengths/fractions to match `torch.utils.data.random_split` and `torchgeo.datasets.splits`. This would support train/val/test splits in addition to train/test splits.

calebrob6:
I'm not trying to mimic torch.utils or torchgo.datasets.splits here. This is meant to be a simple drop in replacement that doesn't depend on all of scikit-learn

torchgeo/datamodules/utils.py
calebrob6: torchgeo.datasets.split doesn't support lengths -- https://github.com/microsoft/torchgeo/blob/main/torchgeo/datasets/utils.py#L397

adamjstewart: That's just a utility used by functions in `torchgeo.datasets.splits`, it isn't even exported to the user.

adamjstewart:
I'm talking about `torchgeo/datasets/splits.py`.

If we're going to write our own splitting utility, I don't see why we shouldn't follow the PyTorch style instead of the sklearn style. Could even put it in `torchgeo/datasets/splits.py` and export it to the user.

calebrob6:
Okay, I see what you're saying now! Sorry that took me a minute -- the disconnect was because I wasn't seeing this as something that operated on torch Datasets. It _could_ go in `torchgeo/datasets/splits.py` but it isn't really specific to `Datasets`.

adamjstewart:
That's fair. I'm fine with keeping this internal-only and using a different style since it doesn't yet support passing in a NonGeoDataset. But if we can figure out a stable API that would support that it would be pretty cool.

Want me to merge this as is and save this for another day?

torchgeo/datamodules/utils.py
```diff
@@ -52,3 +54,75 @@ def dataset_split(
...
+
+    Raises:
+        ValueError if `train_size` and `test_size` do not sum to 1, aren't in the range
+            (0,1), or are both None.
+        ValueError if the number of training or testing groups turns out to be 0.
+    """
+    if train_size is None and test_size is None:
+        raise ValueError("You must specify `train_size`, `test_size`, or both.")
+    if (train_size is not None and test_size is not None) and (
+        train_size + test_size != 1.0
```
adamjstewart, line 90:
```suggestion
        not math.isclose(train_size + test_size, 1)
```
Never compare float equality

torchgeo/datamodules/utils.py
```diff
@@ -52,3 +54,75 @@ def dataset_split(
...
+        test_size = 1 - train_size
+
+    assert train_size is not None and test_size is not None
+
+    if train_size <= 0 or train_size >= 1 or test_size <= 0 or test_size >= 1:
+        raise ValueError("`train_size` and `test_size` must be in the range (0,1).")
+
+    group_vals = set(groups)
+    n_groups = len(group_vals)
+    n_test_groups = math.ceil(n_groups * test_size)
```
adamjstewart, line 106: We're using `math.round` everywhere else in TorchGeo, except `torchgeo.datasets.splits` where we follow PyTorch and use floor with round-robin assignment of the remaining indices.

torchgeo/datamodules/utils.py
```diff
@@ -52,3 +54,75 @@ def dataset_split(
...
+        list(group_vals), size=n_train_groups, replace=False
+    )
+
+    train_idxs = []
+    test_idxs = []
+    for i, group_val in enumerate(groups):
+        if group_val in train_group_vals:
+            train_idxs.append(i)
+        else:
+            test_idxs.append(i)
```
adamjstewart, line 126: Feels like there should be a faster way to do this. Could use pandas groupby, although that's an optional dep. If the input is an np.ndarray, couldn't we use array indexing to grab all entries of a certain value?

adamjstewart (changes requested):

torchgeo/datamodules/utils.py
calebrob6: Changed to use set lookup in case someone has a ton of group vals. This will be O(n log m) where n is the input size and m is the number of unique groups in the train split. As this will not be called with any high frequency I don't think it is worth the effort of optimizing any.

calebrob6:
Tests fail with `round(...)` I expect the behavior guaranteed by ceiling.

adamjstewart:
> I expect the behavior guaranteed by ceiling.

I would expect it to behave similarly to our other splitting functions.

torchgeo/datamodules/utils.py
```diff
@@ -52,3 +55,75 @@ def dataset_split(
             [train_length, val_length, test_length],
             generator=Generator().manual_seed(0),
         )
+
+
+def group_shuffle_split(
+    groups: Iterable[Any],
```
adamjstewart, line 61:
```suggestion
    groups: Iterable[object],
```
Any means disable type checking, object means must support all possible input types. See if this passes. If not, revert it.

calebrob6:
can you finish this?

adamjstewart (approved):

PR #1065, opened 2023-01-30
Fix SSL issue with EuroSAT

### LINKED ISSUE #1064: EuroSAT Dataset URL not being secure anymore
### Description

As `url = "https://madm.dfki.de/files/sentinel/EuroSATallBands.zip"` is not secure anymore, downloading the EuroSat dataset is blocked by default by certify.

We might need to either change the zip source or remove the `https`

### Steps to reproduce

Load the EuroSAT dataset
```python
from torchgeo.datasets import EuroSAT
eurosat_train = EuroSAT(download=True)
```

### Version

0.4.0

julien-blanchon:
Change https to http as the source SSL certificate is not valid anymore

Fixes #1064

### DIFF
--- a/torchgeo/datasets/eurosat.py
+++ b/torchgeo/datasets/eurosat.py
@@ -51,7 +51,8 @@ class EuroSAT(NonGeoClassificationDataset):
     * https://ieeexplore.ieee.org/document/8519248
     """
 
-    # TODO: Change to https after https://github.com/phelber/EuroSAT/issues/10 is resolved
+    # TODO: Change to https after https://github.com/phelber/EuroSAT/issues/10 is
+    # resolved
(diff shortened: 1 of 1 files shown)


calebrob6:
@adamjstewart any reason why we shouldn't switch to http until EuroSAT figures their cert out?

adamjstewart:
Not that I can think of. Just to be safe, let's change checksum to default to True until https is working again.

calebrob6 (approved):

calebrob6 (approved):

adamjstewart (approved):

adamjstewart:
Confirmed that the MD5 hasn't changed.

PR #1082, opened 2023-02-02
Add VHR10 datamodule

ashnair1:


conf/vhr10.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+experiment:
+  task: "vhr10"
+  name: "vhr10_test"
+  module:
+    model: "faster-rcnn"
+    backbone: "resnet18"
+    pretrained: True
+    num_classes: 11
+    learning_rate: 1.3e-5
```
calebrob6, line 21: 👀

torchgeo/datamodules/vhr10.py
```diff
@@ -0,0 +1,110 @@
...
+        num_workers: int = 0,
+        val_split_pct: float = 0.2,
+        test_split_pct: float = 0.2,
+        **kwargs: Any,
+    ) -> None:
+        """Initialize a new VHR10DataModule instance.
+
+        Args:
+            batch_size: Size of each mini-batch.
+            patch_size: Size of each patch, either ``size`` or ``(height, width)``.
```
calebrob6, line 55: patch_size only takes an int in the constructor definition

torchgeo/datamodules/vhr10.py
```diff
@@ -0,0 +1,110 @@
...
+                :class:`~torchgeo.datasets.VHR10`.
+        """
+        super().__init__(VHR10, batch_size, num_workers, **kwargs)
+
+        self.val_split_pct = val_split_pct
+        self.test_split_pct = test_split_pct
+        self.patch_size = _to_tuple(patch_size)
+
+        self.collate_fn = collate_fn
+        self.kwargs["transforms"] = self.preprocess
```
adamjstewart, line 69: Using instance methods for preprocessing in a parallel data loader is a big no no, see #886 for what happens when you do this. If possible, you should use AugmentationSequential with a series of Kornia augmentations.

adamjstewart (changes requested):

calebrob6:
@ashnair1 I can do this if you're busy

ashnair1:
Appreciate it. Might be a while before I can get back to this.

torchgeo/transforms/transforms.py
```diff
@@ -62,10 +63,20 @@ def forward(self, batch: Dict[str, Tensor]) -> Dict[str, Tensor]:
             dtype[key] = batch[key].dtype
             batch[key] = batch[key].float()
 
+        # Convert shape of boxes from [N, 4] to [N, 4, 2]
```
adamjstewart, line 66: Is it possible to do this in the individual datasets instead of `AugmentationSequential`? We're trying to get rid of this entirely and use the Kornia implementation in the future. Kornia has keys for bbox, bbox_xyxy, and bbox_xywh to choose from. Also see #985 where I'm trying to develop a standard that matches Kornia. Since there are so few object detection datasets, we should try to remove anything specific to boxes from this function to help slowly phase it out.

torchgeo/trainers/detection.py
```diff
@@ -21,6 +21,9 @@
 
 from ..datasets.utils import unbind_samples
 
+plt.switch_backend("agg")
```
adamjstewart, line 24: This shouldn't be here, we don't importing this file to inadvertently change matplotlib backends.

torchgeo/datamodules/vhr10.py
```diff
@@ -0,0 +1,185 @@
...
+                K.Normalize(mean=self.mean, std=self.std),
+                K.Resize(self.patch_size),
+                data_keys=["image", "boxes", "masks"],
+            ),
+            batch_size,
+        )
+
+        self.kwargs["download"] = True
+
+    def preprocess(self, sample: Dict[str, Any]) -> Dict[str, Any]:
```
adamjstewart, line 144: Seems like this function isn't used anymore, I would remove it.

torchgeo/datamodules/vhr10.py
```diff
@@ -0,0 +1,185 @@
...
+        self.aug = _AugPipe(
+            AugmentationSequential(
+                K.Normalize(mean=self.mean, std=self.std),
+                K.Resize(self.patch_size),
+                data_keys=["image", "boxes", "masks"],
+            ),
+            batch_size,
+        )
+
+        self.kwargs["download"] = True
```
adamjstewart, line 142: We should never download things automatically, it should be chosen by the user by passing this in via kwargs.

torchgeo/datamodules/vhr10.py
```diff
@@ -0,0 +1,185 @@
...
+    """Custom object detection collate fn to handle variable boxes.
+
+    Args:
+        batch: list of sample dicts return by dataset
+
+    Returns:
+        batch dict output
+    """
+    output: Dict[str, Any] = {}
+    output["image"] = [sample["image"] for sample in batch]
```
adamjstewart, line 81: Do all keys need to be split, or only boxes?

torchgeo/datamodules/vhr10.py
```diff
@@ -0,0 +1,185 @@
...
+
+from ..datasets import VHR10
+from ..samplers.utils import _to_tuple
+from ..transforms import AugmentationSequential
+from .geo import NonGeoDataModule
+
+# kornia DataKeys must accept MASKS
+
+
+class _AugPipe(Module):
```
adamjstewart, line 24: Why is this needed?

torchgeo/datamodules/geo.py
```diff
@@ -30,7 +30,7 @@ class GeoDataModule(LightningDataModule):  # type: ignore[misc]
     """
 
     mean = torch.tensor(0)
-    std = torch.tensor(255)
+    std = torch.tensor(1)
```
adamjstewart, line 33: I actually kind of like this change, although we might want to override the default in other datamodules. Might want to check the PR where I introduced this class to see which datamodules previously used 255.

adamjstewart (commented):
I know this is still a WIP but wanted to give intermediate feedback before you spend too long going in one direction.

torchgeo/datamodules/geo.py
ashnair1: Seems like `DeepGlobeLandCover`, `GID15`, `Vaihingen` and `Potsdam2D` are the ones that used 255. Might move this to a separate PR.

torchgeo/datamodules/vhr10.py
ashnair1: Each image will have N boxes, N labels and N instance segmentation masks. Since N varies image to image, they can't be stacked so they need to be split.

adamjstewart: How does Kornia normally handle this?

ashnair1: From the detection examples I've seen, like the one [here](https://github.com/kornia/tutorials/blob/355d33b068db877beb39f9e6c5866dd6dfa10c96/scripts/training/object_detection/main.py#L74), it's done in a similar manner i.e. via list of dicts. There aren't much (if any) examples on instance segmentation.

For example, in `AugmentationSequential` while list of boxes work, list of masks doesn't. By default the MASK datakey assumes a segmentation style mask tensor.

This is why I had to add MASKS as an alias in `kornia.constants`  so that we could differentiate between mask and masks in our `AugmentationSequential` and handle it appropriately.

ashnair1:
@grantcurell
It's been a while since I worked on this PR and the codebase has changed quite a bit since then. I'm currently in the process of updating this PR and a functioning version should be available soon (hopefully this week).

adamjstewart:
@ashnair1 just mark this as "Ready for review" when you want me to review things or ping me on slack if I forget.

conf/vhr10.yaml
adamjstewart, line 1: We no longer store files here, you can just delete this.

pyproject.toml
```diff
@@ -267,6 +267,7 @@ filterwarnings = [
     # Unexpected warnings, worth investigating
     # Lightning is having trouble inferring the batch size for ChesapeakeCVPRDataModule and CycloneDataModule for some reason
     "ignore:Trying to infer the `batch_size` from an ambiguous collection:UserWarning",
+    "ignore: Tight layout not applied. The bottom and top margins cannot be made large enough:UserWarning",
```
adamjstewart, line 270: What causes this error and why should we ignore it?

tests/conf/vhr10.yaml
```diff
@@ -0,0 +1,20 @@
+model:
+  class_path: ObjectDetectionTask
+  init_args:
+    model: "faster-rcnn"
+    backbone: "resnet50"
+    num_classes: 11
+    lr: 2.5e-5
+    patience: 10
+
```
adamjstewart, line 9: Can you delete all these empty lines to match other files?

tests/data/vhr10/data.py
```diff
@@ -54,12 +53,7 @@ def generate_test_data(root: str, n_imgs: int = 3) -> str:
             "segmentation": [[1, 1, 2, 2, 3, 3, 4, 5, 5]],
             "iscrowd": 0,
         }
-        if i != 0:
-            ANNOTATION_FILE["annotations"].append(annot)
-        else:
-            noseg_annot = deepcopy(annot)
-            del noseg_annot["segmentation"]
-            ANNOTATION_FILE["annotations"].append(noseg_annot)
+        ANNOTATION_FILE["annotations"].append(annot)
```
adamjstewart, line 56: Did something change in the dataset such that all samples now have segmentations? I didn't think that was the case for the negative set.

torchgeo/datamodules/vhr10.py
```diff
@@ -0,0 +1,151 @@
...
+"""NWPU VHR-10 datamodule."""
+
+from typing import Any, Callable, Union
+
+import kornia.augmentation as K
+import torch
+from einops import rearrange
+from torch import Generator, Tensor
+from torch.nn import Module
+from torch.utils.data import random_split
```
adamjstewart, line 13: Let's just use `.utils.dataset_split` instead so we don't need to bump the minimum version of torch/torchvision.

torchgeo/datamodules/vhr10.py
adamjstewart: Why isn't this needed for NASA Marine Debris?

torchgeo/datamodules/vhr10.py
adamjstewart: Can we move this to `.utils` so that vhr10 and nasa_marine_debris can share it? If you think this is common enough, we could even move it into `torchgeo.datasets.utils` and export it like we do with our other custom collation functions.

pyproject.toml
ashnair1: This was originally caused by calling `plt.tight_layout`  but I was able to resolve it now by setting the `constrained` flag in the subplot function.

torchgeo/datamodules/vhr10.py
ashnair1: Documentation of that function says it was deprecated in 0.4 but I imagine the aim is to replace it later everywhere in one go.

tests/data/vhr10/data.py
ashnair1: No, this check was implemented to simulate a case where segmentations were missing from the dataset. However recently I got to check the dataset and the positive set did all have segmentations making this unnecessary. The negative set has no annotations (so no boxes. labels, segmentations etc)

pyproject.toml
ashnair1: Ok, so setting layout as a parameter is only possible for `matplotlib>=3.5.0`. This means that we either bump matplotlib's min version or ignore this warning.

tests/data/vhr10/data.py
adamjstewart: So don't we still need this code to simulate the negative set?

ashnair1: No. The negative set is already simulated by having a directory of images in a negative test folder. This bit was used to simulate a situation where there was no segmentations in the positive set. Which is not the case in the dataset since all samples in the positive set do have segmentations.

torchgeo/datamodules/vhr10.py
```diff
@@ -0,0 +1,147 @@
...
+    output["boxes"] = [sample["boxes"] for sample in batch]
+    output["labels"] = [sample["labels"] for sample in batch]
+    output["masks"] = [sample["masks"] for sample in batch]
+    return output
+
+
+class VHR10DataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the VHR10 dataset.
+
+    .. versionadded:: 0.5
```
adamjstewart, line 87:
```suggestion
    .. versionadded:: 0.6
```

torchgeo/datasets/vhr10.py
```diff
@@ -297,6 +291,7 @@ def _load_image(self, id_: int) -> Tensor:
             tensor = torch.from_numpy(array)
             # Convert from HxWxC to CxHxW
             tensor = tensor.permute((2, 0, 1))
+            tensor = tensor.float() / tensor.max()
```
adamjstewart, line 294: I don't think we want to do normalization like this, it should be left up to the user.

pyproject.toml
adamjstewart: Can we instead change the size of the figure before calling tight_layout? Would like to avoid bumping matplotlib version if we don't need to.

tests/conf/vhr10.yaml
adamjstewart: Still one empty line left

torchgeo/datamodules/geo.py
adamjstewart: Let's separate this out

torchgeo/datamodules/vhr10.py
```diff
@@ -0,0 +1,144 @@
...
+import kornia.augmentation as K
+from torch import Tensor
+
+from ..datasets import VHR10
+from ..samplers.utils import _to_tuple
+from ..transforms import AugmentationSequential
+from .geo import NonGeoDataModule
+from .utils import AugPipe, dataset_split
+
+# class _AugPipe(Module):
```
adamjstewart, line 17: Can delete this

torchgeo/datamodules/vhr10.py
adamjstewart: Still need to move

torchgeo/datamodules/geo.py
```diff
@@ -30,7 +30,7 @@ class BaseDataModule(LightningDataModule):
...
```
adamjstewart, line 33: Can we separate this change out into a separate PR?

torchgeo/datamodules/utils.py
```diff
@@ -19,6 +22,81 @@ class MisconfigurationException(Exception):
     """Exception used to inform users of misuse with Lightning."""
 
 
+class AugPipe(Module):
```
adamjstewart, line 25: Can't remember if I already suggested this, but alternative idea: what if we use `VHR10(transforms=...)`? Would that be the same as what `AugPipe` is designed to do? So `transforms` would be for sample-wise changes and `self.aug` would be for batch-wise changes.

torchgeo/datamodules/utils.py
```diff
@@ -19,6 +22,81 @@ class MisconfigurationException(Exception):
...
+            if "masks" in batch:
+                batch["masks"][bs] = batch_dict["masks"]
+
+        # Stack images
+        batch["image"] = rearrange(batch["image"], "b () c h w -> b c h w")
+
+        return batch
+
+
+def collate_fn_detection(batch: list[dict[str, Tensor]]) -> dict[str, Any]:
```
adamjstewart, line 76: You definitely already explained this to me, but why can't we make these changes in the dataset instead of during collation?

torchgeo/datamodules/vhr10.py
```diff
@@ -0,0 +1,83 @@
...
+            batch_size,
+        )
+
+    def setup(self, stage: str) -> None:
+        """Set up datasets.
+
+        Args:
+            stage: Either 'fit', 'validate', 'test', or 'predict'.
+        """
+        self.dataset = VHR10(**self.kwargs)
```
adamjstewart, line 80: So we never use the negative examples? Should we?

adamjstewart (commented):
Coming back to this PR after a very long time so apologies if I've already asked and you've already answered these questions.

torchgeo/datamodules/vhr10.py
ashnair1: Correct. We do not use them.

Whether we should is an interesting question. Practically, there isn't a way to pass 'no-object' images into detectors like Faster-RCNN (or any other that I know of). This follows an assumption that those detectors already account for negative samples by considering regions of the image where the object is absent.  Additionally negative regions (where object is absent) usually outnumber positive regions (object is present). 

So, if you were building a cat detector, you don't need to pass in images with no cats at all since parts of the image that don't contain a cat is enough for the network to distinguish between what is a cat and what is not. Whether those negative regions alone are enough for a detector deployed in the wild varies case by case.

torchgeo/datamodules/utils.py
ashnair1: The dataset is concerned with individual samples. Our problem is not with the dataset, it's with how we represent the collection of samples. Default collate function does not work with boxes and instance masks. 

You can stack two image tensors to be of shape `(2,C,H,W)` but stacking two box tensors of shape `(N1, 4)` and `(N2, 4)` into a `(N1+N2, 4)` tensor would leave us with not knowing which boxes correspond to which image. Same for the instance masks.

So, we need to organise the samples into a list of tensors as opposed to a tensor itself.

adamjstewart: Gotcha. We're thinking about switching to torchvision transforms v2 someday, but not any time soon, so will keep this in mind if that makes anything simpler.

torchgeo/datamodules/utils.py
```diff
@@ -19,6 +22,81 @@ class MisconfigurationException(Exception):
...
+    """Custom collate fn for object detection & instance segmentation.
+
+    Args:
+        batch: list of sample dicts return by dataset
+
+    Returns:
+        batch dict output
+    """
+    output: dict[str, Any] = {}
```
adamjstewart, line 85: Could we instead use `torchgeo.datasets.unbind_samples` and add one extra line to modify labels?

adamjstewart (commented):
I think I'm mostly okay with this for now, although I would love to figure out if there's an easier way to handle object detection directly in Kornia instead of writing our own custom code like this. Will also ask for at least one more person to review.

torchgeo/datamodules/utils.py
ashnair1: `unbind_samples` is used to convert dict of lists into list of dicts. Here we're doing the opposite.

torchgeo/datamodules/vhr10.py
ashnair1: It is needed and has been added.

torchgeo/datamodules/utils.py
adamjstewart: Then what about concat/merge_samples? Those do the opposite.

ashnair1: Do you mean `stack_samples`? I think concat and merge samples were for `GeoDatasets`.

Either way, those functions can't be used since it assumes the samples can be stacked/concatenated. This is not the case here.

adamjstewart: Both GeoDatasets and NonGeoDatasets use the same `sample` dict format, so all collation functions are appropriate for both. Although these tend to be more used with GeoDatasets.

I see now that your output dict has lists of Tensors instead of Tensors, which is indeed different than all other builtin collation functions. I guess this is required because each sample may have a different number of objects. Not sure how people normally handle collation for that case in PyTorch.

torchgeo/transforms/transforms.py
adamjstewart: Ping on this question.

ashnair1: Perhaps but I'm not sure why that's needed at this stage. Essentially this line is to reshape the boxes before running the augmentations. This PR adds the same for instance masks as well. This is in addition to the line that was already there for default (semantic segmentation) masks. So if were to phase out this function all three would have to be removed. The standardization of sample keys is a good idea but I think it might be better suited in a separate PR.

The `[N,4,2]` format for boxes is only needed while doing augmentations with kornia. Everywhere else (like plotting or calculating MAP via torchmetrics) we use `[N, 4]`. Hence why this was done here.

adamjstewart: Alright, if Kornia and torchmetrics also disagree on this, then I'm fine with this hack. As long as we're choosing one of the two standards.

adamjstewart:
RtD CI will be fixed by #1827

adamjstewart (approved):
I think there are ways that this could be improved, but not without modifying kornia itself. I actually do want to do that someday, but we may also drop kornia entirely someday and switch back to torchvision-style transforms, so I'm not trying to rock the boat. This adds a much-needed VHR-10 data module, and the parts I'm uncertain about aren't public anyway, so I say we merge this and improve transform compatibility and sample consistency in the distant future.

PR #1084, opened 2023-02-03
Fix load_state_dict for all timm models

### LINKED ISSUE #1049: utils.load_state_dict() does not support ViT architecture
### Description

When using the new pretrained weights, the `load_state_dict()` function is looking for a "conv1" in the model to determine the expected input channels to the models. However, only ResNet Architecture weights have this. ViTs begin with "patch_embed". I think the trainer `test_classification.py` as well as the others are only testing with a ResNet18, so that is why the tests didn't catch it.

### Steps to reproduce

```python
from torchgeo.models import ViTSmall16_Weights
from torchgeo.trainers import ClassificationTask

task = ClassificationTask(
    model="vit_small_patch16_224",
    weights=ViTSmall16_Weights.SENTINEL2_ALL_DINO,
    num_classes=10,
    in_channels=13,
)
```

### Version

0.4.0

nilsleh:
This PR closes #1049 , by implementing Isaac's [solution](https://github.com/microsoft/torchgeo/issues/1049#issuecomment-1406949155).

### DIFF
--- a/tests/trainers/test_classification.py
+++ b/tests/trainers/test_classification.py
@@ -162,8 +162,7 @@ class TestClassificationTask:
         model_kwargs["model"] = weights.meta["model"]
         model_kwargs["in_channels"] = weights.meta["in_chans"]
         model_kwargs["weights"] = weights
-        with pytest.warns(UserWarning):
-            ClassificationTask(**model_kwargs)
+        ClassificationTask(**model_kwargs)
 
     @pytest.mark.slow
     def test_weight_str(
@@ -172,8 +171,7 @@ class TestClassificationTask:
         model_kwargs["model"] = weights.meta["model"]
         model_kwargs["in_channels"] = weights.meta["in_chans"]
         model_kwargs["weights"] = str(weights)
-        with pytest.warns(UserWarning):
-            ClassificationTask(**model_kwargs)
+        ClassificationTask(**model_kwargs)
 
     def test_invalid_loss(self, model_kwargs: Dict[str, Any]) -> None:
         model_kwargs["loss"] = "invalid_loss"
--- a/tests/trainers/test_classification.py
+++ b/tests/trainers/test_classification.py
@@ -140,6 +140,7 @@ class TestClassificationTask:
     def test_weight_enum(
         self, model_kwargs: Dict[str, Any], mocked_weights: WeightsEnum
     ) -> None:
+        model_kwargs["model"] = mocked_weights.meta["model"]
         model_kwargs["in_channels"] = mocked_weights.meta["in_chans"]
         model_kwargs["weights"] = mocked_weights
         with pytest.warns(UserWarning):
@@ -148,6 +149,7 @@ class TestClassificationTask:
     def test_weight_str(
         self, model_kwargs: Dict[str, Any], mocked_weights: WeightsEnum
     ) -> None:
+        model_kwargs["model"] = mocked_weights.meta["model"]
         model_kwargs["in_channels"] = mocked_weights.meta["in_chans"]
         model_kwargs["weights"] = str(mocked_weights)
         with pytest.warns(UserWarning):
--- a/tests/trainers/test_classification.py
+++ b/tests/trainers/test_classification.py
@@ -113,9 +113,9 @@ class TestClassificationTask:
 
     @pytest.fixture(
         params=[
-            (model, weight)
+            weights
             for model in list_models()
-            for weight in get_model_weights(model)
+            for weights in get_model_weights(model)
         ]
     )
     def weights(self, request: SubRequest) -> WeightsEnum:
@@ -123,15 +123,14 @@ class TestClassificationTask:
 
     @pytest.fixture
     def mocked_weights(
-        self, tmp_path: Path, monkeypatch: MonkeyPatch, weights: Tuple[str, WeightsEnum]
+        self, tmp_path: Path, monkeypatch: MonkeyPatch, weights: WeightsEnum
     ) -> WeightsEnum:
-        model_name, weight = weights
-        path = tmp_path / f"{weight}.pth"
-        model = timm.create_model(model_name, in_chans=weight.meta["in_chans"])
+        path = tmp_path / f"{weights}.pth"
+        model = timm.create_model(weights.meta["model"], in_chans=weights.meta["in_chans"])
         torch.save(model.state_dict(), path)
-        monkeypatch.setattr(weight, "url", str(path))
+        monkeypatch.setattr(weights, "url", str(path))
         monkeypatch.setattr(torchvision.models._api, "load_state_dict_from_url", load)
-        return weight
+        return weights
 
     def test_weight_file(self, model_kwargs: Dict[str, Any], checkpoint: str) -> None:
         model_kwargs["weights"] = checkpoint
@@ -141,6 +140,7 @@ class TestClassificationTask:
     def test_weight_enum(
         self, model_kwargs: Dict[str, Any], mocked_weights: WeightsEnum
     ) -> None:
+        model_kwargs["in_channels"] = mocked_weights.meta["in_chans"]
         model_kwargs["weights"] = mocked_weights
         with pytest.warns(UserWarning):
             ClassificationTask(**model_kwargs)
@@ -148,6 +148,7 @@ class TestClassificationTask:
     def test_weight_str(
         self, model_kwargs: Dict[str, Any], mocked_weights: WeightsEnum
     ) -> None:
+        model_kwargs["in_channels"] = mocked_weights.meta["in_chans"]
         model_kwargs["weights"] = str(mocked_weights)
         with pytest.warns(UserWarning):
             ClassificationTask(**model_kwargs)
--- a/tests/trainers/test_classification.py
+++ b/tests/trainers/test_classification.py
@@ -9,6 +9,7 @@ import pytest
 import timm
 import torch
 import torchvision
+from _pytest.fixtures import SubRequest
 from _pytest.monkeypatch import MonkeyPatch
 from omegaconf import OmegaConf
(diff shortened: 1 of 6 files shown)


tests/trainers/test_classification.py
```diff
@@ -110,6 +111,21 @@ def model_kwargs(self) -> Dict[str, Any]:
             "weights": None,
         }
 
+    @pytest.fixture(params=[*ResNet50_Weights])
+    def weights(self, request: SubRequest) -> WeightsEnum:
+        return request.param
+
+    @pytest.fixture
+    def mocked_weights(
```
adamjstewart, line 119: If you use the same name as the below fixture, it's just going to be redefined and this won't be used. You'll want to have a single fixture that iterates over all enums. I can hack on this later if you can't figure it out.

torchgeo/trainers/utils.py
```diff
@@ -54,6 +54,25 @@ def extract_backbone(path: str) -> Tuple[str, "OrderedDict[str, Tensor]"]:
     return name, state_dict
 
 
+def get_input_layer_name_and_module(model: Module) -> Tuple[str, Module]:
+    """Retrieve the input layer name and module from a timm model.
+
+    Args:
+        model: timm model
+
+    .. versionadded: 0.4.1
```
adamjstewart, line 63: Patch releases should not include any new features. I would prefix this with underscore and keep it private for now.

tests/trainers/test_classification.py
nilsleh: I think I am doing something wrong, at least cannot figure out the failing tests. I have also run this, which works with the new load_state_dict for all models except the seco weights, because the module key is not in the state dict so looking into that.

```
for model in torchgeo.models.list_models():
    for weight in torchgeo.models.get_model_weights(model):
        try:
            task = ClassificationTask(
                model=model,
                weights=weight,
                num_classes=10,
                in_channels=13,
                loss="ce"
            )
        except:
            print(model, weight)
```

So I think I need your help.

adamjstewart: Fixed. You were extremely close. The only thing missing was changing the "model"/"in_channels" from `model_kwargs` to match `mocked_weights`. Also simplified things a bit, the `WeightsEnum` contains the model name. Also added tests to make sure these work with the real weights (only run on release branch).

adamjstewart: There are still issues with the SeCo weights as you noted. I'm surprised that the first layer seems to be missing from the pretrained model, this is usually the most important layer. Can you look into this? Worst case scenario, we can copy the first layer from an ImageNet pretrained model (will obvs need to test this to make sure it works).

Also, can you copy these tests to the other trainers?

torchgeo/trainers/byol.py
```diff
@@ -326,13 +330,18 @@ def __init__(self, **kwargs: Any) -> None:
...
+                used in BYOL Task
 
         Raises:
             ValueError: if kwargs arguments are invalid
 
         .. versionchanged:: 0.4
            The *backbone_name* parameter was renamed to *backbone*. Change backbone
            support from torchvision.models to timm.
+
+        .. versionchanged:: 0.4.1
```
adamjstewart, line 343: New features shouldn't be added in patch releases. Is this absolutely required? If so, we may have to wait until 0.5 to fix this bug...

adamjstewart: I could also pretend that I didn't see this and we could document it as 0.5 instead of 0.4.1...

nilsleh: The ViT architectures require an input size of 224x224 and the current version it is 256x256 so currently you cannot run BYOL with a pretrained ViT, because that line will always throw an error.

nilsleh:
I looked into the seco weights again. Since they are originally saved as part of a pytorch-lightning module, the keys have different names then the default timm keys. Looking at the seco code there are a "q" and a "k" network and the "q" network is used as a pretrained backbone for downstream tasks. The `encoder_q.0.weight` is the only weight that has the same shape as the default timm model and there is also no bias at this stage, so I think `encoder_q.0.weight` is the `conv1.weight` we look for. I can rename it and reupload to huggingface and then I think we should be good. Or do you have a suggestion?

```
import torch
import timm

timm_model = timm.create_model("resnet18")
checkpoint = torch.load("path/to/seco_resnet18_1m.ckpt", map_location="cpu")

state_dict = checkpoint["state_dict"]
assert timm_model.conv1.weight.shape == state_dict["encoder_q.0.weight"].shape

for key, val in checkpoint["state_dict"].items():
    if key not in ["encoder_q.0.weight", "encoder_k.0.weight"]:
        if state_dict["encoder_q.0.weight"].shape == val.shape:
            print(key)

state_dict["conv1.weight"] = state_dict.pop("encoder_q.0.weight")
# save this state dict with just the encoder_q weights and renamed 0.weight and upload to huggingface
```

adamjstewart:
Your guess is as good as mine. Do the authors have any code for loading the pretrained model like your group does? If not, then I think your analysis makes sense. If you really want to be sure, you can try to train a model with those pretrained weights and make sure if converges quickly.

calebrob6:
You could use something like this -- https://gist.github.com/calebrob6/44f2e42017e2d192e837f0a1cd526c50 -- to make sure that linear-probing on a downstream dataset with the model achieves good performance. This is the notebook that I used to verify one of the SSL4EO weights I think.

nilsleh:
Yes, so the encoder_q I got from this [code](https://github.com/ServiceNow/seasonal-contrast/blob/8285173ec205b64bc3e53b880344dd6c3f79fa7a/main_bigearthnet.py#L63) where they load a backbone from their pretrained model for a downstream task.

nilsleh:
I will try to investigate again tomorrow. This is what I am using as a [script](https://gist.github.com/nilsleh/5f6c175e12169015e68d4f1b66b0349a) (takes about 3 minutes to run on cpu locally).

nilsleh:
I downloaded the bigearthnet dataset to try the linear probing script on that since that is the dataset they also report in their paper. For bigearthnet with 10,000 samples (not the full dataset), I get the following scores:

- seco resnet18 weights: 0.4983
- timm imagenet: 0.4512
- timm random: 0.442

adamjstewart:
Interesting evidence that the weights may not be very transferable...

adamjstewart:
But this should be sufficient proof that your approach to extracting the first layer of weights is correct and we can move forward with this PR.

nilsleh:
Screenshot from 2023-02-10 19-57-31

In the [paper](https://arxiv.org/pdf/2103.16607.pdf), they report quiet a significant improvement when using Seco in linear probing, so I must be doing something wrong. I can also contact the authors to sort it out.

adamjstewart:
That's prob for MSI, not RGB

nilsleh:
I think everything is only RGB, at least that is how I interpret it when they state "Although the collected dataset contains up to 12 spectral bands, in this work we focus on the RGB channels since it is a more general modality."

nilsleh:
For the moment I updated the seco weights on huggingface, and the loading works for all weights now.

adamjstewart:
I'm still seeing the same issue:
```console
$ pytest -m slow tests/trainers/
...
FAILED tests/trainers/test_byol.py::TestBYOLTask::test_weight_enum_download[ResNet18_Weights.SENTINEL2_RGB_SECO] - KeyError: 'conv1.weight'
FAILED tests/trainers/test_byol.py::TestBYOLTask::test_weight_enum_download[ResNet50_Weights.SENTINEL2_RGB_SECO] - KeyError: 'conv1.weight'
FAILED tests/trainers/test_byol.py::TestBYOLTask::test_weight_str_download[ResNet18_Weights.SENTINEL2_RGB_SECO] - KeyError: 'conv1.weight'
FAILED tests/trainers/test_byol.py::TestBYOLTask::test_weight_str_download[ResNet50_Weights.SENTINEL2_RGB_SECO] - KeyError: 'conv1.weight'
FAILED tests/trainers/test_classification.py::TestClassificationTask::test_weight_enum_download[ResNet18_Weights.SENTINEL2_RGB_SECO] - KeyError: 'conv1.weight'
FAILED tests/trainers/test_classification.py::TestClassificationTask::test_weight_enum_download[ResNet50_Weights.SENTINEL2_RGB_SECO] - KeyError: 'conv1.weight'
FAILED tests/trainers/test_classification.py::TestClassificationTask::test_weight_str_download[ResNet18_Weights.SENTINEL2_RGB_SECO] - KeyError: 'conv1.weight'
FAILED tests/trainers/test_classification.py::TestClassificationTask::test_weight_str_download[ResNet50_Weights.SENTINEL2_RGB_SECO] - KeyError: 'conv1.weight'
FAILED tests/trainers/test_regression.py::TestRegressionTask::test_weight_enum_download[ResNet18_Weights.SENTINEL2_RGB_SECO] - KeyError: 'conv1.weight'
FAILED tests/trainers/test_regression.py::TestRegressionTask::test_weight_enum_download[ResNet50_Weights.SENTINEL2_RGB_SECO] - KeyError: 'conv1.weight'
FAILED tests/trainers/test_regression.py::TestRegressionTask::test_weight_str_download[ResNet18_Weights.SENTINEL2_RGB_SECO] - KeyError: 'conv1.weight'
FAILED tests/trainers/test_regression.py::TestRegressionTask::test_weight_str_download[ResNet50_Weights.SENTINEL2_RGB_SECO] - KeyError: 'conv1.weight'
```

nilsleh:
Mhm I am not getting those errors. Maybe, the old weights are still cached in your torch/hub? I had to delete those so it would reload the new ones after I uploaded them to huggingface.

adamjstewart:
Oh mine are prob cached, let me delete

adamjstewart:
Yep, works now. Thanks!

adamjstewart (approved):

Issue #367, opened 2022-01-25
No dockerfile for torchgeo


MATRIX4284:
there is no docker images for torchgeo

adamjstewart:
What would be the benefit of having a Docker image for TorchGeo? Most software, especially Python libraries, isn't distributed via Docker. We already have source code and wheels available on PyPI, and support package managers like pip/conda/spack. TorchGeo itself is pure-Python, so as long as you have one of the aforementioned package managers installed, you can easily install TorchGeo.

If we did provide Docker images, we would need to provide them for many OSes, CUDA versions, and Python versions. I don't see a benefit of this over package managers. But I don't use Docker heavily other than for reproducing bugs, so please correct me if I'm wrong.

MATRIX4284:
i added this because I tried to install torchgeo in my OS 4 times but was unable to import datasets when trying to run with Jupyter.Later on i came to know that is happening as i need to set my PROJ_LIB to the correct one before i run the jupyter.
To ease the usability while using open source jupyter to download datsets  and visualize torcheo outputs the idea of torchgeo docker was provided.
It will enable the users to quickly get the environment with torchgeo up and running to use with jupyter notebook without needing to go for much OS isssues by trying to install on bare OS .This is enable more people in geospatial community to adapt torchgeo.
Moreover going forward geospatial will be used n cloud where containerization will play an important and anyhow people need to dockerize first to use torchgeo within micro-services framework.Again in that case this docker wil come as a useful feature.

adamjstewart:
> Later on i came to know that is happening as i need to set my PROJ_LIB to the correct one before i run the jupyter.

Yes, I've actually encountered this same issue myself. Package managers like Spack and Conda [explicitly set this for you](https://github.com/spack/spack/blob/v0.17.1/var/spack/repos/builtin/packages/proj/package.py#L111,L117). If you're having problems with pip, I would suggest using a more powerful package manager like Spack or Conda. Pip isn't designed to install non-Python dependencies.

From what I can tell, the issue you reported is specific to PROJ and has nothing to do with TorchGeo. Maybe it would make more sense to add a Dockerfile to the PROJ library?

adamjstewart:
Closing as it doesn't sound like this is a feature we really need in TorchGeo.

PR #1085, opened 2023-02-03
✨ Add devcontainer

julien-blanchon:
Add `.devcontainer` directory with basic working devcontainer


adamjstewart:
What is this and why would we want it?

calebrob6:
https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/adding-a-dev-container-configuration/introduction-to-dev-containers

adamjstewart:
How is this different from #367/#368 which was rejected?

julien-blanchon:
Hi, devcontainer make it easier to contribute, basically you just define the what's the basic/minimal development system (here ubuntu+python and requirements), and open source contributor could directly launch the repo on this predefine system.

The usual usage are:
- Open VS-Code with devcontainer extensions installed, user click on open in devcontainer and the VS-Code launch in the devcontainer system with all requirements installed.
- Open on cloud provider such as codespace/gitpod ... The cloud provider will directly read the devcontainer in order to repo with in the dev system.

So it's kind like your conda env, redondent with ./requirements but help a lot in the user/dev perspective.

It's up to you, I will at least keep it on my branch as it help me standardize my development system

adamjstewart:
I've never heard of or used any of these things before, so I guess it's difficult for me to want to maintain these files. I've at least used pip/conda/spack before, so I don't mind supporting those installation methods. But any file we add to this git repo is something we're committing to maintain, and I'm trying to keep that burden minimal.

Like the docker PR, this feels very specific (what if someone wants to use RHEL instead of Ubuntu? or macOS or Windows? or Python 3.8?), and we don't want to maintain a file like this for every system and Python version. Maybe we could just add this file to `.gitignore` so you can keep it in your local copy without worrying about it getting sucked into PRs?

At least it doesn't duplicate the list of places I need to update dependencies. I would like to get rid of `requirements/` but it's needed for CI. If I could, I would delete `environment.yml` too but I think @calebrob6 is using it 😄 

So this isn't a no, but I'm not yet convinced.

calebrob6:
@julien-blanchon thanks for the contribution! I've never used Codespaces before, is the idea that if you open `microsoft/torchgeo` (or a fork of it) in a Codespace environment, then it will automatically run these files?

@adamjstewart I don't think this is a huge overhead (similar in scope to the pre-commit hook) and generally makes it easier for some people to edit. The only thing we would need to change is what is `pip install`ed. I also don't think it is some slippery slope either -- if someone doesn't want to use it then they don't have and if someone wants it to be different then they can open a PR.

adamjstewart:
> similar in scope to the pre-commit hook

I want to get rid of this too 😆

calebrob6:
Why -- you don't have to maintain it? (put differently, these things don't get released / tested -- if they break they are going to break for technical people that are already trying to contribute and who will tell us about it)

Is there a like this nagging feeling that you get when something isn't under the comforting umbrella of `pytest` 🙂?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,55 @@
...
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	"features": {
+		"ghcr.io/devcontainers/features/git:1": {"version": "os-provided"}
+	},
+
+	// The arguments required when starting in the container.
+	"runArgs": [
+    // "--ipc=host",
+    // "--gpus=all" // Enable GPU
+  ],
```
calebrob6, line 17: indenting is weird

adamjstewart:
> you don't have to maintain it?

Someone has to maintain it. precommit becomes quickly out of date any time a style or mypy dep changes. I just had to upgrade the black version the other day.

> Is there a like this nagging feeling that you get when something isn't under the comforting umbrella of pytest 🙂?

Yes 🙂

calebrob6:
Yeah but that's only because you have an insane memory for these things. If you forgot, then nothing bad would have happened and ashnair would have gotten to it (https://github.com/microsoft/torchgeo/commits/main/.pre-commit-config.yaml). If ashnair didn't get to it then someone else would have.

calebrob6:
(@julien-blanchon sorry for meta-discussion here, do you mind signing the CLA?)

julien-blanchon:
Yes sure, 

> @julien-blanchon please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

julien-blanchon:
Here is a little example on the vscode desktop app. It should also work with Intellij, expect for the extensions part.

https://user-images.githubusercontent.com/11278197/216736959-5a2711c1-c9a7-42c3-a6d8-1e9f6783c8e4.mov


CPU and GPU configuration must be split, at least for now.

About codespace, here is the setup example: 

Screenshot 2023-02-04 at 02 07 00

The image build could be quite long, expecially with codespace. So you can enable image prebuild in "settings/codespaces/prebuild_configurations/new". So any contributor will be able to launch the project with the specify configuration (same extensions, same black/mypy/... conf ...)

.devcontainer/devcontainer.json
julien-blanchon: Yes, CPU and GPU are in separate file now

.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,13 @@
...
+sudo apt-get -y --force-yes update
+sudo apt-get -y --force-yes install libgdal-dev
+echo CPLUS_INCLUDE_PATH=/usr/include/gdal >> ~/.bashrc
+echo C_INCLUDE_PATH=/usr/include/gdal >> ~/.bashrc
+pip install --upgrade pip
+pip install --user -r ./requirements/datasets.txt
+pip install --user -r ./requirements/docs.txt
+pip install --user -r ./requirements/required.txt
+pip install --user -r ./requirements/style.txt
+pip install --user -r ./requirements/tests.txt
```
adamjstewart, line 12: These files are not intended to be used outside of CI. Would `pip install .[datasets,docs,style,tests]` work? That might install TorchGeo itself though, not sure if that would be problematic. There is also `pip install -e`.

pyproject.toml
```diff
@@ -7,7 +7,7 @@ build-backend = "setuptools.build_meta"
 
 [tool.black]
 target-version = ["py38", "py39", "py310"]
-color = true
+# color = true # Problem with vscode autoformat see: https://github.com/microsoft/vscode-python/issues/16341
```
adamjstewart, line 10: It appears that color is only used for `--diff`. Can you delete this line and add `--color` to `.github/workflows/style.yaml`?

.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+export DEBIAN_FRONTEND=noninteractive
+sudo apt-get -y --force-yes update
+sudo apt-get -y --force-yes install libgdal-dev
+echo CPLUS_INCLUDE_PATH=/usr/include/gdal >> ~/.bashrc
+echo C_INCLUDE_PATH=/usr/include/gdal >> ~/.bashrc
```
adamjstewart, line 6: Wheels for rasterio/fiona/etc. should come with GDAL, did you encounter issues with installing TorchGeo's dependencies?

adamjstewart (commented):
Still not convinced that this is something we want to maintain (why stop at codespaces when there are dozens of other editors and environments with dotfiles we could add?) but figured I would comment on some of the issues I see.

.devcontainer/postCreateCommand.sh
julien-blanchon: Yes, on my side gdal was needed to use TorchGeo in my container

.devcontainer/postCreateCommand.sh
julien-blanchon: As it's a devcontainer, `docs` `style` and `tests` requirements are usefull.
The goal of the devcontainer is move some CI post commit processing (such as styling and test) on the dev IDE side.
Feature like format on save are quite usefull and deploy by default in the devcontainer

pyproject.toml
julien-blanchon: > Still not convinced that this is something we want to maintain (why stop at codespaces when there are dozens of other editors and environments with dotfiles we could add?) but figured I would comment on some of the issues I see.

`.devcontainer` are not exclusive to Codespace.
It work with many local IDE (VS-Code, PyCharm ...).
And many cloud service, Codespaces, GitPod

.devcontainer/postCreateCommand.sh
adamjstewart: Yes, and the pip install step listed above will also install all of those optional dependencies. See https://torchgeo.readthedocs.io/en/stable/user/installation.html

.devcontainer/postCreateCommand.sh
adamjstewart: What was the error message when GDAL was not installed?

adamjstewart:
I'm slowly warming up to the idea of adding this. Other repos (nbmake) also ship devcontainers, and it can make it easier to reproduce bugs.

calebrob6:
@adamjstewart I'd like to merge this, I don't see any of the open issues as blockers and it makes it easier to for people to contribute to torchgeo

adamjstewart:
I see all of the above as blockers, shouldn't be hard to investigate these.

.devcontainer/postCreateCommand.sh
julien-blanchon: Yes the purpose of the container is to ship a ready development container without the package.

In order to use a `pip install -e .` we will need to copy the whole torchgeo project in the container which is not great.

The workflow is:
1) The editor will launch the devcontainer image with only Python + public requirements.
2) The editor will then bind the torchgeo source code into the container.
3) Maybe add in `postStartCommand ` a `pip install -e .` that will run after the start of the devcontainer and install the current torchgeo lib.

At the end the devcontainer docker image will not change much over the time (only when dependencies upgrade).

julien-blanchon:
Here is some improvement we can do in term of .devcontainer:
- [ ] Remove gdal manual installation and let pip install it
- [ ] Add/Remove VSCode specific extensions for a unified experience: For community project like this one the best is to keep extensions as minimal as possible, and only ship extensions that are part of the formatting/linting process such as black, isort ...
- [ ] Validate that the formatting/linting process work as espected. I've do my best to match your specific configuration for black, pylint, flake8 ...
- [ ] Add `pip install -e .` of torch in `postStartCommand`, so the current torchgeo codebase will be install at the start of the container. Note that requirements library has already been install in the container.

.devcontainer/postCreateCommand.sh
adamjstewart: > we will need to copy the whole torchgeo project in the container which is not great.

> The editor will then bind the torchgeo source code into the container.

How are these different?

julien-blanchon: Oups my bad, your right !!
The `pip install` must be done at the Dockerfile level (which is not specified here).
As you said the nbmake example is pretty good

This way of doing will work but is quite suboptimal.

I'm going to refactor this PR, in something more simple and robust

julien-blanchon:
Fiona seems to not install gdal properly :(

.devcontainer/postCreateCommand.sh
julien-blanchon: > What was the error message when GDAL was not installed?

Failed to get options via gdal-config: 
WARNING:root:Failed to get options via gdal-config: [Errno 2] No such file or directory: 'gdal-config'

When installing fiona

julien-blanchon:
I did some small improvment but the whole formatting/linting process still need to be tested

.devcontainer/gpu/devcontainer.json
```diff
@@ -0,0 +1,76 @@
...
+        "python.formatting.blackPath": "/usr/local/py-utils/bin/black",
+        "python.linting.enabled": true,
+        "python.linting.pylintEnabled": true,
+        "python.linting.pylintPath": "/usr/local/py-utils/bin/pylint",
+        "python.linting.flake8Enabled": true,
+        "python.linting.flake8Path": "/usr/local/py-utils/bin/flake8",
+        "python.linting.mypyEnabled": true,
+        "python.linting.mypyPath": "/usr/local/py-utils/bin/mypy",
+        "python.linting.mypyArgs": [
+          "--follow-imports=silent",
```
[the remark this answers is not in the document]
julien-blanchon: I'm changing it

.devcontainer/postStartCommand.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+export DEBIAN_FRONTEND=noninteractive
+yes | pip install -e .
```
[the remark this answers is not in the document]
julien-blanchon: It might happen when conflict appear. I will try without `yes` but keep `DEBIAN_FRONTEND=noninteractive` as it's usefull for possible update of the script (apt install for example)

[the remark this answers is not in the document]
adamjstewart: I usually do `apt install -y` for that, it's less verbose.

adamjstewart:
@julien-blanchon are you still working on this? Would love to get this merged someday so we can use codespaces for PR reviews.

julien-blanchon:
@adamjstewart Could you have a look ! It's way simpler than before and work both on CPU and GPU

julien-blanchon:
Screenshot 2024-03-01 at 23 20 26

Look like almost all the test pass inside the devcontainer expect some dataset and transform ones. I'm not familiar enouth with torchgeo. Could you have a look into that before merging ?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,52 @@
...
+  // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+  "image": "mcr.microsoft.com/devcontainers/python:3.10",
+
+  // Features to add to the dev container. More info: https://containers.dev/features.
+  "features": {
+    "ghcr.io/devcontainers/features/git:1": { "version": "os-provided" }
+  },
+
+  "postCreateCommand": "pip install .[datasets,docs,style,tests]",
+  "postStartCommand": "pip install -e .",
```
adamjstewart, line 13: Can these two be combined? Maybe we can use:
```
pip3 install -e .[all]
```
Not sure what the different between create and start are.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,52 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+{
+  "name": "Torchgeo DevContainer",
+  // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+  "image": "mcr.microsoft.com/devcontainers/python:3.10",
```
adamjstewart, line 5: Maybe bump to 3.12 since that's the latest supported version

.devcontainer/devcontainer.json
julien-blanchon: `postCreateCommand` execute on the container creation when `postStartCommand ` execute on the container start. If you think the `.[datasets,docs,style,tests]` are likelly to change often then it's a way better practice to make a `pip3 install -e .[all]`.

To be honest I used to make this seperation back in the time, but it doesn't save much and I think `pip3 install -e .[all]` may be better

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,52 @@
...
+  },
+
+  "customizations": {
+    "vscode": {
+      "extensions": [
+        "ms-python.python",
+        "ms-python.isort",
+        "ms-python.black-formatter",
+        "ms-python.vscode-pylance",
+        "ms-python.pylint",
```
adamjstewart, line 27: Do we need these? Not sure if I want to enforce any style constraints that aren't already enforced in CI.

adamjstewart (approved):
Don't know much about this, but someday I'll learn how to use Codespaces and VS Code.

.devcontainer/devcontainer.json
julien-blanchon: That's extension, I would keep them espectially pylance. This give many IDE feedback to the user even before the CI. For example the current settings will display pylint error on the IDE side before any CI. It will also auto format the code with the same specification as the CI. This may put less work on the CI side and enforce best practice directly in the IDE

adamjstewart: Yes, I definitely want to keep black/mypy/isort/pydocstyle/flake8 because we also enforce that in CI. I'm specifically commenting on things like pylance and pylint which we don't use.

adamjstewart:
Can you tell me more about the failing tests? Do those tests fail for you when manually run in your local environment, or do they only fail in VS Code?

.devcontainer/devcontainer.json
julien-blanchon: You can remove pylint. I would keep pylance anyway because it's just visual, it doesn't enforce anything. 
Btw for torchgeo in general you can have a look to [ruff](https://docs.astral.sh/ruff/) that bundle everything in a single package.

adamjstewart: See https://github.com/microsoft/torchgeo/pull/1908#issuecomment-1964000416 for my thoughts on ruff. Not strongly opposed in principle, but let's discuss that in a separate issue.

adamjstewart: Does pylance flag anything that the other tools don't flag?

julien-blanchon:
> Can you tell me more about the failing tests? Do those tests fail for you when manually run in your local environment, or do they only fail in VS Code?

Seems to be due to the Python version. I think it's gone with the 3.12

adamjstewart:
Let's merge as is. We can always make improvements in future PRs. Want to squeeze this in before our next release.

julien-blanchon:
> Let's merge as is. We can always make improvements in future PRs. Want to squeeze this in before our next release.

I'm going to investigate next week about the tests failed. I don't think it's blocking anyway

PR #1095, opened 2023-02-07
One extra to rule them all

adamjstewart:
This adds an "all" extra that allows you to install all optional dependencies. The following:
```console
$ pip install torchgeo[all]
```
is now equivalent to:
```console
$ pip install torchgeo[datasets,docs,style,tests]
```
Note that this requires pip 21.2+. Users on older versions of pip can still install torchgeo normally, but won't be able to install `torchgeo[all]`.

See https://hynek.me/articles/python-recursive-optional-dependencies/.


calebrob6 (approved):

calebrob6:
neat!

PR #1097, opened 2023-02-08
CI: test notebooks when modified

adamjstewart:
Previously, our notebooks were only tested when creating a release, which made it difficult to add or modify tutorials. With this PR, our notebooks are now also tested when modified.

Unfortunately it doesn't seem possible to reuse the tests in `releases.yaml`, we have to duplicate them: https://github.com/orgs/community/discussions/46680.


adamjstewart:
This is prob safe to merge. One of the tutorials doesn't pass, and hasn't passed for a very long time, but we're still trying to figure out why. By having separate tests for these files, it should be easier to debug and iterate on. Once this is merged, I have several ideas for how we can speed up the tests.

isaaccorley (approved):

PR #1099, opened 2023-02-08
Fix folder name for labels_root in the FAIR1M dataset

### LINKED ISSUE #1098: FAIR1M Dataset labels_root name
### Description

I just downloaded the FAIR1M dataset and found that the labelXmls.zip files extract to a folder named `labelXml` instead of the folder expected by `torchgeo.datasets.FAIR1M` shown below. 

https://github.com/microsoft/torchgeo/blob/d897e3374bf15450a02d0bde4344ea64b66ae5fe/torchgeo/datasets/fair1m.py#L160

I can get around this by extracting to the expected folder myself but this is not how the zip extracts by default. Did this maybe change recently in the dataset?

### Steps to reproduce

1. Download FAIR1M dataset from the Google Drive on their website (https://www.gaofen-challenge.com/benchmark)
2. Create dataset object
```
from torchgeo.datasets import FAIR1M
ds = FAIR1M(root='/data/FAIR1M/train/part2')
```
3. Check data root and see the folder name mismatch. This occurs in all the data folders.
```
$ ls /data/FAIR1M/train/part2
images  images.zip  labelXml  labelXmls.zip
$ ls /data/FAIR1M/train/part1
images  images.zip  labelXml  labelXmls.zip
$ ls /data/FAIR1M/validation
images  images.zip  labelXml  labelXmls.zip
```

### Version

0.4.0

SpontaneousDuck:
Change name of expected extracted folder for FAIR1M labels_root. Fixes #1098


adamjstewart:
Thanks for the fix!

Where did you even download the dataset? The dataset homepage seems to be down.

In order to get the tests to pass, you'll also need to change the fake data in `tests/data/fair1m` to have the same directory structure (both the directory and the zip file of that directory). Let me know if you need help.

SpontaneousDuck:
I was able to access the dataset's website here: [https://www.gaofen-challenge.com/benchmark](https://www.gaofen-challenge.com/benchmark). Since it requires an account to get the links to the dataset, my colleague who is from China was able to get that for me. The Google Drive hosted dataset is a public folder here: [https://drive.google.com/drive/folders/1lCZibAl3k9sI5d7ahRm_5GA3g7OCLXmY](https://drive.google.com/drive/folders/1lCZibAl3k9sI5d7ahRm_5GA3g7OCLXmY).

SpontaneousDuck:
I'll go fix those tests! Let me know if anything else needs to be fixed

SpontaneousDuck:
@microsoft-github-policy-service agree company="Kostas Research Institute at Northeastern University"

adamjstewart:
I think you need to update the zip file too.

adamjstewart:
The name of the zip file is correct, but the contents of the zip file in our test data are not correct.

SpontaneousDuck:
Ah, got it! I just changed the contents and made a new commit

torchgeo/datasets/fair1m.py
```diff
@@ -157,7 +157,7 @@ class FAIR1M(NonGeoDataset):
     }
 
     image_root: str = "images"
-    labels_root: str = "labelXmls"
+    labels_root: str = "labelXml"
     filenames = ["images.zip", "labelXmls.zip"]
     md5s = ["a460fe6b1b5b276bf856ce9ac72d6568", "ca8666dc43a553f8d65e5dc671a8ac3c"]
```
adamjstewart, line 162: md5 prob changed too, you can get the new one with `md5 labelXmls.zip` on the command line.

adamjstewart:
> I was able to access the dataset's website here: https://www.gaofen-challenge.com/benchmark.

Can you update the URL in the `FAIR1M` docstring? The old one gives a 404.

adamjstewart:
Is there a difference in directory structure between the 1.0 and 2.0 versions of the dataset? I'm wondering if this PR only supports 2.0 and the old version only supports 1.0. It would be easy to support both if you have the ability to download both and test them.

SpontaneousDuck:
So the current version of the dataset has two different parts to the training dataset as well as a validation set. (train/part1 and train/part2). It looks like the md5 for images.zip matches the part1 folder so I think what they did is leave part1 as the original dataset and added the part2 folder as well as the validation folder as version 2.0. The md5 for `part1/labelXmls.zip` so there must have been some change to the original labels when they updated. Maybe the folder name change in the zip was changed in the update and that also affected the original dataset? 

From what I can tell, they left the original zips from 1.0 in the 2.0 dataset and added more data zips to extend it. Should I extend this PR to support version 2.0 or make that a separate PR? Either way I am getting the 1.0 md5s and tests fixed now 😊

SpontaneousDuck:
Directory structure for FAIR1M2.0 is as below with the exception of the `labelXmls` folders which I renamed to work with the current release version of `torchgeo.datasets.FAIR1M`
```
.
├── test
│   ├── images0.zip
│   ├── images1.zip
│   └── images2.zip
├── train
│   ├── part1
│   │   ├── images
│   │   ├── images.zip
│   │   ├── labelXmls
│   │   └── labelXmls.zip
│   └── part2
│       ├── images
│       ├── images.zip
│       ├── labelXmls
│       └── labelXmls.zip
└── validation
    ├── images.zip
    └── labelXmls.zip
```

adamjstewart:
So what you're saying is that our dataset is for version 1.0, but the authors seem to have changed the directory structure for 1.0 (and therefore the dataset broke and the MD5s changed), and your PR fixes support for version 1.0? That sounds fine to me, we can add support for version 2.0 in a separate PR.

SpontaneousDuck:
Great! Yes, they actually changed the 1.0 version download link to just link to v2.0/train/part1 folder. So, if anyone downloads the 1.0 version, this dataset code will work. This dataset code will also work if the version 2.0 dataset is downloaded and `root="train/part1` is set for `torchgeo.datasets.FAIR1M`.

adamjstewart:
Closing and reopening to try to fix the coverage checks...

adamjstewart (approved):

PR #1102, opened 2023-02-09
Support for radiant-mlhub 0.5+

### LINKED ISSUE #711: Radiant MLHub datasets - FileNotFoundError due to `radiant_mlhub` version 0.5.0 release
### Description

@calebrob6 flagged an issue in merging #510  that the `utils.download_radiant_mlhub_collection(...)` method is resulting in a `FileNotFoundError`. This is because the method in `torchgeo.utils` is expecting the `radiant_mlhub.Dataset.download(...)` method to fetch a **.tar.gz** archive of the entire dataset or for an individual collection. As of v0.5.0+ of the `radiant-mlhub` Python Client, this is no longer the case, there are no **.tar.gz** archives downloaded except for the archive that contains the STAC catalog. This change was made to support asset level filtering and downloading (please see https://github.com/radiantearth/radiant-mlhub/pull/104).

As is the case for all MLHub dataset classes in Torchgeo (BeninSmallHolderCashews, CV4AKenyaCropType, TropicalCycloneWindEstimation, NASAMarineDebris), they anticipate an archive of the entire dataset, or archives of each collection. For example with NASAMarineDebris, similar to CloudCoverDetection, these are assigned as class properties for the archives and their md5 checksums:

```python
    dataset_id = "nasa_marine_debris"
    directories = ["nasa_marine_debris_source", "nasa_marine_debris_labels"]
    filenames = ["nasa_marine_debris_source.tar.gz", "nasa_marine_debris_labels.tar.gz"]
    md5s = ["fe8698d1e68b3f24f0b86b04419a797d", "d8084f5a72778349e07ac90ec1e1d990"]
```

**Proposed Short Term Solution:**
It seems the easiest solution is to fix the version of the `radiant-mlhub` Python Client to v0.4.x, prior to the release which introduced the bug in torchgeo.

**Long Term Solution:**
Ideally any existing and new torchgeo datasets that are sourced from Radiant MLHub will be updated to reflect the latest `Dataset.download(...)` functionality. This is the sequence of events from calling that method now:

1. Download the catalog archive with all STAC objects for the dataset
2. Uncompress the STAC catalog archive directory
3. Load the STAC catalog into a Sqlite database on disk
4. Apply spatial, temporal and band filters to the SQL table
5. Construct list of assets to download from query
6. Download all assets locally into catalog directory

### Steps to reproduce

1. Run the following code to reproduce:

```python
from torchgeo.datasets import BeninSmallHolderCashews

cashews = BeninSmallHolderCashews(
    root='/data',
    bands=('B02','B03','B04'),
    download=True
)

# the same result will happen with CV4AKenyaCropType, TropicalCycloneWindEstimation, NASAMarineDebris, CloudCoverDetection
```

### Version

0.4.0.dev0

### LINKED ISSUE #1101: NASAMarineDebris Dataset download process changed
### Description

It appears like the format of what is downloaded by `radiant_mlhub` is different now and the extracted datasets are downloaded directly instead of the tarballed datasets. When `torchgeo.Datasets.NASAMarineDebris.__init__()` calls `self._verify()`, the verification fails because the files it checks for (`filenames = ["nasa_marine_debris_source.tar.gz", "nasa_marine_debris_labels.tar.gz"]`) are not being created and their contact jsut downloaded directly.

### Steps to reproduce

1. Try to instantiate `NASAMarineDebris` with `download=True`
```
from torchgeo.datasets import NASAMarineDebris
NASAMarineDebris(root='test', download=True, api_key = "key")
```
2. The following error occurs after downloading:
```
File torchgeo/datasets/nasa_marine_debris.py:86, in NASAMarineDebris.__init__(self, root, transforms, download, api_key, checksum, verbose)
     84 self.checksum = checksum
     85 self.verbose = verbose
---> 86 self._verify()
     87 self.files = self._load_files()

File torchgeo/datasets/nasa_marine_debris.py:216, in NASAMarineDebris._verify(self)
    214 for filename in self.filenames:
    215     filepath = os.path.join(self.root, filename)
--> 216     extract_archive(filepath)

File torchgeo/datasets/utils.py:124, in extract_archive(src, dst)
    122 for suffix, extractor in suffix_and_extractor:
    123     if src.endswith(suffix):
--> 124         with extractor(src, "r") as f:
    125             f.extractall(dst)
    126         return

File lib/python3.10/tarfile.py:1632, in TarFile.open(cls, name, mode, fileobj, bufsize, **kwargs)
   1630     saved_pos = fileobj.tell()
...
--> 174     fileobj = self.myfileobj = builtins.open(filename, mode or 'rb')
    175 if filename is None:
    176     filename = getattr(fileobj, 'name', '')
```
3. Inspecting what is downloaded and the code, when the download code is called:
https://github.com/microsoft/torchgeo/blob/d897e3374bf15450a02d0bde4344ea64b66ae5fe/torchgeo/datasets/utils.py#L215-L216
The below tree is created:
```
test
├── nasa_marine_debris
│   ├── catalog.json
│   ├── err_report.csv
│   ├── mlhub_stac_assets.db
│   ├── nasa_marine_debris_labels
│   └── nasa_marine_debris_source
└── nasa_marine_debris.tar.gz
```

### Version

0.4.0

SpontaneousDuck:
Quick changes to several datasets and the tests to support the new download file structures in `radiant-mlhub>=0.5`.

Quick changes for now. Currently no checksum checking for the dataset since `radiant-mlhub` just downloads the individual files directly. Any ideas how I can check for partial dataset downloads? `radiant-mlhub` says it will auto-resume dataset downloads if requested but this code does not currently check or ask it to resume. 

Closes #711 
Closes #1101

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -38,7 +38,7 @@ dependencies:
     - pytorch-lightning>=1.5.1
     - git+https://github.com/pytorch/pytorch_sphinx_theme
     - pyupgrade>=2.4
-    - radiant-mlhub>=0.2.1
+    - radiant-mlhub>=0.2.1,<0.6
     - rtree>=1
     - scikit-image>=0.18
     - scikit-learn>=0.22
--- a/environment.yml
+++ b/environment.yml
@@ -38,7 +38,7 @@ dependencies:
     - pytorch-lightning>=1.5.1
     - git+https://github.com/pytorch/pytorch_sphinx_theme
     - pyupgrade>=2.4
-    - radiant-mlhub>=0.2.1,<0.5
+    - radiant-mlhub>=0.2.1
     - rtree>=1
     - scikit-image>=0.18
     - scikit-learn>=0.22
--- a/environment.yml
+++ b/environment.yml
@@ -38,7 +38,7 @@ dependencies:
     - pytorch-lightning>=1.5.1
     - git+https://github.com/pytorch/pytorch_sphinx_theme
     - pyupgrade>=2.4
-    - radiant-mlhub>=0.2.1,<0.6
+    - radiant-mlhub>=0.2.1
     - rtree>=1
     - scikit-image>=0.18
     - scikit-learn>=0.22
--- a/setup.cfg
+++ b/setup.cfg
@@ -85,7 +85,7 @@ datasets =
     pyvista>=0.25.2,<0.39
     # radiant-mlhub 0.2.1+ required for api_key bugfix:
     # https://github.com/radiantearth/radiant-mlhub/pull/48
-    radiant-mlhub>=0.2.1
+    radiant-mlhub>=0.2.1,<0.6
     # rarfile 4+ required for wheels
     rarfile>=4,<5
     # scikit-image 0.18+ required for numpy 1.17+ compatibility
--- a/setup.cfg
+++ b/setup.cfg
@@ -85,9 +85,7 @@ datasets =
     pyvista>=0.25.2,<0.39
     # radiant-mlhub 0.2.1+ required for api_key bugfix:
     # https://github.com/radiantearth/radiant-mlhub/pull/48
-    # radiant-mlhub 0.5+ changed download behavior:
-    # https://github.com/radiantearth/radiant-mlhub/pull/104
-    radiant-mlhub>=0.2.1,<0.5
+    radiant-mlhub>=0.2.1
     # rarfile 4+ required for wheels
     rarfile>=4,<5
     # scikit-image 0.18+ required for numpy 1.17+ compatibility
--- a/tests/datasets/test_benin_cashews.py
+++ b/tests/datasets/test_benin_cashews.py
@@ -16,15 +16,15 @@ from torch.utils.data import ConcatDataset
 from torchgeo.datasets import BeninSmallHolderCashews
 
 
-class Dataset:
+class Collection:
     def download(self, output_dir: str, **kwargs: str) -> None:
         glob_path = os.path.join("tests", "data", "ts_cashew_benin", "*.tar.gz")
         for tarball in glob.iglob(glob_path):
             shutil.copy(tarball, output_dir)
 
 
-def fetch(dataset_id: str, **kwargs: str) -> Dataset:
-    return Dataset()
+def fetch(dataset_id: str, **kwargs: str) -> Collection:
+    return Collection()
 
 
 class TestBeninSmallHolderCashews:
@@ -33,7 +33,7 @@ class TestBeninSmallHolderCashews:
         self, monkeypatch: MonkeyPatch, tmp_path: Path
     ) -> BeninSmallHolderCashews:
         radiant_mlhub = pytest.importorskip("radiant_mlhub", minversion="0.2.1")
-        monkeypatch.setattr(radiant_mlhub.Dataset, "fetch", fetch)
+        monkeypatch.setattr(radiant_mlhub.Collection, "fetch", fetch)
         source_md5 = "255efff0f03bc6322470949a09bc76db"
         labels_md5 = "ed2195d93ca6822d48eb02bc3e81c127"
         monkeypatch.setitem(BeninSmallHolderCashews.image_meta, "md5", source_md5)
--- a/torchgeo/datasets/benin_cashews.py
+++ b/torchgeo/datasets/benin_cashews.py
@@ -17,7 +17,7 @@ from rasterio.crs import CRS
 from torch import Tensor
 
 from .geo import NonGeoDataset
-from .utils import check_integrity, download_radiant_mlhub_dataset, extract_archive
+from .utils import check_integrity, download_radiant_mlhub_collection, extract_archive
 
 
 # TODO: read geospatial information from stac.json files
@@ -56,6 +56,7 @@ class BeninSmallHolderCashews(NonGeoDataset):
     """
 
     dataset_id = "ts_cashew_benin"
+    collection_ids = ["ts_cashew_benin_source", "ts_cashew_benin_labels"]
     image_meta = {
         "filename": "ts_cashew_benin_source.tar.gz",
         "md5": "957272c86e518a925a4e0d90dab4f92d",
@@ -416,7 +417,8 @@ class BeninSmallHolderCashews(NonGeoDataset):
             print("Files already downloaded and verified")
             return
 
-        download_radiant_mlhub_dataset(self.dataset_id, self.root, api_key)
+        for collection_id in self.collection_ids:
+            download_radiant_mlhub_collection(collection_id, self.root, api_key)
 
         image_archive_path = os.path.join(self.root, self.image_meta["filename"])
         target_archive_path = os.path.join(self.root, self.target_meta["filename"])
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,9 +9,6 @@ updates:
     schedule:
       interval: "daily"
     ignore:
-      # radiant-mlhub 0.5+ changed download behavior:
-      # https://github.com/radiantearth/radiant-mlhub/pull/104
-      - dependency-name: "radiant-mlhub"
       # setuptools releases new versions almost daily
       - dependency-name: "setuptools"
         update-types: ["version-update:semver-patch"]
--- a/tests/datasets/test_cloud_cover.py
+++ b/tests/datasets/test_cloud_cover.py
@@ -15,7 +15,7 @@ from _pytest.monkeypatch import MonkeyPatch
 from torchgeo.datasets import CloudCoverDetection
 
 
-class Dataset:
+class Collection:
     def download(self, output_dir: str, **kwargs: str) -> None:
         glob_path = os.path.join(
             "tests", "data", "ref_cloud_cover_detection_challenge_v1", "*.tar.gz"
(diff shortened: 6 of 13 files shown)


adamjstewart:
We'll need to update all datasets, not just NASA Marine Debris.

> radiant-mlhub says it will auto-resume dataset downloads if requested

Can we just ask to resume whenever `download=True`? Will it fail if the download is 0% or 100% complete and only work when 1<x<99% complete? Or will it handle things gracefully?

environment.yml
```diff
@@ -38,7 +38,7 @@ dependencies:
     - pytorch-lightning>=1.5.1
     - git+https://github.com/pytorch/pytorch_sphinx_theme
     - pyupgrade>=2.4
-    - radiant-mlhub>=0.2.1,<0.5
+    - radiant-mlhub>0.5
```
adamjstewart, line 41:
```suggestion
    - radiant-mlhub>=0.5
```
You'll also need to update `setup.cfg`, `requirements/datasets.txt`, and `requirements/min-reqs.old`

SpontaneousDuck: Got it!

SpontaneousDuck:
Got it, working on all the datatsets now!

SpontaneousDuck:
Ok! So what I ended up doing is falling back to download the old file structure by grabbing the `radiant_mlhub.Collection` instead of the `radiant_mlhub.Dataset`. This basically matches the old way of doing it but calls `tochgeo.datasets.utils.download_radiant_mlhub_collection`  instead to download each individual tarball for each dataset. This allows us to grab files individually and checksum each tarball after downloading.

adamjstewart:
Does that approach work for older radiant-mlhub or is 0.5.0 the minimum version where that technique works?

SpontaneousDuck:
Good idea! Checked with `radiant-mlhub=0.2.2` and this new method still works. Changed `environment.yml` and `setup.cfg` to just remove the version cap instead of bumping the required version

setup.cfg
```diff
@@ -85,9 +85,7 @@ datasets =
     pyvista>=0.25.2,<0.39
     # radiant-mlhub 0.2.1+ required for api_key bugfix:
     # https://github.com/radiantearth/radiant-mlhub/pull/48
-    # radiant-mlhub 0.5+ changed download behavior:
-    # https://github.com/radiantearth/radiant-mlhub/pull/104
-    radiant-mlhub>=0.2.1,<0.5
+    radiant-mlhub>=0.2.1
```
adamjstewart, line 88:
```suggestion
    radiant-mlhub>=0.2.1,<0.6
```
0.6 hasn't been released yet and may break backwards compatibility according to semver.

tests/datasets/test_benin_cashews.py
```diff
@@ -16,15 +16,15 @@
 from torchgeo.datasets import BeninSmallHolderCashews
 
 
-class Dataset:
+class Collection:
     def download(self, output_dir: str, **kwargs: str) -> None:
```
adamjstewart, line 20: When radiant-mlhub 0.5 was first released, we actually didn't notice when things broke because we replace their download method with our own. Do you know if it's possible to monkeypatch something else farther internal so that we can actually test their download method but still use a local file?

torchgeo/datasets/benin_cashews.py
```diff
@@ -17,7 +17,7 @@
 from torch import Tensor
 
 from .geo import NonGeoDataset
-from .utils import check_integrity, download_radiant_mlhub_dataset, extract_archive
+from .utils import check_integrity, download_radiant_mlhub_collection, extract_archive
```
adamjstewart, line 20: Let's remove the `download_radiant_mlhub_dataset` method, it's broken for newer versions of radiant-mlhub

adamjstewart (changes requested):
Can you also update `requirements/datasets.txt` to `radiant-mlhub==0.5.5` and remove radiant-mlhub from `.github/dependabot.yml`?

torchgeo/datasets/benin_cashews.py
SpontaneousDuck: So `download_radiant_mlhub_dataset` is still required by SpaceNet6 as its `Collection` interface is different. That one still works as is now but it does not have md5 checks currently.

tests/datasets/test_benin_cashews.py
SpontaneousDuck: I actually have never used monkeypatch! So I'm not sure how that would work. It looks like the `download` call just creates a `requests. Session` and download chunks for the dataset in a threadpool. We would probably have to over-ride something in requests? But then our test data would need to be chunked I think.

SpontaneousDuck:
Can we loosen the `shapely` requirements for CI?

adamjstewart:
Bleh, that's annoying. Let me see if I can convince them to support newer shapely. I would rather focus on testing newer shapely than on newer radiant-mlhub. So we can test the latest version of radiant-mlhub 0.4 and prevent dependabot from updating it until radiant-mlhub supports shapely 2.

adamjstewart:
Done: https://github.com/radiantearth/radiant-mlhub/pull/195

Hopefully the next release will support newer shapely. Until then, let's keep testing old radiant-mlhub.

SpontaneousDuck:
> Done: [radiantearth/radiant-mlhub#195](https://github.com/radiantearth/radiant-mlhub/pull/195)
> 
> Hopefully the next release will support newer shapely. Until then, let's keep testing old radiant-mlhub.

So that means changing `.github/dependabot.yml` back? Do any other files need to be changed back?

adamjstewart:
You can actually leave your changes to `dependabot.yml`, when the new release comes out we want to try bumping the version we test.

`setup.cfg` and `environment.yml` should be updated, but `requirements/datasets.txt` should not.

requirements/datasets.txt
```diff
@@ -5,7 +5,7 @@ opencv-python==4.7.0.68
 pandas==1.5.3
 pycocotools==2.0.6
 pyvista==0.38.1
-radiant-mlhub==0.4.1
+radiant-mlhub==0.5.5
```
adamjstewart, line 8:
```suggestion
radiant-mlhub==0.4.1
```

setup.cfg
```diff
@@ -85,8 +85,6 @@ datasets =
     pyvista>=0.25.2,<0.39
     # radiant-mlhub 0.2.1+ required for api_key bugfix:
     # https://github.com/radiantearth/radiant-mlhub/pull/48
-    # radiant-mlhub 0.5+ changed download behavior:
-    # https://github.com/radiantearth/radiant-mlhub/pull/104
     radiant-mlhub>=0.2.1,<0.5
```
adamjstewart, line 88:
```suggestion
    radiant-mlhub>=0.2.1,<0.6
```

adamjstewart: Also need to update environment.yml

SpontaneousDuck: Is the above change to environment.yml sufficient?

adamjstewart: I think you understood the reverse of my intentions.

With this PR, the correct dependency versions are `>=0.2.1,<0.6`. This should be the case in all files, except `requirements/datasets.txt` where we still need to test the old version until newer shapely is supported. So you should still remove the upper bound in `environment.yml`

SpontaneousDuck: Got it! Sorry for the confusion. How about now?

environment.yml
```diff
@@ -38,7 +38,7 @@ dependencies:
     - pytorch-lightning>=1.5.1
     - git+https://github.com/pytorch/pytorch_sphinx_theme
     - pyupgrade>=2.4
-    - radiant-mlhub>=0.2.1,<0.5
+    - radiant-mlhub>=0.2.1,<0.6
```
adamjstewart, line 41:
```suggestion
    - radiant-mlhub>=0.2.1
```
We don't usually add upper bounds for conda since this file isn't used in production

adamjstewart:
Everything looks good at first glance, testing this locally now...

adamjstewart:
This is working for most datasets but doesn't seem to work for SpaceNet 6. Maybe @ashnair1 can comment on why only SpaceNet 6 is different.

SpontaneousDuck:
So it looks like SpaceNet provides a direct download link shown here: [https://spacenet.ai/sn6-challenge/](https://spacenet.ai/sn6-challenge/). Since it is 39GB, it looks like ml-hub does not support downloading that dataset directly anymore and only allow the new method of downloading each image individually using the `dataset` API in `radiant-mlhub`. This method as is should work, might just need some retires for data that does not download. Oooor, we could change to directly downloading the 39GB tar file from s3 and using that.

adamjstewart:
I'll let @ashnair1 decide, he wrote all of our SpaceNet data loaders. I think the current approach is probably fine.

adamjstewart (approved):

ashnair1:
So the difference in SpaceNet6 is primarily caused by the way it was archived. Refer to radiantearth/radiant-mlhub#93

While writing this dataset, `download_radiant_mlhub_dataset` worked while `download_radiant_mlhub_collection` did not. It might be possible to download each asset but I haven't tried it out yet.  

Long term, I think it might be best to download all SpaceNet datasets from the official site - https://spacenet.ai/ and not use radiant-mlhub.

adamjstewart:
With this PR both SpaceNet 6 and SpaceNet != 6 seem to download properly for all versions of radiant-mlhub (just ignore the warning messages) so I think this is safe to download. We can consider moving the SpaceNet downloads (or rehosting) later. Would be great if we could download all of our datasets without an API key.

Issue #1042, opened 2023-01-25
Improve model cards for pretrained weights


calebrob6:
With the new v0.4 update to TorchGeo we have the ability to easily add pre-trained model weights (e.g. https://github.com/microsoft/torchgeo/blob/main/torchgeo/models/resnet.py#L28). Currently, each set of weights comes with an [`AugmentationSequential`](https://github.com/microsoft/torchgeo/blob/811a12c105ad839e05343ee243804b3694d36816/torchgeo/transforms/transforms.py#L19) instance that describes the preprocessing that the model expects, but don't include the proper normalization values (or a description of what range of values the model expects as input). These need to be retroactively added for all existing weights, and, ideally, verified (by trying to match reported downstream task performance with the model weights).

For example, the Sentinel 2 models from the SSL4EO papers likely expect uint12 Sentinel 2 values that are divided by 10,000 -- https://github.com/zhu-xlab/SSL4EO-S12/blob/main/src/benchmark/transfer_classification/datasets/EuroSat/eurosat_dataset.py#L97. This step should be added to the  [current set of transforms](https://github.com/microsoft/torchgeo/blob/main/torchgeo/models/resnet.py#L18) and tested.

If verification of the model weights is done on downstream tasks, we should also report those performance numbers vs. relying on the numbers reported in the associated papers.

clkruse:
I'd find this to be quite useful too! Found this issue since I was digging through code to figure out what normalization strategy to use for the SSL4EO pretrained models.

calebrob6:
It seems like this is a popular idea, see https://github.com/zhu-xlab/SSL4EO-S12/issues/2 ;)

PR #1105, opened 2023-02-11
Weights: add sha256 suffix

adamjstewart:
This PR includes the following changes:

- [x] Add the first 8 digits of the sha256 to the filename suffix
- [x] Standardize on .pth file extension
- [x] Sort weights alphabetically
- [x] Keep URLs on a single line

All old weights still exist at the same URLs for backwards compatibility. All new weights should follow this `<name>-<sha256>.pth` format.

References:

* https://github.com/pytorch/vision/issues/7210
* https://github.com/pytorch/vision/pull/7219
* https://pytorch.org/docs/stable/hub.html#torch.hub.load_state_dict_from_url

We won't actually be able to use this checksum until torchvision 0.15.0, but might as well do the heavy lifting now.

### DIFF
--- a/torchgeo/models/resnet.py
+++ b/torchgeo/models/resnet.py
@@ -136,7 +136,7 @@ class ResNet50_Weights(WeightsEnum):  # type: ignore[misc]
     )
 
     SENTINEL2_RGB_SECO = Weights(
-        url="https://huggingface.co/torchgeo/resnet50_sentinel2_rgb_seco/blob/main/resnet50_sentinel2_rgb_seco-584035db.pth",  # noqa: E501
+        url="https://huggingface.co/torchgeo/resnet50_sentinel2_rgb_seco/resolve/main/resnet50_sentinel2_rgb_seco-584035db.pth",  # noqa: E501
         transforms=nn.Identity(),
         meta={
             "dataset": "SeCo Dataset",
--- a/torchgeo/models/resnet.py
+++ b/torchgeo/models/resnet.py
@@ -35,10 +35,7 @@ class ResNet18_Weights(WeightsEnum):  # type: ignore[misc]
     """
 
     SENTINEL2_ALL_MOCO = Weights(
-        url=(
-            "https://huggingface.co/torchgeo/resnet18_sentinel2_all_moco/"
-            "resolve/main/resnet18_sentinel2_all_moco.pth"
-        ),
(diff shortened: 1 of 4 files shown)


torchgeo/models/resnet.py
```diff
@@ -67,10 +61,7 @@ class ResNet18_Weights(WeightsEnum):  # type: ignore[misc]
     )
 
     SENTINEL2_RGB_SECO = Weights(
```
adamjstewart, line 63: This weight has a problem but I can't figure out what it is:
```console
$ pytest -m slow tests/models/test_resnet.py::TestResNet50::test_resnet_download[ResNet50_Weights.SENTINEL2_RGB_SECO]
...
_pickle.UnpicklingError: invalid load key, '<'.
```
The error doesn't occur with the previous weight file, and both the old and new weight files have the same checksum. We'll need to figure out what's going on here before this PR can be merged.

torchgeo/models/resnet.py
```diff
@@ -108,59 +96,47 @@ class ResNet50_Weights(WeightsEnum):  # type: ignore[misc]
...
+            "in_chans": 3,
             "model": "resnet50",
             "publication": "https://arxiv.org/abs/2211.07044",
             "repo": "https://github.com/zhu-xlab/SSL4EO-S12",
-            "ssl_method": "dino",
+            "ssl_method": "moco",
         },
     )
 
     SENTINEL2_RGB_SECO = Weights(
```
adamjstewart, line 138: This weight has a problem but I can't figure out what it is:
```console
$ pytest -m slow tests/models/test_resnet.py::TestResNet50::test_resnet_download[ResNet50_Weights.SENTINEL2_RGB_SECO]
...
_pickle.UnpicklingError: invalid load key, '<'.
```
The error doesn't occur with the previous weight file, and both the old and new weight files have the same checksum. We'll need to figure out what's going on here before this PR can be merged.

nilsleh, line 138: We have defined the `zhu_lab_transforms`, and I was looking into the Seco code again and at least for the bigearthnet datamodule on which they run experiments, augmentations are defined [here](https://github.com/ServiceNow/seasonal-contrast/blob/8285173ec205b64bc3e53b880344dd6c3f79fa7a/datasets/bigearthnet_datamodule.py#L64). Do you want to add those?

adamjstewart: Yes! Can you add it in a separate PR though? That can fix #1042.

nilsleh (approved):

PR #1117, opened 2023-02-15
Chesapeake CVPR experiment: fix import

adamjstewart:
Still not quite sure why this happened, but I ran `pytest` on our new cluster and it picked up the `test_chesapeakecvpr_models.py` script and failed on an import that no longer exists. Renamed the file to avoid this and fixed the import.


calebrob6 (approved):

calebrob6:
that's weird

adamjstewart:
I think it was picked up because I did `pytest .` instead of `pytest` and the explicit path overrides the path in `testpaths` and then recursively searches for any file named `test_*.py` or `*_test.py`. But we should make it work regardless.

PR #1119, opened 2023-02-17
Add seco transforms and zhu normalization to pretrained weights

### LINKED ISSUE #1042: Improve model cards for pretrained weights

nilsleh:
This PR closes #1042 by adding the normalization transform to the Zhu_lab pretrained weights and a transform to the seco weights.


torchgeo/models/resnet.py
```diff
@@ -16,7 +16,51 @@
 __all__ = ["ResNet50_Weights", "ResNet18_Weights"]
 
 _zhu_xlab_transforms = AugmentationSequential(
```
adamjstewart, line 18: If you want to add links to where the transforms come from, you can use:

https://github.com/zhu-xlab/SSL4EO-S12/blob/d2868adfada65e40910bfcedfc49bc3b20df2248/src/benchmark/transfer_classification/linear_BE_moco.py#L167

(I think they are all the same?)

Normalization seems to be either divide by 10K or apply channel-wise normalizations: 

https://github.com/zhu-xlab/SSL4EO-S12/blob/d2868adfada65e40910bfcedfc49bc3b20df2248/src/benchmark/transfer_classification/datasets/EuroSat/eurosat_dataset.py#L95,L97

We could try both and default to whichever works better.

torchgeo/models/resnet.py
```diff
@@ -16,7 +16,51 @@
...
-    K.Resize(256), K.CenterCrop(224), data_keys=["image"]
+    K.Resize(256),
+    K.CenterCrop(224),
+    K.Normalize(mean=0, std=10000),
+    data_keys=["image"],
+)
+
+# https://github.com/ServiceNow/seasonal-contrast/
+# blob/8285173ec205b64bc3e53b880344dd6c3f79fa7a/datasets/bigearthnet_dataset.py#L13
```
adamjstewart, line 26: I dislike breaking long URLs across multiple lines. You can usually add `# noqa: E501` to the end of the line to get flake8 to ignore it. I only do this in situations where readability is improved by having a long line. Doing the same thing to our weight URLs in my weight rename PR.

torchgeo/models/resnet.py
nilsleh: I think @calebrob6 ran an experiment where he found that divide by 10K seems to work well. Not sure if he has tried anything else. 

I also understand that the transforms are the same across their models.

adamjstewart: Let's also add a comment for https://github.com/zhu-xlab/SSL4EO-S12/blob/d2868adfada65e40910bfcedfc49bc3b20df2248/src/benchmark/transfer_classification/datasets/EuroSat/eurosat_dataset.py#L97

torchgeo/models/resnet.py
```diff
@@ -7,16 +7,63 @@
...
 from torchvision.models._api import Weights, WeightsEnum
 
 from ..transforms import AugmentationSequential
 
 __all__ = ["ResNet50_Weights", "ResNet18_Weights"]
 
+
+# https://github.com/zhu-xlab/SSL4EO-S12/blob/d2868adfada65e40910bfcedfc49bc3b20df2248/src/benchmark/transfer_classification/linear_BE_moco.py#L167 # noqa: E501
+# https://github.com/zhu-xlab/SSL4EO-S12/blob/d2868adfada65e40910bfcedfc49bc3b20df2248/src/benchmark/transfer_classification/datasets/EuroSat/eurosat_dataset.py#L97 # noqa: E501
+# Normalization either by 10K or channel-wise with band statistics
```
adamjstewart, line 21: Before we merge this we should get a sense for either which the authors recommend (what do they prefer in their own code or paper or experiments?) or which seems to work better (maybe @calebrob6 can run some experiments with both)

calebrob6: authors confirmed that they / 10000 -- https://github.com/zhu-xlab/SSL4EO-S12/issues/2

adamjstewart (approved):

PR #1125, opened 2023-02-18
Fix broken URLs

adamjstewart:
TIL you can run:
```console
$ cd docs
$ make linkcheck
```
and Sphinx will parse and check all external links in your docs and warn you which links are broken. Unfortunately there are too many false positives to make this a CI test, but this may be something we want to check before each release.

### DIFF
--- a/docs/tutorials/trainers.ipynb
+++ b/docs/tutorials/trainers.ipynb
@@ -19,7 +19,7 @@
    "source": [
     "# PyTorch Lightning Trainers\n",
     "\n",
-    "In this tutorial, we demonstrate TorchGeo trainers to train and test a model. Specifically, we use the [Tropical Cyclone dataset](https://torchgeo.readthedocs.io/en/latest/api/datasets.html#tropical-cyclone-wind-estimation-competition) and train models to predict cyclone wind speed given imagery of the cyclone. \n",
+    "In this tutorial, we demonstrate TorchGeo trainers to train and test a model. Specifically, we use the [Tropical Cyclone dataset](https://torchgeo.readthedocs.io/en/latest/api/datasets.html#tropical-cyclone) and train models to predict cyclone wind speed given imagery of the cyclone.\n",
     "\n",
     "It's recommended to run this notebook on Google Colab if you don't have your own GPU. Click the \"Open in Colab\" button above to get started."
    ]
--- a/torchgeo/datasets/chesapeake.py
+++ b/torchgeo/datasets/chesapeake.py
@@ -28,8 +28,8 @@ from .utils import BoundingBox, download_url, extract_archive
 class Chesapeake(RasterDataset, abc.ABC):
     """Abstract base class for all Chesapeake datasets.
 
-    `Chesapeake Bay Program Land Use/Land Cover Data Project
-    <https://www.chesapeakeconservancy.org/conservation-innovation-center/high-resolution-data/lulc-data-project-2022/>`_
+    `Chesapeake Bay High-Resolution Land Cover Project
+    <https://www.chesapeakeconservancy.org/conservation-innovation-center/high-resolution-data/land-cover-data-project/>`_
     dataset.
 
     This dataset was collected by the Chesapeake Conservancy's Conservation Innovation
--- a/torchgeo/datasets/chesapeake.py
+++ b/torchgeo/datasets/chesapeake.py
@@ -28,23 +28,14 @@ from .utils import BoundingBox, download_url, extract_archive
 class Chesapeake(RasterDataset, abc.ABC):
     """Abstract base class for all Chesapeake datasets.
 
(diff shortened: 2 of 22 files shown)


docs/tutorials/trainers.ipynb
```diff
@@ -19,7 +19,7 @@
    "source": [
     "# PyTorch Lightning Trainers\n",
     "\n",
-    "In this tutorial, we demonstrate TorchGeo trainers to train and test a model. Specifically, we use the [Tropical Cyclone dataset](https://torchgeo.readthedocs.io/en/latest/api/datasets.html#tropical-cyclone-wind-estimation-competition) and train models to predict cyclone wind speed given imagery of the cyclone. \n",
```
adamjstewart, line 22: Dataset was renamed and link broke

torchgeo/datasets/chesapeake.py
```diff
@@ -28,23 +28,14 @@
 class Chesapeake(RasterDataset, abc.ABC):
     """Abstract base class for all Chesapeake datasets.
 
-    `Chesapeake Bay High-Resolution Land Cover Project
-    <https://www.chesapeakeconservancy.org/conservation-innovation-center/high-resolution-data/land-cover-data-project/>`_
+    `Chesapeake Bay Program Land Use/Land Cover Data Project
+    <https://www.chesapeakeconservancy.org/conservation-innovation-center/high-resolution-data/lulc-data-project-2022/>`_
```
adamjstewart, line 32: @calebrob6 I didn't look into this too much but it seems like there's a new version of this dataset. Not sure if we should update the link or link to the old one instead, I don't know how different they are. Let me know if I should revert this.

torchgeo/datasets/chesapeake.py
```diff
@@ -28,23 +28,14 @@
...
     Bay watershed (~100,000 square miles of land).
-
-    For more information, see:
-
-    * `User Guide
-      <https://chesapeakeconservancy.org/wp-content/uploads/2017/01/LandCover101Guide.pdf>`_
-    * `Class Descriptions
-      <https://chesapeakeconservancy.org/wp-content/uploads/2020/03/LC_Class_Descriptions.pdf>`_
-    * `Accuracy Assessment
-      <https://chesapeakeconservancy.org/wp-content/uploads/2017/01/Chesapeake_Conservancy_Accuracy_Assessment_Methodology.pdf>`_
```
adamjstewart, line 47: Some of these links are broken. Most of these docs are linked on the main dataset page, so I don't think we need to explicitly reference them.

torchgeo/datasets/enviroatlas.py
```diff
@@ -35,7 +35,7 @@ class EnviroAtlas(GeoDataset):
     This dataset was organized to accompany the 2022 paper, `"Resolving label
     uncertainty with implicit generative models"
     <https://openreview.net/forum?id=AEa_UepnMDX>`_. More details can be found at
-    https://github.com/estherrolf/qr_for_landcover.
+    https://github.com/estherrolf/implicit-posterior.
```
adamjstewart, line 38: Many of these are repo renames. The old URL still works and redirects to the correct page, but figured we should update them anyway.

torchgeo/datasets/vhr10.py
```diff
@@ -159,10 +159,7 @@ class VHR10(NonGeoDataset):
         "md5": "d30a7ff99d92123ebb0b3a14d9102081",
     }
     target_meta = {
-        "url": (
-            "https://raw.githubusercontent.com/chaozhong2010/VHR-10_dataset_coco/"
-            "master/NWPU%20VHR-10_dataset_coco/annotations.json"
-        ),
+        "url": "https://raw.githubusercontent.com/chaozhong2010/VHR-10_dataset_coco/ce0ba0f5f6a0737031f1cbe05e785ddd5ef05bd7/NWPU%20VHR-10_dataset_coco/annotations.json",  # noqa: E501
```
adamjstewart, line 162: We should never download files from the master branch, they may change or move to a different location. Instead, we should use permalinks when available.

torchgeo/datasets/chesapeake.py
calebrob6: It looks like the new version has different classes

torchgeo/datasets/chesapeake.py
calebrob6: Fine with me

calebrob6 (approved):

torchgeo/datasets/chesapeake.py
adamjstewart: Alright, let me undo this change and we can create a v2 someday if we want.

adamjstewart:
Tests aren't going to pass until the notebook tests are passing. But that test isn't required, this PR can be merged whenever.

calebrob6 (approved):

PR #1126, opened 2023-02-19
LandCover.ai: add GeoDataset version

adrianboguszewski:
This PR renames LandCoverAI class to LandCoverAINonGeo (benchmark dataset, created with `split.py` script) and introduces the abstract base LandCoverAI class and  LandCoverAIGeo (raster dataset).

This is my proposition, but I am open to the discussion of how the structure, inheritance, and naming should look.

### DIFF
--- a/docs/api/datasets.rst
+++ b/docs/api/datasets.rst
@@ -94,7 +94,7 @@ iNaturalist
 .. autoclass:: INaturalist
 
 LandCover.ai
-^^^^^^^^^^^^^^^^
+^^^^^^^^^^^^
 
 .. autoclass:: LandCoverAI
 .. autoclass:: LandCoverAIGeo
@@ -222,7 +222,7 @@ Inria Aerial Image Labeling
 .. autoclass:: InriaAerialImageLabeling
 
 LandCover.ai
-^^^^^^^^^^^^^^^^
+^^^^^^^^^^^^^
 
 .. autoclass:: LandCoverAI
 .. autoclass:: LandCoverAINonGeo
--- a/docs/api/datasets.rst
+++ b/docs/api/datasets.rst
@@ -387,3 +387,12 @@ Collation Functions
 .. autofunction:: concat_samples
 .. autofunction:: merge_samples
 .. autofunction:: unbind_samples
+
+Splitting Functions
+^^^^^^^^^^^^^^^^^^^
+
+.. autofunction:: random_bbox_assignment
+.. autofunction:: random_bbox_splitting
+.. autofunction:: random_grid_cell_assignment
+.. autofunction:: roi_split
+.. autofunction:: time_series_split
--- a/docs/api/datasets.rst
+++ b/docs/api/datasets.rst
@@ -93,6 +93,12 @@ iNaturalist
 
 .. autoclass:: INaturalist
 
+LandCover.ai
+^^^^^^^^^^^^^^^^
+
+.. autoclass:: LandCoverAI
+.. autoclass:: LandCoverAIGeo
+
 Landsat
 ^^^^^^^
 
@@ -216,9 +222,10 @@ Inria Aerial Image Labeling
 .. autoclass:: InriaAerialImageLabeling
 
 LandCover.ai
-^^^^^^^^^^^^
+^^^^^^^^^^^^^^^^
 
 .. autoclass:: LandCoverAI
+.. autoclass:: LandCoverAINonGeo
 
 LEVIR-CD+
 ^^^^^^^^^
--- a/docs/api/datasets.rst
+++ b/docs/api/datasets.rst
@@ -93,10 +93,10 @@ iNaturalist
 
 .. autoclass:: INaturalist
 
-LandCover.ai
-^^^^^^^^^^^^
+LandCover.ai Geo
+^^^^^^^^^^^^^^^^
 
-.. autoclass:: LandCoverAI
+.. autoclass:: LandCoverAIBase
 .. autoclass:: LandCoverAIGeo
 
 Landsat
@@ -222,10 +222,9 @@ Inria Aerial Image Labeling
 .. autoclass:: InriaAerialImageLabeling
 
 LandCover.ai
-^^^^^^^^^^^^^
+^^^^^^^^^^^^
 
 .. autoclass:: LandCoverAI
-.. autoclass:: LandCoverAINonGeo
 
 LEVIR-CD+
 ^^^^^^^^^
--- a/docs/api/geo_datasets.csv
+++ b/docs/api/geo_datasets.csv
@@ -12,6 +12,7 @@ Dataset,Type,Source,Size (px),Resolution (m)
 `GBIF`_,Points,Citizen Scientists,-,-
 `GlobBiomass`_,Masks,Landsat,"45,000x45,000",100
 `iNaturalist`_,Points,Citizen Scientists,-,-
+`LandCover.ai`_,"Imagery, Masks",Aerial,-,0.25--0.5
 `Landsat`_,Imagery,Landsat,"8,900x8,900",30
 `NAIP`_,Imagery,Aerial,"6,100x7,600",1
 `Open Buildings`_,Geometries,"Maxar, CNES/Airbus",-,-
--- a/docs/api/geo_datasets.csv
+++ b/docs/api/geo_datasets.csv
@@ -12,7 +12,7 @@ Dataset,Type,Source,Size (px),Resolution (m)
 `GBIF`_,Points,Citizen Scientists,-,-
 `GlobBiomass`_,Masks,Landsat,"45,000x45,000",100
 `iNaturalist`_,Points,Citizen Scientists,-,-
-`LandCover.ai`_,"Imagery, Masks",Aerial,-,0.25--0.5
+`LandCover.ai Geo`_,"Imagery, Masks",Aerial,-,0.25--0.5
 `Landsat`_,Imagery,Landsat,"8,900x8,900",30
 `NAIP`_,Imagery,Aerial,"6,100x7,600",1
 `Open Buildings`_,Geometries,"Maxar, CNES/Airbus",-,-
--- a/docs/api/geo_datasets.csv
+++ b/docs/api/geo_datasets.csv
@@ -12,7 +12,7 @@ Dataset,Type,Source,Size (px),Resolution (m)
 `GBIF`_,Points,Citizen Scientists,-,-
 `GlobBiomass`_,Masks,Landsat,"45,000x45,000",100
 `iNaturalist`_,Points,Citizen Scientists,-,-
-`LandCover.ai Geo`_,"Imagery, Masks",Aerial,-,0.25--0.5
+`LandCover.ai Geo`_,"Imagery, Masks",Aerial,"~4,200x4,700-~9000x9500",0.25--0.5
 `Landsat`_,Imagery,Landsat,"8,900x8,900",30
 `NAIP`_,Imagery,Aerial,"6,100x7,600",1
 `Open Buildings`_,Geometries,"Maxar, CNES/Airbus",-,-
--- a/docs/api/geo_datasets.csv
+++ b/docs/api/geo_datasets.csv
@@ -12,7 +12,7 @@ Dataset,Type,Source,Size (px),Resolution (m)
 `GBIF`_,Points,Citizen Scientists,-,-
 `GlobBiomass`_,Masks,Landsat,"45,000x45,000",100
 `iNaturalist`_,Points,Citizen Scientists,-,-
-`LandCover.ai Geo`_,"Imagery, Masks",Aerial,"~4,200x4,700-~9000x9500",0.25--0.5
+`LandCover.ai Geo`_,"Imagery, Masks",Aerial,"4,200--9,500",0.25--0.5
 `Landsat`_,Imagery,Landsat,"8,900x8,900",30
 `NAIP`_,Imagery,Aerial,"6,100x7,600",1
 `Open Buildings`_,Geometries,"Maxar, CNES/Airbus",-,-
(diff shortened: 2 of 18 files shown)


adrianboguszewski:

torchgeo/datasets/landcoverai.py
```diff
@@ -50,7 +52,223 @@ class LandCoverAI(NonGeoDataset):
 
     If you use this dataset in your research, please cite the following paper:
 
-    * https://arxiv.org/abs/2005.02264v3
+    * https://openaccess.thecvf.com/content/CVPR2021W/EarthVision/html/Boguszewski_LandCover.ai_Dataset_for_Automatic_Mapping_of_Buildings_Woodlands_Water_and_CVPRW_2021_paper.html
```
adamjstewart, line 55:
```suggestion
    * https://doi.org/10.1109/cvprw53098.2021.00121
```
Links to the same paper but with a shorter URL. This should fix the flake8 tests.

torchgeo/datasets/landcoverai.py
```diff
@@ -50,7 +52,223 @@ class LandCoverAI(NonGeoDataset):
...
+            if show_titles:
+                axs[2].set_title("Predictions")
+
+        if suptitle is not None:
+            plt.suptitle(suptitle)
+        return fig
+
+
+class LandCoverAIGeo(LandCoverAI, RasterDataset):
+    r"""LandCover.ai Geo dataset.
```
adamjstewart, line 186:
```suggestion
    """LandCover.ai Geo dataset.
```
No need to make it a raw string

torchgeo/datasets/landcoverai.py
```diff
@@ -50,7 +52,223 @@ class LandCoverAI(NonGeoDataset):
...
+        }
+
+        if self.transforms is not None:
+            sample = self.transforms(sample)
+
+        return sample
+
+
+class LandCoverAINonGeo(LandCoverAI, NonGeoDataset):
+    r"""LandCover.ai NonGeo dataset.
```
adamjstewart, line 269:
```suggestion
    """LandCover.ai NonGeo dataset.
```

docs/api/datasets.rst
```diff
@@ -93,6 +93,12 @@ iNaturalist
 
 .. autoclass:: INaturalist
 
+LandCover.ai
+^^^^^^^^^^^^
+
+.. autoclass:: LandCoverAI
```
adamjstewart, line 99: RtD tests are failing because this class is documented twice (Sphinx doesn't know which place to link to). Maybe only document it here and remove the reference below.

docs/api/datasets.rst
```diff
@@ -93,6 +93,12 @@ iNaturalist
 
 .. autoclass:: INaturalist
 
+LandCover.ai
```
adamjstewart, line 96: This name will need to be unique, this is the link that the names in the CSV files search for

torchgeo/datasets/landcoverai.py
```diff
@@ -170,47 +370,21 @@ def _load_target(self, id_: str) -> Tensor:
...
-        self._extract()
-
-    def _download(self) -> None:
-        """Download the dataset."""
-        download_url(self.url, self.root, md5=self.md5 if self.checksum else None)
+    def _verify_data(self) -> bool:
+        """Verify if the images and masks are present."""
+        images = os.path.join(self.root, "output", "*_*.jpg")
+        masks = os.path.join(self.root, "output", "*_*_m.png")
+        return len(glob.glob(images)) > 0 and len(glob.glob(images)) == len(
```
adamjstewart, line 377: This globs the images list twice, let's glob it once and save that as the `images` variable to speed this up

torchgeo/datasets/landcoverai.py
```diff
@@ -50,7 +52,223 @@ class LandCoverAI(NonGeoDataset):
...
+
+
+class LandCoverAIGeo(LandCoverAI, RasterDataset):
+    r"""LandCover.ai Geo dataset.
+
+    See the abstract base LandCoverAI class to find out more.
+    """
+    filename_glob = "images/*.tif"
+    filename_regex = ".*tif"
+    crs = CRS.from_epsg(2180)
```
adamjstewart, line 192: This is automatically detected from the files, you can remove this. But we should add a `crs` variant to the class so that users can easily change it.

torchgeo/datasets/landcoverai.py
```diff
@@ -50,7 +52,223 @@ class LandCoverAI(NonGeoDataset):
...
+            plt.suptitle(suptitle)
+        return fig
+
+
+class LandCoverAIGeo(LandCoverAI, RasterDataset):
+    r"""LandCover.ai Geo dataset.
+
+    See the abstract base LandCoverAI class to find out more.
+    """
+    filename_glob = "images/*.tif"
```
adamjstewart, line 190:
```suggestion
    filename_glob = os.path.join("images", "*.tif")
```
This will work better on Windows.

I've never tried using a path in `filename_glob` before but I think it should work.

torchgeo/datasets/landcoverai.py
```diff
@@ -50,7 +52,223 @@ class LandCoverAI(NonGeoDataset):
...
+
+    def _verify_data(self) -> bool:
+        """Verify if the images and masks are present."""
+        images = os.path.join(self.root, "images", "*.tif")
+        masks = os.path.join(self.root, "masks", "*.tif")
+        return len(glob.glob(images)) > 0 and len(glob.glob(images)) == len(
+            glob.glob(masks)
+        )
+
+    def __getitem__(self, query: BoundingBox) -> Dict[str, Any]:
```
adamjstewart, line 232: This looks like exactly what we want, I'll point contributors to this dataset if they encounter https://github.com/microsoft/torchgeo/discussions/1104#discussioncomment-5001426

adamjstewart (commented):
This looks correct modulo some documentation hiccups. Let me mull over the class names. Some options:

* LandCoverAINonGeo and LandCoverAIGeo
* LandCoverAI and LandCoverAIGeo
* LandCoverAIBenchmark and LandCoverAIUncurated

I think this will be the first dataset where we have both a geo and non-geo version, so we have an opportunity to set a new standard naming scheme.

calebrob6:
For naming, I prefer something like `LandCoverAI` for the GeoDataset and `LandCoverAIFixed` / `LandCoverAIPatched` for the NonGeoDataset.

adamjstewart:
I think @isaaccorley said he prefers LandCoverAI for non-geo and LandCoverAIGeo for geo. Not sure how to come to agreement on the names. Would be great if we can decide on an approach that is consistent for all datasets, there are many others that could be both geo and non-geo.

calebrob6:
Yeah on second thought I like that (LandCoverAI for non-geo and LandCoverAIGeo for geo) a bit better:
- Doesn't break existing code
- It makes the non-geo dataset seem like the default (which is likely what people want if they're just trying to compare to published numbers)

torchgeo/datasets/landcoverai.py
adrianboguszewski: I prefer open access to our paper, so I will restore the arxiv link. I don't like the IEEE paywall ;)

adrianboguszewski:
Thank you guys for the review and a valuable name discussion. `backwards-incompatible` label can be removed.

torchgeo/datasets/landcoverai.py
adamjstewart: Agreed. If you really want to keep the open access version, you could add `  # noqa: E501` to the closing `"""` so that flake8 ignores the long line. But arXiv is also fine with me.

docs/api/geo_datasets.csv
```diff
@@ -12,6 +12,7 @@ Dataset,Type,Source,Size (px),Resolution (m)
 `GBIF`_,Points,Citizen Scientists,-,-
 `GlobBiomass`_,Masks,Landsat,"45,000x45,000",100
 `iNaturalist`_,Points,Citizen Scientists,-,-
+`LandCover.ai Geo`_,"Imagery, Masks",Aerial,-,0.25--0.5
```
adamjstewart, line 15: Can you add the size of each image (in pixels)? Can be a range if it varies by a lot

torchgeo/datasets/landcoverai.py
```diff
@@ -2,31 +2,33 @@
...
+from .geo import NonGeoDataset, RasterDataset
+from .utils import BoundingBox, download_url, extract_archive, working_dir
 
 
-class LandCoverAI(NonGeoDataset):
-    r"""LandCover.ai dataset.
+class LandCoverAIBase(Dataset[Dict[str, Any]], abc.ABC):
+    r"""Abstract base class for LandCover.ai Geo and NonGeo datasets.
 
     The `LandCover.ai <https://landcover.ai/>`__ (Land Cover from Aerial Imagery)
```
adamjstewart, line 28: Can you update this URL to the new URL? I did this in https://github.com/microsoft/torchgeo/pull/1125/files#diff-cfa6063281a8e2b83b8fa6c9304b312bd95a6e2d29d206b4c4b062995cc6c8cf but it's probably easier to do it here.

torchgeo/datasets/landcoverai.py
```diff
@@ -50,7 +52,224 @@ class LandCoverAI(NonGeoDataset):
...
+
+        if suptitle is not None:
+            plt.suptitle(suptitle)
+        return fig
+
+
+class LandCoverAIGeo(LandCoverAIBase, RasterDataset):
+    """LandCover.ai Geo dataset.
+
+    See the abstract LandCoverAIBase class to find out more.
```
adamjstewart, line 188: Can you add:
```rst
.. versionadded:: 0.5
```
This will document that the dataset didn't exist until torchgeo 0.5.0

torchgeo/datasets/landcoverai.py
```diff
@@ -50,7 +52,224 @@ class LandCoverAI(NonGeoDataset):
 
     If you use this dataset in your research, please cite the following paper:
 
-    * https://arxiv.org/abs/2005.02264v3
+    * https://arxiv.org/abs/2005.02264v4
```
adamjstewart, line 55: Can you add:
```rst
.. versionadded:: 0.5
```
This will document that the base class didn't exist until torchgeo 0.5.0

torchgeo/datasets/landcoverai.py
```diff
@@ -50,7 +52,224 @@ class LandCoverAI(NonGeoDataset):
...
+
+        Args:
+            sample: a sample returned by :meth:`__getitem__`
+            show_titles: flag indicating whether to show titles above each panel
+            suptitle: optional string to use as a suptitle
+
+        Returns:
+            a matplotlib Figure with the rendered sample
+
+        .. versionadded:: 0.2
```
adamjstewart, line 152: You can probably remove this line, the entire class didn't exist until 0.5

torchgeo/datasets/landcoverai.py
```diff
@@ -2,31 +2,33 @@
...
+from .utils import BoundingBox, download_url, extract_archive, working_dir
 
 
-class LandCoverAI(NonGeoDataset):
-    r"""LandCover.ai dataset.
+class LandCoverAIBase(Dataset[Dict[str, Any]], abc.ABC):
+    r"""Abstract base class for LandCover.ai Geo and NonGeo datasets.
 
-    The `LandCover.ai <https://landcover.ai/>`__ (Land Cover from Aerial Imagery)
+    The `LandCover.ai <https://landcover.ai.linuxpolska.com/>`__ (Land Cover from Aerial Imagery)
```
adamjstewart, line 28: flake8 is complaining about this line being too long, can you wrap it onto the next line?

torchgeo/datasets/landcoverai.py
```diff
@@ -50,7 +52,227 @@ class LandCoverAI(NonGeoDataset):
...
+    def plot(
+        self,
+        sample: Dict[str, Tensor],
+        show_titles: bool = True,
+        suptitle: Optional[str] = None,
+    ) -> plt.Figure:
+        """Plot a sample from the dataset.
+
+        Args:
+            sample: a sample returned by :meth:`__getitem__`
```
adamjstewart, line 148: RtD tests are failing because this method isn't defined. Can you add an abstract method that the subclasses are responsible for overriding? COWC has an example of this.

torchgeo/datasets/landcoverai.py
adrianboguszewski: Yeah, I missed that

torchgeo/datasets/landcoverai.py
adamjstewart: Bleh. Hmm... Here are some things you can try, in order of preference.

1. Make the input type `Union[BoundingBox, int]`
2. Make the input type `Any`
3. Use `typing.overload`
4. Give up on abc's and convert ``:meth:`__getitem__` `` to ``` ``__getitem__`` ```

adrianboguszewski: It has to be Any. Otherwise, mypy returns an error related to inheritance.

docs/api/geo_datasets.csv
```diff
@@ -12,6 +12,7 @@ Dataset,Type,Source,Size (px),Resolution (m)
 `GBIF`_,Points,Citizen Scientists,-,-
 `GlobBiomass`_,Masks,Landsat,"45,000x45,000",100
 `iNaturalist`_,Points,Citizen Scientists,-,-
+`LandCover.ai Geo`_,"Imagery, Masks",Aerial,"~4,200x4,700-~9000x9500",0.25--0.5
```
adamjstewart, line 15:
```suggestion
`LandCover.ai Geo`_,"Imagery, Masks",Aerial,"4,200--9,500",0.25--0.5
```
Just to make it shorter so it doesn't stretch onto multiple lines.

adamjstewart (approved):

PR #1127, opened 2023-02-20
Change BigEarthNet class mapping order to match original code

lucastao:
It seems like the mapping from 43 to 19 classes in the BigEarthNet dataset code is incorrect. 

**Example**
The below image corresponds to the RGB visualization of S2B_MSIL2A_20180502T093039_42_25. The provided textual labels are 'Broad-leaved forest', 'Natural grassland', and 'Transitional woodland/shrub'. Per the mapping provided [here](https://github.com/microsoft/torchgeo/blob/main/torchgeo/datasets/bigearthnet.py#L145-L190) , this corresponds to indices 5, 24, and 39. After converting these from 43 class labels to 19 class labels using the [label converter](https://github.com/microsoft/torchgeo/blob/main/torchgeo/datasets/bigearthnet.py#L193-L226), we end up with 5 being discarded, 24 (Natural grassland) mapping to 10 (Mixed forest) and 39 (Transitional woodland/shrub) mapping to 17 (Inland waters). All three of these are incorrect. 5 (Broad-leaved forest) should be mapped to 8 (Broad-leaved forest), 24 (Natural grassland) should actually map to 11 (Natural grassland and sparsely vegetated areas) and 39 (Transitional woodland/shrub) should map to 13 (Transitional woodland, shrub).

S2B_MSIL2A_20180502T093039_42_25

**Root Cause and Fix**
The root cause of this issue appears to be an alphabetically sorted list here: https://github.com/microsoft/torchgeo/blob/main/torchgeo/datasets/bigearthnet.py#L145-L190 when it should instead match the order provided by the BigEarthNet code: https://git.tu-berlin.de/rsim/BigEarthNet-S2_19-classes_models/-/blob/master/label_indices.json#L2-46. 

1. Look at the BigEarthNet code for 43 to 19 class mappings provided at https://bigearth.net/. Specifically, this file: https://git.tu-berlin.de/rsim/BigEarthNet-S2_19-classes_models/-/blob/master/label_indices.json#L2-46. Here, the mapping are not sorted in alphabetical order. 

2. Look at the BigEarthNet code for 43 to 19 class mappings in torchgeo: https://github.com/microsoft/torchgeo/blob/main/torchgeo/datasets/bigearthnet.py#L145-L190. These are sorted in alphabetical order. 

3. However, the mappings from 43 to 19 classes remain index based and are the same as the indexes used in: https://github.com/microsoft/torchgeo/blob/main/torchgeo/datasets/bigearthnet.py#L193-L226 and https://git.tu-berlin.de/rsim/BigEarthNet-S2_19-classes_models/-/blob/master/label_indices.json#L47-67. To fix this, either the ordering of the 43 classes or the indexing dictionary must be changed to fix this mismatch.

This commit adjusts the order of the 43 classes to match the original. After testing, I find that the above example now has the correct mapping: 
```
Textual label: ['Broad-leaved forest', 'Natural grassland', 'Transitional woodland/shrub']
43 class: [22, 25, 28]
19 class: [8, 11, 13]
```

### DIFF
--- a/torchgeo/datasets/bigearthnet.py
+++ b/torchgeo/datasets/bigearthnet.py
@@ -45,50 +45,50 @@ class BigEarthNet(NonGeoDataset):
 
     Dataset classes (43):
 
-    0. Agro-forestry areas
-    1. Airports
-    2. Annual crops associated with permanent crops
-    3. Bare rock
-    4. Beaches, dunes, sands
-    5. Broad-leaved forest
-    6. Burnt areas
-    7. Coastal lagoons
-    8. Complex cultivation patterns
-    9. Coniferous forest
-    10. Construction sites
-    11. Continuous urban fabric
-    12. Discontinuous urban fabric
-    13. Dump sites
-    14. Estuaries
-    15. Fruit trees and berry plantations
-    16. Green urban areas
-    17. Industrial or commercial units
-    18. Inland marshes
-    19. Intertidal flats
-    20. Land principally occupied by agriculture, with significant
+    0: Continuous urban fabric
+    1: Discontinuous urban fabric
+    2: Industrial or commercial units
+    3: Road and rail networks and associated land
+    4: Port areas
+    5: Airports
+    6: Mineral extraction sites
+    7: Dump sites
+    8: Construction sites
+    9: Green urban areas
+    10: Sport and leisure facilities
+    11: Non-irrigated arable land
+    12: Permanently irrigated land
+    13: Rice fields
+    14: Vineyards
+    15: Fruit trees and berry plantations
+    16: Olive groves
+    17: Pastures
+    18: Annual crops associated with permanent crops
+    19: Complex cultivation patterns
+    20: Land principally occupied by agriculture, with significant
         areas of natural vegetation
-    21. Mineral extraction sites
-    22. Mixed forest
-    23. Moors and heathland
-    24. Natural grassland
-    25. Non-irrigated arable land
-    26. Olive groves
-    27. Pastures
-    28. Peatbogs
-    29. Permanently irrigated land
-    30. Port areas
-    31. Rice fields
-    32. Road and rail networks and associated land
-    33. Salines
-    34. Salt marshes
-    35. Sclerophyllous vegetation
-    36. Sea and ocean
-    37. Sparsely vegetated areas
-    38. Sport and leisure facilities
-    39. Transitional woodland/shrub
-    40. Vineyards
-    41. Water bodies
-    42. Water courses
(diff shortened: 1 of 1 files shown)


lucastao:

lucastao:

torchgeo/datasets/bigearthnet.py
```diff
@@ -45,50 +45,50 @@ class BigEarthNet(NonGeoDataset):
...
-    12. Discontinuous urban fabric
-    13. Dump sites
-    14. Estuaries
-    15. Fruit trees and berry plantations
-    16. Green urban areas
-    17. Industrial or commercial units
-    18. Inland marshes
-    19. Intertidal flats
-    20. Land principally occupied by agriculture, with significant
+    0: Continuous urban fabric
```
adamjstewart, line 48: Can you use periods instead of colons? Then Sphinx will convert this to a numbered list.

torchgeo/datasets/bigearthnet.py
```diff
@@ -143,50 +143,50 @@ class BigEarthNet(NonGeoDataset):
             "Marine waters",
```
adamjstewart, line 143: Can you add a comment pointing to https://git.tu-berlin.de/rsim/BigEarthNet-S2_19-classes_models/-/blob/master/label_indices.json at the top of all of this so we know where the definitive source is?

torchgeo/datasets/bigearthnet.py
```diff
@@ -113,6 +113,11 @@ class BigEarthNet(NonGeoDataset):
     17. Inland waters
     18. Marine waters
 
+    The source for the above dataset classes, their respective ordering, and
+    43-to-19-class mappings can be found here:
+
+    * https://git.tu-berlin.de/rsim/BigEarthNet-S2_19-classes_models/-/blob/master/label_indices.json
```
adamjstewart, line 119: flake8 is failing because it doesn't like the long line. If you add `  # noqa: E501` at the end of the closing `"""` it should ignore this.

torchgeo/datasets/bigearthnet.py
```diff
@@ -113,11 +113,16 @@ class BigEarthNet(NonGeoDataset):
...
+
     If you use this dataset in your research, please cite the following paper:
 
     * https://doi.org/10.1109/IGARSS.2019.8900532
 
-    """
+    """ # noqa: E501
```
adamjstewart, line 125:
```suggestion
    """  # noqa: E501
```

adamjstewart (approved):

calebrob6:
Thanks @lucastao!

Issue #1074, opened 2023-01-31
Trainers tutorial: switch dataset


adamjstewart:
### Summary

We should switch the dataset for our [PyTorch Lightning Trainers](https://torchgeo.readthedocs.io/en/stable/tutorials/trainers.html) tutorial.

### Rationale

Our current dataset is downloaded from Radiant Earth, requiring additional dependencies and preventing users from running the tutorial without first signing up for an account and modifying the notebook. It would be better to use a dataset that does not require these things so that the tutorial works out of the box.

### Implementation

Most of our other current tutorials use EuroSAT, so we could use that.

### Alternatives

Should we just choose the smallest and fastest to download dataset and use that for all tutorials?

adamjstewart:
Another benefit: secrets aren't shared with unprivileged actors, so our tutorials tests fail when anyone other than a maintainer opens a PR that modifies the notebooks. By switching to a public dataset, the tests will always work.

PR #1130, opened 2023-02-21
EuroSAT100: add new dataset/datamodule

adamjstewart:
#### What?

This is a subset of the EuroSAT MS dataset containing only 100 images. Yes, you heard that right, 100 images. This dataset maintains the 60-20-20 train-val-test split from https://arxiv.org/abs/1911.06721. There are 10 images (6 train, 2 val, 2 test) for all 10 classes, for a total of 100 images.

#### But why?

Our tutorials download a lot of data. So much in fact that we crash the meager 14 GB SSD in our [GitHub Actions runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources), causing our notebook tests to fail for the last year or so. Between NAIP, Chesapeake, Sentinel, EuroSAT, and TropicalCyclone, we're talking several minutes of downloads. The TropicalCyclone download doesn't even work unless you sign up for an API key (#1074). A quick comparison:

| Dataset | Compressed | Uncompressed | Download Time |
| --- | --- | --- | --- |
| EuroSAT | 1.9G | 4.7G | 3m 7s |
| EuroSAT100 | 7.4M | 18M | 1s |

Yes, 1 second, down to the millisecond:
```console
$ time wget https://huggingface.co/datasets/torchgeo/eurosat/resolve/main/EuroSAT100.zip
...
real	0m1.000s
user	0m0.055s
sys	0m0.110s
```
Our tutorials should not take several minutes just on data prep, nor should they fill up your hard drive, nor should they require an API key.

#### Why EuroSAT?

EuroSAT was chosen for the following reasons:

* Many of our existing tutorials already use EuroSAT, making them easy to port
* It contains Sentinel-2 images, which our tutorials rely on to show off things like NDVI or pre-trained models
* It is distributed under a [permissive license](https://github.com/phelber/EuroSAT/issues/10#issuecomment-1409891796)
* Images are georeferenced, meaning we could add a GeoDataset version of this someday

#### The milestone says 0.4.1 but versionadded says 0.5...

Technically, new features shouldn't be added in patch releases. But this may fix our chronically failing notebook tests. So I would like to add this to the next patch release for the sake of testing, but we can just pretend that it was actually added in 0.5.

#### Hey, you forgot to update the docs!

I'll open a separate PR to add this to the docs that we can save for the 0.5 release.

### DIFF
--- a/torchgeo/datamodules/eurosat.py
+++ b/torchgeo/datamodules/eurosat.py
@@ -10,6 +10,42 @@ import torch
 from ..datasets import EuroSAT, EuroSAT100
 from .geo import NonGeoDataModule
 
+MEAN = torch.tensor(
+    [
+        1354.40546513,
+        1118.24399958,
+        1042.92983953,
+        947.62620298,
+        1199.47283961,
+        1999.79090914,
+        2369.22292565,
+        2296.82608323,
+        732.08340178,
+        12.11327804,
+        1819.01027855,
+        1118.92391149,
+        2594.14080798,
+    ]
+)
+
+STD = torch.tensor(
+    [
+        245.71762908,
+        333.00778264,
+        395.09249139,
+        593.75055589,
+        566.4170017,
+        861.18399006,
+        1086.63139075,
+        1117.98170791,
+        404.91978886,
+        4.77584468,
+        1002.58768311,
+        761.30323499,
+        1231.58581042,
+    ]
+)
+
 
 class EuroSATDataModule(NonGeoDataModule):
     """LightningDataModule implementation for the EuroSAT dataset.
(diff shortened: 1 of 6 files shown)


torchgeo/datamodules/eurosat.py
```diff
@@ -67,3 +67,59 @@ def __init__(
                 :class:`~torchgeo.datasets.EuroSAT`.
         """
         super().__init__(EuroSAT, batch_size, num_workers, **kwargs)
+
+
+class EuroSAT100DataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the EuroSAT100 dataset.
+
```
calebrob6, line 74: Can you add the "Intended for tutorials and demonstrations, not for benchmarking." disclaimer here too

calebrob6:
Could not reproduce the 1 second download, it took my laptop 1.971 seconds.

torchgeo/datamodules/eurosat.py
adamjstewart: Done.

adamjstewart:
I tried the download several times and it's always between 0.8s and 1.8s on my laptop. Either way, drastic improvement over the previous 3m download for EuroSAT and 10+ min download for Cyclone.

calebrob6 (approved):

calebrob6:
That's more like it! We emphasize test metric distributions in torchgeo!

PR #1131, opened 2023-02-21
Add MapInWild dataset

### LINKED ISSUE #1096: MapInWild dataset
### Summary

Hi! 

MapInWild is a large-scale multi-modal dataset curated for the novel task of wilderness mapping from space, introduced in our recent [paper](https://ieeexplore.ieee.org/document/10089830). We would be glad to see our dataset in `torchgeo`!

- The dataset is originally hosted on [harvard dataverse](https://dataverse.harvard.edu/dataverse/mapinwild). Following @calebrob6's suggestion, I have mirrored it on [huggingface](https://huggingface.co/datasets/burakekim/mapinwild) to make things easier for everyone. 

- The original GitHub repository of the dataset is [here](https://github.com/burakekim/MapInWild).

- Wilderness mapping is a supervised pixel-wise classification task. Although the annotations contain four classes/pixel values `{0:Background, 1:Strict nature reserve, 2:Wilderness Area, 3:National Park]}`, we perform ```annotation[annotation!= 0] = 1``` and frame this task as a binary classification problem where `0` is `Background` and `1`  is `Wilderness Area`. 

- The annotations (in the form of polygons) are derived from World Database of Protected Areas.

- The dataset is around 350 GB in size. That is why each modality is zipped separately, allowing the users to pick among the modalities they want to work on.    

There are 1018 areas in the dataset and each area contains the following modalities in the shape of 1920 x 1920 pixels. 
```
MapInWild
├── Dual Pol Sentinel-1 (2 Bands)
├──  Sentinel-2 (10 Bands)  
│   ├── Spring
│   ├── Autumn
│   ├── Winter
│   ├── Summer
│   └── Single Temporal Subset 
├──  VIIRS Night Time Light (1 Band)
└── ESA WorldCover (1 Band)
```

- As explained in the paper, the `single temporal subset` includes the most informative Sentinel-2 season for each area, suggested to the users who are not interested in the multi-seasonality aspect of the dataset.

Following the `torchgeo` logic, here are the bands and their modality-level combinations.

```
    BAND_SETS: Dict[str, Tuple[str, ...]] = {
        "all": (
            "VV",
            "VH",
            "B2",
            "B3",
            "B4",
            "B5",
            "B6",
            "B7",
            "B8",
            "B8A",
            "B11",
            "B12",
            "2020_Map",
            "avg_rad"), 
        "s1": ("VV", "VH"),
        "s2-rgb":(
            "B4",
            "B3",
            "B2"),
        "s2-all": (
            "B2",
            "B3",
            "B4",
            "B5",
            "B6",
            "B7",
            "B8",
            "B8A",
            "B11",
            "B12"),
        "esa_wc": {"2020_Map"},
        "viirs":{"avg_rad"}
    }
```

### Rationale

_No response_

### Implementation

After asked by the user `MapInWild(root="data/", modalities=[...], download=True)`, any modality can be loaded as below:

```
mask= load_dataset("burakekim/mapinwild", data_dir="mask")
s1 = load_dataset("burakekim/mapinwild", data_dir="s1")
viirs = load_dataset("burakekim/mapinwild", data_dir="viirs")
esa_wc = load_dataset("burakekim/mapinwild", data_dir="esa_wc")
s2 = load_dataset("burakekim/mapinwild", data_dir="s2_temporal_subset")
s2_autumn= load_dataset("burakekim/mapinwild", data_dir="s2_autumn")
s2_spring= load_dataset("burakekim/mapinwild", data_dir="s2_spring")
s2_winter= load_dataset("burakekim/mapinwild", data_dir="s2_winter")
s2_summer = load_dataset("burakekim/mapinwild", data_dir="s2_summer")
```
The s1 and s2 are bigger than 50 GB and they are split into two zip files.  For these modalities, the `num_proc=2` [argument](load_dataset) in the `load_dataset` can be used. 

### Alternatives

_No response_

### Additional information

_No response_

burakekim:
Closes #1096 

Here is the MapInWild dataset instance. There are things I solved with ugly fixes and quick workarounds but I hope this pull request still is a good-enough starting point. 

_TODO and issues_
- Checksum: So far no checksum. I can not seem to find the MD5s from Hugging Face. There are only SH256s.
- Now it is not really straightforward to distinguish the modalities in the `__getitem__` making it harder to apply specific normalization steps (i.e., s2/10000).
- Documentation and tests

_Nice to do_
- Picking modalities band-wise to enable the use of various combinations of bands. This could even supersede the modality-wise picking. 
- Improving the `plot` function. A normalization function (i.e., [percentile_normalization](https://github.com/microsoft/torchgeo/blob/main/torchgeo/datasets/sen12ms.py#L340)) is needed for at least the Sentinel-2 data. For that, an argument and modality-aware logic inside this function could be helpful. 
- Remove the `pandas` and move to a lower-level library. 

Any help is appreciated. Thanks in advance for your time in reviewing this pull request. 

P.S. I am by no means a GitHub pro, so any heads-up is appreciated. 

The signature:
`mapinwild = MapInWild(root="data", download=True, modality=["mask","esa_wc","s2_temporal_subset","viirs","s1"], split="train", transforms=None, checksum=False)
`

Some sample images: 

VIIRS Nighttime light band
viirs
Sentinel-2 single temporal subset (the Sentinel-2 season with the highest score) 
s2
ESA WorldCover map
esa_wc
Sentinel-1 
s1


torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,328 @@
+import os
```
adamjstewart, line 1: Needs license header copied from other files at the top

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,328 @@
...
+import rasterio
+import numpy as np
+import torch
+import pandas as pd
+import matplotlib.pyplot as plt
+
+from .geo import NonGeoDataset
+from .utils import check_integrity, extract_archive, download_url
+
+"""MapInWild dataset."""
```
adamjstewart, line 15: I would put this before the imports

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,328 @@
...
+
+class MapInWild(NonGeoDataset):
+    """MapInWild dataset.
+
+        The `MapInWild <https://arxiv.org/abs/2212.02265>` is curated for the task of wilderness mapping on a pixel-level.
```
adamjstewart, line 20: This is indented too far. Also, link is missing an underscore:
```suggestion
        The `MapInWild <https://arxiv.org/abs/2212.02265>`_ is curated for the task of wilderness mapping on a pixel-level. 
```
All lines should be less than 89 chars to make flake8 happy.

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,328 @@
...
+from .utils import check_integrity, extract_archive, download_url
+
+"""MapInWild dataset."""
+
+class MapInWild(NonGeoDataset):
+    """MapInWild dataset.
+
+        The `MapInWild <https://arxiv.org/abs/2212.02265>` is curated for the task of wilderness mapping on a pixel-level. 
+        MapInWild is a multi-modal dataset and comprises various geodata acquired and formed from different RS sensors over 1018 locations:
+        dual-pol Sentinel1, four-season Sentinel-2 with 10 bands, ESA WorldCover map, and Visible Infrared Imaging Radiometer Suite NightTime Day/Night band.
```
adamjstewart, line 22:
```suggestion
        dual-pol Sentinel-1, four-season Sentinel-2 with 10 bands, ESA WorldCover map, and Visible Infrared Imaging Radiometer Suite NightTime Day/Night band.
```

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,328 @@
...
+        The dataset consists of 8144 images with the shape of 1920 × 1920 pixels. 
+        The images are weakly annotated from the World Database of Protected Areas(WDPA).
```
adamjstewart, line 24:
```suggestion
        The images are weakly annotated from the World Database of Protected Areas (WDPA).
```

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,328 @@
...
+            "B5",
+            "B6",
+            "B7",
+            "B8",
+            "B8A",
+            "B11",
+            "B12"
+        ),
+        "esa_wc":{"2020_Map"},
+        "viirs":{"avg_rad"}
```
adamjstewart, line 68: I'm guessing these should be tuples instead of sets:
```suggestion
        "esa_wc": ("2020_Map",),
        "viirs": ("avg_rad",),
```

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,328 @@
...
+    "s2_summer" : {"https://huggingface.co/datasets/burakekim/mapinwild/resolve/main/s2_summer/s2_summer_part1.zip",
+                    "https://huggingface.co/datasets/burakekim/mapinwild/resolve/main/s2_summer/s2_summer_part2.zip"},
+     "s2_winter" : {"https://huggingface.co/datasets/burakekim/mapinwild/resolve/main/s2_winter/s2_winter_part1.zip",
+                    "https://huggingface.co/datasets/burakekim/mapinwild/resolve/main/s2_winter/s2_winter_part2.zip"}
+    }
+    
+    split_url = "https://huggingface.co/datasets/burakekim/mapinwild/resolve/main/split_IDs/split_IDs.csv"
+    main_directory = os.path.join(os.getcwd(), "data")
+    mask_palette = {1 : (0,153,0),
+                0 : (255,255,255)}
```
adamjstewart, line 110: A lot of the formatting for these is messed up. You can run `black` on these files to auto-fix most of it. See https://torchgeo.readthedocs.io/en/stable/user/contributing.html#linters for details.

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,328 @@
...
+                
+            for modal in modality:
+                if len(self.modality_urls[modal]) > 1:
+                    self.merge_parts(self.main_directory, modal)
+
+        self.list_modals = []
+        
+    def __getitem__(self, index: int) -> Dict[str, Tensor]:
+        """Return an index within the dataset.
+        Args:
```
adamjstewart, line 163: Missing spaces between sections, this happens when you copy-n-paste code from GitHub

adamjstewart (changes requested):
This looks like a great start! Let me know if you have any specific questions.

> Checksum: So far no checksum. I can not seem to find the MD5s from Hugging Face. There are only SH256s.

You can calculate them yourself using `md5 <filename>` on the command line. We'll likely move from MD5 to SHA256 someday.

> Remove the pandas and move to a lower-level library.

You should probably just keep pandas. I've been thinking about making it a required dep of TorchGeo because so many datasets use it.

burakekim:
@microsoft-github-policy-service agree company="unibwmunich"

burakekim:
Thanks a bunch for your revisions! 

I applied some quick changes and tests. I will make sure there is a full test coverage after I applying the necessery changes, which are for now:

adamjstewart:
> Do you happen to know why the line `pyupgrade --py37-plus $(find . -name "*.py")` outputs the error below?

Nope, that syntax should be correct. Note that our tests use the more complicated:
```console
$ pyupgrade --py38-plus $(find . -path ./docs/src -prune -o -name "*.py" -print)
```

> I shortened the plain texts that are going over 89 characters and applied `# noqa: E501` for the links, is this OK?

Yes this is perfect.

adamjstewart:
Note that we just dropped Python 3.8 support, so `--py38-plus` is now `--py39-plus`

burakekim:
@adamjstewart I have improved the dataset class (_mapinwild.py_) and added the code to create test data (_mapinwild\data.py_) with the last commit.  

Now, I assume the next step is to populate the _test_mapinwild.py_. 

The progress is quite slow as I have very little bandwidth for this side hustle -sorry for that!  In fact, feel free to contribute to the _test_mapinwild.py_ to wrap up things faster, otherwise, it could take up quite some time for me.

adamjstewart:
No rush. Most of us are currently working towards a paper deadline, so we prob won't be able to help much either. But might be able to help after June.

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,491 @@
...
+# Licensed under the MIT License.
+
+import os
+import shutil
+from collections.abc import Sequence
+from typing import Any, Callable, Optional
+
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
```
adamjstewart, line 13: Pandas is currently an optional dependency, it won't always be installed. You need to use a lazy import instead of importing it at the top level. See any of these datasets for examples of what this looks like:
```console
> grep -l pandas *.py
eddmaps.py
gbif.py
idtrees.py
inaturalist.py
openbuildings.py
reforestree.py
usavars.py
western_usa_live_fuel_moisture.py
```
This is why the Read the Docs tests are failing.

burakekim: Thanks for the heads up!

burakekim:
Finally did some improvements and fixes! Hope it is better now. Next, will improve the coverage from 77 to 100. 

P.S. We were extending the dataset with [Naturalness Index](https://www.sciencedirect.com/science/article/pii/S0303243421003299) patches and had to keep this PR on hold -but now I can wrap it up quickly!

burakek:
Done! Hope it is good enough. Please let me know what you think.

burakekim:
Done! Hope it is good enough. Please let me know what you think.

adamjstewart:
(sorry about the delay in review, working on a rebuttal right now, but this is my top priority once that is done)

tests/datasets/test_mapinwild.py
```diff
@@ -0,0 +1,158 @@
...
+            "mask",
+            "viirs",
+            "esa_wc",
+            "s2_winter",
+            "s1",
+            "s2_summer",
+            "s2_spring",
+            "s2_autumn",
+            "s2_temporal_subset",
+        ]  # noqa: E501
```
adamjstewart, line 56: Most of the noqa comments in this file can be removed. We generally want to break long lines anyway. The only exceptions are for long URLs where you should use this comment

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,513 @@
...
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://ieeexplore.ieee.org/document/10089830
+
+    .. note::
+       This dataset requires the following additional library to be installed:
+
+       * `pandas <https://pypi.org/project/pandas/>`_ to load CSV files
+
+    .. versionadded:: 0.3
```
adamjstewart, line 54:
```suggestion
    .. versionadded:: 0.5
```
Your dataset will first be available in TorchGeo 0.5.0 (expected late August/early September)

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,513 @@
...
+            "https://huggingface.co/datasets/burakekim/mapinwild/resolve/main/s2_summer/s2_summer_part1.zip",  # noqa: E501
+            "https://huggingface.co/datasets/burakekim/mapinwild/resolve/main/s2_summer/s2_summer_part2.zip",  # noqa: E501
+        },
+        "s2_winter": {
+            "https://huggingface.co/datasets/burakekim/mapinwild/resolve/main/s2_winter/s2_winter_part1.zip",  # noqa: E501
+            "https://huggingface.co/datasets/burakekim/mapinwild/resolve/main/s2_winter/s2_winter_part2.zip",  # noqa: E501
+        },
+    }
+
+    split_url = "https://huggingface.co/datasets/burakekim/mapinwild/resolve/main/split_IDs/split_IDs.csv"  # noqa: E501
```
adamjstewart, line 114: Might make more sense to have a single url:
```python
url = "https://huggingface.co/datasets/burakekim/mapinwild/resolve/main/"
```
and then a set/list of relative paths to the actual files. Then you won't have so many long lines and it will be easier to change if you ever move your dataset somewhere else.

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,513 @@
...
+        "s2_spring_part2.zip": "5845dcae0ab3cdc174b7c41edd4283a9",
+        "s2_summer_part1.zip": "73ca8291d3f4fb7533636220a816bb71",
+        "s2_summer_part2.zip": "5b5816bbd32987619bf72cde5cacd032",
+        "s2_winter_part1.zip": "ca958f7cd98e37cb59d6f3877573ee6d",
+        "s2_winter_part2.zip": "e7aacb0806d6d619b6abc408e6b09fdc",
+    }
+
+    mask_palette = {1: (0, 153, 0), 0: (255, 255, 255)}
+
+    wc_palette = {
```
adamjstewart, line 154: We usually call these cmap instead of pallelete since that's what GDAL and matplotlib call them

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,513 @@
...
+        self.ids = [int(i) for i in self.ids]
+
+        for modal in modality:
+            for modality_link in self.modality_urls[modal]:
+                self._verify(modality_link, self.md5s[os.path.split(modality_link)[1]])
+                if checksum:
+                    if not self._check_integrity(
+                        modality_link, self.md5s[os.path.split(modality_link)[1]]
+                    ):
+                        raise RuntimeError("Dataset not found or corrupted.")
```
adamjstewart, line 224: We usually handle checksumming in `_verify` so that `__init__` isn't super long and confusing

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,513 @@
...
+
+            if modal not in os.listdir(root) and len(self.modality_urls[modal]) == 2:
+                self.merge_parts(root, modal)  # same modality two times
+
+        if "mask" in self.modality:
+            self.modality.remove("mask")
```
adamjstewart, line 230: Might be worth adding more detailed comments to clarify why this is needed

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,513 @@
...
+        Returns:
+            data and label at that index
+        """
+        self.list_modals = []
+        id = self.ids[index]
+
+        mask = self._load_raster(id, "mask")
+        mask[mask != 0] = 1
+
+        if "s2_temporal_subset" in self.modality:
```
adamjstewart, line 247: Couldn't all these if-statements be replaced with a for-loop? Something like:
```python
for mode in self.modality:
    data = self._load_raster(id, mode)
    if mode.startswith("s2_"):
        data = torch.index_select(data, dim=0, index=self.band_indices)
    self.list_modals.append(data)
```

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,513 @@
...
+    def __getitem__(self, index: int) -> dict[str, Tensor]:
+        """Return an index within the dataset.
+
+        Args:
+            index: index to return
+
+        Returns:
+            data and label at that index
+        """
+        self.list_modals = []
```
adamjstewart, line 241: This is never used outside this function so no real reason to save it to self

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,513 @@
...
+
+        if self.transforms is not None:
+            im = np.einsum("ijk->jki", image.numpy())
+            msk = np.einsum("ijk->jki", mask.numpy())
+            sample_: dict[str, Tensor] = {"image": im, "mask": msk}
+
+            transformed = self.transforms(sample_)
+
+            image = torch.Tensor(np.einsum("ijk->kij", transformed["image"]))
+            mask = torch.Tensor(np.einsum("ijk->kij", transformed["mask"]))
```
adamjstewart, line 304: Preferably all tensors would be in the shape expected by our transforms without reshaping. Also, no need to convert to numpy and back to PyTorch just to reshape, you can reshape directly in PyTorch using einops.

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,513 @@
...
+
+        Args:
+            filename: name of the file to load
+            source: the filename of the modality
+
+        Returns:
+            the raster image or target
+        """
+        with rasterio.open(
+            os.path.join(self.root, f"{source}", f"{filename}.tif")
```
adamjstewart, line 329:
```suggestion
            os.path.join(self.root, source, f"{filename}.tif")
```
Could also save the filename on a separate line so it doesn't become long enough to require a line break

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,513 @@
...
+        ) as f:
+            array = f.read()
+            if array.dtype == np.uint16:
+                array = array.astype(np.int32)
+            tensor = torch.from_numpy(array)
+            return tensor
+
+    def _verify(self, url: str, md5: str) -> None:
```
adamjstewart, line 337: Need to document input parameters, or remove input parameters and access them via self

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,513 @@
...
+            )
+        else:
+            # Download the dataset
+            self._download(url, md5)
+            self._extract(url)
+
+    def _verify_split(self) -> None:
+        """Verify the integrity of the split file."""
+        # Download the dataset
+        download_url(
```
adamjstewart, line 364: This shouldn't download unless `download=True`

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,513 @@
...
+            filename=os.path.split(url)[1],
+            md5=md5 if self.checksum else None,
+        )
+
+    def _extract(self, url: str) -> None:
+        """Extract the dataset."""
+        filepath = os.path.join(self.root, os.path.split(url)[1])
+        extract_archive(filepath)
+
+    def _check_integrity(self, filename: str, md5: str) -> bool:
```
adamjstewart, line 385: What's the difference between verify and check_integrity? Torchvision calls it check_integrity and we call it verify but the result is usually the same. I don't think we have any other datasets that have both

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,513 @@
...
+
+        Returns:
+            True if dataset files are found and/or MD5s match, else False
+        """
+        filepath = os.path.join(self.root, os.path.split(filename)[1])
+        if not check_integrity(filepath, md5 if self.checksum else None):
+            return False
+        return True
+
+    def merge_parts(self, source_path: str, modality: str) -> None:
```
adamjstewart, line 396: Could probably start with an underscore to make it hidden. Need to document parameters

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,513 @@
...
+        if len(modality.split("_")) == 2:
+            rename_dest = os.path.join(
+                dest_split[0],
+                dest_split[1].split("_")[0] + "_" + dest_split[1].split("_")[1],
+            )
+        if "_" not in modality:
+            rename_dest = os.path.join(dest_split[0], dest_split[1].split("_")[0])
+        os.rename(destination_folder, rename_dest)
+
+    def convert_to_color(self, arr_2d: Any, palette: Any) -> Any:
```
adamjstewart, line 421: Any should be avoided unless the function is required to be dynamically typed. Let me know if you need any help with type hints

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,513 @@
...
+        arr_3d = np.zeros((arr_2d.shape[0], arr_2d.shape[1], 3), dtype=np.uint8)
+
+        for c, i in palette.items():
+            m = arr_2d == c
+            arr_3d[m] = i
+        return arr_3d
+
+    def get_bands(
+        self, image: Any, all_bands: Sequence[str], select_bands: Sequence[str]
+    ) -> Any:  # noqa: E501
```
adamjstewart, line 432: No need for noqa

torchgeo/datasets/mapinwild.py
burakekim: Ah, nice catch!

torchgeo/datasets/mapinwild.py
burakekim: Good point -somehow missed this.

torchgeo/datasets/mapinwild.py
burakekim: Indeed! Was planning to provide the user with a modality-specific band argument with these if-statements in the future but we can keep it like this for the first version, it is much more readable and clean.

torchgeo/datasets/mapinwild.py
burakekim: Revised the flow by unifying the checks -now it is verify

adamjstewart:
It looks like the tests are failing

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,468 @@
...
+        arr_3d = np.zeros((arr_2d.shape[0], arr_2d.shape[1], 3), dtype=np.uint8)
+
+        for c, i in cmap.items():
+            m = arr_2d == c
+            arr_3d[m] = i
+        return arr_3d
+
+    def get_bands(
+        self,
+        image: "np.typing.NDArray[np.float32]",
```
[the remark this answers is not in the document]
burakekim: @adamjstewart Did create the review message above a couple of days back and now realize "Pending" means it has not been submitted yet -my bad.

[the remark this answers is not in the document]
adamjstewart: The mypy tests seem to be passing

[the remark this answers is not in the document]
adamjstewart: Oh, but the RtD tests aren't. Maybe make it a hidden method so it doesn't get documented.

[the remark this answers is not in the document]
adamjstewart: I would use one underscore instead of two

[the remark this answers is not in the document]
burakekim: Done :)

tests/data/mapinwild/data.py
```diff
@@ -0,0 +1,183 @@
...
+    "s2_spring_part2": np.iinfo(np.uint16).max,
+    "s2_summer_part1": np.iinfo(np.uint16).max,
+    "s2_summer_part2": np.iinfo(np.uint16).max,
+    "s2_winter_part1": np.iinfo(np.uint16).max,
+    "s2_winter_part2": np.iinfo(np.uint16).max,
+}
+
+folder_path = os.path.join(os.getcwd(), "tests", "data", "mapinwild")
+
+dict_all__ = {
```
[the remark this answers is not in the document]
adamjstewart: Still curious about this, I would just remove the trailing underscores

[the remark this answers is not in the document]
adamjstewart: If `folder_path` will always be joined you could remove it from the dict and only join when you need it. Would save a lot of space

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,467 @@
...
+        if "mask" in self.modality:
+            self.modality.remove("mask")
+
+        if "split_ids" in self.modality:
+            self.modality.remove("split_ids")
+
+        if os.path.exists(os.path.join(self.root, "split_IDs.csv")):
+            split_dataframe = pd.read_csv(os.path.join(self.root, "split_IDs.csv"))
+            self.ids = split_dataframe[split].dropna().values.tolist()
+            self.ids = [int(i) for i in self.ids]
```
adamjstewart, line 230: Might be simpler:
```suggestion
            self.ids = list(map(int, self.ids))
```
Could also tell `read_csv` the correct dtype for this column so you don't have to do it later.

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,467 @@
...
+
+        image = torch.cat(list_modals, dim=0)
+
+        if self.transforms is not None:
+            sample_: dict[str, Tensor] = {"image": image, "mask": mask}
+
+            transformed = self.transforms(sample_)
+
+            image = Rearrange("h w c -> c h w")(transformed["image"])
+            mask = Rearrange("h w c -> c h w")(transformed["mask"])
```
adamjstewart, line 260: Up above you can use:
```python
from einops import rearrange
```
Then here you can do:
```suggestion
            image = rearrange(transformed["image"], "h w c -> c h w")
            mask = rearrange(transformed["mask"], "h w c -> c h w")
```

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,467 @@
...
+        filepath = os.path.join(self.root, url.split("/")[-1])
+        if os.path.isfile(filepath) and filepath.endswith(".zip"):
+            if self.checksum and not check_integrity(filepath, md5):
+                raise RuntimeError("Dataset found, but corrupted.")
+            self._extract(url)
+
+        # Check if the user requested to download the dataset
+        if not self.download:
+            raise RuntimeError(
+                "Dataset not found in `root` directory and `download=False`, "
```
adamjstewart, line 310:
```suggestion
                f"Dataset not found in `root={self.root}` directory and `download=False`, "
```

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,467 @@
...
+
+        Args:
+            url: url to the file
+            md5: md5 of the file to be verified
+
+        Raises:
+            RuntimeError: if dataset is not found
+        """
+        filepath = os.path.join(self.root, url.split("/")[-1])
+        if os.path.isfile(filepath) and filepath.endswith(".zip"):
```
adamjstewart, line 302: Does this extract the zip file every time? We should only extract the zip file if the unzipped file is missing. Otherwise loading the dataset will be very slow every time.

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,467 @@
...
+            the raster image or target
+        """
+        with rasterio.open(os.path.join(self.root, source, f"{filename}.tif")) as f:
+            array = f.read()
+            if array.dtype == np.uint16:
+                array = array.astype(np.int32)
+            tensor = torch.from_numpy(array)
+            return tensor
+
+    def _verify(self, url: str, md5: Optional[str] = None) -> None:
```
adamjstewart, line 291: Normally these methods don't have any arguments, just a for-loop that checks/downloads/extracts all parts.

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,467 @@
...
+        fname_p2 = modality + "_part2"
+        source_folder = os.path.join(root, fname_p1)
+        destination_folder = os.path.join(root, fname_p2)
+        for file_name in os.listdir(source_folder):
+            source = os.path.join(source_folder, file_name)
+            destination = os.path.join(destination_folder, file_name)
+            if os.path.isfile(source):
+                shutil.move(source, destination)
+
+        shutil.rmtree(source_folder)
```
adamjstewart, line 355: Would this result in the dataset having to be re-downloaded after every time it is used?

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,467 @@
...
+            filename=os.path.split(url)[1],
+            md5=md5 if self.checksum else None,
+        )
+
+    def _extract(self, url: str) -> None:
+        """Extract the dataset."""
+        filepath = os.path.join(self.root, os.path.split(url)[1])
+        extract_archive(filepath)
+
+    def _merge_parts(self, root: str, modality: str) -> None:
```
adamjstewart, line 338: I don't really understand what this function does. It just moves both parts into the same directory? Couldn't the dataset be loaded into memory and then joined afterwards without changing the file structure? That would make it easier to reuse without having to re-download the dataset each time

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,467 @@
...
+        if len(modality.split("_")) == 2:
+            rename_dest = os.path.join(
+                dest_split[0],
+                dest_split[1].split("_")[0] + "_" + dest_split[1].split("_")[1],
+            )
+        if "_" not in modality:
+            rename_dest = os.path.join(dest_split[0], dest_split[1].split("_")[0])
+        os.rename(destination_folder, rename_dest)
+
+    def convert_to_color(
```
adamjstewart, line 368: Should probably be a hidden function that starts with an underscore. Also, parameter and return value documentation is missing.

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,467 @@
...
+
+        Args:
+            sample: a sample image-mask pair returned by :meth:`__getitem__`
+            show_titles: flag indicating whether to show titles above each panel
+            suptitle: optional string to use as a suptitle
+
+        Returns:
+            a matplotlib Figure with the rendered sample
+        """
+        image = np.einsum("ijk->jki", sample["image"])
```
adamjstewart, line 419: Could prob use einops.rearrange here too

adamjstewart (changes requested):

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,467 @@
...
+            "avg_rad",
+        ),
+        "s1": ("VV", "VH"),
+        "s2": ("B2", "B3", "B4", "B5", "B6", "B7", "B8", "B8A", "B11", "B12"),
+    }
+
+    url = "https://huggingface.co/datasets/burakekim/mapinwild/resolve/main/"
+
+    modality_urls = {
+        "esa_wc": {os.path.join(url, "esa_wc/ESA_WC.zip")},
```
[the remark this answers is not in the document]
burakekim: Ah, sure!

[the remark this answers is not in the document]
burakekim: Long lines are triggering the flake8 line too long error and using E501 is making the line break. Removing the unused E501's then causing the lines to become long again. Any ideas?

torchgeo/datasets/mapinwild.py
burakekim: The source folder (i.e., s1_part1) has been moved to the destination folder (i.e., s1_part2) before the deletion of source folder. Then, we rename the destination folder like destination_folder.split('_')[0]. So yeah, it is not removing any data.

torchgeo/datasets/mapinwild.py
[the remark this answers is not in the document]
adamjstewart: If the line looked like:
```python
modality_suffixes = {
    "esa_wc": {"esa_wc/ESA_WC.zip"},
    ...
}
```
then it should never be too long. Maybe push the change and then show me what you're talking about?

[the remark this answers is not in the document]
burakekim: The modalities with two links (s1 and s2) are the ones causing this problem. Sure, will push along with the others later today.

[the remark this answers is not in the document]
burakekim: Ah alright solved it.

torchgeo/datasets/mapinwild.py
burakekim: Changed the logic here. Now unzipping takes place if the file is missing

adamjstewart:
> Do you have any suggestions?

The problem is that `_convert_to_color` says it takes in a Tensor but the image you're giving it as a numpy NDArray. Personally I would just do everything on Tensors if you can.

> Is this acceptable, or should I reduce it by changing the image size in pixels?

Image size is probably fine, maybe reduce the number of sample images from 6 to 1 or 2?

Will review the rest of the PR in more detail hopefully tomorrow.

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,479 @@
...
+    download_url,
+    extract_archive,
+    percentile_normalization,
+)
+
+
+class MapInWild(NonGeoDataset):
+    """MapInWild dataset.
+
+    The `MapInWild <https://arxiv.org/abs/2212.02265>`_ dataset is curated for the task
```
adamjstewart, line 30:
```suggestion
    The `MapInWild <https://arxiv.org/abs/2212.02265>`__ dataset is curated for the task
```
Otherwise the link in the https://torchgeo--1131.org.readthedocs.build/en/1131/api/datasets.html#non-geospatial-datasets table will take you to the paper instead of the API docs.

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,479 @@
...
+    of wilderness mapping on a pixel-level. MapInWild is a multi-modal dataset and
+    comprises various geodata acquired and formed from different RS sensors
+    over 1018 locations: dual-pol Sentinel-1, four-season Sentinel-2 with 10 bands,
+    ESA WorldCover map, and Visible Infrared Imaging Radiometer Suite
+    NightTime Day/Night band. The dataset consists of 8144 images with the shape
+    of 1920 × 1920 pixels. The images are weakly annotated from the
+    World Database of Protected Areas (WDPA).
+
+    Dataset features:
```
adamjstewart, line 39: Need empty line before start of list, otherwise it doesn't actually create a list: https://torchgeo--1131.org.readthedocs.build/en/1131/api/datasets.html#mapinwild

tests/datasets/test_mapinwild.py
```diff
@@ -0,0 +1,233 @@
...
+            "s2_temporal_subset_part1.zip": "0398bc292d9c3e8513444c96c5b0da96",
+            "s2_temporal_subset_part2.zip": "c063386289476dc2342872abb7c57f1c",
+            "s2_winter_part1.zip": "4f67835837f865f03cda9841b0da0717",
+            "s2_winter_part2.zip": "b6654fb29a315bddaca8b3e7a9f34930",
+        }
+        monkeypatch.setattr(MapInWild, "md5s", md5s)
+
+        urls = os.path.join("tests", "data", "mapinwild")
+
+        modality_urls = {
```
adamjstewart, line 55: You can now replace this entire thing with a one-line monkeypatch of `MapInWild.url`

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,479 @@
...
+            "s2_temporal_subset/s2_temporal_subset_part2.zip",
+        },
+        "s2_autumn": {"s2_autumn/s2_autumn_part1.zip", "s2_autumn/s2_autumn_part2.zip"},
+        "s2_spring": {"s2_spring/s2_spring_part1.zip", "s2_spring/s2_spring_part2.zip"},
+        "s2_summer": {"s2_summer/s2_summer_part1.zip", "s2_summer/s2_summer_part2.zip"},
+        "s2_winter": {"s2_winter/s2_winter_part1.zip", "s2_winter/s2_winter_part2.zip"},
+        "split_ids": {"split_IDs/split_IDs.csv"},
+    }
+
+    filenames = [
```
adamjstewart, line 97: Seems like this is no longer used and can be removed

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,479 @@
...
+                "pandas is not installed and is required to use this dataset"
+            )
+
+        modality.append("split_ids")
+
+        for modal in modality:
+            for modality_link in self.modality_urls[modal]:
+                os.path.join(self.url, modality_link)
+                if modality_link.endswith("csv"):
+                    self._verify(url=modality_link, md5=None)
```
adamjstewart, line 195: Why no md5 for the csv file?

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,479 @@
...
+                else:
+                    self._verify(
+                        url=modality_link,
+                        md5=self.md5s[os.path.split(modality_link)[1]],
```
adamjstewart, line 199: If you use -1 instead of 1 it will work even for the csv file which isn't in a subdir

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,479 @@
...
+                    )
+
+            #  Merge modalities downloaded in two parts
```
adamjstewart, line 202:
```suggestion
            # Merge modalities downloaded in two parts
```

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,479 @@
...
+            if (
+                download
+                and modal not in os.listdir(self.root)
```
adamjstewart, line 205: I would call it modality or mode instead of modal

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,479 @@
...
+                and len(self.modality_urls[modal]) == 2
+            ):
+                self._merge_parts(modal)
+
+        # Masks will be loaded seperately in the :meth:`__getitem__`
+        if "mask" in self.modality:
+            self.modality.remove("mask")
+
+        # Split IDs has been downloaded and not needed in the list
```
adamjstewart, line 214:
```suggestion
        # Split IDs has been downloaded and is not needed in the list
```

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,479 @@
...
+    def __getitem__(self, index: int) -> dict[str, Tensor]:
+        """Return an index within the dataset.
+
+        Args:
+            index: index to return
+
+        Returns:
+            data and label at that index
+        """
+        list_modals = []
```
adamjstewart, line 232: Maybe modes or modalities instead of modals

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,479 @@
...
+        mask[mask != 0] = 1
+
+        for mode in self.modality:
+            mode = mode.upper() if mode in ["esa_wc", "viirs"] else mode
+            data = self._load_raster(id, mode)
+            list_modals.append(data)
+
+        image = torch.cat(list_modals, dim=0)
+
+        if self.transforms is not None:
```
adamjstewart, line 245: Transforms should be applied _after_ rearranging the dimensions. All of our built-in transforms assume the data is already in the right shape. You can also get rid of the sample_trans and transformed variables and just have a single sample variable that gets updated.

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,479 @@
...
+        """Verify the integrity of the dataset.
+
+        Args:
+            url: url to the file
+            md5: md5 of the file to be verified
+
+        Raises:
+            RuntimeError: if dataset is not found
+        """
+        modality_folder_name = os.path.split(url)[-1]
```
adamjstewart, line 292: URLs are not really "paths". On Windows, the path separator is \, not /. It would be better to split by / regardless of OS.

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,479 @@
...
+            raise RuntimeError(
+                "Dataset not found in `root` directory and `download=False`, "
+                "either specify a different `root` directory or use `download=True` "
+                "to automatically download the dataset."
+            )
+        elif self.download and url.endswith(".csv"):
+            # Download the split file
+            download_url(url, self.root, filename=os.path.split(url)[1], md5=None)
+
+        elif self.download and not url.endswith(".csv"):
```
adamjstewart, line 322:
```suggestion
        elif url.endswith(".csv"):
            # Download the split file
            download_url(url, self.root, filename=os.path.split(url)[1], md5=None)

        else:
```
`self.download` is guaranteed to be true thanks to the first if-block

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,479 @@
...
+            # Download the dataset
+            self._download(url, md5)
+            self._extract(url)
+
+    def _download(self, url: str, md5: Optional[str]) -> None:
+        """Download the dataset."""
```
adamjstewart, line 328: All parameters should be documented in the docstring

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,479 @@
...
+        download_url(
+            url,
+            self.root,
+            filename=os.path.split(url)[1],
+            md5=md5 if self.checksum else None,
+        )
+
+    def _extract(self, url: str) -> None:
+        """Extract the dataset."""
```
adamjstewart, line 337: Need to document url. Also, this isn't a url, it's a path, right? os.path.join doesn't make sense for URLs, only for paths.

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,479 @@
...
+        Returns:
+            a matplotlib Figure with the rendered sample
+        """
+        image = rearrange(sample["image"], "h w c -> w c h")
+        image = image.numpy()
+        mask = sample["mask"].squeeze()
+        color_mask = self._convert_to_color(mask, cmap=self.mask_cmap)
+
+        # Land cover
+        if np.all(np.isin(image, np.arange(0, 110, 10))) and image.shape[-1] == 1:
```
adamjstewart, line 436: Feels like there should be a better way to check whether or not the image is land cover. What if the image is really dark (all low numbers)? Could be a false positive

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,479 @@
...
+            image = self._convert_to_color(image.squeeze(), cmap=self.wc_cmap)
+        # Sentinel-1
+        elif image.shape[-1] == 2:
+            image = image[:, :, 0]
+            image = percentile_normalization(image)
```
adamjstewart, line 441: This line is duplicated many times

adamjstewart (changes requested):

torchgeo/datasets/mapinwild.py
adamjstewart: As long as the dataset doesn't try to redownload the files because source folder is empty

torchgeo/datasets/mapinwild.py
adamjstewart: By "loading into memory" I just mean maintaining either a list of all file paths or actually loading all files in `__init__`. The former is probably sufficient and more memory efficient. In fact, you already do this by creating `self.ids`. There's nothing in `__getitem__` that requires all files to be in the same directory, right? I guess I just don't think moving all these files around has any benefit and adds a lot of extra code.

adamjstewart:
Just a heads up, we're planning on finalizing TorchGeo 0.5.0 in the next week, so if you want this dataset in the next release, now is the time to finish. Let me know if there's anything I can help with.

burakekim:
Hi Adam, I was out of the office for the last two weeks, will finalize the PR earlier this week. Thanks for the heads-up!

tests/data/mapinwild/data.py
```diff
@@ -0,0 +1,163 @@
...
+for source in modality_download_list:
+    directory = os.path.join(folder_path, source)
+
+    # Remove old data
+    if os.path.exists(directory):
+        shutil.rmtree(directory)
+    os.makedirs(directory, exist_ok=True)
+
+    # Random images
+    for i in range(1, 7):
```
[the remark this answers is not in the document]
burakekim: Turns out we can get away with 2! It just needs to be >1 as we have "*_part1" and "*_part2" for some modalities.

tests/datasets/test_mapinwild.py
burakekim: I did not quite get this one. I monkeypatch the attribute "modality_url" below and there is no "MapInWild.url" attribute. Could you help me out on this?

adamjstewart: There is a `MapInWild.url` attribute, it's value is:
```python
url = "https://huggingface.co/datasets/burakekim/mapinwild/resolve/main/"
```

torchgeo/datasets/mapinwild.py
burakekim: Would it be fine to keep it this way? I somehow needed to start off the PR this way to have higher degree of freedom in the __init__. Could try to change it now though.

torchgeo/datasets/mapinwild.py
burakekim: Refactored it a bit but your scenario still holds. Could reach the metadata and statistics to add more checks on resolution and mean/std values but it could introduce some overhead. Do you think this is an edge scenario or just needs to be covered?

torchgeo/datasets/mapinwild.py
adamjstewart: It would be fine. If we ever need to force all of them to have the same signature I can update this to match in the future.

torchgeo/datasets/mapinwild.py
adamjstewart: Does `self.modality` tell you what you need to know? I think that's a lot safer than relying on image values/shapes.

torchgeo/datasets/mapinwild.py
burakekim: Do you mean the elif blocks? They are to check what modality and apply specific viz logics.

adamjstewart: I'm just talking about the `percentile_normalization(...)` part. You could just add a block at the end like:
```python
if not land cover:
    image = percentile_normalization(image)
```

torchgeo/datasets/mapinwild.py
burakekim: Not explicitly. It is a list of modalities set by the user. I can expand on the check with `and esa_wc in self.modality` but it will not clear out the inter-confusion.

torchgeo/datasets/mapinwild.py
burakekim: Oh right. Thanks!

torchgeo/datasets/mapinwild.py
adamjstewart: Hmm, correct me if I'm wrong but the current plotting method doesn't really work if multiple modalities are selected, correct? Maybe it would make more sense to have a for loop over `self.modalities` and use that to index which channels of the image are which modality. Then you can add a plot for each.

tests/datasets/test_mapinwild.py
adamjstewart: You might need to add a subdirectory to your tarballs to match.

torchgeo/datasets/mapinwild.py
burakekim: Added! :)

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,429 @@
...
+                modality_channels[modality] = 10
+
+        start_idx = 0
+        split_images = {}
+
+        for modality in self.modality:
+            end_idx = start_idx + modality_channels[modality]  # Start + n of channels
+            split_images[modality] = sample["image"][start_idx:end_idx, :, :]  # Slicing
+            start_idx = end_idx  # Update the iterator
+
```
burakekim, line 382: Above this line could be a function that outputs the `split_images`. I have no preference

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,429 @@
...
+        # Plot each modality in its respective axis
+        for i, (modality, image) in enumerate(split_images.items()):
+            ax = axs[i]
+            img = np.transpose(image, (1, 2, 0)).squeeze()
+            # Apply transformations based on modality type
+            if modality.startswith("s2"):
+                img = img[:, :, [4, 3, 2]]
+            if modality == "esa_wc":
+                img = self._convert_to_color(torch.as_tensor(img), cmap=self.wc_cmap)
+            if modality == "s1":
```
burakekim, line 403: Is there a go-to viz scheme for Sentinel-1? I simply use the VV band here but can do a color composite as well.

adamjstewart: I like color composites (see the Sentinel1 dataset) but had trouble getting them to work for other scales (see the SSL4EOS12 dataset). I think either is fine.

burakekim: Alright will keep it this way for now then. Will be doing side experiments for this on the side.

torchgeo/datasets/mapinwild.py
burakekim: Will look into this soon but can also do so after the release. Please let me know if you have a preference

burakekim:
Could you please help me understand the check "license/cla Expected — Waiting for status to be reported"? I can not seem to complete that. 

Also, please let me know if there is anything else to be done from my side before the release. Thank you for your time in helping me out here, it was a great learning curve for me. I could never have done it without your great reviews :) I plan to contribute more to the library as much as I can!

adamjstewart:
@calebrob6 any idea what's wrong with the CLA bot? It was accepted here: https://github.com/microsoft/torchgeo/pull/1131#issuecomment-1439709629

Glad my harsh reviews didn't scare you away 😆. I also see that you're in Munich? I just started a postdoc at TUM, maybe I'll see you around!

I'll review everything in detail again tomorrow, but I think we're probably pretty close now.

calebrob6:
No idea, let's see if the ol' close and reopen trick works :)

adamjstewart:
I hate how reliable ol' close and reopen is...

calebrob6:
Yep, I think that did it! Thanks for all the hard work on this @burakekim!

burakekim:
> I also see that you're in Munich? I just started a postdoc at TUM, maybe I'll see you around!

@adamjstewart Yes, I am doing my PhD here in Munich. Definitely, it would be great to meet up! I will shoot you an e-mail :)

burakekim:
> Thanks for all the hard work on this @burakekim!

Thanks a bunch, @calebrob6! Contributing to the library has been an enjoyable experience :)

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,427 @@
...
+        self.root = root
+        self.transforms = transforms
+        self.modality = modality
+        self.download = download
+
+        try:
+            import pandas as pd  # noqa: F401
+        except ImportError:
+            raise ImportError(
+                "pandas is not installed and is required to use this dataset"
```
adamjstewart, line 148: See #1586, I just made pandas required, so you can actually get rid of the lazy imports and pandas tests and just treat it like normal.

tests/datasets/test_mapinwild.py
```diff
@@ -0,0 +1,168 @@
...
+            "viirs",
+            "esa_wc",
+            "s2_winter",
+            "s1",
+            "s2_summer",
+            "s2_spring",
+            "s2_autumn",
+            "s2_temporal_subset",
+        ]
+        MapInWild(root=str(tmp_path), modality=modality, download=True)
```
adamjstewart, line 134:
```suggestion
    def test_already_downloaded(self, dataset: MapInWild) -> None:
        MapInWild(root=dataset.root, modality=dataset.modality, download=True)
```

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,427 @@
...
+
+import matplotlib.pyplot as plt
+import numpy as np
+import rasterio
+import torch
+from matplotlib.figure import Figure
+from torch import Tensor
+
+from torchgeo.datasets.geo import NonGeoDataset
+from torchgeo.datasets.utils import (
```
adamjstewart, line 18:
```suggestion
from .geo import NonGeoDataset
from .utils import (
```

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,427 @@
...
+            length of the dataset
+        """
+        return len(self.ids)
+
+    def _load_raster(self, filename: int, source: str) -> Tensor:
+        """Load a single raster image or target.
+
+        Args:
+            filename: name of the file to load
+            source: the filename of the modality
```
adamjstewart, line 222:
```suggestion
            source: the directory of the modality
```

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,427 @@
...
+        # Check if the user requested to download the dataset
+        if not self.download:
+            raise RuntimeError(
+                f"Dataset not found in `root={self.root}` directory and `download=False`, "  # noqa: E501
+                "either specify a different `root` directory or use `download=True` "
+                "to automatically download the dataset."
+            )
+        elif url.endswith(".csv"):
+            # Download the split file
+            download_url(url, self.root, filename=os.path.split(url)[1], md5=None)
```
adamjstewart, line 273: Why do we download this even if the user requested `download=False`?

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,427 @@
...
+
+        Args:
+            root: root directory where dataset can be found
+            modality: the filename of the modality
+        """
+        # Create a new folder named after the 'modality' variable
+        modality_folder = os.path.join(self.root, modality)
+        os.makedirs(
+            modality_folder, exist_ok=True
+        )  # Will not raise an error if the folder already exists
```
adamjstewart, line 314: Can you put this comment on the line before so it doesn't get stretched to 3 lines?

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,427 @@
...
+
+        Returns:
+            a matplotlib Figure with the rendered sample
+        """
+        modality_channels = {"viirs": 1, "esa_wc": 1, "s1": 2}
+
+        # Assign a number of channels to all 's2' modalities
+        for modality in self.modality:
+            if modality.startswith("s2"):
+                modality_channels[modality] = 10
```
adamjstewart, line 370: If you just want this to be 10 for all modes not listed, you could use a `defaultdict(10)`

calebrob6:
What are the requested changes here?

adamjstewart:
Need to re-review tonight

tests/datasets/test_mapinwild.py
```diff
@@ -0,0 +1,128 @@
...
+import torchgeo.datasets.utils
+from torchgeo.datasets import MapInWild
+
+
+def download_url(url: str, root: str, *args: str, **kwargs: str) -> None:
+    shutil.copy(url, root)
+
+
+class TestMapInWild:
+    @pytest.fixture(params=["train", "validation", "test"])
```
adamjstewart, line 25: You aren't actually using this parameterization anywhere are you? I think you need to add a request parameter to the function and use it to define the split. At the moment, you're just testing the "train" split 3 times lol.

torchgeo/datasets/mapinwild.py
```diff
@@ -0,0 +1,407 @@
...
+        80: (0, 70, 200),
+        90: (0, 220, 130),
+        95: (0, 150, 120),
+        100: (255, 235, 175),
+    }
+
+    def __init__(
+        self,
+        root: str = "data",
+        modality: list[str] = ["mask", "esa_wc", "viirs", "s2_summer"],
```
adamjstewart, line 111:
```suggestion
        modality: Sequence[str] = ["mask", "esa_wc", "viirs", "s2_summer"],
```
This can be imported from typing at the top of the file. This will allow people to use either a list or tuple if they want to.

adamjstewart (changes requested):
Two remaining easy changes and I think this will be ready to merge!

tests/datasets/test_mapinwild.py
burakekim: Done :)

torchgeo/datasets/mapinwild.py
burakekim: I do append on the `modality` and that does not work with tuples. Doing a type changing in the init but it does not seem to work. Any suggestions?

adamjstewart: Ah, if append is required then it has to be a list. Let's keep it as is.

tests/datasets/test_mapinwild.py
adamjstewart: Can you push this change?

tests/datasets/test_mapinwild.py
```diff
@@ -0,0 +1,137 @@
...
+import os
+import shutil
+from pathlib import Path
+
+import matplotlib.pyplot as plt
+import pytest
+import torch
+import torch.nn as nn
+from _pytest.fixtures import SubRequest
+from _pytest.monkeypatch import MonkeyPatch
```
adamjstewart, line 14:
```suggestion
from pytest import MonkeyPatch                                                           
```

adamjstewart (approved):
Great work on this! Glad to have your dataset in TorchGeo!

adamjstewart:
Bumping tests...

PR #1132, opened 2023-02-21
Trainers: refactor tests

adamjstewart:
These fake models have nothing to do with testing `torchgeo/trainers/utils.py` and should either be in the files where they are used or in `conftest.py`.


calebrob6 (approved):

calebrob6:
sure!

PR #1136, opened 2023-02-22
Update pyupgrade Python version support

adamjstewart:
Follow-up to #1058, forgot to update one section of the docs.


calebrob6 (approved):

calebrob6:
good catch

PR #1144, opened 2023-02-24
Fix version selection button in docs

adamjstewart:
Want to see if this fixes the broken version selection button.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,9 @@ updates:
       # setuptools releases new versions almost daily
       - dependency-name: "setuptools"
         update-types: ["version-update:semver-patch"]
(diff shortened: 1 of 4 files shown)


adamjstewart:
Okay, 5.3.0 works, let's try 6.0.0 to confirm that that's the version that broken pytorch-sphinx-theme.

adamjstewart:
And 6.0.0 breaks, will open an issue.

adamjstewart:
https://github.com/pytorch/pytorch_sphinx_theme/issues/175

adamjstewart:
We may want to reconsider switching to trojanzoo_sphinx_theme since it seems to actually be supported.

calebrob6 (approved):

calebrob6:
This also fixes the floating menus (these break with the version selector)

PR #1148, opened 2023-02-26
Speed up detection tests

calebrob6:
Same idea as https://github.com/microsoft/torchgeo/pull/344, but for the object detection tests.

See https://github.com/microsoft/torchgeo/issues/1088 for discussion.

On my machine this reduced the time taken for `tests/trainers/test_detection.py` from 22 seconds to 11 seconds.

### DIFF
--- a/tests/trainers/test_detection.py
+++ b/tests/trainers/test_detection.py
@@ -5,9 +5,13 @@ import os
 from typing import Any, Dict, Type, cast
 
 import pytest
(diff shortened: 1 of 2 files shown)


adamjstewart (approved):

adamjstewart:
This should also save on memory. The difference on Linux isn't huge, but it's much bigger on macOS/Windows.

PR #1149, opened 2023-02-28
Add dtype field to RasterDataset

calebrob6:
In the great DataModule overhaul of winter 2022/2023 (https://github.com/microsoft/torchgeo/pull/992) we made sure that all the dtypes returned by our datasets played nicely with Kornia. As part of this, we made the assumption in RasterDataset that all "mask" layers should be `torch.long` and all images should be `torch.float`. As a result, our RasterDatasets can essentially not be used for real valued regression tasks.

This PR aims to fix this by introducing a new field to RasterDataset called `dtype` that, if set, forces a cast to that type right before the dataset level transform is called. With this we can fix the dtype expectations on a per dataset level.

### DIFF
--- a/torchgeo/datasets/chesapeake.py
+++ b/torchgeo/datasets/chesapeake.py
@@ -40,7 +40,6 @@ class Chesapeake(RasterDataset, abc.ABC):
     """
 
     is_image = False
-    dtype = torch.long
 
     # subclasses use the 13 class cmap by default
     cmap = {
--- a/torchgeo/datasets/chesapeake.py
+++ b/torchgeo/datasets/chesapeake.py
@@ -40,6 +40,7 @@ class Chesapeake(RasterDataset, abc.ABC):
     """
 
     is_image = False
+    dtype = torch.long
 
     # subclasses use the 13 class cmap by default
     cmap = {
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -437,7 +437,7 @@ class RasterDataset(GeoDataset):
             data = data.to(self.dtype)
 
         if self.is_image:
-            sample["image"] = data
+            sample["image"] = data.float()
         else:
             sample["mask"] = data
 
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -437,7 +437,7 @@ class RasterDataset(GeoDataset):
             data = data.to(self.dtype)
 
         if self.is_image:
-            sample["image"] = data.float()
+            sample["image"] = data
         else:
             sample["mask"] = data
 
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -302,6 +302,8 @@ class RasterDataset(GeoDataset):
 
         Returns:
             the dtype of the dataset
+
+        .. versionadded:: 5.0
         """
         if self.is_image:
             return torch.float32
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -296,6 +296,9 @@ class RasterDataset(GeoDataset):
     #: Color map for the dataset, used for plotting
     cmap: dict[int, tuple[int, int, int, int]] = {}
 
+    #: dtype to force onto the dataset (overrides the dtype of the file via a cast)
+    dtype: Optional[torch.dtype] = None
+
     def __init__(
         self,
         root: str = "data",
@@ -429,10 +432,14 @@ class RasterDataset(GeoDataset):
             data = self._merge_files(filepaths, query, self.band_indexes)
 
         sample = {"crs": self.crs, "bbox": query}
+
+        if self.dtype is not None:
+            data = data.to(self.dtype)
+
         if self.is_image:
             sample["image"] = data.float()
         else:
-            sample["mask"] = data.long()
+            sample["mask"] = data
 
         if self.transforms is not None:
             sample = self.transforms(sample)
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -298,7 +298,7 @@ class RasterDataset(GeoDataset):
 
     @property
     def dtype(self) -> torch.dtype:
-        """dtype of the dataset (overrides the dtype of the data file via a cast).
+        """dtype of the dataset (overrides the dtype of the data file via a cast).  # noqa: D403, E501
 
         Returns:
             the dtype of the dataset
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -434,13 +434,11 @@ class RasterDataset(GeoDataset):
 
         sample = {"crs": self.crs, "bbox": query}
 
-        if self.dtype is not None and not self.is_image:
-            data = data.to(self.dtype)
-        if self.dtype is not None and self.is_image:
-            warn(
-                "Custom dtype is explicitely set, however the current RasterDataset is"
-                + " an image, so no action will be taken."
-            )
+        if self.dtype is not None:
+            if self.is_image:
+                warn("Custom dtype is explicitly set, but dtype is only valid for mask RasterDatasets.")
+            else:
+                data = data.to(self.dtype)
 
         if self.is_image:
             sample["image"] = data.float()
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -296,9 +296,13 @@ class RasterDataset(GeoDataset):
     #: Color map for the dataset, used for plotting
     cmap: dict[int, tuple[int, int, int, int]] = {}
 
-    #: dtype to force onto the dataset (overrides the dtype of the file via a cast)
     @property
     def dtype(self) -> torch.dtype:
+        """dtype of the dataset (overrides the dtype of the data file via a cast).
+
+        Returns:
+            the dtype of the dataset
+        """
         if self.is_image:
             return torch.float32
         else:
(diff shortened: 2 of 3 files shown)


torchgeo/datasets/geo.py
```diff
@@ -427,10 +430,14 @@ def __getitem__(self, query: BoundingBox) -> Dict[str, Any]:
             data = self._merge_files(filepaths, query, self.band_indexes)
 
         sample = {"crs": self.crs, "bbox": query}
+
+        if self.dtype is not None:
+            data = data.to(self.dtype)
+
         if self.is_image:
             sample["image"] = data.float()
```
adamjstewart, line 438: We should remove this cast too for consistency.

adamjstewart (commented):
Technically this is an API change (we introduce a new class attribute) so it should wait until 0.5.0, but it also fixes an unintended consequence of #992, so it could also be argued that it should go in 0.4.1. Not sure which to choose.

We'll need to take this PR into account when thinking about #985.

torchgeo/datasets/geo.py
calebrob6: all images should be float though

calebrob6: let's see what happens though :)

calebrob6: It looks like this assumption is built into RasterDataset -- https://github.com/microsoft/torchgeo/blob/main/tests/datasets/test_geo.py#L201

There is a test that has a custom dataset with uint32 dtype that checks to make sure it has been converted by RasterDataset to float

adamjstewart: We can remove that test if we don't plan on unifying dtypes across datasets.

adamjstewart:
Just to complicate things, different bands of the same image may have different dtypes: https://github.com/microsoft/torchgeo/discussions/1182

Not sure how to help with this one. We can't stack all bands of the image into one tensor unless the tensor has a single shared dtype.

calebrob6:
> Just to complicate things, different bands of the same image may have different dtypes: #1182

That needs to be handled by casting in a pre-processing step. It doesn't make sense for RasterDataset to have different dtypes for different bands because the input to a neural network needs to be a float32, 16, 8, .. not a int/float mashup.

torchgeo/datasets/geo.py
calebrob6: I think it makes sense to cast images to floats because that is what will be passed to a network

adamjstewart:
In the case of #1182 it may make more sense for the PIXEL_QA band to be read as a mask instead of an image.

torchgeo/datasets/geo.py
adamjstewart: And I thought it would make sense to cast masks to long because that is what will be passed to torchmetrics 😆 

I think if a user specifies `dtype=torch.float16` when constructing the dataset and we instead use `torch.float32` just because it's an image, they'll be quite confused and will open a bug report. We should be consistent.

adamjstewart:
The only thing holding this PR back is that `dtype` affects masks but not images. This is very counterintuitive, isn't documented, and there's no warning letting you know this. Is there any reason `dtype` shouldn't affect both images and masks?

calebrob6:
`Conv2d` expects inputs to be float32s, therefore it makes a lot of sense to always make inputs float32s. If the naming is confusing then we can rename it `mask_dtype` or something.

> Is there any reason dtype shouldn't affect both images and masks?

Often the masks will be longs while the inputs will be float32s.

You can try running this:
```
import torch
import torch.nn as nn

conv = nn.Conv2d(3, 3, 3, bias=False)

for dtype in [
    torch.float32,
    torch.float64,
    torch.complex64,
    torch.complex128,
    torch.float16,
    torch.bfloat16,
    torch.uint8,
    torch.int8,
    torch.int16,
    torch.int32,
    torch.int64,
    torch.bool,
]:
    x = torch.ones(3, 9, 9, dtype=dtype)
    try:
        y = conv(x)
        print(f"{dtype} works")
    except:
        print(f"{dtype} doesn't work")
```

You should get:
```
torch.float32 works
torch.float64 doesn't work
torch.complex64 doesn't work
torch.complex128 doesn't work
torch.float16 doesn't work
torch.bfloat16 doesn't work
torch.uint8 doesn't work
torch.int8 doesn't work
torch.int16 doesn't work
torch.int32 doesn't work
torch.int64 doesn't work
torch.bool doesn't work
```

adamjstewart:
Kinda surprised float64 doesn't work.

I would be fine with calling it `mask_dtype`. We can always rename it later if someone provides a use case for changing image dtype.

calebrob6:
Actually, calling it `mask_dtype` doesn't make sense -- RasterDatasets are either images or masks, they only have one dtype.

adamjstewart:
Let's call it dtype then. I would either let it apply to both images and masks, or raise an error when someone uses not-float32 for an image.

calebrob6:
How about a UserWarning? Raising an all-out error seems extreme to me when everything should otherwise work.

calebrob6:
This should help with https://github.com/microsoft/torchgeo/issues/849 btw

torchgeo/datasets/geo.py
```diff
@@ -427,10 +431,19 @@ def __getitem__(self, query: BoundingBox) -> Dict[str, Any]:
...
 
         sample = {"crs": self.crs, "bbox": query}
+
+        if self.dtype is not None and not self.is_image:
+            data = data.to(self.dtype)
+        if self.dtype is not None and self.is_image:
+            warn(
+                "Custom dtype is explicitely set, however the current RasterDataset is"
+                + " an image, so no action will be taken."
+            )
```
adamjstewart, line 441:
```suggestion
        if self.dtype is not None:
            if self.is_image:
                warn("Custom dtype is explicitly set, but dtype is only valid for mask RasterDatasets.")
            else:
                data = data.to(self.dtype)
```
This feels cleaner

adamjstewart:
Needs tests, see https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html

torchgeo/datasets/chesapeake.py
```diff
@@ -40,6 +40,7 @@ class Chesapeake(RasterDataset, abc.ABC):
     """
 
     is_image = False
+    dtype = torch.long
```
adamjstewart, line 43: I'm guessing we need to set this for all other current mask datasets?
```console
$ grep 'is_image = False' *.py
agb_live_woody_density.py:    is_image = False
astergdem.py:    is_image = False
cdl.py:    is_image = False
chesapeake.py:    is_image = False
cms_mangrove_canopy.py:    is_image = False
esri2020.py:    is_image = False
eudem.py:    is_image = False
globbiomass.py:    is_image = False
nlcd.py:    is_image = False
```

torchgeo/datasets/geo.py
```diff
@@ -429,10 +433,20 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
             data = self._merge_files(filepaths, query, self.band_indexes)
 
         sample = {"crs": self.crs, "bbox": query}
+
+        if self.dtype is not None:
+            if self.is_image:
+                warn(
+                    "Custom dtype is explicitly set, but dtype is only valid for mask"
+                    + " RasterDatasets."
```
adamjstewart, line 441:
```suggestion
                    " RasterDatasets."
```
Fun fact: Python strings automatically concatenate even without the +.

adamjstewart (commented):
I'm fine with this approach. We could also use something like:
```python
@property
def dtype(self) -> torch.dtype:
    if self.is_image:
        return torch.float32
    else:
        return torch.long
```
Then we would only need to override this for the small number of datasets that have a different dtype.

torchgeo/datasets/geo.py
calebrob6: Yeah, I'm not a fan lol, if you forget a comma in your list of strings it will happily fail :)

torchgeo/datasets/chesapeake.py
calebrob6: only datasets where the dtype of the mask needs to be overridden

calebrob6: > I'm fine with this approach. We could also use something like:
> 
> ```python
> @property
> def dtype(self) -> torch.dtype:
>     if self.is_image:
>         return torch.float32
>     else:
>         return torch.long
> ```
> 
> Then we would only need to override this for the small number of datasets that have a different dtype.

But the whole point is that we sometimes want to allow for non `torch.long` masks

adamjstewart: Yes, and adding `dtype = torch.float32` to a subclass will override this property. If `dtype` isn't set in a subclass, the property will be used instead.

calebrob6: And then always cast to dtype in `__getitem__(...)`? Why is that better than what we're currently doing in this PR? 

Also, we could set this for the other mask datasets if we wanted to be explicit! Let me know.

adamjstewart: Many raster files are stored in uint8–uint32. Instead of adding `dtype = torch.long` or `dtype = torch.float32` to all of those datasets, we can set it once in the base class and override it only when a dataset needs to be different (none of our current datasets, but future pixel-wise regression datasets).

adamjstewart, line 43: This can be removed now

torchgeo/datasets/geo.py
```diff
@@ -296,6 +296,18 @@ class RasterDataset(GeoDataset):
     #: Color map for the dataset, used for plotting
     cmap: dict[int, tuple[int, int, int, int]] = {}
 
+    @property
+    def dtype(self) -> torch.dtype:
+        """The dtype of the dataset (overrides the dtype of the data file via a cast).
+
+        Returns:
+            the dtype of the dataset
```
adamjstewart, line 304: Can you add versionadded?

Issue #496, opened 2022-04-03
Better plotting of true color images


adamjstewart:
I never know the best way to create true color images from satellite imagery. QGIS normalizes images based on percentiles, but that blows up depending on whether or not your image contains clouds/snow. It seems like the best way is to divide by some constant, but I never know what constant to choose. You would think that it would be dividing by the maximum possible value, but that usually leads to an image that is much to dark. Is there a more scientific way to go about this?

I found https://www.sciencedirect.com/science/article/pii/S0034425719300422, which seems very well thought out. But I feel like implementing this for every dataset would be an entire research project, it's non-trivial to apply. Even having an "official" constant to divide by would be better than our current approach.

adamjstewart:
I wonder if this would help simplify our plotting code: https://earthpy.readthedocs.io/en/latest/api/earthpy.plot.html#earthpy.plot.plot_rgb

adamjstewart:
I don't think we should do this, it's too much work. Let's do #1263 instead.

PR #1151, opened 2023-02-28
SSL4EO-S12: add new dataset/datamodule

adamjstewart:
Adds the [SSL4EO-S12](https://github.com/zhu-xlab/SSL4EO-S12) dataset.

Some things the dataset doesn't currently support because I don't need them, but could potentially be added someday:

* Automatic downloading: same as SEN12MS and So2Sat, we haven't yet figured out automatic downloads from the TUM servers
* RGB version, 100-patch MSI subset, 50K RGB subset: alternate versions of the dataset on OneDrive and Google Drive
* Combining s1, s2c, or s2a: we don't yet have an easy model for supporting this in trainers and transforms
* Spectral band subsets: adds complexity to the dataset
* GeoDataset: each file has geographic metadata, we could build a GeoDataset

Plotting seems broken at the moment, will upload plots once that's working better.


tests/data/ssl4eo/data.py
```diff
@@ -0,0 +1,130 @@
...
+            18.588542158464236,
+        ),
+    }
+
+    if path.endswith("VH.tif") or path.endswith("VV.tif"):
+        profile["dtype"] = "float32"
+
+    if path.endswith("B1.tif") or path.endswith("B9.tif"):
+        profile["width"] = profile["height"] = SIZE // 6
+    elif (
```
adamjstewart, line 88: There's probably a better way to do this...

torchgeo/datasets/ssl4eo.py
```diff
@@ -0,0 +1,229 @@
...
+    .. note::
+
+       This dataset can be downloaded using:
+
+       .. code-block:: console
+
+          $ RSYNC_PASSWORD=m1660427.001 rsync -av rsync://m1660427.001@dataserv.ub.tum.de/m1660427.001/ .
+
+       The dataset is about 1.5 TB when compressed and 3.7 TB when uncompressed, and
+       takes roughly 36 hrs to download, 1 hr to checksum, and XX hrs to extract.
```
adamjstewart, line 45: Popcorn not included. BYOP.

Still trying to extract the full dataset, will update XX once I know the full time it takes.

torchgeo/datasets/ssl4eo.py
```diff
@@ -0,0 +1,229 @@
...
+        Args:
+            index: index to return
+
+        Returns:
+            sample with an "image" in 5xCxHxW format where the 5 indexes over the same
+                patch sampled from different points in time by the SeCo method
+        """
+        directory = os.path.join(self.root, self.split, f"{index // self.times:07}")
+        subdirs = sorted(os.listdir(directory))
+        directory = os.path.join(directory, subdirs[index % self.times])
```
adamjstewart, line 141: This is somewhat hard-coded, but is much faster than storing a list of every directory like we do in SeCo.

torchgeo/datasets/ssl4eo.py
```diff
@@ -0,0 +1,229 @@
...
+
+        images = []
+        for band in self.metadata[self.split]["bands"]:
+            filename = os.path.join(directory, f"{band}.tif")
+            with rasterio.open(filename) as f:
+                image = f.read(out_shape=(1, self.size, self.size)).astype(np.float32)
```
adamjstewart, line 147: Each Sentinel band is in its original resolution (10–60 m) and needs to be resampled to the higher resolution.

torchgeo/datasets/ssl4eo.py
```diff
@@ -0,0 +1,229 @@
...
+        Args:
+            sample: a sample returned by :meth:`__getitem__`
+            show_titles: flag indicating whether to show titles above each panel
+            suptitle: optional string to use as a suptitle
+
+        Returns:
+            a matplotlib Figure with the rendered sample
+        """
+        if self.split == "s1":
+            # See Sentinel1.plot
```
adamjstewart, line 206: Wish we could call the plot methods of Sentinel1 and Sentinel2 directly so we didn't have to copy the same logic into every dataset. I bet we could figure out a way to do this...

torchgeo/datasets/ssl4eo.py
```diff
@@ -0,0 +1,231 @@
...
+            co_polarization = sample["image"][0]
+            cross_polarization = sample["image"][1]
+            ratio = co_polarization / cross_polarization
+
+            co_polarization = torch.clamp(co_polarization / 0.3, min=0, max=1)
+            cross_polarization = torch.clamp(cross_polarization / 0.05, min=0, max=1)
+            ratio = torch.clamp(ratio / 25, min=0, max=1)
```
adamjstewart, line 215: @RitwikGupta this plotting trick isn't working, all I see is a black image with red specs. I know you're not the biggest fan of RGB plotting, but this dataset presumably only has one type of image processing so it should be possible to reliably plot. Not sure how to tell if the image is power/decibel/amplitude or gamma/sigma and what that actually means for plotting. Any advice? You can download the [100 patch subset](https://github.com/zhu-xlab/SSL4EO-S12) if you want to test the plotting method.

RitwikGupta: @adamjstewart The Sentinel-1 imagery they pulled from GEE is in the decibel scale. Normally, Sentinel-1 data from Copernicus is in the power scale. The ratio you want to calculate should be in the power scale. This code will work to do visualize this specific data:

```
cross_pol = np.exp(cross_pol / 10)
co_pol = np.exp(co_pol / 10)
ratio = cross_pol / co_pol

fig, ax = plt.subplots(1, 4, figsize=(20, 10))
ax[0].imshow(cross_pol, cmap='bone')
ax[1].imshow(co_pol, cmap='bone')
ax[2].imshow(ratio, cmap='bone')
ax[3].imshow(np.concatenate([cross_pol, co_pol, ratio], axis=2))
```
image

RitwikGupta: Here's the relevant GEE documentation that says they convert to decibel when pulling S1 data in GEE: https://developers.google.com/earth-engine/datasets/catalog/COPERNICUS_S1_GRD
image

adamjstewart: 1. I thought it was supposed to be [co_pol, cross_pol, ratio], not [cross_pol, co_pol, ratio]?
2. Am I not supposed to do [normalization](https://gis.stackexchange.com/a/400780/123758) (co_pol / 0.3, cross_pol / 0.05, ratio / 25)?
3. Even with divide by 10, no normalization, swap co_pol/cross_pol, the images don't look great:
Figure_1

RitwikGupta: Which image are you using? I can test with the same one.
* I think you can keep [co_pol, cross_pol, ratio].
* Are you doing `np.exp(x / 10)` or just `x /10`?
* You should use the log scale values in that post for viz purposes only. It's not required, but it may help. Do min-max norm to those ranges, not just division.

adamjstewart:
S1 plotting is definitely still broken, but I would be okay with merging this as is and fixing S1 plotting in a follow-up PR since I need this dataset soon.

I'm planning on adding a datamodule for this, but will wait until #1168 is merged so I can properly test it.

isaaccorley (dismissed):

torchgeo/datasets/ssl4eo.py
```diff
@@ -0,0 +1,250 @@
...
+
+from .geo import NonGeoDataset
+from .utils import check_integrity, extract_archive
+
+
+class SSL4EOS12(NonGeoDataset):
+    """SSL4EO-S12 dataset.
+
+    The `Self-Supervised Learning for Earth Observation - Sentinel-1/2
+    <https://github.com/zhu-xlab/SSL4EO-S12>`_ dataset is a large-scale mutilmodal
```
nilsleh, line 24: "multimodal" typo

torchgeo/datasets/ssl4eo.py
```diff
@@ -0,0 +1,250 @@
...
+                "B12",
+            ],
+        },
+    }
+
+    def __init__(
+        self,
+        root: str = "data",
+        split: str = "s2c",
+        seasons: int = 1,
```
nilsleh, line 102: As far as I understand "seasons" decides how many seasonal patches to sample and it does so randomly. Would you also want to give control over the seasons in the sense if a user only wants to sample from a particular or a subset of the four seasons?

adamjstewart: That could be a feature worth adding later, but I'm trying to keep things simple until someone actually needs that functionality.

torchgeo/datasets/ssl4eo.py
adamjstewart: Submitted a PR to fix upstream README where I copied this from.

nilsleh (approved):

calebrob6:
Can you post an example from plot across the three different splits?

adamjstewart:
@wangyi111 do you want to review this? We couldn't figure out a good way to plot the images, but the rest of the data loader should work.

torchgeo/datasets/ssl4eo.py
wangyi111: why do you want to use power values to plot? In my paper, I simply plot the dB values (99% percentile and scale each one to [0,1])

wangyi111:
@adamjstewart I think the problem is from `SSL4EO.metadata['s2c']['bands']`. One band 'B10' missing.

adamjstewart:
Good catch! I think I looked at s2a, didn't see B10, and assumed s2c would be the same. Let me fix that.

torchgeo/datasets/ssl4eo.py
wangyi111: then can also plot dB for each channel? e.g. [VV,VH,VV/VH]?

adamjstewart: I think that's what I was doing originally and it didn't work either. I just copied the plotting code from our `Sentinel1` dataset, which worked fine for data downloaded from ASF DAAC, but apparently not for data downloaded from GEE.

calebrob6:
I don't have the S1 downloaded so can't check it
image

adamjstewart:
> I don't have the S1 downloaded so can't check it

You can download the 100-patch subset for testing.

torchgeo/datasets/ssl4eo.py
adamjstewart, line 1: @calebrob6 this directly contradicts the change made in #1027. Should we revert that PR too? Given that:

1. All of our datasets use a different scale factor, ignoring which satellite their images come from
2. All images are normalized before they are returned from the dataset or datamodule, changing the appropriate scale factor

I'm starting to think we should use one of the default enhancements from QGIS:

1. Min / max
2. 2% / 98%
3. ± 2 std dev

There could even be an option to select which of these you want to use. We can decide and implement this later, but I want to make sure we're not contradicting ourselves and won't revert #1027 multiple times until we come up with a uniform policy. This also relates to #496.

calebrob6: dividing by 10k obviously didn't work though

calebrob6:
image

s1 plotting is broken

torchgeo/datasets/ssl4eo.py
adamjstewart: Then why did we merge #1027?

calebrob6:
I'm guessing the images from gee and other places are different

adamjstewart:
Yep. We could just plot 2 grayscale images instead of one RGB image for now and fix this in the future when someone (me) has a better understanding of SAR imagery. I'd prefer not to merge this with broken plotting, but we could open an issue and promise to fix it before 0.5.0 is released.

RitwikGupta:
Sorry all. I was out on vacation and then sick for a bit. I will take a look at this again. GEE imagery is in the decibel scale instead of the power scale most Sentinel-1 imagery is released in. This is standard for analysis and visualization.

adamjstewart:
Any updates @RitwikGupta?

We might want to just merge this PR and fix S1 plotting at a later date.

calebrob6:
Yep, I would vote to make the plot show _something_ and merge

adamjstewart:
Alright, see how this looks:

Figure_1
Figure_2
Figure_3
Figure_4

calebrob6 (approved):

PR #1166, opened 2023-03-09
Models: fix preprocessing transforms

adamjstewart:
Our pre-trained models come with preprocessing transforms that should be used to normalize and reshape images used with them. However, these transforms don't actually run. This PR fixes them and adds tests to ensure that all transforms actually work.

### DIFF
--- a/torchgeo/datamodules/seco.py
+++ b/torchgeo/datamodules/seco.py
@@ -5,10 +5,12 @@
 
 from typing import Any
 
+import kornia.augmentation as K
 import torch
 from einops import repeat
 
 from ..datasets import SeasonalContrastS2
+from ..transforms import AugmentationSequential
 from .geo import NonGeoDataModule
 
 
@@ -17,40 +19,6 @@ class SeasonalContrastS2DataModule(NonGeoDataModule):
 
     .. versionadded:: 0.5
     """
-
-    # https://github.com/ServiceNow/seasonal-contrast/blob/8285173ec205b64bc3e53b880344dd6c3f79fa7a/datasets/bigearthnet_dataset.py#L13  # noqa: E501
-    mean = torch.tensor(
-        [
-            340.76769064,
-            429.9430203,
-            614.21682446,
-            590.23569706,
-            950.68368468,
-            1792.46290469,
-            2075.46795189,
-            2218.94553375,
-            2266.46036911,
-            2246.0605464,
-            1594.42694882,
-            1009.32729131,
-        ]
-    )
-    std = 2 * torch.tensor(
-        [
-            554.81258967,
-            572.41639287,
-            582.87945694,
-            675.88746967,
-            729.89827633,
-            1096.01480586,
-            1273.45393088,
-            1365.45589904,
-            1356.13789355,
-            1302.3292881,
-            1079.19066363,
-            818.86747235,
-        ]
-    )
 
     def __init__(
         self, batch_size: int = 64, num_workers: int = 0, **kwargs: Any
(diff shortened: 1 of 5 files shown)


torchgeo/models/resnet.py
```diff
@@ -22,46 +22,16 @@
...
-                1096.01480586,
-                1273.45393088,
-                1365.45589904,
-                1356.13789355,
-                1302.3292881,
-                1079.19066363,
-                818.86747235,
-            ]
-        ),
+        mean=torch.tensor([590.23569706, 614.21682446, 429.9430203]),
```
adamjstewart, line 33: We only have weights for RGB images

torchgeo/models/resnet.py
```diff
@@ -22,46 +22,16 @@
...
+        std=2 * torch.tensor([675.88746967, 582.87945694, 572.41639287]),
```
adamjstewart, line 34: SeCo and SSL4EO use 2x the std dev for normalization:

* https://github.com/ServiceNow/seasonal-contrast/blob/8285173ec205b64bc3e53b880344dd6c3f79fa7a/datasets/bigearthnet_dataset.py#L112,L113
* https://github.com/zhu-xlab/SSL4EO-S12/blob/d2868adfada65e40910bfcedfc49bc3b20df2248/src/benchmark/pretrain_ssl/datasets/SSL4EO/ssl4eo_dataset.py#L37,L38

torchgeo/models/resnet.py
```diff
@@ -22,46 +22,16 @@
 _zhu_xlab_transforms = AugmentationSequential(
     K.Resize(256),
     K.CenterCrop(224),
-    K.Normalize(mean=0, std=10000),
+    K.Normalize(mean=torch.tensor(0), std=torch.tensor(10000)),
```
adamjstewart, line 25: Values must be torch tensors, not ints

isaaccorley (dismissed):

torchgeo/models/resnet.py
adamjstewart: Fixed by https://github.com/kornia/kornia/pull/2269, but we should still use tensors or floats until the next release is the minimum release we support.

torchgeo/models/resnet.py
nilsleh: https://github.com/ServiceNow/seasonal-contrast/blob/8285173ec205b64bc3e53b880344dd6c3f79fa7a/datasets/bigearthnet_dataset.py#L111 Is this also different from the Kornia Normalize method, as they use some min max operation and also clip to range [0, 255]?

adamjstewart: True, this is actually quite different. They substract (mean - 2 * std), not mean, and they divide by (4 * std), not std. I have no idea why they scale by 255, seems odd unless you're trying to plot something. And we obviously aren't clamping (again, only useful for plotting). 

EDIT: oh, is it because they need to load it as a PIL image? Maybe that removes the 255 when converting to a Tensor.

adamjstewart: Not sure how they're even using PIL here since it only supports RGB, not MSI.

adamjstewart: (I guess that's why SeCo only had RGB weights?)

adamjstewart: Ah yes, they divide by 10K * 255 for S1 but not for S2?

adamjstewart: The more I look at these, the less sure I am that they are correct. May want to reach out to the authors.

adamjstewart: Once we figure this out, we should also fix the SeCO datamodule I recently added. According to @wangyi111, it actually doesn't matter that much what the exact normalization is.

adamjstewart: Opened an issue to get to the bottom of this. I actually think we're using the wrong normalization (this was only used for BigEarthNet, not for any other dataset). My guess is the SeCo normalization, but we'll see if we get a response. If I don't hear anything back, let's just assume SeCo normalization.

https://github.com/ServiceNow/seasonal-contrast/issues/19

nilsleh (dismissed):

torchgeo/datamodules/seco.py
```diff
@@ -63,17 +31,30 @@ def __init__(
...
+            _max = torch.tensor([88, 103, 129])
+            _mean = torch.tensor([0.485, 0.456, 0.406])
+            _std = torch.tensor([0.229, 0.224, 0.225])
+
+            _min = repeat(_min, "c -> (t c)", t=seasons)
+            _max = repeat(_max, "c -> (t c)", t=seasons)
+            _mean = repeat(_mean, "c -> (t c)", t=seasons)
+            _std = repeat(_std, "c -> (t c)", t=seasons)
+
+            self.aug = AugmentationSequential(
```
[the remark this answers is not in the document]
calebrob6: Why are there three normalizations right after each other that use things like "_max - _min" for std:
```
K.Normalize(mean=_min, std=_max - _min),
K.Normalize(mean=torch.tensor(0), std=1 / torch.tensor(255)),
K.Normalize(mean=_mean, std=_std),
```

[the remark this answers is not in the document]
adamjstewart: I wish I knew too. But that's what ServiceNow used, so I think it's best to use the same normalization they did.

torchgeo/datamodules/seco.py
```diff
@@ -63,17 +31,30 @@ def __init__(
...
+        super().__init__(SeasonalContrastS2, batch_size, num_workers, **kwargs)
 
+        bands = kwargs.get("bands", SeasonalContrastS2.rgb_bands)
         seasons = kwargs.get("seasons", 1)
-        self.mean = repeat(self.mean, "c -> (t c)", t=seasons)
-        self.std = repeat(self.std, "c -> (t c)", t=seasons)
 
-        super().__init__(SeasonalContrastS2, batch_size, num_workers, **kwargs)
+        # Normalization only available for RGB dataset
+        # https://github.com/ServiceNow/seasonal-contrast/blob/8285173ec205b64bc3e53b880344dd6c3f79fa7a/datasets/seco_dataset.py  # noqa: E501
```
calebrob6, line 40:
```suggestion
        # See the following for where the normalizations are defined https://github.com/ServiceNow/seasonal-contrast/blob/8285173ec205b64bc3e53b880344dd6c3f79fa7a/datasets/seco_dataset.py  # noqa: E501
```

calebrob6 (dismissed):

calebrob6 (approved):

PR #1168, opened 2023-03-09
SeCo/BYOL: add datamodule, RandomSeasonContrast

adamjstewart:
This PR contains the following changes:

- [x] SeCo: faster initialization
- [x] SeCo: add datamodule
- [x] SeCo: add RandomSeasonContrast
- [x] BYOL: add RandomSeasonContrast

RandomSeasonContrast is the idea from SeCo and SSL4EO where images taken of the same location at different points in time are used as inputs to a contrastive SSL model instead of taking two random crops of the same image.


torchgeo/datasets/seco.py
```diff
@@ -88,57 +97,42 @@ def __init__(
...
+        .. versionchanged:: 0.5
+           Image shape changed from 5xCxHxW to SCxHxW
         """
-        scene_name = self.scenes[index]
-        patch_names = self.scene_to_patches[scene_name]
+        directory = os.path.join(
+            self.root, self.metadata[self.version]["directory"], f"{index:06}"
+        )
+        patch_dirs = glob.glob(os.path.join(directory, "*"))
+        patch_dirs = random.sample(patch_dirs, self.seasons)
```
adamjstewart, line 131: RandomSeasonContrast. User defines how many seasons (patches) they want per location, and we randomly return them. Note that returning all 5 seasons will return them in a random order, but this was already the case in the previous implementation.

torchgeo/trainers/byol.py
```diff
@@ -387,8 +387,16 @@ def training_step(self, *args: Any, **kwargs: Any) -> Tensor:
         """
         batch = args[0]
         x = batch["image"]
+
+        in_channels = self.hyperparams["in_channels"]
+        assert x.size(1) == in_channels or x.size(1) == 2 * in_channels
+        if x.size(1) == in_channels:
```
adamjstewart, line 393: This trainer now supports both datasets with and without RandomSeasonContrast

torchgeo/trainers/byol.py
```diff
@@ -401,33 +409,10 @@ def training_step(self, *args: Any, **kwargs: Any) -> Tensor:
         return loss
 
     def validation_step(self, *args: Any, **kwargs: Any) -> None:
```
adamjstewart, line 411: I don't think it makes sense to include validation/test/predict in an SSL trainer. For datasets without labels, there is no easy way to evaluate performance. For datasets with labels, there is no way to know which task we are trying to evaluate (classification, regression, semantic segmentation, etc.)

tests/trainers/test_byol.py
```diff
@@ -60,20 +49,20 @@ def test_custom_augment_fn(self) -> None:
 
 
 class TestBYOLTask:
-    @pytest.mark.parametrize(
-        "name,classname",
-        [
-            ("chesapeake_cvpr_7", ChesapeakeCVPRDataModule),
-            ("chesapeake_cvpr_prior", ChesapeakeCVPRDataModule),
```
adamjstewart, line 67: We could test any of our 50+ supervised datasets, but I think it makes more sense to test our unsupervised datasets. Right now, we only have SeCo and BYOL, but I'm working on adding SSL4EO and SimCLR/MoCo which will also be tested in a similar fashion.

torchgeo/trainers/byol.py
isaaccorley: I think this is fine. There could be an infinite number of downstream tasks

torchgeo/trainers/byol.py
isaaccorley: Nice

torchgeo/trainers/byol.py
adamjstewart: https://pytorch-lightning.readthedocs.io/en/stable/notebooks/course_UvA-DL/13-contrastive-learning.html defines a validation_step but I can't figure out why

isaaccorley (approved):
LGTM

PR #1169, opened 2023-03-10
Landsat: better default bands

adamjstewart:
For Landsat, B8–B11 only exist for raw imagery and top of atmosphere reflectance, not surface reflectance. I still think it makes sense to default to surface reflectance, but we should not include bands that don't exist for it.

At least on Google Earth Engine (GEE), Landsat 1–5 Multispectral Scanner (MSS) do not have surface reflectance, so default to the band names for raw imagery.

References:

* [SR](https://developers.google.com/earth-engine/datasets/catalog/LANDSAT_LC09_C02_T1_L2)
* [TOA](https://developers.google.com/earth-engine/datasets/catalog/LANDSAT_LC09_C02_T1_TOA)
* [raw](https://developers.google.com/earth-engine/datasets/catalog/LANDSAT_LC09_C02_T1)

### DIFF
--- a/torchgeo/datasets/landsat.py
+++ b/torchgeo/datasets/landsat.py
@@ -203,6 +203,7 @@ class Landsat8(Landsat):
 
 
 class Landsat9(Landsat8):
-    """Landsat 9 Operational Land Imager (OLI) and Thermal Infrared Sensor (TIRS)."""
+    """Landsat 9 Operational Land Imager (OLI-2) and Thermal Infrared Sensor (TIRS-2).
+    """
 
     filename_glob = "LC09_*_{}.*"
--- a/torchgeo/datasets/landsat.py
+++ b/torchgeo/datasets/landsat.py
@@ -138,8 +138,8 @@ class Landsat1(Landsat):
 
     filename_glob = "LM01_*_{}.*"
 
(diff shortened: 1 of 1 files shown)


torchgeo/datasets/landsat.py
```diff
@@ -207,23 +198,11 @@ class Landsat8(Landsat):
...
-        "SR_B10",
-        "SR_B11",
-    ]
+    default_bands = ["SR_B1", "SR_B2", "SR_B3", "SR_B4", "SR_B5", "SR_B6", "SR_B7"]
     rgb_bands = ["SR_B4", "SR_B3", "SR_B2"]
 
 
 class Landsat9(Landsat8):
-    """Landsat 9 Operational Land Imager (OLI) and Thermal Infrared Sensor (TIRS)."""
+    """Landsat 9 Operational Land Imager (OLI-2) and Thermal Infrared Sensor (TIRS-2)."""  # noqa: E501
```
adamjstewart, line 206: Black and pydocstyle require one-line docstrings to have the closing quotes on the same line, but flake8 complains that the line is too long. So we have to ignore the latter in order to appease the former.

isaaccorley (approved):

PR #1191, opened 2023-03-23
RasterDataset: fix bug when separate_files and no date in regex

adamjstewart:
Fixes a bug introduced by yours truly waaaaay back in #75.

The reason we haven't discovered this bug until now is that it has extremely limited scope. It only affects `RasterDataset`s where `separate_files = True` and `filename_regex` does not contain a `date` group. This is exactly 0 of our builtin datasets, but is unfortunately exactly what happened to @TolgaAktas in #1182.

Thanks for reporting this @TolgaAktas!

### DIFF
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -415,7 +415,7 @@ class RasterDataset(GeoDataset):
                     directory = os.path.dirname(filepath)
                     match = re.match(filename_regex, filename)
                     if match:
-                        if "date" in match.groupdict():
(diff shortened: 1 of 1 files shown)


isaaccorley (approved):

adamjstewart:
I've introduced far more bugs than you. This bug is rather minor compared to #786.

TolgaAktas:
Does 0.4.1 fix this issue?

adamjstewart:
Yes, it should. Let me know if you still encounter this issue.

PR #1194, opened 2023-03-24
Add script to download LandSat data via Google Earth Engine

wangyi111:


wangyi111:

adamjstewart:
This looks great! Only problem is I can't actually tell what you changed. I think the first step should be to upload the original script in its current state. Then the next step would be to submit a PR that splits “choosing which patches to sample” and “downloading those patches” into two separate scripts. This allows us to do the first part in serial and the second part in parallel. See our SSL Planning Document for the full list of steps. The end goal is that the script will work for all satellites, not just Landsat.

wangyi111:
> This looks great! Only problem is I can't actually tell what you changed. I think the first step should be to upload the original script in its current state. Then the next step would be to submit a PR that splits “choosing which patches to sample” and “downloading those patches” into two separate scripts. This allows us to do the first part in serial and the second part in parallel. See our SSL Planning Document for the full list of steps. The end goal is that the script will work for all satellites, not just Landsat.

@adamjstewart “choosing which patches to sample” includes "check if the patches exist and get information from GEE", this is the most time-consuming part. Therefore, “choosing which patches to sample” has to be paralleled to speed up. The pure downloading is fast.

The goal to download any satellite from GEE sounds cool, I'll check later. The problem is for each satellite, some information is encoded in different names or bits (e.g. cloud), but this can be documented.

adamjstewart:
> “choosing which patches to sample” has to be paralleled to speed up.

How would this work? Do you divide the Earth up into 1 degree grid cells and allow each one to run in parallel? Obviously most grid cells have no cities, and some have more cities than others. Do you sample the number of patches from each grid cell based on how many top 10K cities exist? Or do you parallelize by city instead of by grid cell? Do you decide on ~1.1M patches first, then put them in an R-tree and remove overlapping, then download, or does the R-tree step come after downloading?

> The problem is for each satellite, some information is encoded in different names or bits (e.g. cloud), but this can be documented.

Yep, we can either make these script arguments or store them in a dictionary.

I think either way, we should probably add the existing script first and then make changes to that script in a separate PR. Can you open a PR to add the existing script from SSL4EO-S12? You'll also need to run the steps from https://torchgeo.readthedocs.io/en/stable/user/contributing.html#linters to get the tests to pass.

adamjstewart:
Follow-up question: what level of parallelism does GEE allow a single account to have? Is there any rate throttling per IP address? We could also have multiple people create GEE accounts and add human-level parallelism if needed to get around that. Normally I would feel bad about cheating the system, but we are indeed an entire team of people working on this.

wangyi111:
> > “choosing which patches to sample” has to be paralleled to speed up.
> 
> How would this work? Do you divide the Earth up into 1 degree grid cells and allow each one to run in parallel? Obviously most grid cells have no cities, and some have more cities than others. Do you sample the number of patches from each grid cell based on how many top 10K cities exist? Or do you parallelize by city instead of by grid cell? Do you decide on ~1.1M patches first, then put them in an R-tree and remove overlapping, then download, or does the R-tree step come after downloading?
> 
> > The problem is for each satellite, some information is encoded in different names or bits (e.g. cloud), but this can be documented.
> 
> Yep, we can either make these script arguments or store them in a dictionary.
> 
> I think either way, we should probably add the existing script first and then make changes to that script in a separate PR. Can you open a PR to add the existing script from SSL4EO-S12? You'll also need to run the steps from https://torchgeo.readthedocs.io/en/stable/user/contributing.html#linters to get the tests to pass.


That's the trade-off. I simply made the whole script parallelized, i.e., from sampling to downloading, with a potential risk of overlap comes between some parallel processors. Can have a final overlap check in the end, hopefully wouldn't affect much.
> Do you decide on ~1.1M patches first, then put them in an R-tree and remove overlapping, then download

This would be a choice to ensure no risk, just you have to repeat it several times to reach the needed number of locations. I haven't implemented this strategy though.

wangyi111:
> Follow-up question: what level of parallelism does GEE allow a single account to have? Is there any rate throttling per IP address? We could also have multiple people create GEE accounts and add human-level parallelism if needed to get around that. Normally I would feel bad about cheating the system, but we are indeed an entire team of people working on this.

Not very sure. I think there's a quote limit for one account. like 100 requests/s (6000 requests/min). I didn't meet this limit, but if parallelizing the script with 100 cores, maybe it'll be met:)

experiments/download_landsat_GEE.py
```diff
@@ -0,0 +1,610 @@
...
+    transform = Affine.translation(coords[0][0] - xres / 2, coords[0][1] + yres / 2) * Affine.scale(xres, -yres)
+    profile = {
+        'driver': 'GTiff',
+        'width': width,
+        'height': height,
+        'count': channels,
+        'crs': '+proj=latlong',
+        'transform': transform,
+        'dtype': img.dtype,
+        'compress': 'None'
```
calebrob6, line 401: why not compress='lzw', predictor=2?

adamjstewart: Would like to know which loads faster, a smaller file that has to be decompressed, or a larger file that is already decompressed.

wangyi111: No idea about that. In the end, we converted all the raw data to a single lmdb file for faster data loading.

PR #1197, opened 2023-03-26
L7 Irish: add new dataset/datamodule

yichiac:
Overview

Adding [Landsat 7 Cloud Cover Assessment Validation Data](https://landsat.usgs.gov/landsat-7-cloud-cover-assessment-validation-data) as a RasterDataset to TorchGeo.

Main changes:

Added functions to initialize the L7 Irish dataset, sample from it, and plot images.
Added data generation script and artificial data corresponding to the same.
Wrote tests to check the functionality of the L7 Irish class.
Integrated L7Irish Data Module and included different splits.
Included dataset details into the docs.

Here is a sample plot of an image and its mask from the dataset:

demo2


.DS_Store
adamjstewart, line 1: Can you remove these `.DS_Store` files? You can add these to your global `.gitignore` to prevent them from being added in the future.

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,253 @@
...
+from .utils import BoundingBox, download_url, extract_archive, working_dir
+
+
+class Landsat7Irish(RasterDataset):
+    
+    """ Landsat 7 Cloud Cover Assessment Validation Data.
+
+    The `Landsat 7 Cloud Cover Assessment Validation Data <https://landsat.usgs.gov/landsat-7-cloud-cover-assessment-validation-data>`__ dataset is based on Landsat 7
+    Enhanced Thematic Mapper Plus (ETM+) Level-1G scenes, displayed in the biomes listed below. Manually generated cloud masks are used to train and validate cloud cover 
+    assessment algorithms, which in turn are intended to compute the percentage of cloud cover in each scene. *Originally, this was reported as 207 masks, but one mask
```
adamjstewart, line 32: This last sentence isn't necessary

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,253 @@
...
+     * Fill
+     * Cloud Shadow
+     * Clear
+     * Thin Cloud
+     * Cloud
+
+     If you use this dataset in your research, please cite the following papers:
+
+     * U.S. Geological Survey, 2016. L7 Irish Cloud Validation Masks. U.S. Geological Survey data release. doi:10.5066/F7XD0ZWC.
+     * Scaramuzza, P.L., Bouchard, M.A. & Dwyer, J.L. (2012). Development of the Landsat data continuity mission cloud-cover assessment algorithms. IEEE Transactions on Geoscience and Remote Sensing, 50(4), 1140-1154. doi:10.1109/TGRS.2011.2164087.
```
adamjstewart, line 49:
```suggestion
     * https://doi.org/10.5066/F7XD0ZWC
     * https://doi.org/10.1109/TGRS.2011.2164087
```

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,253 @@
...
+            Raises:
+                RuntimeError: if ``download=False`` and data is not found, or checksums
+                    don't match
+            """
+        self.root = root
+        self.download = download
+        self.checksum = checksum
+
+        self._verify()
+        super().__init__(root, download, checksum)
```
adamjstewart, line 101: The base class doesn't have a download or checksum argument, this is wrong.

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,253 @@
...
+                f"Dataset not found in `root={self.root}` and `download=False`, "
+                "either specify a different `root` directory or use `download=True` "
+                "to automatically download the dataset."
+            )
+
+        # Download the dataset
+        self._download()
+        self._extract()
+
+    @abc.abstractmethod
```
adamjstewart, line 130: This isn't abstract, you implemented it.

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,253 @@
...
+            "image": img.float(),
+            "mask": mask.long(),
+        }
+
+        if self.transforms is not None:
+            sample = self.transforms(sample)
+
+        return sample
+
+    def _verify(self) -> None:
```
adamjstewart, line 166: You have verify defined twice

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,253 @@
...
+
+        Returns:
+            sample of image, mask and metadata at that index
+
+        Raises:
+            IndexError: if query is not found in the index
+        """
+        hits = self.index.intersection(tuple(query), objects=True)
+        img_filepaths = cast(List[str], [hit.object for hit in hits])
+        mask_filepaths = [path.replace("images", "masks") for path in img_filepaths] # need to change
```
adamjstewart, line 145: Still needs to be changed

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,253 @@
...
+
+        if not img_filepaths:
+            raise IndexError(
+                f"query: {query} not found in index with bounds: {self.bounds}"
+            )
+
+        img = self._merge_files(img_filepaths, query, self.band_indexes)
+        mask = self._merge_files(mask_filepaths, query, self.band_indexes)
```
adamjstewart, line 153: Mask needs to be converted from 0, 64, 128, 192, 255 to 0, 1, 2, 3, 4

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,253 @@
...
+        showing_predictions = "prediction" in sample
+        if showing_predictions:
+            predictions = sample["prediction"].numpy()
+            num_panels += 1
+
+        fig, axs = plt.subplots(1, num_panels, figsize=(num_panels * 4, 5))
+        axs[0].imshow(image)
+        axs[0].axis("off")
+        axs[1].imshow(mask, vmin=0, vmax=4,
+                        cmap=self._lc_cmap, interpolation="none")
```
adamjstewart, line 237: This variable isn't defined, you need to use cmap

adamjstewart (changes requested):
Obviously still needs documentation and tests

adamjstewart:
See https://torchgeo.readthedocs.io/en/stable/user/contributing.html#linters for how to fix the failing tests.

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,253 @@
...
+from .geo import NonGeoDataset, RasterDataset
+from .landsat import Landsat7
+from .utils import BoundingBox, download_url, extract_archive, working_dir
+
+
+class Landsat7Irish(RasterDataset):
+    
+    """ Landsat 7 Cloud Cover Assessment Validation Data.
+
+    The `Landsat 7 Cloud Cover Assessment Validation Data <https://landsat.usgs.gov/landsat-7-cloud-cover-assessment-validation-data>`__ dataset is based on Landsat 7
```
adamjstewart, line 30:
```suggestion
class L7Irish(RasterDataset):
    """L7 Irish dataset.

    The `L7 Irish <https://landsat.usgs.gov/landsat-7-cloud-cover-assessment-validation-data>`__ dataset is based on Landsat 7
```
People usually refer to the dataset as L7 Irish. The original masks were produced by Irish et al.

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,253 @@
...
+    Enhanced Thematic Mapper Plus (ETM+) Level-1G scenes, displayed in the biomes listed below. Manually generated cloud masks are used to train and validate cloud cover 
+    assessment algorithms, which in turn are intended to compute the percentage of cloud cover in each scene. *Originally, this was reported as 207 masks, but one mask 
+    was erroneously counted.
+
+    Dataset format:
+    * 102 scenes from this dataset were used in a cloud validation study described in Foga, Scaramuzza et al. (2017)
```
adamjstewart, line 36: This sentence has nothing to do with the dataset format, you should instead list # images, image size, # spectral bands, Level-1G, etc.

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,253 @@
...
+    classes = ["Fill", "Cloud Shadow", "Clear", "Thin Cloud", "Cloud"]
+    
+    cmap = {
+        0: (0,0,0),
+        64: (64,64,64),
+        128: (128,128,128),
+        192: (192,192,192),
+        255: (255,255,255),
+    }
+
```
adamjstewart, line 77: You're missing filename_glob and filename_regex. Without those, this dataset won't actually locate any files. See https://torchgeo.readthedocs.io/en/stable/tutorials/custom_raster_dataset.html.

.DS_Store
yichiac: .DS_Store files have been removed.

.gitignore
```diff
@@ -144,3 +144,4 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+.DS_Store
```
adamjstewart, line 147: This is an OS-specific file, not a project-specific file. I usually add those to global gitignores (`~/.gitignore`), not project gitignores.

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,251 @@
...
+from .geo import NonGeoDataset, RasterDataset
+from .landsat import Landsat7
+from .utils import BoundingBox, download_url, extract_archive, working_dir
+
+
+class L7Irish(RasterDataset):    
+    """L7 Irish dataset.
+
+    The `L7 Irish <https://landsat.usgs.gov/landsat-7-cloud-cover-assessment-validation-data>`__ dataset is based on Landsat 7
```
adamjstewart, line 29: There's a lot of useful information you used to have in this sentence that is now gone

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,251 @@
...
+     * Cloud
+
+     If you use this dataset in your research, please cite the following papers:
+
+     * https://doi.org/10.5066/F7XD0ZWC
+     * https://doi.org/10.1109/TGRS.2011.2164087
+     """
+    
+    # naming... [need to change]
+    filename_regex = r"""
```
adamjstewart, line 53: This regex is incorrect

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,251 @@
...
+        (?P<wrs_row>\d{3})
+        _(?P<date>\d{8})
+        _(?P<processing_date>\d{8})
+        _(?P<collection_number>\d{2})
+        _(?P<collection_category>[A-Z0-9]{2})
+        _(?P<band>[A-Z0-9_]+)
+        \.
+    """
+
+    tarfile_glob = "*.tar.gz"
```
adamjstewart, line 68: No need to encode this at the class level, it isn't specific to this class. You can just hardcode this where you're planning on using it

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,251 @@
...
+    cmap = {
+        0: (0, 0, 0),
+        64: (64, 64, 64),
+        128: (128, 128, 128),
+        192: (192, 192, 192),
+        255: (255, 255, 255),
+    }
+
+    def __init__(
+        self, root: str = "data", download: bool = False, checksum: bool = False
```
adamjstewart, line 94: Missing most of the args from the base class. Your class should contain the existing args from RasterDataset, plus new ones for download/checksum

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,251 @@
...
+        self.download = download
+        self.checksum = checksum
+    
+        for i in list(self.cmap.keys()):
+            self.cmap[np.ceil(i/64).astype(int)] = self.cmap.pop(i)
+
+        lc_colors = np.zeros((max(self.cmap.keys()) + 1, 3))
+        lc_colors[list(self.cmap.keys())] = list(self.cmap.values())
+        lc_colors = lc_colors / 255
+        _lc_cmap = ListedColormap(lc_colors)
```
adamjstewart, line 121: What does this code do and why is it needed? "I just copied it from CDL" isn't a good answer 😛

torchgeo/datasets/l7irish.py
adamjstewart: Still wrong

torchgeo/datasets/l7irish.py
adamjstewart: Still missing filename_glob

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,251 @@
...
+        Returns:
+            sample of image, mask and metadata at that index
+
+        Raises:
+            IndexError: if query is not found in the index
+        """
+        hits = self.index.intersection(tuple(query), objects=True)
+        img_filepaths = cast(List[str], [hit.object for hit in hits])
+        mask_filepaths = [path.replace("","") for path in img_filepaths]
+        # mask_filepaths = [path.replace("images", "masks") for path in img_filepaths]
```
adamjstewart, line 183: What is this code supposed to do? Because it doesn't work at the moment.

torchgeo/datasets/l7irish.py
adamjstewart: Still need to convert

torchgeo/datasets/l7irish.py
adamjstewart: Still broken

adamjstewart (changes requested):
Still a lot of failing tests. See Contributing docs for how to fix style issues.

.gitignore
```diff
@@ -143,4 +143,4 @@ venv.bak/
 dmypy.json
 
 # Pyre type checker
-.pyre/
+.pyre/
```
adamjstewart, line 146: Can you add back the newline to this file? There should be a text editor setting for this

tests/data/l7irish/data.py
```diff
@@ -0,0 +1,116 @@
...
+import numpy as np
+import rasterio
+from rasterio.crs import CRS
+
+SIZE = 64
+
+np.random.seed(0)
+random.seed(0)
+
+wkt = """
```
adamjstewart, line 20: This can be made simpler:
```pycon
>>> import rasterio
>>> src = rasterio.open("L71226098_09820011112_B10.TIF")
>>> src.crs
CRS.from_epsg(32719)
```
So you can replace this giant string with that one line on line 53.

tests/data/l7irish/data.py
```diff
@@ -0,0 +1,116 @@
...
+    profile = {}
+    profile["driver"] = "GTiff"
+    profile["dtype"] = dtype
+    profile["count"] = num_channels
+    profile["crs"] = CRS.from_wkt(wkt),
+    profile["transform"] = rasterio.transform.from_bounds(0, 0, 1, 1, 1, 1) # may need to change
+    profile["height"] = SIZE
+    profile["width"] = SIZE
+    profile["compress"] = "lzw"
+    profile["predictor"] = 2 # predictor??
```
adamjstewart, line 58: I believe this controls the compression level? @calebrob6 would know.

tests/data/l7irish/data.py
```diff
@@ -0,0 +1,116 @@
...
+        AXIS["(N)",north,ORDER[2],LENGTHUNIT["metre",1]]]
+"""
+
+def create_file(path: str, dtype: str, num_channels: int) -> None:
+    profile = {}
+    profile["driver"] = "GTiff"
+    profile["dtype"] = dtype
+    profile["count"] = num_channels
+    profile["crs"] = CRS.from_wkt(wkt),
+    profile["transform"] = rasterio.transform.from_bounds(0, 0, 1, 1, 1, 1) # may need to change
```
adamjstewart, line 54: Should be:
```pycon
>>> import rasterio
>>> src = rasterio.open("L71226098_09820011112_B10.TIF")
>>> src.transform
Affine(30.0, 0.0, 462884.99999999994,
       0.0, -30.0, 4071915.0)
```

tests/data/l7irish/data.py
```diff
@@ -0,0 +1,116 @@
...
+    profile["height"] = SIZE
+    profile["width"] = SIZE
+    profile["compress"] = "lzw"
+    profile["predictor"] = 2 # predictor??
+    cmap = {
```
adamjstewart, line 59: None of the real files have a cmap, so there's no reason to add them to the files.

tests/data/l7irish/data.py
```diff
@@ -0,0 +1,116 @@
...
+        0: (0, 0, 0),
+        64: (64, 64, 64),
+        128: (128, 128, 128),
+        192: (192, 192, 192),
+        255: (255, 255, 255),
+    }
+
+    Z = np.random.randint(size=(SIZE, SIZE), low=0, high=8)
```
adamjstewart, line 67: For high I usually use `np.iinfo(dtype).max`

tests/data/l7irish/data.py
```diff
@@ -0,0 +1,116 @@
...
+    "polar_north", "polar_south", "subtropical_north", "subtropical_south", 
+    "tropical"]
+patches = [
+    "p226_r98", "p108_r18", "p107_r34", "p100_r82",
+    "p110_r12", "p10_r120", "p118_r40", "p101_r72", 
+    "p111_r55"]
+filenames = [
+    "L71226098_09820011112_", "L71108018_01820010729_", "L71107034_03420010722_B10", "L71100082_08220011212_",
+    "L71110012_01220010727_", "L71010120_12020011119_", "L71118040_04020010703_B10", "L71101072_07220011203_",
+    "L71111055_05520010413_"]
```
adamjstewart, line 86: You're missing a mask file. It's probably sufficient to use 1 dir and 1 patch, otherwise your fake dataset will contain 800+ files that need to be added to git.

tests/data/l7irish/data.py
```diff
@@ -0,0 +1,116 @@
...
+bands = ["B10", "B20", "B30", "B40", "B50", "B61", "B62", "B70", "B80"]
```
adamjstewart, line 87: Some bands have different resolution which you'll need to handle. See the `data.py` I added in my SSL4EO-S12 PR.

tests/datasets/test_l7irish.py
```diff
@@ -0,0 +1,136 @@
...
+# bbox = BoundingBox(minx, maxx, miny, maxy, mint, maxt)
+# sample = ds[bbox]
+# ds.plot(sample)
+# plt.show()
+
+class TestL7Irish:
+    @pytest.fixture
+    def dataset(self, monkeypatch: MonkeyPatch, tmp_path: Path) -> L7Irish:
+        monkeypatch.setattr(torchgeo.datasets.l7irish, "download_url", download_url)
+        # need to create new md5s?
```
adamjstewart, line 38: These are calculated by `data.py`

tests/datasets/test_l7irish.py
```diff
@@ -0,0 +1,136 @@
...
+            "subtropical_south": "8346d73a983396c5d41b577c3a94bc26",
+            "tropical": "abe19b22b5d031e6b609cc7207706c3d",
+        }
+        
+        monkeypatch.setattr(L7Irish, "md5s", md5s)
+        url = os.path.join("tests", "data", "l7irish", "{}.tar.gz")
+        monkeypatch.setattr(L7Irish, "url", url)
+        monkeypatch.setattr(plt, "show", lambda *args: None)
+        root = str(tmp_path)
+        transforms = nn.Identity() # why does it need nn?
```
adamjstewart, line 56: We need some transform to get 100% coverage. It just needs to be an identity function, doesn't need to be PyTorch's.

tests/datasets/test_l7irish.py
```diff
@@ -0,0 +1,136 @@
...
+        return L7Irish(root, transforms=transforms, download=True, checksum=True)
+
+    def test_getitem(self, dataset: L7Irish) -> None:
+        x = dataset[dataset.bounds]
+        assert isinstance(x, dict)
+        assert isinstance(x["crs"], CRS) # this is different from landcoverai
+        assert isinstance(x["mask"], torch.Tensor)
```
adamjstewart, line 63: You should also test that an image is returned

tests/datasets/test_l7irish.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def test_or(self, dataset: L7Irish) -> None:
+        ds = dataset | dataset
+        assert isinstance(ds, UnionDataset)
+
+    def test_full_year(self, dataset: L7Irish) -> None:
+        bbox = dataset.bounds
+        time = datetime(2001, 11, 12).timestamp() # randomly select one time
+        query = BoundingBox(bbox.minx, bbox.maxx, bbox.miny, bbox.maxy, time, time)
+        next(dataset.index.intersection(tuple(query)))
```
adamjstewart, line 77: What is this test designed to do? I think you can remove it after you change your dataset to only contain 1 fake image.

tests/datasets/test_l7irish.py
```diff
@@ -0,0 +1,136 @@
...
+        L7Irish(root)
+
+    def test_plot(self, dataset: L7Irish) -> None:
+        query = dataset.bounds
+        x = dataset[query]
+        dataset.plot(x, suptitle="Test")
+        plt.close()
+
+    def test_plot_prediction(self, dataset: L7Irish) -> None: 
+        query = dataset.bounds # same as cdl, slightly different from landcoverai
```
adamjstewart, line 96: Looks identical to me

tests/datasets/test_l7irish.py
```diff
@@ -0,0 +1,136 @@
...
+
+# make new file for test l7irish in test directory
+# don't need params
+# replace Landsat8 with l7irish
+# monkeypatch: make fake data and checksums and use it for testing
+# create fake data: /tests/data/ to see example, be careful about different resolutions,
+# black
+# isort
+# flake8
+# put pull request @...
```
adamjstewart, line 136: Should remove comments before merging

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,271 @@
...
+
+    The `L7 Irish <https://landsat.usgs.gov/landsat-7-cloud-cover-assessment-validation-data>`__ dataset is based on Landsat 7
+    Enhanced Thematic Mapper Plus (ETM+) Level-1G scenes, displayed in the biomes listed below. Manually generated cloud masks 
+    are used to train and validate cloud cover assessment algorithms, which in turn are intended to compute the percentage of 
+    cloud cover in each scene.
+
+    Dataset features:
+
+    * 206 scenes from Landsat-7 Enhanced Thematic Mapper Plus (ETM+) tiles
+    * Imagery from global tiles between June 2000 - December 2001
```
adamjstewart, line 33:
```suggestion
    * Imagery from global tiles between June 2000--December 2001
```
En dash instead of hyphen to denote a range

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,271 @@
...
+    * 9 Level-1 spectral bands with 30 and 60 m per pixel resolution
+    
+    Dataset format:
+
+    * Images are composed of multiple single channel geotiffs
+    * Labels are multiclass, stored in a single geotiffs file per image
+    * Level-1 metadata (MTL.txt file)
+    * Landsat-7 Enhanced Thematic Mapper Plus (ETM+) bands: (B01, B02, B03, B04, B05, B06-1, B06-2, B07, B08)
```
adamjstewart, line 41: Might be less confusing if you use the same band names as the files

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,271 @@
...
+    192. Thin Cloud
+    255. Cloud
+
+    If you use this dataset in your research, please cite the following papers:
+
+    * https://doi.org/10.5066/F7XD0ZWC
+    * https://doi.org/10.1109/TGRS.2011.2164087
+    """    
+
+    url = "https://huggingface.co/datasets/torchgeo/l7irish/resolve/main/{}.tar.gz" #noqa E501
```
adamjstewart, line 57:
```suggestion
    url = "https://huggingface.co/datasets/torchgeo/l7irish/resolve/main/{}.tar.gz"  # noqa: E501
```

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,271 @@
...
+        
+    classes = ["Fill", "Cloud Shadow", "Clear", "Thin Cloud", "Cloud"]
+    
+    cmap = {
+        0: (0, 0, 0),
+        64: (64, 64, 64),
+        128: (128, 128, 128),
+        192: (192, 192, 192),
+        255: (255, 255, 255),
+    }
```
adamjstewart, line 79: I think the keys should actually be 0–4, see L8 Biome

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,271 @@
...
+    tarfile_glob = "*.tar.gz"
+    filename_glob = "L*.TIF"
+
+    filename_regex = r""" 
+        ^L7[12]
+        (?P<wrs_path>\d{3})
+        (?P<wrs_row>\d{3})
+        _(?P<wrs_row>\d{3})
+        (?P<processing_date>\d{8})
+        _(?P<band>[B10-80])
```
adamjstewart, line 91: This does not mean what you think it means. `[B10-80]` just means it can contain one of the following characters: B, 0, 1, 8, -. You should instead use `B\d{2}`

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,271 @@
...
+        \.TIF
+    """
+
+
+    def __init__(  # check variable here!!!
+        self,
+        root: str = "data",
+        download: bool = False,
+        checksum: bool = False,
```
adamjstewart, line 100: Download and checksum should prob come last

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,271 @@
...
+        res: Optional[float] = None,
+    ) -> None:
+        
+        """Initialize a new Landsat 7 Cloud Cover Assessment Validation dataset instance.
+
+            Args:
+                root: root directory where dataset can be found
+                transforms: a function/transform that takes input sample and its target as
+                    entry and returns a transformed version
+                cache: if True, cache file handle to speed up repeated sampling
```
adamjstewart, line 113: Missing docs for some parameters, just copy-n-paste from a different dataset

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,271 @@
...
+
+        Returns:
+            sample of image, mask and metadata at that index
+
+        Raises:
+            IndexError: if query is not found in the index
+        """
+        hits = self.index.intersection(tuple(query), objects=True)
+        img_filepaths = cast(List[str], [hit.object for hit in hits])
+        mask_filepaths = [path.replace(path.split("/")[-1], path.split("/")[-2]+"_mask2019.TIF") for path in img_filepaths]
```
adamjstewart, line 185: Might be easier to follow this logic if you convert the list comprehension to a for-loop. Also, never use `"/"` since that won't work on Windows. Instead use `os.path.split(path)`.

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,271 @@
...
+        image = np.rollaxis(sample["image"].numpy().astype("uint8").squeeze(), 0, 3)
+        mask = sample["mask"].numpy().astype("uint8").squeeze()
+
+        num_panels = 2
+        showing_predictions = "prediction" in sample
+        if showing_predictions:
+            predictions = sample["prediction"].numpy()
+            num_panels += 1
+        
+        # converting 5 classes to [0,1,2,3,4] for the plot function.
```
adamjstewart, line 246: We can do this when we define cmap. Honestly, instead of using cmap, why not remove it and divide the mask/prediction by 4 so the image is in the 0–1 range? Then we don't need to do any fancy cmap stuff, we just plot the raw grayscale values.

adamjstewart (changes requested):

yichiac:

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,290 @@
...
+        (?P<wrs_row>\d{3})
+        _(?P=wrs_row)
+        (?P<date>\d{8})
+        _(?P<band>B\d{2})
+        \.TIF$
+    """
+    date_format = "%Y%m%d"
+
+    separate_files = True
+    rgb_bands = ["B40", "B20", "B30"]
```
adamjstewart, line 82:
```suggestion
    rgb_bands = ["B30", "B20", "B10"]
```
See https://developers.google.com/earth-engine/datasets/catalog/LANDSAT_LE07_C02_T1_TOA

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,290 @@
...
+        root: str = "data",
+        crs: Optional[CRS] = None,
+        res: Optional[float] = None,
+        transforms: Optional[Callable[[Dict[str, Any]], Dict[str, Any]]] = None,
+        bands: Sequence[str] = all_bands,
+        cache: bool = True,
+        download: bool = False,
+        checksum: bool = False,
+    ) -> None:
+        """Initialize a new Landsat7 Cloud Cover Assessment Validation dataset instance.
```
adamjstewart, line 96:
```suggestion
        """Initialize a new L7Irish instance.
```

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,290 @@
...
+    separate_files = True
+    rgb_bands = ["B40", "B20", "B30"]
+    all_bands = ["B10", "B20", "B30", "B40", "B50", "B61", "B62", "B70", "B80"]
+
+    def __init__(
+        self,
+        root: str = "data",
+        crs: Optional[CRS] = None,
+        res: Optional[float] = None,
+        transforms: Optional[Callable[[Dict[str, Any]], Dict[str, Any]]] = None,
```
adamjstewart, line 90: Bands should come before transforms to keep the same order as the base class

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,290 @@
...
+        bands: Sequence[str] = all_bands,
+        cache: bool = True,
+        download: bool = False,
+        checksum: bool = False,
+    ) -> None:
+        """Initialize a new Landsat7 Cloud Cover Assessment Validation dataset instance.
+
+        Args:
+            root: root directory where dataset can be found
+            transforms: a function/transform that takes input sample and its target as
```
adamjstewart, line 100: These should be in the same order as the function signature

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,290 @@
...
+                entry and returns a transformed version
+            cache: if True, cache file handle to speed up repeated sampling
+            crs: :term:`coordinate reference system (CRS)` to warp to
+            (defaults to the CRS of the first file found)
```
adamjstewart, line 104: Some lines are missing indentation, which is causing the doc tests to fail

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,290 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""Landsat 7 Cloud Cover Assessment Validation Data."""
```
adamjstewart, line 4:
```suggestion
"""L7 Irish dataset."""

```

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,290 @@
...
+    def _verify(self) -> None:
+        """Verify the integrity of the dataset.
+
+        Raises:
+            RuntimeError: if ``download=False`` but dataset is missing or checksum fails
+        """
+        # Check if the extracted files already exist
+        pathname = os.path.join(self.root, "**", self.filename_glob)
+        for fname in glob.iglob(pathname, recursive=True):
+            if not fname.endswith(".tar.gz"):
```
adamjstewart, line 134: This if-statement is unnecessary, the glob already prevents matching files ending in .tar.gz

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,290 @@
...
+        img_filepaths = cast(List[str], [hit.object for hit in hits])
+
+        if not img_filepaths:
+            raise IndexError(
+                f"query: {query} not found in index with bounds: {self.bounds}"
+            )
+
+        data_list: List[Tensor] = []
+
+        for band in self.all_bands:
```
adamjstewart, line 190: See my latest commit to Shradha's PR, this was wrong and I fixed it

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,290 @@
...
+                            start = match.start("band")
+                            end = match.end("band")
+                            filename = filename[: start - 1] + band + filename[end:]
+                    filepath = os.path.join(directory, filename)
+                    band_filepaths.append(filepath)
+                    data_list.append(self._merge_files(band_filepaths, query))
+
+        img = torch.cat(data_list)
+
+        mask_filepaths = [
```
adamjstewart, line 209: This is super hard to read. Can you convert it from a list comprehension to a for-loop?

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,290 @@
...
+            path.replace(
+                path.split(os.sep)[-1], path.split(os.sep)[-2] + "_mask2019.TIF"
```
adamjstewart, line 211: I would use `os.path.split(path)` instead of `path.split(os.sep)`

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,290 @@
...
+            )
+            for path in img_filepaths
+        ]
+
+        mask = self._merge_files(mask_filepaths, query)
+        mask_mapping = {0: 0, 64: 1, 128: 2, 192: 3, 255: 4}
```
adamjstewart, line 217: No reason to map 0 to 0

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,290 @@
...
+        num_panels = 2
+        showing_predictions = "prediction" in sample
+        if showing_predictions:
+            predictions = sample["prediction"].numpy().astype("uint8").squeeze()
+            num_panels += 1
+
+        fig, axs = plt.subplots(1, num_panels, figsize=(num_panels * 4, 5))
+        axs[0].imshow(image)
+        axs[0].axis("off")
+        axs[1].imshow(mask, vmin=0, vmax=4, cmap="gray", interpolation="none")
```
adamjstewart, line 273:
```suggestion
        axs[1].imshow(mask, vmin=0, vmax=4, cmap="gray")
```

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,290 @@
...
+        axs[1].axis("off")
+        if show_titles:
+            axs[0].set_title("Image")
+            axs[1].set_title("Mask")
+
+        if showing_predictions:
+            axs[2].imshow(
+                predictions, vmin=0, vmax=4, cmap="gray", interpolation="none"
```
adamjstewart, line 281:
```suggestion
                predictions, vmin=0, vmax=4, cmap="gray"
```

adamjstewart (changes requested):
Also needs datamodule tests

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,290 @@
...
+        "polar_north": "73923dcaf1b9b79bad82de1aa0740d1e",
+        "polar_south": "3bc9f4c6f8955b10b4d55d23e0ab2da7",
+        "subtropical_north": "f8f039970256902e6e9ebd6747589294",
+        "subtropical_south": "8346d73a983396c5d41b577c3a94bc26",
+        "tropical": "abe19b22b5d031e6b609cc7207706c3d",
+    }
+
+    classes = ["Fill", "Cloud Shadow", "Clear", "Thin Cloud", "Cloud"]
+
+    filename_glob = "L7*.TIF"
```
adamjstewart, line 68:
```suggestion
    filename_glob = "L7*_B10.TIF"
```
Otherwise you will match each scene 9 times

torchgeo/datasets/l7irish.py
yichiac: Yes, I will convert it to a for-loop for easier understanding. Thank you.

adamjstewart:
Shradha's PR was merged, so you'll need to resolve merge conflicts, either on GitHub or via CLI.

.gitignore
adamjstewart, line 1: Can you undo changes to this file?

docs/api/datamodules.rst
```diff
@@ -11,11 +11,19 @@ Chesapeake Land Cover
 
 .. autoclass:: ChesapeakeCVPRDataModule
 
+
+L7 Irish
+^^^^^^^^
+
+.. autoclass:: L7IrishDataModule
+
+
```
adamjstewart, line 20: Can you replace double empty lines with single empty lines to save space?

adamjstewart:
Let me know when you get the tests to pass and I'll review again

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,287 @@
...
+                    if "date" in match.groupdict():
+                        start = match.start("band")
+                        end = match.end("band")
+                        filename = filename[:start] + band + filename[end:]
+                        if band in ["B62", "B70", "B80"]:
+                            filename = filename.replace(filename[2], "2", 1)
+                filepath = os.path.join(directory, filename)
+                band_filepaths.append(filepath)
+            image_list.append(self._merge_files(band_filepaths, query))
+        img = torch.cat(image_list)
```
adamjstewart, line 205: I would call this variable "image" since that's what we call the dict key

adamjstewart (changes requested):
Will investigate the plotting issue you discovered.

tests/data/l7irish/data.py
```diff
@@ -0,0 +1,120 @@
...
+    "B50.TIF",
+    "B61.TIF",
+    "B62.TIF",
+    "B70.TIF",
+    "B80.TIF",
+]
+
+filenames: FILENAME_HIERARCHY = {
+    "austral": {"p226_r98": [], "p227_r98": [], "p230_r95": []},
+    "boreal": {"p2_r27": [], "p128_r25": []},
```
adamjstewart, line 36: Some scenes are split into parts (p186_r32_1, p186_r32_2, etc.). Can you change one or two of these to match that so we know if can handle it properly?

torchgeo/datasets/l7irish.py
yichiac: Indeed, I change the conversion method to zfill().

tests/data/l7irish/data.py
```diff
@@ -0,0 +1,125 @@
...
+]
+
+for land_type, patches in filenames.items():
+    for patch in patches:
+        p, r = patch.split("_r")
+        if r.find("_") == -1:
+            r = r.zfill(3)
+        else:
+            r = r[: r.find("_")].zfill(3)
+        key = p[1:] + r
```
adamjstewart, line 53: Does p need zero padding too? Why not use:
```suggestion
        path, row = patch.split("_")[:2]
        path = path[1:].zfill(3)
        row = row[1:].zfill(3)
        key = path + row
```

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,290 @@
...
+
+from .geo import RasterDataset
+from .utils import BoundingBox, download_url, extract_archive
+
+
+class L7Irish(RasterDataset):
+    """L7 Irish dataset.
+
+    The `L7 Irish <https://landsat.usgs.gov/landsat-7-cloud-cover-assessment-validation-data>`__ dataset
+    is based on Landsat 7 Enhanced Thematic Mapper Plus (ETM+) Level-1G scenes.
```
adamjstewart, line 24: Worth mentioning that it's a validation dataset for cloud cover assessment algorithms

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,290 @@
...
+    ) -> None:
+        """Initialize a new L7Irish instance.
+
+        Args:
+            root: root directory where dataset can be found
+            crs: :term:`coordinate reference system (CRS)` to warp to
+                (defaults to the CRS of the first file found)
+            res: resolution of the dataset in units of CRS
+                (defaults to the resolution of the first file found)
+            bands: bands to return
```
adamjstewart, line 107:
```suggestion
            bands: bands to return (defaults to all bands)
```

adamjstewart (commented):
Can you add a datamodule too?

docs/api/geo_datasets.csv
```diff
@@ -12,6 +12,7 @@ Dataset,Type,Source,Size (px),Resolution (m)
 `GBIF`_,Points,Citizen Scientists,-,-
 `GlobBiomass`_,Masks,Landsat,"45,000x45,000",100
 `iNaturalist`_,Points,Citizen Scientists,-,-
+`L7 Irish`_,"Imagery, Masks",Landsat,"8,400x7,500","15, 30, 60"
```
adamjstewart, line 15:
```suggestion
`L7 Irish`_,"Imagery, Masks",Landsat,"8,400x7,500","15, 30"
```
New version has no 60 m bands

tests/conf/l7irish.yaml
```diff
@@ -0,0 +1,21 @@
...
+    model: "unet"
+    backbone: "resnet18"
+    weights: null
+    learning_rate: 1e-3
+    learning_rate_schedule_patience: 6
+    verbose: false
+    in_channels: 9
+    num_classes: 5
+    num_filters: 1
+    ignore_index: null
```
adamjstewart, line 14:
```suggestion
    ignore_index: 0
```
We want to ignore fill in our calculations. Doesn't matter much for testing, but worth making it realistic.

tests/data/l7irish/data.py
```diff
@@ -0,0 +1,117 @@
...
+    "L71002027_02720010604",
+    "L71143021_02120010803",
+]
+
+for land_type, patches in filenames.items():
+    for patch in patches:
+        path, row = patch.split("_")[:2]
+        path = path[1:].zfill(3)
+        row = row[1:].zfill(3)
+        key = path + row
```
adamjstewart, line 51:
```suggestion
        key = path[1:].zfill(3) + row[1:].zfill(3)
```

tests/data/l7irish/data.py
```diff
@@ -0,0 +1,117 @@
...
+        for prefix in prefixes:
+            if key in prefix:
+                for band in bands:
+                    if band in ["B62.TIF", "B70.TIF", "B80.TIF"]:
+                        prefix = prefix.replace(prefix[2], "2", 1)
+                    filenames[land_type][patch].append(f"{prefix}_{band}")
+
+        filenames[land_type][patch].append(f"L7_{patch}_newmask2015.TIF")
```
adamjstewart, line 59:
```suggestion
        filenames[land_type][patch].append(f"L7_{path}_{row}_newmask2015.TIF")
```
Mask filename doesn't include any optional suffix:
```console
$ ls mid_latitude_north/p186_r32_2/*mask*
mid_latitude_north/p186_r32_2/L7_p186_r32_newmask2015.TIF
```

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,293 @@
...
+    The `L7 Irish <https://www.sciencebase.gov/catalog/item/573ccf18e4b0dae0d5e4b109>`__ dataset
+    is based on Landsat 7 Enhanced Thematic Mapper Plus (ETM+) Level-1G scenes.
+    Manually generated cloud masks are used to train and validate cloud cover assessment algorithms,
+    which in turn are intended to compute the percentage of cloud cover in each scene.
+
+    Dataset features:
+
+    * 206 scenes from Landsat-7 ETM+ tiles
+    * Imagery from global tiles between June 2000--December 2001
+    * 9 Level-1 spectral bands with 15 and 30 mper pixel resolution
```
adamjstewart, line 32:
```suggestion
    * 9 Level-1 spectral bands with 15 and 30 m per pixel resolution
```

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,293 @@
...
+
+    Dataset format:
+
+    * Images are composed of multiple single channel geotiffs
+    * Labels are multiclass, stored in a single geotiffs file per image
+    * Level-1 metadata (MTL.txt file)
+    * Landsat-7 ETM+ bands: (B10, B20, B30, B40, B50, B61, B62, B70, B80)
+
+    Dataset classes (4):
```
adamjstewart, line 41: 5 classes, we're including cloud shadow

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,293 @@
...
+
+    0. Fill
+    1. Clear
+    2. Thin Cloud
+    3. Cloud
+
+    If you use this dataset in your research, please cite the following:
+
+    * https://doi.org/10.5066/F7XD0ZWC
+    * https://doi.org/10.1109/TGRS.2011.2164087
```
adamjstewart, line 51: Can you also add https://www.sciencebase.gov/catalog/item/573ccf18e4b0dae0d5e4b109? That's where most of our data actually comes from

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,293 @@
...
+                band_filepaths.append(filepath)
+            image_list.append(self._merge_files(band_filepaths, query))
+        image = torch.cat(image_list)
+
+        mask_filepaths = []
+        for filepath in filepaths:
+            mask_filepath = filepath.replace(
+                os.path.basename(filepath),
+                "L7_"
+                + os.path.basename(os.path.dirname(filepath))
```
adamjstewart, line 216: See my comment in `data.py`, this should only include the path and row, not any optional suffix. This is different from the 2019 version.

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,293 @@
...
+                filename = os.path.basename(filepath)
+                directory = os.path.dirname(filepath)
+                match = re.match(filename_regex, filename)
+                if match:
+                    if "date" in match.groupdict():
+                        start = match.start("band")
+                        end = match.end("band")
+                        filename = filename[:start] + band + filename[end:]
+                        if band in ["B62", "B70", "B80"]:
+                            filename = filename.replace(filename[2], "2", 1)
```
adamjstewart, line 205: If you wanted to make this more obvious what's going on:
```suggestion
                            filename = filename.replace("L71", "L72")
```

adamjstewart (changes requested):
Almost done. Just a few last changes. The 2015 version also has cloud shadow, so we should add that back.

torchgeo/datasets/l7irish.py
yichiac: Great catch, thanks!

docs/api/geo_datasets.csv
```diff
@@ -12,6 +12,7 @@ Dataset,Type,Source,Size (px),Resolution (m)
 `GBIF`_,Points,Citizen Scientists,-,-
 `GlobBiomass`_,Masks,Landsat,"45,000x45,000",100
 `iNaturalist`_,Points,Citizen Scientists,-,-
+`L7 Irish`_,"Imagery, Masks",Landsat,"8,400x7,500","15, 30"
 `LandCover.ai Geo`_,"Imagery, Masks",Aerial,"4,200--9,500",0.25--0.5
 `Landsat`_,Imagery,Landsat,"8,900x8,900",30
 `L8 Biome`_,"Imagery, Masks",Landsat,"8,900x8,900","15, 30"
```
adamjstewart, line 18: Can you move L8 Biome underneath L7 Irish? I think it was in the wrong location.

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,293 @@
...
+        "mid_latitude_north": "0860e218403d949f4b38e4f9f70e0087",
+        "mid_latitude_south": "c66bbeaa6dbf0ba2cd26b9eea89eb3a4",
+        "polar_north": "4a648c5bcc6b5e5eac0f8c57dce1a57e",
+        "polar_south": "a12e4d7fddaa377259328190f10a1c17",
+        "subtropical_north": "ebdfaee37ffc5ba1bd4763f7f72df97f",
+        "subtropical_south": "3670c9490753efe3d36927329bb87e2f",
+        "tropical": "f60c93d8609c72ac86e858105b6272f2",
+    }
+
+    classes = ["Fill", "Clear", "Thin Cloud", "Cloud"]
```
adamjstewart, line 72: Missing cloud shadow

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,293 @@
...
+        mask_filepaths = []
+        for filepath in filepaths:
+            path, row = os.path.basename(os.path.dirname(filepath)).split("_")[:2]
+            mask_filepath = filepath.replace(
+                os.path.basename(filepath), f"L7_{path}_{row}_newmask2015.TIF"
+            )
+            mask_filepaths.append(mask_filepath)
+
+        mask = self._merge_files(mask_filepaths, query)
+        mask_mapping = {64: 1, 128: 2, 191: 3, 255: 4}
```
adamjstewart, line 222:
```suggestion
        mask_mapping = {64: 1, 128: 2, 192: 3, 255: 4}
```
We didn't catch this bug due to the aforementioned mistake in `data.py`

tests/data/l7irish/data.py
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    if path.endswith("B80.TIF"):
+        profile["transform"] = Affine(
+            15.0, 0.0, 462892.49999999994, 0.0, -15.0, 4071907.5
+        )
+        profile["width"] = profile["height"] = SIZE * 2
+
+    if path.endswith("_newmask2015.TIF"):
+        Z = np.random.randint(5, size=(SIZE, SIZE), dtype=dtype)
```
adamjstewart, line 79: Technically the mask should contain 0, 64, 128, 192, 255, not 0–4. We made the same mistake when creating L8 Biome data. Not sure if we should fix this now or in a follow-up PR or just ignore it.

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,293 @@
...
+from torch import Tensor
+
+from .geo import RasterDataset
+from .utils import BoundingBox, download_url, extract_archive
+
+
+class L7Irish(RasterDataset):
+    """L7 Irish dataset.
+
+    The `L7 Irish <https://www.sciencebase.gov/catalog/item/573ccf18e4b0dae0d5e4b109>`__ dataset
```
adamjstewart, line 23:
```suggestion
    The `L7 Irish <https://landsat.usgs.gov/landsat-7-cloud-cover-assessment-validation-data>`__ dataset
```
Let's link to this site instead. It has a more accurate description of the mask classes and loads faster.

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,293 @@
...
+    is based on Landsat 7 Enhanced Thematic Mapper Plus (ETM+) Level-1G scenes.
+    Manually generated cloud masks are used to train and validate cloud cover assessment algorithms,
+    which in turn are intended to compute the percentage of cloud cover in each scene.
```
adamjstewart, line 26: Some of these lines are over 90 characters. Can you wrap these? I know we ignored line length because of the URL but we shouldn't exceed line length when we don't have to.

adamjstewart (changes requested):

tests/data/l7irish/data.py
yichiac: I change it to np.random.choice() to randomly choose values in [0, 64, 128, 191, 255].

torchgeo/datasets/l7irish.py
yichiac: The mask class is 191 instead of 192 in the 2015 version dataset.

torchgeo/datasets/l7irish.py
```diff
@@ -0,0 +1,294 @@
...
+    url = "https://huggingface.co/datasets/torchgeo/l7irish/resolve/main/{}.tar.gz"  # noqa: E501
+
+    md5s = {
+        "austral": "9c2629884c1e7251e24953e1e5f880de",
+        "boreal": "0a9f50998c0fb47c0cc226faf479f883",
+        "mid_latitude_north": "0860e218403d949f4b38e4f9f70e0087",
+        "mid_latitude_south": "c66bbeaa6dbf0ba2cd26b9eea89eb3a4",
+        "polar_north": "4a648c5bcc6b5e5eac0f8c57dce1a57e",
+        "polar_south": "a12e4d7fddaa377259328190f10a1c17",
+        "subtropical_north": "18a6b9b4684ae91bfdcc7b78ea1f42ee",
```
adamjstewart, line 68:
```suggestion
        "polar_north": "18a6b9b4684ae91bfdcc7b78ea1f42ee",
        "polar_south": "a12e4d7fddaa377259328190f10a1c17",
        "subtropical_north": "ebdfaee37ffc5ba1bd4763f7f72df97f",
```

adamjstewart (changes requested):

adamjstewart (approved):

PR #1200, opened 2023-03-28
Add L8 Biome Dataset

shradhasehgal:
#### Overview
Adds L8 Biome dataset as a RasterDataset to TorchGeo.

#### Main changes:
- Added functions to initialize the L8 Biome dataset, sample from it, and plot images.
- Added data generation script and artificial data corresponding to the same.
- Wrote tests to check the functionality of the L8 Biome class.
- Integrated L8Biome Data Module and included different splits.
- Included dataset details into the docs.

Here is a sample plot of an image and its mask from the dataset: 

L8 Biome


adamjstewart:
> Used getitem and plotting functions from CDL dataset. How should these be changed for L8 Biome?

CDL is useless for `__getitem__` and `plot` since this dataset is drastically different. CDL is only useful for copying the downloading code. For `__getitem__`, LandCoverAIGeo is a better reference. For plotting, every dataset is different, there's isn't a great example, but any land cover mapping dataset should have a similar image + mask plotting setup.

> What will the cmap be for this dataset?

First of all, cmap isn't required for any dataset. But you do need a way to plot the image, and imshow + colormap is the best way to do this. Oftentimes, like in the case of CDL, this cmap is hardcoded into the GeoTIFF file, and TorchGeo will automatically read and use it without listing it in the dataset, but I don't think that's the case with your masks. In the case of this dataset, the authors purposefully chose 0, 64, 128, 192, 255 for the class values instead of 0, 1, 2, 3, 4. I'm guessing that's because these are the cmap values they recommend. Basically, we plot the image in grayscale with each class having these color values. Note that in the `__getitem__` you'll need to undo this and ensure that these classes are 0, 1, 2, 3, 4.

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,190 @@
...
+        255: (255,255,255),
+    }
+    
+    def __init__(
+        self, 
+        root: str = "data",
+        crs: Optional[CRS] = None,
+        res: Optional[float] = None,
+        transforms: Optional[Callable[[Dict[str, Any]], Dict[str, Any]]] = None, 
+        cache: bool = True,
```
adamjstewart, line 76: Need to add download and checksum parameters

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,190 @@
...
+        Args:
+            query: (minx, maxx, miny, maxy, mint, maxt) coordinates to index
+        Returns:
+            sample of image, mask and metadata at that index
+        Raises:
+            IndexError: if query is not found in the index
+        """
+        hits = self.index.intersection(tuple(query), objects=True)
+        img_filepaths = cast(List[str], [hit.object for hit in hits])
+        mask_filepaths = [path.replace("images", "masks") for path in img_filepaths] # need to change
```
adamjstewart, line 124: Need to change

adamjstewart:
For the failing tests, see https://torchgeo.readthedocs.io/en/stable/user/contributing.html#linters

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    cmap = {
+        0: (0,0,0),
+        64: (64,64,64),
+        128: (128,128,128),
+        192: (192,192,192),
+        255: (255,255,255),
+    }
+
```
adamjstewart, line 69: You're missing filename_glob and filename_regex. Without those, this dataset won't actually locate any files. See https://torchgeo.readthedocs.io/en/stable/tutorials/custom_raster_dataset.html.

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,190 @@
...
+class L8Biome(RasterDataset):
+    r"""L8 Biome datasets.
+
+    The L8 Biome dataset `<https://landsat.usgs.gov/landsat-8-cloud-cover-assessment-validation-data>`__ (Landsat 8 Cloud Cover Assessment Validation Data) 
+    is a cloud validation dataset of Pre-Collection Landsat 8 Operational Land Imager (OLI) Thermal Infrared Sensor (TIRS) terrain-corrected (Level-1T) scenes.
+    
+    Dataset features:
+
+    * images evenly divided between eight unique biomes
+    * three cloud cover categories
```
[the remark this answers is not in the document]
shradhasehgal: I believe the cloud cover categories are just 3. 
Are you referring to 5 for the interpretation of bits in the manual mask?
Screen Shot 2023-03-28 at 4 06 37 PM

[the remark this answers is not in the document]
adamjstewart: I think the important thing here is the number of semantic segmentation classes, not the number of those classes containing the word "cloud". It's a 5 class problem, not a 3 class problem.

shradhasehgal:
Latest modifications:

- Fixed L8Biome dataset verification and download
- Resolved comments from previous review
- Changed auxiliary files for dataset addition
- Created L8Biome datamodule

Doubts:
- Used getitem function from LandCoverAI as per [this comment](https://github.com/microsoft/torchgeo/pull/1200#issuecomment-1486112643). Wondering if anything needs to change for L8Biome in particular?
- Altered cmap values to 0-4 as per [this discussion](https://github.com/microsoft/torchgeo/pull/1200#issuecomment-1486112643). What do the values in this dictionary represent and how to obtain them for L8 Biome?
- Some tests are still failing after running linters. How to resolve them?

TODOs:
- Write tests
- Plotting function

Tagging @adamjstewart for review.

docs/api/datasets.rst
```diff
@@ -113,6 +113,7 @@ Landsat
 .. autoclass:: Landsat3
 .. autoclass:: Landsat2
 .. autoclass:: Landsat1
+.. autoclass:: L8Biome
```
adamjstewart, line 116: Can you make a separate L8 Biome section for this dataset? Although it contains Landsat images, it's very different from these uncurated raw data loaders.

docs/api/geo_datasets.csv
```diff
@@ -14,6 +14,7 @@ Dataset,Type,Source,Size (px),Resolution (m)
 `iNaturalist`_,Points,Citizen Scientists,-,-
 `LandCover.ai Geo`_,"Imagery, Masks",Aerial,"4,200--9,500",0.25--0.5
 `Landsat`_,Imagery,Landsat,"8,900x8,900",30
+`L8Biome`_,"Imagery, Masks",Landsat,"8,900x8,900",30
```
adamjstewart, line 17:
```suggestion
`L8Biome`_,"Imagery, Masks",Landsat,"8,900x8,900","15, 30"
```
Technically there are both 15 m and 30 m resolution bands. We should probably be more specific in the other datasets too, but we can do that in a different PR.

torchgeo/datamodules/__init__.py
```diff
@@ -46,6 +47,7 @@
     "GID15DataModule",
     "InriaAerialImageLabelingDataModule",
     "LandCoverAIDataModule",
+    "L8BiomeDataModule",
```
adamjstewart, line 50: Technically L8 sorts higher than LandCover or Landsat, so we should reorder these in all files.

torchgeo/datamodules/l8biome.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class L8BiomeDataModule(GeoDataModule):
+    """LightningDataModule implementation for the L8 Biome dataset.
+    
+    .. versionadded:: 0.5
+    """
+
+    mean = torch.tensor(0)
+    std = torch.tensor(10000)
```
adamjstewart, line 21: This is fine for now, although we'll likely replace this with the SSL4EO-L8 mean and std dev once we finish downloading that dataset.

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,225 @@
...
+    The `L8 Biome <https://landsat.usgs.gov/landsat-8-cloud-cover-assessment-validation-data>`__ dataset
+    is a cloud validation dataset of Pre-Collection Landsat 8 Operational Land Imager (OLI) Thermal Infrared Sensor (TIRS) terrain-corrected (Level-1T) scenes.
+
+    Dataset features:
+    * images evenly divided between eight unique biomes
+    * 5 cloud cover categories
+
+    Dataset format:
+
+    * Each cloud mask is in ENVI binary format.
```
adamjstewart, line 28: I would put a newline between these two sentences

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,225 @@
...
+    If you use this dataset in your research, please cite the following:
+
+    * https://doi.org/10.5066/F7251GDH
+    * https://doi.org/10.1016/j.rse.2017.03.026
+
+    .. versionadded:: 0.5
+    """
+
+    url = "https://huggingface.co/datasets/torchgeo/l8biome/blob/main/{}.tar.gz"
+    # redistributed from https://landsat.usgs.gov/landsat-8-cloud-cover-assessment-validation-data
```
adamjstewart, line 40: This is documented on Hugging Face, and the original homepage is already in the docstring, so I would probably remove this comment.

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,225 @@
...
+    cmap = {
+        0: (0, 0, 0, 0),
+        1: (97, 74, 74, 255),
+        2: (38, 115, 0, 255),
+        3: (0, 197, 255, 255),
+        4: (207, 207, 207, 255),
+    }
+
+    filename_glob = "LC*_B2.TIF"
+    targz_file_glob = "*.tar.gz"
```
adamjstewart, line 62: I don't think you need this as a separate variable, it isn't specific to this dataset

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,225 @@
...
+            query: (minx, maxx, miny, maxy, mint, maxt) coordinates to index
+        Returns:
+            sample of image, mask and metadata at that index
+
+        Raises:
+            IndexError: if query is not found in the index
+        """
+        hits = self.index.intersection(tuple(query), objects=True)
+        img_filepaths = cast(List[str], [hit.object for hit in hits])
+        mask_filepaths = [path.replace("images", "masks") for path in img_filepaths]
```
adamjstewart, line 157: This still needs to be updated to convert the image filename to the mask filename.

adamjstewart (commented):
This is very close to being done (aside from plotting and testing)!

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,225 @@
...
+        "wetlands": "bff0d51db84e26a2a8e776c83ab2d331",
+    }
+
+    cmap = {
+        0: (0, 0, 0, 0),
+        1: (97, 74, 74, 255),
+        2: (38, 115, 0, 255),
+        3: (0, 197, 255, 255),
+        4: (207, 207, 207, 255),
+    }
```
adamjstewart, line 59: Came up with a great idea while reviewing the L7 Irish PR. In `__getitem__`, we convert [0, 64, 128, 192, 255] to [0, 1, 2, 3, 4]. During plotting, instead of using a complicated cmap conversion, all we need to do is divide the mask by 4. Since the pixel values will be in the 0–1 range, we can just let cmap plot the image with the builtin grayscale cmap. We don't need to create our own cmap.

tests/data/l8biome/data.py
```diff
@@ -0,0 +1,106 @@
...
+    }
+}
+
+for band in bands:
+    filenames["barren"][barren_prefix].append(f'{barren_prefix}_{band}')
+    filenames["forest"][forest_prefix].append(f'{forest_prefix}_{band}')
+
+ROOT = f"../../../../BC/{barren_prefix}"
+FILENAME = f"{barren_prefix}_B1.TIF"
+src = rasterio.open(os.path.join(ROOT, FILENAME))
```
adamjstewart, line 56: If I'm understanding correctly, this relies on real data existing on the system. We should instead hardcode the driver/dtype/src/transform in `data.py`. Someday, I'm planning on moving all of this fake data generation into the dataset fixtures so we don't have to store any fake files in the repo.

tests/datasets/test_l8biome.py
```diff
@@ -0,0 +1,70 @@
...
+
+from torchgeo.datasets import BoundingBox, IntersectionDataset, L8Biome, UnionDataset
+
+class TestL8Biome:
+    @pytest.fixture(
+        params=[
+            ["SR_B1", "SR_B2", "SR_B3", "SR_B4", "SR_B5", "SR_B6", "SR_B7"],
+            ["SR_B4", "SR_B3", "SR_B2", "SR_QA_AEROSOL"],
+        ]
+    )
```
adamjstewart, line 23: Your dataset does not have a bands parameter so there's no reason to parametrize over it

tests/datasets/test_l8biome.py
```diff
@@ -0,0 +1,70 @@
...
+    def dataset(self, monkeypatch: MonkeyPatch, request: SubRequest) -> L8Biome:
+        root = os.path.join("tests", "data", "l8biome")
+        bands = request.param
+        transforms = nn.Identity()
+        return L8Biome(root, bands=bands, transforms=transforms)
+
+    def test_separate_files(self, dataset: L8Biome) -> None:
+        assert dataset.index.count(dataset.index.bounds) == 1
+
```
adamjstewart, line 32: This is somewhat specific to the Landsat dataset, you don't have to copy it for your L8 Biome dataset.

tests/datasets/test_l8biome.py
```diff
@@ -0,0 +1,70 @@
...
+        ds = dataset | dataset
+        assert isinstance(ds, UnionDataset)
+
+    def test_plot(self, dataset: L8Biome) -> None:
+        x = dataset[dataset.bounds]
+        dataset.plot(x, suptitle="Test")
+        plt.close()
+
+    def test_plot_wrong_bands(self, dataset: L8Biome) -> None:
+        bands = ("SR_B1",)
```
adamjstewart, line 53: No bands parameter, can remove this test

adamjstewart (changes requested):

torchgeo/datamodules/__init__.py
```diff
@@ -45,6 +46,7 @@
     "FAIR1MDataModule",
     "GID15DataModule",
     "InriaAerialImageLabelingDataModule",
+    "L8BiomeDataModule",
```
adamjstewart, line 49: You should move this up to the GeoDataset section

torchgeo/datamodules/l8biome.py
```diff
@@ -0,0 +1,42 @@
...
+        """
+        super().__init__(L8Biome, batch_size, num_workers, **kwargs)
+
+    def setup(self, stage: str) -> None:
+        """Set up datasets.
+
+        Args:
+            stage: Either 'fit', 'validate', 'test', or 'predict'.
+        """
+        self.dataset = L8Biome(**self.kwargs)
```
adamjstewart, line 42: We'll need to decide on official train/val/test splits before we start training

torchgeo/datamodules/l8biome.py
```diff
@@ -0,0 +1,42 @@
...
+    def __init__(
+        self, batch_size: int = 64, num_workers: int = 0, **kwargs: Any
+    ) -> None:
+        """Initialize a new L8BiomeDataModule instance.
+
+        Args:
+            batch_size: Size of each mini-batch.
+            num_workers: Number of workers for parallel data loading.
+            **kwargs: Additional keyword arguments passed to
+                :class:`~torchgeo.datasets.SSL4EOS12`.
```
adamjstewart, line 32:
```suggestion
                :class:`~torchgeo.datasets.L8Biome`.
```

torchgeo/datasets/__init__.py
```diff
@@ -150,6 +151,7 @@
     "Landsat7",
     "Landsat8",
     "Landsat9",
+    "L8Biome",
```
adamjstewart, line 154: L8 comes before Landsat and LandCover

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,237 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""L8Biome dataset."""
```
adamjstewart, line 4:
```suggestion
"""L8 Biome dataset."""
```

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,237 @@
...
+from matplotlib.colors import ListedColormap
+from PIL import Image
+from rasterio.crs import CRS
+from torch import Tensor
+from torch.utils.data import Dataset
+import matplotlib.pyplot as plt
+import numpy as np
+
+class L8Biome(RasterDataset):
+    r"""L8 Biome datasets.
```
adamjstewart, line 24:
```suggestion
    """L8 Biome datasets.
```
I don't think you're using any escapes anywhere

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,237 @@
...
+
+    The `L8 Biome <https://landsat.usgs.gov/landsat-8-cloud-cover-assessment-validation-data>`__ dataset
+    is a cloud validation dataset of Pre-Collection Landsat 8 Operational Land Imager (OLI) Thermal Infrared Sensor (TIRS) terrain-corrected (Level-1T) scenes.
+
+    Dataset features:
+    * images evenly divided between eight unique biomes
+    * 5 cloud cover categories
+
+    Dataset format:
+
```
adamjstewart, line 34: Too many newlines here

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,237 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+
+class L8Biome(RasterDataset):
+    r"""L8 Biome datasets.
+
+    The `L8 Biome <https://landsat.usgs.gov/landsat-8-cloud-cover-assessment-validation-data>`__ dataset
+    is a cloud validation dataset of Pre-Collection Landsat 8 Operational Land Imager (OLI) Thermal Infrared Sensor (TIRS) terrain-corrected (Level-1T) scenes.
+
+    Dataset features:
```
adamjstewart, line 29: Need a newline here

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,237 @@
...
+    * images evenly divided between eight unique biomes
+    * 5 cloud cover categories
+
+    Dataset format:
```
adamjstewart, line 33: It would be worth describing the mask classes

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,237 @@
...
+    cmap = {
+        0: (0, 0, 0, 0),
+        1: (97, 74, 74, 255),
+        2: (38, 115, 0, 255),
+        3: (0, 197, 255, 255),
+        4: (207, 207, 207, 255),
+    }
+
+    filename_glob = "LC*_B2.TIF"
+    targz_file_extension = "*.tar.gz"
```
adamjstewart, line 69: No need to make this a variable, I would just use the string directly

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,237 @@
...
+        """Extract the dataset."""
+        pathname = os.path.join(self.root, self.targz_file_extension)
+        for tarfile in glob.iglob(pathname):
+            extract_archive(tarfile)
+
+    def __getitem__(self, query: BoundingBox) -> Dict[str, Any]:
+        """Retrieve image/mask and metadata indexed by query.
+        
+        Args:
+            query: (minx, maxx, miny, maxy, mint, maxt) coordinates to index
```
adamjstewart, line 155: Need newline

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,237 @@
...
+
+        Raises:
+            IndexError: if query is not found in the index
+        """
+        hits = self.index.intersection(tuple(query), objects=True)
+        img_filepaths = cast(List[str], [hit.object for hit in hits])
+        mask_filepaths = []
+
+        for path in img_filepaths:
+            mask_file_path = re.sub("B[1-9][0-9]*\.TIF", "fixedmask.img", path)
```
adamjstewart, line 167:
```suggestion
            mask_file_path = path.replace("B2.TIF", "fixedmask.img")
```
Glob uses B2 so this will always be B2

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,237 @@
...
+            print(mask_file_path)
```
adamjstewart, line 168: Can remove print statement if you're done debugging

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,237 @@
...
+            mask_filepaths.append(mask_file_path)
+
+        if not img_filepaths:
+            raise IndexError(
+                f"query: {query} not found in index with bounds: {self.bounds}"
+            )
+
+        img = self._merge_files(img_filepaths, query, self.band_indexes)
+        mask = self._merge_files(mask_filepaths, query, self.band_indexes)
```
adamjstewart, line 177: Still need to convert mask from 0, 64, 128, 192, 255 to 0, 1, 2, 3, 4

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,237 @@
...
+        num_panels = 2
+        showing_predictions = "prediction" in sample
+        if showing_predictions:
+            predictions = sample["prediction"].numpy()
+            num_panels += 1
+
+        fig, axs = plt.subplots(1, num_panels, figsize=(num_panels * 4, 5))
+        axs[0].imshow(image)
+        axs[0].axis("off")
+        axs[1].imshow(mask, vmin=0, vmax=4, cmap=self.cmap, interpolation="none")
```
adamjstewart, line 221:
```suggestion
        axs[1].imshow(mask, vmin=0, vmax=4, cmap="gray")
```

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,237 @@
...
+        "forest": "21505d878abac830890ea84abddc3c46",
+        "grass_crops": "33d0c553357f5a439aa85a45916ac89a",
+        "shrubland": "f19afc6dfa818ee3868e7040441d4c6d",
+        "snow_ice": "d7b56084e6267ee114419efdc7f664c9",
+        "urban": "b5f6aabbb380e108c408a8ea5dae3835",
+        "water": "d143049ef64e6e681cea380dd84680e9",
+        "wetlands": "bff0d51db84e26a2a8e776c83ab2d331",
+    }
+
+    cmap = {
```
adamjstewart, line 60: Delete this

adamjstewart (changes requested):

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,265 @@
...
+        "snow_ice": "d7b56084e6267ee114419efdc7f664c9",
+        "urban": "b5f6aabbb380e108c408a8ea5dae3835",
+        "water": "d143049ef64e6e681cea380dd84680e9",
+        "wetlands": "bff0d51db84e26a2a8e776c83ab2d331",
+    }
+
+    filename_glob = "LC*_B2.TIF"
+
+    separate_files = True
+    rgb_bands = ["B2", "B3", "B4"]
```
adamjstewart, line 67: These are BGR, not RGB. See https://developers.google.com/earth-engine/datasets/catalog/LANDSAT_LC08_C02_T1_TOA

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,265 @@
...
+        for path in img_filepaths:
+            mask_file_path = path.replace("B2.TIF", "fixedmask.img")
+            mask_filepaths.append(mask_file_path)
+
+        if not img_filepaths:
+            raise IndexError(
+                f"query: {query} not found in index with bounds: {self.bounds}"
+            )
+        
+        img = self._merge_files(img_filepaths, query, self.band_indexes)
```
adamjstewart, line 181: Remove `self.band_indexes` and that should solve your latest bug. `band_indexes` is only needed for `separate_files = False`.

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,265 @@
...
+        # print(self.band_indexes)
+        mask = self._merge_files(mask_filepaths, query, self.band_indexes)
+        mask_mapping = {0: 0, 64: 1, 128: 2, 192: 3, 255: 4}
```
adamjstewart, line 184: No need to map 0 to 0, you can leave that one as is.

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,225 @@
...
+
+    cmap = {
+        0: (0, 0, 0, 0),
+        1: (97, 74, 74, 255),
+        2: (38, 115, 0, 255),
+        3: (0, 197, 255, 255),
+        4: (207, 207, 207, 255),
+    }
+
+    filename_glob = "LC*_B2.TIF"
```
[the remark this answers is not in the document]
adamjstewart: Finally figured out the naming convention: https://gisgeography.com/landsat-file-naming-convention/

So I think you want:
```python
# https://gisgeography.com/landsat-file-naming-convention/
filename_glob = "LC8*_B2.TIF"
filename_regex = r"""
    ^LC8
    (?P<wrs_path>\d{3})
    (?P<wrs_row>\d{3})
    (?P<date>\d{7})
    (?P<gsi>[A-Z]{3})
    (?P<version>\d{2})
    \.TIF$
"""
date_format = "%Y%j"
```

docs/api/geo_datasets.csv
```diff
@@ -14,6 +14,7 @@ Dataset,Type,Source,Size (px),Resolution (m)
 `iNaturalist`_,Points,Citizen Scientists,-,-
 `LandCover.ai Geo`_,"Imagery, Masks",Aerial,"4,200--9,500",0.25--0.5
 `Landsat`_,Imagery,Landsat,"8,900x8,900",30
+`L8Biome`_,"Imagery, Masks",Landsat,"8,900x8,900","15,30"
```
adamjstewart, line 17:
```suggestion
`L8Biome`_,"Imagery, Masks",Landsat,"8,900x8,900","15, 30"
```

tests/data/l8biome/barren/LC80420082013220LGN00/LC80420082013220LGN00_B8.TIF
adamjstewart, line 1: You need to re-run `data.py` to regenerate this file with the correct resolution

tests/data/l8biome/data.py
```diff
@@ -0,0 +1,106 @@
...
+bands = [
+    "B1.TIF",
+    "B2.TIF",
+    "B3.TIF",
+    "B4.TIF",
+    "B5.TIF",
+    "B6.TIF",
+    "B7.TIF",
+    "B8.TIF",
+    "B8A.TIF",
```
adamjstewart, line 31: Original dataset doesn't contain B8A

tests/data/l8biome/data.py
```diff
@@ -0,0 +1,106 @@
...
+    "B9.TIF",
+    "B10.TIF",
+    "B11.TIF",
+    "B12.TIF",
```
adamjstewart, line 35: Original dataset doesn't contain B12

tests/data/l8biome/data.py
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    if path.endswith("B8.TIF"):
+        profile["transform"] = Affine(15.0, 0.0, 339892.5, 0.0, -15.0, 8286907.5)
+        profile["width"] = profile["height"] = SIZE * 2
+
+    Z = np.random.randn(SIZE, SIZE).astype(profile["dtype"])
+
+    if path.endswith("fixedmask.img"):
+        Z = np.random.randint(4, size=(SIZE, SIZE), dtype=dtype)
```
adamjstewart, line 73:
```suggestion
        Z = np.random.randint(np.iinfo(dtype).max, size=(SIZE, SIZE), dtype=dtype)
```
This is more realistic

tests/datasets/test_l8biome.py
```diff
@@ -0,0 +1,52 @@
...
+
+from torchgeo.datasets import BoundingBox, IntersectionDataset, L8Biome, UnionDataset
+
+
+class TestL8Biome:
+    @pytest.fixture
+    def dataset(self) -> L8Biome:
+        root = os.path.join("tests", "data", "l8biome")
+        transforms = nn.Identity()
+        print("test")
```
adamjstewart, line 21: Can remove print statements

tests/datasets/test_l8biome.py
adamjstewart, line 1: Missing tests for download behavior, see CDL tests for a good example

torchgeo/datamodules/l8biome.py
adamjstewart, line 1: Missing tests for this file, these should go in `tests/trainers/test_segmentation.py`. Should be only 2 lines of code to get this tested.

torchgeo/datamodules/l8biome.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""L8Biome datamodule."""
```
adamjstewart, line 4:
```suggestion
"""L8 Biome datamodule."""
```

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,260 @@
...
+from torch import Tensor
+
+from .geo import RasterDataset
+from .utils import BoundingBox, download_url, extract_archive
+
+
+class L8Biome(RasterDataset):
+    """L8 Biome datasets.
+
+    The `L8 Biome <https://landsat.usgs.gov/landsat-8-cloud-cover-assessment-validation-data>`__ dataset # noqa: E501
```
adamjstewart, line 22:
```suggestion
    The `L8 Biome <https://landsat.usgs.gov/landsat-8-cloud-cover-assessment-validation-data>`__ dataset
```
This is included in the documentation. Comments inside of docstrings aren't comments.

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,260 @@
...
+    Dataset features:
+
+    * images evenly divided between eight unique biomes
+    * 5 cloud cover categories
+
+    Dataset format:
+
+    * Each cloud mask is in ENVI binary format.
+    * Interpretation for bits in each manual mask are as follows:
+    * 0: Fill, 64: Cloud Shadow, 128: Clear, 192: Thin Cloud, 255: Cloud
```
adamjstewart, line 36:
```suggestion

    Dataset classes:
    
    0. Fill
    1. Cloud Shadow
    2. Clear
    3. Thin Cloud
    4. Cloud
```

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,260 @@
...
+        """Verify the integrity of the dataset.
+
+        Raises:
+            RuntimeError: if ``download=False`` but dataset is missing or checksum fails
+        """
+        # Check if the extracted files already exist
+        pathname = os.path.join(self.root, "**", self.filename_glob)
+        for fname in glob.iglob(pathname, recursive=True):
+            if not fname.endswith(".tar.gz"):
+                return
```
adamjstewart, line 118:
```suggestion
            return
```
Your glob can only possibly match `*.TIF` files, it won't match tarballs.

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,260 @@
...
+                f"query: {query} not found in index with bounds: {self.bounds}"
+            )
+
+        data_list: List[Tensor] = []
+        for band in self.bands:
+            band_filepaths = []
+            filepath = filepaths[0].replace("B2", band)
+            band_filepaths.append(filepath)
+            data_list.append(self._merge_files(band_filepaths, query))
+        img = torch.cat(data_list)
```
adamjstewart, line 176: You don't need any of this. Just call `self._merge_files`. The aforementioned `filename_regex` is required for this to work properly.

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,260 @@
...
+            "mask": mask.long(),
+        }
+
+        if self.transforms is not None:
+            sample = self.transforms(sample)
+
+        return sample
+
+    # Plotting code added as placeholder for now till I get it working.
+    # Using LandCoverAI plotting as reference.
```
adamjstewart, line 202: Can remove this comment

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,260 @@
...
+            sample: a sample returned by :meth:`__getitem__`
+            show_titles: flag indicating whether to show titles above each panel
+            suptitle: optional string to use as a suptitle
+
+        Returns:
+            a matplotlib Figure with the rendered sample
+        """
+        rgb_indices = []
+        for band in self.rgb_bands:
+            # print(band, self.bands)
```
adamjstewart, line 222: Can remove commented out code

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,260 @@
...
+            if band in self.bands:
+                rgb_indices.append(self.bands.index(band))
+            else:
+                raise ValueError("Dataset doesn't contain some of the RGB bands")
+
+        image = sample["image"][rgb_indices].permute(1, 2, 0)
+        print(image.shape)
+        image = torch.clamp(image / 50000, min=0, max=1).numpy()
+
+        # image = sample["image"].numpy().astype("uint16").squeeze()
```
adamjstewart, line 232: Can remove commented out code

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,260 @@
...
+        rgb_indices = []
+        for band in self.rgb_bands:
+            # print(band, self.bands)
+            if band in self.bands:
+                rgb_indices.append(self.bands.index(band))
+            else:
+                raise ValueError("Dataset doesn't contain some of the RGB bands")
+
+        image = sample["image"][rgb_indices].permute(1, 2, 0)
+        print(image.shape)
```
adamjstewart, line 229: Can remove print statement

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,260 @@
...
+        image = torch.clamp(image / 50000, min=0, max=1).numpy()
```
adamjstewart, line 230: 50K seems high... Maybe just copy the scaling used in landsat.py

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,260 @@
...
+        axs[0].axis("off")
+        axs[1].imshow(mask, vmin=0, vmax=4, cmap="gray")
+        axs[1].axis("off")
+        if show_titles:
+            axs[0].set_title("Image")
+            axs[1].set_title("Mask")
+
+        if showing_predictions:
+            axs[2].imshow(
+                predictions, vmin=0, vmax=4, cmap="gray", interpolation="none"
```
adamjstewart, line 252:
```suggestion
                predictions, vmin=0, vmax=4, cmap="gray"
```

adamjstewart (changes requested):

torchgeo/datamodules/l8biome.py
adamjstewart: Let's just do a random 60/20/20 train/val/test split. Add this import up top:
```python
from ..datasets import random_bbox_assignment
```
Then change this to:
```suggestion
        dataset = L8Biome(**self.kwargs)
        generator = torch.Generator().manual_seed(0)
        self.train_dataset, self.val_dataset, self.test_dataset = random_bbox_assignment(
            dataset, [0.6, 0.2, 0.2], generator
        )
```

torchgeo/datasets/l8biome.py
shradhasehgal: This was to combat the warning generated by flake8 (E501 line too long (104 > 88 characters)). Should we still remove it?

adamjstewart: You should add it to the closing `"""` line. You can verify that it is currently being erroneously included in the docs by viewing https://torchgeo--1200.org.readthedocs.build/en/1200/api/datasets.html#l8biome.

docs/api/datasets.rst
```diff
@@ -93,6 +93,11 @@ iNaturalist
 
 .. autoclass:: INaturalist
 
+L8Biome
+^^^^^^^
```
adamjstewart, line 97:
```suggestion
L8 Biome
^^^^^^^^
```

docs/api/datamodules.rst
```diff
@@ -11,6 +11,11 @@ Chesapeake Land Cover
 
 .. autoclass:: ChesapeakeCVPRDataModule
 
+L8Biome
+^^^^^^^
```
adamjstewart, line 15:
```suggestion
L8 Biome
^^^^^^^^
```

tests/data/l8biome/data.py
adamjstewart, line 1: Can you make this file executable? `chmod +x data.py`.

tests/data/l8biome/barren/LC80420082013220LGN00/shs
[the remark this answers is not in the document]
shradhasehgal: Not required. Will remove with next commit.

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,260 @@
...
+
+        if not filepaths:
+            raise IndexError(
+                f"query: {query} not found in index with bounds: {self.bounds}"
+            )
+
+        data_list: List[Tensor] = []
+        for band in self.bands:
+            band_filepaths = []
+            filepath = filepaths[0].replace("B2", band)
```
adamjstewart, line 173: You'll want to use the regex here instead of string replacement. For example, what happens when the dataset is in a directory with "B2" in the name?

torchgeo/datamodules/l8biome.py
```diff
@@ -0,0 +1,48 @@
...
+        Args:
+            stage: Either 'fit', 'validate', 'test', or 'predict'.
+        """
+        dataset = L8Biome(**self.kwargs)
+        generator = torch.Generator().manual_seed(0)
+        (
+            self.train_dataset,
+            self.val_dataset,
+            self.test_dataset,
+        ) = random_bbox_assignment(dataset, [0.6, 0.2, 0.2], generator)
```
adamjstewart, line 48: Oh, I think we also need to define a train/val/test_sampler. See the base class, that should give you an idea of what it needs to look like. We'll also need to add batch_size, patch_size, and length parameters to init for this.

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,278 @@
...
+        "forest": "21505d878abac830890ea84abddc3c46",
+        "grass_crops": "33d0c553357f5a439aa85a45916ac89a",
+        "shrubland": "f19afc6dfa818ee3868e7040441d4c6d",
+        "snow_ice": "d7b56084e6267ee114419efdc7f664c9",
+        "urban": "b5f6aabbb380e108c408a8ea5dae3835",
+        "water": "d143049ef64e6e681cea380dd84680e9",
+        "wetlands": "bff0d51db84e26a2a8e776c83ab2d331",
+    }
+
+    filename_glob = "LC*_B2.TIF"
```
adamjstewart, line 66:
```suggestion
    # https://gisgeography.com/landsat-file-naming-convention/
    filename_glob = "LC8*_B2.TIF"
```

tests/conf/l8biome.yaml
```diff
@@ -0,0 +1,19 @@
...
+  module:
+    loss: "ce"
+    model: "unet"
+    backbone: "resnet18"
+    weights: null
+    learning_rate: 1e-3
+    learning_rate_schedule_patience: 6
+    verbose: false
+    in_channels: 3
+    num_classes: 6
```
adamjstewart, line 12:
```suggestion
    in_channels: 11
    num_classes: 5
```

torchgeo/datamodules/l8biome.py
```diff
@@ -0,0 +1,62 @@
...
+    """LightningDataModule implementation for the L8 Biome dataset.
+
+    .. versionadded:: 0.5
+    """
+
+    mean = torch.tensor(0)
+    std = torch.tensor(10000)
+
+    def __init__(
+        self, batch_size: int = 64, num_workers: int = 0, **kwargs: Any
```
adamjstewart, line 25: Need to add batch_size, patch_size, and length parameters, same as the base class. You'll also want to set these in the yaml file to make them smaller for testing purposes

tests/data/l8biome/data.py
```diff
@@ -0,0 +1,106 @@
...
+
+    with rasterio.open(path, "w", **profile) as src:
+        src.write(Z, 1)
+
+
+def create_directory(directory: str, hierarchy: FILENAME_HIERARCHY) -> None:
+    if isinstance(hierarchy, dict):
+        # Recursive case
+        for key, value in hierarchy.items():
+            print(directory, key)
```
adamjstewart, line 82: Don't forget to remove this print statement

tests/datasets/test_l8biome.py
```diff
@@ -0,0 +1,94 @@
...
+        monkeypatch.setattr(L8Biome, "filenames_to_md5", filenames_to_md5)
+        root = str(tmp_path)
+        transforms = nn.Identity()
+        return L8Biome(root, transforms=transforms, download=True, checksum=True)
+
+    def test_getitem(self, dataset: L8Biome) -> None:
+        x = dataset[dataset.bounds]
+        assert isinstance(x, dict)
+        assert isinstance(x["crs"], CRS)
+        assert isinstance(x["image"], torch.Tensor)
```
adamjstewart, line 44: Should also test that a mask is returned

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,280 @@
...
+import torch
+from rasterio.crs import CRS
+from torch import Tensor
+
+from .geo import RasterDataset
+from .utils import BoundingBox, download_url, extract_archive
+
+
+class L8Biome(RasterDataset):
+    """L8 Biome datasets.
```
adamjstewart, line 21:
```suggestion
    """L8 Biome dataset.
```

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,280 @@
...
+    terrain-corrected (Level-1T) scenes.
+
+    Dataset features:
+
+    * images evenly divided between eight unique biomes
+    * 5 cloud cover categories
+
+    Dataset format:
+
+    * Each cloud mask is in ENVI binary format.
```
adamjstewart, line 35: Can you also add a bullet point for the file format of the images?

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,280 @@
...
+
+    If you use this dataset in your research, please cite the following:
+
+    * https://doi.org/10.5066/F7251GDH
+    * https://doi.org/10.1016/j.rse.2017.03.026
+
+    .. versionadded:: 0.5
+    """  # noqa: E501
+
+    url = "https://huggingface.co/datasets/torchgeo/l8biome/blob/main/{}.tar.gz"
```
adamjstewart, line 53:
```suggestion
    url = "https://huggingface.co/datasets/torchgeo/l8biome/resolve/main/{}.tar.gz"  # noqa: E501
```
Previous code downloaded HTML instead of the raw file

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,280 @@
...
+    separate_files = True
+    rgb_bands = ["B4", "B2", "B3"]
+    all_bands = ["B1", "B2", "B3", "B4", "B5", "B6", "B7", "B8", "B9", "B10", "B11"]
+
+    def __init__(
+        self,
+        root: str = "data",
+        crs: Optional[CRS] = None,
+        res: Optional[float] = None,
+        transforms: Optional[Callable[[Dict[str, Any]], Dict[str, Any]]] = None,
```
adamjstewart, line 89: You should try to keep these in the same order as the base class

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,280 @@
...
+        cache: bool = True,
+        download: bool = False,
+        checksum: bool = False,
+    ) -> None:
+        """Initialize a new L8Biome dataset instance.
+
+        Args:
+
+            root: root directory where dataset can be found
+            transforms: a function/transform that takes input sample and its target as
```
adamjstewart, line 100: transforms is duplicated, crs is missing. Try to keep these in the same order as the function signature

adamjstewart (changes requested):
Hopefully my last review!

Docs look good: https://torchgeo--1200.org.readthedocs.build/en/1200/api/datasets.html#l8-biome

Can you plot an image and add the plot to the PR description so we can make sure everything looks visually correct? Colors might be better now that it isn't only B2 data.

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,280 @@
...
+
+    If you use this dataset in your research, please cite the following:
+
+    * https://doi.org/10.5066/F7251GDH
+    * https://doi.org/10.1016/j.rse.2017.03.026
+
+    .. versionadded:: 0.5
+    """  # noqa: E501
+
+    url = "https://huggingface.co/datasets/torchgeo/l8biome/blob/main/{}.tar.gz"  # noqa: E501
```
adamjstewart, line 54:
```suggestion
    url = "https://huggingface.co/datasets/torchgeo/l8biome/resolve/main/{}.tar.gz"  # noqa: E501
```
You're still downloading the HTML file instead of the raw data.

P.S. If you click "commit changes" you don't have to worry about copying the changes correctly, it will add my suggested change directly to the PR.

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,280 @@
...
+        "grass_crops": "33d0c553357f5a439aa85a45916ac89a",
+        "shrubland": "f19afc6dfa818ee3868e7040441d4c6d",
+        "snow_ice": "d7b56084e6267ee114419efdc7f664c9",
+        "urban": "b5f6aabbb380e108c408a8ea5dae3835",
+        "water": "d143049ef64e6e681cea380dd84680e9",
+        "wetlands": "bff0d51db84e26a2a8e776c83ab2d331",
+    }
+
+    # https://gisgeography.com/landsat-file-naming-convention/
+    filename_glob = "LC8*_B2.TIF"
```
adamjstewart, line 68: Why B2 instead of B1?

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,280 @@
...
+                        end = match.end("band")
+                        filename = filename[:start] + band + filename[end:]
+                filepath = os.path.join(directory, filename)
+                band_filepaths.append(filepath)
+            image_list.append(self._merge_files(band_filepaths, query))
+        image = torch.cat(image_list)
+
+        mask_filepaths = []
+        for filepath in filepaths:
+            mask_filepath = filepath.replace("B2.TIF", "fixedmask.img")
```
adamjstewart, line 208: Also need to change here

torchgeo/datasets/l8biome.py
```diff
@@ -0,0 +1,280 @@
...
+        (?P<date>\d{7})
+        (?P<gsi>[A-Z]{3})
+        (?P<version>\d{2})
+        _(?P<band>B\d{1,2})
+        \.TIF$
+    """
+    date_format = "%Y%j"
+
+    separate_files = True
+    rgb_bands = ["B4", "B2", "B3"]
```
adamjstewart, line 82:
```suggestion
    rgb_bands = ["B4", "B3", "B2"]
```
https://developers.google.com/earth-engine/datasets/catalog/LANDSAT_LC08_C02_T1_TOA

adamjstewart: Can you remake the plot in the description too?

shradhasehgal: Updated plot

adamjstewart (approved):
Looks great, congrats on your first PR! Hopefully the first of many if the extensive review process didn't scare you away 😅

PR #1208, opened 2023-03-29
Add NTXent Loss for SSL pretraining

isaaccorley:


torchgeo/losses/qr.py
```diff
@@ -1,3 +1,6 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
isaaccorley, line 1: This file didn't have the license so I added it

torchgeo/losses/ssl.py
```diff
@@ -0,0 +1,41 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""Loss functions self-supervised learning."""
```
adamjstewart, line 4:
```suggestion
"""Loss functions for self-supervised learning."""
```

torchgeo/losses/ssl.py
```diff
@@ -0,0 +1,41 @@
...
+
+import torch
+import torch.nn.functional as F
+from torch.nn.modules import Module
```
adamjstewart, line 8:
```suggestion
from torch.nn import Module
```
I try to import things at the same level that they get imported in the docs

torchgeo/losses/ssl.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class NTXentLoss(Module):
```
adamjstewart, line 11: Should we name this NT-Xent or InfoNCE? I see both very commonly used (SimCLR calls it NT-Xent, MoCo calls it InfoNCE). I believe MoCo v1 links to a specific paper that "defines" this loss. We should add a citation for that. Regardless of which we call it, we should add an "aka" in the docstring.

torchgeo/losses/ssl.py
```diff
@@ -0,0 +1,41 @@
...
+        """
+        batch_size = z1.shape[0]
+        device = z1.device
+        z1 = F.normalize(z1, dim=-1)
+        z2 = F.normalize(z2, dim=-1)
+        similarity = torch.matmul(z1, z2.T)
+        similarity = similarity * torch.exp(t)
+        targets = torch.arange(batch_size, device=device)
+        loss = F.cross_entropy(similarity, targets)
+        return loss
```
adamjstewart, line 41: I trust that this works, but the MoCo and SimCLR implementations are so different that I want to take some time to go through the math and confirm for myself that all of these are actually equivalent. At one point I searched for other implementations (there are several) and they were all drastically different. Some are more general (support a variable number of positive pairs), faster, and more flexible. For now, as long as it works for the trainers we're planning, it doesn't need to be too fancy.

adamjstewart (commented):
This falls under the category of things that aren't really specific to geospatial data that I wish we didn't have to include in TorchGeo. But like our trainers, there isn't a great alternative library that we already depend on that we could quickly contribute this too. In the long run, we may want to try to contribute this to PyTorch directly. But that's going to be a slow process, so I think it's fine to keep this in TorchGeo until that happens.

torchgeo/losses/ssl.py
isaaccorley: That's a good point. This implementation takes the dot product and then L2 normalized each vector so it's effectively taking the cosine similarity between vectors and then passing that through a cross entropy loss function. The target in this case is the identity matrix which has the assumption that there is only 1 positive pair for each sample in a batch.

tests/losses/test_ssl.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class TestSSLLosses:
+    def test_ntxent_loss(self) -> None:
+        z1 = torch.randn(2, 8)
+        z2 = torch.randn(2, 8)
+        t = torch.tensor(0.5)
+        loss = NTXentLoss()(z1, z2, t)
+        assert loss.ndim == 0
+        assert loss.dtype == torch.float
```
adamjstewart, line 16: We should test the value of `loss` to make sure it is correct, not just that it runs without crashing. This would require using a manually defined matrix, not just a random tensor.

torchgeo/losses/ssl.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def forward(
+        self, z1: torch.Tensor, z2: torch.Tensor, t: torch.Tensor
+    ) -> torch.Tensor:
+        """Computes the NT-Xent loss.
+
+        Args:
+            z1: batch of vectors of shape (B, D).
+            z2: batch of vectors of shape (B, D).
+            t: temperature value to normalize the similarity vectors
```
adamjstewart, line 28: Should we support float temperature instead of requiring it to be a Tensor?

torchgeo/losses/ssl.py
adamjstewart: Tried comparing your implementation vs. the ones from SimCLR and MoCo and miraculously all implementations get different results! Using [this script](https://github.com/microsoft/torchgeo/files/11133316/test.txt), I get:
```
isaac: 1.6350399255752563
simclr: 2.2017433643341064
moco: 0.0
moco v3: 3.2240943908691406
```
Not sure if I copied something wrong, especially for MoCo v1 which has a separate queue that I didn't include and am not sure how to incorporate. We sure figure out why these all have different results.

torchgeo/losses/ssl.py
calebrob6: IMO yes

torchgeo/losses/ssl.py
isaaccorley: NT-Xent is different than InfoNCE. NT-Xent normalizes using a temperature value and also assumes the target is an identity matrix (pairs from the same sample are positives, all else are negatives).

adamjstewart: Gotcha. Can you implement both?

torchgeo/losses/ssl.py
isaaccorley: - the MoCo v1 loss in your script assumes the target indicator is at the 0th index not the diagonal which is why you are getting different values than all others (see `labels = torch.zeros(logits.shape[0], dtype=torch.long)` vs `targets = torch.arange(batch_size, device=device)`
- My loss is closest to MoCov3 except they multiply by 2 at the end for some reason (I think we can probably use this loss instead)
- The SimCLR loss implementation does something different with the target labels in this line `pos_mask = self_mask.roll(shifts=cos_sim.shape[0] // 2, dims=0)`. I'm not sure why this happens but it transforms the identity matrix targets to a different type of target matrix which is likely why we end up with different values. If we don't change the target mask and keep it as the identity matrix the loss ends up being `tensor(1.3000e+16)`

torchgeo/losses/ssl.py
isaaccorley: Yes but I think we should do infoNCE in a separate PR. It will take some discussion because it requires that you know where all the positive pairs are in the batch which can be N-pairs.

torchgeo/losses/ssl.py
adamjstewart: MoCo is InfoNCE and SimCLR is NT-Xent so that may also explain the difference

isaaccorley: The MoCo loss you posted normalizes by a temperature so it's basically NT-Xent and has only 1 positive pair per sample in a batch. It's just at the 0th index instead of the diagonal. There's nothing wrong with doing it this way as long as the 0th index in the similarity matrix represents the cosine similarity of the vectors from each augmented pair

tests/losses/test_ssl.py
isaaccorley: Is it sufficient to insert z1 and z2 where each are just vectors with 1s in them and compute the loss and compare to this? Or is there a specific case you would like to see?

adamjstewart: I would just choose some simple vectors (maybe [0, 1, 2] and [3, 4, 5] or something of that level of complexity), then manually compute what NT-Xent should be using a mathematical formula, then confirm that the output of your loss matches.

isaaccorley:
Closing this because we decided to use lightly's implementations instead of maintaining our own.

adamjstewart:
Should we submit a PR for InfoNCE loss to lightly? How different is it from their current implementation?

isaaccorley:
I don't think we need an infoNCE loss. Couldn't we just use lightly's NT-XentLoss with the memory bank argument set? I guess, what are you proposing that the infoNCE loss needs that the NTXent loss doesn't already provide?

adamjstewart:
Idk, that's what I'm asking. I thought they were the same thing but in this PR you suggested that they're different.

isaaccorley:
They can be implemented differently depending on how many positive pairs you have but for our use case their NTXEntLoss should be good enough.

PR #1209, opened 2023-03-29
add ssl4eo-s12 downloading script

wangyi111:


experiments/download_ssl4eo-s12.py
adamjstewart, line 1: Let's call it `download_ssl4eo.py` since we'll be modifying it to work for all satellites in follow-up PRs.

experiments/download_ssl4eo-s12.py
```diff
@@ -0,0 +1,718 @@
...
+    "B12",
+]
+RGB_BANDS = ["B4", "B3", "B2"]
+ALL_BANDS_GRD = ["VV", "VH"]
+
+
+""" samplers to get locations of interest points"""
+
+
+class GeoSampler:
```
adamjstewart, line 84: We might want to rename this to avoid confusion with `torchgeo.samplers.GeoSampler`

experiments/download_ssl4eo-s12.py
```diff
@@ -0,0 +1,718 @@
...
+
+import ee
+import numpy as np
+import rasterio
+import shapefile
+import urllib3
+from rasterio.transform import Affine
+from shapely.geometry import Point, shape
+
+# from skimage.exposure import rescale_intensity
```
adamjstewart, line 42: You can just delete commented out code

adamjstewart (commented):
To fix the remaining test failures, see https://torchgeo.readthedocs.io/en/stable/user/contributing.html#linters. The black/flake8 errors should be super easy to fix. The mypy failures are harder to fix if you aren't familiar with Python type hints. Let me know if you want help with that part.

experiments/download_ssl4eo.py
```diff
@@ -0,0 +1,738 @@
...
+        point = rng.choice(self.interest_points)
+        std = self.km2deg(self.std)
+        lon, lat = np.random.normal(loc=point, scale=[std, std])
+        return [lon, lat]
+
+    @staticmethod
+    def get_world_cities(
+        download_root: str = os.path.expanduser("./world_cities/"),
+    ) -> List[Dict[str, Any]]:
+        url = "https://simplemaps.com/static/data/world-cities/basic/simplemaps_worldcities_basicv1.71.zip"  # noqa
```
adamjstewart, line 115:
```suggestion
        url = "https://simplemaps.com/static/data/world-cities/basic/simplemaps_worldcities_basicv1.71.zip"  # noqa: E501
```
This is more specific so it only ignores "line too long" errors and nothing else

experiments/download_ssl4eo.py
```diff
@@ -0,0 +1,738 @@
...
+        points = [[float(c["lng"]), float(c["lat"])] for c in cities]
+        return points
+
+    @staticmethod
+    def km2deg(kms: float, radius: float = 6371) -> float:
+        return kms / (2.0 * radius * np.pi / 360.0)
+
+    @staticmethod
+    def deg2km(deg: float, radius: float = 6371) -> float:
+        return deg * (2.0 * radius * np.pi / 360.0)
```
adamjstewart, line 143: Aren't these only true at the equator? We might want to fix this in a follow-up PR.

experiments/download_ssl4eo.py
```diff
@@ -0,0 +1,738 @@
...
+    def sample_point(self) -> List[float]:
+        rng = np.random.default_rng()
+        point = rng.choice(self.interest_points)
+        std = self.km2deg(self.std)
+        lon, lat = np.random.normal(loc=point, scale=[std, std])
+        return [lon, lat]
+
+    @staticmethod
+    def get_world_cities(
+        download_root: str = os.path.expanduser("./world_cities/"),
```
adamjstewart, line 113:
```suggestion
        download_root: str = "world_cities",
```

experiments/download_ssl4eo.py
```diff
@@ -0,0 +1,738 @@
...
+        if self.boundaries.contains(p):
+            return [p.x, p.y]
+        else:
+            return self.sample_point()
+
+    @staticmethod
+    def get_country_boundaries(
+        download_root: str = os.path.expanduser("~/.cache/naturalearth"),
+    ) -> shape:
+        url = "https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/cultural/ne_110m_admin_0_countries.zip"  # noqa
```
adamjstewart, line 167:
```suggestion
        url = "https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/cultural/ne_110m_admin_0_countries.zip"  # noqa: E501
```

experiments/download_ssl4eo.py
```diff
@@ -0,0 +1,738 @@
...
+    # Both flags should be set to zero, indicating clear conditions.
+    mask = qa.bitwiseAnd(cloudBitMask).eq(0)
+    mask = mask.bitwiseAnd(cirrusBitMask).eq(0)
+    return image.updateMask(mask)
+
+
+def get_collection_s2a(cloud_pct: float = 20) -> ee.ImageCollection:
+    collection = ee.ImageCollection("COPERNICUS/S2_SR")
+    collection = collection.filter(ee.Filter.lt("CLOUDY_PIXEL_PERCENTAGE", cloud_pct))
+    # collection = collection.map(maskS2clouds)
```
adamjstewart, line 206: You can remove commented out code

experiments/download_ssl4eo.py
```diff
@@ -0,0 +1,738 @@
...
+        ee.Filter.listContains("transmitterReceiverPolarisation", "VV")
+    ).filter(ee.Filter.listContains("transmitterReceiverPolarisation", "VH"))
+    filtered = filtered.filter(ee.Filter.eq("instrumentMode", "IW"))
+    # filtered = filtered.filter(ee.Filter.eq('orbitProperties_pass', 'ASCENDING'))
+    # filtered = filtered.filter(ee.Filter.eq('orbitProperties_pass', 'DESCENDING'))
+
+    if filtered.size().getInfo() == 0:
+        raise ee.EEException(
+            f"ImageCollection.filter: No suitable images found in \
+            ({coords[1]:.4f}, {coords[0]:.4f}) between {period[0]} and {period[1]}."
```
adamjstewart, line 256:
```suggestion
            "ImageCollection.filter: No suitable images found in "
            f"({coords[1]:.4f}, {coords[0]:.4f}) between {period[0]} and {period[1]}."
```
Otherwise you get a ton of extra spaces in the middle of your error msg

experiments/download_ssl4eo.py
```diff
@@ -0,0 +1,738 @@
...
+        )
+    return filtered
+
+
+def center_crop(
+    img: np.ndarray[Any, np.dtype[Any]], out_size: Tuple[int, int]
+) -> np.ndarray[Any, np.dtype[Any]]:
+    image_height, image_width = img.shape[:2]
+    crop_height, crop_width = out_size
+    crop_top = int((image_height - crop_height + 1) * 0.5)
```
adamjstewart, line 266: Could simplify to `(image_height - crop_height + 1) // 2`

experiments/download_ssl4eo.py
```diff
@@ -0,0 +1,738 @@
...
+    return [
+        [coords[0][0] + (xoff * xres), coords[0][1] - (yoff * yres)],
+        [
+            coords[0][0] + ((xoff + new_size[1]) * xres),
+            coords[0][1] - ((yoff + new_size[0]) * yres),
+        ],
+    ]
+
+
+def get_properties(image: ee.Image) -> Any:
```
adamjstewart, line 287: Any idea what the return type actually is? Any should be avoided whenever possible since it disables type checking

adamjstewart (commented):
Still lots of commented-out code we can remove

experiments/download_ssl4eo.py
wangyi111: Yep, this was extremely simplified.

experiments/download_ssl4eo.py
wangyi111: this one I thought it's good to keep as people also use this cloud masking commonly, I commented out for sake of time, but to be discussed.

experiments/download_ssl4eo.py
wangyi111: It should be a dict, but if I modify that I got the error: "Returning Any from function declared to return "Dict[str, Any]""

adamjstewart: Ah, yes, if ee is missing type hints then it won't know what the real type should be. Correct way to handle this is to cast to the proper type, but that's a lot of work and this code isn't going to be installed in the core library so Any is fine for now.

adamjstewart (approved):
I think this is good enough for now. Next steps (preferably each in separate sequential PRs):

- [ ] Split "sampling strategies" and "check availability and download" into separate scripts
- [ ] Replace gridded overlap check with R-tree
- [ ] Remove Sentinel-specific stuff and make the script more general
- [ ] Improve km <-> degree conversion

experiments/download_ssl4eo.py
adamjstewart: Actually, now that I think about it, the patches we download are in WGS 84 (lat, long), so the conversion is correct if we want to ensure that all patches have the same size.

Is there a reason WGS 84 was chosen? Is that the default projection of the data stored in GEE? It will result in a lot of distortion in higher latitudes, although there aren't many cities at higher latitudes.

wangyi111: GEE does have a default projection WGS 84, but you can export to other projections. The main reason is still to make it simple, as the distortion can be kind of neglected (considering the pixel resolution).

PR #1213, opened 2023-03-31
separate sampling and use rtree based overlap check for ssl4eo-s12 downloading

wangyi111:


experiments/download_ssl4eo.py
```diff
@@ -534,10 +622,12 @@ def fix_random_seeds(seed: int = 42) -> None:
     parser.add_argument("--cloud_pct", type=int, default=20)
     parser.add_argument("--num_workers", type=int, default=8)
     parser.add_argument("--log_freq", type=int, default=10)  # print frequency
-    parser.add_argument("--preview", action="store_true")
     parser.add_argument(
         "--resume", type=str, default=None
     )  # resume from existing coordinates
+    parser.add_argument(
+        "--overlap_check", type=str, default="rtree", choices=["grid", "rtree", None]
```
adamjstewart, line 629: Is there any reason we would want to support both? Seems like the rtree implementation is more technically correct and superior in every way. This file is already gigantic, so we should try to minimize how many lines of code we have to maintain. I would vote for deleting the old logic.

wangyi111: Wasn't very sure, as it is the original implementation and in terms of speed grid is a bit faster. But yes it's not very necessary here, let me delete it.

experiments/download_ssl4eo.py
```diff
@@ -396,23 +393,20 @@ def get_random_patches_grid(
...
-                if distance < (1.5 * radius / 1000):
-                    raise OverlapError
-            grid_dict[gridIndex].add(new_coord)
+        for i in rtree_obj.nearest(new_coord, num_results=1, objects=True):
+            distance = np.sqrt(
+                sampler.deg2km(abs(new_coord[0] - i.bbox[2])) ** 2
+                + sampler.deg2km(abs(new_coord[1] - i.bbox[3])) ** 2
+            )
+            if distance < (1.5 * radius / 1000):
+                raise OverlapError
```
adamjstewart, line 402: There's a much easier way to do this:
```pycon
>>> from rtree.index import Index
>>> idx = Index()
>>> idx.insert(0, (0, 0, 1, 1))
>>> list(idx.intersection((0, 0, 1, 1)))
[0]
>>> list(idx.intersection((2, 2, 3, 3)))
[]
```
Just check if the new bbox intersects with any existing bboxes in the R-tree. R-trees are designed to store boxes, not points (although you can also store points). The above example requires you to store the bounding box of the patch you want to store.

experiments/download_ssl4eo.py
```diff
@@ -396,23 +393,20 @@ def get_random_patches_grid(
...
+        rtree_obj.insert(
+            len(rtree_obj) - 1, (new_coord[0], new_coord[1], new_coord[0], new_coord[1])
```
adamjstewart, line 404: This computes the size of the R-tree every single time, which is probably slow. We can either use a counter variable or just use 0 for all indices (we aren't using the index number anyway).

experiments/download_ssl4eo.py
```diff
@@ -396,23 +393,20 @@ def get_random_patches_grid(
...
+        )
 
     except OverlapError:
-        patches_s1, patches_s2c, patches_s2a, center_coord = get_random_patches_grid(
-            idx, collections, bands, crops, sampler, dates, radius, debug
+        patches_s1, patches_s2c, patches_s2a, center_coord = get_random_patches_rtree(
```
adamjstewart, line 408: Is recursion necessary here? It could possibly end up with a very deep recursion stack. I would personally use a `while True` loop and break out when you find a valid location. Then you don't need to raise and catch errors, should be much simpler.

experiments/download_ssl4eo.py
```diff
@@ -377,7 +374,7 @@ def get_patch_s2(
     )
 
 
-def get_random_patches_grid(
+def get_random_patches_rtree(
```
adamjstewart, line 377: How many times does this function get called? We want to build a single R-tree containing all `--indices_range` non-overlapping locations. We don't want to create a new R-tree every time this function gets called.

adamjstewart (commented):
I'm still trying to understand the current implementation. It might help to write the pseudocode for how it works. I think we can get away with something much simpler than we currently have:
```
idx = Index()
counter = 0
while counter < total_num_patches:
    bbox = choose_random_bbox()
    if list(idx.intersection(bbox)):
        continue
    idx.insert(counter, bbox)
    counter += 1
return idx
```
This creates an R-tree containing total_num_patches non-overlapping locations. A separate (parallelized) script would actually check how many of them have 4 cloud-free images and then download them.

experiments/download_ssl4eo.py
wangyi111: No, didn't think much about the difference. I thought recursion and loop work the same here?

experiments/download_ssl4eo.py
wangyi111: Oh That's a bug here. The rtree object is passed as an argument, so every time using the same tree.

```
patches_s1, patches_s2c, patches_s2a, center_coord = get_random_patches_rtree(
            idx, collections, bands, crops, sampler, dates, radius, debug, rtree_obj
        )
```

experiments/download_ssl4eo.py
adamjstewart: In some compiled languages like OCaml where tail recursion is automatically converted to a while loop, yes. But in Python, recursion has a lot more overhead than a while loop since you need to store all inputs at every step of the recursion stack.

experiments/download_ssl4eo.py
```diff
@@ -647,7 +641,7 @@ def worker(idx: int) -> None:
             if ext_flags[str(idx)] != 0:  # only skip downloaded ids
                 return
 
-        patches_s1, patches_s2c, patches_s2a, center_coord = get_random_patches_grid(
+        (patches_s1, patches_s2c, patches_s2a, center_coord) = get_random_patches_rtree(
```
adamjstewart, line 644: It seems like we're still calling this in parallel for each worker. Sampling should occur in serial, preferably in a separate script.

experiments/download_ssl4eo.py
```diff
@@ -627,17 +614,24 @@ def fix_random_seeds(seed: int = 42) -> None:
     else:
         ext_path = os.path.join(args.save_path, "checked_locations.csv")
 
-    # build the grid from existing coordinates
-    grid_dict: Dict[Tuple[int, int], Any] = {}
+    # build the rtree from existing coordinates
```
adamjstewart, line 617: Why are we build two separate R-trees?

experiments/download_ssl4eo.py
wangyi111: i see, i've updated with while loop

experiments/download_ssl4eo.py
wangyi111: yes that would be the next pull request

experiments/download_ssl4eo.py
wangyi111: only one. This part is to resume/initialize from existing sampled coords. The tree is then used for further sampling.

experiments/download_ssl4eo.py
adamjstewart: It might be useful to do that PR first (see https://github.com/microsoft/torchgeo/pull/1209#pullrequestreview-1367415789). Right now, the logic is very hard to follow.

experiments/download_ssl4eo.py
adamjstewart: Both sections of code insert into an R-tree though?

wangyi111: yes the same tree

experiments/download_ssl4eo.py
wangyi111: alright, I have that already. One thing is that includes extending sentinel-1/2 to other general products, no longer separate PR though. Is it OK like that?

adamjstewart: It would be better to keep each PR small and dedicated to a single change. Otherwise it becomes very difficult to review.

experiments/download_ssl4eo.py
wangyi111: yes that's what i did

experiments/download_ssl4eo.py
wangyi111: ok i'll do that

adamjstewart:
I hate to keep asking for changes, but can you split "split sampling and downloading into separate scripts" and "replace grid overlap checking with rtree" into two separate PRs? This PR contains a lot of changes and it's hard to keep track of which are related to what. It looks like the rtree sampling code is now in both scripts for some reason? Meaning I can't review it without commenting on it in two locations.

See https://github.com/microsoft/torchgeo/pull/1209#pullrequestreview-1367415789 and https://github.com/microsoft/torchgeo/pull/1209#pullrequestreview-1367415789 and the outline in the Google doc.

wangyi111:
> can you split "split sampling and downloading into separate scripts" and "replace grid overlap checking with rtree" into two separate PRs?

Sorry not familiar with the review process. After separating sampling and downloading scripts, Is it still necessary to have PR "replace grid with rtree"? The sampling is in a new script anyway, why not just use rtree for the new script?

>  It looks like the rtree sampling code is now in both scripts for some reason? Meaning I can't review it without commenting on it in two locations.

My thought was to keep it both, meaning the user can choose to:
1. run `sample_ssl4eo.py`, and then `download_ssl4eo.py` to match the sampled coords
2. only run `download_ssl4eo.py`, sample and download together.

That's the reason I put the sampling in both scripts. I can delete the second one if preferred.

So I can reopen 2 PRs:
1. split sampling and downloading scripts, the sampling script now uses rtree.
  - this is the current PR, sampling is in both scripts. If needed, I can delete in one.
3. extend to other satellites

wangyi111:
> > The sampling is in a new script anyway, why not just use rtree for the new script?
> 
> Alright, I'll make an exception since they are in separate scripts. If we can remove the code duplication and make each script as short as possible (see below), things will become easier for me to review. So let's continue with what you're trying to do in this PR.
> 
> Apologies for making everything so difficult. The existing script was rather big and lacked documentation, so I still haven't wrapped my mind around it. I'm trying to be very formal about my reviews, but this script isn't included in the core codebase that gets installed, and is only needed for our paper, so I'll try to streamline the process as much as my software engineering brain allows.
> 
> > My thought was to keep it both
> 
> Code duplication is the root of all evil. If you find a bug or add a new feature in only one copy, they start to diverge and become unmaintainable. The most important part of software engineering is to keep the code as simple as possible. Instead of supporting "grid" or "rtree", or supporting "sample + download" or "download", or supporting "uniform" or "gaussian" sampling, we should only support the bare minimum of things we actually need. With 700+ lines of code, things are too complicated for me to wrap my mind around and guarantee that there aren't any bugs. With <100 lines of code, this becomes possible. My end goal is <100 lines of code for these scripts, and I think it's possible. The sampling script should be on the order of 30 lines of code, and the download script should be on the order of 70 lines of code (not counting imports, comments, docstrings, whitespace, etc.).

Oh no should be my apologies instead, my wild way of software development:) Software that is only used by myself😉. Sorry to bother the reviewers too much, I learned a lot though😉. 

For the sampling script, don't wanna break the rule😜. I understand now that it's better to update as few codes as possible for smoother review, yet still don't get the point of two PRs:(. Maybe there's a mismatch in between. 
1. Saying we have already the original `download_ssl4eo.py` script, in which sampling and downloading are together in one function Foo. Now we want to separate them and have a better overlap checking strategy. 
2. In the new `download_ssl4eo.py`, the Foo is not there anymore, instead a Foo2 works by matching and downloading pre-sampled coords. The whole sampling and overlap checking (was grid) is removed (check this part of codes for review? 1 PR or 2 PRs won't affect this script.)
3. The `sample_ssl4eo.py` is a completely new script that does overlap checking and sampling. No existing codes to compare, thus have to review every line. In this case, writing again grid instead of directly updating it with rtree sounds not necessary?

> Code duplication is the root of all evil. If you find a bug or add a new feature in only one copy, they start to diverge and become unmaintainable.
This is a good point here! Wasn't aware of it at all...  Let me reduce the codes.

> With 700+ lines of code, things are too complicated for me to wrap my mind around and guarantee that there aren't any bugs.

My way of reviewing is to simply run it😄 I have to admit by only reading it's very hard to digest...

Thanks for all the efforts! It must be painful to see my wild updates:)

wangyi111:
> > The sampling is in a new script anyway, why not just use rtree for the new script?
> 
> Alright, I'll make an exception since they are in separate scripts. If we can remove the code duplication and make each script as short as possible (see below), things will become easier for me to review. So let's continue with what you're trying to do in this PR.
> 
> Apologies for making everything so difficult. The existing script was rather big and lacked documentation, so I still haven't wrapped my mind around it. I'm trying to be very formal about my reviews, but this script isn't included in the core codebase that gets installed, and is only needed for our paper, so I'll try to streamline the process as much as my software engineering brain allows.
> 
> > My thought was to keep it both
> 
> Code duplication is the root of all evil. If you find a bug or add a new feature in only one copy, they start to diverge and become unmaintainable. The most important part of software engineering is to keep the code as simple as possible. Instead of supporting "grid" or "rtree", or supporting "sample + download" or "download", or supporting "uniform" or "gaussian" sampling, we should only support the bare minimum of things we actually need. With 700+ lines of code, things are too complicated for me to wrap my mind around and guarantee that there aren't any bugs. With <100 lines of code, this becomes possible. My end goal is <100 lines of code for these scripts, and I think it's possible. The sampling script should be on the order of 30 lines of code, and the download script should be on the order of 70 lines of code (not counting imports, comments, docstrings, whitespace, etc.).

Oh no should be my apologies instead, my wild way of software development:) Software that is only used by myself😉. Sorry to bother the reviewers too much, I learned a lot though😉. 

For the sampling script, don't wanna break the rule😜. I understand now that it's better to update as few codes as possible for smoother review, yet still don't get the point of two PRs:(. Maybe there's a mismatch in between. 
1. Saying we have already the original `download_ssl4eo.py` script, in which sampling and downloading are together in one function Foo. Now we want to separate them and have a better overlap checking strategy. 
2. In the new `download_ssl4eo.py`, the Foo is not there anymore, instead a Foo2 works by matching and downloading pre-sampled coords. The whole sampling and overlap checking (was grid) is removed (check this part of codes for review? 1 PR or 2 PRs won't affect this script.)
3. The `sample_ssl4eo.py` is a completely new script that does overlap checking and sampling. No existing codes to compare, thus have to review every line. In this case, writing again grid instead of directly updating it with rtree sounds not necessary?

> Code duplication is the root of all evil. If you find a bug or add a new feature in only one copy, they start to diverge and become unmaintainable.
This is a good point here! Wasn't aware of it at all...  Let me reduce the codes.

> With 700+ lines of code, things are too complicated for me to wrap my mind around and guarantee that there aren't any bugs.

My way of reviewing is to simply run it😄 I have to admit by only reading it's very hard to digest...

Thanks for all the efforts! It must be painful to see my wild updates:)

wangyi111:
Should be ready now, compared to the original `download_ssl4eo.py` script:
- sampling is moved to a separate script `sample_ssl4eo.py` with rtree based overlap check.
- `download_ssl4eo.py` only works with matching pre-sampled coordinates.

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,265 @@
...
+    --radius 1320 \
+    --num_cities 10000 \
+    --std 50 \
+    --resume ./data/checked_locations.csv \
+    --indices_range 0 250000
+
+### Notes
+# The script will sample locations with rtree overlap search.
+# The script will save the sampled locations to a csv file.
+# By default, GaussianSampler is used to sample locations with a standard deviation of 50 km from top 10000 populated cities. # noqa: E501
```
adamjstewart, line 24: No need for `noqa`, you can just wrap this line onto the next line

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+""" samplers to get locations of interest points"""
+
+
+class UniformSampler:
+    def sample_point(self) -> List[float]:
+        lon = np.random.uniform(-180, 180)
+        lat = np.random.uniform(-90, 90)
+        return [lon, lat]
```
adamjstewart, line 54: This isn't used, you can delete it

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,265 @@
...
+        url = "https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/cultural/ne_110m_admin_0_countries.zip"  # noqa: E501
+        filename = "ne_110m_admin_0_countries.shp"
+        if not os.path.exists(os.path.join(download_root, os.path.basename(url))):
+            download_and_extract_archive(url, download_root)
+        sf = shapefile.Reader(os.path.join(download_root, filename))
+        return shape(sf.shapes().__geo_interface__)
+
+
+class OverlapError(Exception):
+    pass
```
adamjstewart, line 141: This isn't used anymore, I would remove it

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,265 @@
...
+    @staticmethod
+    def km2deg(kms: float, radius: float = 6371) -> float:
+        return kms / (2.0 * radius * np.pi / 360.0)
+
+    @staticmethod
+    def deg2km(deg: float, radius: float = 6371) -> float:
+        return deg * (2.0 * radius * np.pi / 360.0)
+
+
+class BoundedUniformSampler:
```
adamjstewart, line 111: This isn't used anywhere, I would remove it

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,265 @@
...
+        "--indices_range",
+        type=int,
+        nargs=2,
+        default=[0, 250000],
+        help="indices to download",
+    )
+
+    args = parser.parse_args()
+
+    fix_random_seeds(seed=42)
```
adamjstewart, line 208:
```suggestion
    np.random.seed(42)
```
Now you can delete the function and save 4 lines of code.

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,265 @@
...
+    def sample_point(self) -> List[float]:
+        lon = np.random.uniform(-180, 180)
+        lat = np.random.uniform(-90, 90)
+        return [lon, lat]
+
+
+class GaussianSampler:
+    def __init__(
+        self,
+        interest_points: Optional[List[List[float]]] = None,
```
adamjstewart, line 60: This feature is never used, I would remove it

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+class UniformSampler:
+    def sample_point(self) -> List[float]:
+        lon = np.random.uniform(-180, 180)
+        lat = np.random.uniform(-90, 90)
+        return [lon, lat]
+
+
+class GaussianSampler:
```
adamjstewart, line 57: Using a class for this feels like overkill, especially now that we will only define a single sampler. I think you're better off converting each staticmethod to a normal function.

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,265 @@
...
+    idx: int,
+    sampler: Any,
+    radius: float,
+    overlap_ratio: float,
+    rtree_obj: index.Index = index.Index(),
+) -> List[float]:
+    """sample new coord and check overlap --- rtree"""
+    # use rtree to avoid strong overlap
+    count = 0
+    bbox_radius = (1 - overlap_ratio) * radius / 1000
```
adamjstewart, line 154: Let's not call it "radius" since we're using boxes instead of circles now. "size" is what we usually use in TorchGeo to mean width/height

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,265 @@
...
+    while count < 1:
+        new_coord = sampler.sample_point()  # (lon,lat) of top-10000 cities
+        bbox = (
+            new_coord[0] - sampler.km2deg(bbox_radius),
+            new_coord[1] - sampler.km2deg(bbox_radius),
+            new_coord[0] + sampler.km2deg(bbox_radius),
+            new_coord[1] + sampler.km2deg(bbox_radius),
```
adamjstewart, line 161: Instead of calling this function 4 times with the same input, you can call it once and save that to a variable

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,265 @@
...
+
+    rtree_coords = index.Index()
+    bbox_radius = (
+        (1 - args.overlap_ratio) * args.radius / 1000
+    )  # allow little overlap between adjacent patches
+    if args.resume:
+        print("Load existing locations.")
+        for i, key in enumerate(tqdm(ext_coords.keys())):
+            c = ext_coords[key]
+            rtree_coords.insert(
```
adamjstewart, line 235: This is still way more complicated than it needs to be. You're inserting into an R-tree two times for every point you sample, here and on line 165. The design I proposed [here](https://github.com/microsoft/torchgeo/pull/1213#pullrequestreview-1367905066) is only 9 lines of code and should do the same thing but twice as fast. Every single insert is $\mathcal{O}(n)$, and we insert $n$ times. So the overall complexity of this sampling script is $\mathcal{O}(n^2)$, so doubling that is bad.

I think a lot of the fancy abstractions in this script (classes and a multitude of functions) are actually making it much slower. I think you can delete 90% of it and just use a single while loop:
```python
with open('locations.csv', 'w', newline='') as f:
    w = csv.writer(f)
    idx = Index()
    counter = 0
    while counter < total_num_patches:
        bbox = choose_random_bbox()  # this might be worth making a separate function for
        if list(idx.intersection(bbox)):
            continue
        idx.insert(counter, bbox)
        w.writerow(bbox)
        counter += 1
```
This is what I mean when I say the entire script could be ~30 lines of code.

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,265 @@
...
+    )
+
+    args = parser.parse_args()
+
+    fix_random_seeds(seed=42)
+
+    # initialize sampler
+    sampler = GaussianSampler(num_cities=args.num_cities, std=args.std)
+
+    # if resume
```
adamjstewart, line 213: How long does the entire script take to run? If it's ~1 hr, we should just get rid of all the resume logic. If it's ~1 day, then the resume logic is helpful.

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,265 @@
...
+        "--std", type=int, default=50, help="std of gaussian distribution"
+    )
+    parser.add_argument(
+        "--resume", type=str, default=None, help="resume from a previous run"
+    )
+    parser.add_argument(
+        "--indices_range",
+        type=int,
+        nargs=2,
+        default=[0, 250000],
```
adamjstewart, line 202: Why a range instead of a total number of patches to sample?

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+def fix_random_seeds(seed: int = 42) -> None:
+    np.random.seed(seed)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--save_path", type=str, default="./data/", help="dir to save data"
```
adamjstewart, line 178: I usually use `-` instead of `_` for flags since that's what most Unix programs use. Argparse automatically converts these to underscores when defining parsed variables.

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,265 @@
...
+        points = [[float(c["lng"]), float(c["lat"])] for c in cities]
+        return points
+
+    @staticmethod
+    def km2deg(kms: float, radius: float = 6371) -> float:
+        return kms / (2.0 * radius * np.pi / 360.0)
+
+    @staticmethod
+    def deg2km(deg: float, radius: float = 6371) -> float:
+        return deg * (2.0 * radius * np.pi / 360.0)
```
adamjstewart, line 108: This function is never used, I would delete it

adamjstewart (commented):
Only reviewed the new sampling script so far, will review the download script tomorrow. I'll prob focus my "how can we make the download script simpler" review for the follow-up PR in which we make it generic enough to work for any satellite.

P.S. If the time zone lag between you committing and me reviewing is too big I can always push commits directly to your PR if you don't mind. Just didn't want to take over your PR without your permission.

P.P.S. I would recommend creating a separate branch for PRs though, since you won't be able to fast forward your main branch ever again.

experiments/sample_ssl4eo.py
wangyi111: this is also related to resuming. saying you have 0 - 100 already, next time you can start from 100 to 200.

wangyi111:
> If the time zone lag between you committing and me reviewing is too big I can always push commits directly to your PR if you don't mind.

Sure if you are certain pls do so. This could save some time.

> I would recommend creating a separate branch for PRs though

Yes I'll do that. Thought it's only one script that's not related to other codes😅

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,193 @@
...
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--save-path", type=str, default="./data/", help="dir to save data"
+    )
+    parser.add_argument(
+        "--radius", type=float, default=1320, help="patch radius in meters"
+    )
+    parser.add_argument(
+        "--overlap-ratio",
```
adamjstewart, line 107: There should be no overlap

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,193 @@
...
+        type=float,
+        default=0.25,
+        help="max overlap ratio between adjacent patches",
+    )
+    parser.add_argument(
+        "--num-cities", type=int, default=10000, help="number of cities to sample"
+    )
+    parser.add_argument(
+        "--std", type=int, default=50, help="std of gaussian distribution"
```
adamjstewart, line 116:
```suggestion
        "--std", type=int, default=50, help="std dev of gaussian distribution"
```

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,193 @@
...
+    )
+    parser.add_argument(
+        "--resume", type=str, default=None, help="resume from a previous run"
```
adamjstewart, line 119: Why not make this a boolean and instead use `--save-path` for the file to resume from?

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,193 @@
...
+    if args.resume:
+        print("Load existing locations.")
+        for i, key in enumerate(tqdm(ext_coords.keys())):
+            c = ext_coords[key]
+            bbox = create_bbox(c, bbox_size_degree)
+            rtree_coords.insert(i, bbox)
+
+    # sample locations
+    start_time = time.time()
+    indices = range(args.indices_range[0], args.indices_range[1])
```
adamjstewart, line 165:
```suggestion
    indices = range(*args.indices_range)
```

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,193 @@
...
+        nargs=2,
+        default=[0, 250000],
+        help="indices to download",
+    )
+
+    args = parser.parse_args()
+    np.random.seed(42)
+
+    # if resume
+    ext_coords = {}
```
adamjstewart, line 133: Might be faster to use a list instead of a dictionary. We aren't randomly sampling indexing, it's an ordered mapping from integer to value

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,193 @@
...
+                interest_points, args.std
+            )  # (lon,lat) of top-10000 cities
+            bbox = create_bbox(new_coord, bbox_size_degree)
+            if list(rtree_coords.intersection(bbox)):
+                continue
+            rtree_coords.insert(idx, bbox)
+            count += 1
+
+        # save to file
+        with open(ext_path, "a") as f:
```
adamjstewart, line 186: This opens and closes the file hundreds of thousands of times. You should do this outside of the outermost loop so we don't have to reopen the file every time.

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,193 @@
...
+            writer = csv.writer(f)
+            data = [idx, new_coord[0], new_coord[1]]
```
adamjstewart, line 188:
```suggestion
            data = [idx] + new_coord
```

experiments/download_ssl4eo.py
```diff
@@ -627,39 +497,42 @@ def fix_random_seeds(seed: int = 42) -> None:
...
+    # if match from pre-sampled coords (e.g. SSL4EO-S12)
+    if args.match_file:
+        match_coords = {}
+        with open(args.match_file) as csv_file:
+            reader = csv.reader(csv_file)
+            for row in reader:
+                key = row[0]
+                val1 = float(row[1])
+                val2 = float(row[2])
+                match_coords[key] = (val1, val2)  # lon, lat
```
adamjstewart, line 509: Where do these centroid coordinates get converted to the bbox to download? I don't see a size/radius parameter to the script.

experiments/download_ssl4eo.py
```diff
@@ -528,16 +398,19 @@ def fix_random_seeds(seed: int = 42) -> None:
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument("--save_path", type=str, default="./data/")  # dir to save data
```
adamjstewart, line 400: Can we make these dashes instead of underscores too?

experiments/sample_ssl4eo.py
wangyi111: some freedom to load from a customize-named file?

experiments/sample_ssl4eo.py
wangyi111: I think dictionary is faster?

experiments/sample_ssl4eo.py
wangyi111: new_coord is a tuple (lon,lat), so maybe keep it?

experiments/download_ssl4eo.py
wangyi111: Here in the `get_patch_s2()` function for example:
```
    region = (
        ee.Geometry.Point(center_coord).buffer(radius).bounds()
    )  # sample region bound
    patch = image.select(*bands).sampleRectangle(region, defaultValue=0)
```
It samples a rectangle centered around the centroid of the bbox.

experiments/sample_ssl4eo.py
adamjstewart: In that case you can use:
```suggestion
            data = [idx, *new_coord]
```

adamjstewart:
Can you make these files executable? `chmod +x <file>`

adamjstewart:
Also need a shebang and license at the top of each file:
```python
#!/usr/bin/env python3                                                                   
                                                                                         
# Copyright (c) Microsoft Corporation. All rights reserved.                              
```

wangyi111:
> Can you make these files executable? `chmod +x <file>`

I run it from my local terminal and commit the file?

adamjstewart:
> I run it from my local terminal and commit the file?

Yep

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,191 @@
...
+    url = "https://simplemaps.com/static/data/world-cities/basic/simplemaps_worldcities_basicv1.71.zip"  # noqa: E501
+    filename = "worldcities.csv"
+    if not os.path.exists(os.path.join(download_root, os.path.basename(url))):
+        download_and_extract_archive(url, download_root)
+    with open(os.path.join(download_root, filename), encoding="UTF-8") as csvfile:
+        reader = csv.DictReader(csvfile, delimiter=",", quotechar='"')
+        cities = []
+        for row in reader:
+            row["population"] = (
+                row["population"].replace(".", "") if row["population"] else "0"
```
adamjstewart, line 63: This is a bug. For example, the population of Crowsnest Pass, Canada is 5589, but the population in the CSV file is 5589.00. So it's a decimal place, you shouldn't remove it and multiple the number by 100.

Honestly I would use pandas for CSV file reading, it's much more powerful.

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,191 @@
...
+            )
+            cities.append(row)
+    return cities
+
+
+def get_interest_points(
+    cities: List[Dict[str, str]], size: int = 10000
+) -> List[List[float]]:
```
adamjstewart, line 71: Might be better to use a tuple and type this as `List[Tuple[float, float]]` everywhere, this is more specific than `List[List[float]]` which could have any number of points per tuple.

adamjstewart (commented):
Couple minor comments and this should be good to go!

experiments/sample_ssl4eo.py
adamjstewart: Okay, technically array > dict > list, and technically Python's list implementation is an arraylist, which I think is still faster than a dict, but I'm not sure. It's prob fine as is.

experiments/sample_ssl4eo.py
wangyi111: true👍

wangyi111: One bug is there's one city 'Tecax' whose population in the csv is 50.789 but in fact should be 50789. The original code would work in this case. I've fixed them with pandas.

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+def get_world_cities(download_root: str = "world_cities") -> pd.DataFrame:
+    url = "https://simplemaps.com/static/data/world-cities/basic/simplemaps_worldcities_basicv1.71.zip"  # noqa: E501
+    filename = "worldcities.csv"
+    if not os.path.exists(os.path.join(download_root, os.path.basename(url))):
+        download_and_extract_archive(url, download_root)
+    cities = pd.read_csv(os.path.join(download_root, filename))
+    cities.at[8436, "population"] = 50789  # fix one bug (Tecax) in the csv file
+    cities = cities[["city", "lat", "lng", "population"]]
```
adamjstewart, line 61:
```suggestion
    cols = ["city", "lat", "lng", "population"]
    cities = pd.read_csv(os.path.join(download_root, filename), usecols=cols)
    cities.at[8436, "population"] = 50789  # fix one bug (Tecax) in the csv file
```
This will make reading faster

adamjstewart:
It's hard to test this because you didn't create a separate branch for the PR. For the next PR, it would help to create a branch.

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,189 @@
...
+    return cities
+
+
+def get_interest_points(
+    cities: pd.DataFrame, size: int = 10000
+) -> List[Tuple[float, float]]:
+    cities = cities.sort_values(by=["population"], ascending=False).head(size)
+    points = [(float(c["lng"]), float(c["lat"])) for _, c in cities.iterrows()]
```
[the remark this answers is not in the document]
wangyi111: Let me do that.

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,178 @@
...
+    return bbox
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--save-path", type=str, default="./data/", help="dir to save data"
+    )
+    parser.add_argument(
+        "--size", type=float, default=1320, help="half patch size in meters"
```
adamjstewart, line 99: If I want to download 264 x 264 px image patches, would I use 264 * 30 (For Landsat with 30 m resolution)?

experiments/sample_ssl4eo.py
```diff
@@ -0,0 +1,178 @@
...
+    )
+
+    args = parser.parse_args()
+
+    # if resume
+    ext_coords = {}
+    if args.resume:
+        ext_path = args.resume
+        with open(ext_path) as csv_file:
+            reader = csv.reader(csv_file)
```
adamjstewart, line 125: Could use pandas here too (for all of ext_coords)

adamjstewart (approved):
This is probably good enough for now. I can always make additional changes once it's merged and I start testing things. Let me know if you have any last changes to make before I merge. Then we can get started on making it generic enough to support any satellite.

wangyi111:
No further updates from my side, and yes let's go to next step!

experiments/sample_ssl4eo.py
wangyi111: landsat 8 has two bands with 15 m resolution, if want to match them, should be 264*15 / 2 = 1980

adamjstewart: We'll prob match the 30 m bands and downsample the 15 m bands.

PR #1214, opened 2023-04-01
Deps: make omegaconf/tensorboard optional

adamjstewart:
### Why optional?

TorchGeo does not import or require either of these dependencies. The only thing these deps are required for is our trainer tests, for [a single notebook](https://github.com/microsoft/torchgeo/pull/1189), and for train.py and friends. None of these files get installed when you run `pip install torchgeo`, so the dependency shouldn't be required.

### Why now?

Omegaconf was always an issue (every time you update the conda-forge recipe it complains that it isn't used), but the lightning/tensorboard issue is relatively new. Starting with lightning 1.9, tensorboard was moved from a required dependency to [an optional extra](https://github.com/Lightning-AI/lightning/pull/16349). However, our tests need tensorboard for coverage, so we added a dependency on `lightning[extra]` to pull in tensorboard. However, this pulls in [a lot of other deps](https://github.com/Lightning-AI/lightning/blob/master/requirements/pytorch/extra.txt) in addition to tensorboard. To minimize the deps we install, we should avoid using `[extra]`.

Also, Spack has trouble concretizing `py-lightning+extra` due to conflicting setuptools versions required.

### Why tensorboard and not tensorboardX?

Lightning used to install tensorboard, then [switched to tensorboardX](https://github.com/Lightning-AI/lightning/pull/15728) (both are compatible with the same API). The reason for this switch was that tensorboardX has fewer dependencies and is easier to install on more platforms. The only other difference I've seen between the two that affects us is that tensorboardX does not support the `%load_ext tensorboard` Jupyter magic that we use in our notebook. Since our notebook tests require tensorboard, I figured we should just go with that.

FYI @bugraaldal

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -72,7 +72,7 @@ jobs:
     - name: Install pip dependencies
       if: steps.cache.outputs.cache-hit != 'true'
       run: |
-        pip install .[docs,tests] planetary_computer pystac pytest-rerunfailure
+        pip install .[docs,tests] planetary_computer pystac pytest-rerunfailures
         pip list
     - name: Run notebook checks
       run: pytest --nbmake --durations=10 --reruns=10 docs/tutorials
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -72,7 +72,7 @@ jobs:
     - name: Install pip dependencies
       if: steps.cache.outputs.cache-hit != 'true'
       run: |
-        pip install .[docs,tests] planetary_computer pystac pytest-rerunfailures tensorboard
+        pip install .[docs,tests] planetary_computer pystac pytest-rerunfailure
         pip list
     - name: Run notebook checks
       run: pytest --nbmake --durations=10 --reruns=10 docs/tutorials
--- a/.github/workflows/tutorials.yaml
+++ b/.github/workflows/tutorials.yaml
@@ -30,7 +30,7 @@ jobs:
     - name: Install pip dependencies
       if: steps.cache.outputs.cache-hit != 'true'
       run: |
-        pip install .[docs,tests] planetary_computer pystac pytest-rerunfailures tensorboard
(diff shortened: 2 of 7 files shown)


.github/workflows/release.yaml
```diff
@@ -72,7 +72,7 @@ jobs:
     - name: Install pip dependencies
       if: steps.cache.outputs.cache-hit != 'true'
       run: |
-        pip install .[docs,tests] planetary_computer pystac pytest-rerunfailures tensorboard
+        pip install .[docs,tests] planetary_computer pystac pytest-rerunfailures
```
adamjstewart, line 75: The rest of these are things that we could subsume into `[tests]`, but they're also dependencies that I would like to remove in the future, and I don't think they're worth tracking.

calebrob6:
I think having tensorboard inside the notebooks isn't necessary (this is why I used CSV logger at one point).

calebrob6 (dismissed):

adamjstewart:
You should try the new trainers tutorial, it's beautiful. The CSV logger is an extra 20 lines of code and makes it seem like using TorchGeo is way harder than it actually is. No user should ever have to manually parse a CSV file and make their own loss/accuracy plots.

calebrob6:
I get that it looks cool :), but notebooks aren't _really_ the environment for running organized experiments. In my mind the tutorials are to show off what torchgeo can do, the point of showing the loss at all is "it goes down, training is working" not showing the logging capabilities of lightning

adamjstewart:
For sure, but the point of the tutorials is also not to teach you how to parse and plot a CSV file. The tutorials should be the bare minimum lines of code. The previous tutorial had more lines for plotting than it did for training. Now plotting is a single line of tensorboard.

calebrob6:
(but, as a consequence, brings in more dependencies)

calebrob6 (approved):

calebrob6:
Also, I don't care one way or the other, just suggesting a path for using tensorboardX

PR #1218, opened 2023-04-05
Plot landsat bands

adamjstewart:
Tried reproducing [this plot](https://www.usgs.gov/media/images/comparison-landsat-7-and-8-bands-sentinel-2):

dmidS2LS7Comparison

Still a far cry from the original, but I think it's a plot we'll want for our paper:

Figure_1

Issues so far:

* Apparently matplotlib doesn't support axis breaks? In 2023?? So you have to roll your own with 2 subplots
* Because of this, we have to use subxlabel instead of xlabel, meaning the position and size of the xlabel differ from ylabel
* Some bands are too narrow to visualize, not sure how USGS cheated to make them easier to see
* Need to figure out how to add band numbers to boxes
* Don't know where to find atmospheric transmission data


adamjstewart:
CSV files lack the structure of dictionaries. I could store it in an XML file, but those are pretty verbose. I don't envision this being used by any other files, it's just a plot for the paper.

calebrob6:
okay, a json file then

adamjstewart:
What would be the benefit of a separate file?

calebrob6:
Its much easier to understand the data that you are using for the plots (so, e.g. I can try to make a similar plot)

adamjstewart:
You can always import the dict from this script. The dict in Python will be identical to the dict in JSON, so there's no readability improvement.

calebrob6:
Right, but I like to keep data and logic separate. Noone wants to import "plot_landsat_bands.py" to get the data. What happens if I want to add the Sentinel-2 data? What happens if I want to take the data and use it in a JS script?

adamjstewart:
Alright, that's a pretty good point. You win. Also, it's not importable at the moment without running all lines of code cuz I didn't check `__name__`.

calebrob6:
Hmmm this isn't really what I meant... let me take a stab

calebrob6:
This way of organizing makes it easier to write out the band numbers if you want:

image

adamjstewart:
Figure_1

adamjstewart:
Figure_2

calebrob6 (approved):

PR #1223, opened 2023-04-07
Download generic products from GEE

wangyi111:
I hope I am doing it correctly this time:)


experiments/download_ssl4eo.py
```diff
@@ -393,98 +288,111 @@ def update(self, delta: int = 1) -> int:
...
+    )
+    # collection properties
+    parser.add_argument(
+        "--collection", type=str, default="COPERNICUS/S2", help="GEE collection name"
+    )
+    parser.add_argument(
+        "--qa-band", type=str, default="QA60", help="qa band name (optional)"
+    )  # optional
+    parser.add_argument(
+        "--qa-cloud-bit", type=int, default=10, help="qa band cloud bit (optional)"
```
adamjstewart, line 304: Landsat has several cloud-related bits (https://developers.google.com/earth-engine/datasets/catalog/LANDSAT_LC08_C02_T1_TOA). Should we only consider bit 3, or other cloud bits too?

experiments/download_ssl4eo.py
```diff
@@ -277,69 +205,36 @@ def get_patch_s2(
...
         if debug:
             print(e)
-        return None, None, None, coords
+        return None, coords
 
-    return patches_s1, patches_s2c, patches_s2a, center_coord
+    return patches, coords
 
 
 def save_geotiff(
```
adamjstewart, line 240: I know @calebrob6 had a question about image compression. We should quickly benchmark to see whether no compression or high compression is better for I/O. Note that HPC clusters tend to have extremely slow filesystems, so a smaller compressed file may actually be faster to load even if it needs to be decompressed. Also makes downloads faster. I know you guys used HDF5 or some other file format for faster loading, but I think we'll want to keep geospatial metadata if possible and make the files easy to load.

experiments/download_ssl4eo.py
```diff
@@ -587,10 +471,11 @@ def worker(idx: int) -> None:
 
     # set indices
     if args.indices_range is not None:
-        indices = range(args.indices_range[0], args.indices_range[1])
+        indices = list(range(args.indices_range[0], args.indices_range[1]))
```
adamjstewart, line 474:
```suggestion
        indices = list(range(*args.indices_range))
```
Does it have to be a list? This will be a pretty large list.

experiments/download_ssl4eo.py
```diff
@@ -576,7 +460,7 @@ def worker(idx: int) -> None:
         # add to existing checked locations
         with open(ext_path, "a") as f:
             writer = csv.writer(f)
-            if patches_s2c is not None:
+            if patches is not None:
                 success = 1
             else:
                 success = 0
```
adamjstewart, line 466: Could use `*center_coord` below here

experiments/download_ssl4eo.py
```diff
@@ -576,7 +460,7 @@ def worker(idx: int) -> None:
         # add to existing checked locations
         with open(ext_path, "a") as f:
```
adamjstewart, line 461: It's a shame we have to open and close a file for every single patch, but I don't see a better way to do it if we want to support parallelism.

experiments/download_ssl4eo.py
```diff
@@ -519,52 +427,28 @@ def worker(idx: int) -> None:
...
-                        raster=patch["raster"],
-                        coords=patch["coords"],
-                        metadata=patch["metadata"],
-                        path=location_path_s2a,
-                    )
-                # s1
-                location_path_s1 = os.path.join(args.save_path, "s1", f"{idx:06d}")
-                os.makedirs(location_path_s1, exist_ok=True)
-                for patch in patches_s1:
+                location_path = os.path.join(args.save_path, "imgs", f"{idx:06d}")
```
adamjstewart, line 444: Doesn't your SSL4EO-S12 dataset use 7 digits instead of 6? I would like to remain consistent if we can.

experiments/download_ssl4eo.py
```diff
@@ -393,98 +288,111 @@ def update(self, delta: int = 1) -> int:
...
+            "B11",
+            "B12",
+        ],
+        help="bands to download",
+    )
+    parser.add_argument(
+        "--crops",
+        type=int,
+        nargs="+",
+        default=[44, 264, 264, 264, 132, 132, 132, 264, 132, 44, 44, 132, 132],
```
adamjstewart, line 351: Won't this always be equal to the radius? Or is this in case the radius is off by half a pixel or something? I wonder if we could combine these two args

experiments/download_ssl4eo.py
```diff
@@ -393,98 +288,111 @@ def update(self, delta: int = 1) -> int:
...
+        help="crop size for each band",
+    )
+    parser.add_argument("--dtype", type=str, default="float32", help="data type")
```
adamjstewart, line 354: Could we use the dtype of the data that gets downloaded instead of overriding?

adamjstewart (commented):
Mostly questions. I'm not super familiar with GEE internals so I can't provide a very thorough review.

Once this is merged I'm planning on doing some benchmarking to check for bugs and determine:

* Optimal parallelism we're allowed to use
* Percent of patches that we can expect to be cloud-free
* Whether or not multiple people running on the same IP at the same time helps

Then we can gauge how big of an R-tree we need to build and how long it will take to download each sensor.

calebrob6:
I ran the sample/download scripts and am seeing it is extremely slow. Is this the type of output that you all get when running this?

image

calebrob6:
Note: at this rate it would take 11 days to download 250k patches.

calebrob6:
Also, let me check to see if I understand the logic of this script:

We have these reference dates: "2021-09-22", "2021-06-21", "2021-03-20", "2020-12-21"
Now for each (lat, lon) point we:
- Randomly subtract a different random number of days between 0 and 365 from each reference date
- Check for data within a 15 day window around that new date
- If there are no valid scenes with cloud cover < 20% in each of the 30 day windows, then we discard that (lat, lon) point as not having valid data

Is this correct?

adamjstewart:
What parallelism are you using?

calebrob6:
8

adamjstewart:
Our node has 128 cores, so as long as there's no rate limit, it would only take us 16 hrs to download. And a few of us have compute nodes we have access to, so if there is a rate limit we can spread it out across people.

experiments/download_ssl4eo.py
```diff
@@ -42,120 +68,65 @@
...
     date2 = date + timedelta(days=days / 2)
-    return date2str(date1), date2str(date2)
+    date3 = date1 - timedelta(days=365)
+    date4 = date2 - timedelta(days=365)
+    return (
+        date2str(date1),
+        date2str(date2),
+        date2str(date3),
+        date2str(date4),
+    )  # two-years buffer
```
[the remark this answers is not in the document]
wangyi111: Nope, sampling whenever should also be fine

experiments/download_ssl4eo.py
wangyi111: This I don't know actually:) Also, cloud-related bits are used in the "mask2clouds" function (replaces the cloud pixels with zero), which is not used by default. I guess only cloud filtering is important, either masking or not should be fine.

experiments/download_ssl4eo.py
wangyi111: Not a must, i think I made it a list to be consistent with the other option.
```
# set indices
if args.indices_range is not None:
    indices = list(range(args.indices_range[0], args.indices_range[1]))
else:
    indices = []
    for key in match_coords.keys():
        indices.append(int(key))
```

experiments/download_ssl4eo.py
wangyi111: yeah...

experiments/download_ssl4eo.py
wangyi111: yep, sure

experiments/download_ssl4eo.py
wangyi111: yes equal to  radius*2. the reason here is that different bands may have different resolutions which are not fixed...

wangyi111:
> I ran the sample/download scripts and am seeing it is extremely slow. Is this the type of output that you all get when running this?
> 
> image

unfortunately this is similar to my downloading😄

experiments/download_ssl4eo.py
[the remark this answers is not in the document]
wangyi111: > 

One potential limit is the max query frequency from GEE: https://developers.google.com/earth-engine/guides/usage#adjustable_quota_limits. I didn't meet this issue, but if using all >100 cores this may show up.

wangyi111:
> Our node has 128 cores, so as long as there's no rate limit, it would only take us 16 hrs to download. And a few of us have compute nodes we have access to, so if there is a rate limit we can spread it out across people.

One potential limit is the max query frequency from GEE: https://developers.google.com/earth-engine/guides/usage#adjustable_quota_limits. I didn't meet this issue, but if using all >100 cores this may show up.

calebrob6:
> This was the logic for seco and ssl4eo. In the current downloading script I made it simpler for potentially more candidates:
> 
> four reference dates in e.g. 2021
> for each date, check for +/- 30 days in 2021 and 2020
> If there are no valid scenes with cloud cover < 20% in the window, discard this point
> reference dates, temporal window size, how many years as a buffer, cloud cover percentage are all arguments that can be defined.

I see now that logic is changed in this PR! Thanks for clarifying!

experiments/download_ssl4eo.py
adamjstewart: Ah, I see. So `--meta-cloud-name` is the flag that actually matters. I see Landsat uses `CLOUD_COVER`. Not sure how that's calculated, but I'm fine with trusting it to be useful.

experiments/download_ssl4eo.py
adamjstewart: I had to get our inode limit increased, but we might be okay. I'm hoping to avoid this but we'll see if it's too slow.

experiments/download_ssl4eo.py
adamjstewart: Is there a way to read the band resolutions from GEE? We're already passing the radius in geospatial coords to `get_patch`. Won't that automatically return patches with the right dimensions?

experiments/download_ssl4eo.py
adamjstewart: Nah, there's no reason to use anything higher than float32. Is the data returned float64, or is the default dtype for float ndarrays float64?

experiments/download_ssl4eo.py
wangyi111: Most patches will already return the right sizes, but some will return slightly different (e.g. 264x265). Thus the "crop" argument is also optional (can be None).

adamjstewart:
Doing some benchmarking now. First bug I noticed: `sample_ssl4eo.py` fails if the `--save-path` doesn't exist. Needs an `os.makedirs(args.save_path, exist_ok=True)`.

adamjstewart:
Next bug: `sample_ssl4eo.py` saves `world_cities` in the current directory instead of in `--save-path`.

adamjstewart:
Third bug: in `download_ssl4eo.py`, we accept any image with cloud percentage < 20, but at least for Landsat 8, a negative value means that a cloud pct wasn't calculated. So we should instead use 0 <= cloud pct <= 20.

adamjstewart:
I'm seeing a new problem now. Let me know if I'm doing something wrong. I want 264 x 264 px patches, at 30 m resolution, so I'm using:
```console
$ python3 experiments/sample_ssl4eo.py --save-path data/ssl4eo-l8-l1 --size $((30 * 264 / 2)) --num-cities 10000 --std 50 --indices-range 0 1000
$ python3 experiments/download_ssl4eo.py --save-path data/ssl4eo-l8-l1 --collection LANDSAT/LC08/C02/T1_TOA --qa-band QA_PIXEL --qa-cloud-bit 3 --meta-cloud-name CLOUD_COVER --radius $((30 * 264 / 2)) --bands B1 B2 B3 B4 B5 B6 B7 B8 B9 B10 B11 --crops 264 264 264 264 264 264 264 $((264 * 2)) $((264 * 2)) 264 264 --num-workers 8 --match-file data/ssl4eo-l8-l1/sampled_locations.csv --indices-range 0 1000 --debug
```
However, I'm seeing:
```
Image.sampleRectangle: Too many pixels in sample; must be <= 262144. Got 287832.
no suitable image for location 64.
Image.sampleRectangle: Too many pixels in sample; must be <= 262144. Got 279312.
no suitable image for location 193.
ImageCollection.filter: No suitable images found in (22.0921, 111.2269) between 2021-08-23 and 2021-10-22.
no suitable image for location 98.
Image.sampleRectangle: Too many pixels in sample; must be <= 262144. Got 284089.
no suitable image for location 0.
Image.sampleRectangle: Too many pixels in sample; must be <= 262144. Got 287832.
no suitable image for location 33.
```
The total number of pixels should be $264^2 = 69696$, but instead it's giving ~300K. And the number changes depending on location. Could it be something to do with the meters to degrees conversion?

adamjstewart:
Fixed the first 3 bugs, not sure what's the deal with the image size bug.

adamjstewart:
Alternatively, could we download all bands at 30 m resolution?

wangyi111:
> Which line of code actually does the download? I know `Export.image.toDrive` has a `maxPixels` parameter that can be set to increase the upper limit.

the download is line 182: 
```
features = patch.getInfo() # the actual download
```
I think the bottleneck here is the sampleRectangle function.

> Alternatively, could we download all bands at 30 m resolution?

This probably won't solve the sampleRectangle issue, as we have to download the data first then upsample or downsample. I just checked the GEE API documentation, one other solution might be sample the rectangle (or bounding box) ourself? 

```
## an example bbox region
bbox = ee.Geometry.BBox(-121.55, 39.01, -120.57, 39.38)
patch = image.select(*bands).sample(region=bbox)
```
Haven't checked if this would work.

Other way is to filter the collection early with the bounding box:
```
# current, line 129
filtered = filtered.filterBounds(ee.Geometry.Point(coords)) # filter region
# new?
filtered = filtered.filterBounds(ee.Geometry.BBox(bbox)) # filter region
```
This may require some changes later.

wangyi111:
maybe the simplest solution is to reduce the image size a bit? e.g. 256 for 30m and 512 for 15m. In ssl4eo, we cropped the input image to 224 for training in the end, and 256 is also commonly used.

adamjstewart:
I would like to use the same size if possible. Let me poke around and see if there's anything we can do. I would be surprised if you can't downsample an image in GEE before downloading.

wangyi111:
emm, it could be possible with [`ee.Image.reproject`](https://developers.google.com/earth-engine/apidocs/ee-image-reproject):

```
patch = image.select('B8').reproject(crs=image.select('B8').projection().crs(), scale=30).sampleRectangle(region, defaultValue=0)
```

Note that 'B8' has a different projection with other bands, so have to split the bands here.

adamjstewart:
Yep, I don't see a way to increase the maxPixels for these functions. That's unfortunate.

How much work would it be to modify the script to download and write a single file containing all bands at the same resolution and CRS? This would also help us reduce inodes. CRS and res could either be script flags, or we could take them from the first band.

If this is a lot of work we'll just have to reduce the patch size and do this manually locally...

wangyi111:
I think it's doable like mentioned before (using a loop to reproject every band to a certain resolution, and concatenate the bands to save a whole image), but would slow the downloading time (i did a brief test, the time is doubled).

> ```
> patch = image.select('B8').reproject(crs=image.select('B8').projection().crs(), scale=30).sampleRectangle(region, defaultValue=0)
> ```

adamjstewart:
Is it double just for B8 or double for the entire scene? The former I can live with but the latter is painful.

Will do some benchmarking with 256 just to see how long it takes to download.

adamjstewart:
256 is still too big, need 240 to avoid limit.

adamjstewart:
Tried visualizing some of the images and there are at least a few with significant cloud cover. For example:

* 99,-87.3207855306882,42.467810108587045
* LC08_187024_20210907 looks clear, LC08_188024_20210306 is completely cloudy

Not sure if I'm using the wrong variable or what.

Screenshot 2023-04-19 at 12 16 38 PM
Screenshot 2023-04-19 at 12 16 50 PM

adamjstewart:
There are also some images where a significant portion of the image is nodata pixels:

Screenshot 2023-04-19 at 12 19 17 PM

wangyi111:
> Is it double just for B8 or double for the entire scene? The former I can live with but the latter is painful.
> 
> Will do some benchmarking with 256 just to see how long it takes to download.

for the entire scene, as to make it generic have to reproject every band (to itself if already 30m resolution).

wangyi111:
> EDIT: seems like it's for the entire image, it's just barely below 20%:

Yes for the whole scene, as the cloud cover metadata is for a whole scene. It can be kind of improved by uncommenting line 110:
```
# Uncomment the following line if you want to apply cloud masking.
# collection = collection.map(maskS2clouds, args)
```

However, the cloud metadata is not very accurate, same as cloud mask. Therefore, cloudy patches can not be totally avoided unfortunately.

adamjstewart:
For some reason the script runs fine when I run it in an interactive job, but hangs without feedback when I run it in a batch script. Any ideas what I'm doing wrong?

wangyi111:
Probably not able to help here. I never ran in a batch job, but some computing nodes do not have internet access.

adamjstewart:
> for the entire scene, as to make it generic have to reproject every band (to itself if already 30m resolution).

What if we hack it just for this time so that it only reprojects B8? I think I like that option best. Then we get 1 file per scene instead of 11 (for L8 TOA).

wangyi111:
> What if we hack it just for this time so that it only reprojects B8?

I just committed one that should work for L8. Note this temporary script is not generic.

adamjstewart:
How much would things slow down if we use the cloud mask instead of the overall image cloud percentage?

experiments/download_ssl4eo.py
```diff
@@ -576,21 +482,22 @@ def worker(idx: int) -> None:
...
             else:
                 success = 0
-            data = [idx, center_coord[0], center_coord[1], success]
+            data = [idx, *center_coord, success]
             writer.writerow(data)
 
         return
 
     # set indices
     if args.indices_range is not None:
```
adamjstewart, line 495: I don't think it's possible for this to be None

experiments/download_ssl4eo.py
```diff
@@ -519,52 +449,28 @@ def worker(idx: int) -> None:
...
             match_coords=match_coords,
         )
 
-        if (
-            patches_s1 is not None
-            and patches_s2c is not None
-            and patches_s2a is not None
-        ):
+        if patches is not None:
             if args.save_path is not None:
```
adamjstewart, line 465: I don't think it's possible for this to be None

wangyi111:
> How much would things slow down if we use the cloud mask instead of the overall image cloud percentage?

depends on the scene, can not tell for sure, but like ~15% maybe.

experiments/download_ssl4eo.py
wangyi111: if args.indices_range is not specified, the script will just match the whole sampled coords

experiments/download_ssl4eo.py
wangyi111: it is possible, see line 251-254:

```
except (ee.EEException, urllib3.exceptions.HTTPError) as e:
        if debug:
            print(e)
        return None, coords
```

adamjstewart: It's possible for patches, but not for save_path

experiments/download_ssl4eo.py
adamjstewart: But indices_range has a default value, it can't be set to None, right?

adamjstewart:
> depends on the scene, can not tell for sure, but like ~15% maybe.

Are you talking about calculating the cloud cover of the entire scene from the pixel qa band? Or calculating the cloud cover of the patch only? I'm talking about the latter. Might make things way slower since we can't do it at a collection level, it needs to be done for every patch on the fly.

adamjstewart (approved):

PR #1225, opened 2023-04-08
Use torch.compile if available

adamjstewart:
See:

* https://lightning.ai/pages/blog/training-compiled-pytorch-2.0-with-pytorch-lightning/
* https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html

This should hopefully make our benchmarking much quicker. Anyone want to benchmark the difference?

@isaaccorley you'll also want to add this to any other scripts you create for the pipeline.

@yichiac @shradhasehgal you'll want to use this to make benchmarking faster.

### DIFF
--- a/evaluate.py
+++ b/evaluate.py
@@ -146,6 +146,9 @@ def main(args: argparse.Namespace) -> None:
     model.freeze()
     model.eval()
 
+    if hasattr(torch, "compile"):
+        model = torch.compile(model)
+
     dm = DATAMODULE(
         seed=args.seed,
         root=args.root,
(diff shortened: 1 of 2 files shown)


isaaccorley:
In the blog you posted, they found that using the arg: `mode="reduce-overhead")` resulted in better performance. Should we set that as well?

adamjstewart:
According to the [torch.compile tutorial](https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html):

> Since our model and data are small, we want to reduce overhead as much as possible, and so we chose `"reduce-overhead"`.

So I'm concerned what might happen for large models and large datasets (like we'll be using for SSL pre-training).

isaaccorley:
Ah I missed that part thanks.

isaaccorley (approved):

PR #1232, opened 2023-04-11
Lightning: add support for v2

adamjstewart:
Forgot to do this in #1178


adamjstewart:
Confirmed that all other upper bounds are correct.

isaaccorley (approved):

PR #1235, opened 2023-04-12
SeCo newer version bug fix

isaaccorley:
I may have a new version of SeCo downloaded, but each patch directory in the version I have contains RGB tifs + directories with individual band tifs. for example the folder `seasonal_contrast_100k/01999` contains the following:

```
20191101T080051_20191101T080824_T36QZH/
20200125T080119_20200125T081343_T36QZH/
20200414T075609_20200414T081030_T36QZH/
20200713T075609_20200713T081050_T36QZH/
20200921T075629_20200921T080225_T36QZH/
20191101T080051_20191101T080824_T36QZH.tif
20200125T080119_20200125T081343_T36QZH.tif
20200414T075609_20200414T081030_T36QZH.tif
20200713T075609_20200713T081050_T36QZH.tif
20200921T075629_20200921T080225_T36QZH.tif
```

I think the current dataset expects there to not be RGB tifs. This simple fix makes it so that it ignores the RGB tifs since we load RGB bands with the individual band files anyway.


adamjstewart:
Can you update the test data?

It could also be that the 100K version has a different layout than the 1M version. I think that was the case with SSL4EO-S12 too.

adamjstewart (approved):

Issue #228, opened 2021-11-08
Add train entrypoint


adamjstewart:
Python has a concept of "entrypoints" that allow you to do something like:
```console
$ torchgeo train conf=...
```
and run `train.py`. We should look into incorporating this. We'll also want to install the `conf` files somewhere and teach TorchGeo how to find them. The tricky part is making both of the following work:
```console
$ python train.py conf=...
$ torchgeo train conf=...
```
since both the training file and the conf files will be in different locations in those two cases.

adamjstewart:
Some docs on how to implement this: https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html

adamjstewart:
More docs:

* [python-packaging](https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html)
* [python `__main__`](https://docs.python.org/3/library/__main__.html)
* [setuptools entry points](https://setuptools.pypa.io/en/latest/userguide/entry_point.html#entry-points)
* [LightningCLI API](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.cli.LightningCLI.html#lightning.pytorch.cli.LightningCLI)
* [LightningCLI Tutorials](https://lightning.ai/docs/pytorch/stable/cli/lightning_cli.html)

PR #1237, opened 2023-04-12
Refactor train.py

isaaccorley:
This PR refactors `train.py` to use `hydra.utils.instantiate` to define the trainer and datamodule inside the config file without having to do any config file magic. This allows a user to define a config like below:

```yaml
trainer:
  _target_: pytorch_lightning.Trainer
  accelerator: gpu
  devices: 1
  min_epochs: 15
  max_epochs: 40

module:
  _target_: torchgeo.trainers.ClassificationTask
  loss: "ce"
  model: "resnet18"
  learning_rate: 1e-3
  weights: null
  in_channels: 3
  num_classes: 45

datamodule:
  _target_: torchgeo.datamodules.RESISC45DataModule
  root: "data/resisc45"
  batch_size: 128
  num_workers: 4
```


adamjstewart:
We talked about this in the meeting today, but we're probably going to try to combine pretrain/train/evaluate/predict into a single main.py that can handle all of the above, and store things like datamodule and trainer classes in the YAML config file.

adamjstewart:
Actually, we could name it `torchgeo/__main__.py` so that users can run `python -m torchgeo --train --datamodule=... --task=...`. This might be sufficient to close #228.

train.py
```diff
@@ -44,32 +43,39 @@
...
-    "naipchesapeake": (SemanticSegmentationTask, NAIPChesapeakeDataModule),
-    "nasa_marine_debris": (ObjectDetectionTask, NASAMarineDebrisDataModule),
-    "potsdam2d": (SemanticSegmentationTask, Potsdam2DDataModule),
-    "resisc45": (ClassificationTask, RESISC45DataModule),
-    "sen12ms": (SemanticSegmentationTask, SEN12MSDataModule),
-    "so2sat": (ClassificationTask, So2SatDataModule),
-    "spacenet1": (SemanticSegmentationTask, SpaceNet1DataModule),
-    "ucmerced": (ClassificationTask, UCMercedDataModule),
-    "vaihingen2d": (SemanticSegmentationTask, Vaihingen2DDataModule),
+DATASET_TO_DATAMODULES_MAPPING: Dict[str, Type[LightningDataModule]] = {
```
adamjstewart, line 49: Can we automatically infer these? They follow a consistent naming scheme.

isaaccorley: How would you infer these? I mentioned using hydra to instantiate from a config. Do you have any other suggestions?

adamjstewart: Can we use the full CamelCase name in the config? I'm fine with using hydra.

isaaccorley: It's possible to use either `torchgeo.trainers.ClassificationTask` or `ClassificationTask` in the config

isaaccorley: The next question is should we make a user only have to define `ClassificationTask` or prefer them to explicitly define the full path like `torchgeo.trainers.ClassificationTask`

adamjstewart: Ideally either one would work. The former because it's short, and the latter because a user may want to define their own trainer/datamodule. But if it's too much code, I would prefer the former for now.

isaaccorley: I'm going to use the latter because it basically will remove code but also not require us to always update the `train.py` script when we add a new datamodule, also allowing a user to define their own trainer/datamodule.

adamjstewart: Why would it require us to update `train.py`? Couldn't we just use `getattr(torchgeo.datamodules, "LandCoverAIDataModule")`?

calebrob6:
I guess we can get rid of evaluate now that we have testing in `train.py`

train.py
isaaccorley: Yes that's right.

adamjstewart:
> ```yaml
> trainer:
>   _target_: pytorch_lightning.Trainer
> ```

It should actually be `lightning.pytorch.Trainer`, we no longer uses pytorch-lightning.

Why don't any of the config files use Trainer?

adamjstewart:
Still need to look into hydra, but I'm wondering if we can remove our omegaconf dep and replace it with hydra? Trying to minimize the deps we have to keep track of.

isaaccorley:
I was having them inherit from conf/defaults.yaml but you're right it's a bit confusing so it's probably better explicitly define it in each config

conf/bigearthnet.yaml
```diff
@@ -1,22 +1,24 @@
...
+datamodule:
+  _target_: torchgeo.datamodules.BigEarthNetDataModule
+  root: "data/bigearthnet"
+  bands: "all"
+  num_classes: ${experiment.module.num_classes}
+  batch_size: 128
+  num_workers: 4
+
 trainer:
+  _target_: lightning.Trainer
```
adamjstewart, line 20:
```suggestion
  _target_: lightning.pytorch.Trainer
```
Some classes are aliased and some aren't. I like to be consistent and explicit.

requirements/min-reqs.old
```diff
@@ -46,6 +46,7 @@ pydocstyle[toml]==6.1.0
 pyupgrade==2.8.0
 
 # tests
+hydra-core==1.1.0
 mypy==0.900
 nbmake==1.3.3
 omegaconf==2.1.0
```
adamjstewart, line 52: Have you found a way to remove the omegaconf dep?

setup.cfg
```diff
@@ -117,6 +117,8 @@ style =
     # pyupgrade 2.8+ required for --py39-plus flag
     pyupgrade>=2.8,<4
 tests =
+    # hydra-core 1.1.0+ required for recursive instantiation
```
adamjstewart, line 120: Are we using recursive instantiation anywhere?

train.py
```diff
@@ -6,70 +6,17 @@
...
-    "nasa_marine_debris": (ObjectDetectionTask, NASAMarineDebrisDataModule),
-    "potsdam2d": (SemanticSegmentationTask, Potsdam2DDataModule),
-    "resisc45": (ClassificationTask, RESISC45DataModule),
-    "sen12ms": (SemanticSegmentationTask, SEN12MSDataModule),
-    "so2sat": (ClassificationTask, So2SatDataModule),
-    "spacenet1": (SemanticSegmentationTask, SpaceNet1DataModule),
-    "ucmerced": (ClassificationTask, UCMercedDataModule),
-    "vaihingen2d": (SemanticSegmentationTask, Vaihingen2DDataModule),
-}
+from torchgeo import datamodules, trainers  # noqa: F401
```
adamjstewart, line 18: Why is this needed?

train.py
```diff
@@ -202,18 +114,20 @@ def main(conf: DictConfig) -> None:
...
-    ######################################
-    # Run experiment
-    ######################################
+    # Train
     trainer.fit(model=task, datamodule=datamodule)
-    trainer.test(ckpt_path="best", datamodule=datamodule)
+
+    # Test
+    if hasattr(task, "test_step") and hasattr(datamodule, "test_dataloader"):
+        trainer.test(ckpt_path="best", datamodule=datamodule)
```
adamjstewart, line 130: Unfortunately, all trainers have a test_step (might be a no-op) and all datamodules have a test_dataloader (might raise an error). I would instead use:
```suggestion
    try:
        trainer.test(ckpt_path="best", datamodule=datamodule)
    except MisconfigurationException:
        pass
```
Can add more exceptions if other errors might be raised

setup.cfg
isaaccorley: Not currently but a user is going to want this. I'll add it to the test config. Recursive instantiation allows you to define the callbacks, logger, augmentations inside the config file itself.

train.py
isaaccorley: We could also check if test_dataset=None as well but yeah I think your solution works best.

train.py
isaaccorley: Since importing torchgeo doesn't import the datamodules, trainers, etc. Hydra will try to instantiate the module and datamodules objects but they won't be defined. So we need to import datamodules and trainers. Simple tradeoff imo.

isaaccorley: Actually this is solved because I have to import MisconfigurationException from torchgeo.datamodules

requirements/min-reqs.old
isaaccorley: Just removed. It can be imported like `from hydra.compose import OmegaConf, DictConfig` instead.

isaaccorley: Had to upgrade from `hydra-core>=1.1.0` to `hydra-core>=1.1.1` for this

tests/conf/trainer.yaml
```diff
@@ -0,0 +1,26 @@
+trainer:
```
isaaccorley, line 1: @adamjstewart Here's an example of recursive instantiation using hydra. This lets you define callbacks, loggers, etc inside the config itself and not having to edit the actual `train.py` script to change these.

train.py
adamjstewart: That won't work either because it could be using `self.dataset` instead of `self.test_dataset`.

adamjstewart (commented):
Like our existing code, there are a ton of discrepancies as to whether we should call a LightningModule a module, model, task, or trainer. I won't comment on those since we never came to a conclusion on #996.

Config files look amazing! Now just need to decide how closely we can align our training script to LightningCLI.

requirements/tests.txt
```diff
@@ -1,7 +1,7 @@
 # tests
+hydra-core>=1.1.1
```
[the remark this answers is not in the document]
isaaccorley: Can you give more context to this?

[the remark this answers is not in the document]
adamjstewart: Files in `requirements` are used for CI, not for installation. We pin all dependencies in these files and manage them with dependabot. You'll notice that all other deps in this file are pinned to the latest version.

train.py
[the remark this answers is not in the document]
isaaccorley: That seems like a pretty large refactor and I think we should do that in a separate PR. This PR currently refactors train.py so that one can more easily use it with all available or custom modules and datamodules.

setup.cfg
adamjstewart, line 120: What's the minimum version that works without recursive instantiation? Users can install a newer version if they want to, but if older versions work for most cases then I think we should set our minimum to that.

adamjstewart:
I think everything looks good to me. Just want to further minimize the hydra version to the lowest version that gets our tests to pass.

adamjstewart (approved):

calebrob6:
Pausing a bit to think here. Just so I understand correctly -- the reason we're bringing in hydra as a dependency is for `instantiate`? And this is to 1.) solve the problem of needing to maintain an explicit list of "name" --> (class, task) in `train.py` which 2.) makes `train.py` a lot more flexible?

adamjstewart:
Correct. This also lets users create their own datamodule and/or trainer and use it with `train.py` without having to edit `train.py`. We're planning on moving `train.py` to `torchgeo/__main__.py` so that it gets installed and users can use `python -m torchgeo fit config_file=...`. This will become even more important when we do that.

calebrob6:
Okay cool! I don't see why `train.py` needs to be moved to `torchgeo/__main__.py` actually, but I think that's a discussion for another day.

adamjstewart:
It relates to #228, just provides an easy way for people to `pip install torchgeo` and user our `train.py` script.

adamjstewart:
Yep, I'm planning on integrating this and adding a torchgeo command in the next release.

PR #1238, opened 2023-04-12
UCMerced: fix image shape bug

isaaccorley:
Previously we performed some preprocessing transforms for each individual sample. In https://github.com/microsoft/torchgeo/pull/992 we changed this to perform all transforms on a batch.

However, some of the UCMerced images vary in size, e.g. 247, 254, 248. This causes an issue when loading batches of the imagery because they aren't the same shape and aren't being resized. This PR adds a default resize transform so that if a user doesn't specify the transform as a kwarg, the dataset will still work with a dataloader.


calebrob6:
literally just ran into this 

> However, some of the UCMerced images vary in size, e.g. 247, 254, 248. 
253

tests/datamodules/test_ucmerced.py
adamjstewart, line 1: This datamodule is already tested by `tests/trainers/test_segmentation.py`, I'm trying to get rid of datamodule tests because they don't actually test anything: https://github.com/microsoft/torchgeo/issues/978

adamjstewart (changes requested):
Can you change the fake data for UC Merced to reflect the fact that the image size isn't consistent? This will let us test things properly.

tests/datamodules/test_ucmerced.py
isaaccorley: I removed it

isaaccorley: Also UCMerced is a classification dataset so it shouldn't be tested in that script I think.

adamjstewart:
Is it worth resizing in the dataset instead of the datamodule?

isaaccorley:
> Is it worth resizing in the dataset instead of the datamodule?

This is essentially resizing in the dataset by passing a default `transforms`. A user can override this and we want to encourage users to use the datamodule since it uses the splits.

adamjstewart:
I guess the benefit of doing it in the dataset is that users who use the dataset without a datamodule won't also have to do this. Are all the images approximately the same size? Looking at our existing code, we have 2 datasets (cyclone, seco) and 0 datamodules that do resizing, unless I miscounted.

isaaccorley:
Yes they are all within a few pixels from 256x256. I'm good either way so I'll leave it up to you, just let me know so I can implement it.

calebrob6:
I like it in the dataset so users don't have to fool around with it. Without it someone (me) is going to do:

```
ds = UCMerced(...)
dl = DataLoader(ds, ...)

for batch in dl:
    ...
```
then get the error about size mismatch, groan, go lookup how to do resize and pass that as a transform, and continue with whatever they were doing.

torchgeo/datasets/ucmerced.py
```diff
@@ -143,6 +144,21 @@ def __init__(
...
+
+        Args:
+            index: index to return
+        Returns:
+            the image
+            the image class label
+
+        .. versionadded:: 0.4.2
+        """
+        img, label = NonGeoClassificationDataset._load_image(self, index)
```
adamjstewart, line 158:
```suggestion
        img, label = super()._load_image(index)
```

torchgeo/datasets/ucmerced.py
```diff
@@ -143,6 +144,21 @@ def __init__(
             is_valid_file=is_in_split,
         )
 
+    def _load_image(self, index: int) -> Tuple[Tensor, Tensor]:
+        """Load a single image and it's class label.
+
+        Args:
+            index: index to return
```
adamjstewart, line 151: Needs a newline

torchgeo/datasets/ucmerced.py
```diff
@@ -143,6 +144,21 @@ def __init__(
...
+        Returns:
+            the image
+            the image class label
+
+        .. versionadded:: 0.4.2
```
adamjstewart, line 156: We don't need to document this, it isn't public and technically doesn't change the API since the base class also had this method.

torchgeo/datasets/ucmerced.py
```diff
@@ -143,6 +144,20 @@ def __init__(
...
 
+    def _load_image(self, index: int) -> Tuple[Tensor, Tensor]:
+        """Load a single image and it's class label.
+
+        Args:
+            index: index to return
+
+        Returns:
+            the image
+            the image class label
```
adamjstewart, line 155:
```suggestion
            the image and class label
```
Python functions only return one thing, in this case a tuple.

isaaccorley:
Something wrong with the pyvista docs

torchgeo/datasets/ucmerced.py
```diff
@@ -143,6 +144,20 @@ def __init__(
...
         )
 
+    def _load_image(self, index: int) -> Tuple[Tensor, Tensor]:
+        """Load a single image and it's class label.
+
+        Args:
+            index: index to return
+
+        Returns:
+            the image
```
adamjstewart, line 154: Remove this line

adamjstewart (commented):
Need to rebase now that we are using Python 3.9+ type hints

torchgeo/datasets/ucmerced.py
```diff
@@ -143,6 +144,19 @@ def __init__(
             is_valid_file=is_in_split,
         )
 
+    def _load_image(self, index: int) -> Tuple[Tensor, Tensor]:
```
adamjstewart, line 147:
```suggestion
    def _load_image(self, index: int) -> tuple[Tensor, Tensor]:
```
We now only support 3.9+

adamjstewart (approved):

PR #1240, opened 2023-04-13
returning the centroid of each cell in the sample dict

estherrolf:
Modifies the sample to additionally return the centroid location for each grid cell in the original label data csvs.


torchgeo/datasets/usavars.py
```diff
@@ -165,6 +165,9 @@ def __getitem__(self, index: int) -> Dict[str, Tensor]:
                 [self.label_dfs[lab].loc[id_][lab] for lab in self.labels]
             ),
             "image": self._load_image(os.path.join(self.root, "uar", tif_file)),
+            "centroid_loc": Tensor(
+                self.label_dfs[self.labels[0]].loc[id_][["lon", "lat"]]
```
adamjstewart, line 169: I feel like lat/lon is more common than lon/lat. Should we call the key centroid_log or centroid?

calebrob6: fiona returns lon/lats (I think this is standard in geospatial libraries)
E.g.

image

When we write coordinates we usually put latitude first though, e.g. 47.661435 N, -122.116502 E.

calebrob6 (approved):

torchgeo/datasets/usavars.py
estherrolf: I'll just return lat and lon separately so that it's as clear as possible - thoughts/objections?

adamjstewart: That's fine too. The current solution is fine as long as we document it.

adamjstewart (approved):

PR #1241, opened 2023-04-13
Add PixelwiseRegressionTask

### LINKED ISSUE #849: a version of the semantic segmentation task for regression
### Summary

semantic segmentation models does support regression if 'classes' are changed to 1 and the 'activation' is changed to 'identity'. it will be useful to have a task created for this in torchgeo.

Also, it would be useful if segmentation models supported more recent architectures (not a torchgeo problem)- out of my depth for me to implement. 

### Rationale

_No response_

### Implementation

_No response_

### Alternatives

_No response_

### Additional information

_No response_

isaaccorley:
This PR adds a `PixelwiseRegressionTask` which can be used for regression on 2D imagery, e.g. height estimation or other continuous per-pixel variables. It's basically a mixture of the current `RegressionTask` and `SemanticSegmentationTask` in that it performs regression but uses smp models e.g. U-Net with a single output channel and L1 or L2 loss for training.

Unless I'm mistaken, we don't have any datasets to actually test this on.

Closes #849


isaaccorley:
I vote for regression because it's basically the `RegressionTask` but with a smp base model instead of a timm model.

torchgeo/trainers/regression.py
```diff
@@ -219,3 +225,129 @@ def configure_optimizers(self) -> Dict[str, Any]:
...
+            raise ValueError(
+                f"Model type '{self.hyperparams['model']}' is not valid. "
+                f"Currently, only supports 'unet', 'deeplabv3+' and 'fcn'."
+            )
+
+        self.loss: nn.Module
+        if self.hyperparams["loss"] == "mse":
+            self.loss = nn.MSELoss()
+        elif self.hyperparams["loss"] == "mae":
+            self.loss = nn.L1Loss()
```
adamjstewart, line 273: If you also define `self.loss` in `RegressionTask` you'll be able to remove all the steps below, right?

torchgeo/trainers/regression.py
```diff
@@ -219,3 +225,129 @@ def configure_optimizers(self) -> Dict[str, Any]:
                 "monitor": "val_loss",
             },
         }
+
+
+class DenseRegressionTask(RegressionTask):
```
adamjstewart, line 230: Is Dense Regression a name that people know this by? I would call it Pixel-Wise Regression personally.

isaaccorley: I think it could go either way.

isaaccorley: I've seen several depth estimation papers refer to it as dense regression but I'm not attached to the name.

adamjstewart: I wonder if we can do the equivalent of a google trends search but for google scholar. I feel like @calebrob6 showed me a plot like that once.

isaaccorley: After googling both `dense regression remote sensing` and `pixelwise regression remote sensing` it seems that more RS papers refer to it as `pixelwise`. I'm biased towards `dense` because I've seen many monocular depth estimation papers for natural images refer to it as `dense`. I think the RS community tends towards `pixelwise` so we will go with that.

torchgeo/trainers/regression.py
isaaccorley: Refactored to reduce code duplication

torchgeo/trainers/regression.py
```diff
@@ -35,8 +36,10 @@ class RegressionTask(LightningModule):  # type: ignore[misc]
         print(timm.list_models())
     """
 
-    def config_task(self) -> None:
-        """Configures the task based on kwargs parameters."""
+    target_key: str = "label"
```
isaaccorley, line 39: This allows to not have to duplicate all the train/val/test steps just to change `label` to `mask`. Let me know if you have any other suggestions.

tests/trainers/test_regression.py
```diff
@@ -198,3 +218,73 @@ def test_predict(self, model_kwargs: dict[Any, Any], fast_dev_run: bool) -> None
...
+        with pytest.raises(ValueError, match=match):
+            RegressionTask(**model_kwargs)
+
+
+class TestPixelwiseRegressionTask:
+    @pytest.mark.parametrize(
+        "name,classname,batch_size,loss",
+        [
+            ("inria", InriaAerialImageLabelingDataModule, 1, "mse"),
+            ("inria", InriaAerialImageLabelingDataModule, 2, "mae"),
```
isaaccorley, line 234: Testing regression on Inria binary [0, 1] masks for now since we don't have a readily available pixelwise regression datamodule.

torchgeo/trainers/regression.py
```diff
@@ -108,13 +130,15 @@ def training_step(self, *args: Any, **kwargs: Any) -> Tensor:
         """
         batch = args[0]
         x = batch["image"]
-        y = batch["label"].view(-1, 1)
+        y = batch[self.target_key]
         y_hat = self(x)
 
-        loss = F.mse_loss(y_hat, y)
+        if y_hat.ndim != y.ndim:
```
isaaccorley, line 136: If `num_outputs=1` the target variable ground truth is missing the necessary channel dim e.g.
- (b,) instead of (b, 1)
- (b, h, w) instead of (b, 1, h, w)

while the output of the models will be:
- (b, 1)
- (b, 1, h, w)

torchgeo/trainers/regression.py
```diff
@@ -108,13 +130,15 @@ def training_step(self, *args: Any, **kwargs: Any) -> Tensor:
...
+            y = y.unsqueeze(dim=1)
 
+        loss = self.loss(y_hat, y.to(torch.float))
         self.log("train_loss", loss)  # logging to TensorBoard
-        self.train_metrics(y_hat, y)
+        self.train_metrics(y_hat, y.to(torch.float))
```
isaaccorley, line 141: Cast to float only for loss and metrics in case the plotting expects a different dtype

adamjstewart:
Looks like pixel-wise regression won the poll.

Can you rebase and use the new hydra-style configs?

calebrob6 (approved):

adamjstewart:
Dimension stuff looks super confusing. Wonder if there's a way to simplify that.

We should consider abstracting the segmentation model stuff into a shared utility.

Will update things to the new style in a separate PR.

isaaccorley:
The dimension lines are essentially the same as the
```
.view(-1, 1)
```

except it only does it if necessary.

PR #1243, opened 2023-04-14
NLCD2016 Tree Canopy

isaaccorley:
This PR adds the NLCD2016 Tree Canopy dataset

See https://www.mrlc.gov/data/nlcd-2016-usfs-tree-canopy-cover-conus


torchgeo/datasets/nlcd.py
```diff
@@ -0,0 +1,86 @@
...
+    Landsat imagery and other available ground and ancillary information.
+
+    If you use this dataset in your research, please cite it using the following format:
+
+    * https://doi.org/10.14358/PERS.78.7.715
+
+    .. note::
+
+       This dataset can be automatically downloaded and converted to geotiffs from the
+       ERDAS IMAGINE .img format using the following bash script:
```
adamjstewart, line 30: Any reason to convert to GeoTIFFs? Should also load fine as .img files (although more slowly).

torchgeo/datasets/nlcd.py
```diff
@@ -0,0 +1,86 @@
...
+
+import matplotlib.pyplot as plt
+
+from .geo import RasterDataset
+
+
+class NLCD2016TreeCanopy(RasterDataset):
+    """National Land Cover Database 2016 (NLCD2016) - Tree Canopy dataset.
+
+    The `National Land Cover Database <https://www.mrlc.gov/>`_ provides 30m tree
```
adamjstewart, line 16: I would link to the tree canopy page, not the NLCD page

torchgeo/datasets/nlcd.py
isaaccorley: I wasn't able to load them without rasterio errors but I'll try again tomorrow

torchgeo/datasets/nlcd.py
calebrob6:
```suggestion
    The `Multi-Resolution Land Characteristics (MRLC) Consortium <https://www.mrlc.gov/>`_ provides 30m tree
```

calebrob6: Here's the specific page for reference https://www.mrlc.gov/data/nlcd-2016-usfs-tree-canopy-cover-conus

calebrob6 (approved):

calebrob6:
Comparison of the IMG format to COG format:
- IMG is uncompressed, just the tree canopy dataset is 20GB on disk. It is 16832104560 pixels that are 1 byte each + overviews :)
- IMG format consists of a .html, .ige, .img, and .img.xml file
- COG is 4.2 GB total, no extra files, lossless compression
- Random windowed reads on the IMG data is ~.6 seconds per 1000
- Random windowed reads on the COG data is 2.1 seconds per 1000

image

adamjstewart:
Surprised img is faster than COGs, I thought COGs were the gold standard.

adamjstewart:
This will need to be rebased once #1244 is merged.

calebrob6:
> Surprised img is faster than COGs, I thought COGs were the gold standard.

I'm guessing the difference is compression related (COG is 5x smaller and 3x slower to read). It is apples to oranges as if these were hosted on a remote server, you could still do windowed reading quickly with a COG.

calebrob6:
Confirmed that the difference is entirely compression related:

image

calebrob6:
(for completeness, because I was curious)

> It is apples to oranges as if these were hosted on a remote server, you could still do windowed reading quickly with a COG.

Not quite actually, you can still do windowed reading from remote files with the Erdas Imagine format, but it is 2x slower than COGs. Also, compression vs. no compression doesn't seem to matter when reading from remote files (it looks like compressed is slightly faster, which makes sense) as the time it takes to transfer the data is going to dominate.

image

torchgeo/datasets/nlcd.py
calebrob6: More info than you ever wanted to know about img vs. cog below

adamjstewart:
We now have a generic `NLCD` dataset, this is likely something we should add to `nlcd.py` instead of making it its own unrelated dataset.

isaaccorley:
Closing because this PR is very out of sync with the main branch. Would be better to just start a fresh one.

PR #1244, opened 2023-04-14
Add National Land Cover Database (NLCD) dataset

nilsleh:
This PR adds the [NLCD dataset](https://www.usgs.gov/centers/eros/science/national-land-cover-database), with selective download and using masks for desired years.

Dataset format:
- single channel `.img` file

Dataset features:
- 16 classes for US
- remapped the class labels to ordinal labels

Example:

Note: 
If the year selection scheme is approved, I would open another PR to change CDL accordingly, so that users don't have to download data for all years.


tests/data/nlcd/data.py
```diff
@@ -0,0 +1,60 @@
...
+SIZE = 32
+
+np.random.seed(0)
+
+dir = "nlcd_2019_land_cover_l48_20210604"
+
+
+def create_file(path: str, dtype: str):
+    """Create the testing file."""
+    profile = {}
```
adamjstewart, line 22: This whole thing could be a dict created at a single time instead of constructing line by line, see the L7 Irish data.py for an example

tests/data/nlcd/nlcd_2019_land_cover_l48_20210604.zip
adamjstewart, line 1: We'll also want to include the uncompressed file if we don't want the eventual trainer tests to decompress this file and add a new file tracked by git.

torchgeo/datasets/nlcd.py
```diff
@@ -0,0 +1,318 @@
...
+    Detailed descriptions of the classes can be found
+    `here <https://www.mrlc.gov/data/legends/national-land-cover-database-class-legend-and-description>`_.
+
+    Dataset format:
+
+    * single channel .img file with integer class labels
+
+    If you use this dataset in your research, please use the corresponding citation:
+
+    * (2001): https://doi.org/10.5066/P9MZGHLF
```
adamjstewart, line 64: No need to put in parentheses I think

adamjstewart (changes requested):
Plot looks great! Also need to add to docs.

docs/api/datasets.rst
adamjstewart, line 1: Also need an entry in `docs/api/geo_datasets.csv`

torchgeo/datasets/nlcd.py
```diff
@@ -0,0 +1,310 @@
...
+    #. Mixed Forest
+    #. Shruf/Scrub
+    #. Grassland/Herbaceous
+    #. Pasture/Hay
+    #. Cultivated Crops
+    #. Woody Wetlands
+    #. Emergent Herbaceous Wetlands
+
+    Detailed descriptions of the classes can be found
+    `on this website <https://www.mrlc.gov/data/legends/national-land-cover-database-class-legend-and-description>`_.
```
adamjstewart, line 49: If you wanted to make "here" work, you would have to replace the trailing _ with __. I still hate rST...

torchgeo/datasets/nlcd.py
```diff
@@ -0,0 +1,310 @@
...
+from .geo import RasterDataset
+from .utils import BoundingBox, download_url, extract_archive
+
+
+class NLCD(RasterDataset):
+    """National Land Cover Database (NLCD) dataset.
+
+    The `NLCD dataset
+    <https://www.usgs.gov/centers/eros/science/national-land-cover-database>`_
+    is a land cover product that covers the United States and Puerto Rico. It is
```
adamjstewart, line 23: The URLs you're using only covers CONUS, not Alaska/Hawaii/Puerto Rico. Interestingly, the 2019 version does not yet have downloads for the latter. Not sure if we should add logic to also download the latter when available or just document that this dataset is only for CONUS until someone actually needs the latter.

torchgeo/datasets/nlcd.py
```diff
@@ -0,0 +1,310 @@
...
+    a joint effort between the United States Geological Survey
+    (`USGS <https://www.usgs.gov/>`_) and the Multi-Resolution Land Characteristics
+    Consortium (`MRLC <https://www.mrlc.gov/>`_) which released the first product
+    in 2001 with new updates every five years since then.
+
+    The dataset contains the following 15 classes:
+
+    #. Open Water
+    #. Pereniial Ice/Snow
```
adamjstewart, line 32:
```suggestion
    #. Perennial Ice/Snow
```

torchgeo/datasets/nlcd.py
```diff
@@ -0,0 +1,310 @@
...
+    #. Developed, Open Space
+    #. Developed, Low Intensity
+    #. Developed, Medium Intensity
+    #. Developed, High Intensity
+    #. Barren Land (Rock/Sand/Clay)
+    #. Deciduous Forest
+    #. Evergreen Forest
+    #. Mixed Forest
+    #. Shruf/Scrub
```
adamjstewart, line 41:
```suggestion
    #. Shrub/Scrub
```

torchgeo/datasets/nlcd.py
```diff
@@ -0,0 +1,310 @@
...
+
+    md5s = {
+        2001: "538166a4d783204764e3df3b221fc4cd",
+        2006: "67454e7874a00294adb9442374d0c309",
+        2011: "ea524c835d173658eeb6fa3c8e6b917b",
+        2016: "452726f6e3bd3f70d8ca2476723d238a",
+        2019: "82851c3f8105763b01c83b4a9e6f3961",
+    }
+
+    ordinal_label_map = {
```
adamjstewart, line 84: Can you modify `ChesapeakeCVPR` to import and use this attribute so we don't have to duplicate it?

torchgeo/datasets/nlcd.py
```diff
@@ -0,0 +1,310 @@
...
+
+        Returns:
+            sample of image/mask and metadata at that index
+
+        Raises:
+            IndexError: if query is not found in the index
+        """
+        sample = super().__getitem__(query)
+
+        # replace mask values with ordinal values
```
adamjstewart, line 181: I wonder how much faster this is than a simple for-loop like we used in L7 Irish/L8 Biome. It's certainly less legible...

nilsleh: I checked and it is actually not faster (current approach 0.011 sec mean across 100 samples, the simple for loop 0.004).

torchgeo/datasets/nlcd.py
nilsleh: Sorry, I am not sure I understand what you mean here.

torchgeo/datasets/nlcd.py
nilsleh: I now added a statement that the current implementation is only for CONUS.

torchgeo/datasets/nlcd.py
adamjstewart: It looks like we aren't currently mapping NLCD? Let's do this in a separate PR.

torchgeo/datasets/nlcd.py
```diff
@@ -0,0 +1,304 @@
...
+            checksum: if True, check the MD5 after downloading files (may be slow)
+
+        Raises:
+            FileNotFoundError: if no files are found in ``root``
+            RuntimeError: if ``download=False`` but dataset is missing or checksum fails
+            AssertionError: if ``year`` is invalid
+        """
+        assert set(years).issubset(
+            self.md5s.keys()
+        ), f"NLCD data product only exists for the following years: {self.md5s.keys()}."
```
adamjstewart, line 157:
```suggestion
        ), f"NLCD data product only exists for the following years: {list(self.md5s.keys())}."
```
Otherwise it looks like:
```
dict_keys([2001, 2006, 2011, 2016, 2019])
```
Might need to wrap onto multiple lines.

torchgeo/datasets/nlcd.py
```diff
@@ -0,0 +1,304 @@
...
+        self.root = root
+        self.download = download
+        self.checksum = checksum
+
+        self._verify()
+
+        super().__init__(root, crs, res, transforms=transforms, cache=cache)
+
+    def __getitem__(self, query: BoundingBox) -> Dict[str, Any]:
+        """Retrieve image/mask and metadata indexed by query.
```
adamjstewart, line 168:
```suggestion
        """Retrieve mask and metadata indexed by query.
```

torchgeo/datasets/nlcd.py
```diff
@@ -0,0 +1,304 @@
...
+            a matplotlib Figure with the rendered sample
+        """
+        mask = sample["mask"].squeeze().numpy()
+        ncols = 1
+
+        cmap: "np.typing.NDArray[np.int_]" = np.zeros((max(self.cmap) + 1, 4), np.int_)
+        for idx, cmap_val in self.cmap.items():
+            cmap[idx, :] = cmap_val
+
+        mask = cmap[mask]
```
adamjstewart, line 277: Can't we just pass the cmap in to imshow directly?

torchgeo/datasets/nlcd.py
```diff
@@ -0,0 +1,304 @@
...
+
+        showing_predictions = "prediction" in sample
+        if showing_predictions:
+            pred = sample["prediction"].squeeze().numpy()
+            pred = cmap[pred]
+            ncols = 2
+
+        fig, axs = plt.subplots(nrows=1, ncols=ncols, figsize=(ncols * 4, 4))
```
adamjstewart, line 285: If you use `squeeze=False`, it will prevent `axs` from changing shape based on the value of `ncols`, making it easier to index and making the below code simpler.

torchgeo/datasets/nlcd.py
```diff
@@ -0,0 +1,305 @@
...
+
+        Returns:
+            a matplotlib Figure with the rendered sample
+        """
+        mask = sample["mask"].squeeze().numpy()
+        ncols = 1
+
+        cmap: "np.typing.NDArray[np.int_]" = np.array(
+            [self.cmap[i] for i in range(len(self.cmap))]
+        )
```
adamjstewart, line 276: What I mean is can be use something like [ListedColormap](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.ListedColormap.html) and use `imshow(..., cmap=cmap)`

torchgeo/datasets/nlcd.py
```diff
@@ -0,0 +1,303 @@
...
+    is a land cover product that covers the United States and Puerto Rico. The current
+    implementation supports maps for the continental United States only. The product is
+    a joint effort between the United States Geological Survey
+    (`USGS <https://www.usgs.gov/>`_) and the Multi-Resolution Land Characteristics
+    Consortium (`MRLC <https://www.mrlc.gov/>`_) which released the first product
+    in 2001 with new updates every five years since then.
+
+    The dataset contains the following 15 classes:
+
+    #. Open Water
```
adamjstewart, line 32: It looks like the mask returned by `__getitem__` will start with 0, 1, 2, whereas this list will start with 1, 2, 3. Let's just manually label it 0, 1, 2 so that it matches and there is no confusion.

adamjstewart: Or is 0 a fill class? Maybe we should mention that there are actually 17 classes (0 being fill).

nilsleh: Yes 0 is a background class, I believe, will adapt the docs.

torchgeo/datasets/nlcd.py
```diff
@@ -0,0 +1,304 @@
...
+            query: (minx, maxx, miny, maxy, mint, maxt) coordinates to index
+
+        Returns:
+            sample of mask and metadata at that index
+
+        Raises:
+            IndexError: if query is not found in the index
+        """
+        sample = super().__getitem__(query)
+        print(sample["mask"].min())
```
adamjstewart, line 183: Can you remove this print statement?

nilsleh: uups, sorry!

torchgeo/datasets/nlcd.py
```diff
@@ -0,0 +1,304 @@
...
+    is a land cover product that covers the United States and Puerto Rico. The current
+    implementation supports maps for the continental United States only. The product is
+    a joint effort between the United States Geological Survey
+    (`USGS <https://www.usgs.gov/>`_) and the Multi-Resolution Land Characteristics
+    Consortium (`MRLC <https://www.mrlc.gov/>`_) which released the first product
+    in 2001 with new updates every five years since then.
+
+    The dataset contains the following 17 classes:
+
+    0. Background
```
adamjstewart, line 32: Genius!

torchgeo/datasets/nlcd.py
```diff
@@ -0,0 +1,304 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""NLCD dataset."""
+
+import os
+from typing import Any, Callable, Dict, List, Optional
```
adamjstewart, line 7: Can you run:
```console
$ pyupgrade --py39-plus torchgeo/datasets/nlcd.py
```
You'll also need to remove the unused imports:
```suggestion
from typing import Any, Callable, Optional
```
Yay for Python 3.9!

adamjstewart:
Missing test coverage for some lines: https://app.codecov.io/gh/microsoft/torchgeo/pull/1244/blob/torchgeo/datasets/nlcd.py

nilsleh:
> Missing test coverage for some lines: https://app.codecov.io/gh/microsoft/torchgeo/pull/1244/blob/torchgeo/datasets/nlcd.py

not sure why it is not working, I thought the following test would cover it but doesn't:
```
    def test_already_downloaded(self, tmp_path: Path) -> None:
        pathname = os.path.join(
            "tests", "data", "nlcd", "nlcd_2019_land_cover_l48_20210604.zip"
        )
        root = str(tmp_path)
        shutil.copy(pathname, root)
        NLCD(root, years=[2019])
```

nilsleh:
found my mistake, and tested again with downloading the real data from scratch and retrieving samples.

adamjstewart (approved):

PR #1246, opened 2023-04-14
Drop Python 3.8 support

adamjstewart:
[NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table) has officially deprecated Python 3.8 support as of today. In the past we've been hesitant to follow NEP 29 because Google Colab was still running Python 3.7, but they've since upgraded to 3.9 already, so that isn't an issue anymore. PyTorch ostensibly follows NEP 29, and many of our other dependencies will soon drop support for 3.8 in their next release, complicating testing. Also, 3.9 and 3.10 introduce native support for type hints, meaning most of our `from typing import *` lines will be removed.

### DIFF
--- a/docs/user/contributing.rst
+++ b/docs/user/contributing.rst
@@ -58,14 +58,14 @@ For example, if you add a new dataset in ``torchgeo/datasets/foo.py``, you'll ne
 
    $ pytest --cov=torchgeo/datasets --cov-report=term-missing tests/datasets/test_foo.py
    ========================= test session starts =========================
-   platform darwin -- Python 3.8.11, pytest-6.2.4, py-1.9.0, pluggy-0.13.0
+   platform darwin -- Python 3.10.11, pytest-6.2.4, py-1.9.0, pluggy-0.13.0
    rootdir: ~/torchgeo, configfile: pyproject.toml
    plugins: mock-1.11.1, anyio-3.2.1, cov-2.8.1, nbmake-0.5
    collected 7 items
 
    tests/datasets/test_foo.py .......                              [100%]
 
-   ---------- coverage: platform darwin, python 3.8.11-final-0 -----------
+   ---------- coverage: platform darwin, python 3.10.11-final-0 -----------
    Name                                      Stmts   Miss  Cover   Missing
    -----------------------------------------------------------------------
    torchgeo/datasets/__init__.py                26      0   100%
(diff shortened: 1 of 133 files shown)


adamjstewart:
Discovered a bug in Sphinx in the process of upgrading: https://github.com/sphinx-doc/sphinx/issues/11327

torchgeo/datasets/utils.py
```diff
@@ -3,6 +3,9 @@
 
 """Common dataset utilities."""
 
+# https://github.com/sphinx-doc/sphinx/issues/11327
+from __future__ import annotations
```
adamjstewart, line 7: Had to use this hack to get our docs to build properly. This file gives a preview of what type hints will look like in Python 3.10 when `Optional` and `Union` become obsolete.

docs/user/contributing.rst
```diff
@@ -58,14 +58,14 @@ For example, if you add a new dataset in ``torchgeo/datasets/foo.py``, you'll ne
...
-   platform darwin -- Python 3.8.11, pytest-6.2.4, py-1.9.0, pluggy-0.13.0
+   platform darwin -- Python 3.10.11, pytest-6.2.4, py-1.9.0, pluggy-0.13.0
    rootdir: ~/torchgeo, configfile: pyproject.toml
    plugins: mock-1.11.1, anyio-3.2.1, cov-2.8.1, nbmake-0.5
    collected 7 items
 
    tests/datasets/test_foo.py .......                              [100%]
 
-   ---------- coverage: platform darwin, python 3.8.11-final-0 -----------
+   ---------- coverage: platform darwin, python 3.10.11-final-0 -----------
```
calebrob6, line 68: Is this a find and replace artifact?

torchgeo/datasets/utils.py
calebrob6: Looks cool!

calebrob6 (dismissed):

calebrob6:
Lol 130 files! I didn't check to see if you missed any "Tuple" --> "tuple", etc.

calebrob6:
image

docs/user/contributing.rst
adamjstewart: Nah, just noticed it while grepping around and figured we should pretend it was run on a supported version of Python.

adamjstewart:
> Lol 130 files! I didn't check to see if you missed any "Tuple" --> "tuple", etc.

pyupgrade did, this was mostly auto-generated. The only thing I had to do manually was non-Python files and removing imports we no longer use.

calebrob6 (approved):

PR #1252, opened 2023-04-15
Add SimCLR trainer

isaaccorley:
This PR adds a SimCLR trainer which uses implementations from the `lightly` package

Reboot of #1195


requirements/required.txt
adamjstewart, line 1: Also need to update `min-reqs.old`

setup.cfg
```diff
@@ -32,6 +32,8 @@ install_requires =
     kornia>=0.6.5,<0.7
     # lightning 1.8+ is first release
     lightning>=1.8,<3
+    # lightly 1.4.1+ required for fastsiam
```
adamjstewart, line 35: We aren't using fastsiam

adamjstewart (commented):
Messaged on Slack, need to update to new style.

torchgeo/trainers/simclr.py
```diff
@@ -0,0 +1,242 @@
...
+            augmentations: Data augmentation.
+        """
+        super().__init__()
+
+        self.save_hyperparameters(ignore=["augmentations"])
+        self.augmentations = augmentations
+
+        # Create backbone
+        self.backbone = timm.create_model(
+            model, in_chans=in_channels, pretrained=weights is True
```
isaaccorley, line 137: Should we set num_classes=0 which would then return the backbone features before the classification layer?

torchgeo/trainers/simclr.py
```diff
@@ -0,0 +1,242 @@
...
+            if isinstance(weights, WeightsEnum):
+                state_dict = weights.get_state_dict(progress=True)
+            elif os.path.exists(weights):
+                _, state_dict = utils.extract_backbone(weights)
+            else:
+                state_dict = get_weight(weights).get_state_dict(progress=True)
+            self.backbone = utils.load_state_dict(self.backbone, state_dict)
+
+        # Create projection head
+        input_dim = list(self.backbone.named_children())[-1][1].out_features
```
isaaccorley, line 151: Most timm models have a `num_features` attr. Should we prefer this over `list(self.backbone.named_children())[-1][1].out_features`?

torchgeo/trainers/simclr.py
```diff
@@ -0,0 +1,242 @@
...
+            hidden_dim = input_dim
+        if output_dim is None:
+            output_dim = input_dim
+
+        self.projection_head = SimCLRProjectionHead(
+            input_dim, hidden_dim, output_dim, layers
+        )
+
+        # Define loss function
+        self.criterion = NTXentLoss(temperature, memory_bank_size, gather_distributed)
```
isaaccorley, line 162: v1 doesn't use a memory bank.  We should check if a user set `version=1` and set `memory_bank_size` to None

torchgeo/trainers/simclr.py
```diff
@@ -0,0 +1,242 @@
...
+        model: str = "resnet50",
+        weights: Optional[Union[WeightsEnum, str, bool]] = None,
+        in_channels: int = 3,
+        version: int = 2,
+        layers: int = 3,
+        hidden_dim: Optional[int] = None,
+        output_dim: Optional[int] = None,
+        lr: float = 4.8,
+        weight_decay: float = 1e-4,
+        max_epochs: int = 100,
```
isaaccorley, line 101: I have a feeling users will forget to set to explicitly set this parameter and will only set the number of epochs in the trainer. I'm curious if we can grab it from the trainer itself and pass it in like `self.trainer.max_epochs`

torchgeo/trainers/simclr.py
adamjstewart: Is that what that does? I can't even find documentation for anything in timm. Also not sure if I know TF well enough to determine whether or not they replace the last layer or not.

torchgeo/trainers/simclr.py
adamjstewart: I replaced this because our fask classification model doesn't have this attr, but we could add it. Why didn't we go with `num_features` in the weight loading PR? If this long complicated thing is more reliable we should go with that.

torchgeo/trainers/simclr.py
adamjstewart: I guess it depends on how flexible we want to be. We could force layers, weight_decay, and memory_bank_size to match the version, although this limits flexibility if you want to experiment with one feature but not another. At the moment, we aren't actually using version for anything, but we could.

torchgeo/trainers/simclr.py
adamjstewart: You're a genius: https://github.com/Lightning-AI/lightning/discussions/6723

torchgeo/trainers/simclr.py
isaaccorley: I don't think it's more reliable because it probably makes assumptions about the layer you indexed which may not be the case for other timm models

torchgeo/trainers/simclr.py
isaaccorley: We should at least put a note in the docstring. My concern is based on user blindly assume changing `version=1` would literally mean that they are using SimCLR v1 when in fact they still need to change the `weight decay` and `memory_bank` args which seems to defeat the purpose of even having a `version` arg.

adamjstewart: Should we compromise on a warning message?

isaaccorley: I'm good with that!

torchgeo/trainers/simclr.py
adamjstewart: Any idea if SimCLR does this too? I'll keep digging

isaaccorley: What I'm saying is you NEED to set num_classes=0 otherwise num_classes defaults to 1000 for imagenet and will return classifier logits. By setting num_classes=0 it will return a model without the classifier head which is what we want.

adamjstewart: Lightly example also does this. Let me make the change.

adamjstewart:
Lot of warnings and errors when I use both `num_classes=0` and `num_features`...

adamjstewart:
Yep, timm uses `nn.Identity` when `num_classes = 0`.

adamjstewart:
Oddly it works fine on the command line:
```pycon
$ python
Python 3.10.10 (main, Mar 31 2023, 13:44:49) [Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
>>> import timm
>>> m = timm.create_model('resnet18', num_classes=0)
>>> m.num_features
512
```

adamjstewart:
Kinda stuck here, could use some help.

isaaccorley:
I'll take a look at it

isaaccorley (commented):
This LGTM

torchgeo/trainers/simclr.py
```diff
@@ -0,0 +1,259 @@
...
+
+        # TODO
+        # v1+: add global batch norm
+        # v2: add selective kernels, channel-wise attention mechanism
+
+    def forward(self, batch: Tensor) -> Tensor:
+        """Forward pass of the model.
+
+        Args:
+            batch: Mini-batch of images.
```
nilsleh, line 189: In the `training_step` "batch" refers to the output of the dataloader and in `forward` the argument batch just refers to the input images to the model as it is a tensor. Maybe forward should have a different argument name to differentiate it?

adamjstewart: What name would you suggest? Image?

adamjstewart: Went with "x" to match what we put in and h/z that we output.

torchgeo/trainers/simclr.py
```diff
@@ -0,0 +1,259 @@
...
+        layers.append((hidden_dim, output_dim, nn.BatchNorm1d(output_dim), None))
+        super().__init__(layers)
+
+
+class SimCLRTask(LightningModule):  # type: ignore[misc]
+    """SimCLR: a simple framework for contrastive learning of visual representations.
+
+    Reference implementation:
+
+    * https://github.com/google-research/simclr
```
calebrob6, line 80: We are referencing lightly, not the OG implementation. Can we change this to reflect that (maybe "Original implementation -- google-research/simclr")

torchgeo/trainers/simclr.py
```diff
@@ -0,0 +1,259 @@
...
+        temperature: float = 0.07,
+        memory_bank_size: int = 64000,
+        gather_distributed: bool = False,
+        augmentations: nn.Module = AUG,
+    ) -> None:
+        """Initialize a new SimCLRTask instance.
+
+        Args:
+            model: Name of the timm model to use.
+            weights: Either a weight enum, the string representation of a weight enum,
```
calebrob6, line 110:
```suggestion
            weights: Initial model weights. Either a weight enum, the string representation of a weight enum,
```

torchgeo/trainers/simclr.py
```diff
@@ -0,0 +1,259 @@
...
+        Returns:
+            Optimizer and learning rate scheduler.
+        """
+        # Original paper uses LARS optimizer, but this is not defined in PyTorch
+        optimizer = Adam(
+            self.parameters(),
+            lr=self.hparams["lr"],
+            weight_decay=self.hparams["weight_decay"],
+        )
+        lr_scheduler = CosineAnnealingLR(
```
calebrob6, line 256: Why is `T_max` the total number of epochs? (I've seen this elsewhere, just don't know why it is)

calebrob6 (changes requested):

calebrob6:
Has anyone tried to train a model with this?

torchgeo/trainers/simclr.py
adamjstewart: By "reference implementation" I mean "the implementation of a standard to be used as a definitive interpretation for the requirements in that standard", not "thing I looked at while writing this". To be fair, I did reference the reference implementation to find all the hyperparams and fix the incorrect lightly implementation: https://github.com/lightly-ai/lightly/pull/1150

torchgeo/trainers/simclr.py
adamjstewart: See the mathematical definition [here](https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.CosineAnnealingLR.html). $T_{cur}$ is the number of epochs since the start (restarts are not implemented). Once $T_{cur}$ reaches $T_{max}$, the right half of the equation cancels out and the learning rate $\eta_t$ becomes $\eta_{min}$, which is 0.

adamjstewart:
> Has anyone tried to train a model with this? I'd love to at least see a tensorboard loss curve that's going down :)

Our cluster is down at the moment 😢

torchgeo/trainers/simclr.py
calebrob6: I know what it is, I don't know why it defaults to the total number of epochs (i.e. one cycle through the cosine)

adamjstewart: It's only a half-cycle ($[0, \pi]$). Convergence isn't guaranteed unless learning rate decays to zero. If you did a full-cycle, learning rate would go back to the initial learning rate. You could cycle multiple times and still end up at 0, but I think that's the point of warm restarts.

calebrob6:
Do you have a config file that I can use to test it?

adamjstewart:
Nope. The default hyperparams should work. Just throw your favorite dataset in there.

adamjstewart:
Updating config files to new train.py format...

calebrob6:
Running this with `train.py` results in:

```
RuntimeError: Early stopping conditioned on metric `val_loss` which is not available. Pass in or modify your `EarlyStopping` callback to use any of the following: `train_loss`
Epoch 0: 100%|██████████| 198/198 [01:17<00:00,  2.54it/s, loss=3.77, v_num=0_1]
```

calebrob6:
Also are both the datamodule _and_ SimCLR augmentations going to be applied here?

calebrob6:
I'm testing this with RESISC45 where I commented out the augmentations in the existing DataModule and the following config:

```
module:
  _target_: torchgeo.trainers.SimCLRTask
  model: "resnet50"
  lr: 0.15
  weights: False
  in_channels: 3

datamodule:
  _target_: torchgeo.datamodules.RESISC45DataModule
  root: "data/resisc45"
  batch_size: 128
  num_workers: 8

trainer:
  _target_: lightning.pytorch.Trainer
  accelerator: gpu
  devices:
    - 6
  min_epochs: 15
  max_epochs: 200

program:
  seed: 0
  output_dir: output
  data_dir: data
  log_dir: logs/ssl_tests/
  overwrite: True
```

adamjstewart:
> Also are both the datamodule _and_ SimCLR augmentations going to be applied here?

Yes, augmentations are tightly coupled to SSL methods since they need to be applied on different _views_ of the same image. I'm mostly envisioning SSL trainers being used with SSL datasets like SeCo and SSL4EO. These datamodules don't contain any augs other than normalization. This is the first trainer where augs can be passed in as parameters. We could do the same for all trainers and datamodules to make this more flexible.

calebrob6:
Added logging for the average l2 normalized stdev of the feature dimension as a way to monitor mode collapse during training. See 4.1 of https://arxiv.org/pdf/2011.10566.pdf

adamjstewart:
Is that something we want to add to all SSL trainers?

calebrob6:
Maybe, but it isn't important right now because I haven't been convinced that the BYOL one actually works.

For this one, training is in progress -- I changed the above config to make min_epochs much larger so EarlyStopping wouldn't kick in. Interestingly it looks like after EarlyStopping tries to stop, logging changes to every batch...

image

calebrob6:
Hmmm this run didn't work

image

adamjstewart:
I think the default hyperparameters are what was used in the paper on ImageNet

calebrob6 (dismissed):

torchgeo/trainers/simclr.py
```diff
@@ -0,0 +1,272 @@
...
+        version: int = 2,
+        layers: int = 3,
+        hidden_dim: Optional[int] = None,
+        output_dim: Optional[int] = None,
+        lr: float = 4.8,
+        weight_decay: float = 1e-4,
+        temperature: float = 0.07,
+        memory_bank_size: int = 64000,
+        gather_distributed: bool = False,
+        augmentations: nn.Module = AUG,
```
adamjstewart, line 105: Not saying we shouldn't make this a parameter, but the docs for this look rough...

adamjstewart:
@calebrob6 I added linear warmup (this is used in the paper but missing from our implementation). This might improve learning, assuming I implemented it correctly...

adamjstewart:
> Is that something we want to add to all SSL trainers?

I didn't add this to MoCo, but I can if we decide this should be added to all SSL trainers. If we do, we can add a utility function for it.

adamjstewart:
Really sick of https://github.com/codecov/codecov-action/issues/837

calebrob6:
I tried this trainer with multigpu and got `ValueError: bad value(s) in fds_to_keep`, however the MultiLabel trainer worked fine with the same DataModule.

adamjstewart:
> I tried this trainer with multigpu and got `ValueError: bad value(s) in fds_to_keep`, however the MultiLabel trainer worked fine with the same DataModule.

With or without `gather_distributed`?

calebrob6:
I don't touch `gather_distributed` as I would expect this to work by specifying multiple GPUs in the trainer.

adamjstewart:
What's the stack trace for the error? Only thing I can do is search repos for that error message.

adamjstewart:
Our node is _still_ down. Been going back and forth with the sys admins for days 😭

torchgeo/trainers/simclr.py
```diff
@@ -0,0 +1,274 @@
...
+            x1 = self.augmentations(x1)
+            x2 = self.augmentations(x2)
+
+        z1, h1 = self(x1)
+        z2, h2 = self(x2)
+
+        loss = self.criterion(z1, z2)
+
+        # Calculate the mean normalized standard deviation over features dimensions.
+        # If this is << 1 / sqrt(h1.shape[1]), then the model is not learning anything.
```
adamjstewart, line 227: @calebrob6 do you want this in every SSL trainer or no SSL trainers? Want to be consistent.

calebrob6: It is a good thing to monitor in training per the SimSiam paper so we shouldn't remove it just for convenience, however it also isn't urgent to add it to other trainers

calebrob6 (approved):

torchgeo/trainers/simclr.py
adamjstewart: I'll add it to MoCo, should only take a minute.

PR #1253, opened 2023-04-17
Add SustainBench Crop Yield Dataset

nilsleh:
This PR adds the [Sustainbench Crop Yield Dataset](https://sustainlab-group.github.io/sustainbench/docs/datasets/sdg2/crop_yield.html) as a dataset with a corresponding datamodule.

Dataset Format:
- .npz file of stacked samples

Dataset Features:
- MODIS band histograms across time with regression target of soybean yield in metric tonnes

Questions:
- the dataloading from the .nzp files seems quiet slow, as it always has to index the .npz files but I am not sure how to speed it up

Example:


torchgeo/datamodules/sustainbench_crop_yield_prediction.py
```diff
@@ -0,0 +1,53 @@
...
+    ) -> None:
+        """Initialize a new SustainbenchCropYieldDataModule instance.
+
+        Args:
+            batch_size: Size of each mini-batch.
+            num_workers: Number of workers for parallel data loading.
+            **kwargs: Additional keyword arguments passed to
+                :class:`~torchgeo.datasets.SustainBenchCropYieldPrediction`.
+
+        .. versionadded:: 0.5
```
adamjstewart, line 26: This should be in the class docstring. It's not just `__init__` that was added, it's the entire class

docs/api/datamodules.rst
```diff
@@ -125,6 +125,11 @@ SSL4EO
 
 .. autoclass:: SSL4EOS12DataModule
 
+SustainbenchCropYieldDataModule
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
adamjstewart, line 129:
```suggestion
SustainBench Crop Yield
^^^^^^^^^^^^^^^^^^^^^^^
```

torchgeo/datamodules/sustainbench_crop_yield_prediction.py
```diff
@@ -0,0 +1,54 @@
...
+        super().__init__(
+            SustainBenchCropYieldPrediction, batch_size, num_workers, **kwargs
+        )
+
+    def setup(self, stage: str) -> None:
+        """Set up datasets.
+
+        Called at the beginning of fit, validate, test, or predict. During distributed
+        training, this method is called from every process across all the nodes. Setting
+        state here is recommended.
```
adamjstewart, line 38: I usually drop this paragraph from datamodules because it's documented in the base class.

torchgeo/datasets/sustainbench_crop_yield_prediction.py
adamjstewart, line 1: Would it make sense to add all SustainBench datasets to a single file, or should they be in separate files?

adamjstewart (changes requested):
Lots of inconsistencies in naming. Need to decide between SustainBenchCropYield and SustainBenchCropYieldPrediction (I prefer the former because it's shorter). Also it should be SustainBench, not Sustainbench.

torchgeo/datasets/sustainbench_crop_yield_prediction.py
nilsleh: Not sure, there are 15 SustainBench datasets in total so one file would probably grow very large?

adamjstewart: I guess it depends on if they follow a consistent file format like SpaceNet or not. If they can't share a base class, then there's no reason to group them

adamjstewart (commented):
Still a bunch of directories and filenames containing "prediction"

torchgeo/datasets/sustainbench_crop_yield_prediction.py
nilsleh: the tasks are all different (some have overlap) but they cover segmentation, classification, regression, ... https://sustainlab-group.github.io/sustainbench/about/

torchgeo/datamodules/sustainbench_crop_yield.py
```diff
@@ -0,0 +1,48 @@
...
+class SustainBenchCropYieldDataModule(NonGeoDataModule):
+    """LightningDataModule for SustainBench Crop Yield dataset.
+
+    .. versionadded:: 0.5
+    """
+
+    def __init__(
+        self, batch_size: int = 32, num_workers: int = 0, **kwargs: Any
+    ) -> None:
+        """Initialize a new DataModule instance.
```
adamjstewart, line 21:
```suggestion
        """Initialize a new SustainBenchCropYieldDataModule instance.
```

torchgeo/datamodules/sustainbench_crop_yield.py
```diff
@@ -0,0 +1,48 @@
...
+        super().__init__(SustainBenchCropYield, batch_size, num_workers, **kwargs)
+
+    def setup(self, stage: str) -> None:
+        """Set up datasets.
+
+        Args:
+            stage: Either 'fit', 'validate', 'test', or 'predict'.
+        """
+        if stage in ["fit"]:
+            self.train_dataset = self.dataset_class(  # type: ignore[call-arg]
```
adamjstewart, line 38: Let's use `SustainBenchCropYield` here to avoid the type ignore

torchgeo/datasets/sustainbench_crop_yield.py
```diff
@@ -0,0 +1,274 @@
...
+            split: one of "train", "dev", or "test"
+            countries: which countries to include in the dataset
+            transforms: a function/transform that takes an input sample
+                and returns a transformed version
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 after downloading files (may be slow)
+
+        Raises:
+            AssertionError: if ``countries`` contains invalid countries
+            FileNotFoundError: if no files are found in ``root``
```
adamjstewart, line 79: Never raised

torchgeo/datasets/sustainbench_crop_yield.py
```diff
@@ -0,0 +1,274 @@
...
+            root: root directory where dataset can be found
+            split: one of "train", "dev", or "test"
+            countries: which countries to include in the dataset
+            transforms: a function/transform that takes an input sample
+                and returns a transformed version
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 after downloading files (may be slow)
+
+        Raises:
+            AssertionError: if ``countries`` contains invalid countries
```
adamjstewart, line 78: Or if split is invalid

adamjstewart (dismissed):

PR #1258, opened 2023-04-18
fix hardcoding of batch size in datamodules

isaaccorley:


adamjstewart:
Ah, now I remember why I did things this way. If you look, you'll see that all of these data modules pass `batch_size` to `_NRandomCrop`. This transform takes in a single large image and crops a number of smaller patches from it. This helps handle images that are too large to pass into a model, and images of varying size. So the number of small image patches that goes into the model is still `batch_size`. If you make this change, it will instead be `batch_size` x `batch_size`.

isaaccorley:
Ahh I see it now. Thanks for clarifying.

adamjstewart:
I can't guarantee that it's perfect, and may not make sense at all during validation and testing, but it allowed me to get the tests to pass. We likely need a better approach, similar to GridGeoSampler. I tried something but it ran into memory issues so I dropped it during the big refactor PR.

PR #1259, opened 2023-04-18
Add ChaBuD Dataset

isaaccorley:
This PR adds the [Change detection for Burned area Delineation (ChaBuD) dataset](https://huggingface.co/spaces/competitions/ChaBuD-ECML-PKDD2023) which is being used as one of the discovery challenges for the [ECML-PKDD2023 conference](https://2023.ecmlpkdd.org/submissions/discovery-challenge/challenges/)

The dataset contains pairs of pre/post disaster multispectral Sentinel-2 imagery (12 bands) 512x512 px with binary segmentation masks.

Sample plots (some are clear and several samples in the dataset are incredibly tiny):

chabud1
chabud2
chabud3
chabud4
chabud5
chabud6

@lccol @DarthReca


tests/datasets/test_chabud.py
```diff
@@ -0,0 +1,114 @@
...
+import pytest
+import torch
+import torch.nn as nn
+from _pytest.fixtures import SubRequest
+from _pytest.monkeypatch import MonkeyPatch
+
+import torchgeo.datasets.utils
+from torchgeo.datasets import ChaBuD
+
+pytest.importorskip("h5py", minversion="2.6")
```
adamjstewart, line 20: Technically our min version is 3 now. I've been forgetting to update these, will submit a PR to do this.

tests/datasets/test_chabud.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def download_url(url: str, root: str, filename: str, *args: str, **kwargs: str) -> None:
+    output = shutil.copy(url, os.path.join(root, filename))
+    print(output)
```
adamjstewart, line 25: Can remove print statement

torchgeo/datamodules/chabud.py
```diff
@@ -0,0 +1,83 @@
...
+        _min = self.min[band_indices]
+        _max = self.max[band_indices]
+
+        # Change detection, 2 images from different times
+        _min = repeat(_min, "c -> (t c)", t=2)
+        _max = repeat(_max, "c -> (t c)", t=2)
+
+        self.aug = AugmentationSequential(
+            K.Normalize(mean=_min, std=_max - _min), data_keys=["image", "mask"]
+        )
```
adamjstewart, line 70: Could instead set `self.mean = _min` and `self.std = _max - _min` and then you won't have to set `self.aug`

torchgeo/datamodules/chabud.py
```diff
@@ -0,0 +1,83 @@
...
+            batch_size: Size of each mini-batch.
+            num_workers: Number of workers for parallel data loading.
+            **kwargs: Additional keyword arguments passed to
+                :class:`~torchgeo.datasets.ChaBuD`.
+        """
+        super().__init__(ChaBuD, batch_size, num_workers, **kwargs)
+
+        self.bands = kwargs.get("bands", ChaBuD.rgb_bands)
+        band_indices = [ChaBuD.all_bands.index(b) for b in self.bands]
+        _min = self.min[band_indices]
```
adamjstewart, line 61: According to [PEP 8](https://peps.python.org/pep-0008/#descriptive-naming-styles) you should use a trailing underscore instead of a leading underscore to avoid conflicts with builtin functions. Also, you should use the same thing for the class attributes.

torchgeo/datamodules/chabud.py
```diff
@@ -0,0 +1,83 @@
...
+
+        Args:
+            stage: Either 'fit', 'validate', 'test', or 'predict'.
+        """
+        if stage in ["fit", "validate"]:
+            self.train_dataset = ChaBuD(split="train", **self.kwargs)
+            self.val_dataset = ChaBuD(split="val", **self.kwargs)
+        if stage in ["test"]:
+            # No test set currently released
+            self.test_dataset = ChaBuD(split="val", **self.kwargs)
```
adamjstewart, line 83: Same comment as other PR, I would either remove the test stage or split val

torchgeo/datasets/chabud.py
```diff
@@ -0,0 +1,303 @@
...
+from .geo import NonGeoDataset
+from .utils import download_url, percentile_normalization
+
+
+class ChaBuD(NonGeoDataset):
+    """ChaBuD dataset.
+
+    The `ChaBuD <https://huggingface.co/spaces/competitions/ChaBuD-ECML-PKDD2023>`__
+    dataset is a dataset for Change detection for Burned area Delineation and is used
+    as for the ChaBuD ECML-PKDD 2023 Discovery Challenge.
```
adamjstewart, line 23:
```suggestion
    for the ChaBuD ECML-PKDD 2023 Discovery Challenge.
```

torchgeo/datasets/chabud.py
```diff
@@ -0,0 +1,303 @@
...
+            split: one of "train" or "val"
+            bands: the subset of bands to load
+            transforms: a function/transform that takes input sample and its target as
+                entry and returns a transformed version
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+
+        Raises:
+            AssertionError: if ``split`` argument is invalid
+            AssertionError: if ``bands`` argument is invalid
```
adamjstewart, line 96:
```suggestion
            AssertionError: if ``split`` or ``bands`` arguments are invalid
```

torchgeo/datasets/chabud.py
```diff
@@ -0,0 +1,303 @@
...
+        return tensor
+
+    def _load_target(self, index: int) -> Tensor:
+        """Load the target mask for a single image.
+
+        Args:
+            index: index to return
+
+        Returns:
+            the target mask and label for each mask
```
adamjstewart, line 201: Is it not just the mask?

calebrob6 (approved):

isaaccorley:
@adamjstewart Looks like the minimum test is the only thing failing here due to the warning:

`ignore:The .*dataloader.* does not have many workers which may be a bottleneck:lightning.fabric.utilities.warnings.PossibleUserWarning:lightning`.

adamjstewart:
Looking into it

torchgeo/datasets/chabud.py
```diff
@@ -0,0 +1,303 @@
...
+        Raises:
+            RuntimeError: if ``download=False`` but dataset is missing or checksum fails
+        """
+        # Check if the files already exist
+        if os.path.exists(self.filepath):
+            return
+
+        # Check if the user requested to download the dataset
+        if not self.download:
+            raise RuntimeError(
```
adamjstewart, line 225: Can you use the new `DatasetNotFoundError` here? See #1714 for details.

docs/api/non_geo_datasets.csv
```diff
@@ -3,6 +3,7 @@ Dataset,Task,Source,# Samples,# Classes,Size (px),Resolution (m),Bands
 `Benin Cashew Plantations`_,S,Airbus Pléiades,70,6,"1,122x1,186",10,MSI
 `BigEarthNet`_,C,Sentinel-1/2,"590,326",19--43,120x120,10,"SAR, MSI"
 `BioMassters`_,R,Sentinel-1/2 and Lidar,,,256x256, 10, "SAR, MSI"
+`ChaBuD`_,CD,Sentinel-2,"356",2,512x512,10,MSI
```
adamjstewart, line 6: 356 doesn't need quotes since there's no comma

tests/conf/chabud.yaml
```diff
@@ -0,0 +1,18 @@
...
+    num_filters: 1
+    ignore_index: null
+data:
+  class_path: ChaBuDDataModule
+  init_args:
+    batch_size: 2
+  dict_kwargs:
+    root: "tests/data/chabud"
+    bands: ["B4", "B3", "B2"]
+    download: true
```
adamjstewart, line 18: Can you add an EOL char? Probably easiest to configure your editor to do this automatically.

tests/datasets/test_chabud.py
```diff
@@ -0,0 +1,113 @@
...
+import shutil
+from pathlib import Path
+from typing import Any
+
+import matplotlib.pyplot as plt
+import pytest
+import torch
+import torch.nn as nn
+from _pytest.fixtures import SubRequest
+from _pytest.monkeypatch import MonkeyPatch
```
adamjstewart, line 15:
```suggestion
from pytest import MonkeyPatch
```
This is public now

torchgeo/datamodules/chabud.py
```diff
@@ -0,0 +1,77 @@
...
+from einops import repeat
+
+from ..datasets import ChaBuD
+from .geo import NonGeoDataModule
+
+
+class ChaBuDDataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the ChaBuD dataset.
+
+    Uses the train/val splits from the dataset
```
adamjstewart, line 18: No test split? Should we make one by subdividing val?

torchgeo/datamodules/chabud.py
```diff
@@ -0,0 +1,77 @@
...
+        band_indices = [ChaBuD.all_bands.index(b) for b in self.bands]
+        mins = self.min[band_indices]
+        maxs = self.max[band_indices]
+
+        # Change detection, 2 images from different times
+        mins = repeat(mins, "c -> (t c)", t=2)
+        maxs = repeat(maxs, "c -> (t c)", t=2)
+
+        self.min = mins
+        self.max = maxs - mins
```
adamjstewart, line 67: These never actually get used? Maybe you meant to set mean and std?

torchgeo/datasets/chabud.py
```diff
@@ -0,0 +1,294 @@
...
+
+from .geo import NonGeoDataset
+from .utils import DatasetNotFoundError, download_url, percentile_normalization
+
+
+class ChaBuD(NonGeoDataset):
+    """ChaBuD dataset.
+
+    The `ChaBuD <https://huggingface.co/spaces/competitions/ChaBuD-ECML-PKDD2023>`__
+    dataset is a dataset for Change detection for Burned area Delineation and is used
```
adamjstewart, line 23:
```suggestion
    `ChaBuD <https://huggingface.co/spaces/competitions/ChaBuD-ECML-PKDD2023>`__
    is a dataset for Change detection for Burned area Delineation and is used
```

torchgeo/datasets/chabud.py
```diff
@@ -0,0 +1,294 @@
...
+        "B3",
+        "B4",
+        "B5",
+        "B6",
+        "B7",
+        "B8",
+        "B8A",
+        "B9",
+        "B11",
+        "B12",
```
adamjstewart, line 67: In other S2 datasets I think we standardized on `B01` instead of `B1`

torchgeo/datasets/chabud.py
```diff
@@ -0,0 +1,294 @@
...
+    folds = {"train": [1, 2, 3, 4], "val": [0]}
+    url = "https://huggingface.co/datasets/chabud-team/chabud-ecml-pkdd2023/resolve/main/train_eval.hdf5"  # noqa: E501
+    filename = "train_eval.hdf5"
+    md5 = "15d78fb825f9a81dad600db828d22c08"
+
+    def __init__(
+        self,
+        root: str = "data",
+        split: str = "train",
+        bands: list[str] = rgb_bands,
```
adamjstewart, line 79: Should probably default to all bands, right?

torchgeo/datasets/chabud.py
```diff
@@ -0,0 +1,294 @@
...
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+
+        Raises:
+            AssertionError: if ``split`` or ``bands`` argument are invalid
+            DatasetNotFoundError: If dataset is not found and *download* is False.
+        """
+        assert split in self.folds
+        for band in bands:
+            assert band in self.all_bands
```
adamjstewart, line 101: Could be done in one line if you want:
```suggestion
        assert set(bands) <= set(self.all_bands)
```
This is optional though, whatever you think is clearer.

torchgeo/datasets/chabud.py
```diff
@@ -0,0 +1,294 @@
...
+        import h5py
+
+        uuid = self.uuids[index]
+        with h5py.File(self.filepath, "r") as f:
+            pre_array = f[uuid]["pre_fire"][:]
+            post_array = f[uuid]["post_fire"][:]
+
+        # index specified bands and concatenate
+        pre_array = np.take(pre_array, indices=self.band_indices, axis=-1)
+        post_array = np.take(post_array, indices=self.band_indices, axis=-1)
```
adamjstewart, line 185: Didn't know about `np.take`, very fancy! Maybe this also works?
```python
pre_array = pre_array[..., self.band_indices]
```

adamjstewart (changes requested):
I believe @robmarkcole is trying to standardize all CD datasets to return two images instead of returning a single stacked image. We'll then add a new CD trainer specifically for working with that data.

torchgeo/datamodules/chabud.py
isaaccorley: I'll have to check back in on the competition. I'm not sure if they made a test set public or if it just has no labels.

torchgeo/datamodules/chabud.py
isaaccorley: Good catch, I'll fix this

torchgeo/datasets/chabud.py
isaaccorley: I don't remember why I used take. Probably was feeling fancy

isaaccorley:
> I believe @robmarkcole is trying to standardize all CD datasets to return two images instead of returning a single stacked image. We'll then add a new CD trainer specifically for working with that data.

I'm aware of this, however I think we should wait until we have a CD trainer to update these to use `image1/image2` otherwise we have no way to test that the datamodules work with a trainer. Right now it works with the segmentation trainer.

torchgeo/datamodules/chabud.py
isaaccorley: I'll add a test set in a new PR. Looks like they changed the format of the test file so it will take some significant changes.

torchgeo/datamodules/chabud.py
```diff
@@ -0,0 +1,77 @@
...
+        band_indices = [ChaBuD.all_bands.index(b) for b in self.bands]
+        mins = self.min[band_indices]
+        maxs = self.max[band_indices]
+
+        # Change detection, 2 images from different times
+        mins = repeat(mins, "c -> (t c)", t=2)
+        maxs = repeat(maxs, "c -> (t c)", t=2)
+
+        self.mean = mins
+        self.std = maxs - mins
```
adamjstewart, line 67: This needs to happen before super()

torchgeo/datamodules/chabud.py
```diff
@@ -0,0 +1,77 @@
...
+
+        Args:
+            batch_size: Size of each mini-batch.
+            num_workers: Number of workers for parallel data loading.
+            **kwargs: Additional keyword arguments passed to
+                :class:`~torchgeo.datasets.ChaBuD`.
+        """
+        super().__init__(ChaBuD, batch_size, num_workers, **kwargs)
+
+        self.bands = kwargs.get("bands", ChaBuD.rgb_bands)
```
adamjstewart, line 57: Needs to change to all to match default of dataset

adamjstewart (approved):

torchgeo/datasets/chabud.py
```diff
@@ -0,0 +1,293 @@
...
+    * single hdf5 dataset containing images and masks
+
+    Dataset classes:
+
+    0. no change
+    1. burned area
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://doi.org/10.1016/j.rse.2021.112603
```
adamjstewart, line 44: A bit confused why this cites the old ZueriCrop dataset paper...

torchgeo/datasets/chabud.py
```diff
@@ -0,0 +1,293 @@
...
+from torch import Tensor
+
+from .geo import NonGeoDataset
+from .utils import DatasetNotFoundError, download_url, percentile_normalization
+
+
+class ChaBuD(NonGeoDataset):
+    """ChaBuD dataset.
+
+    `ChaBuD <https://huggingface.co/spaces/competitions/ChaBuD-ECML-PKDD2023>`__
```
adamjstewart, line 22: Homepage is now dead, updating...

torchgeo/datasets/chabud.py
isaaccorley: Oh just classic copy/paste from yours truly

adamjstewart: No actually, the official paper citation on the dataset homepage is also the ZueriCrop dataset.

isaaccorley: Oh interesting then yeah that's definitely incorrect

torchgeo/datasets/chabud.py
lccol: I can recover the MD files for the competition description and dataset description if you want... maybe I can create a github repo with just that and update the link accordingly

adamjstewart: I changed it to the hugging face repo. That has some description, but the more the better.

lccol: Dumped the page which was on HF here: [lccol/chabud-challenge-description](https://github.com/lccol/chabud-challenge-description/tree/main)

adamjstewart: Are we supposed to cite the CaBuAr dataset or the ZueriCrops dataset?

lccol: CaBuAr definitely, which is the full version of the dataset.

Arxiv: https://arxiv.org/abs/2401.11519
from IEEE: https://ieeexplore.ieee.org/document/10261881

adamjstewart: Do you want to submit a PR to fix the citation URL? We can include this in the 0.8.1 release.

lccol: ok, doing that

lccol: here: https://github.com/torchgeo/torchgeo/pull/3130

PR #1260, opened 2023-04-18
DataModule: common base class to reduce code duplication

adamjstewart:
A lot of the code in our base classes was shared, so I aggregated it.

Originally I was just planning on adding better error messages, but figured I would do this first.

### DIFF
--- a/torchgeo/datamodules/geo.py
+++ b/torchgeo/datamodules/geo.py
@@ -23,14 +23,75 @@ from ..transforms import AugmentationSequential
 from .utils import MisconfigurationException
 
 
-class GeoDataModule(LightningDataModule):  # type: ignore[misc]
-    """Base class for data modules containing geospatial information.
+class DataModule(LightningDataModule):  # type: ignore[misc]
+    """Base class for all TorchGeo data modules.
 
-    .. versionadded:: 0.4
+    .. versionadded:: 0.5
     """
 
     mean = torch.tensor(0)
     std = torch.tensor(255)
+
+    def prepare_data(self) -> None:
+        """Download and prepare data.
+
+        During distributed training, this method is called only within a single process
+        to avoid corrupted data. This method should not set state since it is not called
+        on every device, use :meth:`setup` instead.
+        """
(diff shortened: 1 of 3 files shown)


torchgeo/datamodules/geo.py
```diff
@@ -23,15 +23,76 @@
 from .utils import MisconfigurationException
 
 
-class GeoDataModule(LightningDataModule):  # type: ignore[misc]
-    """Base class for data modules containing geospatial information.
+class DataModule(LightningDataModule):  # type: ignore[misc]
```
calebrob6, line 26: Should we call this a TorchgeoDataModule or something?

calebrob6 (commented):
This feels kind of like premature optimization (is premature abstracting a thing?) as it only saves a net few lines (just repeated twice), and "DataModule" doesn't seem like the correct name. It is good because I can't think of a case why we would want to go back and make different `plot`, `on_after_batch_transfer` or `prepare_data` methods for the Geo and NonGeo data modules.

adamjstewart:
For context, the reason I made this change was because I wanted to improve the datamodule error messages. The current error message just says:
```
Foo.setup does not define a 'foo_dataset'
```
but the actual error could be any of:

* Foo.setup does not define a foo_dataset or dataset
* Foo.setup does not define a foo_sampler or sampler or foo_batch_sampler or batch_sampler
* The dataset length is 0
* The sampler length is 0

Since this is a bit of code and requires access to self, I figured it would make sense to add this check to a base class. When I started working on a base class, that's when I realized how much duplicated code there was. So I figured it would be easiest to add the base class in one PR and improve error messages in a second PR.

torchgeo/datamodules/geo.py
adamjstewart: Could, I don't think it matters too much. The user should never use this class.

calebrob6:
Totally fine! I just think the name "DataModule" might be confusing, maybe "BaseDataModule"? Else, I'm cool with it.

calebrob6 (approved):

PR #1262, opened 2023-04-18
Add Western USA Live Fuel Moisture Dataset

nilsleh:
This PR adds the [Western USA Live Fuel Moisture Dataset](https://mlhub.earth/data/su_sar_moisture_content_main) from MLHub. It is a "new" dataset to torchgeo in the sense that it is a Tabular dataset that has satellite sensor derived variables as features for a regression task.

Dataset Format:
- .geojson file for each datapoint

Dataset Features:
- 138 remote sensing derived variables, some with a time dependency
- 2615 datapoints with regression target of predicting fuel moisture


torchgeo/datasets/western_usa_live_fuel_moisture.py
```diff
@@ -0,0 +1,357 @@
...
+            "input": torch.tensor(data.drop([self.label_name]), dtype=torch.float32),
+            "label": torch.tensor(data[self.label_name], dtype=torch.float32),
+        }
+
+        if self.transforms is not None:
+            sample = self.transforms(sample)
+
+        return sample
+
+    def _load_data(self):
```
nilsleh, line 291: I am not sure how to handle the return type here, with the conditional pandas import.

torchgeo/datasets/western_usa_live_fuel_moisture.py
```diff
@@ -0,0 +1,353 @@
...
+    def __init__(
+        self,
+        root: str = "data",
+        input_features: list[str] = all_variable_names,
+        transforms: Optional[Callable[[dict[str, Any]], dict[str, Any]]] = None,
+        download: bool = False,
+        api_key: Optional[str] = None,
+        checksum: bool = False,
+    ) -> None:
+        """Initialize a new Tropical Cyclone Wind Estimation Competition Dataset.
```
adamjstewart, line 205: Need to update dataset name

torchgeo/datasets/western_usa_live_fuel_moisture.py
```diff
@@ -0,0 +1,353 @@
...
+from typing import Any, Callable, Optional
+
+import torch
+from torch import Tensor
+
+from .geo import NonGeoDataset
+from .utils import download_radiant_mlhub_collection, extract_archive
+
+
+class WesternUSALiveFuelMoisture(NonGeoDataset):
```
adamjstewart, line 18: Could also make it a GeoDataset if you want, that's what I did for GBIF/iNaturalist/EDDMapS. Depends on how you intend to use the dataset and whether or not you need to combine it with other data sources.

torchgeo/datasets/western_usa_live_fuel_moisture.py
adamjstewart: Try this:
```suggestion
    def _load_data(self) -> "pd.DataFrame":
```
The quotes prevent it from being evaluated at run-time.

torchgeo/datasets/western_usa_live_fuel_moisture.py
```diff
@@ -0,0 +1,353 @@
...
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://doi.org/10.1016/j.rse.2020.111797
+
+    .. note::
+
+       This dataset requires the following additional library to be installed:
+
+       * `radiant-mlhub <https://pypi.org/project/radiant-mlhub/>`_ to download the
+         imagery and labels from the Radiant Earth MLHub
```
adamjstewart, line 45: Need versionadded

torchgeo/datasets/western_usa_live_fuel_moisture.py
nilsleh: At the moment I think I want to use just by itself and index the datapoints in "standard tabular machine learning" fashion. I also discovered that there is an image-based version from the corresponding paper: https://github.com/kkraoj/lfmc_from_sar so looking to add that as well maybe

torchgeo/datasets/western_usa_live_fuel_moisture.py
nilsleh: I still get flake8:
`torchgeo/datasets/western_usa_live_fuel_moisture.py:289:29: F821 undefined name 'pd'` 

and 

mypy:
`torchgeo/datasets/western_usa_live_fuel_moisture.py:289: error: Name "pd" is not defined  [name-defined]`

adamjstewart: What about `"pandas.DataFrame"`? We had to hack things to get pyvista working in idtrees.py and ended up using:
```
# type: ignore[name-defined] # noqa: F821
```
to ignore the issue. At some point in the future I may make pandas a required dep so the issue will go away.

adamjstewart (dismissed):
Assuming tests pass

docs/api/datasets.rst
```diff
@@ -329,6 +329,11 @@ VHR-10
 
 .. autoclass:: VHR10
 
+WesternUSALiveFuelMoisture
```
adamjstewart, line 332: I would separate this out into separate words. Will also need to change in the table. This will help split the title onto multiple lines in the table.

torchgeo/datasets/western_usa_live_fuel_moisture.py
```diff
@@ -0,0 +1,355 @@
...
+                entry and returns a transformed version
+            download: if True, download dataset and store it in the root directory
+            api_key: a RadiantEarth MLHub API key to use for downloading the dataset
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+
+        Raises:
+            AssertionError: if ``split`` argument is invalid
+            RuntimeError: if ``download=False`` but dataset is missing or checksum fails
+            RuntimeError: if ``input_features`` contains invalid variable names
+            ImportError: if pandas is not installed
```
adamjstewart, line 222:
```suggestion
            AssertionError: if ``input_features`` contains invalid variable names
            ImportError: if pandas is not installed
            RuntimeError: if ``download=False`` but dataset is missing or checksum fails
```

PR #1265, opened 2023-04-18
add FireRisk dataset

isaaccorley:
This PR adds the [FireRisk](https://github.com/CharmonyShen/FireRisk) dataset and datamodule for risk image classification of 320x320 crops from NAIP tiles in .png format with 7 risk categories. 

[Paper](https://arxiv.org/abs/2303.07035)

Some example plots:

firerisk1
firerisk2
firerisk3
firerisk4

@CharmonyShen


tests/data/fire_risk/data.py
```diff
@@ -0,0 +1,93 @@
...
+        "val",
+        "Very_High",
+        "35502941_5_-122.968467383602_40.2960022654498.png",
+    ),
+]
+
+
+def create_file(path: str) -> None:
+    Z = np.random.rand(SIZE, SIZE, 3) * 255
+    img = Image.fromarray(Z.astype("uint8")).convert("RGB")
```
adamjstewart, line 71:
```suggestion
    Z = np.random.randint(255, size=(SIZE, SIZE, 3), dtype=np.uint8)
    img = Image.fromarray(Z).convert("RGB")
```

tests/conf/fire_risk.yaml
```diff
@@ -0,0 +1,15 @@
...
+    loss: "ce"
+    model: "resnet18"
+    learning_rate: 1e-3
+    learning_rate_schedule_patience: 6
+    weights: null
+    in_channels: 3
+    num_classes: 7
+  datamodule:
+    root: "tests/data/fire_risk"
+    download: true
```
adamjstewart, line 13:
```suggestion
    download: false
```

torchgeo/datamodules/fire_risk.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class FireRiskDataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the RESISC45 dataset.
+
+    .. versionadded:: 0.5
+    """
+
+    mean = torch.tensor([0, 0, 0])
+    std = torch.tensor([255.0, 255.0, 255.0])
```
adamjstewart, line 23: These are the defaults in the base class already, prob don't need to add here

torchgeo/datamodules/fire_risk.py
```diff
@@ -0,0 +1,73 @@
...
+import kornia.augmentation as K
+import torch
+
+from ..datasets import FireRisk
+from ..transforms import AugmentationSequential
+from .geo import NonGeoDataModule
+
+
+class FireRiskDataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the RESISC45 dataset.
```
adamjstewart, line 17:
```suggestion
    """LightningDataModule implementation for the FireRisk dataset.
```

torchgeo/datamodules/fire_risk.py
```diff
@@ -0,0 +1,73 @@
...
+        Args:
+            batch_size: Size of each mini-batch.
+            num_workers: Number of workers for parallel data loading.
+            **kwargs: Additional keyword arguments passed to
+                :class:`~torchgeo.datasets.FireRisk`.
+        """
+        super().__init__(FireRisk, batch_size, num_workers, **kwargs)
+        self.aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std), data_keys=["image"]
+        )
```
adamjstewart, line 39: This is already added by the base class, no need to add here

torchgeo/datamodules/fire_risk.py
```diff
@@ -0,0 +1,73 @@
...
+            K.ColorJitter(p=0.5, brightness=0.1, contrast=0.1, saturation=0.1, hue=0.1),
+            data_keys=["image"],
+        )
+
+    def setup(self, stage: str) -> None:
+        """Set up datasets.
+
+        Called at the beginning of fit, validate, test, or predict. During distributed
+        training, this method is called from every process across all the nodes. Setting
+        state here is recommended.
```
adamjstewart, line 56: I don't usually add this paragraph to subclasses, it's already documented in the base class

torchgeo/datamodules/fire_risk.py
```diff
@@ -0,0 +1,73 @@
...
+            )
+        if stage in ["fit", "validate"]:
+            self.val_dataset = self.dataset_class(  # type: ignore[call-arg]
+                split="val", **self.kwargs
+            )
+        if stage in ["test"]:
+            # FireRisk has no test set
+            self.test_dataset = self.dataset_class(  # type: ignore[call-arg]
+                split="val", **self.kwargs
+            )
```
adamjstewart, line 73: I would either use no test_dataset or split val

torchgeo/datamodules/fire_risk.py
```diff
@@ -0,0 +1,73 @@
...
+
+        Called at the beginning of fit, validate, test, or predict. During distributed
+        training, this method is called from every process across all the nodes. Setting
+        state here is recommended.
+
+        Args:
+            stage: Either 'fit', 'validate', 'test', or 'predict'.
+        """
+        if stage in ["fit"]:
+            self.train_dataset = self.dataset_class(  # type: ignore[call-arg]
```
adamjstewart, line 62: You can use FireRisk instead of dataset_class to avoid the type ignore

torchgeo/datasets/fire_risk.py
```diff
@@ -0,0 +1,174 @@
...
+    ) -> None:
+        """Initialize a new FireRisk dataset instance.
+
+        Args:
+            root: root directory where dataset can be found
+            split: one of "train" or "val"
+            transforms: a function/transform that takes input sample and its target as
+                entry and returns a transformed version
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
```
adamjstewart, line 82: Raises AssertionError or RuntimeError

torchgeo/datasets/fire_risk.py
```diff
@@ -0,0 +1,174 @@
...
+        # Check if zip file already exists (if so then extract)
+        filepath = os.path.join(self.root, self.filename)
+        if os.path.exists(filepath):
+            self._extract()
+            return
+
+        # Check if the user requested to download the dataset
+        if not self.download:
+            raise RuntimeError(
+                "Dataset not found in `root` directory and `download=False`, "
```
adamjstewart, line 115:
```suggestion
                f"Dataset not found in `root={self.root}` directory and `download=False`, "
```

adamjstewart (changes requested):

torchgeo/datamodules/fire_risk.py
isaaccorley: I'd prefer not to randomly split into a test set. However having no test set doesn't work currently because our datamodule tests perform `trainer.test()` which requires a test set. I'll modify them to skip if a  test set is not available.

torchgeo/datamodules/fire_risk.py
```diff
@@ -0,0 +1,56 @@
...
+        Args:
+            stage: Either 'fit', 'validate', 'test', or 'predict'.
+        """
+        if stage in ["fit"]:
+            self.train_dataset = FireRisk(split="train", **self.kwargs)
+        if stage in ["fit", "validate"]:
+            self.val_dataset = FireRisk(split="val", **self.kwargs)
+        if stage in ["test"]:
+            # FireRisk has no test set
+            self.test_dataset = None
```
adamjstewart, line 56: You can delete this, it's already set by the base class

isaaccorley: Done

isaaccorley:
I'm not sure why this is failing now. It works fine for me locally.

isaaccorley:
Nvm figured it out

tests/conf/fire_risk.yaml
adamjstewart: One more

adamjstewart (approved):

PR #1266, opened 2023-04-19
Windows CI: replace pacman with choco

adamjstewart:
Honestly not sure how this ever worked. My text editor marks `\b` as the ASCII backspace character. Using single quotes instead of double quotes avoids backslash being interpreted as a special character. I'm guessing the reason it works is because the YAML interpreter GitHub uses doesn't register `\b`? Anyway, this fix makes my editor happy.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -67,7 +67,7 @@ jobs:
     - name: Install pacman dependencies (Windows)
       shell: cmd
       run: |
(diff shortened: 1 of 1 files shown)


adamjstewart:
Found an easier option, assuming it works: https://docs.github.com/en/actions/using-github-hosted-runners/customizing-github-hosted-runners#installing-software-on-windows-runners

isaaccorley (approved):

PR #1272, opened 2023-04-21
CI: fix minimum test build failure

adamjstewart:


calebrob6:
did ubuntu-latest change?

adamjstewart:
I think this is working now. My best guess as to what broke is the latest setuptools release (v67.7.1) which came out last night. This specifically contains https://github.com/pypa/setuptools/pull/3900, which "fixes" https://github.com/pypa/setuptools/issues/3898, which looks suspiciously similar to our issue.

Our minimum tests _should_ pin us to an older version of setuptools, but pip is allowed to install a newer version of setuptools in the background that is used only at build time. I tried constraining us to an older setuptools but that didn't help, so maybe constraints files only affect the final version that gets installed. Anyway, using a newer version of pycocotools avoids the use of deprecated features, so I think it's worth upgrading anyway. Hopefully someday they'll ship wheels.

calebrob6 (approved):

PR #1290, opened 2023-04-25
freeze_backbone and freeze_decoder in Trainers

isaaccorley:
This PR adds a `freeze_backbone` arg to all trainers and a `freeze_decoder` arg to the `SemanticSegmentationTask` and `PixelwiseRegressionTask`. This is particularly useful if wanting to fine-tune (aka linear probe) only the classifier head of a pretrained model or only the decoder or segmentation head in an encoder/decoder architecture.


adamjstewart:
Can you do other trainers too? For this paper, all of our tasks will be semantic segmentation, but might as well do all trainers if it's easy.

torchgeo/trainers/classification.py
```diff
@@ -44,7 +44,11 @@ class ClassificationTask(LightningModule):  # type: ignore[misc]
     """
 
     def config_model(self) -> None:
-        """Configures the model based on kwargs parameters passed to the constructor."""
+        """Configures the model based on kwargs parameters passed to the constructor.
+
+        .. versionchanged:: 0.5
+           Added *freeze_backbone* parameter.
```
adamjstewart, line 50: This function has no parameters

torchgeo/trainers/classification.py
```diff
@@ -90,9 +101,12 @@ def __init__(self, **kwargs: Any) -> None:
             in_channels: Number of input channels to model
             learning_rate: Learning rate for optimizer
             learning_rate_schedule_patience: Patience for learning rate scheduler
+            freeze_backbone: Fine-tune the cls head by freezing the backbone
 
         .. versionchanged:: 0.4
            The *classification_model* parameter was renamed to *model*.
```
adamjstewart, line 107: Need to check the docs, might need an empty line between these

torchgeo/trainers/classification.py
```diff
@@ -90,9 +101,12 @@ def __init__(self, **kwargs: Any) -> None:
             in_channels: Number of input channels to model
             learning_rate: Learning rate for optimizer
             learning_rate_schedule_patience: Patience for learning rate scheduler
+            freeze_backbone: Fine-tune the cls head by freezing the backbone
```
adamjstewart, line 104:
```suggestion
            freeze_backbone: Fine-tune the classifier head by freezing the backbone
```

isaaccorley: linear probing is the action of actually fine-tuning. freezing the backbone is what is happening when you set this arg to True. Imo makes more sense to call it `freeze_backbone`.

torchgeo/trainers/classification.py
isaaccorley: Should I remove the version changed completely then? Or should I change the comment to something like `Added ability freeze_backbone` ?

adamjstewart: Yes, remove it. I'm actually planning on getting rid of this function entirely in a refactor, or at least making it private

calebrob6 (approved):
sick

nilsleh (approved):

torchgeo/trainers/classification.py
```diff
@@ -90,9 +97,13 @@ def __init__(self, **kwargs: Any) -> None:
...
 
         .. versionchanged:: 0.4
            The *classification_model* parameter was renamed to *model*.
+
+        .. versionchanged:: 0.5
+           Added *freeze_backbone* parameter.
```
adamjstewart, line 106:
```rst
.. versionadded:: 0.5
   The *freeze_backbone* parameter.
```

torchgeo/trainers/classification.py
```diff
@@ -90,9 +97,13 @@ def __init__(self, **kwargs: Any) -> None:
             in_channels: Number of input channels to model
             learning_rate: Learning rate for optimizer
             learning_rate_schedule_patience: Patience for learning rate scheduler
+            freeze_backbone: Fine-tune the cls head by freezing the backbone
```
adamjstewart, line 100: Fine-tune the classifier head. Also, fine-tune is misleading since this is only used in the case where you aren't fine-tuning the model, you're doing linear probing.

torchgeo/trainers/regression.py
```diff
@@ -294,3 +305,17 @@ def config_model(self) -> None:
...
+
+        # Freeze backbone
+        if self.hyperparams.get("freeze_backbone", False) and self.hyperparams[
+            "model"
+        ] in ["unet", "deeplabv3+"]:
+            for param in self.model.encoder.parameters():
+                param.requires_grad = False
+
+        # Freeze decoder
+        if self.hyperparams.get("freeze_decoder", False) and self.hyperparams[
```
adamjstewart, line 317: This parameter isn't documented

torchgeo/trainers/segmentation.py
```diff
@@ -110,7 +124,9 @@ class and used with 'ce' loss
            *encoder_weights* to *weights*.
 
         .. versionadded: 0.5
-            The *class_weights* parameter.
+            The *class_weights*, *freeze_backbone*,
+            and *freeze_decoder* parameters.
```
adamjstewart, line 128: Not documented.

torchgeo/trainers/segmentation.py
```diff
@@ -80,6 +80,20 @@ def config_task(self) -> None:
                 f"Currently, supports 'ce', 'jaccard' or 'focal' loss."
             )
 
+        # Freeze backbone
+        if self.hyperparams.get("freeze_backbone", False) and self.hyperparams[
+            "model"
+        ] in ["unet", "deeplabv3+"]:
```
adamjstewart, line 86: What about fcn?

torchgeo/trainers/classification.py
```diff
@@ -64,6 +64,13 @@ def config_model(self) -> None:
                 state_dict = get_weight(weights).get_state_dict(progress=True)
             self.model = utils.load_state_dict(self.model, state_dict)
 
+        # Freeze backbone and unfreeze classifier head
+        if self.hyperparams.get("freeze_backbone", False):
+            for param in self.model.parameters():
+                param.requires_grad = False
+            for param in self.model.get_classifier().parameters():
```
adamjstewart, line 71: What does `get_classifier` look like for various models? Is it always everything except the last layer?

adamjstewart (commented):
Not sure about the parameter names. Also not sure if freezing the ResNet encoder in a U-Net counts as linear probing or not. Same with classification/regression. Might be easier just to freeze the entire model and add a single layer MLP to the end.

isaaccorley:
Can you reference some papers that freeze the entire UNet and add fine-tune an MLP on top? I have never seen this before. Also it would make more sense to fine tune a conv layer not a MLP.

Could you also suggest better parameter names?

torchgeo/trainers/segmentation.py
isaaccorley: FCN doesn't use a backbone.

torchgeo/trainers/classification.py
isaaccorley: They added this because sometimes they name the head differently, e.g. head, fc, cls, etc.

adamjstewart:
I'm only talking about classification/regression. No idea what the usual practice is for linear probing of semantic segmentation.

Will think about parameter names, don't have a better name off the top of my head. I still kind of like linear_probing because it makes it clear _why_ you would want to use the parameter.

torchgeo/trainers/segmentation.py
adamjstewart: But you might still want to evaluate linear probing on it, right?

isaaccorley:
Disagree about the naming. Freeze_backbone is the action we are performing. Linear probing is just the downstream task a user may choose to perform with the classifier with a frozen backbone. There may be other potential reasons for freezing the backbone other than linear probing.

PR #1291, opened 2023-04-26
fix typo in docs: ucmerced n=2,100 not n=21,000

urbanophile:
I would like to fix a small typo in the documentation. The UC Merced land-use dataset consists of 2,100 images, not 21,000 (although it would be nice if it were more extensive). As far as I've seen, n=21,000 is only referenced in the table, while the actual UC Merced dataset class refers to the correct n=2,100.


calebrob6 (approved):

calebrob6:
Good catch @urbanophile, thanks!!

PR #1296, opened 2023-04-30
SSL4EO: more generic download script

adamjstewart:
#1223 included https://github.com/microsoft/torchgeo/pull/1223/commits/53a776e67f5a7abb24d38c31dce95e03b670029e, which hacked the script to resize the panchromatic band on Landsat 8. Unfortunately, this resulted in the script only working for Landsat 8 Level-1 data. This PR further generalizes this feature to support both resampling all bands to a single resolution and keeping all bands in their original resolution. It also includes scripts to bash scripts to download data for each satellite.

I tested all scripts to make sure they are able to download data and the images look correct (all bands have the same dimensions and spatial extent). There are still a few garbage images (clouds, nodata pixels, noise) but there isn't much we can do about that at this stage.

Note that there is almost no data for Landsat 5, especially near populated cities. It's unclear if we'll be able to train or evaluate models for Landsat 5 MSS.

@wangyi111

### DIFF
--- a/experiments/ssl4eo/download_l5_raw.sh
+++ b/experiments/ssl4eo/download_l5_raw.sh
@@ -1,4 +1,6 @@
-#!/usr/bin/env bash -euo pipefail
+#!/usr/bin/env bash
+
+set -euo pipefail
 
 # User-specific parameters
 SAVE_PATH=data/ssl4eo-l5-raw
--- a/experiments/ssl4eo/download_l5_raw.sh
+++ b/experiments/ssl4eo/download_l5_raw.sh
@@ -6,7 +6,7 @@ set -euo pipefail
 ROOT_DIR=data
 SAVE_PATH="$ROOT_DIR/ssl4eo-l5-raw"
 MATCH_FILE="$ROOT_DIR/ssl4eo-l-60/sampled_locations.csv"
-NUM_WORKERS=28
+NUM_WORKERS=40
 START_INDEX=0
 END_INDEX=10
 
--- a/experiments/ssl4eo/download_l5_raw.sh
+++ b/experiments/ssl4eo/download_l5_raw.sh
@@ -18,6 +18,7 @@ YEAR=1991  # MSS data acquisitions over the United States ceased in 1992
 BANDS=(B1 B2 B3 B4)
 ORIGINAL_RESOLUTIONS=(60 60 60 30)
 NEW_RESOLUTIONS=60
+DEFAULT_VALUE=0
 
 # Generic parameters
 SCRIPT_DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
@@ -39,6 +40,7 @@ time python3 "$SCRIPT_DIR/download_ssl4eo.py" \
     --original-resolutions ${ORIGINAL_RESOLUTIONS[@]} \
     --new-resolutions $NEW_RESOLUTIONS \
     --dtype $DTYPE \
+    --default-value $DEFAULT_VALUE \
     --num-workers $NUM_WORKERS \
     --log-freq $LOG_FREQ \
     --match-file "$MATCH_FILE" \
--- a/experiments/ssl4eo/download_l5_raw.sh
+++ b/experiments/ssl4eo/download_l5_raw.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash -euo pipefail
+
+# User-specific parameters
+SAVE_PATH=data/ssl4eo-l5-raw
+NUM_WORKERS=28
+MATCH_FILE=data/ssl4eo-l-60/sampled_locations.csv
+START_INDEX=0
+END_INDEX=10
+
+# Satellite-specific parameters
+COLLECTION=LANDSAT/LM05/C02/T1
+QA_BAND=QA_PIXEL
+QA_CLOUD_BIT=3
+META_CLOUD_NAME=CLOUD_COVER
+YEAR=1991  # MSS data acquisitions over the United States ceased in 1992
+BANDS=(B1 B2 B3 B4)
+ORIGINAL_RESOLUTIONS=(60 60 60 30)
+NEW_RESOLUTIONS=60
+
+# Generic parameters
+SCRIPT_DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
+CLOUD_PCT=20
+SIZE=264
+DTYPE=float32
+LOG_FREQ=1000
+
+time python3 "$SCRIPT_DIR/download_ssl4eo.py" \
+    --save-path "$SAVE_PATH" \
+    --collection $COLLECTION \
+    --qa-band $QA_BAND \
+    --qa-cloud-bit $QA_CLOUD_BIT \
+    --meta-cloud-name $META_CLOUD_NAME \
+    --cloud-pct $CLOUD_PCT \
+    --dates $YEAR-03-20 $YEAR-06-21 $YEAR-09-23 $YEAR-12-21 \
+    --radius $(($NEW_RESOLUTIONS * $SIZE / 2)) \
+    --bands ${BANDS[@]} \
+    --original-resolutions ${ORIGINAL_RESOLUTIONS[@]} \
+    --new-resolutions $NEW_RESOLUTIONS \
+    --dtype $DTYPE \
+    --num-workers $NUM_WORKERS \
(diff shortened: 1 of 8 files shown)


experiments/ssl4eo/download_l5_raw.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash -euo pipefail
+
+# User-specific parameters
```
adamjstewart, line 3: These are parameters that people will need to change depending on where and how they will run the script.

experiments/ssl4eo/download_ssl4eo.py
```diff
@@ -135,10 +133,10 @@ def filter_collection(
 
 
 def center_crop(
-    img: np.ndarray[Any, np.dtype[Any]], out_size: tuple[int, int]
+    img: np.ndarray[Any, np.dtype[Any]], out_size: int
```
adamjstewart, line 136: We only support square crops, no reason to make this function more complicated

experiments/ssl4eo/download_ssl4eo.py
```diff
@@ -163,78 +161,63 @@ def adjust_coords(
     ]
 
 
-def get_properties(image: ee.Image) -> Any:
-    return image.getInfo()
```
adamjstewart, line 167: 1-liner functions actually take up more lines than not using them.

experiments/ssl4eo/download_ssl4eo.py
```diff
@@ -163,78 +161,63 @@ def adjust_coords(
...
-        if band == reproj_bands:
-            img = np.atleast_3d(features_reproj["properties"][band])
-        else:
-            img = np.atleast_3d(features_other["properties"][band])
-        if crop is not None:
-            img = center_crop(img, out_size=crop[band])
-        raster[band] = img.astype(dtype)
-
-    coords0 = np.array(features_other["geometry"]["coordinates"][0])
+    image = collection.sort("CLOUD_COVER").first()
```
adamjstewart, line 173: In my mind, it's better to sort by least cloud cover than by more recent images.

experiments/ssl4eo/download_ssl4eo.py
```diff
@@ -163,78 +161,63 @@ def adjust_coords(
...
+        indices, bands = zip(*value)
+        patch = image.select(*bands)
+        if orig_res != new_res:
+            patch = patch.reproject(patch.projection().crs(), scale=new_res)
+        patch = patch.sampleRectangle(region, defaultValue=0)
+        features = patch.getInfo()
+        for i, band in zip(indices, bands):
+            x = features["properties"][band]
+            x = np.atleast_3d(x)
+            x = center_crop(x, out_size=2 * radius // new_res)
```
adamjstewart, line 193: Instead of the user providing the crop size, we compute the appropriate crop size based on the resolution of each band.

experiments/ssl4eo/download_ssl4eo.py
```diff
@@ -340,7 +327,8 @@ def update(self, delta: int = 1) -> int:
         "--dates",
         type=str,
         nargs="+",
-        default=["2021-12-21", "2021-09-22", "2021-06-21", "2021-03-20"],
+        # https://www.weather.gov/media/ind/seasons.pdf
+        default=["2021-12-21", "2021-09-23", "2021-06-21", "2021-03-20"],
```
adamjstewart, line 331: Adjusted the September date a bit.

experiments/ssl4eo/download_ssl4eo.py
```diff
@@ -455,31 +465,34 @@ def worker(idx: int) -> None:
...
+            new_resolutions,
             dtype,
             dates,
             radius=args.radius,
             debug=args.debug,
             match_coords=match_coords,
         )
 
-        if patches is not None:
-            if args.save_path is not None:
```
adamjstewart, line 467: `args.save_path` can't be None, it is given a default value by argparse.

experiments/ssl4eo/download_ssl4eo.py
```diff
@@ -494,12 +507,7 @@ def worker(idx: int) -> None:
         return
 
     # set indices
-    if args.indices_range is not None:
```
adamjstewart, line 497: `args.indices_range` can't be None, it is given a default value by argparse.

wangyi111 (dismissed):
Looks good, I tested L8_L1 on my PC and works well.

adamjstewart:
Oddly, -9999 works for L1 but not for L2. Investigating now...

adamjstewart:
Might modify the bash scripts to spawn several processes in the background and see if that's faster.

adamjstewart:
There's a serious bug in this script. Occasionally the saved file has more bands than expected (12 instead of 11, or 12 instead of 6). No idea how that's even possible.

adamjstewart:
> Need to check if that was the case before this PR or not, or if serial vs. parallel has anything to do with it.

Bug was introduced by this PR, which narrows it down. Has nothing to do with multiprocessing.

adamjstewart:
> 12 instead of 6

In this example, all bands are duplicated (i.e., 1==2, 3==4, ...).

experiments/ssl4eo/download_l5_raw.sh
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
```
calebrob6, line 1: this isn't working on my windows machine??

calebrob6 (dismissed):

experiments/ssl4eo/download_l5_raw.sh
adamjstewart: What error are you getting?

calebrob6: oh I'm joking, its a bash script

adamjstewart:
40 seems to be the optimal number of cores, at least on our system and for Landsat 7 Level-2.
l7-l2

adamjstewart:
For Landsat 8 Level-1, it doesn't seem to matter how many threads are used, all threads produce the same rates.

adamjstewart:
For expected timing, it should be somewhere between:

* Landsat 7 Level-2 (least bands): 1.7 days
* Landsat 8 Level-1 (most bands): 6.3 days

for 500K scenes. This might be overkill, but a significant number of scenes either don't have a match or contain nodata pixels. Obviously this can be sped up by having multiple people download a fraction of the dataset in parallel.

experiments/ssl4eo/download_ssl4eo.py
isaaccorley: Does it support sorting by both like pandas does?

experiments/ssl4eo/sample_30.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# User-specific parameters
+SAVE_PATH=data/ssl4eo-l-30
+START_INDEX=0
+END_INDEX=10
+
+# Generic parameters
+SCRIPT_DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
+RES=30
+SIZE=264
```
isaaccorley, line 13: We download at 264 so we can center crop or something right?

experiments/ssl4eo/download_ssl4eo.py
```diff
@@ -163,78 +161,65 @@ def adjust_coords(
...
+    for (orig_res, new_res), value in band_groups.items():
+        indices, bands_group = zip(*value)
+        patch = image.select(*bands_group)
+        if orig_res != new_res:
+            patch = patch.reproject(patch.projection().crs(), scale=new_res)
+        patch = patch.sampleRectangle(region, defaultValue=default_value)
+        features = patch.getInfo()
+        for i, band in zip(indices, bands_group):
+            x = features["properties"][band]
+            x = np.atleast_3d(x)
```
isaaccorley, line 193: TIL about `np.atleast3d`

experiments/ssl4eo/download_ssl4eo.py
adamjstewart: No idea, but cloud cover is float so I doubt we'll ever find two that match.

experiments/ssl4eo/sample_30.sh
adamjstewart: Yep, also to match previous papers

calebrob6 (approved):

PR #1298, opened 2023-05-02
Add script to compute dataset statistics

adamjstewart:
This script computes the mean and standard deviation of all images in a dataset, even in the presence of nodata pixels. If we remove nodata pixels from the dataset, the implementation could become even simpler.

There are a couple of ways in which we could potentially speed up this script:

* [multiprocessing](https://docs.python.org/3/library/multiprocessing.html): I don't have a lot of experience with this
* [rasterio.statistics](https://rasterio.readthedocs.io/en/stable/api/rasterio.io.html#rasterio.io.DatasetReader.statistics): not sure if these stats would help

At the current rate, it would take about 10 hrs to run on 1M images. So it might be worth speeding up.

I confirmed that the current implementation matches what you get when you stack all images in the dataset and compute np.mean and np.std on the stack.


calebrob6 (dismissed):
You can make args.nan default to f.nodata instead of 0 if it isn't not set if you want

calebrob6:
Would change the name to something other than normalize -- you're computing dataset stats

adamjstewart:
What would be a better name? stats.py? statistics.py? It isn't going in core so name isn't too important.

calebrob6:
idk, `compute_dataset_statistics.py`?

isaaccorley (dismissed):

adamjstewart:
Going to try to parallelize this...

adamjstewart:
Going to add min/max and remove support for NaNs. Should be able to make this script much faster.

adamjstewart:
With 40 threads, it takes about 6 hrs to compute stats for 1.4M images. I have a feeling 1 training epoch is going to take even longer... Although we're going to convert from float32 to uint8 and compress, so maybe it will improve.

experiments/ssl4eo/compute_dataset_statistics.py
```diff
@@ -0,0 +1,69 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("directory", help="directory to recursively search for files")
+    parser.add_argument("--suffix", default=".tif", help="file suffix")
+    parser.add_argument("--num-workers", type=int, default=10, help="number of threads")
+    args = parser.parse_args()
+
+    def compute(path: str) -> "np.typing.NDArray[np.float32]":
+        """Compute the minimum and maximum values in a dataset.
```
nilsleh, line 23: Isn't this just to compute the min and max of a single tile within a dataset?

experiments/ssl4eo/compute_dataset_statistics.py
```diff
@@ -0,0 +1,69 @@
...
+            out_list.append(compute(path))
+        out = np.array(out_list)
+
+    minimum = np.amin(out[:, :, 0], axis=0)
+    maximum = np.amax(out[:, :, 1], axis=0)
+
+    mu_d = out[:, :, 2]
+    mu = np.mean(mu_d, axis=0)
+    sigma_d = out[:, :, 3]
+    N_d = 264**2
```
nilsleh, line 56: Should this be something not hardcoded?

adamjstewart:
@isaaccorley `approx=True` is indeed faster, but the stats are drastically different.

### `approx=False`

```
min: array([-1.3871045e-02,  2.4574728e+02,  5.4600620e+01,  1.0121686e+02], dtype=float32)
max: array([2.6160342e-01, 3.2517932e+02, 7.1928284e+01, 1.3391820e+02], dtype=float32)
mean: array([5.3258397e-02, 2.9934772e+02, 6.5858688e+01, 1.2290101e+02], dtype=float32)
std: array([9.1137975e-02, 2.9981570e+02, 6.5956573e+01, 1.2308110e+02], dtype=float32)
```

### `approx=True`

```
min: array([ 4.42728177e-02,  3.72961313e-02, -1.00180114e-04, -1.39481146e-02, -1.81622729e-02,  2.40351334e+02,  2.41102325e+02, -1.85363255e-02,  1.85722783e-02], dtype=float32)
max: array([  1.0669258,   1.0974499,   1.1490365,   1.217688 ,   0.9447322, 328.44455  , 322.08057  ,   0.8553141,   1.1938525], dtype=float32)
mean: array([1.3747187e-01, 1.3059923e-01, 1.3965788e-01, 2.2442502e-01, 2.1674435e-01, 2.9580356e+02, 2.9580475e+02, 1.4759095e-01, 1.7730184e-01], dtype=float32)
std: array([ 0.07018409,  0.08147372,  0.1101771 ,  0.12716562,  0.14547655, 11.659259  , 11.64575   ,  0.12016387,  0.10631502], dtype=float32)
```

experiments/ssl4eo/compute_dataset_statistics.py
```diff
@@ -0,0 +1,70 @@
...
+            out_list.append(compute(path))
+        out = np.array(out_list)
+
+    minimum = np.amin(out[:, :, 0], axis=0)
+    maximum = np.amax(out[:, :, 1], axis=0)
+
+    mu_d = out[:, :, 2]
+    mu = np.mean(mu_d, axis=0)
+    sigma_d = out[:, :, 3]
+    N_d = args.size**2
```
calebrob6, line 57: just compute this

calebrob6 (dismissed):

calebrob6 (dismissed):

PR #1301, opened 2023-05-02
RandomGrayscale: add new transform

adamjstewart:
Each of these images are scaled from 0–1 for plotting, so the brightness of the image may not be comparable.

avg
rgb
pca

### DIFF
--- a/tests/transforms/test_color.py
+++ b/tests/transforms/test_color.py
@@ -1,0 +1,58 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import pytest
+import torch
+from torch import Tensor
+
+from torchgeo.transforms import AugmentationSequential, RandomGrayscale
+
+
+@pytest.fixture
+def sample() -> dict[str, Tensor]:
+    return {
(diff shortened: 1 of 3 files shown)


calebrob6:
This seems tricky to use if you aren't steeped in geospatial and torch stuff. Can we provide reasonable defaults to weights somehow?

adamjstewart:
It's certainly harder to use than Kornia's RandomGrayscale, but I wouldn't describe it as harder to use than Kornia's Normalize.

It's possible to set a default for average/uniform weighting, but it isn't possible to set any other defaults without the user explicitly saying which bands are RGB or what the first principal component is. If we set the default to average/uniform weighting, we'll need to support scalar inputs (we won't know the # of channels until run-time) and move the rescaling to forward.

calebrob6:
That's why I said `.from_example_sample(ds[0], weighting='uniform')` where you pass an example that has the number of channels

adamjstewart:
But that won't work for RGB or PCA unless you want to give it a list of indices or the entire dataset. Idk, feels like a lot more code and would necessitate bringing sklearn back as a dep.

calebrob6 (approved):

PR #1303, opened 2023-05-03
Revert "PyVista: fix intersphinx link"

adamjstewart:
Reverts microsoft/torchgeo#1251

Seems like the docs were moved back??


isaaccorley (approved):

adamjstewart:
Okay it was fixed, we should be good. I would rather explicitly use stable so I'll close this.

PR #1306, opened 2023-05-03
PixelwiseRegressionTask pretrained weights

isaaccorley:
Similarly to #1046, this adds the ability to load pretrained encoder weights from our weight enums and checkpoint files to `PixelwiseRegressionTask`.


adamjstewart (approved):
Not sure if or where we should clarify that FCN and ViTs are not supported, but it's prob fine.

PR #1311, opened 2023-05-05
Add uniform sampling locations from CONUS script

nilsleh:
This PR adds a sampling script to generate uniformly sampled locations within conus to yield a csv table which can be used with the `download_ssl4eo.py` script. I tried integrating it but thought it is more understandable if it is self contained.

This is plot of running the script for 250 desired samples:
Screenshot from 2023-05-05 16-40-39

### DIFF
--- a/experiments/ssl4eo/sample_conus.py
+++ b/experiments/ssl4eo/sample_conus.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import argparse
+import csv
+import os
+import random
+
+import fiona
+from shapely.geometry import Point, shape
+from torchvision.datasets.utils import download_and_extract_archive
+
+
+def get_uniform_points_within_conus(
+    download_root: str, num_samples: int
+) -> list[tuple[float]]:
+    random.seed(0)
+    nation_url = (
+        "https://www2.census.gov/geo/tiger/GENZ2022/shp/cb_2022_us_nation_5m.zip"
+    )
+    nation_filename = "cb_2022_us_nation_5m.shp"
+    if not os.path.exists(os.path.join(download_root, os.path.basename(nation_url))):
+        download_and_extract_archive(nation_url, download_root)
+
+    state_url = "https://www2.census.gov/geo/tiger/GENZ2018/shp/cb_2018_us_state_5m.zip"
+    state_filename = "cb_2018_us_state_5m.shp"
+    if not os.path.exists(os.path.join(download_root, os.path.basename(state_url))):
+        download_and_extract_archive(state_url, download_root)
+
+    exclude_states = [
+        "United States Virgin Islands",
+        "Commonwealth of the Northern Mariana Island",
+        "Puerto Rico",
+        "Alaska",
+        "Hawaii",
+        "American Samoa",
+        "Guam",
+    ]
+    with fiona.open(os.path.join(download_root, state_filename), "r") as shapefile:
+        excluded = []
+        for feature in shapefile:
+            name = feature["properties"]["NAME"]
+            if name in exclude_states:
+                excluded.append(shape(feature["geometry"]))
+
+    with fiona.open(os.path.join(download_root, nation_filename), "r") as shapefile:
+        conus = shape(shapefile[0]["geometry"])
+        x_min, y_min, x_max, y_max = conus.bounds
+        points: list[tuple[float]] = []
+        while len(points) < num_samples:
+            x = random.uniform(x_min, x_max)
+            y = random.uniform(y_min, y_max)
+            point = Point(x, y)
+            if conus.contains(point) and not any(
+                [polygon.contains(point) for polygon in excluded]
+            ):
+                points.append((point.x, point.y))
+
+    return points
+
+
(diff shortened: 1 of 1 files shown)


experiments/ssl4eo/sample_conus.py
```diff
@@ -0,0 +1,89 @@
...
+        for feature in shapefile:
+            name = feature["properties"]["NAME"]
+            if name in exclude_states:
+                excluded.append(shape(feature["geometry"]))
+
+    with fiona.open(os.path.join(download_root, nation_filename), "r") as shapefile:
+        conus = shape(shapefile[0]["geometry"])
+        x_min, y_min, x_max, y_max = conus.bounds
+        points: list[tuple[float]] = []
+        while len(points) < num_samples:
```
adamjstewart, line 52: Can you wrap this with tqdm? It looks like it's possible even for while-loops: https://stackoverflow.com/a/45808255/5828163

experiments/ssl4eo/sample_conus.py
```diff
@@ -0,0 +1,89 @@
...
+import random
+
+import fiona
+from shapely.geometry import Point, shape
+from torchvision.datasets.utils import download_and_extract_archive
+
+
+def get_uniform_points_within_conus(
+    download_root: str, num_samples: int
+) -> list[tuple[float, ...]]:
```
adamjstewart, line 18: `...` is only for variable-length tuples. These all have static length, right?

experiments/ssl4eo/sample_conus.py
```diff
@@ -0,0 +1,89 @@
...
+    random.seed(0)
```
adamjstewart, line 19: I like this. We should probably add this to the other sampling script too. Either in this PR or a separate PR.

experiments/ssl4eo/sample_conus.py
```diff
@@ -0,0 +1,89 @@
...
+    nation_url = (
+        "https://www2.census.gov/geo/tiger/GENZ2022/shp/cb_2022_us_nation_5m.zip"
+    )
+    nation_filename = "cb_2022_us_nation_5m.shp"
+    if not os.path.exists(os.path.join(download_root, os.path.basename(nation_url))):
```
adamjstewart, line 24: No need for these if-statements, torchvision already checks to see if the file has already been downloaded and exits immediately if so: https://github.com/pytorch/vision/blob/v0.15.1/torchvision/datasets/utils.py#L127

experiments/ssl4eo/sample_conus.py
```diff
@@ -0,0 +1,89 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--save-path", type=str, default="./data/", help="dir to save data"
+    )
+    parser.add_argument(
+        "--num-samples",
+        type=int,
+        default=1000,
+        help="number of uniform samples to draw from conus",
+    )
```
adamjstewart, line 82: You're missing the `--size` and R-tree stuff from the other sampling script. We want to ensure that none of our patches overlap. Also might want to convert `--num-samples` to `--indices-range` in case we want to add more samples, although the script itself is fast enough that we could always just resample the entire thing.

adamjstewart (changes requested):
Biggest change is integrating R-trees to avoid overlap. Most of the rest of the script looks good.

adamjstewart:
@nilsleh what's the status of this? Would like to finish downloading by the end of the week

nilsleh:
I believe I addressed your requested changes, except for the tqdm bar, which doesn't work yet.

adamjstewart:
What's the issue with tqdm? Not important, but might not be hard to solve.

adamjstewart (commented):
Can you also add a `sample_conus.sh` script similar to the others?

wangyi111:
looks good except `--indices-range` (which also relates to tqdm stuff).

experiments/ssl4eo/sample_conus.py
wangyi111: See https://github.com/microsoft/torchgeo/pull/1318#issuecomment-1540891280, might solve the issue here.

nilsleh:
I tried adapting the script to #1318 now, but waiting for a decision on how to handle the random seed. Also added a bash script.

adamjstewart (changes requested):
I'm guessing you haven't tested the download script since it's missing a parameter. Would be good to make sure this actually works. These changes you can make regardless. We'll discuss in tomorrow's meeting if there's anything else we want to change.

experiments/ssl4eo/sample_conus.py
```diff
@@ -0,0 +1,121 @@
...
+        for feature in shapefile:
+            name = feature["properties"]["NAME"]
+            if name in excluded_states:
+                excluded.append(shape(feature["geometry"]))
+
+    with fiona.open(os.path.join(download_root, nation_filename), "r") as shapefile:
+        usa = shape(shapefile[0]["geometry"])  # still usa
+
+    excluded_polygons = unary_union(excluded)
+    conus = usa.difference(excluded_polygons)
```
adamjstewart, line 51: Instead of computing the union of the exclusions and removing them from USA, could we instead compute the union of CONUS and not even need the USA bounds?

adamjstewart (approved):

PR #1319, opened 2023-05-08
SSL4EO download script: fix type hints

adamjstewart:
Found that a lot of these type hints were wrong.

@wangyi111


wangyi111 (approved):
Forgive me😄. I'm also not confident in reviewing every type hint, trust your knowledge here😂

isaaccorley (approved):

PR #1320, opened 2023-05-10
Add script to delete mismatched scenes

adamjstewart:
Given directories of downloaded data, this script deletes any scenes that are not common between all datasets. This lets us build a parallel corpus for all satellites. We delete all locations that are not shared, and all times that are not shared between Level-1 and Level-2 within a single satellite (L7 and L8 are from different decades so it doesn't make sense to require the same dates).

I didn't try parallelizing this script. I'm not sure how long it will take to delete all files, but a dry run on 600K L8 files takes 3.5 min (99% of time spent on looping over all directories to compare dates).

Unfortunately, this script requires all datasets to be on one system. We have enough storage/inodes for this, but transferring everything to our cluster will be slow. Alternatively, we could support lists of files containing the same filesystem hierarchy.

### DIFF
--- a/experiments/ssl4eo/delete_mismatch.py
+++ b/experiments/ssl4eo/delete_mismatch.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python3
+
+import argparse
+import os
+import shutil
+
+
+def delete_scene(directories: list[str], scene_id: str) -> None:
+    """Delete scene_id from all directories.
+
+    Args:
+        directories: directories to check
+        scene_id: scene to delete
+    """
+    print(f"Removing {scene_id}")
+    for directory in directories:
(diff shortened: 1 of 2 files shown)


calebrob6:
Didn't really check this, going to assume it is correct based on you using it and looking at the output

calebrob6:
My only comment would be to write what the script does in the script (more/less embed your PR description)

adamjstewart:
If it was user-facing, I definitely would. Since it's restricted to us, I'm going to leave it as is for now. We can think about making it better documented via README or argparse if users start asking how to reproduce our paper. We also have bash scripts that automate most of the work, so knowing what the script does or how to use it isn't as important as it would be otherwise.

PR #1322, opened 2023-05-11
Typo in sample_conus.py script

nilsleh:
"Commonwealth of the Northern Mariana Island" was missing an S for "Islands".


adamjstewart (approved):
This would have taken me forever to spot, good job!

PR #1324, opened 2023-05-11
fix SeCo transforms

isaaccorley:
The current SeCo transforms do the following:

- resize
- center crop
- normalize using the quantiles
- multiply by 255
- normalize using imagenet stats

However in the [script](https://github.com/ServiceNow/seasonal-contrast/blob/8285173ec205b64bc3e53b880344dd6c3f79fa7a/datasets/seco_dataset.py) they actually do the following:

- resize
- center crop
- normalize using the quantiles
- multiply by 255
- clip to [0. 255]
- convert to uint8
- divide by 255
- normalize using imagenet stats

### DIFF
--- a/torchgeo/models/resnet.py
+++ b/torchgeo/models/resnet.py
@@ -39,7 +39,7 @@ _seco_transforms = AugmentationSequential(
     K.Normalize(mean=_min, std=_max - _min),
     K.Normalize(mean=torch.tensor(0), std=1 / torch.tensor(255)),
     Lambda(lambda x: torch.clamp(x, min=0.0, max=255.0)),
-    Lambda(lambda x: x.to(torch.uint8))
+    Lambda(lambda x: x.to(torch.uint8)),
     K.Normalize(mean=torch.tensor(0), std=torch.tensor(255)),
     K.Normalize(mean=_mean, std=_std),
     data_keys=["image"],
--- a/torchgeo/models/resnet.py
+++ b/torchgeo/models/resnet.py
@@ -8,6 +8,7 @@ from typing import Any, Optional
 import kornia.augmentation as K
 import timm
 import torch
+from kornia.contrib import Lambda
 from timm.models import ResNet
 from torchvision.models._api import Weights, WeightsEnum
 
@@ -37,6 +38,8 @@ _seco_transforms = AugmentationSequential(
     K.CenterCrop(224),
     K.Normalize(mean=_min, std=_max - _min),
     K.Normalize(mean=torch.tensor(0), std=1 / torch.tensor(255)),
+    Lambda(lambda x: torch.clamp(x, min=0.0, max=255.0)),
+    K.Normalize(mean=torch.tensor(0), std=torch.tensor(255)),
     K.Normalize(mean=_mean, std=_std),
     data_keys=["image"],
 )
--- a/torchgeo/models/resnet.py
+++ b/torchgeo/models/resnet.py
@@ -39,7 +39,7 @@ _seco_transforms = AugmentationSequential(
     K.Normalize(mean=_min, std=_max - _min),
     K.Normalize(mean=torch.tensor(0), std=1 / torch.tensor(255)),
     Lambda(lambda x: torch.clamp(x, min=0.0, max=255.0)),
-    Lambda(lambda x: x.to(torch.uint8).to(torch.float)),
+    Lambda(lambda x: x.to(torch.uint8).to(torch.float)),  # type: ignore[no-any-return]
     K.Normalize(mean=torch.tensor(0), std=torch.tensor(255)),
     K.Normalize(mean=_mean, std=_std),
     data_keys=["image"],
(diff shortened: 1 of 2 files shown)


torchgeo/models/resnet.py
```diff
@@ -37,6 +38,9 @@
     K.CenterCrop(224),
     K.Normalize(mean=_min, std=_max - _min),
     K.Normalize(mean=torch.tensor(0), std=1 / torch.tensor(255)),
+    Lambda(lambda x: torch.clamp(x, min=0.0, max=255.0)),
```
adamjstewart, line 41: I don't see a lot of value in this. All we're doing is losing information, right? We're already scaling to the same range, do we really care if there are values less/greater than what they trained on?

torchgeo/models/resnet.py
```diff
@@ -37,6 +38,9 @@
...
+    Lambda(lambda x: x.to(torch.uint8).to(torch.float)),  # type: ignore[no-any-return]
```
adamjstewart, line 42: Does this actually do anything?

torchgeo/models/resnet.py
```diff
@@ -37,6 +38,9 @@
...
+    K.Normalize(mean=torch.tensor(0), std=torch.tensor(255)),
```
adamjstewart, line 43: This is actually the exact opposite of what we want. This would divide by 255, not multiply. The above normalization is the correct one.

adamjstewart (changes requested):
I would actually disagree with all changes in this PR. You might be able to convince me to keep the clamp if you can show evidence that it actually improves downstream accuracy.

torchgeo/models/resnet.py
isaaccorley: To be clear, I'm not agreeing with this, but these are the transforms used for SeCo. @calebrob6 and I have done some KNN experiments with SeCo and it's actually very sensitive to these transforms (significant changes in downstream performance if you don't use these).  https://github.com/ServiceNow/seasonal-contrast/blob/8285173ec205b64bc3e53b880344dd6c3f79fa7a/datasets/seco_dataset.py#L51

torchgeo/models/resnet.py
isaaccorley: Yes it converts from float to uint8 to reduce bit resolution. The conversion back to float is so that the other kornia augmentations don't complain about it being float

torchgeo/models/resnet.py
isaaccorley: `transforms.ToTensor(),` right before `imagenet_normalization()`

isaaccorley:
Whether you agree or not with the changes, our current seco transforms are incorrect. I'm just taking what they have from their script.

isaaccorley:
I don't actually know how to fix the mypy errors other than ignoring them.
- I can't type hint a lambda but kornia doesn't have a clip augmentation.
- `K.AugmentationSequential` complains about the `K.contrib.Lambda` input arg

torchgeo/models/resnet.py
adamjstewart: This feels like the same situation as the clip above. It reduces information from the image and makes the job harder for downstream tasks. This may help during SSL because image comparison is harder, but I have no a priori reason to believe that this would help with downstream tasks. If you perform an ablation study and find that it helps, I'm fine with adding it.

isaaccorley: I agree with you that it's bad to throw away information, but we should make another transform function if we want to remove certain pieces. Our current seco transform isn't correct which is what we are advertising by attaching it to the weights.

torchgeo/models/resnet.py
adamjstewart: Good catch. In that case, I would remove all 4 lines, so we neither multiply by 255 nor divide by 255. This will make the transform faster.

adamjstewart:
> Whether you agree or not with the changes, our current seco transforms are incorrect. I'm just taking what they have from their script.

In my mind, the transforms supplied with the weights aren't designed to reproduce SeCo pre-training, they're designed to use the SeCo weights for downstream tasks. If these changes don't improve performance on downstream tasks and only make transforms slower, or even harm performance on downstream tasks, then we shouldn't add them. That's why an ablation study is needed to ensure that these changes are beneficial.

isaaccorley:
In that case then the transforms are completely wrong, because SeCo doesn't use these for downstream tasks at all. They only divide eurosat by 255, they use different BigEarthNet mean and std. They also don't resize to 224x224. I would argue that you would need to show an ablation study of why you shouldn't use the same transforms as those used during training since that's less intuitive. It's trained to take 8-bit inputs so I would expect performance would degrade if you feed it 12-bit inputs.

adamjstewart:
Let's hold off on this PR until after the deadline so we have time to do a proper ablation study on no transforms, current transforms, and proposed transforms.

isaaccorley:
I'm good with this just wanted to use it for comparisons to SeCo

adamjstewart:
Added this to the 0.4.2 milestone so we don't forget about it. At the bare minimum, we should remove the multiply by 255 stuff. The rest is more controversial and will need ablation studies that we don't have time to work on until after the deadline.

adamjstewart:
Is this still a work in progress?

isaaccorley:
@adamjstewart I can finish this today. It fell off my radar.

adamjstewart:
No rush. If we can squeeze it into 0.5.2, great. If not, we can continue to punt it. Just trying to finish up old PRs.

adamjstewart:
I'm no longer opposed to the use of `torch.clamp` and added a Kornia-style `_Clamp` transform in #2038 you can use.

isaaccorley:
Closing because this PR is very out of sync with the main branch. Would be better to just start a fresh one.

PR #1325, opened 2023-05-11
GASSL ResNet50 Weights

isaaccorley:
This PR adds the ResNet50 RGB weights trained on fMoW (mostly Maxar WorldView imagery) using the  [Geography Aware Self-supervised Learning (GASSL)](https://arxiv.org/abs/2011.09980) method. This is basically a modified MoCo-v2 with Geolocation as a pretext task (Geo) as well as temporal pairs contrastive learning (TP).

Weights taken from [this repo](https://github.com/sustainlab-group/geography-aware-ssl) and uploaded to huggingface.

Adding these weights is essential since it's being compared as a baseline in both the [SatMAE](https://arxiv.org/abs/2207.08051) and [Scale-MAE](https://arxiv.org/abs/2212.14532) papers.

### DIFF
--- a/torchgeo/models/resnet.py
+++ b/torchgeo/models/resnet.py
@@ -41,6 +41,17 @@ _seco_transforms = AugmentationSequential(
     data_keys=["image"],
 )
 
+# Normalization only available for RGB dataset, defined here:
+# https://github.com/sustainlab-group/geography-aware-ssl/blob/main/moco_fmow/main_moco_geo%2Btp.py#L287  # noqa: E501
+_mean = torch.tensor([0.485, 0.456, 0.406])
+_std = torch.tensor([0.229, 0.224, 0.225])
+_gassl_transforms = AugmentationSequential(
+    K.Resize(224),
+    K.Normalize(mean=torch.tensor(0), std=torch.tensor(255)),
+    K.Normalize(mean=_mean, std=_std),
+    data_keys=["image"],
+)
+
 # https://github.com/pytorch/vision/pull/6883
 # https://github.com/pytorch/vision/pull/7107
 # Can be removed once torchvision>=0.15 is required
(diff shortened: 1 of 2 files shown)


torchgeo/models/resnet.py
```diff
@@ -170,6 +170,19 @@ class ResNet50_Weights(WeightsEnum):  # type: ignore[misc]
         },
     )
 
+    FMOW_RGB_GASSL = Weights(
+        url="https://huggingface.co/torchgeo/resnet50_fmow_rgb_gassl/resolve/main/resnet50_fmow_rgb_gassl-44b4461b.pth",  # noqa: E501
```
adamjstewart, line 174: This is the MD5 checksum. torch.hub.download_url uses the SHA256 checksum. Can you change it to da43d987? You'll also need to rename the file on HF.

torchgeo/models/resnet.py
```diff
@@ -170,6 +170,19 @@ class ResNet50_Weights(WeightsEnum):  # type: ignore[misc]
         },
     )
 
+    FMOW_RGB_GASSL = Weights(
```
adamjstewart, line 173: I would put this first since it comes first alphabetically

adamjstewart (changes requested):
Need to add to `docs/api/resnet_pretrained_weights.csv`. We should start thinking about how we want to handle multiple satellites. I'm thinking we'll want different tables for each satellite since the benchmark datasets will differ. We can punt that to the PR where we add all our new SSL4EO-L models.

isaaccorley:
Should I add the transforms as well? They only train on the fmow RGB data which is uint8 and they just divide by 255 and use imagenet normalization.

isaaccorley:
No idea why coverage is failing

calebrob6:
This LGTM, @adamjstewart any problems with merging this?

adamjstewart:
Give me some time to review. How urgent is this? Would prefer to focus on the paper until the deadline, but can review sooner if needed. I mostly want to verify the transforms and filename and such.

calebrob6:
I've checked the sha256sum and transforms so I think we're good
image

image

adamjstewart:
Do we want to call it fMoW or Maxar? Might be more clear what the weights can be used for with the latter.

calebrob6:
The current name fits with the existing scheme `{model}_{dataset}_{bands}_{paper/source}_{hash}` -- regardless, users aren't interacting with the actual file.

calebrob6 (approved):

isaaccorley:
Fmow RGB is more than just Maxar though.

adamjstewart (approved):

adamjstewart:
Why did you choose Apache 2.0 for the license?

calebrob6:
Where are you seeing "Apache 2.0"?

calebrob6:
It seems the weights aren't licensed -- but just to be clear, we are _not_ uploading the weights to this repo. It is always good to ask the authors though.

isaaccorley:
I'll change the license on HF

calebrob6:
Already did

PR #1326, opened 2023-05-11
Add script to compress dataset files

adamjstewart:
This is hopefully the final script we need for dataset preprocessing. It allows us to map from the raw float32 files we downloaded to uint8 files while still capturing the majority of information. Reflectance bands were scaled using GEE visualization parameters. Temperature bands were scaled using NASA visualization parameters. Validation was done by computing statistics on the new datasets and visualizing the resulting bands in QGIS. Dataset size reduction is on the order of 15% of the original size.

### DIFF
--- a/experiments/ssl4eo/compress_dataset.py
+++ b/experiments/ssl4eo/compress_dataset.py
@@ -31,7 +31,7 @@ if __name__ == "__main__":
     args.min = np.array(args.min)[:, np.newaxis, np.newaxis]
     args.max = np.array(args.max)[:, np.newaxis, np.newaxis]
 
-    def compress(src_path: str) -> "np.typing.NDArray[np.float_]":
+    def compress(src_path: str) -> None:
         """Rescale, convert to uint8, and compress an image.
 
         Args:
--- a/experiments/ssl4eo/compress_dataset.py
+++ b/experiments/ssl4eo/compress_dataset.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import argparse
+import glob
+import os
+
+import numpy as np
+import rasterio as rio
+from tqdm import tqdm
+from tqdm.contrib.concurrent import thread_map
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    # Can be same directory for in-place compression
+    parser.add_argument("src_dir", help="directory to recursively search for files")
+    parser.add_argument("dst_dir", help="directory to save compressed files in")
+    parser.add_argument("--suffix", default=".tif", help="file suffix")
+    # Could be min/max, 2%/98%, mean ± 2 * std, etc.
+    parser.add_argument(
+        "--min", nargs="+", type=float, required=True, help="minimum range"
+    )
+    parser.add_argument(
+        "--max", nargs="+", type=float, required=True, help="maximum range"
+    )
+    parser.add_argument("--num-workers", type=int, default=10, help="number of threads")
+    args = parser.parse_args()
+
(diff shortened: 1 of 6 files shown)


calebrob6:
Do these need to be compressed? Going from 12 or 14 bit (don't remember which) to 8 bit removes a lot of information.

adamjstewart:
> Do these need to be compressed? Going from 12 or 14 bit (don't remember which) to 8 bit removes a lot of information.

We're going from float32 to uint8, it's a much bigger compression. ImageNet and CIFAR are also uint8. The SSL4EO-S12 paper actually experimented with this and found that uint8 resulted in better pre-trained models than uint16, see Appendix C1 and Table XVII of https://arxiv.org/pdf/2211.07044.pdf.

adamjstewart:
Histograms:

### L7 L1

l7-l1

### L7 L2

l7-l2

### L8 L1

l8-l1

### L8 L2

l8-l2

Script used to generate in case I need to redo:
```python
import glob

import numpy as np
import rasterio as rio

data = []
for file in glob.iglob("data/ssl4eo-l7-l2-v2/imgs/*/*/all_bands.tif"):
    with rio.open(file) as f:
        x = f.read().flatten()
        data.append(x)

plt.hist(data, bins=256, range=(0, 255))
plt.show()
```
This was for all 644 scenes I have downloaded on my laptop. I can make separate plots per band or for a smaller set of images if you prefer.

calebrob6:
Just over 1 band, and I think this is doing something weird, can you do `f.read(1)` and `data = np.concatenate(data, axis=0)`, then maybe log scale the yaxis?

calebrob6 (approved):

calebrob6:
> We're going from float32 to uint8, it's a much bigger compression. ImageNet and CIFAR are also uint8. The SSL4EO-S12 paper actually experimented with this and found that uint8 resulted in better pre-trained models than uint16, see Appendix C1 and Table XVII of https://arxiv.org/pdf/2211.07044.pdf.

They are reporting best performance with unnormalized uint8 values as input to the network?

adamjstewart:
I believe so, @wangyi111?

calebrob6:
> I think this is doing something weird

The different colors on the plots make me think it is plotting a histogram for each image separately. You want, e.g., a single histogram of all the values in the red band.

wangyi111:
> They are reporting best performance with unnormalized uint8 values as input to the network?

Yes, but I won't say that's the "best". I would go for "similar", while uint8 would save tons of storage. 

On the other hand, I checked only land cover classification task though. The raw float32 reflectance values might be useful for specific fine-grained tasks, which I am not so familiar.

wangyi111:
> Dataset size reduction is on the order of 15% of the original size.

Why only 15%? from 32bit to 8bit I suppose much more compression.

adamjstewart:
I don't mean 100% to 85%, I mean 100% to 15%. So 75% decrease due to float32 -> uint8 and another 10% due to LZW compression.

isaaccorley:
Running this doesn't actually print any progress bar for me. Is there something else we should be passing to `thread_map` to show progress?

adamjstewart:
That's weird, it prints a progress bar for me. I'm using tqdm 4.65.0.

adamjstewart:
@calebrob6 you were correct, here are the correct histograms for B1:

### L7 L1

l7-l1

### L7 L2

l7-l2

### L8 L1

l8-l1

### L8 L2

l8-l2

isaaccorley:
Nevermind about the progress bar. It acts different inside my vscode terminal but in a regular terminal it prints fine.

calebrob6 (approved):

calebrob6:
No changes, but from the histograms you can see the values bunched up over a relatively small range. I'm guessing this is a place where you could get incremental improvement by playing with the percentiles you use for min and max (as it would spread these distributions out).

Issue #1245, opened 2023-04-14
Concerns with GridGeoSampler for evaluation


adamjstewart:
A couple of issues that may affect the usage of GridGeoSampler for benchmarking.

### Overlapping patches

Back in #630, we modified GridGeoSampler to ensure that every part of the image is sampled from, even if the height/width of the image is not a multiple of the stride. At the time, I decided that we should adjust the stride of the last row/col in order to avoid sampling outside of the bounds of the image. In hindsight, I think this was a mistake.

The problem is that we end up with the last row/col overlapping with the second-to-last row/col, resulting in some areas being double counted when computing performance metrics. It also makes stitching together prediction patches unnecessarily complicated.

I think we should modify GridGeoSampler to avoid adjusting stride and instead sample outside the bounds of the image. I believe rasterio will simply return nodata pixels for areas outside of the image. @remtav are you okay with this solution? I believe this was actually the first idea you implemented, apologies for pushing that PR in the wrong direction.

Technically this issue also occurs when multiple images in the dataset intersect, but this is harder to mitigate without storing all predictions in one giant tensor and computing performance only on the final predicted mask. I think we would run out of memory very quickly.

### `ignore_index` weighting

This one may also affect training for other GeoSamplers as well, although I'm most concerned about evaluation.

When sampling from large tiles, many patches will contain partial or complete nodata pixels. TorchMetrics allows us to ignore these areas using `ignore_index`. However, it's unclear to me if all patches are weighted equally when computing the final performance metrics with Lightning. Ideally, the overall reported accuracy would match regardless of whether we chip up the image into small patches or if we compute accuracy on the entire image/mask in one go.

We could peruse the internals of TorchMetrics and Lightning, but I think it's actually easier to construct a toy example to determine whether or not this issue occurs. Consider an image with width 200 and height 100. Let the first 99 columns of the ground truth mask be 0, the 100th column be 1, and the last 100 columns be 2. Let the predicted mask be a tensor of all 1s. If we use a GridGeoSampler with size 100 and stride 100, and let `ignore_index=0`, the correct performance should be ~1%. If the actual reported performance is 50%, we'll know we have an issue.

adamjstewart:
@calebrob6 any updates on this?

remtav:
With some delay, no trouble with this on my end. Thanks for asking.

adamjstewart:
Closed by #1329 and #1331

FogDrip:
[...]

PR #1329, opened 2023-05-12
GridGeoSampler: don't change stride of last patch

adamjstewart:
Partially reverts #630. See #1245 for motivation.

@remtav

### DIFF
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -487,6 +487,7 @@ class RasterDataset(GeoDataset):
                 indexes=band_indexes,
                 out_shape=out_shape,
                 window=from_bounds(*bounds, src.transform),
+                boundless=True,
             )
         else:
             dest, _ = rasterio.merge.merge(
--- a/tests/samplers/test_single.py
+++ b/tests/samplers/test_single.py
@@ -162,7 +162,16 @@ class TestGridGeoSampler:
     @pytest.fixture(
         scope="function",
         params=product(
-            [(8, 1), (6, 2), (4, 3), (2.5, 3), ((8, 6), (1, 2)), ((6, 4), (2, 3))],
+            [
+                (8, 1),
+                (6, 2),
+                (4, 3),
+                (4, 4),
+                (2, 4),
+                (2.5, 3),
+                ((8, 6), (1, 2)),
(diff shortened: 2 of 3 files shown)


isaaccorley:
Are we sure this works? I thought that to sample outside the bounds of an image we have to pass `boundless=True` like

```python
dest = src.read(
   indexes=band_indexes,
   out_shape=out_shape,
   window=from_bounds(*bounds, src.transform),
   boundless=True
)
```

adamjstewart:
Good catch! It returns an image but crops it to the bounds of the file unless boundless=True is added. I'll fix this.

torchgeo/datasets/geo.py
```diff
@@ -487,6 +487,7 @@ def _merge_files(
                 indexes=band_indexes,
                 out_shape=out_shape,
                 window=from_bounds(*bounds, src.transform),
+                boundless=True,
             )
         else:
             dest, _ = rasterio.merge.merge(
```
adamjstewart, line 493: `merge` has no `boundless` parameter, but doesn't need one. I confirmed that it properly samples outside the bounds of the file and uses nodata pixels for padding.

calebrob6 (approved):

adamjstewart:
This introduces a serious bug. IntersectionDatasets with different CRSs result in the following error when loading data:
```
  File "/Users/Adam/torchgeo/test.py", line 11, in <module>
    ds[ds.bounds]
  File "/Users/Adam/torchgeo/torchgeo/datasets/geo.py", line 899, in __getitem__
    samples = [ds[query] for ds in self.datasets]
  File "/Users/Adam/torchgeo/torchgeo/datasets/geo.py", line 899, in <listcomp>
    samples = [ds[query] for ds in self.datasets]
  File "/Users/Adam/torchgeo/torchgeo/datasets/geo.py", line 443, in __getitem__
  File "/Users/Adam/torchgeo/torchgeo/datasets/geo.py", line 486, in _merge_files
    dest = src.read(
  File "rasterio/_warp.pyx", line 1233, in rasterio._warp.WarpedVRTReaderBase.read
ValueError: WarpedVRT does not permit boundless reads
```
The fix is to use `rasterio.merge.merge` for all I/O instead.

PR #1333, opened 2023-05-14
MonkeyPatch exposed in public API in newer pytest

adamjstewart:
See https://docs.pytest.org/en/7.1.x/changelog.html#id142

Unfortunately, SubRequest is not yet public: https://github.com/pytest-dev/pytest/issues/7469, https://github.com/pytest-dev/pytest/issues/8073, https://github.com/pytest-dev/pytest/issues/8763


calebrob6 (approved):
Nice :), although this is the type of thing that can definitely wait until July!

PR #1335, opened 2023-05-14
SSL4EO download script: optionally download nodata pixels

adamjstewart:
This PR partially reverts #1310.

During downloading, we discovered that ignoring images containing nodata pixels slightly reduces the number of L8 L1 images we download, and drastically reduces the number of L8 L2 images we download. Visual inspection of the L8 L2 images we were skipping shows that those images are perfectly fine. I'm guessing they were removed because the L1 -> L2 process resulted in values that were clipped, or there's a 1 pixel border of nodata pixels, or something like that.

_If_ we need to download more images (e.g., for SSL4EO-L v2 or SSL4EO-M), we'll want to use a different strategy:

1. Download TOA for L5, L7, L8, filtering clouds and nodata pixels
2. Delete all scenes without matching locations
3. Repeat 1 and 2 until we have 250K+ scenes
4. Download SR for L5, L8, from the same locations as TOA, ignoring clouds and nodata pixels
5. Delete all scenes without matching dates

This will result in a parallel corpus requiring the least amount of downloads and samples to search, while still ensuring that all of our imagery avoids large patches of nodata pixels.

### DIFF
--- a/experiments/ssl4eo/download_l8_l2.sh
+++ b/experiments/ssl4eo/download_l8_l2.sh
@@ -22,6 +22,7 @@ YEAR=2022
 BANDS=(SR_B1 SR_B2 SR_B3 SR_B4 SR_B5 SR_B6 SR_B7)
 ORIGINAL_RESOLUTIONS=30
 NEW_RESOLUTIONS=30
+DEFAULT_VALUE=0
 
 # Generic parameters
 SCRIPT_DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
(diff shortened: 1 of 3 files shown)


experiments/ssl4eo/download_l8_l2.sh
```diff
@@ -22,6 +22,7 @@ YEAR=2022
 BANDS=(SR_B1 SR_B2 SR_B3 SR_B4 SR_B5 SR_B6 SR_B7)
 ORIGINAL_RESOLUTIONS=30
 NEW_RESOLUTIONS=30
+DEFAULT_VALUE=0
```
nilsleh, line 25: the python script says `default_value` is supposed to be `float` but this is integer, or do you not use floats in bash scripts?

nilsleh (approved):

PR #1336, opened 2023-05-15
Create matching CDL/NLCD chips for downstream tasks

nilsleh:
This PR adds a script to create matching mask chips for downstream evaluation tasks based on CONUS data. @adamjstewart 

Remaining Questions:
- what should be the proper directory structure for a sample, still the numerate/landsat_id/*.tif?
- how to add the no-data mask?

Issues:
- Currently all masks I am chipping are just plain zeros :/ and not sure why

Screenshot from 2023-05-15 11-19-47

### DIFF
--- a/experiments/ssl4eo/chip_landsat_downstream.py
+++ b/experiments/ssl4eo/chip_landsat_downstream.py
@@ -17,7 +17,7 @@ from tqdm import tqdm
 
 def retrieve_mask_chip(
     img_src: DatasetReader, mask_src: DatasetReader
-) -> np.typing.NDArray[np.float_]:
+) -> "np.typing.NDArray[np.float_]":
     """Retrieve the mask for a given landsat image.
 
     Args:
@@ -45,20 +45,25 @@ if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     # Can be same directory for in-place compression
     parser.add_argument(
-        "--landsat-dir", help="directory to recursively search for files"
+        "--landsat-dir", help="directory to recursively search for files", required=True
     )
-    parser.add_argument("--mask-path", help="path to downstream task mask to chip")
     parser.add_argument(
-        "--save-dir", help="directory where to save masks, defaults to 'landsat-dir'"
+        "--mask-path", help="path to downstream task mask to chip", required=True
+    )
+    parser.add_argument(
+        "--save-dir", help="directory where to save masks", required=True
     )
     parser.add_argument("--suffix", default=".tif", help="file suffix")
     args = parser.parse_args()
 
-    os.makedirs(args.save_dir, exist_ok=True)
-
-    paths = glob.iglob(
+    paths = glob.glob(
         os.path.join(args.landsat_dir, "**", f"all_bands{args.suffix}"), recursive=True
     )
+
+    if "nlcd" in args.mask_path:
+        layer_name = "nlcd"
+    else:
+        layer_name = "cdl"
 
     for img_path in tqdm(paths):
         with rasterio.open(img_path) as img_src, rasterio.open(
@@ -70,19 +75,20 @@ if __name__ == "__main__":
             # retrieve mask
             mask = retrieve_mask_chip(img_src, mask_src)
 
-            # directory structure mask <7-digit id>/<scene_id>/<cdl>_<year>.tif*.tif
-            digit_id = os.path.dirname(img_path).split("/")[-2]
-            scene_id = os.path.dirname(img_path).split("/")[-1]
+            # directory structure mask <7-digit id>/<scene_id>/<cdl>_<year>.tif
+            digit_id = img_path.split(os.sep)[-3]
+            scene_id = img_path.split(os.sep)[-2]
             year = scene_id.split("_")[-1][:4]
-            mask_dir = os.path.join(args.save_dir, digit_id, scene_id, year)
+            mask_dir = os.path.join(args.save_dir, digit_id, scene_id)
             os.makedirs(mask_dir, exist_ok=True)
 
             # write mask tif
             profile = img_src.profile
             profile["count"] = 1
+            profile["dtype"] = mask_src.profile["dtype"]
 
             with rasterio.open(
-                os.path.join(mask_dir, "mask.tif"), "w", **profile
+                os.path.join(mask_dir, f"{layer_name}_{year}.tif"), "w", **profile
             ) as dst:
                 dst.write(mask)
                 dst.write_colormap(1, mask_src.colormap(1))
--- a/experiments/ssl4eo/chip_landsat_downstream.py
+++ b/experiments/ssl4eo/chip_landsat_downstream.py
@@ -12,6 +12,7 @@ import rasterio
 from rasterio.io import DatasetReader
 from rasterio.vrt import WarpedVRT
 from rasterio.windows import from_bounds
+from tqdm import tqdm
 
 
 def retrieve_mask_chip(img_src: DatasetReader, mask_src: DatasetReader) -> np.ndarray:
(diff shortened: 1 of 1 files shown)


adamjstewart:
> what should be the proper directory structure for a sample, still the numerate/landsat_id/*.tif?

Landsat is:

```
*-l{5,7,8}-l{1,2}/<scene_id>/<data>/all_bands.tif
```

Since CDL/NLCD are annual and shared by multiple Landsat splits, it should be:
```
{cdl,nlcd}/<scene_id>/<year>/*.tif
```
This way we can store l7-l1, l8-l2, cdl, and nlcd in separate tarballs. We could further separate cdl/nlcd by year (l5-l1 will require a different year) but I think it's fine.

adamjstewart:
> how to add the no-data mask?

Actually, let's skip this for now. We don't want to include nodata pixels in the files since they are shared by both l7 and l8 (which doesn't have nodata pixels). We'll likely want to do this in the dataset instead. The idea will be to copy 0 pixels from the image to the mask. Technically this isn't correct because some 0 values will be real 0 values, not nodata. But I don't see another easy way of using -9999 while still keeping uint8 compression.

adamjstewart:
> Currently all masks I am chipping are just plain zeros :/ and not sure why

Weird. I would visualize images at each step of the pipeline and see where things break. Maybe the warping/indexing is incorrect and you're actually sampling from outside the US?

nilsleh:
> Weird. I would visualize images at each step of the pipeline and see where things break. Maybe the warping/indexing is incorrect and you're actually sampling from outside the US?

After the warping, the bounds actually line up so I think that step is correct, because calling `mask_src.statistics()` before and after the warping step shows the same statistics. But after the `mask_src.read()` the array is just zeros, so maybe the transform is doing something weird?

nilsleh:
*-l{5,7,8}-l{1,2}/<scene_id>/<data>/all_bands.tif

With the raw data download script we get paths like "ssl4eo-l7-l1/imgs/0000000/LE07_044030_20190815/all_bands.tif`  so is 0000000 your scene_id and `LE07_044030_20190815` data? Or do you cut out the 0000000 and add a `data` directory after?

adamjstewart:
Sorry, that was a typo. By data I meant date. So `LE07_044030_20190815` is the _date_, kinda. We'll cut out the `imgs` directory.

nilsleh:
So I found the "only zeros" error (used img_src.transform instead of mask_src.transform), so we get mask images now with proper values, however, inspecting them visually doesn't seem correct.

Looks like a potential match:
Screenshot from 2023-05-17 11-42-20

Not matching:

Screenshot from 2023-05-17 11-50-18

adamjstewart (changes requested):
I can confirm that the images and masks don't line up. If you run gdalinfo on the files or load them in QGIS, you can confirm that they don't overlap. I'll see if I can figure out why now.

experiments/ssl4eo/chip_landsat_downstream.py
```diff
@@ -0,0 +1,93 @@
...
+        os.makedirs(mask_year_dir, exist_ok=True)
+
+        # write mask tif
+        profile = img_src.profile
+        profile["count"] = 1
+
+        with rasterio.open(
+            os.path.join(mask_year_dir, "mask.tif"), "w", **profile
+        ) as dst:
+            dst.write(mask)
```
adamjstewart, line 93: Can you also copy the colormap? See https://github.com/microsoft/torchgeo/tree/main/tests/data#raster-data

adamjstewart:
Even after manually reprojecting CDL to the same CRS as Landsat, it still has the same issue. Going to keep debugging...

experiments/ssl4eo/chip_landsat_downstream.py
```diff
@@ -0,0 +1,93 @@
...
+
+    args = parser.parse_args()
+
+    mask_root_dir = os.path.join(args.landsat_dir, "masks")
+    os.makedirs(mask_root_dir, exist_ok=True)
+
+    # find all files in landsat dir
+    paths = sorted(
+        glob.glob(
+            os.path.join(args.landsat_dir, "imgs", "*", "*", f"*{args.suffix}"),
```
adamjstewart, line 59:
```suggestion
            os.path.join(args.landsat_dir, "**", f"*{args.suffix}"),
```
This allows it to work regardless of which directory is specified.

experiments/ssl4eo/chip_landsat_downstream.py
```diff
@@ -0,0 +1,93 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    # Can be same directory for in-place compression
+    parser.add_argument(
+        "--landsat_dir", help="directory to recursively search for files"
+    )
+    parser.add_argument("--mask_path", help="path to downstream task mask to chip")
+    parser.add_argument("--suffix", default=".tif", help="file suffix")
+    # masks will be added as separate band to the *landsat_dir*
```
adamjstewart, line 49: Would be better to add a new flag for save directory

experiments/ssl4eo/chip_landsat_downstream.py
```diff
@@ -0,0 +1,93 @@
...
+    )
+
+    mask_src = rasterio.open(args.mask_path)
+    img_crs = rasterio.open(paths[0]).crs
+
+    if mask_src.crs != img_crs:
+        mask_src = WarpedVRT(mask_src, crs=img_crs)
+
+    for img_path in paths:
+        img_src = rasterio.open(img_path)
```
adamjstewart, line 71: This should be a with-statement to avoid file handle leaks.

experiments/ssl4eo/chip_landsat_downstream.py
```diff
@@ -0,0 +1,93 @@
...
+    # find all files in landsat dir
+    paths = sorted(
+        glob.glob(
+            os.path.join(args.landsat_dir, "imgs", "*", "*", f"*{args.suffix}"),
+            recursive=True,
+        )
+    )
+
+    mask_src = rasterio.open(args.mask_path)
+    img_crs = rasterio.open(paths[0]).crs
```
adamjstewart, line 65: Should also be with-statements to avoid file handle leaks

adamjstewart:
Some important clues: 

Reprojecting CDL beforehand doesn't help, so it's not a WarpedVRT issue.

If I only chip a single image (one known to have different bounds than its mask), it has the correct bounds. So there could be some issue related to variable reuse. Unfortunately it's unrelated to the file handle leak issue.

experiments/ssl4eo/chip_landsat_downstream.py
adamjstewart: P.S. If you want to avoid too much indentation, you can combine these like so:
```python
with rio.open('foo.tif') as f, rio.open('bar.tif') as g, ... :
    ...
```

adamjstewart:
Another important clue: the mask that has the wrong location matches the location of another image. This supports the theory that a variable is leaking somewhere.

adamjstewart:
> Let's change this to `<7-digit id>/<scene_id>/<cdl>_<year>.tif` to match the image directory structure.

This is both the source of the bug and the fix. Not all `<scene_id>` are unique, so for samples whose `<scene_id>` matches, we were overwriting the previous correct mask with the new mask.

How did I find this? There were 2 scenes with bounds that matched each faulty mask. The incorrect mask location was always close to the correct location (same WRS path/row). There were fewer masks than there were images.

nilsleh:
brilliant!

experiments/ssl4eo/chip_landsat_downstream.py
```diff
@@ -0,0 +1,88 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    # Can be same directory for in-place compression
+    parser.add_argument(
+        "--landsat-dir", help="directory to recursively search for files"
+    )
+    parser.add_argument("--mask-path", help="path to downstream task mask to chip")
+    parser.add_argument(
+        "--save-dir", help="directory where to save masks, defaults to 'landsat-dir'"
```
adamjstewart, line 52: It does not default to anything. Also, you should add `required=True` to all of these since they aren't optional.

experiments/ssl4eo/chip_landsat_downstream.py
```diff
@@ -0,0 +1,88 @@
...
+        ) as mask_src:
+            if mask_src.crs != img_src.crs:
+                mask_src = WarpedVRT(mask_src, crs=img_src.crs)
+
+            # retrieve mask
+            mask = retrieve_mask_chip(img_src, mask_src)
+
+            # directory structure mask <7-digit id>/<scene_id>/<cdl>_<year>.tif*.tif
+            digit_id = os.path.dirname(img_path).split("/")[-2]
+            scene_id = os.path.dirname(img_path).split("/")[-1]
```
adamjstewart, line 75: Any reason to use os.path.dirname if you're going to split anyway? Also, you should use `img_path.split(os.sep)` so it also works on Windows. You only need to split once, then you can index the list twice.

experiments/ssl4eo/chip_landsat_downstream.py
```diff
@@ -0,0 +1,88 @@
...
+        with rasterio.open(img_path) as img_src, rasterio.open(
+            args.mask_path
+        ) as mask_src:
+            if mask_src.crs != img_src.crs:
+                mask_src = WarpedVRT(mask_src, crs=img_src.crs)
+
+            # retrieve mask
+            mask = retrieve_mask_chip(img_src, mask_src)
+
+            # directory structure mask <7-digit id>/<scene_id>/<cdl>_<year>.tif*.tif
```
adamjstewart, line 73:
```suggestion
            # directory structure mask <7-digit id>/<scene_id>/<cdl>_<year>.tif
```

experiments/ssl4eo/chip_landsat_downstream.py
```diff
@@ -0,0 +1,88 @@
...
+        "--landsat-dir", help="directory to recursively search for files"
+    )
+    parser.add_argument("--mask-path", help="path to downstream task mask to chip")
+    parser.add_argument(
+        "--save-dir", help="directory where to save masks, defaults to 'landsat-dir'"
+    )
+    parser.add_argument("--suffix", default=".tif", help="file suffix")
+    args = parser.parse_args()
+
+    os.makedirs(args.save_dir, exist_ok=True)
```
adamjstewart, line 57: Don't actually need this line since we makedirs below

experiments/ssl4eo/chip_landsat_downstream.py
```diff
@@ -0,0 +1,88 @@
...
+
+    paths = glob.iglob(
+        os.path.join(args.landsat_dir, "**", f"all_bands{args.suffix}"), recursive=True
+    )
+
+    for img_path in tqdm(paths):
```
adamjstewart, line 63: If I recall, tqdm doesn't work with iglob since it has no way of knowing how long the list is. Can you check?

experiments/ssl4eo/chip_landsat_downstream.py
```diff
@@ -0,0 +1,88 @@
...
+            # directory structure mask <7-digit id>/<scene_id>/<cdl>_<year>.tif*.tif
+            digit_id = os.path.dirname(img_path).split("/")[-2]
+            scene_id = os.path.dirname(img_path).split("/")[-1]
+            year = scene_id.split("_")[-1][:4]
+            mask_dir = os.path.join(args.save_dir, digit_id, scene_id, year)
+            os.makedirs(mask_dir, exist_ok=True)
+
+            # write mask tif
+            profile = img_src.profile
+            profile["count"] = 1
```
adamjstewart, line 82: Should also copy the dtype from mask_src in case they are different. Anything else we might want to copy?

experiments/ssl4eo/chip_landsat_downstream.py
```diff
@@ -0,0 +1,88 @@
...
+                mask_src = WarpedVRT(mask_src, crs=img_src.crs)
+
+            # retrieve mask
+            mask = retrieve_mask_chip(img_src, mask_src)
+
+            # directory structure mask <7-digit id>/<scene_id>/<cdl>_<year>.tif*.tif
+            digit_id = os.path.dirname(img_path).split("/")[-2]
+            scene_id = os.path.dirname(img_path).split("/")[-1]
+            year = scene_id.split("_")[-1][:4]
+            mask_dir = os.path.join(args.save_dir, digit_id, scene_id, year)
```
adamjstewart, line 77: I don't think we want a year dir, right? It should go into the filename.

adamjstewart (commented):
Let me know if you need help solving mypy issues.

experiments/ssl4eo/chip_landsat_downstream.py
nilsleh: You don't get the full progress bar, just a running number, but also the list of paths is not that long so probably okay to take glob.glob and have a full progress bar.

experiments/ssl4eo/chip_landsat_downstream.py
```diff
@@ -0,0 +1,94 @@
...
+        window=from_bounds(
+            query.left, query.bottom, query.right, query.top, mask_src.transform
+        ),
+    )
+    return mask_chip
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    # Can be same directory for in-place compression
```
adamjstewart, line 46: This comment can be removed

adamjstewart (approved):
LGTM, just test it and make sure it behaves as expected.

nilsleh:
I have run the script and randomly checked a few resulting pairs of images, they all align now and look okay, but maybe there should be another test to check what sort of inputs we are getting? because maybe some of the input images are crap?

adamjstewart:
All input images should be good. At least they will all be in the US and square and mostly be cloud free. You could try sorting images by average value and manually remove cloudy images if you're concerned. Could also copy the pixel qa band cloud pixels and convert those to fill values in the mask.

adamjstewart (approved):

PR #1337, opened 2023-05-15
Add year filter functionality CDL

nilsleh:
This PR adds `years` argument to select a number of years for which to download CDL or use CDL so the user doesn't have to download all 15 years. This is how we decided to implement it for NLCD as well.

### DIFF
--- a/torchgeo/datasets/cdl.py
+++ b/torchgeo/datasets/cdl.py
@@ -350,8 +350,8 @@ class CDL(RasterDataset):
             FileNotFoundError: if no files are found in ``root``
             RuntimeError: if ``download=False`` but dataset is missing or checksum fails
 
-        ..versionchanged:: 0.5
-            Added *years* parameter
+        .. versionadded:: 0.5
+           The *years* parameter.
         """
         assert set(years).issubset(self.md5s.keys()), (
             "CDL data product only exists for the following years: "
--- a/torchgeo/datasets/cdl.py
+++ b/torchgeo/datasets/cdl.py
@@ -3,6 +3,7 @@
 
 """CDL dataset."""
 
+import glob
 import os
 from typing import Any, Callable, Optional
 
@@ -375,15 +376,12 @@ class CDL(RasterDataset):
         exists = []
         for year in self.years:
             filename_year = self.filename_glob.replace("*", str(year))
-            pathname = os.path.join(
-                self.root, filename_year.removesuffix(".tif"), filename_year
-            )
-            if os.path.exists(pathname):
-                exists.append(True)
-            else:
-                exists.append(False)
+            pathname = os.path.join(self.root, "**", filename_year)
+            for fname in glob.iglob(pathname, recursive=True):
+                if not fname.endswith(".zip"):
+                    exists.append(True)
 
-        if all(exists):
+        if len(exists) == len(self.years):
             return
 
         # Check if the zip files have already been downloaded
--- a/torchgeo/datasets/cdl.py
+++ b/torchgeo/datasets/cdl.py
@@ -376,7 +376,7 @@ class CDL(RasterDataset):
         for year in self.years:
             filename_year = self.filename_glob.replace("*", str(year))
             pathname = os.path.join(
-                self.root, filename_year.strip(".tif"), filename_year
+                self.root, filename_year.removesuffix(".tif"), filename_year
             )
             if os.path.exists(pathname):
                 exists.append(True)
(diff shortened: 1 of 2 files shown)


torchgeo/datasets/cdl.py
```diff
@@ -325,6 +324,7 @@ def __init__(
         root: str = "data",
         crs: Optional[CRS] = None,
         res: Optional[float] = None,
+        years: list[int] = [2019],
```
adamjstewart, line 327: If we're going to default to something let's default to the most recent year

adamjstewart (changes requested):
Needs more coverage

torchgeo/datasets/cdl.py
```diff
@@ -363,15 +372,31 @@ def _verify(self) -> None:
...
         """
         # Check if the extracted files already exist
-        pathname = os.path.join(self.root, "**", self.filename_glob)
-        for fname in glob.iglob(pathname, recursive=True):
-            if not fname.endswith(".zip"):
-                return
+        exists = []
+        for year in self.years:
+            filename_year = self.filename_glob.replace("*", str(year))
+            pathname = os.path.join(self.root, filename_year)
```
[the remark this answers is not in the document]
nilsleh: The implementation here, is how it is currently done for `nlcd.py`.

torchgeo/datasets/cdl.py
```diff
@@ -347,7 +348,15 @@ def __init__(
         Raises:
             FileNotFoundError: if no files are found in ``root``
             RuntimeError: if ``download=False`` but dataset is missing or checksum fails
+
+        ..versionchanged:: 0.5
+            Added *years* parameter
```
adamjstewart, line 353:
```suggestion
        .. versionadded:: 0.5
           The *years* parameter.
```

torchgeo/datasets/cdl.py
```diff
@@ -363,15 +372,33 @@ def _verify(self) -> None:
...
+        exists = []
+        for year in self.years:
+            filename_year = self.filename_glob.replace("*", str(year))
+            pathname = os.path.join(
+                self.root, filename_year.removesuffix(".tif"), filename_year
+            )
+            if os.path.exists(pathname):
+                exists.append(True)
+            else:
+                exists.append(False)
```
adamjstewart, line 384:
```suggestion
            exists.append(os.path.exists(pathname))
```

torchgeo/datasets/cdl.py
```diff
@@ -363,15 +372,33 @@ def _verify(self) -> None:
...
         # Check if the extracted files already exist
-        pathname = os.path.join(self.root, "**", self.filename_glob)
-        for fname in glob.iglob(pathname, recursive=True):
-            if not fname.endswith(".zip"):
-                return
+        exists = []
+        for year in self.years:
+            filename_year = self.filename_glob.replace("*", str(year))
+            pathname = os.path.join(
+                self.root, filename_year.removesuffix(".tif"), filename_year
```
adamjstewart, line 379: We should keep the recursive glob so that people can specify a higher or lower root directory (e.g., `root=/path/to/2019_30m_cdls/` or `root=/path/`)

adamjstewart (changes requested):

torchgeo/datasets/cdl.py
```diff
@@ -363,15 +373,30 @@ def _verify(self) -> None:
             RuntimeError: if ``download=False`` but dataset is missing or checksum fails
         """
         # Check if the extracted files already exist
```
adamjstewart, line 375: This is **very** unimportant but I wonder if we should instead have a single for-loop over all versions and only download/extract the exact versions we need to. Imagine a situation where someone downloaded and extracted 2021, but then 2022 came out and they wanted to use that too. At the moment, Spack would try to re-download (torchvision will skip this for us) and re-extract the zip file for 2021 even though it's already extracted.

torchgeo/datasets/cdl.py
```diff
@@ -43,23 +43,23 @@ class CDL(RasterDataset):
     is_image = False
```
adamjstewart, line 43: I think the glob/regex will still pick up versions other than the ones requested if they are found on the system. You _could_ fix this by editing the `filename_regex` during `__init__`. We do something similar to this in our Landsat datasets where we patch `filename_glob` based on the bands requested.

adamjstewart (approved):
Remaining issues are very trivial. Feel free to merge this if you want to focus on different things.

PR #1338, opened 2023-05-15
SSL4EO Landsat Downstream Dataset/module CDL, NLCD

nilsleh:
This PR adds datasets and module for downstream evaluation of SSL methods.

Sensors:
- L7-L1
- L7-L2
- L8-L1
- L8-L2

Masks:
- CDL
- NLCD

This is a `NonGeoDataset` relying on datasets to be created with the help of #1336.

Example CDL TM-TOA:

Screenshot from 2023-05-24 22-33-21


torchgeo/datasets/ssl4eo_downstream_landsat.py
```diff
@@ -0,0 +1,195 @@
...
+
+    * input landsat image and single channel mask
+
+    Each patch has the following properties:
+
+    * 264 x 264 pixels
+    * Resampled to 30 m resolution (7920 x 7920 m)
+    * Single multispectral GeoTIFF file
+
+    .. versionadded:: 0.5
```
adamjstewart, line 35: We'll add citation info after we publish

torchgeo/datasets/ssl4eo_downstream_landsat.py
```diff
@@ -0,0 +1,195 @@
...
+import numpy as np
+import rasterio
+import torch
+from torch import Tensor
+
+from .geo import NonGeoDataset
+from .utils import check_integrity, extract_archive
+
+
+class SSL4EODownstream(NonGeoDataset):
```
adamjstewart, line 20: Trying to decide if we want 1 class (SSL4EO-L CDL/NLCD) or 2 (SSL4EO-L CDL and SSL4EO-L NLCD).

torchgeo/datasets/ssl4eo_downstream_landsat.py
```diff
@@ -0,0 +1,195 @@
...
+    * Single multispectral GeoTIFF file
+
+    .. versionadded:: 0.5
+    """
+
+    valid_input_sensors = ["l5-l1", "l7-l1", "l7-l2", "l8-l1", "l8-l2"]
+    valid_mask_products = ["cdl", "nlcd"]
+    valid_splits = ["train", "val", "test"]
+
+    image_root = "ssl4eo-*-conus"
```
adamjstewart, line 42: This will change but is fine for now

torchgeo/datasets/ssl4eo_downstream_landsat.py
```diff
@@ -0,0 +1,195 @@
...
+    Each patch has the following properties:
+
+    * 264 x 264 pixels
+    * Resampled to 30 m resolution (7920 x 7920 m)
+    * Single multispectral GeoTIFF file
+
+    .. versionadded:: 0.5
+    """
+
+    valid_input_sensors = ["l5-l1", "l7-l1", "l7-l2", "l8-l1", "l8-l2"]
```
adamjstewart, line 38: These should eventually match the SSL4EO-L dataset sensors/products

tests/data/ssl4eo_downstream_landsat/l5-l1/data.py
```diff
@@ -0,0 +1,116 @@
...
+    for mask_name in MASKS:
+        mask_dir = MASK_DIR.replace("*", mask_name)
+        if os.path.isdir(mask_dir):
+            shutil.rmtree(mask_dir)
+        os.makedirs(mask_dir, exist_ok=True)
+
+        for subdir in SUBDIRS:
+            mask_path = os.path.join(os.getcwd(), mask_dir, subdir[0], subdir[1])
+            os.makedirs(mask_path)
+            create_mask(os.path.join(mask_path, f"{mask_name}_2011.tif"))
```
adamjstewart, line 110: Other `data.py` files define a `FILENAME_HIERARCHY` and have a recursive file creation method. Not sure whether that would be easier here.

tests/data/ssl4eo_downstream_landsat/l5-l1/data.py
```diff
@@ -0,0 +1,116 @@
...
+        "blockysize": 1,
+        "tiled": False,
+        "compress": "lzw",
+        "interleave": "pixel",
+    }
+
+    Z = np.random.randint(low=0, high=255, size=(NUM_BANDS, SIZE, SIZE))
+
+    with rasterio.open(path, "w", **profile) as src:
+        src.write(Z)
```
adamjstewart, line 51: Would be nice if we could copy the cmap too, but not crucial

tests/data/ssl4eo_downstream_landsat/data.py
```diff
@@ -0,0 +1,194 @@
...
+SIZE = 36
+
+np.random.seed(0)
+
+FILENAME_HIERARCHY = Union[dict[str, "FILENAME_HIERARCHY"], list[str]]
+
+filenames: FILENAME_HIERARCHY = {
+    "tm_toa": {
+        "0000002": {
+            "LE07_172034_20010526": ["all_bands.tif"],
```
adamjstewart: These directory names should be different

tests/data/ssl4eo_downstream_landsat/data.py
```diff
@@ -0,0 +1,194 @@
...
+            #     key = key.split("_")[0] # only keep toa, etm, oil
+            path = os.path.join(directory, key)
+            os.makedirs(path, exist_ok=True)
+            create_mask_directory(path, value, mask_product)
+    else:
+        # Base case
+        for value in hierarchy:
+            path = os.path.join(directory, value)
+            year = years[path.split(os.sep)[1].split("_")[2]]
+            create_mask(path.replace("all_bands", f"{mask_product}_{year}.tif"))
```
adamjstewart: Lot of code duplication here. If you add the masks to the FILENAME_HIERARCHY and change the few things that are different based on filename, this shouldn't be necessary.

tests/data/ssl4eo_downstream_landsat/data.py
```diff
@@ -0,0 +1,187 @@
...
+from rasterio import Affine
+from rasterio.crs import CRS
+
+IMG_DIR = "ssl4eo-l5-l1-conus"
+MASK_DIR = "l5-*-2011"
+MASKS = ["cdl", "nlcd"]
+
+SUBDIRS = [("0000000", "LT05_045030_20110723"), ("0000001", "LT05_040032_20110805")]
+
+NUM_BANDS = 7
```
adamjstewart, line 22: These vars aren't used anymore

torchgeo/datasets/ssl4eo_downstream_landsat.py
adamjstewart: We should at least include L in the dataset name since it is only for Landsat. I'm hoping we don't have to do this kind of thing again in the future since other satellites have more downstream datasets already available.

torchgeo/datasets/ssl4eo_downstream_landsat.py
```diff
@@ -0,0 +1,224 @@
...
+    * Single multispectral GeoTIFF file
+
+    .. versionadded:: 0.5
+    """
+
+    valid_input_sensors = ["tm_toa", "etm_toa", "etm_sr", "oli_tirs_toa", "oli_sr"]
+    valid_mask_products = ["cdl", "nlcd"]
+    valid_splits = ["train", "val", "test"]
+
+    image_root = "ssl4eo_l_*_benchmark"
```
adamjstewart, line 41: Benchmark or downstream? We should be consistent with file/directory/class names to avoid confusion.

torchgeo/datasets/ssl4eo_downstream_landsat.py
```diff
@@ -0,0 +1,224 @@
...
+
+    year_dict = {
+        "tm_toa": 2011,
+        "etm_toa": 2019,
+        "etm_sr": 2019,
+        "oli_tirs_toa": 2019,
+        "oli_sr": 2019,
+    }
+
+    RGB_INDICES = {
```
adamjstewart, line 58: This should prob be lowercase. It's a class attribute, not a global variable.

torchgeo/datasets/ssl4eo_downstream_landsat.py
```diff
@@ -0,0 +1,224 @@
...
+        "etm_toa": [2, 1, 0],
+        "etm_sr": [2, 1, 0],
+        "oli_tirs_toa": [3, 2, 1],
+        "oli_sr": [3, 2, 1],
+    }
+
+    def __init__(
+        self,
+        root: str = "data",
+        input_sensor: str = "etm_toa",
```
adamjstewart, line 69: I would default to oli_sr, it's the most widely used (see Figure 1 in https://doi.org/10.1016/j.rse.2022.113195)

torchgeo/datasets/ssl4eo_downstream_landsat.py
```diff
@@ -0,0 +1,224 @@
...
+        transforms: Optional[Callable[[dict[str, Tensor]], dict[str, Tensor]]] = None,
+    ) -> None:
+        """Initialize a new SSL4EODownstream instance.
+
+        Args:
+            root: root directory where dataset can be found
+            input_sensor: one of ['etm_toa', 'etm_sr', 'oli_tirs_toa, 'oli_sr']
+            mask_product: mask target matched to input_sensor
+            transforms: a function/transform that takes input sample and its target as
+                entry and returns a transformed version
```
adamjstewart, line 81: Missing docs for some params

torchgeo/datasets/ssl4eo_downstream_landsat.py
```diff
@@ -0,0 +1,224 @@
...
+
+        Raises:
+            AssertionError: if ``input_sensor`` argument is invalid
+            AssertionError: if ``mask_product`` argument is invalid
+            AssertionError: if ``split`` argument is invalid
```
adamjstewart, line 86:
```suggestion
            AssertionError: if any arguments are invalid
```

torchgeo/datasets/ssl4eo_downstream_landsat.py
```diff
@@ -0,0 +1,224 @@
...
+        """Return the number of data points in the dataset.
+
+        Returns:
+            length of the dataset
+        """
+        return len(self.sample_collection)
+
+    def retrieve_sample_collection(self) -> list[tuple[str]]:
+        """Retrieve paths to samples in data directory."""
+        img_data_dir = self.image_root.replace("*", self.input_sensor)
```
adamjstewart, line 137: If we're only using `*` as something to replace, why not remove it and concatenate instead of replace?

torchgeo/datasets/ssl4eo_downstream_landsat.py
```diff
@@ -0,0 +1,224 @@
...
+        Returns:
+            a matplotlib Figure with the rendered sample
+        """
+        ncols = 2
+        image = (
+            sample["image"][self.RGB_INDICES[self.input_sensor]]
+            .permute(1, 2, 0)
+            .numpy()
+            / 255
+        )
```
adamjstewart, line 199: Can you split this into 2 lines so it doesn't get reformatted to 6 lines?

adamjstewart (changes requested):

tests/conf/ssl4eo_l_downstream_cdl.yaml
```diff
@@ -0,0 +1,20 @@
...
+  num_filters: 1
+  ignore_index: 0
+
+datamodule:
+  _target_: torchgeo.datamodules.SSL4EOLDownstreamDataModule
+  root: "tests/data/ssl4eo_downstream_landsat"
+  input_sensor: "tm_toa"
+  mask_product: "cdl"
+  batch_size: 2
+  num_workers: 0
```
adamjstewart, line 20: Can you add an EOL char before this EOF char? That will remove the red circle

tests/data/ssl4eo_downstream_landsat/data.py
nilsleh: Not sure what you mean here.

tests/data/ssl4eo_downstream_landsat/data.py
nilsleh: I tried something else before, but I thought this was more clear, even though there is code duplication.

tests/conf/ssl4eo_l_downstream_cdl.yaml
adamjstewart: In the files changed tab, not in the file. Also, you'll see it in the git diff, it will say "no EOL character" or something like that.

tests/data/ssl4eo_downstream_landsat/data.py
adamjstewart: TM is Landsat 5, not Landsat 7

nilsleh: But those are the 5 different names for the different sensors we have and I want to create fake data for all of them with one script like in the other `ssl4eo/data.py`.

tests/conf/ssl4eo_l_benchmark_nlcd.yaml
```diff
@@ -0,0 +1,20 @@
...
+  learning_rate_schedule_patience: 6
+  in_channels: 7
+  num_classes: 17
+  num_filters: 1
+  ignore_index: 0
+
+datamodule:
+  _target_: torchgeo.datamodules.SSL4EOLBenchmarkDataModule
+  root: "tests/data/ssl4eo_benchmark_landsat"
+  input_sensor: "tm_toa"
```
adamjstewart, line 17: Let's use a different sensor for at least one of the tests

tests/conf/ssl4eo_l_benchmark_cdl.yaml
```diff
@@ -0,0 +1,20 @@
+module:
+  _target_: torchgeo.trainers.SemanticSegmentationTask
+  loss: "ce"
+  model: "unet"
+  backbone: "resnet18"
+  weights: null
+  learning_rate: 1e-3
+  learning_rate_schedule_patience: 6
+  in_channels: 7
+  num_classes: 17
```
adamjstewart, line 10: CDL actually has far more classes than this, although it doesn't matter too much for the tests

tests/datasets/test_ssl4eo_benchmark_landsat.py
```diff
@@ -0,0 +1,135 @@
...
+def download_url(url: str, root: str, *args: str, **kwargs: str) -> None:
+    shutil.copy(url, root)
+
+
+class TestSSL4EOLBenchmark:
+    @pytest.fixture(
+        params=product(
+            ["tm_toa", "etm_toa", "etm_sr", "oli_tirs_toa", "oli_sr"],
+            ["cdl", "nlcd"],
+            ["train", "val", "test"],
```
adamjstewart, line 31: How fast is this? In other datasets we've just used zip (with items in shorter lists duplicated to make them all the same size) so that we only test 5 times instead of 5 * 2 * 3 times. If it only takes a few seconds then it doesn't really matter though.

tests/datasets/test_ssl4eo_benchmark_landsat.py
```diff
@@ -0,0 +1,135 @@
...
+        with pytest.raises(AssertionError):
+            SSL4EOLBenchmark(split="foo")
+
+    def test_invalid_input_sensor(self) -> None:
+        with pytest.raises(AssertionError):
+            SSL4EOLBenchmark(split="foo")
+
+    def test_invalid_mask_product(self) -> None:
+        with pytest.raises(AssertionError):
+            SSL4EOLBenchmark(split="foo")
```
adamjstewart, line 102: These aren't actually testing sensor/mask

torchgeo/datasets/ssl4eo_benchmark_landsat.py
adamjstewart, line 1: Let's actually call these files `ssl4eo_benchmark.py`. Our pre-training datasets are also stored in `ssl4eo.py`, with classes for each satellite. I don't envision us making benchmark datasets for other satellites, but you never know. If we do, we can store them in the same file and make an ABC to avoid code duplication.

torchgeo/datasets/ssl4eo_benchmark_landsat.py
```diff
@@ -0,0 +1,348 @@
...
+from .geo import NonGeoDataset
+from .utils import download_url, extract_archive
+
+
+class SSL4EOLBenchmark(NonGeoDataset):
+    """SSL4EO Landsat Benchmark Evaluation Dataset.
+
+    Dataset is intended to be used for evaluation of SSL techniques. Each
+    benchmark dataset consists of 25,000 images with corresponding land
+    cover classification masks.
```
adamjstewart, line 28: Should mention number of classes in one of the below bullet points. I forget how many there are for CDL but it's ~100 I think. I guess you'll figure that out once you finish the mapping code.

torchgeo/datasets/ssl4eo_benchmark_landsat.py
```diff
@@ -0,0 +1,348 @@
...
+        download_url(
+            self.url.format(self.img_dir_name, self.img_dir_name),
+            self.root,
+            md5=self.img_md5s[self.input_sensor] if self.checksum else None,
+        )
+        # download mask
+        download_url(
+            self.url.format(self.mask_dir_name, self.mask_dir_name),
+            self.root,
+            md5=self.mask_md5s[self.input_sensor.split("_")[0]][self.mask_product]
```
adamjstewart, line 230: Not sure why mypy is failing here. I would try calculating some of these variables first instead of doing everything on one line.

torchgeo/datasets/ssl4eo_benchmark_landsat.py
```diff
@@ -0,0 +1,348 @@
...
+
+    def __len__(self) -> int:
+        """Return the number of data points in the dataset.
+
+        Returns:
+            length of the dataset
+        """
+        return len(self.sample_collection)
+
+    def retrieve_sample_collection(self) -> list[tuple[str]]:
```
adamjstewart, line 272:
```suggestion
    def retrieve_sample_collection(self) -> list[tuple[str, str]]:
```
This will fix most other mypy errors

torchgeo/datamodules/ssl4eo_benchmark_landsat.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class SSL4EOLBenchmarkDataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the SSL4EO-L dataset.
+
+    .. versionadded:: 0.5
+    """
+
+    mean = 0
+    std = 255
```
adamjstewart, line 22: I believe these need to be tensors for older versions of Kornia (I fixed this in newer versions). Anyway, these are already the defaults in the base class, so you don't have to change them

torchgeo/datamodules/ssl4eo_benchmark_landsat.py
```diff
@@ -0,0 +1,64 @@
...
+        self.test_aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
+            K.CenterCrop(224),
+            data_keys=["image", "mask"],
+        )
+        self.predict_aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
+            K.CenterCrop(224),
+            data_keys=["image", "mask"],
+        )
```
adamjstewart, line 56: The dataset doesn't have a predict split so no need to add this

torchgeo/datamodules/ssl4eo_benchmark_landsat.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def setup(self, stage: str) -> None:
+        """Set up datasets.
+
+        Args:
+            stage: Either 'fit', 'validate', 'test', or 'predict'.
+        """
+        self.dataset = SSL4EOLBenchmark(**self.kwargs)
```
adamjstewart, line 64: I think we want the default base class function here. We need to pass in the split based on stage.

tests/datasets/test_ssl4eo_benchmark_landsat.py
nilsleh: Running this test file takes 10-11 seconds.

torchgeo/datamodules/ssl4eo_benchmark_landsat.py
adamjstewart: It could be because the fake data doesn't have enough examples to do the current split ratio without one split ending up with 0 images? Remind me to improve the error messages to clarify why the data loaders don't work, right now the message is really generic.

torchgeo/datamodules/ssl4eo_benchmark.py
```diff
@@ -0,0 +1,48 @@
...
+
+import kornia.augmentation as K
+
+from ..datasets import SSL4EOLBenchmark
+from ..transforms import AugmentationSequential
+from .geo import NonGeoDataModule
+
+
+class SSL4EOLBenchmarkDataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the SSL4EO-L dataset.
```
adamjstewart, line 16:
```suggestion
    """LightningDataModule implementation for the SSL4EO-L Benchmark dataset.
```
If this doesn't fit on one line we'll have to get creative to satisfy pydocstyle

torchgeo/datamodules/ssl4eo_benchmark.py
```diff
@@ -0,0 +1,48 @@
...
+
+    .. versionadded:: 0.5
+    """
+
+    def __init__(
+        self, batch_size: int = 64, num_workers: int = 0, **kwargs: Any
+    ) -> None:
+        """Initialize a new SSL4EOLDownstreamDataModule instance.
```
adamjstewart, line 24:
```suggestion
        """Initialize a new SSL4EOLBenchmarkDataModule instance.
```

torchgeo/datamodules/ssl4eo_benchmark_landsat.py
nilsleh: I created 3 data examples and monkeypatch the split as [1/3,1/3, 1/3], it was always working before.

torchgeo/datasets/cdl.py
adamjstewart, line 1: Can rebase/merge now that CDL PR has been merged

torchgeo/datasets/ssl4eo_benchmark.py
```diff
@@ -0,0 +1,369 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""Self-Supervised Learning for Earth Observation Landsat Benchmark Evaluation."""
```
adamjstewart, line 4:
```suggestion
"""Self-Supervised Learning for Earth Observation Benchmark Datasets."""
```
No longer specific to Landsat with the more general naming scheme. Could add other satellites in the future

torchgeo/datasets/ssl4eo_benchmark.py
```diff
@@ -0,0 +1,369 @@
...
+class SSL4EOLBenchmark(NonGeoDataset):
+    """SSL4EO Landsat Benchmark Evaluation Dataset.
+
+    Dataset is intended to be used for evaluation of SSL techniques. Each
+    benchmark dataset consists of 25,000 images with corresponding land
+    cover classification masks.
+
+    Dataset format:
+
+    * input landsat image and single channel mask
```
adamjstewart, line 32:
```suggestion
    * Input Landsat image and single channel mask
```

torchgeo/datasets/ssl4eo_benchmark.py
```diff
@@ -0,0 +1,369 @@
...
+        "etm_sr": [2, 1, 0],
+        "oli_tirs_toa": [3, 2, 1],
+        "oli_sr": [3, 2, 1],
+    }
+
+    split_percentages = [0.7, 0.15, 0.15]
+
+    ordinal_label_map = {"nlcd": NLCD.ordinal_label_map, "cdl": CDL.ordinal_label_map}
+
+    cmaps = {"nlcd": NLCD.cmap, "cdl": CDL.cmap}
```
adamjstewart, line 103: This is so nice

torchgeo/datasets/ssl4eo_benchmark.py
```diff
@@ -0,0 +1,369 @@
...
+
+        Raises:
+            RuntimeError: if ``download=False`` but dataset is missing or checksum fails
+        """
+        # Check if the extracted files already exist
+        img_pathname = os.path.join(self.root, self.img_dir_name, "**", "all_bands.tif")
+        exists = []
+        for fname in glob.iglob(img_pathname, recursive=True):
+            exists.append(True)
+            break
```
adamjstewart, line 182: You always break immediately if glob returns multiple results? Why not use:
```python
exists.append(bool(glob.glob(...)))
```

torchgeo/datasets/ssl4eo_benchmark.py
```diff
@@ -0,0 +1,369 @@
...
+            self.root,
+            self.mask_dir_name,
+            "**",
+            f"{self.mask_product}_{self.year_dict[self.input_sensor]}.tif",
+        )
+        for fname in glob.iglob(mask_pathname, recursive=True):
+            exists.append(True)
+            break
+        if exists and all(exists):
+            return
```
adamjstewart, line 193: Won't this be true if image is found or mask is found but also if only one or the other is found?

torchgeo/datasets/ssl4eo_benchmark.py
```diff
@@ -0,0 +1,369 @@
...
+        """Return the number of data points in the dataset.
+
+        Returns:
+            length of the dataset
+        """
+        return len(self.sample_collection)
+
+    def retrieve_sample_collection(self) -> list[tuple[str, str]]:
+        """Retrieve paths to samples in data directory."""
+        img_paths = glob.glob(
```
adamjstewart, line 275: Should we sort this to ensure reproducibility? Don't want to get a different order in the datamodule just because we ran `glob.glob` 3 different times and got 3 different orders. I don't think the order is controlled by np.random or random, it's whatever the OS returns.

torchgeo/datamodules/ssl4eo_benchmark_landsat.py
adamjstewart: You only monkey-patched the dataset, not the datamodule. Maybe we could make those params of the datamodule/dataset? Also, with a batch size of 2, you'll need at least 6 data examples with 1/3 splits.

nilsleh: batch_size is always a "max" or no? So it would at most take that many samples but if there is only less available it will  just take those.

nilsleh: should I just create more fake samples then so the test_segmentation script always has enough samples?

torchgeo/datamodules/ssl4eo_benchmark.py
```diff
@@ -0,0 +1,50 @@
...
+
+        Args:
+            batch_size: Size of each mini-batch.
+            num_workers: Number of workers for parallel data loading.
+            **kwargs: Additional keyword arguments passed to
+                :class:`~torchgeo.datasets.SSL4EOLBenchmark`.
+        """
+        super().__init__(SSL4EOLBenchmark, batch_size, num_workers, **kwargs)
+
+        self.train_aug = AugmentationSequential(
```
adamjstewart, line 36: If you use the same aug for all splits, you can just set `self.aug` instead. However, we might want different augs for train to improve performance.

torchgeo/datasets/ssl4eo_benchmark.py
```diff
@@ -0,0 +1,370 @@
...
+        exists.append(bool(glob.glob(mask_pathname, recursive=True)))
+        if all(exists):
+            return
+        # Check if the tar.gz files have already been downloaded
+        exists = []
+        img_pathname = os.path.join(self.root, f"{self.img_dir_name}.tar.gz")
+        if os.path.exists(img_pathname):
+            exists.append(True)
+        else:
+            exists.append(False)
```
adamjstewart, line 196:
```suggestion
        exists.append(os.path.exists(img_pathname))
```
Same below

torchgeo/datasets/ssl4eo_benchmark.py
```diff
@@ -0,0 +1,370 @@
...
+        exists.append(bool(glob.glob(img_pathname, recursive=True)))
+        mask_pathname = os.path.join(
+            self.root,
+            self.mask_dir_name,
+            "**",
+            f"{self.mask_product}_{self.year_dict[self.input_sensor]}.tif",
+        )
+        exists.append(bool(glob.glob(mask_pathname, recursive=True)))
+        if all(exists):
+            return
```
adamjstewart, line 189: Let's put a blank line between sections

torchgeo/datasets/ssl4eo_benchmark.py
```diff
@@ -0,0 +1,370 @@
...
+        return len(self.sample_collection)
+
+    def retrieve_sample_collection(self) -> list[tuple[str, str]]:
+        """Retrieve paths to samples in data directory."""
+        img_paths = sorted(
+            glob.glob(
+                os.path.join(self.root, self.img_dir_name, "**", "all_bands.tif"),
+                recursive=True,
+            )
+        )
```
adamjstewart, line 279: Could prob turn a lot of these into 2 lines just by defining the glob first then actually calling `sorted(glob.glob(..., recursive=True))` on the second line. Also helps to make things easier to read

torchgeo/datasets/ssl4eo_benchmark.py
```diff
@@ -0,0 +1,370 @@
...
+        ncols = 2
+        image = sample["image"][self.rgb_indices[self.input_sensor]].permute(1, 2, 0)
+        image = image.numpy() / 255
+
+        plt_cmap = ListedColormap(
+            np.stack(
+                [np.array(val) / 255 for val in self.cmaps[self.mask_product].values()],
+                axis=0,
+            )
+        )
```
adamjstewart, line 344: Forgot to say this during the CDL review, but don't we also need to map this to 134 classes before plotting? Should have asked you to add an example plot for CDL.

nilsleh: updated CDL example at the top. Each available class has an assigned color already, I don't think you need to remap.

nilsleh: I guess you would have to remap the ordinal values to then access the proper cmap values.

torchgeo/datamodules/ssl4eo_benchmark_landsat.py
adamjstewart: Yes, but 15% of 2 is 0.

torchgeo/datasets/ssl4eo_benchmark.py
adamjstewart: I meant between the return statement and the comment below

adamjstewart (approved):
I could nitpick for days but I think it's good enough now. Would rather merge it soon so people can start benchmarking. Feel free to merge if you also think it's good enough or address the remaining comments and I'll review again.

torchgeo/datamodules/ssl4eo_benchmark.py
nilsleh: Yes, I wrote them out because I thought it remains to be discussed whether we have additional training augs.

torchgeo/datasets/ssl4eo_benchmark.py
```diff
@@ -0,0 +1,366 @@
...
+    ordinal_label_map = {"nlcd": NLCD.ordinal_label_map, "cdl": CDL.ordinal_label_map}
+
+    cmaps = {"nlcd": NLCD.cmap, "cdl": CDL.cmap}
+
+    def __init__(
+        self,
+        root: str = "data",
+        input_sensor: str = "oli_sr",
+        mask_product: str = "cdl",
+        split: str = "train",
```
adamjstewart, line 110: In hindsight, I would have probably called these parameters `sensor` and `product` and removed the `split` parameter. We should probably do splitting in the data module to match our other datasets. But this can wait until 0.5, no rush to do it before the deadline.

adamjstewart: Also, I think we should put all datasets in one directory on HF to reduce the clutter.

nilsleh: also post deadline or want to do it now?

adamjstewart: Post deadline

PR #1339, opened 2023-05-15
Add empirical sampling mode to the RCF model

calebrob6:
The MOSAIKS paper describes RCF, which we've already implemented, and this method (which I'll call MOSAIKS). This method is the same setup as RCF but instead of initializing the conv kernels randomly, you sample patches from the dataset, do ZCA whitening, and use those as the kernels. In some little experiments on EuroSAT and UCM this gives +2 to 5% accuracy over simple RCF kernels.

### DIFF
--- a/torchgeo/models/rcf.py
+++ b/torchgeo/models/rcf.py
@@ -20,6 +20,10 @@ class RCF(Module):
     RCFs are used in Multi-task Observation using Satellite Imagery & Kitchen Sinks
     (MOSAIKS) method proposed in https://www.nature.com/articles/s41467-021-24638-z.
 
+    This class can operate in two modes, "gaussian" and "empirical". In "gaussian" mode,
+    the filters will be sampled from a Gaussian distribution, while in "empirical" mode,
+    the filters will be sampled from a dataset.
+
     .. note::
 
         This Module is *not* trainable. It is only used as a feature extractor.
@@ -35,6 +39,8 @@ class RCF(Module):
         kernel_size: int = 3,
         bias: float = -1.0,
         seed: Optional[int] = None,
+        mode: str = "gaussian",
+        dataset: Optional[NonGeoDataset] = None,
     ) -> None:
         """Initializes the RCF model.
 
@@ -44,19 +50,27 @@ class RCF(Module):
         .. versionadded:: 0.2
            The *seed* parameter.
 
+        .. versionadded:: 0.5.0
+            The *mode* and *dataset* parameters.
+
         Args:
             in_channels: number of input channels
             features: number of features to compute, must be divisible by 2
             kernel_size: size of the kernel used to compute the RCFs
             bias: bias of the convolutional layer
             seed: random seed used to initialize the convolutional layer
+            mode: "empirical" or "gaussian"
+            dataset: a NonGeoDataset to sample from when mode is "empirical"
         """
         super().__init__()
-
+        assert mode in ["empirical", "gaussian"]
+        if mode == "empirical" and dataset is None:
+            raise ValueError("dataset must be provided when mode is 'empirical'")
         assert features % 2 == 0
+        num_patches = features // 2
 
         if seed is None:
-            generator = None
+            generator = torch.Generator()
         else:
             generator = torch.Generator().manual_seed(seed)
 
@@ -67,7 +81,7 @@ class RCF(Module):
         self.register_buffer(
             "weights",
             torch.randn(
-                features // 2,
+                num_patches,
                 in_channels,
                 kernel_size,
                 kernel_size,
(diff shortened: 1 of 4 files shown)


isaaccorley (approved):
Just need to add to docs

calebrob6:
Lol need to figure out some type hints and stuff too

calebrob6:
probably tests too huh?

isaaccorley:
Nah who needs tests???

calebrob6:
Okay nowwww it LGTM

adamjstewart:
Will review post paper

estherrolf:
chiming in at Caleb's request on the naming of the option with Gaussian patches and the option with patches drawn from the dataset:

I would call both of these random convolutional features, but in the first the random patches are drawn from a Gaussian distribution and in the second from the empirical distribution of image patches. Thus, I would propose "RCF - Gaussian" and "RCF - Empirical" to distinguish between them, noting that the latter (RCF - empirical) is what what we use in the MOSAIKS paper. 

LMK if I can help clarify further!

isaaccorley:
Would RCF-ZCA or RCF-Empirical-ZCA be better since it's more explicit? I'm assuming we could probably use other techniques than ZCA to estimate the kernels using the dataset.

estherrolf:
Yeah using ZCA to whiten the images is an optional preprocessing step but we've found that it can help quite a bit. I still think the most straightforward would be to have one RCF class with arguments for patch_distribution as either "gaussian" or "empirical" and have the ZCA whitening step as an optional (but default?) processing step. Alternatively having two classes (RCF_Gaussian) and (RCF_empirical) and for both of them having ZCA whitening step as an argument make sense.

torchgeo/models/rcf.py
```diff
@@ -104,3 +107,111 @@ def forward(self, x: Tensor) -> Tensor:
             assert len(x1a.shape) == 2
             output = torch.cat((x1a, x1b), dim=1)
             return output
+
+
+class MOSAIKS(RCF):
+    """This model extracts MOSAIKS features from its input.
+
+    MOSAIKS features are described in Multi-task Observation using Satellite Imagery &
+    Kitchen Sinks https://www.nature.com/articles/s41467-021-24638-z. Briefly, this
```
adamjstewart, line 116: Should we use a link here? Would like the format of RCF and MOSAIKS to match. What if we instead use something like:
```rst

* https://www.nature.com/articles/s41467-021-24638-z
```

torchgeo/models/rcf.py
```diff
@@ -104,3 +107,111 @@ def forward(self, x: Tensor) -> Tensor:
...
+
+    def _normalize(
+        self,
+        patches: "np.typing.NDArray[np.float32]",
+        min_divisor: float = 1e-8,
+        zca_bias: float = 0.001,
+    ) -> "np.typing.NDArray[np.float32]":
+        """Does ZCA whitening on a set of input patches.
+
+        Copied from https://github.com/Global-Policy-Lab/mosaiks-paper/blob/7efb09ed455505562d6bb04c2aaa242ef59f0a82/code/mosaiks/featurization.py#L120
```
adamjstewart, line 133: Pretty sure https://github.com/Global-Policy-Lab/mosaiks-paper#use-of-code-and-data lets us do this but I'm no legal expert

torchgeo/models/rcf.py
```diff
@@ -104,3 +107,111 @@ def forward(self, x: Tensor) -> Tensor:
...
+        features: int = 16,
+        kernel_size: int = 3,
+        bias: float = -1.0,
+        seed: Optional[int] = None,
+    ):
+        """Initializes the MOSAIKS model.
+
+        This is the main model used in Multi-task Observation using Satellite Imagery
+        & Kitchen Sinks (MOSAIKS) method proposed in
+        https://www.nature.com/articles/s41467-021-24638-z.
```
adamjstewart, line 186: Probably don't need to reference this paper in both the class description and the init docstring.

torchgeo/models/rcf.py
```diff
@@ -104,3 +107,111 @@ def forward(self, x: Tensor) -> Tensor:
...
+
+        Args:
+            dataset: a torch dataset that returns dictionaries with an "image" key
+            in_channels: number of input channels
+            features: number of features to compute, must be divisible by 2
+            kernel_size: size of the kernel used to compute the RCFs
+            bias: bias of the convolutional layer
+            seed: random seed used to initialize the convolutional layer
+
+        .. versionadded:: 0.5.0
```
adamjstewart, line 196:
```suggestion
        .. versionadded:: 0.5
```
Also, this should be in the class docstring, not the `__init__` docstring. The entire class is new, not just the `__init__` method.

torchgeo/models/rcf.py
```diff
@@ -104,3 +107,111 @@ def forward(self, x: Tensor) -> Tensor:
...
+        """
+        super().__init__(in_channels, features, kernel_size, bias, seed)
+
+        # sample dataset patches
+        generator = np.random.default_rng(seed=seed)
```
adamjstewart, line 201: Not sure if we should convert all this numpy logic to torch or not. It's probably fine as is.

adamjstewart (commented):
Since we added RCF first, all files and documentation are named after RCF. I haven't read the paper, but my only question is whether we should rename things to be under the RCF or MOSAIKS umbrella.

torchgeo/models/rcf.py
calebrob6: Done

calebrob6:
@adamjstewart @estherrolf -- I combined the functionality of both under a single RCF class that takes `mode="empirical"` or `mode="gaussian"`. I actually like this way better than having two classes and, as usual, am thankful we do thorough peer review :). In particular, I can imagine us adding a few more parameters to this class in the future to allow for easy experimentation over dimensions like "normalization schemes". Having the code in one class is better for this.

torchgeo/models/rcf.py
```diff
@@ -73,9 +95,85 @@ def __init__(
             ),
         )
         self.register_buffer(
-            "biases", torch.zeros(features // 2, requires_grad=False) + bias
+            "biases", torch.zeros(num_patches, requires_grad=False) + bias
+        )
+
+        if mode == "empirical":
+            assert dataset is not None
```
adamjstewart, line 102: This line can be dropped, we already check this above

torchgeo/models/rcf.py
```diff
@@ -5,17 +5,29 @@
...
 from torch import Tensor
 from torch.nn.modules import Module
 
+from ..datasets import NonGeoDataset
+
 
 class RCF(Module):
     """This model extracts random convolutional features (RCFs) from its input.
 
     RCFs are used in Multi-task Observation using Satellite Imagery & Kitchen Sinks
```
adamjstewart, line 20: "in _the_ MOSAIKS method" sounds better than "in MOSAIKS method"

adamjstewart (commented):
I'm fine with the new organization

torchgeo/models/rcf.py
calebrob6: mypy complains

torchgeo/models/rcf.py
calebrob6, line 20:
```suggestion
    RCFs are used in the Multi-task Observation using Satellite Imagery & Kitchen Sinks
```

torchgeo/models/rcf.py
adamjstewart: Ah, because it's a string value thing, not catchable by a static type analyzer.

adamjstewart (approved):

PR #1343, opened 2023-05-17
Add L7 Irish experiment

yichiac:
Running experiments on training models for the L7Irish dataset.


experiments/torchgeo/run_l7irish_experiments.py
```diff
@@ -0,0 +1,75 @@
...
+            command = (
+                "python train.py"
+                + f" config_file={config_file}"
+                + f" experiment.name={experiment_name}"
+                + f" experiment.module.segmentation_model={model}"
+                + f" experiment.module.learning_rate={lr}"
+                + f" experiment.module.loss={loss}"
+                + f" experiment.module.weights={weights}"
+                + f" experiment.module.backbone={backbone}"
+                + f" experiment.datamodule.weights={weights}"
```
adamjstewart, line 59: There is no experiment level anymore, these should just be module and datamodule. Right now they aren't being used at all.

adamjstewart:
What's the status of this PR?

adamjstewart:
Don't think this is needed anymore

PR #1344, opened 2023-05-17
Fix reprojection issues

### LINKED ISSUE #1341: UnionDataset does not propagate .res
### Description

TLDR: setting .res on `UnionDataset` should set it for all children.

A problematic scenario: assume `a` is a raster dataset with res=1 and `b, c` are vector datasets with res=0.0001. We want a union with res=1. If we write
```python
ab = a | b
abc = ab | c
```
`b` will be cast to res of `a` and then `c` to resolution of `ab` -> all datasets have the same res=1. However if we write

```python
bc = b | c
abc = a | bc
```
we will create `bc` with res=0.0001 and then cast it to the res of `a`. But since `res` is just a member variable of `bc`, it doesn't run any code to propagate down to `b` and `c`. When we query resulting `abc` we will get answers at differing resolutions, which in my case crashes the kernel with an attempt of rasterizing `b` and `c` at ridiculously high resolutions.

### Steps to reproduce

I hope the analysis above is sufficient to understand the issue. I can provide a runnable repro if I am pointed to a way to reasonably generate fake data (I'm not sure how to do it best, such that the fake vector and raster datasets overlap and agree on metadata).

### Version

torchgeo==0.4.1

adamjstewart:
This PR fixes 2 issues related to reprojection:

- [x] RasterDataset: fix loading of files requiring reprojection

I broke this in #1329. When loading any file that requires changing the CRS (and therefore WarpedVRT), we see the following error:
```
  File "/Users/Adam/torchgeo/test.py", line 11, in <module>
    ds[ds.bounds]
  File "/Users/Adam/torchgeo/torchgeo/datasets/geo.py", line 899, in __getitem__
    samples = [ds[query] for ds in self.datasets]
  File "/Users/Adam/torchgeo/torchgeo/datasets/geo.py", line 899, in <listcomp>
    samples = [ds[query] for ds in self.datasets]
  File "/Users/Adam/torchgeo/torchgeo/datasets/geo.py", line 443, in __getitem__
  File "/Users/Adam/torchgeo/torchgeo/datasets/geo.py", line 486, in _merge_files
    dest = src.read(
  File "rasterio/_warp.pyx", line 1233, in rasterio._warp.WarpedVRTReaderBase.read
ValueError: WarpedVRT does not permit boundless reads
```
The fix is to always use `rasterio.merge.merge` instead of `src.read`, even if there is only a single file. This bug wasn't detected because we don't actually load real files and change the CRS during the RasterDataset tests. I added real files and ensured that we test loading those files with a different CRS.

- [x] Intersection/UnionDataset: fix crs/res propagation

This bug was reported by @jatentaki in #1341. Basically, the following works fine:
```python
ds = (ds1 & ds2) & ds3
ds = (ds1 | ds2) | ds3
```
However, the following is completely broken:
```python
ds = ds1 & (ds2 & ds3)
ds = ds1 | (ds2 | ds3)
```
The fix for this was much more involved. We need to add a res getter/setter to GeoDataset, similar to our crs getter/setter. Then we need to override this in Intersection/UnionDataset to update both datasets. This is updated recursively to ensure that all datasets have matching crs/res. This is a pretty uncommon use case (`ds1 & ds2 & ds3` maps to the former), so that's why no one noticed until now. I added tests for both of the above situations so this doesn't break again.

Thanks @jatentaki for reporting this!

Fixes #1341


torchgeo/datasets/geo.py
```diff
@@ -213,12 +211,10 @@ def bounds(self) -> BoundingBox:
 
     @property
     def crs(self) -> CRS:
-        """:term:`coordinate reference system (CRS)` for the dataset.
+        """:term:`coordinate reference system (CRS)` of the dataset.
 
         Returns:
-            the :term:`coordinate reference system (CRS)`
-
-        .. versionadded:: 0.2
```
adamjstewart, line 221: Although the getter/setter were added in 0.2, the attribute itself has always been present, so technically there was no API change as far as the user is concerned.

tests/data/raster/uint16/corrupted.tif
```diff
@@ -0,0 +1 @@
+not a tif file
```
calebrob6, line 1: needs a newline

(joking)

torchgeo/datasets/geo.py
```diff
@@ -477,22 +494,7 @@ def _merge_files(
...
-                indexes=band_indexes,
-                out_shape=out_shape,
-                window=from_bounds(*bounds, src.transform),
-                boundless=True,
-            )
-        else:
-            dest, _ = rasterio.merge.merge(
-                vrt_fhs, bounds, self.res, indexes=band_indexes
-            )
+        dest, _ = rasterio.merge.merge(vrt_fhs, bounds, self.res, indexes=band_indexes)
```
calebrob6, line 497: My only concern here is on performance but let's sort that out later. I'm imagining that we can have something like `benchmark.py` run during integration tests to test for dataloader performance regressions.

calebrob6 (approved):

torchgeo/datasets/geo.py
adamjstewart: I recall it being slightly slower (that's why we added the if-statement in the first place), but I'm unaware of any other way of doing this with src.read, so for now if we want boundless reading (needed for GridGeoSampler) then we have to use rasterio.merge.merge.

calebrob6: Sure, but it still stands that this is a place that needs revisiting

calebrob6 (approved):

Issue #409, opened 2022-02-16
GeoDataset: avoid unnecessary reprojection


adamjstewart:
This issue is meant to serve as a proposal for how to fix #278. Please try to poke holes in this proposal and find ways in which it won't work, or suggest alternative solutions.

### Rationale

Geospatial data files are stored in a particular CRS. In order to build a single R-tree index of a geospatial dataset, the bounds of all files must be reprojected to the same CRS. During sampling, we also reproject the file itself to a common CRS so that we can stitch together multiple files from the same layer, or combine multiple layers in an `IntersectionDataset`.

Unfortunately, reprojection/resampling is very slow, and can negatively affect data loader performance (see section 4.2 of https://arxiv.org/abs/2111.08872). It can also lead to large distortions of images in polar regions. We would like to be able to avoid reprojection if possible.

### Proposal

I propose to only reproject files during sampling when absolutely necessary. This includes any situation in which we sample from multiple files at once and all files are not in the same CRS, such as:

* a single `GeoDataset` containing overlapping files on the border of two UTM zones
* an `IntersectionDataset` with datasets in different CRSes
* a `UnionDataset` with overlapping files and datasets in different CRSes

The details of how this can be done is left to the "Implementation" section below, but the basic idea is to add an additional parameter for `__getitem__` that contains the CRS to project to. The sampler is then responsible for setting this CRS in a way that minimizes reprojection. 

There is one issue that arises with this implementation. If we choose our bounding box in one CRS and reproject it to a different CRS, the bounding box will be rotated, and the enclosing bounding box may be larger than the original. This can result in tensors of different size that cannot be concatenated without padding. The "Implementation" section describes how this problem can be avoided.

### Implementation

The details of what will need to be changed are as follows:

1. `GeoDataset.index`: instead of storing `{bbox: filename}`, store `{bbox: [(filename, CRS), ...]}`
2. `{Intersection|Union}Dataset.index`: instead of storing `{bbox}`, store `{bbox: [(filename, CRS), ...]}`
3. `GeoDataset.__getitem__`: instead of taking a `BoundingBox(minx, maxx, miny, maxy, mint, maxt)`, accept a `BoundingBox(x, y, t, width, height, time)` and CRS, then project (x, y) to that CRS before sampling
4. `GeoSampler`: look at the native CRS of all files in the R-tree and select the mode as the CRS to return

[...]

PR #1353, opened 2023-05-20
TileDatasets

calebrob6:
- Introduces a new class of datasets called TileDatasets that are indexed by filename, xoffset, yoffset, and patch size.
- Implements samplers for these
- Implements a L7IrishDataModule using this scheme

### DIFF
--- a/conf/l7irishtile.yaml
+++ b/conf/l7irishtile.yaml
@@ -1,0 +1,36 @@
+module:
+  _target_: torchgeo.trainers.SemanticSegmentationTask
+  loss: "ce"
+  model: "unet"
+  backbone: "resnet18"
+  weights: true
+  learning_rate: 1e-3
+  learning_rate_schedule_patience: 6
+  in_channels: 9
+  num_classes: 5
+  num_filters: 256
+  ignore_index: 0
+
+datamodule:
+  _target_: torchgeo.datamodules.L7IrishTileDataModule
+  root: "/home/calebrobinson/ssdprivate/torchgeo_dataset_integration_tests/data/L7Irish/"
+  batch_size: 32
+  patch_size: 256
+  train_batches_per_epoch: 500
+  val_batches_per_epoch: 100
+  num_workers: 6
+
+trainer:
+  _target_: lightning.pytorch.Trainer
+  accelerator: gpu
+  devices:
+    - 2
+  min_epochs: 15
+  max_epochs: 40
+
+program:
+  seed: 0
+  output_dir: output
+  log_dir: logs/l7irish/
+  overwrite: True
+  experiment_name: run1
--- a/torchgeo/datamodules/__init__.py
+++ b/torchgeo/datamodules/__init__.py
@@ -15,7 +15,7 @@ from .fire_risk import FireRiskDataModule
 from .geo import BaseDataModule, GeoDataModule, NonGeoDataModule
 from .gid15 import GID15DataModule
 from .inria import InriaAerialImageLabelingDataModule
-from .l7irish import L7IrishDataModule
+from .l7irish import L7IrishDataModule, L7IrishTileDataModule
 from .l8biome import L8BiomeDataModule
 from .landcoverai import LandCoverAIDataModule
 from .loveda import LoveDADataModule
@@ -41,6 +41,7 @@ __all__ = (
     # GeoDataset
     "ChesapeakeCVPRDataModule",
     "L7IrishDataModule",
+    "L7IrishTileDataModule",
     "L8BiomeDataModule",
     "NAIPChesapeakeDataModule",
     # NonGeoDataset
--- a/torchgeo/datamodules/l7irish.py
+++ b/torchgeo/datamodules/l7irish.py
@@ -86,7 +86,7 @@ class L7IrishTileDataModule(LightningDataModule):
 
         mask_mapping = {64: 1, 128: 2, 191: 3, 255: 4}
         if "mask" in sample:
-            mask = sample["mask"]
+            mask = sample["mask"].squeeze()
             for k, v in mask_mapping.items():
                 mask[mask == k] = v
             sample["mask"] = mask
@@ -106,12 +106,13 @@ class L7IrishTileDataModule(LightningDataModule):
                 mask_fns.append(os.path.join(root,area,path_row,f"L7_{path}_{row}_newmask2015.TIF"))
         return image_fns, mask_fns
 
-    def __init__(self, root, batch_size=1, patch_size=32, batches_per_epoch=None, num_workers=0, seed=0):
+    def __init__(self, root, batch_size=1, patch_size=32, train_batches_per_epoch=None, val_batches_per_epoch=None, num_workers=0, seed=0):
         super().__init__()
         self.image_fns, self.mask_fns = self._get_all_the_fns(root)
         self.batch_size = batch_size
         self.patch_size = patch_size
-        self.batches_per_epoch = batches_per_epoch
+        self.train_batches_per_epoch = train_batches_per_epoch
+        self.val_batches_per_epoch = val_batches_per_epoch
         self.num_workers = num_workers
 
         generator = torch.Generator().manual_seed(seed)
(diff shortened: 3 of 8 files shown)


adamjstewart:
Thanks for opening this proof of concept!

There's a few questions here:

1. Do we want to add this base class to TorchGeo?
2. Do we want all of our _curated benchmark_ GeoDatasets to subclass this?
3. Do we want to search for files or pass in a list of files?

My current opinions:

1. I think we'll likely want _something_ like this in TorchGeo
2. This one is more nuanced. Using this for all benchmark GeoDatasets lets us avoid reprojection, but prevents us from combining those datasets with other GeoDatasets. We'll have to decide how important this functionality is. Depending on how much time I have this summer, I may try to tackle #409, which would make the gains less.
3. I know you've been looking for something like this for a long time. I would like to be consistent here. Either Raster/Vector/Tile search for files, or get passed a list of files, or support both options.

calebrob6:
> Do we want to add this base class to TorchGeo?

Maybe -- although it feels like we should be able to get there by making RasterDataset and the Samplers more complex. I think I sketched out a method for allowing RasterDataset `__getitem__` to take in two different types of bounding boxes, the current one, and the one used in this implementation (call these `BoundingBox` and `Patch` or something -- note you can convert from `Patch` to `BoundingBox`). If we're in a situation where there are overlapping datasets or something you can just stick with the current implementation.

> Do we want all of our curated benchmark GeoDatasets to subclass this?

Depends on the first question

> Do we want to search for files or pass in a list of files?

We've talked about this a bunch of times before -- I almost never want to search a file system and keep things that match a regex. The nice thing about this is that you can rename `fns` to `uris` and pass it lists of COG files on remote servers and it will work, while you definitely can't do that with glob.

experiments/ssl4eo/run_l7irish.py
```diff
@@ -0,0 +1,81 @@
...
+    ):
+        if model == "fcn" and not weights:
+            continue
+
+        if model != "unet":
+            experiment_name = f"{model}_{backbone}_{lr}_{loss}_{wd}_{weights}_{seed}"
+        else:
+            experiment_name = f"{model}_{lr}_{loss}_{wd}_{weights}_{seed}"
+
+        config_file = os.path.join("conf", "l7irishtile.yaml")
```
nilsleh, line 56: Should we make one `run_{downstream_task}.py` script that has the config file name as an additional variable at the beginning of the file?

isaaccorley:
I'd like to revive this one but maybe we can also make the samplers work with non georeferenced images. Some of the datasets we have like GID-15, LEVIR-CD, etc. are large images that I may want to sample smaller patches from for training but I don't want to manually preprocess them to a specific patch size beforehand in case I want to run an ablation by varying the patch size.

calebrob6:
@adamjstewart -- do you have any more recent thoughts on this? I use these types of datasets somewhat frequently in my day-to-day work and would love to have them in TorchGeo, but also don't have a burning passion to merge it

adamjstewart:
Overall, I think I would be okay with this approach. It's kind of like what other libraries did before TorchGeo existed. I'm very okay with this for datasets like GID-15 and LEVIR-CD where the images are large but there are no geocoordinates. I'm slightly okay with this for datasets like L7 Irish where the images are large and there are geocoordinates. 

Fancy is cool, but fast is better. Especially if you're already regularly using datasets like this. I want to make sure TorchGeo is actually useful to its intended audience, including its maintainers. If the maintainers aren't using the builtin datasets, we're doing something wrong.

> Depending on how much time I have this summer, I may try to tackle GeoDataset: avoid unnecessary reprojection #409, which would make the gains less.

Well this did not age well...

>  it feels like we should be able to get there by making RasterDataset and the Samplers more complex.

I don't think this is possible. If we subclass from GeoDataset, we are declaring that the dataset can interoperate with all other GeoDatasets. TileDataset really does deserve its own base class and samplers. I'm not even sure if samplers are the right approach here. This requires a lot of rethinking. How do people normally use these kind of datasets for ML? Are there other ideas from high-res imagery we can borrow from? Presumably we are not the only ones using large images.

> Do we want to search for files or pass in a list of files?

We now support this, yay!

> Implements a L7IrishDataModule using this scheme

Note that I converted L7 Irish to an IntersectionDataset in 0.6.0. This significantly cuts down on the code (no more getitem). Just an aside, not really relevant to the discussion.

calebrob6:
Wanted to bring this back up -- I continue to run into cases where I can use a TileDataset. As one example, I rewrote `microsoft/building-damage-assessment` to use TileDataset to deal with the bug we currently have where xres and yres are treated as identical. If you want to see what replacing GeoDataset logic with TileDataset logic you can see this PR -- https://github.com/microsoft/building-damage-assessment/pull/7

If it helps anyone think about this -- a NonGeoDataset is essentially a very constrained TileDataset that doesn't have any flexibility in sampling. For NonGeoDatasets that have patches with different sizes we currently jump through hoops like NCrops in the datamodule when we _could_ just treat them as TileDatasets :)

adamjstewart:
That actually does help, to think of TileDataset as a strong NonGeoDataset instead of as a weak GeoDataset.

I think you're coming from the point of view that TileDataset is a _supplement_ to RasterDataset and lets you squeeze out more performance. Where I'm coming from is TileDataset is a _competitor_ to GeoDataset and NonGeoDataset and adds a [third competing standard](https://xkcd.com/927/). I'm worried that when people create new subclasses like Landsat8, they won't know if it should be a RasterDataset or a TileDataset. I don't want Landsat8RasterDataset and Landsat8TileDataset for all existing datasets. I know that's not your intention, but it does raise some questions of compatibility.

If it's okay with you, I would like to save this for the big TorchGeo 1.0 refactor. What I'm hoping to do is find more ways to unify all 3 "standards". I would like to track geospatial metadata whenever it exists, but not require its usage. This goes along with what I presented before about allowing GeoDataset to take either a bounding box or an integer index as input. This will allow you to combine a NonGeoDataset and a GeoDataset, or to use our spatial splitting methods on things like EuroSAT to construct something like EuroSATSpatial on the fly. I think it should be possible to incorporate some of the ideas from this PR without losing support for reprojection entirely. I'll continue to think it over though.

calebrob6:
I agree that it'd be cool to add the geospatial information wherever possible (and think it'd be especially cool to have EuroSAT be a test case -- e.g., how good can you do with 5 training samples per class and accessory geospatial information compared to the standard benchmark?)

> I think you're coming from the point of view that TileDataset is a supplement to RasterDataset
I think the opposite is true, I'm imagining that this replaces NonGeoDataset -- very specifically I'm proposing to refactor the NonGeoDataset base class to something similar to the TileDataset I have implemented here. TileDatasets are indexed by (file index, x coord, y coord, patch height, patch width). There should be a subclass called "RegularTileDataset" that is simply indexed by (file index, 0, 0, patch size, patch size) that _most_ of the current NonGeoDatasets can be refactored to use. ChesapeakeCVPR and any dataset that currently has non-uniform patch sizes fit nicely into this abstraction without having to resort to warping or weird datamodule level augmentations. If we eventually figure out some nicer/faster way to sample from GeoDatasets without going through WarpedVRT then that might cover 90% of the cases, however I also think we'll run into benchmark datasets with different patch sizes and no geospatial metadata that we'll want to support.

adamjstewart:
Gotcha. As long as int sampling also works for the majority of datasets, this seems fine to me. But let's still save this for the 1.0 refactor. What I'm imagining is something like:
```python
# Geographic coords, float
GeoSlice = slice | tuple[slice] | tuple[slice, slice] | tuple[slice, slice, slice]
class GeoDataset(Dataset):
    def __getitem__(self, index: int | GeoSlice) -> dict[str, Tensor]:
        ...

# Pixel coords, int
NonGeoSlice = tuple[int, slice] | tuple[int, slice, slice]
class NonGeoDataset(Dataset):
    def __getitem__(self, index: int | NonGeoSlice) -> dict[str, Tensor]:
        ...
```
So both classes will be compatible with PyTorch's builtin sampler (in cases of pre-chipped datasets) and we may need to make additional `NonGeoSamplers`. The API of both is getting closer and closer, so it may be possible to create a single shared base class if we're clever.

isaaccorley:
This also covers the case where several datasets are derived from Google Earth images as jpg without any georeferencing but still are too large to be fed to a model directly without patching.

calebrob6:
@adamjstewart @isaaccorley, I want to bring this back up (but without a full NonGeoDataset refactor). This would also solve the cases we've seen for "NCrops" (i.e. benchmark datasets with irregular sized inputs)

calebrob6:
Closing in favor of #3237

PR #1354, opened 2023-05-21
SSL4EO-L MoCo experiment

isaaccorley:
WIP

### DIFF
--- a/conf/ssl4eo-l-mocov2.yaml
+++ b/conf/ssl4eo-l-mocov2.yaml
@@ -1,0 +1,38 @@
+module:
+  _target_: torchgeo.trainers.MoCoTask
+  model: resnet50
+  weights: True
+  in_channels: 9
+  version: 2
+  layers: 2
+  hidden_dim: 2048
+  output_dim: 128
+  lr: 0.03
+  weight_decay: 1e-4
+  momentum: 0.9
+  schedule: [120, 160]
+  temperature: 0.07
+  memory_bank_size: 65536
+  moco_momentum: 0.999
+  gather_distributed: True
+  size: 224
+  grayscale_weights: null
+
+datamodule:
+  _target_: torchgeo.datamodules.SSL4EOLDataModule
+  root: /workspace/storage/data/ssl4eo/imgs
+  split: etm_toa
+  seasons: 1
+  batch_size: 256
+  num_workers: 128
+
+trainer:
+  _target_: lightning.pytorch.Trainer
+  accelerator: gpu
+  devices: 2
+  limit_val_batches: 0.0
+  max_epochs: 200
+  log_every_n_steps: 5
+
+program:
+  overwrite: True
--- a/torchgeo/transforms/color.py
+++ b/torchgeo/transforms/color.py
@@ -30,7 +30,7 @@ class RandomGrayscale(IntensityAugmentationBase2D):
 
     def __init__(
         self,
-        weights: Tensor,
+        weights: Optional[Tensor],
         p: float = 0.1,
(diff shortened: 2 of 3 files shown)


torchgeo/transforms/color.py
```diff
@@ -30,7 +30,7 @@ class RandomGrayscale(IntensityAugmentationBase2D):
 
     def __init__(
         self,
-        weights: Tensor,
+        weights: Optional[Tensor],
```
isaaccorley, line 33: As a user, I want to use the averaged grayscale by default. I also don't want to have to know how many channels to compute the average. This is different if I actually want to use weights.

conf/ssl4eo-l-mocov2.yaml
```diff
@@ -0,0 +1,38 @@
...
+  split: etm_toa
+  seasons: 1
+  batch_size: 256
+  num_workers: 128
+
+trainer:
+  _target_: lightning.pytorch.Trainer
+  accelerator: gpu
+  devices: 2
+  limit_val_batches: 0.0
```
isaaccorley, line 33: I'm not sure why but even w/o a defined val_step, the validation still gets run??? Force turning this off using `limit_val_batches: 0.0`

torchgeo/transforms/color.py
adamjstewart: This is a lot of extra code just for that...

isaaccorley: It's only a few lines so I don't have to define the weights. I initially made it a separate class. Maybe that's cleaner.

adamjstewart: Nah, separate class is all code duplication

adamjstewart: You'll prob want to make this a separate PR to save time. It affects both MoCo and SimCLR

isaaccorley: Sounds good. I actually thought of a better way to handle this so I'll push an update shortly.

isaaccorley: I actually didn't see you already it working if `grayscale_weights=None`, in that case, let's just forget this entirely

torchgeo/transforms/color.py
```diff
@@ -70,7 +70,7 @@ def apply_transform(
         Returns:
             The augmented input.
         """
-        weights = flags["weights"][..., :, None, None]
+        weights = flags["weights"][..., :, None, None].to(input.device)
```
isaaccorley, line 73: This fixes an issue with the weights being on cpu while the input is on gpu

conf/ssl4eo-l-mocov2.yaml
```diff
@@ -0,0 +1,38 @@
...
+  lr: 0.03
+  weight_decay: 1e-4
+  momentum: 0.9
+  schedule: [120, 160]
+  temperature: 0.07
+  memory_bank_size: 65536
+  moco_momentum: 0.999
+  gather_distributed: True
+  size: 224
+  grayscale_weights: null
```
adamjstewart, line 19: You shouldn't need to override any of these if you just want to use the defaults. They aren't required anymore like they were with the other trainers. I'll update the other trainers before 0.5.

conf/ssl4eo-l-mocov2.yaml
```diff
@@ -0,0 +1,38 @@
+module:
+  _target_: torchgeo.trainers.MoCoTask
+  model: resnet50
+  weights: True
+  in_channels: 9
+  version: 2
```
adamjstewart, line 6: Why version 2?

PR #1357, opened 2023-05-22
MoCo/SimCLR transforms update

isaaccorley:
- add random brightness and random contrast
- fix random grayscale wrong device bug

### DIFF
--- a/torchgeo/trainers/moco.py
+++ b/torchgeo/trainers/moco.py
@@ -62,6 +62,8 @@ def moco_augmentations(
             T.RandomGrayscale(weights=weights, p=0.2),
             # Not appropriate for multispectral imagery, seasonal contrast used instead
             # K.ColorJitter(brightness=0.4, contrast=0.4, saturation=0.4, hue=0.4, p=1)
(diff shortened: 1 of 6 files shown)


torchgeo/trainers/moco.py
```diff
@@ -62,6 +62,8 @@ def moco_augmentations(
             T.RandomGrayscale(weights=weights, p=0.2),
             # Not appropriate for multispectral imagery, seasonal contrast used instead
             # K.ColorJitter(brightness=0.4, contrast=0.4, saturation=0.4, hue=0.4, p=1)
+            K.RandomBrightness(brightness=(0.8, 1.0), p=0.8),
+            K.RandomContrast(contrast=(0.8, 1.0), p=0.8),
```
adamjstewart, line 66: I still need to look into the details of the implementations, but shouldn't this be `(0.6, 1.4)` and `p=1`?

adamjstewart:
Looks like RandomBrightness was added in Kornia 0.6.9. You'll need to update the minimum version in setup.cfg/environment.yml/min-reqs.old

adamjstewart (approved):
Thanks for fixing/improving this!

PR #1358, opened 2023-05-23
Satlas pretrained Swin Transformer v2 models

isaaccorley:
Adds Swin Transformer v2 Base pretrained models from the [Satlas](https://github.com/allenai/satlas) dataset


adamjstewart:
Will review after deadline

calebrob6:
is the `timm` swin transformer compatible?

isaaccorley:
No it's not. The layers are named differently unfortunately.

isaaccorley:
Wonder if I can map them to each other.

adamjstewart:
If it's not to hard, that would be useful since it would let us use it in our trainers. Wonder if we could add some kind of logic to support both timm and torchvision in our trainers...

isaaccorley:
It is hard unfortunately.

calebrob6 (approved):

isaaccorley:
I have no idea why the figure plotting in `torchgeo/trainers/detection.py` is somehow indirectly not covered via codecov

requirements/min-reqs.old
adamjstewart, line 1: Also need to update `pyproject.toml`

torchgeo/models/swin.py
```diff
@@ -0,0 +1,93 @@
...
+
+"""Pre-trained Swin v2 Transformer models."""
+
+from typing import Any, Optional
+
+import kornia.augmentation as K
+import torch
+import torchvision
+from torchvision.models._api import Weights, WeightsEnum
+from torchvision.models.swin_transformer import SwinTransformer
```
adamjstewart, line 12:
```suggestion
from torchvision.models import SwinTransformer
```
The sub-sub-modules aren't really public.

torchgeo/models/swin.py
```diff
@@ -0,0 +1,93 @@
...
+
+from ..transforms import AugmentationSequential
+
+__all__ = ["Swin_V2_B_Weights"]
+
+
+# https://github.com/allenai/satlas/blob/bcaa968da5395f675d067613e02613a344e81415/satlas/cmd/model/train.py#L42 # noqa: E501
+# All Satlas imagery is uint8 and normalized to the range (0, 1) by dividing by 255 # noqa: E501
```
adamjstewart, line 20: This doesn't need noqa, you can just wrap the line

adamjstewart (changes requested):

adamjstewart:
Turns out the decrease in test coverage is actually very important!

If you remove the try-except from ObjectDetectionTask, you'll find that there is a bug in our implementation:
```
FAILED tests/trainers/test_detection.py::TestObjectDetectionTask::test_trainer[True-faster-rcnn-nasa_marine_debris] - ValueError: Boxes need to be in (xmin, ymin, xmax, ymax) format. Use torchvision.ops.box_convert to convert them
FAILED tests/trainers/test_detection.py::TestObjectDetectionTask::test_trainer[True-fcos-nasa_marine_debris] - ValueError: Boxes need to be in (xmin, ymin, xmax, ymax) format. Use torchvision.ops.box_convert to convert them
FAILED tests/trainers/test_detection.py::TestObjectDetectionTask::test_trainer[True-retinanet-nasa_marine_debris] - ValueError: Boxes need to be in (xmin, ymin, xmax, ymax) format. Use torchvision.ops.box_convert to convert them
FAILED tests/trainers/test_detection.py::TestObjectDetectionTask::test_no_rgb[True] - ValueError
```
These were being masked by our try-except. The code works fine for torchvision 0.13 but breaks for any newer version. We didn't notice this because the combined coverage didn't change. From what I can tell, this has always been broken...

adamjstewart:
Can you try to fix that in a separate PR so we can backport it to 0.5.2? Or one of us can do it, just let me know if you're busy.

isaaccorley:
> Can you try to fix that in a separate PR so we can backport it to 0.5.2? Or one of us can do it, just let me know if you're busy.

I'll push a PR here shortly

adamjstewart (approved):
Can merge as soon as coverage issue is fixed.

docs/api/misc_pretrained_weights.csv
```diff
@@ -1,2 +1,4 @@
 Weight,Channels,Source,Citation
 ResNet50_Weights.FMOW_RGB_GASSL, 3,`link <https://github.com/sustainlab-group/geography-aware-ssl>`__,`link <https://arxiv.org/abs/2011.09980>`__
+Swin_V2_B_Weights.NAIP_RGB_SATLAS,3,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__
+Swin_V2_B_Weights.SENTINEL2_RGB_SATLAS,3,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__
```
adamjstewart, line 4: If we know what satellite they were trained on, these should have their own categories/files. So Sen 2 should go under Sen 2, and we should have a new category/file for NAIP.

adamjstewart (approved):

PR #1359, opened 2023-05-23
SSL4EO-L MoCo-v3 Experiment

isaaccorley:


conf/ssl4eo-l.yaml
```diff
@@ -0,0 +1,36 @@
+module:
+  _target_: torchgeo.trainers.MoCoTask
+  model: resnet50
+  weights: True
```
adamjstewart, line 4: Do we want to start from ImageNet weights?

conf/ssl4eo-l.yaml
```diff
@@ -0,0 +1,36 @@
+module:
+  _target_: torchgeo.trainers.MoCoTask
+  model: resnet50
```
adamjstewart, line 3: Ideally we'll want pre-trained models for ResNet-18, ResNet-50, and ViT-S. Would probably prioritize the last 2. If we don't have any ViTs I think the reviewers will laugh at us.

conf/ssl4eo-l.yaml
```diff
@@ -0,0 +1,36 @@
...
+  gather_distributed: True
+  size: 224
+  grayscale_weights: null
+
+datamodule:
+  _target_: torchgeo.datamodules.SSL4EOLDataModule
+  root: /workspace/storage/data/ssl5eo_l_etm_toa
+  split: etm_toa
+  seasons: 2
+  batch_size: 256
```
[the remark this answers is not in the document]
isaaccorley: I haven't read the v3 paper in awhile, but I'm curious if the memory bank (MoCo-v2 settings) is better in the case you can't push the batch size to 4096, the memory bank will still allow you to sample many negatives

conf/ssl4eo-l.yaml
isaaccorley: I think we should focus on ResNet50 and ViT-S first. The only other SSL paper I've seen recently that trains ResNet18 is SeCo.

conf/ssl4eo-l.yaml
isaaccorley: `True` should use imagenet weights. See [link](https://github.com/microsoft/torchgeo/blob/3196dab4dbb9df15d47f266e0fa2c7ec6eb415a3/torchgeo/trainers/moco.py#L227)

conf/ssl4eo-l.yaml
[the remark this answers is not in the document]
adamjstewart: Yeah, not sure either. Let's stick to vanilla v3 for now. We can add memory bank later if necessary.

adamjstewart:
Do we still need this PR?

adamjstewart:
Ping @isaaccorley

adamjstewart:
None of these files even exist anymore, I think we can close this.

PR #1362, opened 2023-05-24
Adjust naming scheme for SSL4EO dataset

nilsleh:
This PR makes two changes:
- tm_sr -> etm_sr
- correct L5 naming in data.py

I suppose we would also have to change the name of the actual dataset to not break anything.


adamjstewart (approved):

adamjstewart:
> I suppose we would also have to change the name of the actual dataset to not break anything.

Can you clarify this?

nilsleh:
> Can you clarify this?

I was thinking that somewhere inside the real dataset filenames or directories etc there would be a `tm_sr` so if the dataset class is now looking for `etm_sr` there will be an error?

adamjstewart:
Nah, the data I uploaded to Azure and will eventually upload to HF all uses etm_sr.

PR #1364, opened 2023-05-24
CDL ordinal label map

nilsleh:
This PR adds an ordinal label map so there are no label errors during training. The raw labels go from 0-255 but there are "only" 133 labels.


torchgeo/datasets/cdl.py
```diff
@@ -28,6 +28,9 @@ class CDL(RasterDataset):
     annually using moderate resolution satellite imagery and extensive agricultural
     ground truth.
 
+    The dataset contains 133 classes, for a description of the classes see the
+    xls file at the top of `this page <https://www.nass.usda.gov/Research_and_Science/Cropland/sarsfaqs2.php>`_.
```
adamjstewart, line 32: Can you move URL to another line so it isn't as long?

adamjstewart: 134 classes if you count the background class

adamjstewart (approved):
Now you can import and use this in your other dataset. The NLCD dataset already has this.

PR #1368, opened 2023-05-25
Adapt CDL cmap to match ordinal label mapping

nilsleh:
new cmap with ordinal labels:
```
new_cmap = {}
for key, val in cmap.items():
    if key in ordinal_label_map:
        new_cmap[ordinal_label_map[key]] = val
```
Also saw that `return fig` was missing from plot.


adamjstewart:
Can you upload an example plot?

adamjstewart:
Hmm, the plot looks a little odd. Can you use the geolocation from the sample image and view the raw CDL file in QGIS at the same location? Especially here in the Midwest, almost everything should be corn (yellow) or soybean (green), with a small amount of urban (gray) and more grassland (light green) as you move further west.

nilsleh:
I haven't used QGIS, so I tried and the image falls somewhere here. don't know how to exactly align the above image within QGIS: 
Screenshot from 2023-05-26 10-19-30

adamjstewart:
Yeah, that second plot is what it should look like. What image and mask year is it? I'll zoom in on QGIS until I get the exact plot it should look like.

adamjstewart:
If it's 2019, the plot should look something like this:
Screenshot 2023-05-26 at 9 57 47 AM

nilsleh:
I had 2011

calebrob6:
Sup fellas, I think the `ax.imshow()`s need a vmin=0, vmax=255, and interpolate="none".

Also, would gently recommend not spending time on this :)

adamjstewart:
2011 version is a bit more sporadic, closer to what you're seeing:
Screenshot 2023-05-26 at 11 29 05 AM
Looks like corn is being mapped from yellow to white and soybean is being mapped from green to yellow.

adamjstewart:
> Sup fellas, I think the ax.imshow()s need a vmin=0, vmax=255, and interpolate="none".

You are correct, although vmax should be len(cmap) - 1, not 255.

> Also, would gently recommend not spending time on this :)

I needed a break :)

adamjstewart:
I plotted SSL4EOLBenchmark, NLCD, and CDL and compared them against the QGIS visualizations and everything looks correct.

ssl4eo_nlcd
nlcd
ssl4eo_cdl
cdl

adamjstewart (approved):
@nilsleh see what you think of this solution

adamjstewart:
@calebrob6 does Chesapeake not map mask values to ordinal ranges? Should it? Trying to decide whether or not we need cmap to be a dict or if it can be a much simpler list.

adamjstewart:
Decided to keep it as a dict to match `rasterio.DatasetReader.colormap()`

adamjstewart (dismissed):

adamjstewart:
I can split the changes to each file into separate PRs if you want. Might make it easier to backport only CDL (the rest will have to wait until 0.5).

adamjstewart (approved):

PR #1371, opened 2023-05-26
L8 Biome: clarify dataset description

adamjstewart:
The dataset isn't for validating clouds, it's for validating the algorithms used to detect them.

@shradhasehgal


shradhasehgal (approved):

shradhasehgal:
> The dataset isn't for validating clouds, it's for validating the algorithms used to detect them.
> 
> @shradhasehgal

Yup! Nice catch, thanks for updating it.

PR #1376, opened 2023-05-28
Revert "GASSL ResNet50 Weights"

adamjstewart:
Reverts microsoft/torchgeo#1325

PRs have been failing ever since this PR was merged. Curious to see if the errors go away if we remove it. Failure is always on Windows, and always looks like:
```
RuntimeError: [enforce fail at ..\c10\core\impl\alloc_cpu.cpp:72] data. DefaultCPUAllocator: not enough memory: you tried to allocate 4194304 bytes.
```
The failing tests are always when our trainers try to load our mocked model weights. Not sure what we can do other than load fewer weights. We might be able to only load 1 ResNet, but we'll still have to load all of them on the release branch, so we can't just avoid the problem. Maybe we need to clear memory after each test somehow? We can't really fake the model to something smaller since that won't work on the release branch. I wonder if we could get Microsoft to pay for larger GitHub Actions runners... (@calebrob6?)

This PR is half test and half serious. Like I don't want to remove these weights, but we also can't let our tests keep failing forever. We could revert until we have time to figure out how to get this working properly.


adamjstewart:
Closing in favor of #1377

PR #1378, opened 2023-05-29
Improve customizability of Landsat band plots

adamjstewart:
Follow-up to #1218

### Changes

- [x] Add RBV, ETM, and OLI-2/TIRS-2 sensors
- [x] Don't hardcode y locations or labels (got sick of adjusting them every time)
- [x] Allow customization of list of sensors to be plotted
- [x] Use same typeface and font size as NeurIPS template
- [x] Convert subxlabel to xlabel to fix font size and spacing

### Main text plot (the sensors we care about)

```console
$ python3 plot_landsat_bands.py RBV MSS ETM --fig-height=3
```
sensors

### Supplementary materials plot (all sensors)

```console
$ python3 plot_landsat_bands.py
```
all_sensors


adamjstewart:
Haven't yet decided whether I want to combine all MSS, and OLI/TIRS 1 and 2, or whether to keep them separate. The wavelengths/res are the same, the only difference is band # or sensor name.

adamjstewart:
I'm leaning towards the latter. We can always clarify in the caption.

adamjstewart:
Yep, that looks better:

all_sensors

PR #1383, opened 2023-05-30
Add script to plot Landsat mission timeline

adamjstewart:
The goal is to get something similar to [this](https://www.usgs.gov/media/images/landsat-missions-timeline):

timeline

Here is what I have so far:

timeline

Crosshatched regions represent partial or complete sensor failure.


adamjstewart:
Should we artificially make the Landsat 6 bar wider so it isn't just a single line? 

The original plot doesn't include dates after satellites went offline or went on standby mode. My bars go all the way til the satellite was decommissioned. This is most obvious for Landsat 4. Thoughts on which we should end with?

PR #1384, opened 2023-05-31
Script for downstream tasks

nilsleh:


experiments/ssl4eo/conf/tm_toa_cdl.yaml
```diff
@@ -0,0 +1,34 @@
...
+  accelerator: cpu
+  min_epochs: 50
+  max_epochs: 100
+
+program:
+  seed: 0
+  output_dir: output/tm_toa_cdl/
+  log_dir: logs/tm_toa_cdl/
+  overwrite: True
+  experiment_name: test_downstream_tm_toa
```
adamjstewart, line 34: What text editor do you use? There should be a way to prevent this line ending bug

experiments/ssl4eo/run_ssl4eo_benchmark.py
```diff
@@ -0,0 +1,79 @@
...
+# Licensed under the MIT License.
+
+"""Runs the train script with a grid of hyperparameters."""
+import itertools
+import os
+import subprocess
+from multiprocessing import Process, Queue
+
+# list of GPU IDs that we want to use, one job will be started for every ID in the list
+GPUS = [1, 2, 3, 4]
```
adamjstewart, line 12: Can we rename gpu to device everywhere? For example, on macOS, I would use MPS, not GPU

calebrob6: Can you do this in a big rename PR?

experiments/ssl4eo/conf/tm_toa_cdl.yaml
adamjstewart, line 1: This file is the same as the `ssl4eo_benchmark_tm_toa_cdl.yaml` file I added later, right? Maybe we can combine them. I'm also open to reducing the number of files I added, I think I may have been too gung ho.

calebrob6:
@adamjstewart @nilsleh does this need to be merged?

adamjstewart:
I don't think this is needed

PR #1387, opened 2023-06-01
GitHub Actions: disable macOS Python 3.11 cache

adamjstewart:
https://github.com/actions/cache/issues/1176


adamjstewart:
Confirmed that only macOS Python 3.11 caching is skipped. This should be ready to merge.

isaaccorley (approved):

PR #1390, opened 2023-06-02
Add copyright to files we forgot

adamjstewart:


experiments/ssl4eo/compute_dataset_pca.py
```diff
@@ -12,7 +15,6 @@
     parser = argparse.ArgumentParser()
     parser.add_argument("directory", help="directory to recursively search for files")
     parser.add_argument("--ext", default="tif", help="file extension")
-    parser.add_argument("--nan", default=0, type=float, help="fill value")
```
adamjstewart, line 15: No longer used

PR #1399, opened 2023-06-05
Allow GeoDataset to list files in VSI path(s)

### LINKED ISSUE #1165: Reading S3File using RasterDataset
### Summary

Hello,
First of all thanks for your work.

Could it be possible to add an option from reading files from S3 using RasterDataset? 

```
class BDOrtho(RasterDataset):
    filename_glob = "*.jp2"
    is_image = True
    separate_files = True
```
    
`TypeError: expected str, bytes or os.PathLike object, not S3File`

Kind regards,
Alexandra

### Rationale

I need to create my own datasets reader

### Implementation

_No response_

### Alternatives

_No response_

### Additional information

_No response_

### LINKED ISSUE #1398: Support reading from GDAL virtual file systems (e.g. cloud storage)
https://github.com/microsoft/torchgeo/blob/9e57f278188ca36348ce8d5c30d5ae2acb19107c/torchgeo/datasets/geo.py#L363-L367

[GDAL virtual file systems](https://gdal.org/user/virtual_file_systems.html) such as reading directly from Google Buckets are natively supported by rasterio (through gdal).

```python
with rasterio.open("vsigs/my_bucket/.../my_image.tif") as src:
    src.read()  # etc.
```

The glob-matching (source code linked above) is the only thing stopping this currently.

What do you think the best way is to do this? My initial guess is that supporting the glob-matching for all the different file systems would take some effort.

The quickest fix (for me at least) would be to add an optional parameter `filenames:List` that is iterated, and the (already existing) try/except would handle if the filename is wrong.

adriantre:
Fix #1398 
Fix #1165

This PR remove blockers for reading raster files directly from [Cloud Storage (buckets)](https://gdal.org/user/virtual_file_systems.html) and other GDAL Virtual File Systems.


adamjstewart:
Hey, sorry it took us so long to review this! We're normally much more responsive to issues/PRs but we just finished a paper deadline.

This looks great! We've been wanting better support for cloud (AWS/GCP/Azure) for a long time. The biggest thing holding that up is my own perfectionism (what about VectorDataset or NonGeoDataset?). There isn't an easy way to add support for cloud for every dataset without refactoring the entire library. I was considering doing this at the same time as porting to TorchData, but we're still on the fence about when or if that will actually happen. So in my opinion, starting with Raster/VectorDataset for now and worrying about the others should be fine.

In terms of the current implementation, here are my preliminary thoughts:

1. Is it possible to use the same logic for VectorDataset?
2. Can VSI vs. non-VSI be detected based on whether or not `root` starts with `/vsi`? Or would it start with https://? We can check for `/vsizip/`, `/vsitar/`, etc. if we need to be more explicit. I can't imagine too many users will happen to have a `/vsi*` directory on their root filesystem, that's the only issue I can think of.
3. The listdir stuff seems fragile, especially the error handling. What happens if the URL/path is not actually a real VSI filesystem? Does it return an invalid root directory? Is this only a fiona feature or does rasterio have the same feature?
4. Does this work for all cloud providers? Some of us may have access to others if you need help checking this.
5. Testing this will be fun. For the normal tests that run on all PRs, we try to avoid any tests that require internet access. So if we can create a local VSI file to test this that would be ideal. We could potentially add tests that only get run during the release process if we want to check remote cloud providers, but idk how stable those would be.

@calebrob6 will likely want to check this to make sure this works for him on Azure/Planetary Computer.

P.S. Thanks for the PR! This would be a fantastic first contribution.

adriantre:
> There isn't an easy way to add support for cloud for every dataset without refactoring the entire library.

As Sean mentioned in my [Fiona feature request](https://github.com/Toblerity/Fiona/issues/1275) maybe [libcloud](https://libcloud.apache.org/) is a good option. But we could start with the simple `fiona.listdir` and if users report incompatibilities we can put more efforts into the "perfect" solution?

> 1. Is it possible to use the same logic for VectorDataset?
Looks like the behaviour will be the same. Everything that can be opened using fiona/rasterio/gdal will support vsi.

> 2. Can VSI vs. non-VSI be detected based on whether or not `root` starts with `/vsi`? Or would it start with https://? We can check for `/vsizip/`, `/vsitar/`, etc. if we need to be more explicit. I can't imagine too many users will happen to have a `/vsi*` directory on their root filesystem, that's the only issue I can think of.

`fiona.listdir` will work for local filesystems as well. Root will start with `/vsi`. If it is a problem that users for some reason create a local folder `/vsi` then TorchGeo datasets already have this problem as they read files using fiona and rasterio. We can use [is_remote or valid_vsi](https://github.com/Toblerity/Fiona/blob/b23fd3b94741599d4dec4f7d85238a54d362bd97/fiona/vfs.py#L32) from fiona instead of the vsi kwarg in my suggestion.

> 3. The listdir stuff seems fragile, especially the error handling. What happens if the URL/path is not actually a real VSI filesystem? Does it return an invalid root directory? Is this only a fiona feature or does rasterio have the same feature?

Looks like rasterio does not have this method. Do you think the above mentioned `valid_vsi` will be sufficient regarding teh fragility?

> 4. Does this work for all cloud providers? Some of us may have access to others if you need help checking this.

This (should) work with the cloud providers listen in the [gdal docs](https://gdal.org/user/virtual_file_systems.html#network-based-file-systems). [fiona.listdir use](https://github.com/Toblerity/Fiona/blob/b23fd3b94741599d4dec4f7d85238a54d362bd97/fiona/ogrext.pyx#L1918) method [VSIReadDir](https://gdal.org/doxygen/cpl__vsi_8h.html#abbdfabc91fccb19c331e5b3ea520743d) from gdal/ogr internally, which I assume is thoroughly tested. I have tested my implementation on Azure and GS.

> 5. Testing this will be fun. For the normal tests that run on all PRs, we try to avoid any tests that require internet access. 

Looks like [moto](https://github.com/rasterio/rasterio/issues/1785) can mock virtual file systems.

adamjstewart:
We're trying to limit additional dependencies but I'll keep libcloud and moto in mind.

I agree that we should start with a simple implementation and worry about making it "perfect" later. Especially since it wouldn't be an API change if we modify the implementation later.

`valid_vsi` looks good to me, let's use that and remove the user-defined parameter.

isaaccorley:
We had previously discussed possibly making a `from_s3` or `from_files` method instead so a user could instantiate a dataset like `dataset = RasterDataset.from_files(...)`. This would also reduce code in the constructor. Do we want to take this approach instead?

adamjstewart:
We should definitely have a `RasterDataset.from_files` classmethod constructor, @calebrob6 has been asking for that for years.

This is somewhat orthogonal to this PR. We should add `from_files` in a separate PR. Then we can decide whether we want `root` to support both local and remote file systems or just tell users to use `from_files`.

@adriantre would you be interested in implementing `from_files`? It would share most of the code with `__init__` (we can move it to a subfunction) but accept a list of files instead of a single directory. I'll open an issue so we can track this, it should be relatively easy. Also let me know what you think about requiring `from_files` for remote file systems or not.

adriantre:
I will give it a shot. I'm initially thinking that, analogously to `rasterio.open`, we shouldn't need to have different constructors for remote or local files. We may look to e.g. `pandas.read_file` to see how they handle imports and drivers for different file types.

adriantre:
The vsi can be passed in two ways. E.g. for azure:
1. `/vsiaz/my/root` is [OGR VSI](https://gdal.org/user/virtual_file_systems.html)
2. `az://my/root` is [Apache` Common VFS](https://commons.apache.org/proper/commons-vfs/filesystems.html#Zip_Jar_and_Tar)

Looks like [2 is preferred](https://rasterio.readthedocs.io/en/stable/topics/datasets.html), at least somewhat documented, although internally 2. is converted to 1. before it is passed to the reader.

For either we would need to use `fiona.listdir` (or alternatives) to list the contents.

I realised that we cannot use `valid_vsi` from fiona. It only supports 2. and is not up-to-date with all supported vsi.

Lastly, another vsi complicating the matter is zip-archives, which `fiona.listdir` supports to a lesser degree. It works only if the root is the actual zip-archive, so my implementation `listdir_vsi_recursive` won't be sufficient if one wants all tif's in all zip-archives within the root.

I'm starting to feel like `RasterDataset.from_files` is the most flexible way of solving this. But at the same time it feels like it should be _possible_ to make it work for root-directory discovery as well, but we would have to narrow the scope somehow.

Thoughts?

adamjstewart:
It's certainly possible, but as you mention, the scope makes it difficult. I think `from_files` should be added regardless. If someone asks for native support in `__init__`, we can consider adding it, but let's see if `from_files` is sufficient for most people for now. We'll also want to document it so people know it actually exists. Maybe add a new tutorial. Not required for the initial PR unless you're really gung ho about it.

calebrob6:
I think `.from_uris` is more what I was thinking about, then the user could pass vsi strings, https links to COGs, local paths, etc.

adriantre:
I have now refactored this and rebased on #1442, so all changes from that PR is reflected here until it is merged.

torchgeo/datasets/geo.py
```diff
@@ -358,11 +365,22 @@ def __init__(
...
+            root = [root]
+
+        filespaths: list[str] = []
+        for dir_or_file in root:
+            if os.path.exists(dir_or_file) and os.path.isfile(dir_or_file):
+                filespaths.append(dir_or_file)
+            else:
+                filespaths.extend(
+                    list_directory_recursive(dir_or_file, self.filename_glob)
+                )
```
adriantre, line 378: This is the new change in this file

adriantre:
Is now outdated wrt. to #1442 but still works as proof of concept. Any feedback?

torchgeo/datasets/utils.py
```diff
@@ -737,3 +742,48 @@ def percentile_normalization(
         (img - lower_percentile) / (upper_percentile - lower_percentile + 1e-5), 0, 1
     )
     return img_normalized
+
+
+def _path_is_vsi(path: str) -> bool:
+    from rasterio._path import SCHEMES
+
+    prefix = path.split("://")[0]
+    schemes = prefix.split("+")
```
adriantre, line 751: This, because vsi can be chained, e.g. "zip+az://..." for zip-files in azure.

adamjstewart:
> Any feedback?

I'm hesitant to add too much complexity if #1442 supports something similar. I would say let's keep pushing on #1442 and once that's done we can decide whether this is functionality we need or whether the user should be responsible for calling this function themselves and passing the input as a list of paths to the dataset.

adriantre:
All right, I have an idea. In #1442 I have an else statement in `list_files` that hits if the path is not a local path. There we can call a method that raises until the user overrides it with their own logic.

adamjstewart:
Is that logic needed for remote files or only for remote directories? I was under the impression that if a user implemented a method to find remote files, then GDAL could open those without issue.

adriantre:
Indeed, gdal can open them if they reach `rasterio.open`. But with my implementation of allowing list of files, the [code checks os.exists](https://github.com/microsoft/torchgeo/blob/e2fb866d078f0ced1e56979c5a26405323807970/torchgeo/datasets/geo.py#L441), which it does not. So we need to handle them separately, or not check if it exists. Thoughts?

adamjstewart:
We don't need to check `os.exists`. If it exists, we can check if it's a directory. If it's a directory, we recursively search. If it's a file or doesn't exist, we assume it's a path and proceed as normal.

adriantre:
That works :)

adamjstewart:
Ton of failing tests.

adriantre:
Fixed the most obvious failing tests.

Three tests fail on GH Action, but not locally on my machine. Any ideas? I have the same version of fiona locally as in required.txt (1.9.6).

```python
FAILED tests/datasets/test_geo.py::TestGeoDataset::test_files_property_for_virtual_files - AttributeError: module 'fiona' has no attribute 'listdir'
FAILED tests/datasets/test_geo.py::TestGeoDataset::test_files_property_ordered - AttributeError: module 'fiona' has no attribute 'listdir'
FAILED tests/datasets/test_geo.py::TestGeoDataset::test_files_property_deterministic - AttributeError: module 'fiona' has no attribute 'listdir'
```

adamjstewart:
The failing tests use our current minimum supported versions, listed in https://github.com/microsoft/torchgeo/blob/main/requirements/min-reqs.old. Currently this is fiona 1.8.21. We can increase this if we need to, as long as it isn't too new.

torchgeo/datasets/utils.py
```diff
@@ -772,6 +776,47 @@ def path_is_vsi(path: Path) -> bool:
...
 
+def listdir_vsi_recursive(root: Path) -> list[Path]:
+    dirs = [root]
+    files = []
+    while dirs:
+        dir = dirs.pop()
+        try:
+            subdirs = fiona.listdir(dir)
+            dirs.extend([os.path.join(dir, subdir) for subdir in subdirs])
+        except FionaValueError:
```
adriantre, line 787: @adamjstewart the test [test_files_property_for_virtual_files](https://github.com/microsoft/torchgeo/blob/8c02cec3801672947455db92f76537072b15dd4a/tests/datasets/test_geo.py#L181) checks vsi paths that (currently) does not exist. If I was to raise error upon 'does not exist', the test would fail.

Three options:
1. Allow non-existing vsi to pass through without raising errors
2. Remove the test, and rely on the new test I added that checks only zip
3. Add tests for all the vsi listed in the test.

For now I went with number 1.

adamjstewart:
> Thanks. Skimmed the release logs, and could not find any breaking api changes. Updated `min-reqs.old` to fiona version 1.9.6.
> 
> I'm not familiar with the intention of minimum supported versions, so I might have missed something.
> 
> It does seem like there are a lot of important bugfixes in [1.9.4](https://github.com/Toblerity/Fiona/releases/tag/1.9.4), but it introduces new GDAL version (?)

We want our users to be able to install TorchGeo as easily as possible with a wide range of supported dependency versions. `pyproject.toml` documents all of our dependencies and minimum versions. This is what `pip install` looks for when someone installs TorchGeo.

We also test this range of versions in CI to make sure it's accurate. We test both the latest version (`requirements/*.txt`) and minimum version (`requirements/*.old`) of all of our dependencies.

Fiona 1.9.6 is probably too new to set as a minimum supported version. It looks like `fiona.listdir` was added in fiona 1.9.0, can you try that version? If it works, you'll also need to update `pyproject.toml`.

torchgeo/datasets/geo.py
```diff
@@ -307,12 +306,11 @@ def files(self) -> list[Path]:
...
-            if os.path.isdir(path):
-                pathname = os.path.join(path, '**', self.filename_glob)
-                files |= set(glob.iglob(pathname, recursive=True))
-            elif os.path.isfile(path) or path_is_vsi(path):
+            if os.path.exists(path) and os.path.isfile(path):
                 files.add(path)
-            elif not hasattr(self, 'download'):
+            else:
+                files |= set(list_directory_recursive(path, self.filename_glob))
+            if not files and not hasattr(self, 'download'):
```
adamjstewart, line 313: `not files` is going to become False as soon as any files are found. If you pass in 2 directories, 1 containing files and 1 containing nothing, the warning will not issue for the second directory.

adamjstewart (changes requested):

torchgeo/datasets/utils.py
```diff
@@ -772,6 +776,47 @@ def path_is_vsi(path: Path) -> bool:
...
+        root: directory to list. For vsi these will have prefix
+            e.g. /vsiaz or az:// for azure blob storage
+        filename_glob: filename pattern to filter filenames
+    """
+    files: list[Path]
+    if path_is_vsi(root):
+        # Change type to match expected input to filter
+        files_as_str: list[str] = [str(file) for file in listdir_vsi_recursive(root)]
+        # Prefix glob with wildcard to ignore directories
+        files = cast(list[Path], fnmatch.filter(files_as_str, '*' + filename_glob))
```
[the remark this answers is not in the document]
adriantre: Tried to make the method(s) respect the `Path: TypeAlias = str | pathlib.Path` type all the way, but changed it now to always return list[str].

torchgeo/datasets/utils.py
adriantre: Agreed. Removed the test, added new tests for files within zip archive, and made GeoDataset.files warn the same way as for non-existing (regular) directores.

torchgeo/datasets/geo.py
adriantre: Thanks! I assigned the result of listing the current directory to a variable, and checked that, before warning/adding to set.

tests/datasets/test_geo.py
```diff
@@ -84,6 +85,19 @@ def __len__(self) -> int:
         return 2
 
 
+@pytest.fixture(scope='module')
+def temp_archive(request: SubRequest) -> Generator[tuple[str, str], None, None]:
```
adriantre, line 89: @adamjstewart This temp archive fixture will run only once per input parameter (directory). The input comes from the test that uses it, hence the `indirect=True` in `pytest.mark.parametrize`.

pyproject.toml
```diff
@@ -39,7 +39,8 @@ dependencies = [
     # einops 0.3+ required for einops.repeat
     "einops>=0.3",
     # fiona 1.8.21+ required for Python 3.10 wheels
-    "fiona>=1.8.21",
+    # fiona 1.9.0+ required for fiona.listdir
+    "fiona>=1.9.0",
```
adamjstewart, line 43:
```suggestion
    # fiona 1.9+ required for fiona.listdir
    "fiona>=1.9",
```
Just to make things simpler

tests/datasets/test_geo.py
adamjstewart: I've been looking for a way to do this for a long time, had no idea about `indirect=True`. Didn't expect to learn new pytest tricks while reviewing this PR, thanks!

torchgeo/datasets/utils.py
```diff
@@ -772,6 +776,74 @@ def path_is_vsi(path: Path) -> bool:
...
+    Args:
+        root: directory to list. These must contain the prefix for the VSI
+            e.g. '/vsiaz/' or 'az://' for azure blob storage
+                 '/vsizip/' or 'zip://' for zipped archives
+
+    Returns:
+        A list of all file paths matching filename_glob in the root VSI directory or its
+        subdirectories.
+
+    .. versionadded:: #TODO
```
adamjstewart, line 791: You can use 0.6 for now, that's the most likely candidate. If we need to we can change it later.

tests/datasets/test_geo.py
adriantre: I figured I would try to do this properly ;)

torchgeo/datasets/geo.py
```diff
@@ -307,12 +307,16 @@ def files(self) -> list[Path]:
...
+                    files.add(path)
+                    file_not_found = False
+            elif files_found := set(list_directory_recursive(path, self.filename_glob)):
+                files |= files_found
+                file_not_found = False
+
+            if file_not_found and not hasattr(self, 'download'):
                 warnings.warn(
                     f"Could not find any relevant files for provided path '{path}'. "
                     f'Path was ignored.',
```
adriantre, line 322: @adamjstewart I went ahead and prepared how this will look adopting changes from #2222. I think this is the best I can do in terms of clean code.

tests/datasets/test_geo.py
```diff
@@ -334,6 +343,132 @@ def test_no_all_bands(self) -> None:
...
+
+        files_zipped = Sentinel2(
+            paths=f'zip://{dir_zipped}', bands=bands, transforms=transforms, cache=cache
+        ).files
+
+        basenames_not_zipped = [Path(path).stem for path in files_not_zipped]
+        basenames_zipped = [Path(path).stem for path in files_zipped]
+        assert basenames_zipped == basenames_not_zipped
+
+    def test_http_files(self) -> None:
```
adriantre, line 444: @sfalkena I added this test, do you think it is sufficient? I found adding more files would make the test significantly slower, so I opted for two.

adamjstewart:
I have a couple paper deadlines in the next couple days, will get back to this afterwards.

adamjstewart:
Can you merge/rebase now that my PR has been merged?

torchgeo/datasets/utils.py
```diff
@@ -772,6 +776,74 @@ def path_is_vsi(path: Path) -> bool:
...
+        subdirectories.
+
+    .. versionadded:: #0.6
+    """
+    dirs = [str(root)]
+    files = []
+    while dirs:
+        dir = dirs.pop()
+        try:
+            subdirs = fiona.listdir(dir)
```
adamjstewart, line 798: Does this work for both raster and vector files?

adriantre: `fiona.listdir` lists any file or directory actually, not only geospatial files.

torchgeo/datasets/geo.py
adriantre: Merged from main with #2222 changes

adriantre:
So to the test failing on Windows. I don't have access to a Windows to debug, but:

Seems like there is a bug with [fiona 1.10a1](https://github.com/Toblerity/Fiona/releases/tag/1.10a1) which might be relevant. 
([This commit](https://github.com/Toblerity/Fiona/pull/1334/files#) contains a test showing the filepath formatting for windows.) But it might also be me who don't know how to respect windows file system structure.

Maybe @burakekim has some intel? :D

tests/datasets/test_geo.py
```diff
@@ -368,6 +377,143 @@ def test_no_all_bands(self) -> None:
...
+        specific_file_not_zipped = list_directory_recursive(
+            dir_not_zipped, filename_glob
+        )[0]
+        filepath_within_dir = specific_file_not_zipped.replace(dir_not_zipped, '')
+
+        specific_file_zipped = f'zip://{dir_zipped}!{filepath_within_dir}'
+        if sys.platform == 'win32':
+            specific_file_zipped = os.path.join(
+                'zip://D:', '{dir_zipped}!{filepath_within_dir}'
+            )
```
adriantre, line 427: Tried my luck looking at [this test in fiona](https://github.com/Toblerity/Fiona/blob/908bf4f8a07d36ec1d55ba45f6550a3699520b6f/tests/test__path.py#L8), but I guess this is only required when using abspath.

Would I need to specify absolute path? My intuition tells me no

adamjstewart:
RE Windows support: this is definitely a best effort kind of thing. The tests have to pass on all platforms, but if there are certain file prefixes that don't work on all platforms, we don't need to test those.

tests/datasets/test_geo.py
adamjstewart: Definitely no need to test abs path. We aren't trying to test fiona, we're trying to test TorchGeo.

adriantre:
I snuck in a skipif windows :)

.gitignore
```diff
@@ -149,3 +149,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# Mac
+.DS_Store
```
adamjstewart, line 154: Can you undo this change? User-specific ignores should be in the global gitignore, not the project gitignore.

torchgeo/datasets/geo.py
```diff
@@ -308,14 +307,16 @@ def files(self) -> list[Path]:
...
         for path in paths:
-            if os.path.isdir(path):
-                pathname = os.path.join(path, '**', self.filename_glob)
-                files |= set(glob.iglob(pathname, recursive=True))
-            elif (os.path.isfile(path) or path_is_vsi(path)) and fnmatch.fnmatch(
-                str(path), f'*{self.filename_glob}'
-            ):
-                files.add(path)
-            elif not hasattr(self, 'download'):
+            file_not_found = True
```
adamjstewart, line 310: What's the purpose of this variable? Can't we just use if-elif-else? There are no loops

torchgeo/datasets/utils.py
```diff
@@ -631,6 +635,74 @@ def path_is_vsi(path: Path) -> bool:
...
+    Args:
+        root: directory to list. For VSI these will have prefix
+            e.g. /vsiaz/ or az:// for azure blob storage
+        filename_glob: filename pattern to filter filenames
+
+    Returns:
+        A list of all file paths matching filename_glob in the root directory or its
+        subdirectories.
+
+    .. versionadded:: #0.6
```
adamjstewart, line 686:
```suggestion
    .. versionadded:: 0.6
```

torchgeo/datasets/utils.py
```diff
@@ -631,6 +635,74 @@ def path_is_vsi(path: Path) -> bool:
...
+        except FionaValueError as e:
+            if 'is not a directory' in str(e):
+                # Assuming dir is a file as it is not a directory
+                # fiona.listdir can throw FionaValueError for only two reasons
+                files.append(dir)
+            else:
+                # fiona.listdir can throw FionaValueError for only two reasons
+                # 1. 'is not a directory'
+                # 2. 'does not exist'
+                raise
```
adamjstewart, line 668: We should be more specific about what we raise here. Is this try-except really the only way to do this? Seems dirty.

torchgeo/datasets/utils.py
```diff
@@ -631,6 +635,74 @@ def path_is_vsi(path: Path) -> bool:
...
+
+    Args:
+        root: directory to list. These must contain the prefix for the VSI
+            e.g. '/vsiaz/' or 'az://' for azure blob storage
+                 '/vsizip/' or 'zip://' for zipped archives
+
+    Returns:
+        A list of all file paths matching filename_glob in the root VSI directory or its
+        subdirectories.
+
```
adamjstewart, line 649: Need to document `Raises:` if we keep it

torchgeo/datasets/utils.py
```diff
@@ -631,6 +635,74 @@ def path_is_vsi(path: Path) -> bool:
     return '://' in str(path) or str(path).startswith('/vsi')
 
 
+def listdir_vsi_recursive(root: Path) -> list[str]:
+    """Lists all files in Virtual File Systems (VSI) recursively.
+
+    Args:
+        root: directory to list. These must contain the prefix for the VSI
+            e.g. '/vsiaz/' or 'az://' for azure blob storage
+                 '/vsizip/' or 'zip://' for zipped archives
```
adamjstewart, line 644:
```suggestion
            (e.g., '/vsiaz/' or 'az://' for azure blob storage, or
            '/vsizip/' or 'zip://' for zipped archives).
```

torchgeo/datasets/utils.py
```diff
@@ -631,6 +635,74 @@ def path_is_vsi(path: Path) -> bool:
...
+            all_files = listdir_vsi_recursive(root)
+        except FionaValueError:
+            # To match the behaviour of glob.iglob we silently return empty list
+            # for non-existing root.
+            pass
+        # Prefix glob with wildcard to ignore directories
+        files = fnmatch.filter(all_files, os.path.join('*', filename_glob))
+    else:
+        pathname = os.path.join(root, '**', filename_glob)
+        files = list(glob.iglob(pathname, recursive=True))
```
adamjstewart, line 702:
```suggestion
        files = glob.glob(pathname, recursive=True)
```

torchgeo/datasets/utils.py
```diff
@@ -631,6 +635,74 @@ def path_is_vsi(path: Path) -> bool:
...
+                files.append(dir)
+            else:
+                # fiona.listdir can throw FionaValueError for only two reasons
+                # 1. 'is not a directory'
+                # 2. 'does not exist'
+                raise
+    return files
+
+
+def list_directory_recursive(root: Path, filename_glob: str) -> list[str]:
```
adamjstewart, line 672: Should we return a set instead of a list? The `files` method is the only place they are used and uses a set.

torchgeo/datasets/utils.py
```diff
@@ -631,6 +635,74 @@ def path_is_vsi(path: Path) -> bool:
...
+        # Change type to match expected input to filter
+        all_files: list[str] = []
+        try:
+            all_files = listdir_vsi_recursive(root)
+        except FionaValueError:
+            # To match the behaviour of glob.iglob we silently return empty list
+            # for non-existing root.
+            pass
+        # Prefix glob with wildcard to ignore directories
+        files = fnmatch.filter(all_files, os.path.join('*', filename_glob))
```
adamjstewart, line 699: I chose not to use `os.path.join` here because the VSI may be a URL and use / even on Windows where `os.path.join` uses \

adamjstewart (commented):
General comments:

1. Code looks kinda hacky, wondering if there's a simpler way
2. Tests look extremely complex, wish we could simplify them greatly

torchgeo/datasets/utils.py
```diff
@@ -631,6 +635,74 @@ def path_is_vsi(path: Path) -> bool:
     return '://' in str(path) or str(path).startswith('/vsi')
 
 
+def listdir_vsi_recursive(root: Path) -> list[str]:
+    """Lists all files in Virtual File Systems (VSI) recursively.
```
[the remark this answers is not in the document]
adriantre: Yeah, this is a bit confusing. I can change to VFS as I have not found any description of what VSI stands for.
https://gdal.org/en/latest/user/virtual_file_systems.html

torchgeo/datasets/geo.py
adriantre: There are two cases where I currently warn:
1. it is a local file, but does not match the filename_glob.
2. It is a directory or a VFS, but no files matching filename_glob is found.

The first one is in nested `if`s so that if the inner if fails, it will not go the the outer elif. It should not go to the outer elif, at it is indeed a file.

I could duplicate the warning, but thought it was cleaner to have the warning defined once. 

Or we could remove the warning for case 1. as in this case the user has provided a path that is ignored?

torchgeo/datasets/utils.py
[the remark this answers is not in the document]
adriantre: Might be "GDAL Virtual File System Interface"
https://rasterio.readthedocs.io/en/latest/topics/vsi.html

torchgeo/datasets/utils.py
adriantre: Would checking the error message and raising our own error be better?

Here is the error that might occur:
https://github.com/Toblerity/Fiona/blob/7490643d145ce0be506a210abd2802fd0fff63f4/fiona/ogrext.pyx#L2171

torchgeo/datasets/utils.py
adriantre: Thats ok by me. The only reason I did not is because similar methods returns a list (iglob, fnmatch etc).

adriantre: We could change this method to remove duplicates (set) and sort it, then return a list as well. Or is it better to be explicit about that in the files-property?

torchgeo/datasets/utils.py
adriantre: Ah, yes. I [see it now](https://github.com/microsoft/torchgeo/blob/042b75ea9c884d2743bb46687ee0bd069dfc2a79/torchgeo/datasets/geo.py#L315).

This should be the reason the reason for the failing test on windows. I need to do the same in [listdir_vfs_recursive](https://github.com/microsoft/torchgeo/blob/51a3d64f6f0339cb26707cea2579b952fd725605/torchgeo/datasets/utils.py#L662)

adriantre: Updated it and removed the skipif windows

torchgeo/datasets/utils.py
adriantre: If Fiona accepts [this](https://github.com/Toblerity/Fiona/pull/1436), would this better? (Would still need some workaround until then).

```python
if not fiona.dir_exists(dir):
    continue
elif fiona.isdir(dir):
    subdirs = fiona.listdir(dir)
    dirs.extend([f'{dir}/{subdir}' for subdir in subdirs])
else:
    files.append(dir)
```

torchgeo/datasets/utils.py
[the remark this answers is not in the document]
adriantre: Changed.

torchgeo/datasets/utils.py
adriantre: Done

adriantre:
> General comments:
> 
> 1. Code looks kinda hacky, wondering if there's a simpler way
> 2. Tests look extremely complex, wish we could simplify them greatly

Let's work our way through this.

torchgeo/datasets/geo.py
adriantre: I did this because if it **is** a file, but does not match filename_glob, it does not need to list_directory. But I guess it is does not hurt to allow it to try listing the file.

torchgeo/datasets/utils.py
adriantre: So the feedback from fiona is that "torchgeo use a dedicated package like fsspec or the Azure Python module." But then we don't get the silver bullet for all of them, and it seems like additional libraries for each file system will have to be installed.

adamjstewart: We're definitely trying to minimize our dependencies when possible (0.6 actually removes more dependencies than it adds) so I would prefer not to do this. Worst case, users can always use these libraries themselves to get a list of files, then pass these to a GeoDataset (like in 0.5).

adriantre: So what do you think, merge this, or go with my first proposal; make it easier to handle "non-local-files" by adding a method that the users can override?

```python
if os.path.isdir(path):
    pathname = os.path.join(path, '**', self.filename_glob)
elif (os.path.isfile(path) or path_is_vsi(path)) and fnmatch.fnmatch(
    str(path), f'*{self.filename_glob}'
):
    files.add(path)
elif not hasattr(self, 'download'):
    self.handle_non_local_file()
    
def handle_non_local_file(self):
    warnings.warn(
        f"Could not find any relevant files for provided path '{path}'. "
        f'Path was ignored.',
        UserWarning,
    )
```

tests/datasets/test_geo.py
```diff
@@ -368,6 +376,88 @@ def test_no_all_bands(self) -> None:
...
+            os.path.join(
+                'tests',
+                'data',
+                'sentinel2',
+                'S2A_MSIL2A_20220414T110751_N0400_R108_T26EMU_20220414T165533.SAFE',
+            )
+        ],
+        indirect=True,
+    )
+    def test_zipped_specific_file_dir(self, temp_archive: tuple[str, str]) -> None:
```
adamjstewart, line 416: To simplify the testing drastically, I think I would rather:

1. Create all uncompressed (.tif) and compressed (.zip) files in `tests/data/<whatever>` and store them in git so we don't have to generate and delete them on the fly
2. Have only 2 test functions (expected to pass, expected to fail)
3. Parametrize each of these test functions with all supported inputs (.tif, zipped file, zipped dir, etc.)

This will remove a lot of the code duplication and make things easier to add new tests for (just a single line in parametrize). What do you think?

torchgeo/datasets/utils.py
```diff
@@ -607,18 +611,22 @@ def percentile_normalization(
     return img_normalized
 
 
-def path_is_vsi(path: Path) -> bool:
-    """Checks if the given path is pointing to a Virtual File System.
+def path_is_gdal_vsi(path: Path) -> bool:
```
adamjstewart, line 614: I'm still considering deleting this function. It's only used in a single place, and it's only 1 line long, so we could just replace the if-statement on line 693 with this line. I would like to keep the documentation somehow though.

torchgeo/datasets/utils.py
```diff
@@ -631,6 +639,74 @@ def path_is_vsi(path: Path) -> bool:
     return '://' in str(path) or str(path).startswith('/vsi')
 
 
+def listdir_vfs_recursive(root: Path) -> list[str]:
```
adamjstewart, line 642: Can we make all of these hidden functions (prefixed with an underscore)? They aren't currently documented or exported via `__all__`, but I don't want to take any chances on letting people use these.

adamjstewart (commented):
Still need to devote some time to peeking into `fiona.listdir` and seeing if there's a way to recursively do this (or add this option to fiona if not) instead of writing our own while-statement, but I'm starting to understand and come to terms with your implementation. Let's focus on simplifying the tests before finalizing the code. If needed, we can also simplify the tests in a separate PR so that this PR only adds a couple new lines to testing (zip files in parametrize).

pyproject.toml
```diff
@@ -39,7 +39,8 @@ dependencies = [
     # einops 0.3+ required for einops.repeat
     "einops>=0.3",
     # fiona 1.8.21+ required for Python 3.10 wheels
-    "fiona>=1.8.21",
+    # fiona 1.9+ required for fiona.listdir
+    "fiona>=1.9",
```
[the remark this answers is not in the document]
adriantre: Yes, I think we are good

adriantre:
> Still need to devote some time to peeking into `fiona.listdir` and seeing if there's a way to recursively do this (or add this option to fiona if not) instead of writing our own while-statement

FYI: https://github.com/Toblerity/Fiona/issues/1275

torchgeo/datasets/utils.py
adriantre: Thats a lot of documentation to have in-line in the other method :P

tests/datasets/test_geo.py
```diff
@@ -84,6 +84,19 @@ def __len__(self) -> int:
         return 2
 
 
+@pytest.fixture(scope='module')
+def temp_archive(request: SubRequest) -> Generator[tuple[str, str], None, None]:
```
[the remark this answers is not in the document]
adriantre: I made a module-scoped fixture instead, but I can opt for class-scoped if it will only be used for TestGeoDataset.

tests/datasets/test_geo.py
adriantre: I'll take a look! I like the idea.

Might be annoying to keep the archives up-to-date with the source if we ever modify them though. The tests does not seem to be much slower when archiving on the fly, but it would reduce code-complexity for the test

adriantre:
I'm very happy with how this now looks 😎

torchgeo/datasets/utils.py
```diff
@@ -990,3 +969,54 @@ def convert_poly_coords(
...
+
+    Args:
+        root: VSI path to list (e.g., ``/vsiaz/container/`` for Azure Blob
+            Storage, ``/vsizip/archive.zip`` for zip archives).
+
+    Returns:
+        A list of all file paths under *root*, or an empty list if *root*
+        does not exist.
+
+    .. versionadded:: 0.9
```
ashnair1:
```diff
-    .. versionadded:: 0.9
+    .. versionadded:: 0.10
```

ashnair1 (changes requested):
This PR looks good to me once the version annotation is fixed.

torchgeo/datasets/utils.py
adriantre: Done!

ashnair1 (dismissed):

calebrob6:
@adamjstewart you need to approve this change -- it looks good to me

calebrob6 (dismissed):

tests/datasets/conftest.py
```diff
@@ -31,6 +33,23 @@ def download_url(monkeypatch: MonkeyPatch, request: SubRequest) -> None:
     monkeypatch.setattr('torchgeo.datasets.copernicus.lcz_s2.download_url', copy)
 
 
+@pytest.fixture(scope='module')
+def module_tmp_path(tmp_path_factory: TempPathFactory) -> Path:
```
adamjstewart: This doesn't seem to be used anywhere than inside temp_archive, could it be moved there instead?

tests/datasets/test_geo.py
```diff
@@ -288,29 +288,28 @@ def test_files_property_for_non_existing_file_or_dir(self, tmp_path: Path) -> No
...
+
+    def test_files_property_deterministic(self, tmp_path: Path) -> None:
         """Ensure that the list of files is consistent regardless of their original
         order.
         """
-        paths1 = ['file://file3.tif', 'file://file1.tif', 'file://file2.tif']
-        paths2 = ['file://file2.tif', 'file://file3.tif', 'file://file1.tif']
+        files = ['file3.tif', 'file1.tif', 'file2.tif']
+        paths1 = [tmp_path / fake_file for fake_file in files]
+        paths2 = paths1[::-1]  # reverse order
```
adamjstewart, line 311: Why reverse order? I'll admit I have no idea what these tests are for...

tests/datasets/conftest.py
```diff
@@ -31,6 +33,23 @@ def download_url(monkeypatch: MonkeyPatch, request: SubRequest) -> None:
...
+    return tmp_path_factory.mktemp('module_tmp')
+
+
+@pytest.fixture(scope='module')
+def temp_archive(
```
adamjstewart, line 37: Instead of doing this on the fly, couldn't we store both raw data and a zipped version in git?

torchgeo/datasets/utils.py
```diff
@@ -990,3 +969,54 @@ def convert_poly_coords(
...
+        does not exist.
+
+    .. versionadded:: 0.10
+    """
+    try:
+        entries = pyogrio.vsi_listtree(str(root))
+    except NotADirectoryError:
+        return [str(root)]
+    except FileNotFoundError:
+        return []
```
adamjstewart, line 1010: Is this something we want? Why not raise an error if root doesn't exist?

torchgeo/datasets/geo.py
```diff
@@ -321,21 +321,16 @@ def files(self) -> list[str]:
...
+            if found:
+                files.update(found)
+            elif not os.path.isdir(path) and not hasattr(self, 'download'):
                 warnings.warn(
                     f"Could not find any relevant files for provided path '{path}'. "
                     f'Path was ignored.',
                     UserWarning,
                 )
 
-        # Sort the output to enforce deterministic behavior.
```
adamjstewart, line 338: Can we keep these comments? It's important that this remains deterministic if we want datasets to be reproducible. Without this comment, we'll quickly forget why we are sorting.

adamjstewart (changes requested):
I'll fully admit I don't really understand most of the tests you added...

tests/datasets/conftest.py
adriantre: I did it like this to allow flexibility. When writing test you can pick and choose which ever dataset you need. And to avoid having to keep archive up-to-date with data-creation.

Why do you feel it is better to track zipped version(s)?

tests/datasets/test_geo.py
adriantre: The merge looks a bit messy, please use split-layout i github. 

Image

The test is (still) verifying that ordering is deterministic. Reversing order makes it clear that it is the same input with different ordering, and that the output should be exactly the same regardless.

Removing the `file://` prefix as the tmp_path fixture + touch ensure they exist.

torchgeo/datasets/utils.py
adriantre: This mirrors what happens for non-vsi listing. We have never raised if the root does not exist, or am I mistaken? 

But if no files are found, we raise in property `files` like before.
```python
found = set(find_files(path, self.filename_glob))
if found:
    files.update(found)
elif not os.path.isdir(path) and not hasattr(self, 'download'):
    warn
```

torchgeo/datasets/geo.py
adriantre: Fully agree, this slipped when refactoring

tests/datasets/conftest.py
adriantre: Collapsed it now. I thought I had to override scope for it, but I see now that it is not needed

tests/datasets/conftest.py
adamjstewart: Mostly consistency and simplicity. I agree we could someday do all test data creation on the fly, although it's a bit slower.

torchgeo/datasets/utils.py
adamjstewart: Alright, I think we've talked about this before, trying to jog my memory.

I think it makes sense to warn about non-existent `paths` and error if all `paths` are non-existent? There is a slight difference if we are going to download things or not, but that shouldn't affect VSI, right?

tests/datasets/test_geo.py
adriantre: All tests are verifying intention. Does it find files and only when it's supposed to, and does it find the same files inside zip-archives as regular directories.

How can I help in explaining it?

I do believe this is the minimal test suite for coverage and to document intention.

torchgeo/datasets/utils.py
adriantre: Sounds good! Please verify my understanding:
> warn for all provided paths that do not exist, and raise if none of them exists

The only thing missing for this is a final step of files property:
```python
if not files and not hasattr(self,'download'): raise
```

Datasets having `self.download` will never warn og raise. If you use VSI with a class that has self.download you won't get a warning. But that problem is on the concept of `self.download` than VSI, in my opinion.

adamjstewart: Does this behavior make sense?

* Downloadable datasets: If _anything_ is missing, download it if download is True, raise if not.
* Non-downloadable datasets: If _anything_ is missing, warn. If _everything_ is missing, raise.

If we assume that some VSI datasets could be downloadable, then I agree this behavior should be independent of whether or not we are VSI. Only reason I bring it up in this PR is because we are refactoring things in the same code.

adriantre: This makes sense and is essentially what this does, after adding the raise at the end.

But the raising of not able to download should happen at download-time, not here, agree?

Or did you mean that we should raise here if everything is missing, regardless of downloadable?

adamjstewart: Nah, as long as it happens _somewhere_, it doesn't have to happen here. The code logic is just becoming complex enough that I don't know if or where these warnings/errors will actually happen.

adriantre: Hehe. Then I will put in the raise as suggested above. After that I feel this PR is complete

adriantre: I implemented the Raise in a new PR. Can not raise at this point because downloadable datasets use `files` as existance check and would have to try except. Let's merge this as is.

[The new PR](https://github.com/adriantre/torchgeo/pull/2) alleviates `files` the responsibility of file existence during download logic. Will rebase it on main when this PR is merged. 

I used a lot of brain cells here, and I'm very confident that I have found a good solution.

tests/datasets/conftest.py
adriantre: Are you willing to accept this as is? I understand your point, but I strongly believe that we should make static archives at a later point if there is a need outside these couple of tests.

torchgeo/datasets/utils.py
adamjstewart: I appreciate the use of brain cells, those are even harder to come by than tokens these days 😄

torchgeo/datasets/utils.py
```diff
@@ -8,6 +8,8 @@
 
 import bz2
 import contextlib
+import fnmatch
+import glob as glob_module
```
adamjstewart: Why the rename?

adriantre: That is a good question. No need for it here, I suspect Claude, when helping me find the vsi-listing functionality inside pyogrio, made this alias when refactoring. I've seen it used in Claude plugin source code.

Btw, I would never have found a good alternative to fiona.listdir without Claude. It was super effective in finding an alternative that perfectly matched my needs.

adamjstewart: FYI, once our AI policy is adopted, comments like "I don't know, AI did it" will probably not go over well. AI is a good starting point but not a good stopping point for code refactors. This also explains why the tests are illegible...

adriantre: I think the tests are fine. This is a PR I started on Jun 5, 2023 and probably before that. The tests and code has only become cleaner after I got help from AI.

If you don't like it, feel free to close this PR as a won't-fix.

adriantre:
Last try from my side. I have done my best to make the tests easy to understand.

If you want any changes from my side, I expect more constructive guidance than "tests are illegible...". Ok?

torchgeo/datasets/utils.py
```diff
@@ -875,30 +878,6 @@ def quantile_normalization(
     return torch.clamp(img, 0, 1)
 
 
-def path_is_vsi(path: Path) -> bool:
```
isaaccorley, line 878: Let's not remove this since we will want to deprecate it since it's not a public method

isaaccorley (changes requested):
Small change but so far lgtm

torchgeo/datasets/utils.py
adamjstewart: Technically don't need to deprecate _because_ it isn't public.

adriantre: This is a chance for me to learn. 

Does that mean that, since it was not included in `__all__` in `__init__.py` (`import *`), it was not public?
And what's the verdict? Is it safe to remove here, or must we deprecate?

adamjstewart: I think this is something that the maintainers disagree on, so it's a chance for all of us to learn.

IMO, things should only be public if they are in the API docs. And the API docs only include things with import aliases in `__init__.py`. We reserve the right to add/remove/rename files/classes/functions if they aren't documented. Everything that isn't documented is an implementation detail.

I think @isaaccorley would consider things that aren't prefixed with an underscore as public? This would prevent us from renaming any files, for example. We could rename half of TorchGeo to use underscores to make this more clear, but IMO it's implied that it isn't public if it isn't documented.

isaaccorley: Yeah I originally said this was public because other functions in this same script were private in the literal sense. However I agree with @adamjstewart on this now that it's not surfaced in the docs so it's not truly "public".

isaaccorley (approved):

adamjstewart (approved):
Two other maintainers approved so don't want to hold this up any longer. The logic in core is very tricky, but we should get bug reports pretty quickly if there are any corner cases we aren't properly handling (#test-in-deployment). The tests still have a bit of complexity, but they look much better now. Thanks for sticking with this one for 3 long years, glad to finally merge it!

calebrob6:
@adriantre do you have a remote data bucket that you use to test this with IRL?

adriantre:
I've used this branch for three years, haha. Not for training, but predict. I will test out the main branch to make sure the last refactors also work the same way 👍🏻

adriantre:
One thing I could have added to the docstring is that the vsi-functionality requires `paths`to be provided as str type. Or more specifically, pathlib.Path will collapse repreadet forward-slash, e.g. `/vsizip//vsiaz//home...` -> `/vsizip/vsiaz/home...` which gdal would not be able to interpret.

What do you guys think? Worth adding to the docstring, or try-excepting in any way?

adamjstewart:
Hmm, I would consider that a bug (feature?) in pathlib, not in TorchGeo. If users complain, we can add explicit documentation. But if not, let's keep it simple.

adriantre:
@calebrob6 

Here is a full script if you want to try it out yourself! You will need a CDSE account.
Fun fact: S1 GRD SAFE products are not projected / GCPs-only, so this script uses everything I did in #3404 too 😎

```python
from collections.abc import Sequence

from dotenv import load_dotenv

from torchgeo.samplers import RandomPatchSampler

# Loads the CDSE S3 config from a local .env, e.g.:
#   AWS_S3_ENDPOINT=eodata.dataspace.copernicus.eu
#   AWS_VIRTUAL_HOSTING=FALSE
#   AWS_ACCESS_KEY_ID=...        # CDSE-issued S3 key, not an AWS account
#   AWS_SECRET_ACCESS_KEY=...
load_dotenv()


class S1SafeGRD(RasterDataset):
    """Sentinel-1 IW GRD measurement tiffs from an ESA/CDSE ``.SAFE`` product.

    The two polarizations differ in both the band token and a trailing image
    index (``-001`` vs ``-002``), so they can't be derived from each other by
    band-token substitution. Use one polarization per dataset and union them.
    """

    filename_glob = 's1*-grd-{}-*.tiff'
    filename_regex = r"""
        ^s1[a-d]-iw-grd
        -(?P<band>[vh]{2})
        -(?P<date>\d{8}t\d{6})
        -\d{8}t\d{6}-\d{6}-[0-9a-f]{6}-\d{3}
        \.tiff?$
    """
    date_format = '%Y%m%dt%H%M%S'
    all_bands = ('vv', 'vh')
    separate_files = True

    def __init__(self, paths: str, bands: Sequence[str] = ('vv',), **kwargs) -> None:
        assert len(bands) == 1, 'use one polarization per dataset; union for both'
        self.filename_glob = self.filename_glob.format(bands[0])
        super().__init__(paths, bands=bands, **kwargs)


if __name__ == '__main__':
    safe = (
        '/vsis3/eodata/Sentinel-1/SAR/IW_GRDH_1S/2026/06/10/'
        'S1A_IW_GRDH_1SDV_20260610T061923_20260610T061948_064903_082DC0_2148.SAFE'
    )

    vv = S1SafeGRD(safe, bands=('vv',))
    vh = S1SafeGRD(safe, bands=('vh',))

    dataset = vv | vh

    sampler = RandomPatchSampler(dataset, size=160)

    # You might find a nodata patch, sample until you're lucky
    index = next(iter(sampler))
    sample = dataset[index]
```

adamjstewart:
Should we change the Sentinel datasets to read from .SAFE products instead of .jp2, or should we wait until they reformat everything in Zarr anyway?

adriantre:
I have only used .SAFE myself. Is the currently supported format something that came out of esa snap? My guess is that most users would preprocess with snap or perform some kind of DEM referencing. Depends on the use case, but we could support both?

SAFE contains the manifest.json metadata file which gdal can read directly. This way the multi band file challenge vanishes. (Same for Sentinel2). The filename_regex would have to extract timestamp from the SAFE folder name, as the filename in this case is manifest.json. Or better yet read the values from the opened dataset metadata. As I did for Sentinel2 footprint

adamjstewart:
The currently supported format came out of my own ignorance of a native .SAFE driver in GDAL. I just saw files that looked like images and designed TorchGeo to process them. But a lot of work goes into figuring out the right paths to each file at each resolution. We could get rid of a lot of complexity if we only support the .SAFE format.

PR #1400, opened 2023-06-05
SSL4EO-L Benchmark ETM+: copy nodata pixels from img to mask

adamjstewart:
We were using the mask from NLCD/CDL but not copying the nodata pixels from the Landsat 7 ETM+ SLC-off image. Therefore, the model was being penalized for not correctly predicting classes where it didn't even have imagery to use.

Example:

cdl_nodata
nlcd_nodata


adamjstewart:
Will also upload to HF and update the checksums once this finishes running.

nilsleh (approved):

isaaccorley (approved):

PR #1413, opened 2023-06-13
fix: Support tensors and arrays for class_weight

ntw-au:
Avoids ambiguous truth value `ValueError` when the class_weight input parameter is either a PyTorch tensor or a NumPy array.

Repro follows—also fails with `np.array` instead `torch.tensor`, succeeds with Python sequence.

```
import torch
from torchgeo.trainers.segmentation import SemanticSegmentationTask

SemanticSegmentationTask(
    model='unet',
    backbone='resnet101',
    loss='ce',
    learning_rate=0.01,
    weights=None,
    class_weights=torch.tensor([0.25, 0.5, 0.25]),
    in_channels=1,
    num_classes=3,
    ignore_index=None
)
```

Expected:

```
<prints PyTorch model>
```

Actual:

```
  File "<path>/torchgeo/torchgeo/trainers/segmentation.py", line 171, in __init__
    self.config_task()
  File "<path>/torchgeo/torchgeo/trainers/segmentation.py", line 69, in config_task
    torch.FloatTensor(self.class_weights) if self.class_weights else None
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

### DIFF
--- a/tests/trainers/test_segmentation.py
+++ b/tests/trainers/test_segmentation.py
@@ -5,6 +5,7 @@ import os
 from pathlib import Path
 from typing import Any, cast
 
+import numpy as np
 import pytest
 import segmentation_models_pytorch as smp
 import timm
@@ -262,3 +263,23 @@ class TestSemanticSegmentationTask:
                 for param in model.model.segmentation_head.parameters()
             ]
         )
+
+    @pytest.mark.parametrize(
+        "class_weights", [torch.tensor([1, 2, 3]), np.array([1, 2, 3]), [1, 2, 3]]
+    )
+    def test_classweights_valid(
+        self, class_weights: Any, model_kwargs: dict[Any, Any]
+    ) -> None:
+        model_kwargs["class_weights"] = class_weights
+        sst = SemanticSegmentationTask(**model_kwargs)
+        assert isinstance(sst.loss.weight, torch.Tensor)
+        assert torch.equal(sst.loss.weight, torch.tensor([1.0, 2.0, 3.0]))
+        assert sst.loss.weight.dtype == torch.float32
+
+    @pytest.mark.parametrize("class_weights", [[], None])
+    def test_classweights_empty(
+        self, class_weights: Any, model_kwargs: dict[Any, Any]
+    ) -> None:
+        model_kwargs["class_weights"] = class_weights
+        sst = SemanticSegmentationTask(**model_kwargs)
+        assert sst.loss.weight is None
--- a/torchgeo/trainers/segmentation.py
+++ b/torchgeo/trainers/segmentation.py
@@ -65,9 +65,12 @@ class SemanticSegmentationTask(LightningModule):
         if self.hyperparams["loss"] == "ce":
             ignore_value = -1000 if self.ignore_index is None else self.ignore_index
 
-            class_weights = (
-                torch.FloatTensor(self.class_weights) if self.class_weights else None
-            )
+            class_weights = None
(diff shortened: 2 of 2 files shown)


ntw-au:
@microsoft-github-policy-service agree company="Geomatic.AI Pty Ltd"

torchgeo/trainers/segmentation.py
```diff
@@ -65,9 +65,12 @@ def config_task(self) -> None:
...
             ignore_value = -1000 if self.ignore_index is None else self.ignore_index
 
-            class_weights = (
-                torch.FloatTensor(self.class_weights) if self.class_weights else None
-            )
+            class_weights = None
+            if isinstance(self.class_weights, torch.Tensor):
+                class_weights = self.class_weights.to(dtype=torch.float32)
+            elif hasattr(self.class_weights, "__array__") or self.class_weights:
+                class_weights = torch.tensor(self.class_weights, dtype=torch.float32)
```
adamjstewart, line 72: We also want to be able to support lists and other sequences. How about:
```suggestion
            if self.class_weights is not None:
                class_weights = torch.tensor(self.class_weights, dtype=torch.float32)
```

adamjstewart (commented):
At some point before the 0.5 release (expected in August) I'm planning on making all arguments explicit (as opposed to using `**kwargs`) so we can add type hints. So remind me to make sure all sequences are permitted, not just Tensors.

tests/trainers/test_segmentation.py
```diff
@@ -262,3 +263,23 @@ def test_freeze_decoder(
                 for param in model.model.segmentation_head.parameters()
             ]
         )
+
+    @pytest.mark.parametrize(
+        "class_weights", [torch.tensor([1, 2, 3]), np.array([1, 2, 3]), [1, 2, 3]]
```
adamjstewart, line 268: Can you also add a normal list here?

ntw-au: The parameters already include a normal list, so covers PyTorch tensor, NumPy array and Python list.

torchgeo/trainers/segmentation.py
ntw-au: Yes, other sequence types (including lists) are already covered by the `or self.class_weights` part of the branch test. This mimics the previous behaviour of the function (that only supported sequences), where `class_weights` would be `None` if the input sequence was non-null but empty.

It became apparent during development that we needed a separate branch to handle PyTorch tensors, as copying a tensor via the `torch.tensor` function raises warnings as follows:

```
>>> import torch
>>> a = torch.arange(10)
>>> b = torch.tensor(a)
<stdin>:1: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
```

adamjstewart (commented):
I don't yet see a simpler way to handle this other than documenting that only lists are allowed. Maybe @nsutezo can review too.

calebrob6 (approved):

adamjstewart (approved):

adamjstewart:
Working on type hints in #1541 and it's actually difficult to support lists, arrays, and tensors. Might just support tensors since that's what `nn.CrossEntropyLoss` supports.

ntw-au:
I think that's a good idea, as a user I'd expect tensors as first preference, and it's easy for callers to change arrays and lists into a tensor themselves.

PR #1424, opened 2023-06-17
SSL4EO-L: add download support

adamjstewart:
Preliminary download support for SSL4EO-L. Still need to test to make sure it works on the real dataset.


torchgeo/datasets/ssl4eo.py
```diff
@@ -157,6 +221,62 @@ def __len__(self) -> int:
...
         return len(self.scenes)
 
+    def _verify(self) -> None:
+        """Verify the integrity of the dataset.
+
+        Raises:
+            RuntimeError: if ``download=False`` but dataset is missing or checksum fails
+        """
+        # Check if the extracted files already exist
+        path = os.path.join(self.subdir, "*", "*", "all_bands.tif")
```
adamjstewart, line 231: This is likely going to be very slow and must be executed every time you instantiate the dataset. Thoughts on how to make this faster?

torchgeo/datasets/ssl4eo.py
```diff
@@ -157,6 +221,62 @@ def __len__(self) -> int:
...
+            download_url(
+                self.url.format(self.split, suffix),
+                self.root,
+                md5=md5 if self.checksum else None,
+            )
+
+    def _extract(self) -> None:
+        """Extract the dataset."""
+
+        # Concatenate all tarballs together
```
adamjstewart, line 270: I'm pretty confident that this won't work unless you have > 50 GB RAM on your system. Need to think about how to do this incrementally. I guess just read/write chunk-by-chunk. It's going to be very slow regardless.

torchgeo/datasets/ssl4eo.py
```diff
@@ -157,6 +221,62 @@ def __len__(self) -> int:
...
+
+    def _extract(self) -> None:
+        """Extract the dataset."""
+        # Concatenate all tarballs together
+        chunk_size = 2**15  # same as torchvision
+        path = self.subdir + ".tar.gz"
+        with open(path, "wb") as f:
+            for suffix in self.checksums[self.split]:
+                with open(path + suffix, "rb") as g:
+                    while chunk := g.read(chunk_size):
```
adamjstewart, line 274: This is my first time using assignment expressions, such a neat feature!

torchgeo/datasets/ssl4eo.py
adamjstewart: It's worse than I thought. 21 minutes to glob everything. There are a few ways to speed this up. We could use `00000*` instead of `*` so we're only checking for a subset of directories. This should work for all datasets. Alternatively, we could change the indices to be 1–250K instead of the original sample indices. This would make a parallel corpus across sensors more difficult, but we would no longer need to use os.listdir or glob.glob. Note that the sorted os.listdir in `__init__` is almost instantaneous so that isn't of concern.

calebrob6:
This looks good to me as long as you check that it works with the real dataset

adamjstewart:
Yep, I checked

calebrob6 (approved):

PR #1441, opened 2023-06-21
DataModules: improve error messages

### LINKED ISSUE #1307: Improve data module error messages
### Summary

We should improve our data module error messages by being more specific about what issues a data module may have.

### Rationale

The current error message just says:
```
Foo.setup does not define a 'foo_dataset'
```
but the actual error could be any of:

* Foo.setup does not define a foo_dataset or dataset
* Foo.setup does not define a foo_sampler or sampler or foo_batch_sampler or batch_sampler
* The dataset length is 0
* The sampler length is 0

We should explain which of the above is actually happening.

### Implementation

We added a new BaseDataModule class. We can add a method to that which will be inherited by subclasses. This method will be given a list of either datasets or sampler, then check to see if at least one of them is defined and has length > 0, and prints the appropriate error message if that is not the case.

### Alternatives

_No response_

### Additional information

_No response_

adamjstewart:
Closes #1307 

This implementation defines two new helper functions:

* `_valid_attribute`: checks if at least one of the datasets or samplers has length > 0
* `_dataloader_factory`: defines a `*_dataloader` depending on the split

The former handles all error messages and provides more specific and helpful error messages. Another bonus is that we avoid the majority of code duplication in these base classes, chopping off 30+ lines of code.

@yichiac I know you've run into this error message too many times, hopefully this will help with debugging.

### DIFF
--- a/torchgeo/datamodules/geo.py
+++ b/torchgeo/datamodules/geo.py
@@ -267,6 +267,7 @@ class GeoDataModule(BaseDataModule):
         batch_size = self._valid_attribute(f"{split}_batch_size", "batch_size")
 
         if isinstance(sampler, BatchGeoSampler):
+            batch_size = 1
             batch_sampler = sampler
             sampler = None
         else:
--- a/torchgeo/datamodules/geo.py
+++ b/torchgeo/datamodules/geo.py
@@ -267,7 +267,7 @@ class GeoDataModule(BaseDataModule):
         if isinstance(sampler, BatchGeoSampler):
             kwargs = {"batch_size": 1, "batch_sampler": sampler}
         else:
-            batch_size = getattr(self, f"{split}_batch_size", self.batch_size)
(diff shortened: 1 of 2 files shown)


torchgeo/datamodules/geo.py
```diff
@@ -220,6 +248,41 @@ def setup(self, stage: str) -> None:
                 self.test_dataset, self.patch_size, self.patch_size
             )
 
+    def _dataloader_factory(self, split: str) -> DataLoader[dict[str, Tensor]]:
+        """Implement one or more PyTorch DataLoaders.
+
+        Args:
+            split: Either 'train', 'val', 'test', or 'predict'.
```
nilsleh, line 255: In setup() there is no case for `predict` if I am not misunderstanding something.

adamjstewart: There is, we just don't set it by default because most datasets don't have a `predict` split.

nilsleh (approved):

PR #1442, opened 2023-06-22
Make RasterDataset accept list of files

### LINKED ISSUE #1427: RasterDataset.from_files
### Summary

We should add a `from_files` classmethod constructor to `RasterDataset` and `VectorDataset`.

### Rationale

Our current implementation allows the user to specify a root directory, which the class will recursively search for geospatial files. However, there are a few issues with this:

* It doesn't work well for remote files (#1399)
* It doesn't work well for drivers that split each scene into multiple auxiliary files
* It doesn't allow users to select a subset of data unless it is separated geographically or in separate directories

A secondary class constructor would alleviate these issues and allow users to specify a list of files any way they want to.

### Implementation

The idea would be to add `@classmethod` constructors to `RasterDataset` and `VectorDataset`. These methods will largely share the same code as `__init__` so we should move those to a common helper function. The only difference will be replacing the glob with a user-supplied list of files.

### Alternatives

We could optionally allow `root` to be a list and assume that this is a list of files, not a list of directories to search. We could even support both if we first check if the contents of the list are directories or files.

### Additional information

@calebrob6 @adriantre

adriantre:
Fix #1427

### DIFF
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -10,7 +10,7 @@ import os
 import re
 import sys
 from collections.abc import Sequence
-from typing import Any, Callable, Optional, cast
+from typing import Any, Callable, Optional, Union, cast
 
 import fiona
 import fiona.transform
@@ -329,7 +329,7 @@ class RasterDataset(GeoDataset):
 
     def __init__(
         self,
-        root: str = "data",
+        root: Union[str, list[str]] = "data",
         crs: Optional[CRS] = None,
         res: Optional[float] = None,
         bands: Optional[Sequence[str]] = None,
@@ -339,7 +339,8 @@ class RasterDataset(GeoDataset):
         """Initialize a new Dataset instance.
 
         Args:
-            root: root directory where dataset can be found
+            root: root directory or list of absolute filepaths where
+                dataset can be found
             crs: :term:`coordinate reference system (CRS)` to warp to
                 (defaults to the CRS of the first file found)
             res: resolution of the dataset in units of CRS
@@ -358,11 +359,22 @@ class RasterDataset(GeoDataset):
         self.bands = bands or self.all_bands
         self.cache = cache
 
+        if not isinstance(root, str):
+            root = [cast(str, root)]
+
+        filespaths = []
+        for dir_or_file in root:
+            _, ext = os.path.splitext(dir_or_file)
+            if ext:
+                filespaths.append(dir_or_file)
+            else:
+                pathname = os.path.join(dir_or_file, "**", self.filename_glob)
+                filespaths.extend(glob.iglob(pathname, recursive=True))
+
         # Populate the dataset index
         i = 0
-        pathname = os.path.join(root, "**", self.filename_glob)
         filename_regex = re.compile(self.filename_regex, re.VERBOSE)
-        for filepath in glob.iglob(pathname, recursive=True):
+        for filepath in filespaths:
             match = re.match(filename_regex, os.path.basename(filepath))
             if match is not None:
                 try:


adriantre:
Here is the inital proposal. If `root` is a string, add it to a list and handle it. But we should allow `root` to be e.g. a `pathlib.Path` too.

Are there any collection of types that are appropriate for checking either "path-like" or "not list-like"?

torchgeo/datasets/geo.py
```diff
@@ -329,7 +329,7 @@ def dtype(self) -> torch.dtype:
 
     def __init__(
         self,
-        root: str = "data",
+        root: Union[str, list[str]] = "data",
```
adamjstewart, line 332: According to [PEP-519](https://peps.python.org/pep-0519/#provide-specific-type-hinting-support), the correct way to support all possible path types is:
```python
Union[os.PathLike, str, bytes]
```
However, there are hundreds of other places in TorchGeo that would need to be modified to support pathlib. Let's stick with str for now and consider supporting pathlib in a separate issue/PR.

adamjstewart (changes requested):
For directory vs. file, what if we use `os.path.exists` to determine whether or not the path is on the local filesystem. If it's local, use `os.path.isdir` or `os.path.isfile`. If it isn't, assume it's a path.

Alternative is to only support `str` directories and `list` files. I would also be satisfied with that solution if you want to keep it simple.

torchgeo/datasets/geo.py
```diff
@@ -339,7 +339,8 @@ def __init__(
         """Initialize a new Dataset instance.
 
         Args:
-            root: root directory where dataset can be found
+            root: root directory or list of absolute filepaths where
+                dataset can be found
```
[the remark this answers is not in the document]
adriantre: Renamed to paths. As you will see from mypy, a lot of datasets implements e.g. `_verify` that expect it to be a (single) directory. Would it be possible to generalise this method and put it in RasterDataset? And then handle paths instead of root.

torchgeo/datasets/geo.py
```diff
@@ -358,11 +359,22 @@ def __init__(
         self.bands = bands or self.all_bands
         self.cache = cache
 
+        if isinstance(root, str):
+            root = [root]
+
+        filespaths = []
+        for dir_or_file in root:
+            _, ext = os.path.splitext(dir_or_file)
```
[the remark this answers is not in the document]
adriantre: Changed this to use `os.isfile`

torchgeo/datasets/geo.py
adriantre: Agree, let's keep str | list for now

torchgeo/datasets/geo.py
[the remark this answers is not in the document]
adriantre: `_verify` methods generally seem to do:

If no file match `self.filename_glob` in root:
-> Do extract archive
If no archive:
-> Do download

And some check_integrity / checksum.

I think I have mostly updated these to allow `self.paths` to be a list, but introducing method `list_files`. The only "breaking" change is that archives are now extracted to the same dir as the archive, not root. This possible do not break stuff though?

I think these methods can be generalised, but this should be a separate issue/pr.

adamjstewart:
I think this is sufficient for a proof of concept. Wondering what @calebrob6 thinks as he's the one who's been asking for this feature for the longest.

torchgeo/datasets/landcoverai.py
```diff
@@ -236,8 +237,10 @@ def __init__(
            RuntimeError: if ``download=False`` and data is not found, or checksums
                don't match
         """
-        LandCoverAIBase.__init__(self, root, download, checksum)
-        RasterDataset.__init__(self, root, crs, res, transforms=transforms, cache=cache)
+        LandCoverAIBase.__init__(self, cast(str, paths), download, checksum)
```
adriantre, line 240: @adamjstewart I added this cast to make mypy happy while this is tested for RasterDatasets only. Do you have a best practice for highlighting this in the code? It will currently crash if LandCoverAIGeo is instantiated with a list, since LandCoverAIBase is not RasterDataset.

tests/datasets/test_geo.py
```diff
@@ -178,6 +178,26 @@ def sentinel(self, request: SubRequest) -> Sentinel2:
         cache = request.param[1]
         return Sentinel2(root, bands=bands, transforms=transforms, cache=cache)
 
+    def test_init_with_list_of_files(self) -> None:
```
adriantre, line 181: Added this quick and dirty test for proof of concept. Open to suggestions on how this should be tested.

torchgeo/datasets/geo.py
```diff
@@ -418,6 +420,36 @@ def __init__(
         self._crs = cast(CRS, crs)
         self._res = cast(float, res)
 
+    def list_files(self, filename_glob: Optional[str] = None) -> list[str]:
```
adriantre, line 423: @calebrob6 Do you have any feedback? This file contains is the main contributions. The rest is mainly refactoring.

torchgeo/datasets/landcoverai.py
adamjstewart: Is it possible for the base class to also support lists? Like we could just check if list and in that case only use the first value and ignore the rest. Something simple like that.

adriantre: Thanks, I added this temporary fix

adriantre:
@adamjstewart I am moving [this discussion](https://github.com/microsoft/torchgeo/pull/1399#issuecomment-1614880284) over here.

If we remove `os.path.exists` it will add non-existing directories to the list. In my opinion, since we are only implementing support for local files at this point, we should let the user know if they are providing non-existing paths. The current implementation does this by raising, and at the same time make it easy for the users to handle non-local files by giving them the possibility to override `handle_nonlocal_path`.

We could overcome this by checking both `if os.path.exists() and os.path.isdir():`, but then users that want to handle non-local files must override the whole `list_files` method instead of only the relevant part of the code.

calebrob6:
Hey all -- first, sorry it has taken me so long to get to this!

I just tested this and it behaves exactly as I would expect, so that's awesome.

My only concern is the situation in which someone overrides `handle_nonlocal_path(dir_or_file)` and then that file is matched to the `filename_regex`. We are using `os.path.basename(filepath)` to get the filename, then match that against `filename_regex`:
```
for filepath in self.list_files():
    ...
```
This might produce unexpected behavior (e.g. if I make a RemoteRasterDataset that takes a list of URLs to Azure blob containers that have SAS tokens appended to the end of them, then a string passed to `re.match(...)` would be something like "m_4512227_se_10_060_20200719.tif?st=2023-07-18T18%.....").

This seems minor and something that we could handle if it comes up.

calebrob6 (approved):

adamjstewart:
I'll try to take a closer look at this later this week. I definitely want to get this into the 0.5.0 release in August.

tests/datasets/test_eudem.py
```diff
@@ -33,21 +33,23 @@ def test_getitem(self, dataset: EUDEM) -> None:
         assert isinstance(x["mask"], torch.Tensor)
 
     def test_extracted_already(self, dataset: EUDEM) -> None:
-        zipfile = os.path.join(dataset.root, "eu_dem_v11_E30N10.zip")
-        shutil.unpack_archive(zipfile, dataset.root, "zip")
-        EUDEM(dataset.root)
+        zipfile = dataset.list_files(filename_glob="eu_dem_v11_E30N10.zip")[0]
+        outdir = os.path.abspath(os.path.join(zipfile, os.pardir))
+        # TODO: This changes the behaviour from unpacking in root to same dir as archive
+        shutil.unpack_archive(zipfile, outdir, "zip")
```
adamjstewart, line 39: Why is this hack necessary? Can't we do:
```suggestion
        zipfile = os.path.join(dataset.paths[0], "eu_dem_v11_E30N10.zip")
        shutil.unpack_archive(zipfile, dataset.paths[0], "zip")
        EUDEM(dataset.root)
```

tests/datasets/test_geo.py
adamjstewart, line 181: Can we do a pytest.parametrize and test all of the following:

* single dir (str)
* multiple dirs (list[str])
* single file (str)
* multiple files (list[str])
* combination of dirs and files (list[str])

Should also rename the test to something more generic. We should either move this to TestGeoDataset or also add something similar to TestVectorDataset.

torchgeo/datasets/geo.py
```diff
@@ -329,7 +329,7 @@ def dtype(self) -> torch.dtype:
 
     def __init__(
         self,
-        root: str = "data",
+        paths: Union[str, list[str]] = "data",
```
adamjstewart, line 332: Can we use `Iterable` instead of `list`? I don't want to tell users they can't use a tuple if they want to. We may also want to support generators someday for online learning. I'm trying to get better at making these types as broad as possible.

torchgeo/datasets/geo.py
```diff
@@ -339,7 +339,8 @@ def __init__(
         """Initialize a new Dataset instance.
 
         Args:
-            root: root directory where dataset can be found
+            paths: root directory or list of absolute filepaths where
+                dataset can be found
```
adamjstewart, line 343: Can we change this to something like:

> paths: one or more directories or files to load

This makes it clear that:

* A list of directories is supported
* A single file is supported

The only thing I don't love is idk how to express that some datasets can be auto-downloaded and this is where they will be downloaded to, not where they are found.

This will break support for glob chars in directories, but I'm not sure if anyone was using that anyway.

torchgeo/datasets/geo.py
```diff
@@ -396,7 +396,9 @@ def __init__(
                     i += 1
 
         if i == 0:
-            msg = f"No {self.__class__.__name__} data was found in `root='{self.root}'`"
+            msg = (
+                f"No {self.__class__.__name__} data was found in `root='{self.paths}'`"
```
adamjstewart, line 400:
```suggestion
                f"No {self.__class__.__name__} data was found in `paths={self.paths!r}`"
```
This will look better if paths is a list

torchgeo/datasets/geo.py
```diff
@@ -418,6 +420,61 @@ def __init__(
         self._crs = cast(CRS, crs)
         self._res = cast(float, res)
 
+    def list_files(self, filename_glob: Optional[str] = None) -> list[str]:
```
adamjstewart, line 423: Can we remove the `filename_glob` parameter for now? This is also being discussed in #1524 and I don't think we've settled on how to handle this yet.

torchgeo/datasets/geo.py
```diff
@@ -418,6 +420,61 @@ def __init__(
...
+        """Get list of files matching filename_glob.
+
+        Args:
+            filename_glob: Defaults to self.filename_glob
+        """
+        # self.paths can be single string (or pathlike) representing root directory.
```
adamjstewart, line 429: Can't be a pathLike yet, let's leave support for that for a future PR

torchgeo/datasets/geo.py
```diff
@@ -418,6 +420,61 @@ def __init__(
...
+        # Make iterable
+        if isinstance(self.paths, str) or not isinstance(self.paths, Iterable):
+            paths = [self.paths]
+        else:
+            paths = cast(list[str], self.paths)
```
adamjstewart, line 434: `cast` shouldn't usually be necessary unless there's something wrong with the type hints. Can we simplify this?
```suggestion
        if isinstance(self.paths, str):
            paths = [self.paths]
        else:
            paths = self.paths
```
Thought about moving it to `__init__` but I'm not sure if I want to do that.

torchgeo/datasets/geo.py
```diff
@@ -418,6 +420,61 @@ def __init__(
...
+                    filepaths.add(dir_or_file)
+                else:
+                    pathname = os.path.join(dir_or_file, "**", filename_glob)
+                    filepaths |= set(glob.iglob(pathname, recursive=True))
+            else:
+                filepaths |= self.handle_nonlocal_path(dir_or_file)
+
+        return list(filepaths)
+
+    def handle_nonlocal_path(self, path: str) -> set[str]:
```
adamjstewart, line 452: I don't want to do this just yet, can we save it for your other PR? I'm actually leaning towards using classmethods like `from_vsi` or `from_s3`, or just telling users to pass in a list of files if they want to use a remote server that can't be globbed. So for now, use something like:
```python
for path in paths:
    if os.path.isdir(path):
        # glob
    else:
        # add
```

torchgeo/datasets/geo.py
adamjstewart: Also wondering if we should change this to:
```python
@propery
def files(self) -> list[str]:
    ...
```
I think `dataset.files` feels cleaner than `dataset.list_files()`

torchgeo/datasets/agb_live_woody_density.py
```diff
@@ -112,15 +111,19 @@ def _verify(self) -> None:
...
-        download_url(self.url, self.root, self.base_filename)
+        download_url(self.url, self.paths, self.base_filename)
 
-        with open(os.path.join(self.root, self.base_filename)) as f:
+        base_file = self.list_files(filename_glob=self.base_filename)[0]
+        with open(base_file) as f:
             content = json.load(f)
 
+        # TODO: This changes the behaviour from
+        #  unpacking in root to same dir as archive
```
adamjstewart, line 121: It's not clear to me exactly how all of this works. If `self.paths` is a list of files, how will `download_url` work? Do we need to add an assert statement somewhere that force paths to be a single dir? Or should we just use os.path.basename if it's a file?

torchgeo/datasets/landcoverai.py
```diff
@@ -86,6 +89,10 @@ def __init__(
             RuntimeError: if ``download=False`` and data is not found, or checksums
                 don't match
         """
+        if isinstance(root, list):
+            # TODO: Workaround until list are fully implemented
```
adamjstewart, line 93: I'm fine with whatever workarounds we need for these special cases for now.

adamjstewart (changes requested):
Apologies for taking so long to review this, I graduated and moved to Germany so I've been very preoccupied.

I still really want to sneak this into 0.5.0, which I still want to finish by tomorrow. I know this is very last minute, so if you need me to I'm happy to take over at any point to finish up this PR. Depending on which time zone you're in this may work very well.

Last request: can we do the same thing for VectorDataset? The code will be basically identical, and we only have 2 of them, so shouldn't take long.

tests/datasets/test_eudem.py
adriantre: Yes, that works if the zip archive is (forever) in `dataset.paths[0]` and not in subdirectory.

tests/datasets/test_geo.py
adriantre: Good idea. If you want to take a crack at testing this and codecov, I can take a look at the VectorDataset support?

torchgeo/datasets/geo.py
adriantre: I like the idea of a property, but it should still re-create the list every time it is called, as the file structure may change, do you agree?

tests/datasets/test_agb_live_woody_density.py
```diff
@@ -54,12 +54,12 @@ def test_getitem(self, dataset: AbovegroundLiveWoodyBiomassDensity) -> None:
 
     def test_no_dataset(self) -> None:
         with pytest.raises(RuntimeError, match="Dataset not found in."):
-            AbovegroundLiveWoodyBiomassDensity(root="/test")
+            AbovegroundLiveWoodyBiomassDensity(paths="/test")
```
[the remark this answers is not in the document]
adriantre: Should be fixed now

torchgeo/datasets/geo.py
adriantre: Fixed

torchgeo/datasets/geo.py
adriantre: I included the possibility to override filename_glob only in order to easier support the code that [looks for zip-files](https://github.com/microsoft/torchgeo/blob/a5d3d6ef9b26690bb632e76ab123cef9466139c9/tests/datasets/test_eudem.py#L36). If we can still solve that, I am happy to remove it.

adamjstewart: I think we can remove that and use what we previously used to find zip files, just using glob.

I'm fine with re-creating the list every time. If it becomes slow or we need to do it multiple times we can cache it.

torchgeo/datasets/geo.py
adriantre: If the users want to use the `dataset.paths` after init, I think it is tidy that the type is not changed. It may lead to runtime errors (caused by the user), but that would be the case either way.

torchgeo/datasets/geo.py
adriantre: I'm ok with that. Now that `list_files` (or what it will be called) is implemented, it is much easier for users (me) to override this part of the code in order to support vsi.

torchgeo/datasets/landcoverai.py
adriantre: All right. Can we merge this and start a new PR for VectorDataset?

adamjstewart: I'm fine with doing VectorDataset in a separate PR. Just need to get all tests passing on this PR. Let me know if/when you want me to take over.

tests/datasets/test_cms_mangrove_canopy.py
```diff
@@ -65,7 +65,9 @@ def test_corrupted(self, tmp_path: Path) -> None:
         ) as f:
             f.write("bad")
         with pytest.raises(RuntimeError, match="Dataset found, but corrupted."):
-            CMSGlobalMangroveCanopy(root=str(tmp_path), country="Angola", checksum=True)
+            CMSGlobalMangroveCanopy(
+                paths=str(tmp_path), country="Angola", checksum=True
```
adamjstewart, line 69:
```suggestion
                str(tmp_path), country="Angola", checksum=True
```

torchgeo/datasets/geo.py
```diff
@@ -350,19 +350,18 @@ def __init__(
             cache: if True, cache file handle to speed up repeated sampling
 
         Raises:
-            FileNotFoundError: if no files are found in ``root``
+            FileNotFoundError: if no files are found in ``paths``
```
adamjstewart, line 353: Need to add something like:
```rst
.. versionchanged:: 0.5
   *root* was renamed to *paths*
```
Same for other datasets

torchgeo/datasets/geo.py
```diff
@@ -418,6 +420,60 @@ def __init__(
         self._crs = cast(CRS, crs)
         self._res = cast(float, res)
 
+    def list_files(self, filename_glob: Optional[str] = None) -> list[str]:
+        """Get list of files matching filename_glob.
+
+        Args:
+            filename_glob: Defaults to self.filename_glob
```
adamjstewart, line 427: Need to add:
```rst
.. versionadded:: 0.5
```
I think we can actually move this method to `GeoDataset` and reuse it in `VectorDataset` instead of duplicating it.

torchgeo/datasets/agb_live_woody_density.py
adriantre: Yes, this is not intuitive. It works only because we know that the code instantiates this class with a str representing root. Why mypy does not complain that the type may not be as expected is unknown to me, any idea? 

This should annoy mypy:
```python
download_url(self.url, self.paths, self.base_filename)
```

I haven't been able come up with an elegant way of using `self.paths` as a root.

torchgeo/datasets/geo.py
adriantre: If i'm adding support for VectorDataset in a separate PR I will save it for that PR.

torchgeo/datasets/geo.py
adamjstewart: I wish. I don't think that's possible.

torchgeo/datasets/geo.py
adriantre: Fixed

tests/conf/l7irish.yaml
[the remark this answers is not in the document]
adriantre: Now this is changed for GeoDataModules that instantiate RasterDatasets. If I did not miss any 😅

torchgeo/datasets/geo.py
adriantre: Fixed

torchgeo/datasets/geo.py
adriantre: Should be fixed

adamjstewart:
> How to find zip-files if we don't want to allow to override in filename_glob

Same way we used to, with a call to glob.glob

torchgeo/datasets/agb_live_woody_density.py
```diff
@@ -80,14 +80,16 @@ def __init__(
             cache: if True, cache file handle to speed up repeated sampling
 
         Raises:
-            FileNotFoundError: if no files are found in ``root``
+            FileNotFoundError: if no files are found in ``paths``
```
adamjstewart, line 83: I think you need a blank line between Raises and versionchanged

adamjstewart:
Too tired to finish this tonight, will work on this tomorrow.

torchgeo/datasets/agb_live_woody_density.py
adriantre: Fixed

torchgeo/datasets/cdl.py
```diff
@@ -292,8 +294,7 @@ def _verify(self) -> None:
         exists = []
         for year in self.years:
             filename_year = self.filename_glob.replace("*", str(year))
-            pathname = os.path.join(self.root, "**", filename_year)
-            for fname in glob.iglob(pathname, recursive=True):
+            for fname in self.list_files(filename_glob=filename_year):
```
adriantre, line 297: @adamjstewart This is also a use case where overriding the `filename_glob` is proposed.

adamjstewart: In this case `self.list_files` isn't significantly better than `glob.iglob`, you're only saving 1 line of code and the result is that it can't be a property anymore.

tests/datasets/test_eudem.py
adamjstewart: It will forever be in this location, at least for our unit tests.

torchgeo/datasets/agb_live_woody_density.py
```diff
@@ -80,14 +80,17 @@ def __init__(
             cache: if True, cache file handle to speed up repeated sampling
```
adriantre, line 80: @adamjstewart can you take a look at the docstring for `download? just above this line?

`download: if True, download dataset and store it in the root directory`

It may be hard to know what that root is

adamjstewart: I think the majority of users will still use a single string in `paths`, and the docs do say that `paths` is one or more root directories, so hopefully they can figure it out. If users report confusion we can make this more clear or add a better error message in a future patch release.

adamjstewart (approved):
I think this is ready to merge now!

At the moment, we don't support passing a list if you also want to download/extract a dataset. I think this limitation makes sense, as people won't have an easy way to get a list of files if they don't already have them extracted anyway. But if someone asks for this feature, we can think about how to do it in a future patch release.

Thanks for a great first contribution, I think this is worth highlighting in the release notes as it unlocks a lot of potential for remote files or file filtering.

adriantre:
Looks good to me! Good job! Hopefully not my last contrib! :)

PR #1446, opened 2023-06-24
Move project configuration from setup.cfg to pyproject.toml

adamjstewart:
Setuptools 61+ adds support for [PEP 621](https://peps.python.org/pep-0621/), which describes a common metadata structure for package configuration. This PR moves our old setup.cfg configuration to pyproject.toml.

I considered moving from setuptools to hatchling but still haven't found a compelling reason other than "everyone else is doing it". So as to avoid continuously jumping from fad to fad, I think we should stick with setuptools until PyPA provides further guidance on how to choose a build backend. Luckily, since most build backends support PEP 621, it will be much easier to make the switch now.

I was hoping we could reduce the number of config files, but Codecov/ReadTheDocs don't support pyproject.toml and the maintainer of flake8/pyupgrade/pre-commit actively refuses to support pyproject.toml.


pyproject.toml
```diff
@@ -1,10 +1,155 @@
...
+license = {file = "LICENSE"}
+authors = [
+    {name = "Adam J. Stewart", email = "ajstewart426@gmail.com"},
+]
+maintainers = [
+    {name = "Adam J. Stewart", email = "ajstewart426@gmail.com"},
+    {name = "Caleb Robinson", email = "Caleb.Robinson@microsoft.com"},
+    {name = "Isaac Corley"},
+    {name = "Nils Lehmann"},
+    {name = "Ashwin Nair"},
```
adamjstewart, line 23: @isaaccorley @nilsleh @ashnair1 I didn't add your emails because they aren't public on GitHub but let me know if you want me to add them. These names/emails are listed on the [PyPI project page](https://pypi.org/project/torchgeo/).

pyproject.toml
```diff
@@ -1,10 +1,155 @@
...
+    # tensorboard 2.9.1+ required by lightning
+    "tensorboard>=2.9.1,<3",
+]
+all = [
+    "torchgeo[datasets,docs,style,tests]",
+]
+
+[project.urls]
+Homepage = "https://github.com/microsoft/torchgeo"
+Documentation = "https://torchgeo.readthedocs.io"
```
adamjstewart, line 151: This is new, we previously only listed a homepage

pyproject.toml
nilsleh: n.lehmann@tum.de

ashnair1: ashnair0007@gmail.com

isaaccorley: isaac.corley@my.utsa.edu

ashnair1 (approved):

PR #443, opened 2022-03-01
Add NAIPCluster dataset and datamodule

calebrob6:
Dataset of NAIP imagery sampled random from the NAIP archive and masks generated by a KMeans clustering of pixels on-the-fly.

image

image

PR #1450, opened 2023-06-28
Feature extracted USAVars Dataset

nilsleh:
This PR adds a feature extracted version of USAVars. In the MOSAIKs paper, they show that RCF extracted features with RidgeRegression can outperform fully trained CNNs. In their code base they also have this feature extraced dataset version available. In order to not have to extract the same fixed RCF features in each epoch, this PR makes a feature extraced dataset version available to reduce the computational cost of running that experiment setting.


docs/api/datamodules.rst
```diff
@@ -151,6 +151,11 @@ USAVars
 
 .. autoclass:: USAVarsDataModule
```
adamjstewart, line 152: Let's just put it under this section, it's basically the same dataset

docs/api/datasets.rst
```diff
@@ -351,6 +351,11 @@ USAVars
 
 .. autoclass:: USAVars
```
adamjstewart, line 352: Same

docs/api/non_geo_datasets.csv
```diff
@@ -36,6 +36,7 @@ Dataset,Task,Source,# Samples,# Classes,Size (px),Resolution (m),Bands
 `Tropical Cyclone`_,R,GOES 8--16,"108,110",-,256x256,4K--8K,MSI
 `UC Merced`_,C,USGS National Map,"2,100",21,256x256,0.3,RGB
 `USAVars`_,R,NAIP Aerial,100K,-,-,4,"RGB, NIR"
+`USAVarsFeaturesExtracted`_,R,NAIP Aerial,100K,-,-,4,"RGB, NIR"
```
adamjstewart, line 39: Same here, no need for a new row, the values are the same and the link will link to the above sections

adamjstewart (commented):
Don't know much about the details but can dig a little deeper after I dig myself out of this paperwork hole.

calebrob6:
Hey @nilsleh -- this is just running normal USAVars through an RCF and ResNet and exporting the results as a dataset?

calebrob6:
I'm a bit hesitant as:
1. there are many parameters that one might want to play with to go from NAIP images in USAVars to the embedded versions (normalization, size of RCF kernels, number of RCF kernels, etc.) and this is simply choosing one such combination
2. caching the results of an RCF or ResNet18 embedding step isn't too difficult, so I don't think this saves much work

adamjstewart:
This smells like #443

nilsleh:
I don't have a strong opinion about it. I agree the embedding step is not difficult, but I thought I would just add the dataloading and module for it to be useable with torchgeo.

PR #1456, opened 2023-06-30
Added pretraining config files for MoCo and SimCLR 

AABNassim:
on oli_tirs_toa with a ResNet-50 and ViT-S backbone


AABNassim:

adamjstewart (approved):
Thanks!

PR #1466, opened 2023-07-06
adding seasonet dataset + tests + doc 

dkosm:
This PR adds the SeasoNet Seasonal Land Cover Segmentation Dataset. 

Huge thanks to @briktor for his work on the dataset and the torchgeo version !!

The SeasoNet [dataset](https://doi.org/10.5281/zenodo.5850306) consists of 1,759,830 multi-spectral Sentinel-2 image patches, taken from 519,547 unique locations, covering the whole surface area of Germany. Annotations are provided in the form of pixel-level land cover and land usage segmentation masks from the German land cover model LBM-DE2018 with land cover classes based on the CORINE Land Cover database (CLC) 2018. The set is split into two overlapping grids, consisting of roughly 880,000 samples each, which are shifted by half the patch size in both dimensions. The images in each of the both grids themselves do not overlap.

Example Plot: 
example_plot

Dataset format:

* images are 16-bit GeoTiffs, split into seperate files based on resolution
* images include 12 spectral bands with 10, 20 and 60 m per pixel resolutions
* masks are single-channel 8-bit GeoTiffs

Paper: [here](https://arxiv.org/abs/2207.09507)


docs/api/datasets.rst
```diff
@@ -297,6 +297,11 @@ Seasonal Contrast
 
 .. autoclass:: SeasonalContrastS2
 
+SeasoNet
+^^^^^^^^^^^^^^^^^
```
adamjstewart, line 301:
```suggestion
SeasoNet
^^^^^^^^
```

tests/datasets/test_seasonet.py
```diff
@@ -0,0 +1,210 @@
...
+from pathlib import Path
+from typing import Union
+
+import matplotlib.pyplot as plt
+import pytest
+import requests
+import torch
+import torch.nn as nn
+from _pytest.fixtures import SubRequest
+from _pytest.monkeypatch import MonkeyPatch
```
adamjstewart, line 16:
```suggestion
from pytest import MonkeyPatch
```
It's part of the public API now

tests/datasets/test_seasonet.py
```diff
@@ -0,0 +1,210 @@
...
+from torch.utils.data import ConcatDataset
+
+import torchgeo.datasets.utils
+from torchgeo.datasets import SeasoNet
+
+
+class MockResponse:
```
adamjstewart, line 23: Why is this class necessary? We haven't used anything like this in other datasets

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,456 @@
...
+from matplotlib.colors import ListedColormap
+from matplotlib.figure import Figure
+from rasterio.enums import Resampling
+from torch import Tensor
+
+from .geo import NonGeoDataset
+from .utils import download_url, extract_archive, percentile_normalization
+
+
+class SeasoNet(NonGeoDataset):  # TODO: Docs
```
adamjstewart, line 26: Is this still a TODO?

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,456 @@
...
+        "Water bodies",
+        "Coastal lagoons",
+        "Estuaries",
+        "Sea and ocean",
+    ]
+    all_seasons = {"Spring", "Summer", "Fall", "Winter", "Snow"}
+    all_bands = ("10m_RGB", "10m_IR", "20m", "60m")
+    band_nums = {"10m_RGB": 3, "10m_IR": 1, "20m": 6, "60m": 2}
+    splits = ["train", "val", "test"]
+    colormap = [
```
adamjstewart, line 144: Let's name this `cmap` and use the same dict structure as CDL and NLCD.

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,456 @@
...
+import os
+import random
+from typing import Callable, Collection, Iterable, Optional
+
+import matplotlib.patches as mpatches
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+import rasterio
+import requests
```
adamjstewart, line 15: We don't yet have a direct dependency on requests, which is why mypy is failing. See below, it might be possible to remove this.

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,456 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""SeasoNet dataset."""
+
+import os
+import random
+from typing import Callable, Collection, Iterable, Optional
```
adamjstewart, line 8: Most of these can be imported from `collections.abc`, which is why pyupgrade tests are failing

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,456 @@
...
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+        """
+        assert split in self.splits
+        for season in seasons:
+            assert season in self.all_seasons
+        for band in bands:
+            assert band in self.all_bands
+        for grid in grids:
+            assert grid in [1, 2]
```
adamjstewart, line 216: Would it be simpler to check using sets instead of for loops? For example:
```python
```

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,456 @@
...
+    ) -> None:
+        """Initialize a new SeasoNet dataset instance.
+
+        Args:
+            root: root directory where dataset can be found
+            split: one of "train", "val" or "test"
+            seasons: list of seasons to load
+            bands: list of bands to load
+            grids: which of the overlapping grids to load
+            concat_seasons: number of seasonal images to return per sample.
```
adamjstewart, line 201: The SeCo and SSL4EO datasets also have a notion of seasons. In those datasets, the `seasons` parameter controls the number of seasons to sample from. I wonder if we should try to be more consistent with parameter names or not.

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,456 @@
...
+                if 1, each seasonal image is returned as its own sample,
+                otherwise seasonal images are randomly picked from the seasons
+                specified in `seasons` and returned as stacked tensors
```
adamjstewart, line 204: Single backticks is italic, did you mean double backticks?

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,456 @@
...
+"""SeasoNet dataset."""
+
+import os
+import random
+from typing import Callable, Collection, Iterable, Optional
+
+import matplotlib.patches as mpatches
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
```
adamjstewart, line 13: We don't (yet) have a required dep on pandas, this should be imported in the dataset class. See gbif and other datasets for examples of how to do this and test this.

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,456 @@
...
+                else missing[0]
+            )
+            raise RuntimeError(
+                f"{missing_str} not found in `root={self.root}` and `download=False`, "
+                "either specify a different `root` directory or use `download=True`"
+                " to automatically download the dataset."
+            )
+
+        # Download missing files
+        record_info = requests.get(self.url).json()["files"]
```
adamjstewart, line 365: Is this necessary? None of our other datasets directly use requests to download, they rely on download_url.

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,456 @@
...
+        )
+
+        if self.concat_seasons > 1:
+            self.files = csv.groupby(["Latitude", "Longitude"])
+            self.files = self.files["Path"].agg(sum)
+            self.files = self.files[
+                self.files.apply(lambda d: len(d) >= self.concat_seasons)
+            ]
+        else:
+            self.files = csv["Path"]
```
adamjstewart, line 254: Might be useful to add comments to `__init__`, at first glance I have no idea what all of this pandas logic is doing.

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,456 @@
...
+
+    def __getitem__(self, index: int) -> dict[str, Tensor]:
+        """Return an index within the dataset.
+
+        Args:
+            index: index to return
+
+        Returns:
+            sample at that index containing the image with shape SCxHxW,
+            where ``S = self.concat_seasons``, and the mask
```
adamjstewart, line 264: What about the mask?

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,456 @@
...
+            for band in self.bands:
+                with rasterio.open(f"{path}_{band}.tif") as f:
+                    array = f.read(
+                        out_shape=[f.count] + list(self.image_size),
+                        out_dtype="int32",
+                        resampling=Resampling.bilinear,
+                    )
+                image = torch.from_numpy(array).float()
+                c = img_idx * self.channels + bnd_idx
+                tensor[c : c + image.shape[0]] = image
```
adamjstewart, line 307: I usually just append to a list and concatenate at the end, not sure if that would be simpler

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,456 @@
...
+    27. Intertidal flats
+    28. Water courses
+    29. Water bodies
+    30. Coastal lagoons
+    31. Estuaries
+    32. Sea and ocean
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://doi.org/10.1109/IGARSS46834.2022.9884079
```
adamjstewart, line 83: Can you add:
```rst
.. versionadded:: 0.5
```

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,456 @@
...
+        if not missing:
+            return
+
+        # Check if the user requested to download the dataset
+        if not self.download:
+            missing_str = (
+                f"{', '.join(missing[:-1])} and {missing[-1]}"
+                if len(missing) > 1
+                else missing[0]
+            )
```
adamjstewart, line 357: Not sure if the error message needs to be this detailed

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,456 @@
...
+                    md5=file_info["checksum"].split(":")[1] if self.checksum else None,
+                )
+                if file_info["type"] == "zip":
+                    extractable.append(file_path)
+
+        # Extract downloaded files
+        for file_path in extractable:
+            extract_archive(file_path)
+
+    def plot(
```
adamjstewart, line 383: Can you add an example plot to the PR description so we can check if plotting works correctly?

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,456 @@
...
+        fig.subplots_adjust(wspace=0.05)
+        for ax, index in enumerate(rgb_indices):
+            image = images[index : index + 3].permute(1, 2, 0).numpy()
+            image = percentile_normalization(image)
+            axs[ax].imshow(image)
+            axs[ax].axis("off")
+            if show_titles:
+                axs[ax].set_title(f"Image {ax+1}")
+
+        axs[ax + 1].imshow(mask, vmin=0, vmax=32, cmap=self.cmap)
```
adamjstewart, line 430: We also found that adding `interpolation="none"` to imshow when plotting masks is important, otherwise classes will bleed over when you zoom in.

adamjstewart (commented):
Thanks for the PR, excited to add your dataset to TorchGeo!

dkosm:
Thanks for the fast reply! We will look into the necessary changes asap!

torchgeo/datasets/seasonet.py
briktor: download_url wouldn't work in this case, because no files are being downloaded here. This is just a call to the Zenodo API to get the correct file URLs and MD5s and is also the intended way to download datasets from Zenodo through scripts. This is also what the MockResponse class in our tests is for, as it mocks the response to this API call.

Technically it would be possible to hard-code all URLs and MD5s into the dataset but in case anything ever changes either on our side regarding the files themselves or on Zenodos side regarding their location, this API call would still provide the correct information.

torchgeo/datasets/seasonet.py
briktor: That may be simpler but afaik this way should be a bit better performance-wise

torchgeo/datasets/seasonet.py
briktor: What `seasons` does in both of those datasets is basically what `concat_seasons` does in our dataset. I do agree with trying to be consistent but I also think it would be a bit counter-intuitive for the `seasons` parameter not to be the one that is used to choose the actual seasons. Though it would require more changes, I could rather see something like `num_seasons` working in all cases.

torchgeo/datasets/seasonet.py
briktor: True, it doesn't have to be. At least the extra logic just to add an "and". I do think it is helpful to print the actual list of missing files though, as the dataset requires multiple very large files to be downloaded and just saying "dataset missing", even when it's just a single file, may be a bit misleading.

torchgeo/datasets/seasonet.py
adamjstewart: How many URLs/MD5s are we talking? The whole point of MD5s is to ensure that the files that get downloaded are the ones expected. For example, if someone hacked your Zenodo account and uploaded a virus in place of the dataset, a change in MD5 would alert users to the issue. I don't think our users are hugely concerned about security, but MD5s are still useful to ensure reproducibility.

adamjstewart: That's basically the approach our current datasets use. The URL and MD5 are hard-coded into the dataset, and we optionally verify the checksum after download. Let's do the same for this dataset for now. If we're concerned that the URL might change frequently, we can do something fancier later. And the MD5 changing is less of a big deal since users can just use `checksum=False` if they don't care, which is the default.

briktor: Afaik the approach most datasets use currently does not include an MD5 check before downloading, which I still think would be nice to have, since it would inform the user that something is wrong before having to download a 60GB file and computing the MD5 on their own machine. But just to make sure, since this is all concerning a very simple API request, just using urllib instead of requests should work, right? Since that is what `download_url` also uses.

adamjstewart: I'm not sure if most websites (other than Zenodo) will let you verify the checksum before you even start the download. For now, it's best to keep things the same as every other dataset. If you want to change how all dataset downloads work, that should be left up to a follow-up PR. Any code unique to this PR requires a lot more careful thought and vetting, whereas using the same code as every other dataset makes it easier for me to tell it will work consistently.

calebrob6:
Thanks @dkosm! Please let us know if you questions about the changes :)

torchgeo/datasets/seasonet.py
dkosm: Done

torchgeo/datasets/seasonet.py
dkosm: Done

torchgeo/datasets/seasonet.py
dkosm: Done

torchgeo/datasets/seasonet.py
dkosm: Done

torchgeo/datasets/seasonet.py
dkosm: Resolved

tests/datasets/test_seasonet.py
dkosm: See other comments. Handled in one of the commits.

tests/datasets/test_seasonet.py
```diff
@@ -0,0 +1,217 @@
...
+        monkeypatch.setitem(
+            SeasoNet.metadata[5],
+            "url",
+            os.path.join("tests", "data", "seasonet", "splits.zip"),
+        )
+        monkeypatch.setitem(
+            SeasoNet.metadata[6],
+            "url",
+            os.path.join("tests", "data", "seasonet", "meta.csv"),
+        )
```
adamjstewart, line 101: Could use a list and a for-loop so there isn't so much code duplication.

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,495 @@
...
+       * `pandas <https://pypi.org/project/pandas/>`_ to load CSV files
+
+    .. versionadded:: 0.5
+    """
+
+    metadata = [
+        {
+            "name": "spring",
+            "ext": ".zip",
+            "url": "https://zenodo.org/api/files/e2288446-9ee8-4b2e-ae76-cd80366a40e1/spring.zip",  # noqa: E501
```
adamjstewart, line 96: Personally I would save a base url like:
```python
url = "https://zenodo.org/api/files/e2288446-9ee8-4b2e-ae76-cd80366a40e1/"
```
and then you can append only when you need the full URL. Then you don't have so much code duplication. Not sure if you really need `ext` either, could just infer that from the filename.

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,495 @@
...
+        31: (166, 255, 230, 255),
+        32: (230, 242, 255, 255),
+    }
+    image_size = (120, 120)
+
+    def __init__(
+        self,
+        root: str = "data",
+        split: str = "train",
+        seasons: Collection[str] = all_seasons,
```
adamjstewart, line 216: How is Collection different from Iterable? How did you decide which to use?

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,495 @@
...
+
+        Args:
+            index: index to return
+
+        Returns:
+            the target mask
+        """
+        path = self.files.iloc[index][0]
+        with rasterio.open(f"{path}_labels.tif") as f:
+            array = f.read() - 1
```
adamjstewart, line 366: Maybe add a comment explaining why - 1?

torchgeo/datasets/seasonet.py
```diff
@@ -0,0 +1,495 @@
...
+        tensor = torch.from_numpy(array).squeeze().long()
+        return tensor
+
+    def _verify(self) -> None:
+        """Verify the integrity of the dataset.
+
+        Raises:
+            RuntimeError: if ``download=False`` but dataset is missing or checksum fails
+            ImportError: if ``download=True`` but requests is not installed
```
adamjstewart, line 375: requests is no longer used, correct?

adamjstewart (dismissed):
This looks fantastic!

Left a couple of optional style suggestions.

adamjstewart (approved):

adamjstewart:
Thanks for the contribution, hopefully the first of many!

Apologies it took so long to get the PR merged. In the meantime I moved to Germany! Starting a postdoc at TUM, maybe we'll see each other around someday.

tests/data/seasonet/data.py
```diff
@@ -0,0 +1,141 @@
...
+from rasterio.transform import Affine
+
+np.random.seed(0)
+
+meta = {
+    "driver": "GTiff",
+    "nodata": None,
+    "crs": CRS.from_epsg(32632),
+    "transform": Affine(10.0, 0.0, 664800.0, 0.0, -10.0, 5342400.0),
+    "compress": "zstd",
```
adamjstewart, line 22: Just want to confirm that the original SeasoNet files are also compressed with ZSTD

briktor: Yes, they are

PR #1482, opened 2023-07-17
SSL4EO-L Weights

nilsleh:
This PR adds the SSL4EO-L Weights as Enums


nilsleh:
@adamjstewart I just added a couple to see if the formatting and naming etc is okay, before adding the rest.

adamjstewart:
I'm thinking we should have a single repo on HF for all Landsat weights. What do you think?

adamjstewart:
Trying to decide if we should rename our model repo to `ssl4eo_l` or our dataset repo to `ssl4eo_landsat`. What do you think?

nilsleh:
I feel like ssl4eo_landsat is a lot more clear and worth the extra characters, I didn't think about it when creating the HF repo, just thought it would be the clearest description

torchgeo/models/resnet.py
```diff
@@ -52,6 +52,13 @@
     data_keys=["image"],
 )
 
+# https://github.com/microsoft/torchgeo/blob/8b53304d42c269f9001cb4e861a126dc4b462606/torchgeo/datamodules/ssl4eo_benchmark.py#L43 # noqa: E501
+_ssl4eo_l_transforms = AugmentationSequential(
+    K.Normalize(mean=torch.tensor(0), std=torch.tensor(255)),
+    K.RandomResizedCrop((224, 224), scale=(0.6, 1.0)),
```
adamjstewart, line 58: I think center crop would be better here. The idea behind these transforms is if you start with a pre-trained model, these are the transforms needed to ensure that the input images match the format expected by the model at **prediction** time, not at **training** time. At least, that's how I think of them, I could be wrong.

torchgeo/models/resnet.py
```diff
@@ -52,6 +52,13 @@
...
+    data_keys=["image"],
```
adamjstewart, line 59: I'm never sure if we're supposed to append all possible data keys here or only the ones we're actually using. Like if we use this with a mask, we want the mask to be cropped too.

nilsleh: Are we defining these transforms just as an "information" guideline or are they somewhere automatically integrated and used? I wouldn't think so since the torchgeo datamodules defined the augmentations. So if people have to write their own datamodule anyway then they will have to adapt the augmentations for their use case as well. Or how is this `Weights` attribute supposed to be used?

adamjstewart: AFAIK, they aren't automatically used by anything, but users can programmatically use them in their code. The idea is that these are reference transforms that can be used with the weights, but users are free to adapt them. Note that not all users will be using Lightning, some may directly import the weights and use the transforms in their existing pipelines.

docs/api/resnet_pretrained_weights.csv
```diff
@@ -2,9 +2,29 @@ Weight,Channels,Source,Citation,BigEarthNet,EuroSAT,So2Sat,OSCD
 ResNet18_Weights.SENTINEL2_ALL_MOCO,13,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,,,,
 ResNet18_Weights.SENTINEL2_RGB_MOCO, 3,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,,,,
 ResNet18_Weights.SENTINEL2_RGB_SECO, 3,`link <https://github.com/ServiceNow/seasonal-contrast>`__,`link <https://arxiv.org/abs/2103.16607>`__,87.27,93.14,,46.94
+ResNet18_Weights.LANDSAT_ETM_SR_MOCO,6,`link <https://github.com/microsoft/torchgeo>`__,`link <https://arxiv.org/abs/2306.09424>`__,,,,
```
adamjstewart, line 5: It doesn't really make sense to add all ResNet metrics to a single table. We can't report model performance of Landsat weights on BigEarthNet/EuroSAT/So2Sat/OSCD because those are completely different datasets. I think we want a separate table for each sensor instead of each model. In this case, that would mean one table for Sentinel 1, one for Sentinel 2, and one for all of Landsat. We can't combine S1 and S2 because the datasets used to evaluate them are different. But we can combine L5–L9 because we have a single SSL4EO-L CDL/NLCD dataset that we can directly compare across.

docs/api/resnet_pretrained_weights.csv
```diff
@@ -2,9 +2,29 @@ Weight,Channels,Source,Citation,BigEarthNet,EuroSAT,So2Sat,OSCD
...
+ResNet18_Weights.LANDSAT_ETM_TOA_MOCO,9,`link <https://github.com/microsoft/torchgeo>`__,`link <https://arxiv.org/abs/2306.09424>`__,,,,
+ResNet18_Weights.LANDSAT_ETM_TOA_SIMCLR,9,`link <https://github.com/microsoft/torchgeo>`__,`link <https://arxiv.org/abs/2306.09424>`__,,,,
+ResNet18_Weights.LANDSAT_OLI_SR_MOCO,7,`link <https://github.com/microsoft/torchgeo>`__,`link <https://arxiv.org/abs/2306.09424>`__,,,,
+ResNet18_Weights.LANDSAT_OLI_SR_SIMCLR,7,`link <https://github.com/microsoft/torchgeo>`__,`link <https://arxiv.org/abs/2306.09424>`__,,,,
+ResNet18_Weights.LANDSAT_OLI_TIRS_TOA_MOCO,11,`link <https://github.com/microsoft/torchgeo>`__,`link <https://arxiv.org/abs/2306.09424>`__,,,,
+ResNet18_Weights.LANDSAT_OLI_TIRS_TOA_SIMCLR,11,`link <https://github.com/microsoft/torchgeo>`__,`link <https://arxiv.org/abs/2306.09424>`__,,,,
+ResNet18_Weights.LANDSAT_TM_TOA_MOCO,5,`link <https://github.com/microsoft/torchgeo>`__,`link <https://arxiv.org/abs/2306.09424>`__,,,,
+ResNet18_Weights.LANDSAT_TM_TOA_SIMCLR,5,`link <https://github.com/microsoft/torchgeo>`__,`link <https://arxiv.org/abs/2306.09424>`__,,,,
```
adamjstewart, line 14: Can you sort these in the same order as the table in our paper? The goal is basically to copy the table from our paper so people know which weights to prefer. We also need to clarify that TM = Landsat 4–5 and ETM+ = L7 and OLI/TIRS = L8–9.

torchgeo/models/resnet.py
```diff
@@ -106,6 +113,110 @@ class ResNet18_Weights(WeightsEnum):  # type: ignore[misc]
         },
     )
 
+    LANDSAT_OLI_SR_MOCO = Weights(
```
adamjstewart, line 116: Sort in same order as table in paper

torchgeo/models/resnet.py
```diff
@@ -106,6 +113,110 @@ class ResNet18_Weights(WeightsEnum):  # type: ignore[misc]
...
+            "dataset": "SSL4EO-L",
+            "in_chans": 7,
+            "model": "resnet18",
+            "publication": "https://arxiv.org/abs/2306.09424",
+            "repo": "https://github.com/microsoft/torchgeo",
+            "ssl_method": "simclr",
+        },
+    )
+
+    LANDSAT_ETM_TOA_MOCO = Weights(
```
adamjstewart, line 142: Missing ETM_SR?

torchgeo/models/resnet.py
```diff
@@ -194,6 +305,110 @@ class ResNet50_Weights(WeightsEnum):  # type: ignore[misc]
...
+            "dataset": "SSL4EO-L",
+            "in_chans": 7,
+            "model": "resnet50",
+            "publication": "https://arxiv.org/abs/2306.09424",
+            "repo": "https://github.com/microsoft/torchgeo",
+            "ssl_method": "simclr",
+        },
+    )
+
+    LANDSAT_ETM_TOA_MOCO = Weights(
```
adamjstewart, line 334: Missing ETM_SR?

adamjstewart (changes requested):
Do we not have any U-Net pre-trained weights we can upload? We at least have the cloud cover models on our cluster. Not sure if the framework is in place to load those weights though...

torchgeo/models/resnet.py
adamjstewart: Still needs to be updated

docs/api/resnet_pretrained_weights.csv
nilsleh: do you want both accuracy and mIoU reported for NLCD and CDL?

nilsleh: do you want both accuracy and mIoU reported for NLCD and CDL?

docs/api/resnet_pretrained_weights.csv
nilsleh: what do you want to do with ResNet50_Weights.FMOW_RGB_GASSL?

adamjstewart: I forget, did we decide it was trained on Planet, or on many different satellites? If the latter, we could choose one of the following options:

* Single table for all miscellaneous weights
* Single table for all high-res RGB weights
* Separate table for all miscellaneous datasets

Since it's only one entry right now, I don't particularly care which we use. We can change it later.

docs/api/resnet_pretrained_weights.csv
adamjstewart: Yeah, let's do that. The table might be too wide but I think it adds an option to scroll. Will have to see what the table looks like in the doc preview.

adamjstewart:
None of the weights are in the right format...
```
missing_keys = ['cls_token', 'pos_embed', 'patch_embed.proj.weight', 'patch_embed.proj.bias', 'blocks.0.norm1.weight', 'blocks.0.norm1.bias', ...]
unexpected_keys = ['state_dict', 'hyper_parameters']
```

adamjstewart (approved):

PR #1484, opened 2023-07-18
Bump minimum support pycocotools version

adamjstewart:
pycocotools is failing to build in #1483. I want to investigate if this is broken in main and what the cause of the issue is.


adamjstewart:
This can't be a coincidence: https://pypi.org/project/Cython/#history

adamjstewart:
Pip does not provide a way to control the versions of build dependencies, so our only option is to change the minimum version of pycocotools we support.

calebrob6 (approved):

nilsleh (approved):

PR #509, opened 2022-04-18
RioXarrayDataset for in-memory geographical xarray.DataArray objects

weiji14:
A torchgeo.dataset based on an in-memory xarray.DataArray! Allows users to directly use a dataset they've already loaded from a GeoTIFF/NetCDF or any other processing pipeline without having to save it to a file first. Requires `rioxarray` as a dependency.

Usage example:

```python
import xarray as xr
import rioxarray

xr_dataarray = xr.DataArray(
    data=np.random.randn(5, 3),
    coords=dict(y=[5.6, 4.5, 3.4, 2.3, 1.2], x=[6.7, 7.8, 8.9]),
    dims=["y", "x"],
)
xr_dataarray.rio.set_crs(input_crs="EPSG:3857")

dataset = RioXarrayDataset(xr_dataarray=xr_dataarray)

sample = dataset[dataset.bounds]
print(sample)
```
produces:
```python
{'image': tensor([[ 1.0317, -0.4566, -0.2855],
        [ 0.6037, -1.5887, -1.4465],
        [-0.8714, -0.1645,  0.7559],
        [ 1.8187, -0.6460, -0.2239],
        [ 0.1100, -0.1918,  0.0911]], dtype=torch.float64), 'crs': CRS.from_epsg(3857), 'bbox': BoundingBox(minx=6.15, maxx=9.450000000000001, miny=0.65, maxy=6.1499999999999995, mint=0.0, maxt=9.223372036854776e+18)}
```

PR #1490, opened 2023-07-20
Xarray Dataset Support 

### LINKED ISSUE #1486: Xarray Dataset
### Summary

I am working with different climate data sources that come in the form of .netcdf files and xarrays. Although, I am not an expert in that domain, it seems that this is the go to data format that is frequently used. Since there are lots of features in Torchgeo that I would like to use with this data without having to reformat to tiff files for example, I think it could be quiet powerful to add support for Xarray datasets, even though it would be another couple dependencies to add to Torchgeo.

### Rationale

This could quiet possibly extend the horizon of users to other communities that work with Xarray data and benefit from all the tools Torchgeo already provides. In the majority of cases climate data also comes in the form of time-series so this would go hand in hand with the planned support for TimeSeries models and dataloading stuff in Torchgeo.

### Implementation

Both in #412 and #509 there was some discussion about this, but nothing was finalized. I am definitely willing to start on this but don't have a detailed plan yet as I first wanted to gather opinions on this.

### Alternatives

_No response_

### Additional information

_No response_

nilsleh:
This PR aims to add support for dataloading based on xarray datasets as the discussion began in #1486. 

The application I have for my project is taking in CMIP6 data from separate files that each define a climate variable and I need to extract sampled time-series. The sampling procedure can come in form of certain samplers as I started working on in #877 . 

The current draft is based on the closed #509 , but instead of a single xr.DataArray takes in a directory to gather files like `RasterDataset` does. 

The climate data for example does not have a CRS but they do come in different resolutions. Additionally, it could be that some data is time-series but others is not, so this needs to be merged correctly. Other climate data could also have a "depth" component.

Since there are a lot of things to consider, I would definitely appreciate input from more experienced users that have a better scope of common required work flows. 

### References

The following references were critical to getting this working correctly:

* https://corteva.github.io/rioxarray/stable/getting_started/crs_management.html
* https://www.unidata.ucar.edu/software/netcdf/examples/files.html

Closes #509
Closes #1486
Closes #2945


adamjstewart:
We discussed this over zoom, but my basic comments were something along the lines of:

* This base class should be generic enough to support any NetCDF-based dataset
* Most NetCDF files won't have a default CRS, we should default to lat/long (EPSG 4326 I believe)
* Not sure if files have a way of expressing their bounds, we should default to the entire Earth
* Subclasses can override the defaults to change to a different CRS or extent. This may require a custom `__getitem__` that parses the extent from a particular layer

Other than that, it's generally difficult to design a useful base class without lots of examples. I didn't write RasterDataset until we already had several examples of the same pattern in TorchGeo. The base class was an attempt to reduce code duplication and share some of the more complicated data access stuff. I think it's fine to design this base class with the few example datasets you have in mind and change it to add new features later.

nilsleh:
I think we would also have to make clear what assumptions we make for the netcdf files people use, for example: 
- should we assume that the dataset files they input are already in the same resolution for example climate data in lat/lon at a certain degree resolution, or should torchgeo support some resolution alignment (https://github.com/pangeo-data/xESMF seems like a interesting package for this), alternatively, some elaborate tutorial could show what exactly torchgeo expects as inputs and how to make that happen
- currently this PR supports netcdf files that are Xarray `Datasets` and not `DataArrays`, the former can include multiple variables at once while the latter can only support one variable, however, one would have to process them slightly different
- I think a lot of these datasets come with a time dimension so taking good care of how to handle this time dimension and aligning support with time-series sampling is important
- w.r.t to time, the dataset I have measures time as the number of hours since a certain date, however, I think since many other formats will be out there it would be beneficial if this was standardized to be a datetime timestamp, so again is this something we would assume that is already done or should torchgeo offer support here?

nilsleh:
I tested the xarray case I have with time-series indexing from #877 and I think this requires a discussion about assumptions about dateformats which is a bit tricky. Currently, the assumption is that the index and the sampler is working with `datetime.timestamps()` and that the xarray dataset are then indexed with a datetime object.

requirements/datasets.txt
```diff
@@ -9,4 +9,5 @@ radiant-mlhub==0.4.1
 rarfile==4.0
 scikit-image==0.21.0
 scipy==1.11.1
+xarray
```
adamjstewart, line 12:
```suggestion
xarray==2023.7.0
```

requirements/min-reqs.old
```diff
@@ -31,6 +31,7 @@ radiant-mlhub==0.3.0
 rarfile==4.0
 scikit-image==0.18.0
 scipy==1.6.2
+xarray
```
adamjstewart, line 34: Will need to determine the minimum version that works before merging

torchgeo/datasets/rioxarray.py
```diff
@@ -0,0 +1,156 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""In-memory geographical xarray.DataArray."""
```
adamjstewart, line 4: Do we only support DataArray or also Dataset? Most of the time data comes bundled together with multiple layers for different features.

tests/data/rioxarray/data.py
```diff
@@ -0,0 +1,65 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
adamjstewart, line 1: I usually add a shebang to data.py scripts and make them executable.

torchgeo/datasets/rioxarray.py
```diff
@@ -0,0 +1,156 @@
...
+
+import glob
+import os
+import re
+import sys
+from datetime import datetime
+from typing import Any, Callable, Optional, cast
+
+import netCDF4  # noqa: F401
```
adamjstewart, line 13: This isn't used at the moment and could probably be removed

torchgeo/datasets/rioxarray.py
```diff
@@ -0,0 +1,156 @@
...
+from rtree.index import Index, Property
+
+from .geo import GeoDataset
+from .utils import BoundingBox
+
+
+class RioXarrayDataset(GeoDataset):
+    """Wrapper for geographical datasets stored as Xarray Datasets.
+
+    Relies on rioxarray.
```
adamjstewart, line 27: You're currently not using rioxarray, nor do you have a dependency on it

torchgeo/datasets/rioxarray.py
```diff
@@ -0,0 +1,156 @@
...
+        root: str,
+        data_variables: list[str],
+        crs: Optional[CRS] = None,
+        res: Optional[float] = None,
+        transforms: Optional[Callable[[dict[str, Any]], dict[str, Any]]] = None,
+    ) -> None:
+        """Initialize a new Dataset instance.
+
+        Args:
+            root: directory with nc files
```
adamjstewart, line 44: Xarray supports many file formats other than NetCDF, let's keep this generic.

torchgeo/datasets/rioxarray.py
```diff
@@ -0,0 +1,156 @@
...
+            data_variables: data variables that should be gathered from the xr_datasets
```
adamjstewart, line 45: Is `xr_datasets` a variable? Or do you mean to say xarray datasets?

torchgeo/datasets/rioxarray.py
```diff
@@ -0,0 +1,156 @@
...
+        for filepath in glob.iglob(pathname, recursive=True):
+            match = re.match(filename_regex, os.path.basename(filepath))
+            if match is not None:
+                with xr.open_dataset(filepath, decode_times=True) as ds:
+                    if crs is None:
+                        crs = ds.rio.crs
+                    if res is None:
+                        res = ds.rio.resolution()[0]
+
+                    (minx, miny, maxx, maxy) = ds.rio.bounds()
```
adamjstewart, line 78: Will these need try-except statements for when this info isn't encoded in the files?

torchgeo/datasets/rioxarray.py
```diff
@@ -0,0 +1,156 @@
...
+
+                if hasattr(ds, "time"):
+                    try:
+                        indices = ds.indexes["time"].to_datetimeindex()
+                    except AttributeError:
+                        indices = ds.indexes["time"]
```
adamjstewart, line 84: Is this a standard attribute name or just something chosen by the authors of your particular NetCDF files? If necessary we can always add a class attribute that lists the names of layers to look for things like time from.

torchgeo/datasets/rioxarray.py
```diff
@@ -0,0 +1,156 @@
...
+        """
+        hits = self.index.intersection(tuple(query), objects=True)
+        items = [hit.object for hit in hits]
+
+        if not items:
+            raise IndexError(
+                f"query: {query} not found in index with bounds: {self.bounds}"
+            )
+
+        data_arrays: list["np.typing.NDArray[np.float32]"] = []
```
adamjstewart, line 125: Will they always be float32?

torchgeo/datasets/rioxarray.py
```diff
@@ -0,0 +1,156 @@
...
+        for item in items:
+            with xr.open_dataset(item, decode_cf=True) as ds:
+                if not ds.rio.crs:
+                    ds.rio.write_crs(self._crs, inplace=True)
```
adamjstewart, line 129: Does rioxarray automatically reproject to the right CRS if files are in multiple or if the user chooses a different CRS than the default (or if IntersectionDataset changes it)?

torchgeo/datasets/rioxarray.py
```diff
@@ -0,0 +1,156 @@
...
+                        time=slice(
+                            datetime.fromtimestamp(query.mint),
+                            datetime.fromtimestamp(query.maxt),
+                        )
+                    )
+                for variable in self.data_variables:
+                    if hasattr(clipped, variable):
+                        data_arrays.append(clipped[variable].data.squeeze())
+
+        sample = {"image": torch.from_numpy(np.stack(data_arrays)), "bbox": query}
```
adamjstewart, line 151: We should support both images and masks. See the `is_image` attribute in `RasterDataset` for how we do this there. You can also copy the `dtype` property to automatically choose what dtype to cast to.

adamjstewart: I'm not convinced this works correctly for multiple overlapping files. We shouldn't be stacking, we should be merging.

tests/data/rioxarray/data.py
```diff
@@ -0,0 +1,65 @@
...
+    ds.to_netcdf(path=filename)
+
+
+if __name__ == "__main__":
+    if os.path.isdir(DIR):
+        shutil.rmtree(DIR)
+    os.makedirs(DIR)
+    for var_name in VAR_NAMES:
+        for lats, lons, cf_time in zip(LATS, LONS, CF_TIME):
+            path = os.path.join(DIR, f"{var_name}_{lats}_{lons}.nc")
```
adamjstewart, line 62: Never thought about the possibility of the filename containing the bounds/res/crs. Have you seen this in the wild? If so, we could add a check for this in the regex and extract it from the filename like we do for time.

torchgeo/datasets/rioxarray.py
adamjstewart, line 1: I would recommend renaming this file. Otherwise the following become very different things:
```python
import rioxarray
import .rioxarray
```
Maybe call it `rioxr.py`? Or just throw it in `geo.py` with the other base classes.

adamjstewart (commented):
Difficult to review without several examples of what these datasets look like in real life. Maybe do a quick literature review and find a few good examples of popular NetCDF datasets from different sources? I want to see the diversity of file metadata. Like, do all datasets encode CRS/res/time/bounds in the same way or are they all different? This greatly influences how much customizability the base class requires.

nilsleh:
> Difficult to review without several examples of what these datasets look like in real life. Maybe do a quick literature review and find a few good examples of popular NetCDF datasets from different sources? I want to see the diversity of file metadata. Like, do all datasets encode CRS/res/time/bounds in the same way or are they all different? This greatly influences how much customizability the base class requires.

This is probably not exhaustive but I tried finding some sources from climate data and satellite optical data to cover some bases. 

- Xarray supports these [conventions](https://cfconventions.org/) for weather and climate data as suggested on their [docs](https://docs.xarray.dev/en/stable/user-guide/weather-climate.html). 
- For a tutorial notebook that works with the climate data that I am also using maybe [this notebook](https://comptools.climatematch.io/tutorials/W1D1_ClimateSystemOverview/student/W1D1_Tutorial5.html) gives insights into the meta data
- This is a [notebook](https://uwgda-jupyterbook.readthedocs.io/en/latest/modules/09_NDarrays_xarray_ERA5/09_NDarrays_xarray_ERA5_Part1_exercises.html) working with ERA reanalysis data which is also very popular
- [this](https://www.earthdatascience.org/courses/use-data-open-source-python/hierarchical-data-formats-hdf/open-MODIS-hdf4-files-python/) and [this](https://nordicesmhub.github.io/forces-2021/learning/example-notebooks/read-MODIS.html) are examples of xarray data from MODIS

torchgeo/datasets/rioxarray.py
nilsleh: The `xr.open_dataset()` function works for both I think so it should support both as far as I can tell and with the examples I have tried.

torchgeo/datasets/rioxarray.py
nilsleh: The functions in the `__getitem__` method like `ds.rio.crs` or `ds.rio.clip_box` require rioxarray.

torchgeo/datasets/rioxarray.py
nilsleh: sorry meant to say xarray datasets

torchgeo/datasets/rioxarray.py
```diff
@@ -0,0 +1,156 @@
...
+        self.index = Index(interleaved=False, properties=Property(dimension=3))
+
+        # Populate the dataset index
+        i = 0
+        pathname = os.path.join(root, self.filename_glob)
+        filename_regex = re.compile(self.filename_regex, re.VERBOSE)
+        for filepath in glob.iglob(pathname, recursive=True):
+            match = re.match(filename_regex, os.path.basename(filepath))
+            if match is not None:
+                with xr.open_dataset(filepath, decode_times=True) as ds:
```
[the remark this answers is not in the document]
nilsleh: As far as I am aware rioxarray only has a `rioxarray.open_rasterio()` function intended to read tif files and uses rasterio under the hood, while `xr.open_dataset()` is the generic form to support more formats.

torchgeo/datasets/rioxarray.py
nilsleh: I have encountered different naming schemes for spatial coordinate `x and y` or `lat and lon` but for the time dimension it was always called `time`. However, the `ds.rio.clip_box()` function in the `__getitem__` method expects the spatial coordinates to be named x and y otherwise it complains and tells you to rename them to x and y.

torchgeo/datasets/rioxarray.py
nilsleh: Good Point, I think it does not automatically reproject. So far I have only used climate data which doesn't explicitly encode or use a CRS, I should check with some MODIS files.

tests/data/rioxarray/data.py
nilsleh: No this was just a dummy naming scheme for test data. Maybe we should explicitly create test data for some of the different data cases like CMIP, ERA5, MODIS and others and write test cases for those.

torchgeo/datasets/rioxarray.py
nilsleh: So I am just doing this with the climate data, where there is no explicit CRS and `ds.rio.reproject()` also only works for 2D and 3D arrays, whereas the CMIP data I have has more dimensions so I get `rioxarray.exceptions.TooManyDimensions: Only 2D and 3D data arrays supported. Data variable: tos`.

torchgeo/datasets/rioxarray.py
nilsleh: You are right, maybe have to rethink the base class thing and have one for xarray climate data that is not using crs explicitly (`class XarrayDataset(GeoDataset)`)  and one that is intended for crs depending data sources like MODIS and more similar to `RasterDataset` and using `rioxarray` so the current naming `class RioXarrayDataset(GeoDataset)` but with the required functionality to handle overlapping files etc.

torchgeo/datasets/rioxarray.py
adamjstewart: I'd rather ignore that warning in pyproject.toml than add a fake import

torchgeo/datasets/rioxarray.py
adamjstewart: Probably best to keep it dynamic if we can't predict 100% of the time.

nilsleh:
I found one approach that merges the data back again, but I think it can be quiet slow potentially. Might want to explore that.

torchgeo/datasets/rioxr.py
```diff
@@ -0,0 +1,233 @@
...
+from rtree.index import Index, Property
+
+from .geo import GeoDataset
+from .utils import BoundingBox
+
+
+class RioXarrayDataset(GeoDataset):
+    """Wrapper for geographical datasets stored as Xarray Datasets.
+
+    Relies on rioxarray.
```
adamjstewart, line 27: Needs versionadded

torchgeo/datasets/rioxr.py
```diff
@@ -0,0 +1,233 @@
...
+    is_image = True
+
+    @property
+    def dtype(self) -> torch.dtype:
+        """The dtype of the dataset (overrides the dtype of the data file via a cast).
+
+        Returns:
+            the dtype of the dataset
+
+        .. versionadded:: 5.0
```
adamjstewart, line 42: Doesn't need versionadded. The property isn't new, the entire class is

adamjstewart (commented):
Tests are failing. Also curious if we can add a dep on only xarray or rioxarray but not both.

nilsleh:
Currently getting a `rasterio.errors.WindowError: Bounds and transform are inconsistent` error with the synthetic test data and trying to figure out why that is to fix the failing test. There should also be more diverse test cases given the variety of available datasets.

nilsleh:
https://github.com/jejjohnson/xrpatcher Potentially interesting utility for data loading

imSanko:
Seems like things are ongoing but does anyone need some help so that I can also provide a solution to fix this issue.

nilsleh:
> Seems like things are ongoing but does anyone need some help so that I can also provide a solution to fix this issue.

I won't have time to work on this in the near future so feel free to suggest an alternative approach or build on top of the ideas found in this PR.

tests/data/hdf5/data.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
adamjstewart: Reminder to self to update copyright

torchgeo/datasets/geo.py
```diff
@@ -622,6 +623,164 @@ def _load_warp_file(self, filepath: Path) -> DatasetReader:
...
+                        src = src.rio.write_crs(crs)
+
+                    if src.rio.crs != crs:
+                        src = src.rio.reproject(crs)
+
+                    filepaths.append(filepath)
+                    datetimes.append((tmin, tmax))
+                    geometries.append(shapely.box(*src.rio.bounds()))
+            except (OSError, ValueError):
+                # Skip files that xarray is unable to read
```
isaaccorley, line 690: Not required for this PR but it probably would be nice to warn a user if we can't read a file instead of silently skipping.

torchgeo/datasets/geo.py
```diff
@@ -622,6 +623,164 @@ def _load_warp_file(self, filepath: Path) -> DatasetReader:
...
+
+        image = self._merge_files(index.filepath, query)
+        sample: dict[str, Any] = {'crs': self.crs, 'bounds': query, 'image': image}
+
+        if self.transforms is not None:
+            sample = self.transforms(sample)
+
+        return sample
+
+    def _merge_files(self, filepaths: Sequence[str], query: GeoSlice) -> Tensor:
```
isaaccorley, line 741: I have a feeling this function can be optimized with some lazy xarray magic but I'd like to merge this first to unblock others and then make improvements later.

torchgeo/datasets/geo.py
```diff
@@ -622,6 +623,164 @@ def _load_warp_file(self, filepath: Path) -> DatasetReader:
             return src
 
 
+class XarrayDataset(GeoDataset):
+    """Abstract base class for :class:`GeoDataset` stored as raster files.
```
isaaccorley, line 627:
```suggestion
    """Abstract base class for :class:`GeoDataset` stored as xarray readable raster files.
```

isaaccorley (dismissed):

torchgeo/datasets/geo.py
adamjstewart: Let's debate this in a different PR. We've never done it before. The problem is things like shapefiles with dozens of auxiliary files that we're supposed to skip.

torchgeo/datasets/geo.py
adamjstewart: All raster files are xarray readable (just about). This should be an even broader class than RasterDataset. I thought about mentioning which I/O library is being used in each dataset, but that's more of an implementation detail. But I agree that it would be nice to somehow differentiate these two classes (even if they eventually merge).

isaaccorley (approved):

torchgeo/datasets/geo.py
```diff
@@ -622,6 +623,164 @@ def _load_warp_file(self, filepath: Path) -> DatasetReader:
...
+                            UserWarning,
+                        )
+                        src = src.rio.write_crs(crs)
+
+                    if src.rio.crs != crs:
+                        src = src.rio.reproject(crs)
+
+                    filepaths.append(filepath)
+                    datetimes.append((tmin, tmax))
+                    geometries.append(shapely.box(*src.rio.bounds()))
```
nilsleh, line 688: Does this handle different coordinate ranges correctly? i.e. some data comes as [-180, 180], other as [0, 360]. Similar question down below when different files get merged. I think I have also encountered xarrays where coordinates are sorted in ascending and descending order

adamjstewart: No, probably not. I'm trying to keep things minimal first and only add extra features as needed for datasets we plan to add. Rasterio 1.4+ seems to better support ascending vs. descending y coordinates (the inconsistent transforms and bounds bug you hit). Maybe it better supports [-180, 180] and [0, 360] too?

nilsleh: yeah not sure, just meant to check, because handling this would require someone to overwrite the base class right? Especially if you have "random" netcdf files and not collections like ERA5 for which we would have an own dataset class.

adamjstewart: I'll update the base class as needed to get ERA5 and CMIP6 working. No one is intended to use the base class directly, although they definitely will.

adamjstewart: Just realized I didn't read your comment in enough detail, although my answer mostly remains the same. Let's start adding datasets like ERA5/CMIP6/etc. I'm fully expecting (but not hoping) that one of these will be custom enough that it requires some kind of hacks like this. When we encounter one, we can add this function to the base class. Or other users can add this in a PR if needed. I just don't want to add stuff for the sake of "what if" until we actually see a need for it.

PR #1516, opened 2023-08-14
SSL4EO: add script to measure flops of models

adamjstewart:


experiments/ssl4eo/flops.py
```diff
@@ -0,0 +1,31 @@
...
+
+for model in models:
+    print(f"Model: {model}")
+
+    m = timm.create_model(model, num_classes=num_classes, in_chans=in_channels)
+
+    # Calculate memory requirements of model
+    mem_params = sum([p.nelement() * p.element_size() for p in m.parameters()])
+    mem_bufs = sum([b.nelement() * b.element_size() for b in m.buffers()])
+    mem = (mem_params + mem_bufs) / 2**20
```
calebrob6, line 25: Where does 2**20 come from? I thought MB was 10**6?

calebrob6 (approved):

experiments/ssl4eo/flops.py
adamjstewart: $2^{20} \approx 10^6$. I've seen both definitions of MB, not sure which is more "official".

adamjstewart: I'll keep it as is for now so we don't have to change the paper, but maybe we should change it to $10^6$ if we use the same script for future papers.

calebrob6: **I would if it was really easy ;) -- merged

PR #1540, opened 2023-09-01
Added functionality for validation split

### LINKED ISSUE #654: Inria dataset validation set
https://project.inria.fr/aerialimagelabeling/contest/ says, "For validation, we suggest to remove the first five images of every location (e.g., austin{1-5}.tif, chicago{1-5}.tif) from the training set."

It would be nice to implement this in our Dataset, then have the option to ignore it or not in the DataModule (i.e. when you want to train with as many labels as possible).

Tarandeep97:
**Closes #654: Implement Inria Dataset Validation Set Split**

**Summary:**
This pull request addresses issue #654, which involves implementing the validation set split for the Inria dataset as specified in the contest guidelines (https://project.inria.fr/aerialimagelabeling/contest/).

**Changes Made:**
Added functionality to the `inria.py` file.

**Details:**
Modified the `inria.py` file to implement the validation set split. The first five images of each location can now be removed by  setting split argument to valid.


Tarandeep97:

adamjstewart:
Can we call it "val" instead of "valid"? That's what most other datasets call it.

Tarandeep97:
changed "valid" to "val", as per convention.

adamjstewart:
Still a lot of failing tests, see https://torchgeo.readthedocs.io/en/stable/user/contributing.html for tips to fix these.

torchgeo/datasets/inria.py
```diff
@@ -89,21 +89,28 @@ def _load_files(self, root: str) -> list[dict[str, str]]:
         Returns:
             list of dicts containing paths for each pair of image and label
         """
-        files = []
-        root_dir = os.path.join(root, self.directory, self.split)
+        train_files = []
+        val_files = []
+        test_files = []
```
adamjstewart, line 94: What's the benefit of having 3 lists if you only ever return a single list?

torchgeo/datasets/inria.py
```diff
@@ -72,7 +72,7 @@ def __init__(
             RuntimeError: if dataset is missing
```
adamjstewart, line 72: Can you add this to the docstring:
```rst
.. versionchanged:: 0.5                                                          
   Added support for a *val* split.   
```

Tarandeep97:
Hi @adamjstewart, sorry for late reply. I was occupied with my thesis project. I am working on the feedback you have provided.

adamjstewart:
@Tarandeep97 no worries, I also just finished my thesis. Hope everything went okay!

torchgeo/datasets/inria.py
Tarandeep97: Docstring added.

adamjstewart:
I made the regex a bit simpler and got the tests working, let me know what you think!

Tarandeep97:
This looks perfect to me @adamjstewart . Thank you for helping!

adamjstewart (approved):

PR #1541, opened 2023-09-01
Refactor trainers

### LINKED ISSUE #1393: Refactor trainers
### Summary

Opening this issue so I don't forget to do this.

We added new MoCo and SimCLR trainers with a new design. We should do the same for our existing Trainers.

### Rationale

Kwargs are evil, default values are nice, type hints are nice

### Implementation

Same as MoCo and SimCLR

### Alternatives

_No response_

### Additional information

_No response_

adamjstewart:
In #1195 we introduced a new structure for our trainers. This PR updates all existing trainers to match.

- [x] Add new base class (reduces code duplication)
- [x] No `*args` or `**kwargs` (prevents argument typos, adds default values and type hints, required for LightningCLI)
- [x] No `typing.cast` (declare type when initialized)
- [x] `__init__` first (should be first thing in docs)
- [x] Added `configure_(losses|metrics|models)` methods (easier to override in subclass)

This is a great time to bring #996 back up. At this point, our "trainers" are referred to as:

* tasks: they all have a `Task` suffix
* trainers: they live in `torchgeo.trainers`
* models: this is what Lightning refers to them as
* modules: our YAML config file list them as `module:` (this will go away in my next PR)

I would love it if we could decide on a single naming scheme and be consistent...

Closes #1393

### DIFF
--- a/conf/deepglobelandcover.yaml
+++ b/conf/deepglobelandcover.yaml
@@ -4,8 +4,8 @@ module:
   model: "unet"
   backbone: "resnet18"
   weights: null
-  learning_rate: 1e-3
-  learning_rate_schedule_patience: 6
+  lr: 1e-3
+  patience: 6
   verbose: false
   in_channels: 3
   num_classes: 7
--- a/conf/deepglobelandcover.yaml
+++ b/conf/deepglobelandcover.yaml
@@ -6,7 +6,6 @@ module:
   weights: null
   lr: 1e-3
   patience: 6
-  verbose: false
   in_channels: 3
   num_classes: 7
   num_filters: 1
--- a/tests/trainers/test_byol.py
+++ b/tests/trainers/test_byol.py
@@ -86,7 +86,7 @@ class TestBYOLTask:
 
     @pytest.fixture
     def model_kwargs(self) -> dict[str, Any]:
-        return {"model": "resnet18", "in_channels": 13, "weights": None}
+        return {"model": "resnet18", "in_channels": 13}
 
     @pytest.fixture
     def weights(self) -> WeightsEnum:
--- a/tests/trainers/test_byol.py
+++ b/tests/trainers/test_byol.py
@@ -85,10 +85,6 @@ class TestBYOLTask:
         trainer.fit(model=model, datamodule=datamodule)
 
     @pytest.fixture
-    def model_kwargs(self) -> dict[str, Any]:
-        return {"model": "resnet18", "in_channels": 13}
-
-    @pytest.fixture
     def weights(self) -> WeightsEnum:
         return ResNet18_Weights.SENTINEL2_ALL_MOCO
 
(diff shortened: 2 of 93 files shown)


tests/trainers/test_byol.py
```diff
@@ -76,7 +74,6 @@ def test_trainer(
 
         # Instantiate model
         model = instantiate(conf.module)
-        model.backbone = SegmentationTestModel(**conf.module)
```
adamjstewart, line 79: `BYOLTask` has never had a `backbone` attribute. This line of code didn't do anything, and the tests are still slower and require more memory than necessary.

torchgeo/trainers/detection.py
```diff
@@ -58,37 +59,70 @@ class ObjectDetectionTask(LightningModule):
...
+       ['resnet18', 'resnet34', 'resnet50', 'resnet101', 'resnet152',
+       'resnext50_32x4d', 'resnext101_32x8d',
+       'wide_resnet50_2', 'wide_resnet101_2']
 
     .. versionadded:: 0.4
     """
 
-    def config_task(self) -> None:
-        """Configures the task based on kwargs parameters passed to the constructor."""
-        backbone_pretrained = self.hyperparams.get("pretrained", True)
```
adamjstewart, line 70: Previously this would use a pretrained backbone and download weights from the internet by default, which doesn't match the behavior of any other trainer.

conf/deepglobelandcover.yaml
```diff
@@ -4,9 +4,8 @@ module:
   model: "unet"
   backbone: "resnet18"
   weights: null
-  learning_rate: 1e-3
-  learning_rate_schedule_patience: 6
-  verbose: false
```
adamjstewart, line 9: Verbose isn't an option

torchgeo/trainers/byol.py
```diff
@@ -279,115 +283,98 @@ def update_target(self) -> None:
...
+        """Initialize a new BYOLTask instance.
+
+        Args:
+            model: Name of the timm model to use.
+            weights: Initial model weights. Either a weight enum, the string
+                representation of a weight enum, True for ImageNet weights, False
+                or None for random weights, or the path to a saved model state dict.
+            in_channels: Number of input channels to model.
+            lr: Learning rate for optimizer.
+            weight_decay: Weight decay (L2 penalty).
```
adamjstewart, line 315: Many of these parameters were not previously documented.

torchgeo/trainers/detection.py
```diff
@@ -274,6 +290,7 @@ def validation_step(self, *args: Any, **kwargs: Any) -> None:
 
     def on_validation_epoch_end(self) -> None:
         """Logs epoch level validation metrics."""
+        # TODO: why is this method necessary?
```
adamjstewart, line 293: None of our other trainers require special handling for metric logging. Is it only because of https://github.com/Lightning-AI/torchmetrics/pull/1832#issuecomment-1623890714?

nilsleh:
> * Should we have a public `configure_models`, `configure_losses`, `configure_metrics` for every trainer to make it easier for people to override these? This would be similar to `configure_optimizers`.

Regarding this point, for another domain library project I am also writing "trainers" that need a model, loss and metrics. When Isaac told me about hydra configs and instantiation I started using that at least for the model, loss and optimizer part. So you would have something like:

```
class BaseModel(LightningModule):
    def __init__(
        self,
        model: nn.Module,
        optimizer: type[torch.optim.Optimizer],
        loss_fn: nn.Module,
    ) -> None:
        self.model = model
        self.optimizer = optimizer
        self.lr_scheduler = lr_scheduler
        self.loss_fn = loss_fn

    def configure_optimizers(self) -> dict[str, Any]:
        """Initialize the optimizer."""
        optimizer = self.optimizer(params=self.parameters())
        return optimizer
```

You could also provide defaults. But what I like is that you have some nice control via config files, so I could have:

```
method:
  _target_: BaseModel
  model:
    _target_: some.MLP # timm.create_model for timm models
    # all arguments to initiate the mlp for example
    n_outputs: 1
    n_hidden: [50]
  optimizer:
    _target_: torch.optim.Adam # can change optimizers here easily
    _partial_: true
    lr: 0.003
  loss_fn:
    _target_: torch.nn.MSELoss
```
So just via config you have control about optimizer and model architecture. Not sure if this is good practice but it has been convenient for running experiments and keeping track of configurations.

adamjstewart:
You're in luck. Once this PR is merged I'll open a follow-up PR that switches everything to [LightningCLI](https://lightning.ai/docs/pytorch/stable/cli/lightning_cli.html) which supports exactly what you're describing without even changing the code: 

* https://lightning.ai/docs/pytorch/stable/cli/lightning_cli_intermediate_2.html#multiple-optimizers

This all relies on jsonargparse which supports command-line, YAML, and JSON configuration.

PR #1549, opened 2023-09-06
Advertise TorchGeo Slack

adamjstewart:
TorchGeo has a Slack channel. We should advertise it better so anyone can join.

The badge can be viewed at https://github.com/adamjstewart/torchgeo/tree/github/slack.

Slack no longer supports permanent invite links, so this link will expire and need to be updated after 400 people join. I don't think that will happen any time soon but if someone says it doesn't work, that's what happened.


adamjstewart:
Spack has a much fancier button: https://github.com/spack/spack

It automatically generates an invite link that never expires and displays the number of users on Slack, which is cool. But it requires a container running in the cloud to host it, which I don't have. Maybe @calebrob6 has access to something like that.

nilsleh (approved):

PR #1560, opened 2023-09-12
Add BioMassters Dataset

nilsleh:
This PR adds the BioMassters Dataset which was part of this [Driven Data Challenge](https://www.drivendata.org/competitions/99/biomass-estimation/page/536/#features_list) and currently under review as a paper. It consists of Sentinel 1 and 2 data as well as corresponding Aboveground Biomass pixelwise targets and is an effort lead by @nascetti-a . If you should have any comments or suggestions, feel welcome to do so.

TODOs:
- waiting to download the full data on my side to finish the directory structure and tests
- automatic download of dataset and verify
- fix plotting of S1

Example Plots:
Screenshot from 2023-09-14 10-58-52
Screenshot from 2023-09-14 10-58-02


torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,344 @@
...
+    * .tif files for Sentinel 1 and 2 data
+    * .tif file for pixel wise AGB target mask
+    * .csv files for metadata regarding features and targets
+
+    Dataset Features:
+
+    * 13,000 target AGB masks of size (256x256px)
+    * 12 months of data per target mask
+    * Sentinel 1 and Sentinel 2 data for each location
+    * Sentinel 1 (specify channels here) available for each month
```
adamjstewart, line 39: Either specify channels or remove band list

torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,344 @@
...
+    * Sentinel 2 (B02, B03, B04, B05, B06, B07, B08, B8A, B11, B12)
```
adamjstewart, line 40: I wonder why they didn't include all bands...

torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,344 @@
...
+      not available for each month due to ESA aquisition halt over the region
+      during particular periods
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * TBD
```
adamjstewart, line 46:
```suggestion
    * https://nascetti-a.github.io/BioMasster/
```
This is the closest thing we have for now, still citable even if it isn't on arXiv yet.

torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,344 @@
...
+
+    url = "https://huggingface.co/datasets/nascetti-a/BioMassters/resolve/main/{}"
+
+    md5s = {"train_features": "123", "test_features": "456", "train_agbm": "789"}
+
+    metadata_filename = "The_BioMassters_-_features_metadata.csv.csv"
+
+    def __init__(
+        self,
+        root: str,
```
adamjstewart, line 62:
```suggestion
        root: str = "data",
```
Always good to have a default, especially if the user will auto-download the dataset anyway. This matches all our other datasets

torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,344 @@
...
+
+        assert set(sensors).issubset(
+            set(self.valid_sensors)
+        ), f"Please choose a subset of valid sensors {self.valid_sensors}."
+        self.sensors = sensors
+        self.as_time_series = as_time_series
+        self.download = download
+        self.checksum = checksum
+
+        try:
```
adamjstewart, line 100: See #1586, we can merge these in either order

torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,344 @@
...
+        self.df = pd.read_csv(os.path.join(self.root, self.metadata_filename))
+
+        # filter sensors
+        self.df = self.df[self.df["satellite"].isin(self.sensors)]
+
+        # filter split
+        self.df = self.df[self.df["split"] == self.split]
+
+        # generate numerical month from filename since first month is September
+        # and has numerical index of 0
```
adamjstewart, line 119: These comments are fantastic, thank you!

torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,344 @@
...
+            )
+            filter_df = filter_df[filter_df["satellite"] == len(self.sensors)].drop(
+                "satellite", axis=1
+            )
+            # guarantee that each sample has corresponding number of images available
+            self.df = self.df.merge(filter_df, on=["chip_id", "month"], how="inner")
+
+            self.df["num_index"] = self.df.groupby(["chip_id", "month"]).ngroup()
+
+    def __getitem__(self, index: int) -> dict[str, Any]:
```
adamjstewart, line 142:
```suggestion
    def __getitem__(self, index: int) -> dict[str, Tensor]:
```
Pretty sure you only return Tensor

torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,344 @@
...
+        # sort by satellite and month to return correct order
+        sample_df.sort_values(
+            by=["satellite", "num_month"], inplace=True, ascending=True
+        )
+
+        filepaths = sample_df["filename"].tolist()
+        sample: dict[str, Tensor] = {}
+        for sens in self.sensors:
+            sens_filepaths = [fp for fp in filepaths if sens in fp]
+            sample[f"image_{sens}"] = self._load_input(sens_filepaths)
```
adamjstewart, line 165: This relates to #985. I would love to be able to avoid custom keys in dicts (it prevents us from using these datasets with our trainers or type hint them). I'm fine with this for now, but we should think of whether or not it's possible to keep these as standard names.

torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,344 @@
...
+    ) -> None:
+        """Initialize a new instance of BioMassters dataset.
+
+        Args:
+            root: root directory where dataset can be found
+            split: train or test split
+            sensors: which sensors to consider for the sample
+            as_time_series: whether or not to return all available
+                time-steps or just a single one for a given target location
+            download: if True, download dataset and store it in the root directory
```
adamjstewart, line 77: I download doesn't get fixed before 0.5 I would rather merge this without downloading and add download support in 0.6.

torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,344 @@
...
+                img = img[-1, ...]
+            if sens == "S2":
+                img = img[[2, 1, 0], ...]
+                img = percentile_normalization(img.transpose(1, 2, 0))
+            else:
+                co_polarization = img[0]  # transmit == receive
+                cross_polarization = img[1]  # transmit != receive
+                ratio = co_polarization / cross_polarization
+
+                # https://gis.stackexchange.com/a/400780/123758
```
adamjstewart, line 311: This only works for certain scales, see: https://github.com/microsoft/torchgeo/pull/1151#discussion_r1121153094

If you can't get it working, you can always plot them as separate grayscale images like I did for SSL4EO-S12. Maybe @RitwikGupta can help if he has time

adamjstewart (changes requested):
Need to add to docs

torchgeo/datasets/biomassters.py
adamjstewart: #1586 has been merged, pandas is now a required dep so you can simplify a lot of this.

tests/data/biomassters/data.py
```diff
@@ -0,0 +1,137 @@
...
+
+import csv
+import hashlib
+import os
+import shutil
+
+import numpy as np
+import rasterio
+
+metadata_train = "The_BioMassters_-_features_metadata.csv.csv"
```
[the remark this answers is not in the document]
nilsleh: Yes, https://huggingface.co/datasets/nascetti-a/BioMassters/tree/main I think it will be corrected but for now its the case.

tests/datasets/test_biomassters.py
```diff
@@ -0,0 +1,64 @@
...
+        self, monkeypatch: MonkeyPatch, tmp_path: Path, request: SubRequest
+    ) -> BioMassters:
+        split, sensors, as_time_series = request.param
+
+        url = os.path.join("tests", "data", "biomassters", "{}")
+        monkeypatch.setattr(BioMassters, "url", url)
+
+        # copy data for now
+        shutil.rmtree(tmp_path)
+        shutil.copytree(os.path.join("tests", "data", "biomassters"), str(tmp_path))
```
adamjstewart, line 32: What if we drop `tmp_path` from the function entirely and just give it the `tests/data/biomassters` directory as the root dir? Then we don't need to bother copying data.

tests/datasets/test_biomassters.py
```diff
@@ -0,0 +1,64 @@
...
+    @pytest.fixture(
+        params=product(["train", "test"], [["S1"], ["S2"], ["S1", "S2"]], [True, False])
+    )
+    def dataset(
+        self, monkeypatch: MonkeyPatch, tmp_path: Path, request: SubRequest
+    ) -> BioMassters:
+        split, sensors, as_time_series = request.param
+
+        url = os.path.join("tests", "data", "biomassters", "{}")
+        monkeypatch.setattr(BioMassters, "url", url)
```
adamjstewart, line 28: You don't need to monkeypatch this anymore, you can remove monkeypatch from the function signature too

torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,287 @@
...
+
+    * https://nascetti-a.github.io/BioMasster/
+
+    .. versionadded:: 0.5
+    """
+
+    valid_splits = ["train", "test"]
+    valid_sensors = ["S1", "S2"]
+
+    url = "https://huggingface.co/datasets/nascetti-a/BioMassters/resolve/main/{}"
```
adamjstewart, line 55: You could remove this variable if you want, or you could keep it, up to you

torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,287 @@
...
+
+    metadata_filename = "The_BioMassters_-_features_metadata.csv.csv"
+
+    def __init__(
+        self,
+        root: str = "data",
+        split: str = "train",
+        sensors: list[str] = ["S1", "S2"],
```
adamjstewart, line 63: Does it have to be a list or can it be a `Sequence`? That would allow people to use tuples too

torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,287 @@
...
+        as_time_series: bool = False,
+    ) -> None:
+        """Initialize a new instance of BioMassters dataset.
+
+        Args:
+            root: root directory where dataset can be found
+            split: train or test split
+            sensors: which sensors to consider for the sample
```
adamjstewart, line 71: Can you document the available options?

torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,287 @@
...
+            as_time_series: whether or not to return all available
+                time-steps or just a single one for a given target location
```
adamjstewart, line 73: How is that single one chosen, most recent or random? Should document better here

torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,287 @@
...
+
+        RuntimeError:
+            AssertionError: if ``split`` or ``sensors`` is invalid
+            ImportError: if pandas is not installed
```
adamjstewart, line 77: Can remove ImportError now

torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,287 @@
...
+
+        return sample
+
+    def __len__(self) -> int:
+        """Return the length of the dataset.
+
+        Returns:
+            length of the dataset
+        """
+        print(self.df)
```
adamjstewart, line 165: Can remove debugging statement

torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,287 @@
...
+        return len(self.df["num_index"].unique())
+
+    def _load_input(self, filenames: list[str]) -> Tensor:
+        """Load the input imagery at the index.
+
+        Args:
+            filenames: list of filenames corresponding to input
```
adamjstewart, line 172: Need to document returns

torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,287 @@
...
+            arr: "np.typing.NDArray[np.float_]" = src.read()
+
+        target = torch.from_numpy(arr).float()
+        return target
+
+    def _verify(self) -> None:
+        """Verify the integrity of the dataset.
+
+        Raises:
+            RuntimeError: if ``download=False`` but dataset is missing or checksum fails
```
adamjstewart, line 202: download isn't an option

torchgeo/datasets/biomassters.py
```diff
@@ -0,0 +1,287 @@
...
+        exists = []
+
+        filenames = [f"{self.split}_features", self.metadata_filename]
+        for filename in filenames:
+            pathname = os.path.join(self.root, filename)
+            exists.append(os.path.exists(pathname))
+        if all(exists):
+            return
+
+        # Check if the user requested to download the dataset
```
adamjstewart, line 214: No download

adamjstewart (changes requested):

torchgeo/datasets/biomassters.py
adamjstewart: Do the filenames of the downloaded files tell you? If not, maybe just remove "(specify channels here)" so our docs don't look like TODO statements

adamjstewart:
Updating the branch may fix the RtD tests

torchgeo/datasets/biomassters.py
nilsleh: `as_time_series` changes the number of samples there are, so if True then a sample consists of all available time-steps but if False, then each time step becomes its own sample with the target always being the same

torchgeo/datasets/biomassters.py
adamjstewart: I assume this is still broken? We could fix it in a follow-up patch release I suppose.

adamjstewart (approved):
I think the S1 plot is the only remaining thing that's broken. If we run out of time to fix this I'm fine with merging as is and fixing it in the future.

PR #1587, opened 2023-09-25
Ensure that none of our minimum version tests are skipped

adamjstewart:
Some of our minimum version tests are being skipped. This doesn't ensure that our minimum versions actually work. Let's fix that.

### DIFF
--- a/tests/datasets/test_landcoverai.py
+++ b/tests/datasets/test_landcoverai.py
@@ -75,7 +75,7 @@ class TestLandCoverAI:
     def dataset(
         self, monkeypatch: MonkeyPatch, tmp_path: Path, request: SubRequest
     ) -> LandCoverAI:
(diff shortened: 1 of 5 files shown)


tests/datasets/test_landcoverai.py
```diff
@@ -75,7 +75,7 @@ class TestLandCoverAI:
     def dataset(
         self, monkeypatch: MonkeyPatch, tmp_path: Path, request: SubRequest
     ) -> LandCoverAI:
-        pytest.importorskip("cv2", minversion="4.4.0.46")
+        pytest.importorskip("cv2", minversion="4.4.0")
```
adamjstewart, line 78: `cv2.__version__` only includes the first 3 components

calebrob6 (approved):

PR #1604, opened 2023-09-30
0.5.0 release

adamjstewart:
#1520

### DIFF
--- a/docs/tutorials/getting_started.ipynb
+++ b/docs/tutorials/getting_started.ipynb
@@ -189,6 +189,7 @@
    "outputs": [],
    "source": [
     "chesapeake_root = os.path.join(tempfile.gettempdir(), \"chesapeake\")\n",
+    "os.makedirs(chesapeake_root, exist_ok=True)\n",
     "chesapeake = ChesapeakeDE(chesapeake_root, crs=naip.crs, res=naip.res, download=True)"
    ]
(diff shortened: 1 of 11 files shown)


torchgeo/datamodules/chesapeake.py
```diff
@@ -93,7 +93,7 @@ def __init__(
         """
         # This is a rough estimate of how large of a patch we will need to sample in
         # EPSG:3857 in order to guarantee a large enough patch in the local CRS.
-        self.original_patch_size = patch_size * 2
+        self.original_patch_size = patch_size * 3
```
adamjstewart, line 96: Just a heads up @calebrob6. Not sure what changed since the last release that made this necessary for the test to pass. We should really convert this to a normal RasterDataset.

docs/tutorials/getting_started.ipynb
```diff
@@ -189,6 +189,7 @@
    "outputs": [],
    "source": [
     "chesapeake_root = os.path.join(tempfile.gettempdir(), \"chesapeake\")\n",
+    "os.makedirs(chesapeake_root, exist_ok=True)\n",
```
adamjstewart, line 192: This is actually a bug in TorchGeo: #1605. Pretty minor so I don't think it's worth holding up the release for.

Issue #99, opened 2021-09-01
Dataset downloading expected behavior pt. 2


calebrob6:
When I instantiate a dataset with `download=False` and `checksum=False` I expect it to assume everything is in place, however our current setup usually checks to make sure that the archive file exists. If the archive file is 100+ GB then it is totally reasonable for users to delete it but keep the downloaded data.

I think `_check_integrity` should be something like this:
```
    def _check_integrity(self) -> bool:
        """Check integrity of dataset.

        Returns:
            True if dataset MD5s match, else False
        """
        return check_integrity(
            os.path.join(self.root, self.filename),
            self.md5,
        )
```
and we only call it if `self.checksum` is true.

Datasets that still use `_check_integrity` (and presumably follow the old convention):

- [ ] Advance
- [x] Benin Cashews
- [ ] CBF
- [ ] COWC
- [x] CV4A Kenya
- [x] Cyclone
- [ ] ETCI 2021
- [ ] EuroCrops
- [ ] Eurosat
- [ ] GID15
- [ ] LEVIRCD
- [ ] Loveda
- [ ] SEN12MS
- [ ] So2Sat
- [x] SpaceNet
- [ ] UCMerced
- [ ] VHR-10

For a good example to copy, I think CDL is the first dataset I updated to use the new download style.

calebrob6:
@estherrolf (one of our first users!) just hit this problem with a manually downloaded version of the ChesapeakeCVPR dataset. I think it is worth making the change as this will happen especially with the larger datasets.

adamjstewart:
I also don't like the "dataset not found or corrupted" error message. Like, which is it??

calebrob6:
I think we need a list here of all datasets that do _not_ follow these conventions / a good example dataset to copy.

adamjstewart:
Datasets that still use `_check_integrity` (and presumably follow the old convention):

- [ ] Advance
- [ ] Benin Cashews
- [ ] CBF
- [ ] COWC
- [ ] CV4A Kenya
- [ ] Cyclone
- [ ] ETCI 2021
- [ ] Eurosat
- [ ] GID15
- [ ] LEVIRCD
- [ ] Loveda
- [ ] VHR10
- [ ] SEN12MS
- [ ] So2Sat
- [ ] SpaceNet
- [ ] UCMerced

For a good example to copy, I think CDL is the first dataset I updated to use the new download style.

PR #1612, opened 2023-10-03
Add check if path is vsi

### LINKED ISSUE #1605: Download support broken for some GeoDatasets if directory does not exist
### Description

@adriantre this is a bug we (mostly I) introduced in #1442/#1597. The issue is that if `os.path.isdir(file)` is False, we automatically add the file to `self.files`, and then `self.files` evaluates to True in `_verify`. I think the safest solution is to not use `self.files` in `_verify`, but you might be able to think of a better solution.

I first noticed this during https://github.com/microsoft/torchgeo/pull/1442/commits/8ec2e932810f55128ca8c0ceb78152389d3c5f16 but didn't realize the impact of it until https://github.com/microsoft/torchgeo/pull/1604/commits/fe546bf160a02acf6275b3f97c1ecd94fec81490. I still want to get this release out so I figured I would give us something to do in 0.5.1. I don't know how many other datasets are impacted.

Current workaround is to create the directory before downloading. Not hard, just annoying.

### Steps to reproduce

```console
> python3 
>>> from torchgeo.datasets import ChesapeakeDE
>>> ChesapeakeDE('fake/path', download=True)
  File "/Users/Adam/torchgeo/torchgeo/datasets/chesapeake.py", line 138, in __init__
    super().__init__(paths, crs, res, transforms=transforms, cache=cache)
  File "/Users/Adam/torchgeo/torchgeo/datasets/geo.py", line 434, in __init__
    raise FileNotFoundError(msg)
FileNotFoundError: No ChesapeakeDE data was found in `paths='fake/path''`
```

### Version

0.5.0

adriantre:
Fix #1605 

In 0.5 we included property `files` in GeoDatasets. `files` tries to list all files matching the `filename_glob`. It also tries to support virtual file systems. In doing so, it accidentally accepts non-directories as valid files.

The proposed solution will do a basic check of the path prefix to see if it looks like a vsi. Paths not matching these criteria will be ignored (to match behaviour pre version 0.5). Could `raise FileNotFound` instead. Then `_verify`-methods of Datasets could `try-except` to decide if files need downloading.

Current limitations:
* Does not check if the vsi file exists, or is a 'directory'
* Does not list files in vsi directories.

### DIFF
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -63,6 +63,14 @@ class CustomNonGeoDataset(NonGeoDataset):
         return 2
 
 
+class MockRasterDataset(RasterDataset):
+    def __init__(self, paths: Union[str, Iterable[str]]):
+        """Override super init because it raises FileNotFoundError.
+        Is used to test behaviour of property `files` for non-local paths.
+        """
+        self.paths = paths
+
+
 class TestGeoDataset:
     @pytest.fixture(scope="class")
     def dataset(self) -> GeoDataset:
@@ -211,6 +219,21 @@ class TestRasterDataset:
     )
     def test_files(self, paths: Union[str, Iterable[str]]) -> None:
         assert 1 <= len(NAIP(paths).files) <= 2
+
+    def test_ignore_non_existing_file_or_dir(self) -> None:
+        paths = ["/non/existing/file.tif", "/non/existing/dir"]
+        assert len(MockRasterDataset(paths).files) == 0
+
+    def test_files_property_for_vsi(self) -> None:
+        # Tests only a subset of schemes and combinations.
+        paths = [
+            "file://directory/file.tif" "az://azure_bucket/prefix/file.tif",
+            "zip+az://azure_bucket/prefix/archive.zip!folder_in_archive/file.tif",
+            "zip://archive.zip!folder/file.tif",
+            "/vsiaz/azure_bucket/prefix/file.tif",
+            "/vsizip//vsiaz/azure_bucket/prefix/archive.zip/folder_in_archive/file.tif",
+        ]
+        assert len(MockRasterDataset(paths).files) == len(paths)
 
     def test_getitem_single_file(self, naip: NAIP) -> None:
         x = naip[naip.bounds]
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -231,7 +231,6 @@ class TestRasterDataset:
         paths = [tmp_path.as_posix(), (tmp_path / "non_existing_file.tif").as_posix()]
         assert len(MockRasterDataset(paths).files) == 0
 
-    @pytest.mark.filterwarnings("ignore:.*Path was ignored.*")
     def test_files_property_for_virtual_files(self) -> None:
         # Tests only a subset of schemes and combinations.
         paths = [
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -65,8 +65,9 @@ class CustomNonGeoDataset(NonGeoDataset):
 
 class MockRasterDataset(RasterDataset):
     def __init__(self, paths: Union[str, Iterable[str]]):
-        """Override super init because it raises FileNotFoundError.
-        Is used to test behaviour of property `files` for non-local paths.
+        """Used to test behaviour of property `files` for non-local paths.
+        This is needed because super init may raise FileNotFoundError because
+        it opens the files. When testing instance properties, we need init to succeed.
         """
         self.paths = paths
 
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -227,10 +227,11 @@ class TestRasterDataset:
     def test_files_property_for_virtual_files(self) -> None:
         # Tests only a subset of schemes and combinations.
         paths = [
-            "file://directory/file.tif" "az://azure_bucket/prefix/file.tif",
-            "zip+az://azure_bucket/prefix/archive.zip!folder_in_archive/file.tif",
+            "file://directory/file.tif",
             "zip://archive.zip!folder/file.tif",
+            "az://azure_bucket/prefix/file.tif",
             "/vsiaz/azure_bucket/prefix/file.tif",
+            "zip+az://azure_bucket/prefix/archive.zip!folder_in_archive/file.tif",
             "/vsizip//vsiaz/azure_bucket/prefix/archive.zip/folder_in_archive/file.tif",
         ]
         assert len(MockRasterDataset(paths).files) == len(paths)
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -155,7 +155,7 @@ class TestGeoDataset:
             dataset & ds2  # type: ignore[operator]
 
     def test_files_property_for_non_existing_file_or_dir(self, tmp_path: Path) -> None:
-        paths = [tmp_path.as_posix(), (tmp_path / "non_existing_file.tif").as_posix()]
+        paths = [str(tmp_path), str(tmp_path / "non_existing_file.tif")]
         with pytest.warns(UserWarning, match="Path was ignored."):
             assert len(CustomGeoDataset(paths=paths).files) == 0
 
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -154,15 +154,10 @@ class TestGeoDataset:
         ):
             dataset & ds2  # type: ignore[operator]
 
-    def test_files_property_warns_non_existing_file(self, tmp_path: Path) -> None:
-        path = (tmp_path / "non_existing_file.tif").as_posix()
-        with pytest.warns(UserWarning, match="Path was ignored."):
-            CustomGeoDataset(paths=path).files
-
-    @pytest.mark.filterwarnings("ignore:.*Path was ignored.*")
     def test_files_property_for_non_existing_file_or_dir(self, tmp_path: Path) -> None:
         paths = [tmp_path.as_posix(), (tmp_path / "non_existing_file.tif").as_posix()]
-        assert len(CustomGeoDataset(paths=paths).files) == 0
+        with pytest.warns(UserWarning, match="Path was ignored."):
+            assert len(CustomGeoDataset(paths=paths).files) == 0
 
     def test_files_property_for_virtual_files(self) -> None:
         # Tests only a subset of schemes and combinations.
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -4,7 +4,7 @@ import os
 import pickle
 from collections.abc import Iterable
 from pathlib import Path
-from typing import Union
+from typing import Optional, Union
 
 import pytest
 import torch
@@ -33,11 +33,13 @@ class CustomGeoDataset(GeoDataset):
         bounds: BoundingBox = BoundingBox(0, 1, 2, 3, 4, 5),
         crs: CRS = CRS.from_epsg(4087),
         res: float = 1,
+        paths: Optional[Union[str, Iterable[str]]] = None,
     ) -> None:
         super().__init__()
         self.index.insert(0, tuple(bounds))
         self._crs = crs
         self.res = res
+        self.paths = paths or []
 
     def __getitem__(self, query: BoundingBox) -> dict[str, BoundingBox]:
         hits = self.index.intersection(tuple(query), objects=True)
@@ -61,15 +63,6 @@ class CustomNonGeoDataset(NonGeoDataset):
 
     def __len__(self) -> int:
         return 2
-
-
-class MockRasterDataset(RasterDataset):
-    def __init__(self, paths: Union[str, Iterable[str]]):
-        """Used to test behaviour of property `files` for non-local paths.
-        This is needed because super init may raise FileNotFoundError because
-        it opens the files. When testing instance properties, we need init to succeed.
-        """
-        self.paths = paths
 
 
 class TestGeoDataset:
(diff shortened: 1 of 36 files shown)


adamjstewart:
Why not check for all files containing `://`? Do `os.path.isdir` or `glob.glob` work for `file://`?

adriantre:
I forgot to push the isfile in `elif os.path.isfile(path) or path_is_vsi(path):`

> Do `os.path.isdir` or `glob.glob` work for `file://`?

I don't think it does. But it is a valid vsi scheme.

> Why not check for all files containing ://?

Should be sufficient, but then we only support the apache/rasterio syntax `az://bucket/blob.tif`. This is equivalent to `/vsiaz/bucket/blob.tif` which is the gdal syntacx. Both syntax is supported by rasterio. In fact, rasterio converts it to the latter before calling the underlying gdal methods.

adriantre:
Could create a feature request in rasterio for non-private methods/classes for doing this verification?

adriantre:
This could be useful, but it will be deprecated in rasterio 1.4. Cannot find any information pointing to why or what will replace it. I'll ask in the rasterio discussion board.

```python
from rasterio.path import ParsedPath
p = ParsedPath.from_uri("zip://*.tif::files.zip")
p.scheme  # 'zip'
p.is_local  # True
p.is_remote  # False
```

I see that rasterio are working on api-changes for opening vsi. Pasting here for reference. I do not yet know if this means will need to change anything when upgrading rasterio.

docs: 
https://rasterio.readthedocs.io/en/latest/topics/vsi.html#python-file-openers
usage in tests: 
https://github.com/rasterio/rasterio/blob/d966440c06f3324aca1fa761d490cc780a9f619c/tests/test_pyopener.py#L20

adriantre:
FYI, I started this 'Q&A' over at rasterio 
https://github.com/rasterio/rasterio/discussions/2930

adriantre:
Rasterio don't want to expose the path-validation, and say this API is prone to change (as we expected). They suggest copying in the relevant parts of this validation, as I do in this PR.

What do we think. Are we on the right track here? I do think it would be practical to give the users the possibility to implement their own logic for files that are not visible to `os` and `glob` packages, somethink like: `else: handle_non_local_path`.

calebrob6:
I'm fine with keeping the check on our side as the changes should be "_someone wants a new VSI in GeoDataset_" vs. an update to `rasterio` breaks us.

adamjstewart:
My concern is that if the list of supported schemes changes according to rasterio version, it will be annoying to validate in torchgeo. What's the downside of treating all paths with a `:` in them as remote files and letting rasterio issue it's own error msg if something doesn't work?

adriantre:
I agree. If the user inputs invalid paths, then rasterio will complain.

`files` property will then: 
- for local files: validate existence and recursively list files matching glob
- for vsi: return without further ado 
    (requires users to perform listing and validation themselves, and provide as `paths`)

This is outside the scope of this bug-fix and is something we can fix when the use of vsi shows/grows. We should inform the user about this difference though. And when rasterio settles on how they treat vsi then we can take a new look at the possibilities it may bring.

torchgeo/datasets/utils.py
```diff
@@ -737,3 +737,23 @@ def percentile_normalization(
         (img - lower_percentile) / (upper_percentile - lower_percentile + 1e-5), 0, 1
     )
     return img_normalized
+
+
+def path_is_vsi(path: str) -> bool:
```
adamjstewart, line 742: Any idea what VSI stands for? Wondering if we should call this vfs instead.

torchgeo/datasets/utils.py
```diff
@@ -737,3 +737,23 @@ def percentile_normalization(
...
+    """Checks if the given path is pointing to a Virtual File System.
+
+    NB! Does not check if the path exists, or if it is a dir or file.
```
adamjstewart, line 745:
```suggestion
    .. note::
       Does not check if the path exists, or if it is a dir or file.
```

torchgeo/datasets/utils.py
```diff
@@ -737,3 +737,23 @@ def percentile_normalization(
...
+
+    VSI can for instance be Cloud Storage Blobs or zip-archives.
+    They will start with a prefix indicating this.
+    For examples of these, see references for the two accepted syntaxes.
+    - https://gdal.org/user/virtual_file_systems.html
+    - https://rasterio.readthedocs.io/en/latest/topics/datasets.html
```
adamjstewart, line 751:
```suggestion

    * https://gdal.org/user/virtual_file_systems.html
    * https://rasterio.readthedocs.io/en/latest/topics/datasets.html
```

tests/datasets/test_geo.py
```diff
@@ -63,6 +63,14 @@ def __len__(self) -> int:
         return 2
 
 
+class MockRasterDataset(RasterDataset):
```
adamjstewart, line 66: Let's remove this and instead test to ensure that an error is raised: https://docs.pytest.org/en/latest/how-to/assert.html#assertions-about-expected-exceptions

torchgeo/datasets/geo.py
```diff
@@ -298,8 +304,10 @@ def files(self) -> set[str]:
             if os.path.isdir(path):
                 pathname = os.path.join(path, "**", self.filename_glob)
                 files |= set(glob.iglob(pathname, recursive=True))
-            else:
+            elif os.path.isfile(path) or path_is_vsi(path):
                 files.add(path)
+            else:
+                continue
```
adamjstewart, line 310: This doesn't do anything and can be removed

torchgeo/datasets/utils.py
adriantre: "Virtual Files System Interface" is the closest I can find to an explanation of the abbreviation. Could also be a play on URI? 😅

What I do like about keeping VSI, is communicating that the intention is to support _GDAL_ virtual file systems, not _any_ virtual filesystem.

torchgeo/datasets/geo.py
adriantre: Do you think it would make sense to rase a warning here? e.g.

```python
warnings.warn(
    "Could not find any relevant files for provided path '{path}'.",
    UserWarning,
)
```

tests/datasets/test_geo.py
adriantre: What I want to test here is the behaviour of the `files`-property: that virtual files are added to the list even though they dont exists locally. 

But that cannot be tested without calling init, which raises. If I change the test to check that it raises, it then communicates to other developers that it is testing something different.

What do you think?

torchgeo/datasets/geo.py
adamjstewart: Might be a good idea. We raise an error if no files were found, but we don't warn the user if multiple paths were given and only some were valid.

tests/datasets/test_geo.py
```diff
@@ -212,6 +221,29 @@ def sentinel(self, request: SubRequest) -> Sentinel2:
     def test_files(self, paths: Union[str, Iterable[str]]) -> None:
         assert 1 <= len(NAIP(paths).files) <= 2
 
+    def test_files_property_warns_non_existing_file(self, tmp_path: Path) -> None:
+        path = (tmp_path / "non_existing_file.tif").as_posix()
+        with pytest.warns(UserWarning, match="Path was ignored."):
+            MockRasterDataset(path).files
+
+    @pytest.mark.filterwarnings("ignore:.*Path was ignored.*")
```
adamjstewart, line 229: We don't want to ignore warnings, we want to ensure warnings: https://docs.pytest.org/en/7.4.x/how-to/capture-warnings.html#warns

tests/datasets/test_geo.py
```diff
@@ -212,6 +221,29 @@ def sentinel(self, request: SubRequest) -> Sentinel2:
...
+    def test_files_property_for_non_existing_file_or_dir(self, tmp_path: Path) -> None:
+        paths = [tmp_path.as_posix(), (tmp_path / "non_existing_file.tif").as_posix()]
+        assert len(MockRasterDataset(paths).files) == 0
+
+    @pytest.mark.filterwarnings("ignore:.*Path was ignored.*")
```
adamjstewart, line 234: This shouldn't warn, right?

adriantre: True! 😅

tests/datasets/test_geo.py
adriantre: I see. the test above this ensures the warning. My idea was that these two tests different things. But I can merge them into one if that is better?

see `test_files_property_warns_non_existing_file`

adamjstewart: Let's merge them. They're both testing the same situation, we're just ensuring two things.

adriantre: Agreed

tests/datasets/test_geo.py
```diff
@@ -212,6 +221,29 @@ def sentinel(self, request: SubRequest) -> Sentinel2:
     def test_files(self, paths: Union[str, Iterable[str]]) -> None:
         assert 1 <= len(NAIP(paths).files) <= 2
 
+    def test_files_property_warns_non_existing_file(self, tmp_path: Path) -> None:
```
adamjstewart, line 224: Since we're testing `GeoDataset.files`, can we move these tests to `TestGeoDataset`? We can also rename `MockRasterDataset` to `MockGeoDataset`.

torchgeo/datasets/utils.py
```diff
@@ -737,3 +737,25 @@ def percentile_normalization(
...
+    For examples of these, see references for the two accepted syntaxes.
+
+    * https://gdal.org/user/virtual_file_systems.html
+    * https://rasterio.readthedocs.io/en/latest/topics/datasets.html
+
+    Args:
+        path: string representing a directory or file
+
+    Returns:
+        True if path is on a virtual file system, else False
```
adamjstewart, line 759: Add this:
```rst
.. versionadded:: 0.6
```
Technically it will be added in 0.5.1 because we're fixing a bug, but 0.6 will be the first feature release where it officially appears.

tests/datasets/test_geo.py
adriantre: Totally makes sense. I pushed a suggestion. GeoDataset does not accept arg `paths`, so I added it to the existing `CustomGeoDataset` in the test. Removed the MockRasterDataset as GeoDataset.__init__ does not try to open the paths.

tests/datasets/test_geo.py
```diff
@@ -33,11 +33,13 @@ def __init__(
         bounds: BoundingBox = BoundingBox(0, 1, 2, 3, 4, 5),
         crs: CRS = CRS.from_epsg(4087),
         res: float = 1,
+        paths: Optional[Union[str, Iterable[str]]] = None,
     ) -> None:
         super().__init__()
         self.index.insert(0, tuple(bounds))
         self._crs = crs
         self.res = res
+        self.paths = paths or []
```
adamjstewart, line 42: Does this work?
```suggestion
        paths: Union[str, Iterable[str]] = [],
    ) -> None:
        super().__init__()
        self.index.insert(0, tuple(bounds))
        self._crs = crs
        self.res = res
        self.paths = paths
```

tests/datasets/test_geo.py
```diff
@@ -152,6 +154,28 @@ def test_and_nongeo(self, dataset: GeoDataset) -> None:
         ):
             dataset & ds2  # type: ignore[operator]
 
+    def test_files_property_warns_non_existing_file(self, tmp_path: Path) -> None:
+        path = (tmp_path / "non_existing_file.tif").as_posix()
+        with pytest.warns(UserWarning, match="Path was ignored."):
+            CustomGeoDataset(paths=path).files
+
+    @pytest.mark.filterwarnings("ignore:.*Path was ignored.*")
+    def test_files_property_for_non_existing_file_or_dir(self, tmp_path: Path) -> None:
```
adamjstewart, line 163: Should be merged. Can we test both a non-existing filename and directory?

tests/datasets/test_geo.py
adriantre: This would do the same, but I have heard that [mutable input defaults should be avoided](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/) :P

tests/datasets/test_geo.py
adriantre: I though we concluded that these tested two different things, but yes I can merge them. 

It does test both file and dir.

adamjstewart: I thought we agreed to merge them: https://github.com/microsoft/torchgeo/pull/1612#discussion_r1372262744

The latter tests both but the former only tests file. Just make sure the merged function tests both.

adriantre: You are right, my bad 😅 I totally read that wrong

tests/datasets/test_geo.py
```diff
@@ -152,6 +154,23 @@ def test_and_nongeo(self, dataset: GeoDataset) -> None:
         ):
             dataset & ds2  # type: ignore[operator]
 
+    def test_files_property_for_non_existing_file_or_dir(self, tmp_path: Path) -> None:
+        paths = [tmp_path.as_posix(), (tmp_path / "non_existing_file.tif").as_posix()]
```
adamjstewart, line 158:
```suggestion
        paths = [str(tmp_path), str(tmp_path / "non_existing_file.tif")]
```
We're working on adding pathlib support in a different PR. For now, let's just cast to a string. We also need to test on Windows which isn't POSIX.

adamjstewart (approved):

adamjstewart:
When I try to download AbovegroundLiveWoodyBiomassDensity, it complains that my directory doesn't exist and will be ignored. This is because we check `self.files` in `_verify`. Wonder if there's a way we could ignore that warning...

adamjstewart:
I would love a common `_verify` method but haven't had the time/energy to do it. This relates to #99.

PR #1647, opened 2023-10-11
Remove configure_callbacks

### LINKED ISSUE #1641: Unable to log_model="all" with wandb_logger
### Description

Configured to log all checkpoints, but only a single is logged

### Steps to reproduce

```python
from lightning.pytorch.loggers import WandbLogger

wandb_logger = WandbLogger(
    project="eurosat-test3", 
    name="imagenet", 
    log_model="all", 
    save_dir = "wandb_logs_tests"
)

trainer = Trainer(
    logger=wandb_logger,
    min_epochs=15,
    max_epochs=25,
)

glob("wandb_logs_tests/eurosat-test3/*/checkpoints/*.ckpt") # returns only a single checkpoint
```

### Version

0.5.0

### LINKED ISSUE #1642: trainers docs show configuring callbacks but these will not be used
### Issue

The [trainer](https://torchgeo.readthedocs.io/en/stable/tutorials/trainers.html) docs show setting `callbacks=[checkpoint_callback, early_stopping_callback],` but if you run this you will get a warning:

```
The following callbacks returned in `LightningModule.configure_callbacks` will override existing 
callbacks passed to Trainer: EarlyStopping, ModelCheckpoint
```

Honestly I am not a fan of this behaviour, but regardless the docs need ammending

### Fix

The docs should state the behaviour of tasks, and show how to use a conventional lightning module in their place when the default behaviour is not desired

### LINKED ISSUE #1645: Checkpoint saving not working as expected
### Description

After migrating to release 0.5.0 noticed that checkpoint saving is not working as expected.

## description
tried different configuration e.g., `checkpoint_callback = ModelCheckpoint(monitor="val_loss", dirpath=ckpt_dir, save_last=True, every_n_epochs=1, save_top_k=1)` for example when running 20-30 epochs for training a model.
after training was completed could not find the ckpt file. what was found was a single ckpt file of the first epoch only, in a wrong directory.

## severance
the bug is very limiting. for example, after hours of training a model, there is no way to load the model from a checkpoint to run inference. the single shot to run inference was during the same run. 

## expected behavior
using a given configuration expected to see:
- checkpoint files saved every number of epoch
- the last epoch checkpoint file
- the checkpoints should have been saved to the given directory

## observed behavior
- after training several epochs only the first was saved.
- the single checkpoint were saved to another directory under the logger output

## initial investigation
1. checkpoint callback created and training fit called
2. later, see image and call stack:
seems like c'tor called again with save_last=None
image

3. when saving later supposed to happen, the save_last is None:
image

4. last checkpoint saving is skipped

### Steps to reproduce

1. create a checkpoint callback and use different checkpoints saving parameters e.g., checkpoint_callback = ModelCheckpoint(monitor="val_loss", dirpath=ckpt_dir, save_last=True, every_n_epochs=1, save_top_k=1)
2. call trainer fit and run several epochs
3. check expected results:
- saving location as expected e.g., under C:\foo 
- check last epoch checkpoint saved - must have last.ckpt
- check how many checkpoints were saved e.g., every 2 etc

### Version

torchgeo version 0.5.0, lightning version 2.0.9

robmarkcole:
The use of configure_callbacks was blocking any use of `callbacks` and caused https://github.com/microsoft/torchgeo/issues/1641

This is now removed. This will require the use to configure these callbacks but this is already documented [here](https://torchgeo.readthedocs.io/en/stable/tutorials/trainers.html)

Fixes #1641 
Fixes #1642 
Fixes #1645


robmarkcole:
Unsure why pytest self cancelled

robmarkcole:
OK some args may lead nowhere after this change, e.g. `patience=6`

adamjstewart:
Let's wait for feedback from https://github.com/Lightning-AI/lightning/issues/18784

adamjstewart:
> Unsure why pytest self cancelled

This has been happening to all of our tests recently and I have no idea why. Anyone who figures this out is my hero.

adamjstewart:
https://github.com/Lightning-AI/lightning/issues/18784 proposed one possible workaround that will allow us to define defaults but also allow the user to override them. But I'm not sure if the added complexity is worth it. Thoughts?

robmarkcole:
It does sound like a lot of extra complexity to take on and manage - I still think leaving it to the users to set their own callbacks whilst providing a few examples of best practice is the way to go

adamjstewart (approved):

PR #1650, opened 2023-10-12
Fix classes in EuroSAT

robmarkcole:
- Closes https://github.com/microsoft/torchgeo/issues/1648
- Correctly (alphabetically) sorts the eurosat classes in docstring
- Removes `classes` attribute from eurosat, ucmerced and reseic45 datasets


torchgeo/datasets/eurosat.py
```diff
@@ -74,16 +74,16 @@ class EuroSAT(NonGeoClassificationDataset):
         "test": "7ae5ab94471417b6e315763121e67c5f",
     }
     classes = [
```
adamjstewart, line 76: I think we should just remove this variable since the base class sets it automatically, but curious what @calebrob6 thinks

robmarkcole: done

adamjstewart (approved):

PR #1654, opened 2023-10-12
Debug failing CI

adamjstewart:
Complete shot in the dark to see if this fixes our canceled tests.


burakekim:
Just a wild guess: Would disabling the coverage report in `pytest --cov=torchgeo --cov-report=xml --durations=10` help in better troubleshooting, given that it introduces additional time overhead? They all exit after 8m and the ones that pass are executed within ~6m.

adamjstewart:
At this point I'm open to any wild guess, let's see.

adamjstewart:
New clue: skip only happens on Ubuntu

adamjstewart:
Next clue: we're running out of space:
```
  File "/opt/hostedtoolcache/Python/3.11.5/x64/lib/python3.11/pathlib.py", line 1116, in mkdir
    os.mkdir(self, mode)
FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/torchgeo/torchgeo/.pytest_cache/v/cache'
During handling of the above exception, another exception occurred:
Error: No space left on device : '/home/runner/runners/2.309.0/_diag/pages/617c6e66-62ab-4661-9c1b-3caf8dabc48f_def3b5b4-f2a1-5e57-62e9-dcf39b95de2d_1.log'
```

adamjstewart:
Results of storage use experiments:

1. We start with 21G of free space remaining
2. Cloning the repo adds 164M
3. Installing Python and apt deps adds <1G
4. **Installing pip deps adds 8G!**
5. **2.7G of this is pip's cache and can be removed**
6. **Running pytest adds 10.6G!**
7. Only 5M of this is in the actual torchgeo git dir, rest must be in tmp?

So adding `pip cache purge` will probably get our tests working again, but we should really investigate where that 10.6G is hiding.

adamjstewart:
Yep, that worked. Will submit a separate PR to keep the commit history clean.

PR #1677, opened 2023-10-18
Add CropHarvest Dataset

GeorgeHuber:
Draft PR to add cropharvest dataset to torchgeo.

- [x] Implement the dataset extending either [GeoDataset](https://torchgeo.readthedocs.io/en/stable/api/datasets.html#torchgeo.datasets.GeoDataset) or [NonGeoDataset](https://torchgeo.readthedocs.io/en/stable/api/datasets.html#torchgeo.datasets.NonGeoDataset)
- [X] Add the dataset definition to torchgeo/datasets/__init__.py
- [x] Add a data.py script to tests/data/<new dataset>/ that generates test data with the same directory structure/file naming conventions as the new dataset
- [x] Add appropriate tests with 100% test coverage to tests/datasets/
- [x] Add the dataset to docs/api/datasets.rst
- [x] Add the dataset metadata to either docs/api/geo_datasets.csv or docs/api/non_geo_datasets.csv

Sample plot for single pixel timeseries over 12 months:
Screenshot 2024-01-08 at 11 05 12 PM


calebrob6:
Thanks for the contribution @GeorgeHuber! Let me know if you want help with the linting / mypy (you can read more about it here https://torchgeo.readthedocs.io/en/latest/user/contributing.html#linters)

GeorgeHuber:
> @GeorgeHuber please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,317 @@
...
+from torch import Tensor
+
+from .geo import NonGeoDataset
+from .utils import download_and_extract_archive, download_url
+
+
+class CropHarvest(NonGeoDataset):
+    """CropHarvest dataset.
+
+    The CropHarvest <https://github.com/nasaharvest/cropharvest> datataset is a
```
adamjstewart, line 25:
```suggestion
    The `CropHarvest <https://github.com/nasaharvest/cropharvest>`_ dataset is a
```

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,317 @@
...
+    classification dataset.
```
adamjstewart, line 26:
```suggestion
    classification dataset for crop classification.
```
or something like that, add a little more detail about the intended use of the dataset

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,317 @@
...
+
+    Dataset features:
+
+    * _ pixel single timeseries with croptype labels
```
adamjstewart, line 30: _?

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,317 @@
...
+
+    Dataset format:
+
+    * images are 12x18 ndarrays with 18 bands over 12 months
+
+    Dataset properties:
+
+    1. is_crop - whether or not a single pixel contains cropland
+    2. classification_label - optional field identifying a specific croptype
+    3. dataset - source datset for the imagery
```
adamjstewart, line 41:
```suggestion
    3. dataset - source dataset for the imagery
```
So different predictions were made by different satellites?

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,317 @@
...
+    4. lat
+    5. lon
+
+
```
adamjstewart, line 45: Can remove a couple empty lines here

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,317 @@
...
+    This dataset requires the following additional library to be installed:
+
+       * `h5py <https://pypi.org/project/h5py/>`_ to load the dataset
+    """
+
+    # *https://github.com/nasaharvest/cropharvest/blob/main/cropharvest/bands.py
+    all_bands = [
+        "VV",
+        "VH",
+        "B2",
```
adamjstewart, line 60: Note that we won't be able to use this data in @yichiac's paper because:

1. It is missing S2 B1/B10 so we can't use any of our pre-trained models
2. It is pixel-wise, so we can't use a CNN

We would have to collect our own imagery if we want to use this.

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,317 @@
...
+        "B9",
+        "B11",
+        "B12",
+        "temperature_2m",
+        "total_precipitation",
+        "elevation",
+        "slope",
+        "NDVI",
+    ]
+    rgb_bands = ["B11", "B8", "B2"]
```
adamjstewart, line 77: Normally we visualize Sentinel-2 imagery using Bands 4, 3, 2. Where did 11, 8, 2 come from?

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,317 @@
...
+        root: str = "data",
+        transforms: Optional[Callable[[dict[str, object]], dict[str, object]]] = None,
+        download: bool = False,
+        checksum: bool = False,
+    ) -> None:
+        """Initialize a new CropHarvest dataset instance.
+
+        Args:
+            root: root directory where dataset can be found
+            split: one of "train" or "val"
```
adamjstewart, line 109: This parameter does not exist

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,317 @@
...
+            transforms: a function/transform that takes input sample and its target as
+                entry and returns a transformed version
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+
+        Raises:
+            RuntimeError: if ``download=False`` but dataset is missing or checksum fails
```
adamjstewart, line 116: See #1714, we added a new error type specifically for this case that you should use

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,317 @@
...
+            index: index to return
+
+        Returns:
+            data and label at that index
+        """
+        files = self.files[index]
+        data = self._load_array(files["chip"])
+
+        label = self._load_label(files["index"], files["dataset"])
+        sample = {"data": data, "label": label}
```
adamjstewart, line 147: We don't really have any other single-pixel datasets so not sure what to call the dict keys.

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,317 @@
...
+            raise ImportError(
+                "h5py is not installed and is required to use this dataset"
+            )
+        filename = os.path.join(path)
+        with h5py.File(filename, "r") as f:
+            array = f.get("array")[()]
+            tensor = torch.from_numpy(array).float()
+            return tensor
+
+    def _load_label(self, idx: str, dataset: str) -> str:
```
adamjstewart, line 233: This returns a string but ML requires ints. If the dataset is binary I would use 0 for None and 1 for Some

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,317 @@
...
+        row = row.to_dict(orient="records")[0]
+        label = "None"
+        if row["properties.label"]:
+            label = row["properties.label"]
+        elif row["properties.is_crop"] == 1:
+            label = "Some"
+
+        return label
+
+    def _check_integrity(self) -> bool:
```
adamjstewart, line 258: This follows the old dataset style, I would instead base this off the newer style. Search similar datasets for `_verify`.

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,317 @@
...
+            filepath = os.path.join(self.root, self.directory, filename)
+            if not os.path.exists(filepath):
+                return False
+        return True
+
+    def _download(self) -> None:
+        """Download the dataset and extract it.
+
+        Raises:
+            AssertionError: if the checksum of split.py does not match
```
adamjstewart, line 275: What is split.py?

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,317 @@
...
+        )
+
+        download_url(
+            self.file_dict["labels"]["url"],
+            self.root,
+            filename=os.path.join(self.directory, self.file_dict["labels"]["filename"]),
+            md5=self.file_dict["labels"]["md5"] if self.checksum else None,
+        )
+
+    def plot(self, sample: dict[str, object], subtitle: Optional[str] = None) -> Figure:
```
adamjstewart, line 297: Can you add an example plot to the PR description so we can make sure plotting works correctly? Not sure if plot even makes sense for a single-pixel dataset.

adamjstewart (changes requested):

torchgeo/datasets/cropharvest.py
GeorgeHuber: I believe this field just refers to where the labels come from while imagery is standardized and downloaded from eo_data according to the cropharvest docs.

tests/datasets/test_cropharvest.py
```diff
@@ -0,0 +1,108 @@
...
+
+pytest.importorskip("h5py", minversion="3")
+
+
+def download_url(url: str, root: str, filename: str, md5: str) -> None:
+    shutil.copy(url, os.path.join(root, filename))
+
+
+class TestCropHarvest:
+    file_dict = {
```
adamjstewart, line 28: Instead of replacing the entire dict, you could replace just the keys you need to replace (url, md5) using `monkeypatch.setitem`

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,673 @@
...
+    classification dataset for crop classification.
+
+    Dataset features:
+
+    * single pixel timeseries with croptype labels
+    * 18 bands per image over 12 months
+
+    Dataset format:
+
+    * images are 12x18 ndarrays with 18 bands over 12 months
```
adamjstewart, line 35:
```suggestion
    * arrays are 12x18 with 18 bands over 12 months
```
We use Torch Tensors anyway

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,673 @@
...
+
+    Dataset properties:
+
+    1. is_crop - whether or not a single pixel contains cropland
+    2. classification_label - optional field identifying a specific croptype
+    3. dataset - source dataset for the imagery
+    4. lat
+    5. lon
```
adamjstewart, line 43:
```suggestion
    4. lat - latitude
    5. lon - longitude
```

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,673 @@
...
+        "B12",
+        "temperature_2m",
+        "total_precipitation",
+        "elevation",
+        "slope",
+        "NDVI",
+    ]
+    rgb_bands = ["B4", "B3", "B2"]
+
+    classes = [
```
adamjstewart, line 79: This is a lot of classes. Do we need to list these in the dataset, or can they be dynamically read from a file?

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,673 @@
...
+        Args:
+            root: root directory where dataset can be found
+            transforms: a function/transform that takes input sample and its target as
+                entry and returns a transformed version
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+
+        Raises:
+            DatasetNotFoundError: if ``download=False`` but
+                dataset is missing or checksum fails
```
adamjstewart, line 468:
```suggestion
            DatasetNotFoundError: If dataset is not found and *download* is False.
```
Just to match the description from other datasets

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,673 @@
...
+
+        Returns:
+            pandas dataframe containing label data for each feature
+        """
+        filename = self.file_dict["labels"]["extracted_filename"]
+        with open(os.path.join(root, filename), encoding="utf8") as f:
+            data = json.load(f)
+
+            pd.json_normalize(data["features"])
+            df = pd.json_normalize(data["features"])
```
adamjstewart, line 546: Why are we running this twice? What does `json_normalize` do that we can't do with `json.load`?

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,673 @@
...
+
+    def _load_array(self, path: str) -> Tensor:
+        """Load an individual single pixel timeseries.
+
+        Args:
+            path: path to the image
+
+        Returns:
+            the image
+            ImportError if h5py is not installed
```
adamjstewart, line 558: Can we move this error doc up to the `__init__` method? Hidden methods (that start with a single underscore) don't actually make it into the documentation.

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,673 @@
...
+    def _load_label(self, idx: str, dataset: str) -> Tensor:
+        """Load the croptype label for a single pixel timeseries.
+
+        Args:
+            idx: sample index in labels.geojson
+            dataset: dataset name to query labels.geojson
+        Returns:
+            the croptype label ,
+            "Other" if no croptype is defined but is_crop is true,
+            or "None" if there are no crops
```
adamjstewart, line 581:
```suggestion
            the crop type label
```

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,673 @@
...
+            tensor = torch.from_numpy(array).float()
+            return tensor
+
+    def _load_label(self, idx: str, dataset: str) -> Tensor:
+        """Load the croptype label for a single pixel timeseries.
+
+        Args:
+            idx: sample index in labels.geojson
+            dataset: dataset name to query labels.geojson
+        Returns:
```
adamjstewart, line 578:
```suggestion

        Returns:
```

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,673 @@
...
+        elif row["properties.is_crop"] == 1:
+            label = "Other"
+
+        return torch.tensor(self.classes.index(label))
+
+    def _verify(self) -> None:
+        """Verify the integrity of the dataset.
+
+        Raises:
+            DatasetNotFoundError: If dataset is not found and *download* is False.
```
adamjstewart, line 601: No need to document this here, it won't make it into the API docs. Let's just do it in the `__init__` only

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,673 @@
...
+
+        Returns:
+            a matplotlib Figure with the rendered sample
+        """
+        fig, axs = plt.subplots()
+        bands = [self.all_bands.index(band) for band in self.rgb_bands]
+        rgb = np.array(sample["array"])[:, bands]
+        normalized = rgb / np.max(rgb, axis=1, keepdims=True)
+        axs.imshow(normalized[None, ...])
+        axs.set_title(f'Croptype: {self.classes[sample["label"]]}')
```
adamjstewart, line 668:
```suggestion
        axs.set_title(f'Crop type: {self.classes[sample["label"]]}')
```

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,673 @@
...
+            suptitle: optional subtitle to use for figure
+
+        Returns:
+            a matplotlib Figure with the rendered sample
+        """
+        fig, axs = plt.subplots()
+        bands = [self.all_bands.index(band) for band in self.rgb_bands]
+        rgb = np.array(sample["array"])[:, bands]
+        normalized = rgb / np.max(rgb, axis=1, keepdims=True)
+        axs.imshow(normalized[None, ...])
```
adamjstewart, line 667: The example plot doesn't look very "wheat" colored. Maybe the normalization is changing the color and we should remove it? Also, we can probably remove at least the y-axis, it doesn't have any meaning. Might be good to add an x-axis label saying that this is the month, and count from 1–12 instead of 0–11.

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,323 @@
...
+    download_url,
+    extract_archive,
+    percentile_normalization,
+)
+
+
+class CropHarvest(NonGeoDataset):
+    """CropHarvest dataset.
+
+    `CropHarvest <https://github.com/nasaharvest/cropharvest>`_ is a
```
adamjstewart, line 30:
```suggestion
    `CropHarvest <https://github.com/nasaharvest/cropharvest>`__ is a
```
rST is the worst file format ever created...

Double underscore needed so that the link in the table points to the class docs instead of to GitHub.

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,323 @@
...
+    * https://openreview.net/forum?id=JtjzUXPEaCu
+
+    This dataset requires the following additional library to be installed:
+
+       * `h5py <https://pypi.org/project/h5py/>`_ to load the dataset
+
+    .. versionadded:: 0.6
+    """
+
+    # *https://github.com/nasaharvest/cropharvest/blob/main/cropharvest/bands.py
```
adamjstewart, line 61:
```suggestion
    # https://github.com/nasaharvest/cropharvest/blob/main/cropharvest/bands.py
```

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,323 @@
...
+        self.classes = np.insert(self.classes, 0, ["None", "Other"])
+
+    def __getitem__(self, index: int) -> dict[str, Tensor]:
+        """Return an index within the dataset.
+
+        Args:
+            index: index to return
+
+        Returns:
+            single pixel timeseries array and label at that index
```
adamjstewart, line 147:
```suggestion
            single pixel time-series array and label at that index
```

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,323 @@
...
+
+class CropHarvest(NonGeoDataset):
+    """CropHarvest dataset.
+
+    `CropHarvest <https://github.com/nasaharvest/cropharvest>`_ is a
+    crop classification dataset.
+
+    Dataset features:
+
+    * single pixel timeseries with croptype labels
```
adamjstewart, line 35:
```suggestion
    * single pixel time series with crop-type labels
```

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,323 @@
...
+            pandas dataframe containing label data for each feature
+        """
+        filename = self.file_dict["labels"]["extracted_filename"]
+        with open(os.path.join(root, filename), encoding="utf8") as f:
+            data = json.load(f)
+            df = pd.json_normalize(data["features"])
+            return df
+
+    def _load_array(self, path: str) -> Tensor:
+        """Load an individual single pixel timeseries.
```
adamjstewart, line 208:
```suggestion
        """Load an individual single pixel time series.
```

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,323 @@
...
+        import h5py
+
+        filename = os.path.join(path)
+        with h5py.File(filename, "r") as f:
+            array = f.get("array")[()]
+            tensor = torch.from_numpy(array)
+            return tensor
+
+    def _load_label(self, idx: str, dataset: str) -> Tensor:
+        """Load the croptype label for a single pixel timeseries.
```
adamjstewart, line 225:
```suggestion
        """Load the crop-type label for a single pixel time series.
```

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,323 @@
...
+    * 18 bands per image over 12 months
+
+    Dataset format:
+
+    * arrays are 12x18 with 18 bands over 12 months
+
+    Dataset properties:
+
+    1. is_crop - whether or not a single pixel contains cropland
+    2. classification_label - optional field identifying a specific croptype
```
adamjstewart, line 45:
```suggestion
    2. classification_label - optional field identifying a specific crop type
```

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,323 @@
...
+
+    def _load_label(self, idx: str, dataset: str) -> Tensor:
+        """Load the croptype label for a single pixel timeseries.
+
+        Args:
+            idx: sample index in labels.geojson
+            dataset: dataset name to query labels.geojson
+
+        Returns:
+            the crop type label
```
adamjstewart, line 232:
```suggestion
            the crop-type label
```

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,323 @@
...
+        Args:
+            root: root dir of dataset
+
+        Returns:
+            pandas dataframe containing label data for each feature
+        """
+        filename = self.file_dict["labels"]["extracted_filename"]
+        with open(os.path.join(root, filename), encoding="utf8") as f:
+            data = json.load(f)
+            df = pd.json_normalize(data["features"])
```
adamjstewart, line 204: Remind me why this is needed? Can't we use `json.load` only? If not, can we instead use `pd.read_json`?

torchgeo/datasets/cropharvest.py
```diff
@@ -0,0 +1,323 @@
...
+            suptitle: optional subtitle to use for figure
+
+        Returns:
+            a matplotlib Figure with the rendered sample
+        """
+        fig, axs = plt.subplots()
+        bands = [self.all_bands.index(band) for band in self.rgb_bands]
+        rgb = np.array(sample["array"])[:, bands]
+        normalized = percentile_normalization(rgb)
+        axs.imshow(normalized[None, ...])
```
adamjstewart, line 314: Still need to remove normalization here so the color matches during plotting

adamjstewart (commented):
Almost finished! Just a couple typo fixes and questions left.

torchgeo/datasets/cropharvest.py
GeorgeHuber: From my testing, pd.read_json fails on the data because the list of objects to make the dataframe is not at the root. So the data must first be loaded and then the dataframe created with ["features"] as the root. The json_normalize call flattens the nested objects into the pandas dataframe to make queries and calls to .unique() possible rather than working with the list directly. Is there an alternative way you would suggest going about this?

adamjstewart: Nah, that seems reasonable to me. Thanks for clarifying!

torchgeo/datasets/cropharvest.py
GeorgeHuber: All of my values for the B4, B3, and B2 bands are in the thousands. Is this a sign of a potential issue or could you clarify what you mean by removing the normalization?

adamjstewart: No, that's normal. Most satellites record values in 12 or 16 bits instead of 8 bits. In the Sentinel-2 dataset, we divide by 10K to get the values to a range that looks reasonable color-wise. Can you try that to see if the colors look more like we would expect (green for most crops)?

GeorgeHuber: Screenshot 2024-01-16 at 6 09 29 PM
The plots are looking quite dark, could I add a fixed value to bring them up to a better range?

adamjstewart: Try dividing by 3K, that's what we did in SSL4EO.

GeorgeHuber: Screenshot 2024-01-17 at 10 09 19 AM
Screenshot 2024-01-17 at 10 18 13 AM

is this looking closer to normal?

adamjstewart: Hmm, closer, but still weird. I would expect it to be pretty green for the whole middle of the year. Are we using the right bands? Does the raw data from the dataset include any normalization? Did you remove the normalization from the plot function and only include the division?

GeorgeHuber: A few interesting things:

- I double checked the bands being used and confirmed there is no normalization of the raw data or during plotting except for division, all seems to be good there
- It seems time-series in the dataset do not start at January, but rather a month defined in the metadata. I can update the plot accordingly but this will only shift the graph, though that might make a little more sense looking at the forage crop plot, knowing it starts in February
- I also believe the maize plot is an anomaly as the date the label was assigned occurred almost two years after the end of the time-series. After investigating further samples, this seems to be uncommon as most labels during or just around the time-series
Screenshot 2024-01-19 at 12 08 04 AM
Screenshot 2024-01-19 at 12 21 52 AM
Screenshot 2024-01-19 at 12 21 59 AM
Screenshot 2024-01-19 at 12 22 15 AM
Screenshot 2024-01-19 at 12 22 22 AM

adamjstewart: Alright, those look less crazy. Let's go with this so as not to sink any more time into this dataset. Thanks for validating!

PR #1678, opened 2023-10-18
Added complete support for pathlib

### LINKED ISSUE #1616: Support for pathlib
### Summary

We should add support for Python's [pathlib](https://docs.python.org/3/library/pathlib.html) module.

### Rationale

Users may want to pass in a `Path` object instead of a string.

### Implementation

The type hints will change from `str` to `Union[str, bytes, os.PathLike]` according to [PEP 519](https://peps.python.org/pep-0519/#provide-specific-type-hinting-support).

### Alternatives

_No response_

### Additional information

_No response_

pioneerHitesh:
Added ``pathlib`` (path object) support for the entire ``dataset module``

Closes #1616

### DIFF
--- a/tests/datasets/test_advance.py
+++ b/tests/datasets/test_advance.py
@@ -33,7 +33,7 @@ class TestADVANCE:
         md5s = ["43acacecebecd17a82bc2c1e719fd7e4", "039b7baa47879a8a4e32b9dd8287f6ad"]
         monkeypatch.setattr(ADVANCE, "urls", urls)
         monkeypatch.setattr(ADVANCE, "md5s", md5s)
-        root = str(tmp_path)
+        root = tmp_path
         transforms = nn.Identity()
         return ADVANCE(root, transforms, download=True, checksum=True)
 
@@ -69,7 +69,7 @@ class TestADVANCE:
 
     def test_not_downloaded(self, tmp_path: Path) -> None:
         with pytest.raises(RuntimeError, match="Dataset not found or corrupted."):
-            ADVANCE(str(tmp_path))
+            ADVANCE(tmp_path)
 
     def test_mock_missing_module(
         self, dataset: ADVANCE, mock_missing_module: None
--- a/tests/datasets/test_advance.py
+++ b/tests/datasets/test_advance.py
@@ -4,6 +4,7 @@
 import builtins
 import os
 import shutil
+from pathlib import Path
 from typing import Any
 
 import matplotlib.pyplot as plt
@@ -14,7 +15,6 @@ from pytest import MonkeyPatch
 
 import torchgeo.datasets.utils
 from torchgeo.datasets import ADVANCE
-from torchgeo.datasets.utils import Path
 
 
 def download_url(url: str, root: str, *args: str) -> None:
--- a/tests/datasets/test_advance.py
+++ b/tests/datasets/test_advance.py
@@ -4,7 +4,6 @@
 import builtins
 import os
 import shutil
-from pathlib import Path
 from typing import Any
 
 import matplotlib.pyplot as plt
@@ -15,6 +14,7 @@ from pytest import MonkeyPatch
 
 import torchgeo.datasets.utils
 from torchgeo.datasets import ADVANCE
+from torchgeo.datasets.utils import Path
 
 
 def download_url(url: str, root: str, *args: str) -> None:
--- a/tests/datasets/test_eudem.py
+++ b/tests/datasets/test_eudem.py
@@ -3,6 +3,7 @@
 
 import os
 import shutil
+from pathlib import Path
 
 import matplotlib.pyplot as plt
 import pytest
@@ -12,7 +13,6 @@ from pytest import MonkeyPatch
 from rasterio.crs import CRS
 
 from torchgeo.datasets import EUDEM, BoundingBox, IntersectionDataset, UnionDataset
-from torchgeo.datasets.utils import Path
 
 
 class TestEUDEM:
@@ -21,7 +21,7 @@ class TestEUDEM:
         md5s = {"eu_dem_v11_E30N10.zip": "ef148466c02197a08be169eaad186591"}
         monkeypatch.setattr(EUDEM, "md5s", md5s)
         zipfile = os.path.join("tests", "data", "eudem", "eu_dem_v11_E30N10.zip")
-        shutil.copy(zipfile, str(tmp_path))
+        shutil.copy(zipfile, tmp_path)
         root = str(tmp_path)
         transforms = nn.Identity()
         return EUDEM(root, transforms=transforms)
@@ -45,7 +45,7 @@ class TestEUDEM:
             EUDEM(str(tmp_path))
 
     def test_corrupted(self, tmp_path: Path) -> None:
-        with open(os.path.join(str(tmp_path), "eu_dem_v11_E30N10.zip"), "w") as f:
+        with open(os.path.join(tmp_path, "eu_dem_v11_E30N10.zip"), "w") as f:
             f.write("bad")
         with pytest.raises(RuntimeError, match="Dataset found, but corrupted."):
             EUDEM(str(tmp_path), checksum=True)
--- a/tests/datasets/test_eudem.py
+++ b/tests/datasets/test_eudem.py
@@ -22,7 +22,7 @@ class TestEUDEM:
         monkeypatch.setattr(EUDEM, "md5s", md5s)
         zipfile = os.path.join("tests", "data", "eudem", "eu_dem_v11_E30N10.zip")
         shutil.copy(zipfile, tmp_path)
-        root = str(tmp_path)
+        root = tmp_path
         transforms = nn.Identity()
         return EUDEM(root, transforms=transforms)
 
@@ -42,13 +42,13 @@ class TestEUDEM:
         shutil.rmtree(tmp_path)
         os.makedirs(tmp_path)
         with pytest.raises(RuntimeError, match="Dataset not found in"):
-            EUDEM(str(tmp_path))
+            EUDEM(tmp_path)
 
     def test_corrupted(self, tmp_path: Path) -> None:
         with open(os.path.join(tmp_path, "eu_dem_v11_E30N10.zip"), "w") as f:
             f.write("bad")
         with pytest.raises(RuntimeError, match="Dataset found, but corrupted."):
-            EUDEM(str(tmp_path), checksum=True)
+            EUDEM(tmp_path, checksum=True)
 
     def test_and(self, dataset: EUDEM) -> None:
         ds = dataset & dataset
--- a/tests/datasets/test_eudem.py
+++ b/tests/datasets/test_eudem.py
@@ -13,6 +13,7 @@ from pytest import MonkeyPatch
 from rasterio.crs import CRS
 
 from torchgeo.datasets import EUDEM, BoundingBox, IntersectionDataset, UnionDataset
+from torchgeo.datasets.utils import check_instance_type
 
 
 class TestEUDEM:
@@ -33,9 +34,9 @@ class TestEUDEM:
         assert isinstance(x["mask"], torch.Tensor)
 
     def test_extracted_already(self, dataset: EUDEM) -> None:
-        assert isinstance(dataset.paths, str)
-        zipfile = os.path.join(dataset.paths, "eu_dem_v11_E30N10.zip")
-        shutil.unpack_archive(zipfile, dataset.paths, "zip")
+        assert check_instance_type(dataset.paths)
+        zipfile = os.path.join(str(dataset.paths), "eu_dem_v11_E30N10.zip")
+        shutil.unpack_archive(zipfile, str(dataset.paths), "zip")
         EUDEM(dataset.paths)
 
     def test_no_dataset(self, tmp_path: Path) -> None:
--- a/tests/datasets/test_eudem.py
+++ b/tests/datasets/test_eudem.py
@@ -3,7 +3,6 @@
 
 import os
 import shutil
-from pathlib import Path
 
 import matplotlib.pyplot as plt
 import pytest
@@ -13,6 +12,7 @@ from pytest import MonkeyPatch
 from rasterio.crs import CRS
 
 from torchgeo.datasets import EUDEM, BoundingBox, IntersectionDataset, UnionDataset
+from torchgeo.datasets.utils import Path
 
 
 class TestEUDEM:
@@ -21,7 +21,7 @@ class TestEUDEM:
         md5s = {"eu_dem_v11_E30N10.zip": "ef148466c02197a08be169eaad186591"}
         monkeypatch.setattr(EUDEM, "md5s", md5s)
         zipfile = os.path.join("tests", "data", "eudem", "eu_dem_v11_E30N10.zip")
-        shutil.copy(zipfile, tmp_path)
+        shutil.copy(zipfile, str(tmp_path))
         root = str(tmp_path)
         transforms = nn.Identity()
         return EUDEM(root, transforms=transforms)
@@ -45,7 +45,7 @@ class TestEUDEM:
             EUDEM(str(tmp_path))
 
     def test_corrupted(self, tmp_path: Path) -> None:
-        with open(os.path.join(tmp_path, "eu_dem_v11_E30N10.zip"), "w") as f:
+        with open(os.path.join(str(tmp_path), "eu_dem_v11_E30N10.zip"), "w") as f:
             f.write("bad")
         with pytest.raises(RuntimeError, match="Dataset found, but corrupted."):
             EUDEM(str(tmp_path), checksum=True)
--- a/tests/datasets/test_landcoverai.py
+++ b/tests/datasets/test_landcoverai.py
@@ -3,6 +3,7 @@
 
 import os
 import shutil
+from pathlib import Path
 
 import matplotlib.pyplot as plt
 import pytest
@@ -14,7 +15,6 @@ from torch.utils.data import ConcatDataset
 
 import torchgeo.datasets.utils
 from torchgeo.datasets import BoundingBox, LandCoverAI, LandCoverAIGeo
-from torchgeo.datasets.utils import Path
 
 
 def download_url(url: str, root: str, *args: str, **kwargs: str) -> None:
--- a/tests/datasets/test_landcoverai.py
+++ b/tests/datasets/test_landcoverai.py
@@ -3,7 +3,6 @@
 
 import os
 import shutil
-from pathlib import Path
 
 import matplotlib.pyplot as plt
 import pytest
@@ -15,6 +14,7 @@ from torch.utils.data import ConcatDataset
 
 import torchgeo.datasets.utils
 from torchgeo.datasets import BoundingBox, LandCoverAI, LandCoverAIGeo
+from torchgeo.datasets.utils import Path
 
 
 def download_url(url: str, root: str, *args: str, **kwargs: str) -> None:
--- a/tests/datasets/test_landcoverai.py
+++ b/tests/datasets/test_landcoverai.py
@@ -29,7 +29,7 @@ class TestLandCoverAIGeo:
         monkeypatch.setattr(LandCoverAIGeo, "md5", md5)
         url = os.path.join("tests", "data", "landcoverai", "landcover.ai.v1.zip")
         monkeypatch.setattr(LandCoverAIGeo, "url", url)
-        root = str(tmp_path)
+        root = tmp_path
         transforms = nn.Identity()
         return LandCoverAIGeo(root, transforms=transforms, download=True, checksum=True)
 
@@ -44,13 +44,13 @@ class TestLandCoverAIGeo:
 
     def test_already_downloaded(self, monkeypatch: MonkeyPatch, tmp_path: Path) -> None:
         url = os.path.join("tests", "data", "landcoverai", "landcover.ai.v1.zip")
-        root = str(tmp_path)
+        root = tmp_path
         shutil.copy(url, root)
-        LandCoverAIGeo(root)
+        LandCoverAIGeo(str(root))
 
     def test_not_downloaded(self, tmp_path: Path) -> None:
         with pytest.raises(RuntimeError, match="Dataset not found"):
-            LandCoverAIGeo(str(tmp_path))
+            LandCoverAIGeo(tmp_path)
 
     def test_out_of_bounds_query(self, dataset: LandCoverAIGeo) -> None:
         query = BoundingBox(0, 0, 0, 0, 0, 0)
@@ -83,7 +83,7 @@ class TestLandCoverAI:
         monkeypatch.setattr(LandCoverAI, "url", url)
         sha256 = "ecec8e871faf1bbd8ca525ca95ddc1c1f5213f40afb94599884bd85f990ebd6b"
         monkeypatch.setattr(LandCoverAI, "sha256", sha256)
-        root = str(tmp_path)
+        root = tmp_path
         split = request.param
         transforms = nn.Identity()
         return LandCoverAI(root, split, transforms, download=True, checksum=True)
@@ -110,13 +110,13 @@ class TestLandCoverAI:
         sha256 = "ecec8e871faf1bbd8ca525ca95ddc1c1f5213f40afb94599884bd85f990ebd6b"
         monkeypatch.setattr(LandCoverAI, "sha256", sha256)
         url = os.path.join("tests", "data", "landcoverai", "landcover.ai.v1.zip")
-        root = str(tmp_path)
+        root = tmp_path
         shutil.copy(url, root)
         LandCoverAI(root)
 
     def test_not_downloaded(self, tmp_path: Path) -> None:
         with pytest.raises(RuntimeError, match="Dataset not found"):
-            LandCoverAI(str(tmp_path))
+            LandCoverAI(tmp_path)
 
     def test_invalid_split(self) -> None:
         with pytest.raises(AssertionError):
--- a/tests/datasets/test_mapinwild.py
+++ b/tests/datasets/test_mapinwild.py
@@ -4,6 +4,7 @@
 import glob
 import os
 import shutil
+from pathlib import Path
 
 import matplotlib.pyplot as plt
 import pytest
@@ -15,7 +16,6 @@ from torch.utils.data import ConcatDataset
 
 import torchgeo.datasets.utils
 from torchgeo.datasets import MapInWild
-from torchgeo.datasets.utils import Path
 
 
 def download_url(url: str, root: str, *args: str, **kwargs: str) -> None:
@@ -118,7 +118,7 @@ class TestMapInWild:
             "tests", "data", "mapinwild", "split_IDs", "split_IDs.csv"
         )
         shutil.copy(splitfile, root)
-        with open(os.path.join(str(tmp_path), "mask.zip"), "w") as f:
+        with open(os.path.join(tmp_path, "mask.zip"), "w") as f:
             f.write("bad")
         with pytest.raises(RuntimeError, match="Dataset found, but corrupted."):
             MapInWild(root=str(tmp_path), download=True, checksum=True)
--- a/tests/datasets/test_mapinwild.py
+++ b/tests/datasets/test_mapinwild.py
@@ -4,7 +4,6 @@
 import glob
 import os
 import shutil
-from pathlib import Path
 
 import matplotlib.pyplot as plt
 import pytest
@@ -16,6 +15,7 @@ from torch.utils.data import ConcatDataset
 
 import torchgeo.datasets.utils
 from torchgeo.datasets import MapInWild
+from torchgeo.datasets.utils import Path
 
 
 def download_url(url: str, root: str, *args: str, **kwargs: str) -> None:
@@ -118,7 +118,7 @@ class TestMapInWild:
             "tests", "data", "mapinwild", "split_IDs", "split_IDs.csv"
         )
         shutil.copy(splitfile, root)
-        with open(os.path.join(tmp_path, "mask.zip"), "w") as f:
+        with open(os.path.join(str(tmp_path), "mask.zip"), "w") as f:
             f.write("bad")
         with pytest.raises(RuntimeError, match="Dataset found, but corrupted."):
             MapInWild(root=str(tmp_path), download=True, checksum=True)
--- a/tests/datasets/test_mapinwild.py
+++ b/tests/datasets/test_mapinwild.py
@@ -53,7 +53,7 @@ class TestMapInWild:
         urls = os.path.join("tests", "data", "mapinwild")
         monkeypatch.setattr(MapInWild, "url", urls)
 
-        root = str(tmp_path)
+        root = tmp_path
         split = request.param
 
         transforms = nn.Identity()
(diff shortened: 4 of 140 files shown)


pioneerHitesh:
> @microsoft-github-policy-service agree

torchgeo/datasets/agb_live_woody_density.py
```diff
@@ -59,7 +59,7 @@ class AbovegroundLiveWoodyBiomassDensity(RasterDataset):
 
     def __init__(
         self,
-        paths: Union[str, Iterable[str]] = "data",
+        paths: Union[str, Iterable[str],os.PathLike] = "data",
```
adamjstewart, line 62: Can also be iterable[pathlike]

torchgeo/datasets/agb_live_woody_density.py
```diff
@@ -115,6 +115,10 @@ def _verify(self) -> None:
 
     def _download(self) -> None:
         """Download the dataset."""
+
+        if(isinstance(self.paths,os.PathLike)):
+            self.paths = str(self.paths)
```
adamjstewart, line 120: Is it possible to make them work if we keep them as pathlib instead of casting them?

torchgeo/datasets/geo.py
```diff
@@ -289,6 +289,8 @@ def files(self) -> set[str]:
         # Make iterable
         if isinstance(self.paths, str):
             paths: Iterable[str] = [self.paths]
+        elif isinstance(self.paths, os.PathLike):
+            paths: Iterable[os.PathLike] = [self.paths]
```
adamjstewart, line 293: Can probably keep this to a single if-statement. Either check both or check "not iterable"

torchgeo/datasets/agb_live_woody_density.py
pioneerHitesh: Unfortunately, it isn't possible to keep it as pathlib because of the method calling `` download_url(self.url, self.paths, self.base_filename)`` which is a torchvision method and only accepts strings.

torchgeo/datasets/geo.py
pioneerHitesh: Yes, you are correct sir, will implement this.

adamjstewart:
@pioneerHitesh can you open a separate issue for the download problems? Maybe the URL changed.

torchgeo/datasets/geo.py
```diff
@@ -72,7 +72,7 @@ class GeoDataset(Dataset[dict[str, Any]], abc.ABC):
        dataset = landsat7 | landsat8
     """
 
-    paths: Union[str, Iterable[str]]
+    paths: Union[str, Iterable[str], os.PathLike, Iterable[os.PathLike]]
```
adamjstewart, line 75: Let's add a type alias to make things simpler. In `torchgeo/datasets/utils.py`, you can add:
```python
import os
from typing import Union, TypeAlias

Path: TypeAlias = Union[str, bytes, os.PathLike[str], os.PathLike[bytes]]
```
Then in all other files you can simply replace `str` with `Path` where needed. This will be less code duplication.

pioneerHitesh: Thank you for your feedback, I will work towards implementing this.

torchgeo/datasets/utils.py
```diff
@@ -17,15 +17,23 @@
...
 from torchvision.datasets.utils import check_integrity, download_url
 from torchvision.utils import draw_segmentation_masks
 
+Path: TypeAlias = Union[
+    str,
+    bytes,
+    os.PathLike[str],
+    os.PathLike[bytes],
+    Iterable[os.PathLike[str]],
+    Iterable[os.PathLike[bytes]],
```
adamjstewart, line 35: Iterables shouldn't be included, most of the time they aren't supported

pioneerHitesh: oh! Thank you sir

pioneerHitesh: Sir you have mentioned to add them in this [comment](https://github.com/microsoft/torchgeo/pull/1678#discussion_r1363698094)?

torchgeo/datasets/agb_live_woody_density.py
```diff
@@ -59,7 +58,7 @@ class AbovegroundLiveWoodyBiomassDensity(RasterDataset):
 
     def __init__(
         self,
-        paths: Union[str, Iterable[str]] = "data",
+        paths: Path = "data",
```
adamjstewart, line 61: Path or Iterable[Path]

torchgeo/datasets/agb_live_woody_density.py
```diff
@@ -5,15 +5,14 @@
 
 import json
 import os
-from collections.abc import Iterable
```
adamjstewart, line 8: `typing.Iterable` is deprecated, should use `collections.abc.Iterable` instead: https://docs.python.org/3/library/typing.html#typing.Iterable

torchgeo/datasets/agb_live_woody_density.py
```diff
@@ -105,7 +104,7 @@ def _verify(self) -> None:
         # Check if the user requested to download the dataset
         if not self.download:
             raise RuntimeError(
-                f"Dataset not found in `root={self.paths}` and `download=False`, "
+                f"Dataset not found in `root={str(self.paths)}` and `download=False`, "
```
adamjstewart, line 107: This should automatically be cast to a string during formatting, were you seeing an error?

torchgeo/datasets/geo.py
```diff
@@ -748,7 +756,7 @@ class NonGeoClassificationDataset(NonGeoDataset, ImageFolder):  # type: ignore[m
 
     def __init__(
         self,
-        root: str = "data",
+        root: Union[Path, Iterable[Path]] = "data",
```
adamjstewart, line 759: Iterables aren't supported here, only Paths

torchgeo/datasets/utils.py
```diff
@@ -17,15 +17,17 @@
...
 
 import numpy as np
 import rasterio
 import torch
 from torch import Tensor
 from torchvision.datasets.utils import check_integrity, download_url
 from torchvision.utils import draw_segmentation_masks
+from typing_extensions import TypeAlias
 
+Path: TypeAlias = Union[str, bytes, os.PathLike[str], os.PathLike[bytes]]
```
adamjstewart, line 30: This should come after `__all__`

torchgeo/datasets/utils.py
```diff
@@ -737,3 +739,15 @@ def percentile_normalization(
         (img - lower_percentile) / (upper_percentile - lower_percentile + 1e-5), 0, 1
     )
     return img_normalized
+
+
+def check_instance_type(objects: Union[Path, Iterable[Path]]) -> bool:
```
adamjstewart, line 744: Why is this needed?

torchgeo/datasets/agb_live_woody_density.py
pioneerHitesh: Yes , The error occurs specifically for byte. ``error: On Python 3 '{}'.format(b'abc') produces "b'abc'", not 'abc'; use '{!r}'.format(b'abc') if this is desired behavior when using formatted string literals ``.

torchgeo/datasets/utils.py
pioneerHitesh: This function help to check for parameterized generic types such as ``PathLike[bytes]`` , the original isinstance fails to check parameterized generic types. Here is the the error ``Second argument to "isinstance" must be a class or tuple of classes TypeVar or generic type with type arguments not allowed``.

torchgeo/datasets/agb_live_woody_density.py
adamjstewart: How do I reproduce this error? It works fine for me:
```pycon
> python3 -Werror
Python 3.11.6 (main, Oct 11 2023, 22:49:14) [Clang 15.0.0 (clang-1500.0.40.1)] on darwin
>>> foo = b"foo"
>>> f"{foo} bar"
"b'foo' bar"
```

torchgeo/datasets/geo.py
```diff
@@ -574,7 +581,7 @@ class VectorDataset(GeoDataset):
 
     def __init__(
         self,
-        paths: Union[str, Iterable[str]] = "data",
+        paths: Path = "data",
```
adamjstewart, line 584:
```suggestion
        paths: Union[Path, Iterable[Path]] = "data",
```

torchgeo/datasets/agb_live_woody_density.py
pioneerHitesh: mypy detects this an error. Here is the code and output
code:
``mypy torchgeo/datasets/agb_live_woody_density.py``
output
``torchgeo/datasets/agb_live_woody_density.py:108: error: If x = b'abc' then f"{x}" or "{}".format(x) produces "b'abc'", not "abc". If this is desired behavior, use f"{x!r}" or "{!r}".format(x). Otherwise, decode the bytes  [str-bytes-safe]
Found 3 errors in 2 files (checked 1 source file)``

torchgeo/datasets/utils.py
adamjstewart: It doesn't have a docstring, but I'm guessing this function returns True if the input is a pathlib.Path or an iterable of pathlib.Path? But then you always cast to str, which doesn't make sense for iterables. Also, `str(object)` will always be an instance of str, so that check doesn't make sense. Can't we just check if it's an instance of `pathlib.Path`?

adamjstewart:
Some of the type ignores are for files that have not yet been updated (this is fine), but some of the type ignores are for files that have been updated (this makes me think there's a bug in the type hints).

pioneerHitesh:
> Some of the type ignores are for files that have not yet been updated (this is fine), but some of the type ignores are for files that have been updated (this makes me think there's a bug in the type hints).
Will investigate this and get back to you.

torchgeo/datasets/agb_live_woody_density.py
adamjstewart: Can we use `{self.paths!r}` or `{self.paths!s}` instead of `{str(self.paths)}`?

torchgeo/datasets/utils.py
pioneerHitesh: if we try to check if the parameter is an instance of pathlib.Path using ``isinstance`` method , Here are the results.

 Code
``isinstance(self.paths, Path)``
Errors:
```plaintext
torchgeo/datasets/geo.py:297: error: Parameterized generics cannot be used with class or instance checks  [misc]
torchgeo/datasets/geo.py:297: error: Argument 2 to "isinstance" has incompatible type "<typing special form>"; 
expected "_ClassInfo"  [arg-type]
```

The cast to str is required to find out if the path's type is str or bytes. Here is an example 
```from os import PathLike
from pathlib import Path
dir_path = Path("/home")
print(isinstance(dir_path,str))  # This will return false 
print(hasattr(dir_path,'decode'))
```

As you have mentioned str(object) will be an instance of str , so that check does not make sense and will incorrectly convert objects to string. Here is what i mean.

```from pathlib import Path
dir_path = Path.cwd()  # return type is <class 'pathlib.PosixPath'>
dir_path = Path("/home")
print(isinstance(str(dir_path), str))  # returns true
```

A better way of checking for strings would be using the ``hasattr(str_to_check, "encode")``.

adamjstewart: I'm guessing that error is because you're using:
```
from pathlib import Path
```
AND
```
from .utils import Path
```
and the names conflict. Not sure if we should use a different name or just use `pathlib.Path`.

Do we need to check whether or not it's bytes? Can we treat str and bytes identically?

pioneerHitesh: No, sir both of them have been tested in seperate files. `from pathlib import Path` and `from .utils import Path` are not imported simultaneously.  We can skip the test if desired so.

adamjstewart: I have the following file:
```python
from pathlib import Path

p = Path('.')

if isinstance(p, Path):
    print("Path")
else:
    print("str or bytes")
```
`mypy --strict` passes on this file. Anything I'm missing here?

I don't think you need to check `Iterable[Path]`, just `Path`.

pioneerHitesh: It throws the same error when i try to run the below code.
```import os
from typing import Union
from typing_extensions import TypeAlias

Path: TypeAlias = Union[str, bytes, os.PathLike[bytes], os.PathLike[str]]

testPath: Path = os.getcwd()

print(isinstance(testPath, Path))```

Comment (adamjstewart) [reply torchgeo/datasets/utils.py]: Yes, because `torchgeo.datasets.utils.Path` and `pathlib.Path` are completely different things.

File: torchgeo/datasets/agb_live_woody_density.py (hunk 8 above)
Comment (pioneerHitesh) [reply torchgeo/datasets/agb_live_woody_density.py:108]: Yes, I have implemented this fix.

File: tests/datasets/test_advance.py [hunk 13]
```diff
@@ -15,6 +14,7 @@
 
 import torchgeo.datasets.utils
 from torchgeo.datasets import ADVANCE
+from torchgeo.datasets.utils import Path
```
adamjstewart, line 17: This is incorrect, the correct type is `pathlib.Path`

pioneerHitesh: Have replaced it with Path , so that it stays consistent with the codebase.

adamjstewart: `torchgeo.datasets.utils.Path` and `pathlib.Path` are two different things. The latter is a subset of the former. We should be as specific as possible.

pioneerHitesh: @adamjstewart ok will revert these changes.

adamjstewart:
Can you replace all instances of `str(tmp_path)` with `tmp_path` in our tests? Should be possible with a 1-line sed command. Then we can actually test to ensure that `pathlib.Path` works correctly.

pioneerHitesh:
> Can you replace all instances of `str(tmp_path)` with `tmp_path` in our tests? Should be possible with a 1-line sed command. Then we can actually test to ensure that `pathlib.Path` works correctly.

@adamjstewart  Unfortunately, it not completely possible to replace ``str(tmp_path)`` because ``download_url()`` method does not support path but i am trying to reduce ``str(tmp_path)``.

pioneerHitesh:
@adamjstewart any additional changes ?

tests/datasets/test_eudem.py
```diff
@@ -33,22 +34,22 @@ def test_getitem(self, dataset: EUDEM) -> None:
         assert isinstance(x["mask"], torch.Tensor)
 
     def test_extracted_already(self, dataset: EUDEM) -> None:
-        assert isinstance(dataset.paths, str)
-        zipfile = os.path.join(dataset.paths, "eu_dem_v11_E30N10.zip")
-        shutil.unpack_archive(zipfile, dataset.paths, "zip")
+        assert check_instance_type(dataset.paths)
```
adamjstewart, line 37: I think we can just assert that this is a `pathlib.Path`

tests/datasets/test_landcoverai.py
```diff
@@ -44,13 +44,13 @@ def test_already_extracted(self, dataset: LandCoverAIGeo) -> None:
 
     def test_already_downloaded(self, monkeypatch: MonkeyPatch, tmp_path: Path) -> None:
         url = os.path.join("tests", "data", "landcoverai", "landcover.ai.v1.zip")
-        root = str(tmp_path)
+        root = tmp_path
         shutil.copy(url, root)
-        LandCoverAIGeo(root)
+        LandCoverAIGeo(str(root))
```
adamjstewart, line 49: Why does this need to be converted to a string?

tests/datasets/test_mapinwild.py
```diff
@@ -98,20 +98,20 @@ def test_invalid_split(self) -> None:
...
 
     def test_downloaded_not_extracted(self, tmp_path: Path) -> None:
         pathname = os.path.join("tests", "data", "mapinwild", "*", "*")
         pathname_glob = glob.glob(pathname)
-        root = str(tmp_path)
+        root = tmp_path
         for zipfile in pathname_glob:
             shutil.copy(zipfile, root)
-        MapInWild(root, download=False)
+        MapInWild(str(root), download=False)
```
adamjstewart, line 109: Why str?

torchgeo/datasets/agb_live_woody_density.py
```diff
@@ -105,7 +105,7 @@ def _verify(self) -> None:
         # Check if the user requested to download the dataset
         if not self.download:
             raise RuntimeError(
-                f"Dataset not found in `root={self.paths}` and `download=False`, "
+                f"Dataset not found in `root={self.paths!r}` and `download=False`, "
```
adamjstewart, line 108:
```suggestion
                f"Dataset not found in `paths={self.paths!r}` and `download=False`, "
```
We forgot to update this in the root -> paths PR, may need to be fixed in multiple files

torchgeo/datasets/astergdem.py
```diff
@@ -88,7 +90,7 @@ def _verify(self) -> None:
             return
 
         raise RuntimeError(
-            f"Dataset not found in `root={self.paths}` "
+            f"Dataset not found in `root={self.paths!r}` "
```
adamjstewart, line 93:
```suggestion
            f"Dataset not found in `paths={self.paths!r}` "
```

torchgeo/datasets/benin_cashews.py
```diff
@@ -202,7 +207,7 @@ def __init__(
         """
         self._validate_bands(bands)
 
-        self.root = os.path.expanduser(root)
+        self.root = str(os.path.expanduser(root))
```
adamjstewart, line 210: For uniformity we prob shouldn't use `os.path.expanduser` in any dataset

torchgeo/datasets/cbf.py
```diff
@@ -108,9 +113,9 @@ def _check_integrity(self) -> bool:
         Returns:
             True if dataset files are found and/or MD5s match, else False
         """
-        assert isinstance(self.paths, str)
+        assert check_instance_type(self.paths)
```
adamjstewart, line 116: Seems like this dataset doesn't actually support iterables. If we change the type hints to just `Path` does mypy complain?

torchgeo/datasets/cdl.py
```diff
@@ -314,7 +314,7 @@ def _verify(self) -> None:
         # Check if the user requested to download the dataset
         if not self.download:
             raise RuntimeError(
-                f"Dataset not found in `root={self.paths}` and `download=False`, "
+                f"Dataset not found in `root={self.paths!r}` and `download=False`, "
```
adamjstewart, line 317:
```suggestion
                f"Dataset not found in `paths={self.paths!r}` and `download=False`, "
```

torchgeo/datasets/cms_mangrove_canopy.py
```diff
@@ -235,24 +236,24 @@ def _verify(self) -> None:
...
+        pathname = os.path.join(str(self.paths), self.zipfile)
         if os.path.exists(pathname):
             if self.checksum and not check_integrity(pathname, self.md5):
                 raise RuntimeError("Dataset found, but corrupted.")
             self._extract()
             return
 
         raise RuntimeError(
-            f"Dataset not found in `root={self.paths}` "
+            f"Dataset not found in `root={self.paths!r}` "
```
adamjstewart, line 248:
```suggestion
            f"Dataset not found in `paths={self.paths!r}` "
```

torchgeo/datasets/geo.py
```diff
@@ -287,16 +294,17 @@ def files(self) -> set[str]:
...
         """
         # Make iterable
-        if isinstance(self.paths, str):
-            paths: Iterable[str] = [self.paths]
+        paths: Iterable[Path]
+        if check_instance_type(self.paths):
+            paths = [self.paths]  # type: ignore[list-item]
         else:
-            paths = self.paths
+            paths = self.paths  # type: ignore[assignment]
```
adamjstewart, line 301: Something is definitely wrong here...

torchgeo/datasets/geo.py
```diff
@@ -392,7 +400,7 @@ def __init__(
         i = 0
         filename_regex = re.compile(self.filename_regex, re.VERBOSE)
         for filepath in self.files:
-            match = re.match(filename_regex, os.path.basename(filepath))
+            match = re.match(filename_regex, os.path.basename(str(filepath)))
```
adamjstewart, line 403: I don't think this needs to be cast to a str, right?

torchgeo/datasets/geo.py
```diff
@@ -432,7 +440,6 @@ def __init__(
             if self.bands:
                 msg += f" with `bands={self.bands}`"
             raise FileNotFoundError(msg)
-
```
adamjstewart, line 435: Can you add this blank line back?

torchgeo/datasets/geo.py
```diff
@@ -629,7 +636,8 @@ def __init__(
                 i += 1
 
         if i == 0:
-            msg = f"No {self.__class__.__name__} data was found in `root='{paths}'`"
+            str_path = str(paths)
+            msg = f"No {self.__class__.__name__} data was found in `root='{str_path}'`"
```
adamjstewart, line 640:
```suggestion
            msg = f"No {self.__class__.__name__} data was found in `paths={paths!r}`"
```

torchgeo/datasets/inaturalist.py
```diff
@@ -44,9 +44,9 @@ def __init__(self, root: str = "data") -> None:
         """
         super().__init__()
 
-        self.root = root
+        self.root = str(root)
 
-        files = glob.glob(os.path.join(root, "**.csv"))
+        files = glob.glob(os.path.join(self.root, "**.csv"))
```
adamjstewart, line 49: This doesn't need to be a str, right?

torchgeo/datasets/nlcd.py
```diff
@@ -209,7 +209,7 @@ def _verify(self) -> None:
         # Check if the user requested to download the dataset
         if not self.download:
             raise RuntimeError(
-                f"Dataset not found in `root={self.paths}` and `download=False`, "
+                f"Dataset not found in `root={str(self.paths)}` and `download=False`, "
```
adamjstewart, line 212:
```suggestion
                f"Dataset not found in `paths={self.paths!r}` and `download=False`, "
```

torchgeo/datasets/openbuildings.py
```diff
@@ -278,14 +278,14 @@ def __init__(
...
-                self.paths, feature["properties"]["tile_url"].split("/")[-1]
+                str(self.paths), feature["properties"]["tile_url"].split("/")[-1]
             )
             self.index.insert(i, coords, filepath)
             i += 1
 
         if i == 0:
             raise FileNotFoundError(
-                f"No {self.__class__.__name__} data was found in '{self.paths}'"
+                f"No {self.__class__.__name__} data was found in '{self.paths!r}'"
```
adamjstewart, line 288:
```suggestion
                f"No {self.__class__.__name__} data was found in `paths={self.paths!r}`"
```

torchgeo/datasets/utils.py
```diff
@@ -45,6 +46,8 @@
     "percentile_normalization",
 )
 
+Path: TypeAlias = Union[str, bytes, os.PathLike[bytes], os.PathLike[str]]
```
adamjstewart, line 49:
```suggestion
Path: TypeAlias = Union[str, bytes, os.PathLike[str], os.PathLike[bytes]]
```
Just to keep the order consistent

torchgeo/datasets/utils.py
```diff
@@ -17,14 +17,15 @@
...
-from typing import Any, cast, overload
+from typing import Any, Union, cast, overload
 
 import numpy as np
 import rasterio
 import torch
 from torch import Tensor
 from torchvision.datasets.utils import check_integrity, download_url
 from torchvision.utils import draw_segmentation_masks
+from typing_extensions import TypeAlias
```
adamjstewart, line 28: We don't have a dep on typing_extensions right now. Let's just remove the `TypeAlias` usage, we don't actually need it.

torchgeo/datasets/utils.py
```diff
@@ -737,3 +740,25 @@ def percentile_normalization(
         (img - lower_percentile) / (upper_percentile - lower_percentile + 1e-5), 0, 1
     )
     return img_normalized
+
+
+def check_instance_type(paths: Path | Iterable[Path]) -> bool:
```
adamjstewart, line 745: I don't love this function:

* The name is way too vague
* The docstring doesn't tell me when it's True and when it's False
* Do we really need this complex of a check? 

It seems like we're really just checking if `paths` is a `Path` (True) or `Iterable[Path]` (False). Can't we just do:
```python
return not isinstance(paths, Iterable)
```
? In that case, we wouldn't even need a function. Basically, if we can figure out a reasonable way to remove this function, I would be happier.

adamjstewart (commented):
At the moment, it seems like 99% of the time, we accept any Path as input, then immediately convert it to a str. I think this is still useful, but it's a lot of extra work to ensure valid types and cast everything. Is torchvision's download_url the only reason we need to cast to a str? I wonder if we could just update download_url to accept any Path as input. I'm willing to submit a PR to torchvision if you think that's worth it. We could even have a wrapper in TorchGeo that just converts the path to a str and calls the torchvision version. Let me know what you think of this idea.

Also, this PR is becoming massive. Is there any way we can split this up into multiple PRs to make it easier on both of us to update/review? Maybe 1 PR to remove `os.path.expanduser`, another PR to fix `root={self.paths}` to `paths={self.paths!r}`, another PR to add support to the datasets, and another PR to update the tests? Or would that be too much work for you?

torchgeo/datasets/agb_live_woody_density.py
pioneerHitesh: Noted, will look into this.

torchgeo/datasets/utils.py
pioneerHitesh: @adamjstewart Unfortunately ``isinstance(paths, Iterable)`` will also return ``True`` for ``bytes and str``. will have to find another way to eliminate this function.

pioneerHitesh:
> At the moment, it seems like 99% of the time, we accept any Path as input, then immediately convert it to a str. I think this is still useful, but it's a lot of extra work to ensure valid types and cast everything. Is torchvision's download_url the only reason we need to cast to a str? I wonder if we could just update download_url to accept any Path as input. I'm willing to submit a PR to torchvision if you think that's worth it. We could even have a wrapper in TorchGeo that just converts the path to a str and calls the torchvision version. Let me know what you think of this idea.
> 
> Also, this PR is becoming massive. Is there any way we can split this up into multiple PRs to make it easier on both of us to update/review? Maybe 1 PR to remove `os.path.expanduser`, another PR to fix `root={self.paths}` to `paths={self.paths!r}`, another PR to add support to the datasets, and another PR to update the tests? Or would that be too much work for you?

In addition to the torchvision's ``download_url`` , the ``os.path.join`` also needs ``str``. If we can drop support for Path being of type ``bytes`` and ``pathlib.Path[bytes]`` then we do not need to cast to ``str``. The problem with ``bytes`` or ``pathlib.Path[bytes]`` is if the second argument passed in the ``join`` method is a string it is unable to combine the paths. It also does not support merging of ``pathlib.Path[bytes]`` and ``bytes``.  Another issue i have discovered is ``pathlib.path`` does not support ``bytes``. So we will have to drop support for atleast ``os.PathLike[bytes]`` and if feasible also ``bytes``. 

My suggestion would be that we add support for ``os.PathLike[str]`` in torchvision's ``download_url`` as that would benefit others as well and will eliminate the need for writing a wrapper function just to cast to ``str``. I concur with you regarding the PR becoming unviable to update/review and will split the PR. However we cannot split the ``datasets`` and ``tests`` separately as that would lead to failing of tests instead we could split ``datasets`` and ``tests`` by limiting the number of ``datasets`` and ``tests`` files changed in a commit .

tests/datasets/test_mapinwild.py
pioneerHitesh: This was probably added earlier to fix mypy error when ``MapInWild`` did not support ``pathlib.path``. Have fixed this and gone through other test files to fix similar issues.

tests/datasets/test_landcoverai.py
pioneerHitesh: This was probably added earlier to fix mypy error when ``LandCoverAIGeo`` did not support ``pathlib.path``. Have fixed this and gone through other test files to fix similar issues.

tests/datasets/test_eudem.py
pioneerHitesh: @adamjstewart  We may not be able to do that because we have to assert for ``str``,``bytes`` , ``os.PathLike[str]`` and
``os.PathLike[bytes]``.

torchgeo/datasets/geo.py
pioneerHitesh: Here the issue is with mypy wrongly throwing an error. The ``check_instance_type()`` will return ``True`` if it is not a iterable 
so inner body of if loop will cast the ``paths`` to a ``list[paths]`` and if it will be ``False`` which means it is iterable, then the paths is directly assigned to ``self.paths``.
Below are the code and error

Code
``paths = [self.paths]``
Output
``` List item 0 has incompatible type "Union[bytes, PathLike[bytes], PathLike[str], Iterable[Union[str, bytes, PathLike[bytes], PathLike[str]]]]"; expected "Union[str, bytes, PathLike[bytes], PathLike[str]]"```

Code
``paths = self.paths`` 
Output
```torchgeo/datasets/geo.py:301: error: Incompatible types in assignment (expression has type "Union[bytes, PathLike[bytes], PathLike[str], Iterable[Union[str, bytes, PathLike[bytes], PathLike[str]]]]", variable has type "Iterable[Union[str, bytes, PathLike[bytes], PathLike[str]]]")  [assignment]```

File: torchgeo/datasets/cbf.py (hunk 20 above)
Comment (pioneerHitesh) [reply torchgeo/datasets/cbf.py:116]: @adamjstewart 

Code:
``assert isinstance(self.paths, Path)``.

Mypy error:
```
Parameterized generics cannot be used with class or instance checks  [misc]
Argument 2 to "isinstance" has incompatible type "<typing special form>"; expected "_ClassInfo"  [arg-type]
```

torchgeo/datasets/benin_cashews.py
pioneerHitesh: @adamjstewart  This is noted , will remove this later.

torchgeo/datasets/inaturalist.py
pioneerHitesh: This will need to be ``str`` as ``join`` does not support concatenation of ``bytes`` and ``str``.

pioneerHitesh:
> > However we cannot split the datasets and tests separately as that would lead to failing of tests instead we could split datasets and tests by limiting the number of datasets and tests files changed in a commit .
> 
> We could fix the datasets in one PR and then update the tests in a second PR after the first one is merged.

Should I start implementing the changes?

torchgeo/datasets/geo.py
pioneerHitesh: This will need to be ``str`` as join does not support concatenation of ``bytes`` and ``str``.

pioneerHitesh: Possibly ``bytes`` and ``os.PathLike`` is not supported in ``re.match``. Either both arguments in ``re.match`` should be ``str`` or ``bytes``.
[More info here](https://docs.python.org/3/library/re.html)

adamjstewart:
Yeah, let's start with PRs for `os.path.expanduser` and `root={self.paths}`, then once those are merged a PR for `Path`.

pioneerHitesh:
> os.path.expanduser
Creating a separate issue would be better? This PR already has 75 comments.

adamjstewart:
We can either close this PR and start new PRs or mark this one as a draft and rebase once other PRs are merged. I have no preference.

pioneerHitesh:
> We can either close this PR and start new PRs or mark this one as a draft and rebase once other PRs are merged. I have no preference.

I think we should convert this to draft.

pioneerHitesh:
> Yeah, let's start with PRs for `os.path.expanduser` and `root={self.paths}`, then once those are merged a PR for `Path`.

should i rename ``root`` to ``path`` in   ``__init__``  methods?

adamjstewart:
For NonGeoDatasets we want to keep using `root`. For GeoDatasets, we should use `paths` everywhere, including in error messages.

pioneerHitesh:
@adamjstewart  please take a look at this [PR](https://github.com/microsoft/torchgeo/pull/1702).

adamjstewart:
Closing in favor of #1731

PR #1681, opened 2023-10-19
EuroSATDataModule: set mean/std based on bands

robmarkcole:
Bands are now used to update the `mean` and `std` attributes used in normalisation. 

Closes https://github.com/microsoft/torchgeo/issues/1634

### DIFF
--- a/torchgeo/datamodules/eurosat.py
+++ b/torchgeo/datamodules/eurosat.py
@@ -10,22 +10,6 @@ import torch
 from ..datasets import EuroSAT, EuroSAT100
 from .geo import NonGeoDataModule
 
-ALL_BANDS = [
-    "B01",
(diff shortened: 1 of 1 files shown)


torchgeo/datamodules/eurosat.py
```diff
@@ -71,6 +80,10 @@ def __init__(
         """
         super().__init__(EuroSAT, batch_size, num_workers, **kwargs)
 
+        bands = kwargs.get("bands", ALL_BANDS)
```
adamjstewart, line 83: Could we use `EuroSAT.all_band_names` instead so we don't have to duplicate the band names?

adamjstewart (approved):
Looks great! Feel free to do the same for other data modules if you want.

PR #1684, opened 2023-10-20
Issue1683

robmarkcole:
Closes https://github.com/microsoft/torchgeo/issues/1683

### DIFF
--- a/torchgeo/datasets/oscd.py
+++ b/torchgeo/datasets/oscd.py
@@ -78,11 +78,31 @@ class OSCD(NonGeoDataset):
 
     colormap = ["blue"]
 
+    all_band_names = (
+        "B01",
+        "B02",
+        "B03",
+        "B04",
+        "B05",
+        "B06",
+        "B07",
+        "B08",
+        "B8A",
+        "B09",
+        "B10",
+        "B11",
+        "B12",
+    )
+
+    rgb_bands = ("B04", "B03", "B02")
+
+    BAND_SETS = {"all": all_band_names, "rgb": rgb_bands}
+
     def __init__(
         self,
         root: str = "data",
         split: str = "train",
-        bands: str = "all",
+        bands: Sequence[str] = BAND_SETS["all"],
         transforms: Optional[Callable[[dict[str, Tensor]], dict[str, Tensor]]] = None,
         download: bool = False,
         checksum: bool = False,
@@ -103,7 +123,9 @@ class OSCD(NonGeoDataset):
                 don't match
         """
         assert split in self.splits
-        assert bands in ["rgb", "all"]
+        self._validate_bands(bands)
+        self.bands = bands
+        self.band_indices = [self.all_band_names.index(b) for b in bands if b in self.all_band_names]
 
         self.root = root
         self.split = split
@@ -170,8 +192,8 @@ class OSCD(NonGeoDataset):
                 )
 
             images1, images2 = get_image_paths(1), get_image_paths(2)
-            if self.bands == "rgb":
-                images1, images2 = images1[1:4][::-1], images2[1:4][::-1]
+            images1 = [images1[i] for i in self.band_indices][::-1]
+            images2 = [images2[i] for i in self.band_indices][::-1]
 
             with open(os.path.join(images_root, region, "dates.txt")) as f:
                 dates = tuple(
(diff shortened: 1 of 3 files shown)


torchgeo/datasets/oscd.py
```diff
@@ -170,8 +200,8 @@ def get_image_paths(ind: int) -> list[str]:
                 )
 
             images1, images2 = get_image_paths(1), get_image_paths(2)
-            if self.bands == "rgb":
-                images1, images2 = images1[1:4][::-1], images2[1:4][::-1]
```
robmarkcole, line 174: Not sure why `[::-1]` was performed here

tests/datasets/test_oscd.py
```diff
@@ -72,12 +74,12 @@ def dataset(
...
-        assert isinstance(x["image"], torch.Tensor)
-        assert x["image"].ndim == 3
+        assert isinstance(x["image1"], torch.Tensor)
+        assert x["image1"].ndim == 3
         assert isinstance(x["mask"], torch.Tensor)
         assert x["mask"].ndim == 2
 
-        if dataset.bands == "rgb":
+        if dataset.bands == OSCD.rgb_bands:
             assert x["image"].shape[0] == 6
```
adamjstewart, line 83: Also needs to be updated

tests/datasets/test_oscd.py
```diff
@@ -72,12 +74,12 @@ def dataset(
     def test_getitem(self, dataset: OSCD) -> None:
         x = dataset[0]
         assert isinstance(x, dict)
-        assert isinstance(x["image"], torch.Tensor)
-        assert x["image"].ndim == 3
+        assert isinstance(x["image1"], torch.Tensor)
+        assert x["image1"].ndim == 3
```
adamjstewart, line 78: Can we also check image2 here?

torchgeo/datasets/oscd.py
```diff
@@ -78,11 +78,31 @@ class OSCD(NonGeoDataset):
 
     colormap = ["blue"]
 
+    all_band_names = (
```
adamjstewart, line 81: Let's call this `all_bands` to match `RasterDataset`

torchgeo/datasets/oscd.py
```diff
@@ -78,11 +78,31 @@ class OSCD(NonGeoDataset):
...
+        "B8A",
+        "B09",
+        "B10",
+        "B11",
+        "B12",
+    )
+
+    rgb_bands = ("B04", "B03", "B02")
+
+    BAND_SETS = {"all": all_band_names, "rgb": rgb_bands}
```
adamjstewart, line 99: Do we really need `BAND_SETS`? We only use it in one place.

torchgeo/datasets/oscd.py
```diff
@@ -103,7 +123,15 @@ def __init__(
                 don't match
         """
         assert split in self.splits
-        assert bands in ["rgb", "all"]
+        self._validate_bands(bands)
```
adamjstewart, line 126: I think we can make this simpler:
```python
```

torchgeo/datasets/oscd.py
```diff
@@ -103,7 +123,15 @@ def __init__(
...
+        self.bands = bands
+        self.all_band_indices = [self.all_band_names.index(b) for b in bands]
+        self.rgb_indices = []
+
+        try:
+            self.rgb_indices = [bands.index(band) for band in self.rgb_bands]
```
adamjstewart, line 132: We only use this in `plot`, so let's create it in plot

torchgeo/datasets/oscd.py
```diff
@@ -130,12 +158,15 @@ def __getitem__(self, index: int) -> dict[str, Tensor]:
         image2 = self._load_image(files["images2"])
         mask = self._load_target(str(files["mask"]))
 
-        image = torch.cat([image1, image2])
-        sample = {"image": image, "mask": mask}
-
         if self.transforms is not None:
```
adamjstewart, line 161: This seems more complicated than it needs to be. Can we not get `self.transforms` to work on the new sample?

torchgeo/transforms/transforms.py
```diff
@@ -42,7 +42,7 @@ def __init__(
 
         keys: list[str] = []
         for key in data_keys:
-            if key == "image":
+            if key in ["image", "image1", "image2"]:
```
adamjstewart, line 45: Can we do:
```suggestion
            if key.startswith("image"):
```

torchgeo/datasets/oscd.py
adamjstewart: Probably to convert BGR to RGB

torchgeo/datasets/oscd.py
```diff
@@ -290,10 +307,15 @@ def plot(
         """
```
adamjstewart, line 307: Can you add a "Raises:" section to the docstring?

adamjstewart (commented):
Looks great! One last documentation change

torchgeo/datasets/oscd.py
```diff
@@ -290,10 +307,15 @@ def plot(
...
         ncols = 2
 
-        rgb_inds = [3, 2, 1] if self.bands == "all" else [0, 1, 2]
+        try:
+            rgb_indices = [self.bands.index(band) for band in self.rgb_bands]
+        except ValueError as e:
+            raise ValueError(
+                "RGB bands must be present to use `plot` with Sentinel-2 imagery."
```
adamjstewart, line 314: Never heard of `raise from` before, neat!

Btw, you'll need to add a simple test to cover this line. Should be plenty of examples in other datasets.

adamjstewart (approved):

PR #1690, opened 2023-10-22
Updated ChesapeakeCVPR to use NLCD colormap

### LINKED ISSUE #1264: ChesapeakeCVPR: use NLCD mapping/cmap
### Summary

We recently added a new NLCD dataset. Instead of duplicating the class mapping and cmap in ChesapeakeCVPR, we should use them from the NLCD dataset.

### Rationale

Code duplication is bad.

### Implementation

```python
from .nlcd import NLCD

cmap = NLCD.cmap
```

### Alternatives

Only reason to not duplicate them is in case we add Alaska to NLCD, which contains classes that don't exist in ChesapeakeCVPR. This doesn't affect cmap, but may affect the class mapping if we want to minimize the number of predicted classes.

### Additional information

_No response_

alhridoy:
Fixes #1264

This PR updates the ChesapeakeCVPR class in `torchgeo/datasets/chesapeake.py` to use the colormap from the NLCD dataset. 

Changes include:
- Set `self.cmap` to `NLCD.cmap` in the `__init__` method of the `ChesapeakeCVPR` class.
- Defined a property `_nlcd_cmap` that returns `self.cmap`.


alhridoy:
> @microsoft-github-policy-service agree

torchgeo/datasets/chesapeake.py
```diff
@@ -618,7 +618,9 @@ def __init__(
                             "prior_from_cooccurrences_101_31_no_osm_no_buildings": prior_fn,  # noqa: E501
                         },
                     )
-
+    @property
+    def _nlcd_cmap(self):
```
adamjstewart, line 622: Why is the property needed?

alhridoy: Provides encapsulation and improving code readability for future changes in how colormap is derived.

adamjstewart: We already have that with the `self.cmap` attribute, I don't think we need a second attribute.

calebrob6: I agree

calebrob6:
I like this change @alhridoy, thanks for the contribution! Let me know if you want help with the linters / mypy / etc.

alhridoy:
Hi @adamjstewart @calebrob6 I've made the changes as per your suggestions. I've removed the `_nlcd_cmap` property and replaced all its references with `self.cmap`. I've also set `self.cmap` to `NLCD.cmap` in the `__init__` method and removed the block of code that was creating `nlcd_colors` and `_nlcd_cmap`. Would you check and give me feedbacks.

torchgeo/datasets/chesapeake.py
```diff
@@ -584,11 +583,6 @@ def __init__(
         lc_colors = lc_colors[:, :3] / 255
         self._lc_cmap = ListedColormap(lc_colors)
 
-        nlcd_colors = np.zeros((max(self.nlcd_cmap.keys()) + 1, 4))
-        nlcd_colors[list(self.nlcd_cmap.keys())] = list(self.nlcd_cmap.values())
```
adamjstewart, line 588: The tests are failing because the cmap isn't in the format expected by matplotlib. It might be simplest to use:
```python
nlcd_colors = np.zeros((max(NLCD.cmap.keys()) + 1, 4))
nlcd_colors[list(NLCD.cmap.keys())] = list(NLCD.cmap.values())
```
instead. Then we don't need to set `self.cmap` since there may be 2 different cmaps for the same dataset. You can also delete `nlcd_cmap` above to avoid duplication.

torchgeo/datasets/chesapeake.py
```diff
@@ -22,11 +22,10 @@
 from matplotlib.figure import Figure
 from rasterio.crs import CRS
 from torch import Tensor
-
 from .geo import GeoDataset, RasterDataset
+from .nlcd import NLCD
```
adamjstewart, line 26: If you run `isort` on the file, it will automatically take care of these for you. See https://torchgeo.readthedocs.io/en/stable/user/contributing.html#linters

torchgeo/datasets/chesapeake.py
```diff
@@ -120,7 +119,6 @@ def __init__(
         self.paths = paths
         self.download = download
         self.checksum = checksum
-
```
adamjstewart, line 123: Let's undo these whitespace changes

torchgeo/datasets/chesapeake.py
```diff
@@ -775,7 +772,7 @@ def plot(
                 img = mask[:, :, 0]
                 mask = mask[:, :, 1:]
                 axs[i].imshow(
-                    img, vmin=0, vmax=95, cmap=self._nlcd_cmap, interpolation="none"
+                    img, vmin=0, vmax=95, cmap=self.cmap, interpolation="none"
```
adamjstewart, line 775: Undo this change now that we're using `self._nlcd_cmap` again

torchgeo/datasets/chesapeake.py
```diff
@@ -22,11 +22,10 @@
 from matplotlib.figure import Figure
 from rasterio.crs import CRS
 from torch import Tensor
-
 from .geo import GeoDataset, RasterDataset
+from .nlcd import NLCD
 from .utils import BoundingBox, DatasetNotFoundError, download_url, extract_archive
 
-
 class Chesapeake(RasterDataset, abc.ABC):
```
adamjstewart, line 29: You can now delete `nlcd_cmap` at the top of this class, it's no longer used.

adamjstewart (changes requested):

adamjstewart:
Ping @alhridoy

adamjstewart:
There are now no changes lol.

alhridoy:
> There are now no changes lol.

Hi @adamjstewart,

I've made the changes as per your suggestions and committed them. Could you please check and confirm if everything is in order now?  Thanks for your guidance and patience.

adamjstewart:
Still need to do this:

> You can now delete `nlcd_cmap` at the top of this class, it's no longer used.

alhridoy:
> Still need to do this:
> 
> > You can now delete `nlcd_cmap` at the top of this class, it's no longer used.

Regarding your suggestion to remove the `nlcd_cmap` attribute from the `Chesapeake` class, I noticed that it's being used in the `ChesapeakeCVPR` class for creating a colormap for the "nlcd" layer.  Could you please confirm if the removal is intended for the ChesapeakeCVPR class instead? Thanks for your guidance.

> You can now delete `nlcd_cmap` at the top of this class, it's no longer used.

adamjstewart:
Yes, for ChesapeakeCVPR. The Chesapeake class doesn't include NLCD data.

alhridoy:
> ### Some checks were not successful
> 11 successful, 8 cancelled, and 2 failing checks

After removing the  nlcd_cmap I've noticed that some of the checks are failing for my recent commit, specifically the `pytest` checks. I've looked into the details but I'm having trouble understanding why these tests are failing. Could you please help me understand what might be causing these failures, and how I can fix them?

adamjstewart:
Seems like an intermittent issue, restarting the test solved it.

adamjstewart (approved):

alhridoy:
> Seems like an intermittent issue, restarting the test solved it.

Thank you for your patient and guidance throughout the PR. I am happy to contribute.

adamjstewart:
Thanks for contributing! Hopefully the first of many 😄

PR #1695, opened 2023-10-24
Implements variably-sized caches in 'RasterDataset'

### LINKED ISSUE #1694: Support variably-sized caches in 'RasterDataset'
### Summary

Currently, 'RasterDataset' caches warp files with a fixed-sized LRU cache with 128 elements. I propose supporting variably-sized caches for subclasses.

### Rationale

When loading large raster files, the fixed-size cache consumes considerable memory. For a given machine, this fixed-size overhead restricts the number of parallel workers usable for DataLoaders. 

In our application, training batch creation is limited by the number of parallel workers rather than data access speeds. If we could reduce the size of caches during training, we could spawn additional dataloader workers and remove the present bottleneck.

### Implementation

We'd add a member to "RasterDataset" and cache "_load_warp_file" during the constructor.

### Alternatives

There may be others, but this one plays (relatively) nicely with MyPy the inability of method decorators to access class- or instance-members. 

### Additional information

_No response_

pmaldonado:
Closes #1694.

### DIFF
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -63,6 +63,10 @@ class CustomNonGeoDataset(NonGeoDataset):
         return 2
 
 
+class CustomRasterDataset(RasterDataset):
+    _cache_size = 32
+
+
 class TestGeoDataset:
     @pytest.fixture(scope="class")
     def dataset(self) -> GeoDataset:
@@ -258,6 +262,13 @@ class TestRasterDataset:
 
         with pytest.raises(AssertionError, match=msg):
             CustomSentinelDataset(root, bands=bands, transforms=transforms, cache=cache)
+
+    def test_cache_size_default(self) -> None:
+        root = os.path.join("tests", "data", "raster", "res_2_epsg_4326")
+        default_cache_load = RasterDataset(root)._cached_load_warp_file
+        custom_cache_load = CustomRasterDataset(root)._cached_load_warp_file
+        assert default_cache_load.cache_info().maxsize == 128
+        assert custom_cache_load.cache_info().maxsize == 32
 
 
 class TestVectorDataset:
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -340,6 +340,9 @@ class RasterDataset(GeoDataset):
     #: Color map for the dataset, used for plotting
     cmap: dict[int, tuple[int, int, int, int]] = {}
 
+    #: Default maxsize for LRU cache.
+    _cache_size = 128
+
     @property
     def dtype(self) -> torch.dtype:
         """The dtype of the dataset (overrides the dtype of the data file via a cast).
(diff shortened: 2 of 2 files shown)


pmaldonado:

pmaldonado:
The current error on `tests / pytest (windows-latest, 3.10)` appears to be unrelated to my changes:

```
unrar v5.30.0 [Approved]
unrar package files install completed. Performing other installation steps.
Attempt to get headers for http://www.rarlab.com/rar/unrarw32.exe failed.
  The remote file either doesn't exist, is unauthorized, or is forbidden for url 'http://www.rarlab.com/rar/unrarw32.exe'. Exception calling "GetResponse" with "0" argument(s): "The remote server returned an error: (404) Not Found."
Downloading unrar 
  from 'http://www.rarlab.com/rar/unrarw32.exe'
ERROR: The remote file either doesn't exist, is unauthorized, or is forbidden for url 'http://www.rarlab.com/rar/unrarw32.exe'. Exception calling "GetResponse" with "0" argument(s): "The remote server returned an error: (404) Not Found." 
```

calebrob6:
Hey @pmaldonado, we're aware that the windows test is failing for unrelated reasons. Going to give it a day to see if it resolves, then look into it.

adamjstewart:
I wonder if we could somehow detect the total RAM size and set the default cache size based on that. I don't really want to add a new dependency on psutil though.

pmaldonado:
> I wonder if we could somehow detect the total RAM size and set the default cache size based on that. I don't really want to add a new dependency on psutil though.

Should "GeoDataset" change it's behavior based on properties of the DataModule (the number of workers) or system (amount of RAM)? 

A lightweight solution to our issue is allow users to overwrite the `_cache_size` member when subclassing "RasterDataset" (or overwriting it at import time, if desirable). It's backwards compatible and allows users to adjust memory usage to their needs, if needed, but doesn't require their judgement if the current behavior is sufficient.

An alternate implementation could be allow the `cache` parameter to the constructor to be of type `bool | int` and interpret an integer value as the desired cache size. This would also be backwards compatible, but would require updating all of the `RasterDataset` subclasses. The one clear benefit would be configuring the cache size upon dataset creation, rather than definition or import. This doesn't appear significant to me, since cache size should vary much more between application/platforms rather than between datasets in a given application.

adamjstewart:
> Should "GeoDataset" change it's behavior based on properties of the DataModule (the number of workers) or system (amount of RAM)?

I _think_ the filename LRU cache is shared across all processes so it isn't affected by the number of workers, but correct me if I'm wrong. That's just an assumption, I have no evidence.

> cache size should vary much more between application/platforms rather than between datasets in a given application.

This depends. If we know that one dataset uses COGs and one doesn't, that may affect how large we want the cache to be. But let's explore `GDAL_CACHEMAX` more first before making a decision here. It's possible that the filename LRU cache can be infinitely sized as long as `GDAL_CACHEMAX` is set appropriately.

calebrob6 (approved):

calebrob6:
This is fine with me!

adamjstewart:
@pmaldonado have you tried `GDAL_CACHEMAX` to see if it's possible to control this via environment variable instead? See your original issue for my suggestions.

adamjstewart:
Ping @pmaldonado

pmaldonado:
Got pulled into a few other things. I'll give your suggestion with `GDAL_CACHEMAX` a shot later this week when I get some time.

adamjstewart:
Is this still a work in progress? We're preparing for a new release and trying to close out old issues/PRs.

PR #1696, opened 2023-10-24
Update levircd

robmarkcole:
Address levircd changes required for https://github.com/microsoft/torchgeo/issues/1691

### DIFF
--- a/tests/datasets/test_levircd.py
+++ b/tests/datasets/test_levircd.py
@@ -41,8 +41,8 @@ class TestLEVIRCDPlus:
         assert isinstance(x["image1"], torch.Tensor)
         assert isinstance(x["image2"], torch.Tensor)
         assert isinstance(x["mask"], torch.Tensor)
-        assert x["image1"].shape[0] == 2
-        assert x["image2"].shape[0] == 2
+        assert x["image1"].shape[0] == 3
+        assert x["image2"].shape[0] == 3
 
     def test_len(self, dataset: LEVIRCDPlus) -> None:
         assert len(dataset) == 2
--- a/tests/datasets/test_levircd.py
+++ b/tests/datasets/test_levircd.py
@@ -38,9 +38,11 @@ class TestLEVIRCDPlus:
     def test_getitem(self, dataset: LEVIRCDPlus) -> None:
         x = dataset[0]
         assert isinstance(x, dict)
-        assert isinstance(x["image"], torch.Tensor)
+        assert isinstance(x["image1"], torch.Tensor)
+        assert isinstance(x["image2"], torch.Tensor)
         assert isinstance(x["mask"], torch.Tensor)
-        assert x["image"].shape[0] == 2
+        assert x["image1"].shape[0] == 2
+        assert x["image2"].shape[0] == 2
 
     def test_len(self, dataset: LEVIRCDPlus) -> None:
         assert len(dataset) == 2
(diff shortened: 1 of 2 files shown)


robmarkcole:
Noticed that levircd images are `sample['image1'].dtype == torch.uint8`, do we want to cast these to `float` @adamjstewart  ?

tests/datasets/test_levircd.py
```diff
@@ -38,9 +38,11 @@ def dataset(
     def test_getitem(self, dataset: LEVIRCDPlus) -> None:
         x = dataset[0]
         assert isinstance(x, dict)
-        assert isinstance(x["image"], torch.Tensor)
+        assert isinstance(x["image1"], torch.Tensor)
+        assert isinstance(x["image2"], torch.Tensor)
         assert isinstance(x["mask"], torch.Tensor)
-        assert x["image"].shape[0] == 2
+        assert x["image1"].shape[0] == 3
```
robmarkcole, line 44: These are RGB, not sure why dim was 2 before

robmarkcole:
Also do we want a channel dimension on the mask in `_load_target`?

robmarkcole:
Created the following datamodule

```python
class LEVIRCDPlusDataModule(NonGeoDataModule):
    """LightningDataModule implementation for the OSCD dataset.

    Uses the train/test splits from the dataset and further splits
    the train split into train/val splits.

    .. versionadded:: 0.2
    """

    def __init__(
        self,
        batch_size: int = 8,
        patch_size: Union[tuple[int, int], int] = 64,
        val_split_pct: float = 0.2,
        num_workers: int = 0,
        **kwargs: Any,
    ) -> None:
        """Initialize a new LEVIRCDPlusDataModule instance.

        Args:
            batch_size: Size of each mini-batch.
            patch_size: Size of each patch, either ``size`` or ``(height, width)``.
                Should be a multiple of 32 for most segmentation architectures.
            val_split_pct: Percentage of the dataset to use as a validation set.
                :class:`~torchgeo.datasets.LEVIRCDPlus`.
        """
        super().__init__(LEVIRCDPlus, 1, num_workers, **kwargs)

        self.patch_size = _to_tuple(patch_size)
        self.val_split_pct = val_split_pct

        self.mean = torch.tensor([0.485, 0.456, 0.406])
        self.std = torch.tensor([0.229, 0.224, 0.225])

            _RandomNCrop(self.patch_size, batch_size),
            data_keys=["image1", "image2", "mask"],
        )

    def setup(self, stage: str) -> None:
        """Set up datasets.

        Args:
        """
        if stage in ["fit", "validate"]:
            self.dataset = LEVIRCDPlus(split="train", **self.kwargs)
            self.train_dataset, self.val_dataset = dataset_split(
                self.dataset, val_pct=self.val_split_pct
            )
        if stage in ["test"]:
            self.test_dataset = LEVIRCDPlus(split="test", **self.kwargs)
```

to test with the [CustomSemanticSegmentationTask](https://github.com/microsoft/torchgeo/pull/1684#issue-1954327984)
but get error 
```
 --> 459 return F.conv2d(input, weight, bias, self.stride,
    460                 self.padding, self.dilation, self.groups)

RuntimeError: expected scalar type Byte but found Float
```
So there is an invalid type somewhere, but I am not sure where

tests/datasets/test_levircd.py
robmarkcole: OK since before `sample['image'].shape == torch.Size([2, 3, 1024, 1024])`

adamjstewart (approved):

PR #1702, opened 2023-10-27
Fixed string formatting and naming issues.

pioneerHitesh:
Fixed ``root={self.paths}`` in the ``datasets module`` and I have also fixed inconsistent ``root`` and ``path`` naming conventions.

### DIFF
--- a/tests/datasets/test_chesapeake.py
+++ b/tests/datasets/test_chesapeake.py
@@ -170,7 +170,7 @@ class TestChesapeakeCVPR:
         assert isinstance(ds, UnionDataset)
 
     def test_already_extracted(self, dataset: ChesapeakeCVPR) -> None:
-        ChesapeakeCVPR(root=dataset.root, download=True)
+        ChesapeakeCVPR(paths=dataset.paths, download=True)
 
     def test_already_downloaded(self, tmp_path: Path) -> None:
         root = str(tmp_path)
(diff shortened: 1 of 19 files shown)


adamjstewart:
We try to use `root` for datasets that require a single directory and `paths` for datasets that support one or more directories or filenames. Most GeoDatasets support `paths`, but not all. That's part of the inconsistency, and we might want to keep that. All NonGeoDatasets only support `root`.

pioneerHitesh:
@adamjstewart  ``mypy`` throws error for paths being of type ``str`` and ``Iterable[str]`` . It seems like we should add ``pathlib.path``  support first then we may embark on other removing ``os.path.expanduser()`` and renaming ``root`` to ``paths``. What is your take on this?

adamjstewart:
The only thing I suggested to rename was `root={self.paths}` to `paths={self.paths!r}`, I didn't suggest renaming all `root` to `paths`. Mypy shouldn't touch this code since it's a string name.

pioneerHitesh:
@adamjstewart can you please take a look at this [PR](https://github.com/microsoft/torchgeo/pull/1704)

pioneerHitesh:
@adamjstewart i have removed ``os.path.expanduser`` in this [PR](https://github.com/microsoft/torchgeo/pull/1705). Can you please take a look?

adamjstewart:
P.S. We're planning a 0.5.1 release next week, would be great to get this PR in if you want to remove the unrelated changes and stick to just updating the error messages.

adamjstewart:
Nevermind, #1704 had all other changes, so I think this can be closed now.

pioneerHitesh:
> P.S. We're planning a 0.5.1 release next week, would be great to get this PR in if you want to remove the unrelated changes and stick to just updating the error messages.

Sorry for the belated reply, but i may not be able to help you out with this.

PR #1703, opened 2023-10-28
Fix all Trainers to support PyTorch DataLoaders

kaybe20:
Fixes an error that occurred when using Trainers with PyTorch DataLoaders instead of a Lightning DataModule.
-> as discussed in #1598


kaybe20:

torchgeo/trainers/classification.py
adamjstewart, line 1: There are actually two places this needs to be done in `classification.py`, there's another class below.

kaybe20: Oh sorry, I missed that

adamjstewart (approved):

PR #1705, opened 2023-10-31
Removed ``os.path.expanduser`` in ``datasets`` module.

pioneerHitesh:
Removed ``os.path.expanduser(root)`` and replaced it with ``root``.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,5 +34,5 @@ repos:
         hooks:
         -   id: mypy
             args: [--strict, --ignore-missing-imports, --show-error-codes]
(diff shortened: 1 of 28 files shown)


adamjstewart:
This also includes all changes to root -> path

pioneerHitesh:
> This also includes all changes to root -> path
Sorry I have done this by mistake and will fix this and then request a review.

adamjstewart:
Looks good now!

I think everyone will agree that consistency is correct. So the only options are to either remove `os.path.expanduser` from all datasets or add it to all datasets. Will leave this open for a while to see if anyone else wants to review. 

Note that all download/extract utilities in torchgeo/torchvision use `os.path.expanduser` already. `torchvision.datasets` has the same inconsistency we have, where most don't use it but some do.

PR #1711, opened 2023-11-03
[WIP] Add ChangeDetectionTask

robmarkcole:
Ref https://github.com/microsoft/torchgeo/discussions/1535

Seeking feedback:
1) Binary change is common, make this default (setting 'num_classes=1')?
2) Include Unet or not? I've found it quite useful as a baseline
3) For binary there usually is a background class and the change class (which you actually care about), can we report metrics specifically for the change?
4) FCSiamDiff appears the simplest, implement next?

Notes
- With `MulticlassAccuracy` and `num_classes=1` an error is raised: `ValueError: Expected argument `num_classes` to be an integer larger than 1, but got 1`
- With `JaccardLoss` and `num_classes=1` an error is raised: `RuntimeError: one_hot is only applicable to index tensor.`
- `BinaryAccuracy` & `BinaryJaccardIndex` do not allow `ignore_index`

### DIFF
--- a/torchgeo/trainers/__init__.py
+++ b/torchgeo/trainers/__init__.py
@@ -5,7 +5,7 @@
 
 from .base import BaseTask
 from .byol import BYOLTask
-from .change import BinaryChangeDetectionTask
+from .change import ChangeDetectionTask
 from .classification import ClassificationTask, MultiLabelClassificationTask
 from .detection import ObjectDetectionTask
 from .moco import MoCoTask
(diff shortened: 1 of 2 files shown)


adamjstewart:
We don't need `ignore_index` if there are only 2 possible classes, binary is the correct loss metric.

robmarkcole:
Right perhaps I've misunderstood the role of `ignore_index` - it is to handle `nodata` values, but for binary they just merge into the 'background' class then?

adamjstewart:
How many of our change detection datasets have nodata values? I think we'll need some way to distinguish them from the background class (there are no false positives or false negatives for nodata values). It might be worth opening an issue/PR with torchmetrics to add `ignore_index` to the binary methods. I don't think the multi-class methods with classes=2 are equivalent unfortunately.

robmarkcole:
None of Levir, xview or oscd have `nodata`

Issue #1263, opened 2023-04-18
Dataset plotting: normalization


adamjstewart:
### Summary

At the moment, our dataset plotting routines are inconsistent. While some plot methods stretch to the range of the image, others simply divide by 3K or 10K and clip to get images to the range of 0–1. I propose we convert the latter to the former and consistently stretch images for all datasets.

### Rationale

While technically correct, many of our plotting methods make it difficult to visualize images. This is especially true for datamodules, where normalization has been applied to all images and images are no longer in the uint8 or float32 range of the original data.

### Implementation

I propose we use one of the default visualization options used by QGIS:

1. Clip to 2% to 98% range (exact percentages TBD)
2. Clip to min/max
3. Clip to mean ± 2 std dev (exact std dev TBD)

We already have a `torchgeo.datasets.utils.percentile_normalization` method we could use or modify for this purpose.

### Alternatives

We could apply an inverse Normization transform during datamodule plotting. This would help with datamodule plotting, but still suffers from inconsistent dataset plotting. It would only be a couple lines of code though, which would make it much easier.

### Additional information

@calebrob6 we discussed this on Slack or somewhere.

Note that this contradicts #496, so we should decide on one approach or another.

adamjstewart:
Before #476, we basically had this functionality for most of our `GeoDatasets` already. `RasterDataset.plot` would normalize all images to the 2% to 98% range using `self.rgb_bands` and `self.all_bands`, and would only need to be overridden for non-image datasets. We should consider bringing this back to save us time.

robmarkcole:
In my own plotting funcs I have been applying the `percentile_normalisation` to 'undo' the effects of normalisation, but the result is not always that great. I think an `un-normalize` followed by `percentile_normalisation` with dataset specific percentiles would make sense. e.g. min/max can then be achieved with 0 & 100 percentiles. Default to 2 & 98 generally works pretty well, and if QGIS do this it is probably pretty sensible

PR #1712, opened 2023-11-04
Fix for Blank Images in plot due to Float Tensor Ranges

menglutao:
**Description:**
Encountered an issue where ucmerced dataset images would render as blank after using the plot function and matplotlib to show the image.

**Issue:**
The underlying cause is that Matplotlib expects float values to be normalized between 0 and 1. However, the float tensor representing our image data occasionally contained values in the 0 to 255 range, leading to unexpected behavior in the rendering process.

**Resolution:**
A conditional check has been introduced. When the maximum value in the image tensor exceeds 1, the tensor is normalized by scaling down to the 0 to 1 range. This ensures compatibility with Matplotlib's expected float value range and resolves the issue of blank images.

**Outcome:**
With the proposed changes, images are displayed correctly, regardless of whether their data originally spans 0 to 1 or 0 to 255. 

I invite the maintainers to review the changes. Please advise if further modifications are required or if the update can be merged.

### DIFF
--- a/torchgeo/datasets/ucmerced.py
+++ b/torchgeo/datasets/ucmerced.py
@@ -214,6 +214,11 @@ class UCMerced(NonGeoClassificationDataset):
         .. versionadded:: 0.2
         """
         image = np.rollaxis(sample["image"].numpy(), 0, 3)
+
+        # Normalize the image if the max value is greater than 1
+        if image.max() > 1:
+            image = image.astype(np.float32) / 255.0 # Scale to [0, 1] 
+
         label = cast(int, sample["label"].item())
         label_class = self.classes[label]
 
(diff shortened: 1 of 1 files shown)


adamjstewart:
This relates to #1263. Depending on whether or not normalization is added, all of our plotting breaks. A better solution would be to implement the ideas in #1263, but that's a lot of work.

Looking through the data, it seems like every image is in the range 0–255. Until we implement #1263, it may be simpler just to always divide by 255.

calebrob6:
Let's merge this, as plotting of images on tensorboard is indeed broken

adamjstewart:
Doesn't tensorboard use the data module that includes normalization which divides all images by 255? If there are images with values greater than 255 that's a whole nother problem.

adamjstewart:
Just checked and there are none greater than 255. So yes, data module plotting is broken, just like for every other dataset. We really need #1263 if we want both dataset plotting and data module plotting to work.

PR #1714, opened 2023-11-05
Datasets: consistent error messages for missing data

### LINKED ISSUE #660: Add DatasetNotFoundError for consistent error messages

adamjstewart:
Closes #660

### DIFF
--- a/torchgeo/datasets/utils.py
+++ b/torchgeo/datasets/utils.py
@@ -70,6 +70,7 @@ class DatasetNotFoundError(FileNotFoundError):
             val = dataset.paths
         else:
             super().__init__(f"{msg}.")
+            return
 
         msg += f" in `{var}={val!r}` and "
 
--- a/torchgeo/datasets/utils.py
+++ b/torchgeo/datasets/utils.py
@@ -23,11 +23,13 @@ import numpy as np
 import rasterio
 import torch
 from torch import Tensor
+from torch.utils.data import Dataset
 from torchvision.datasets.utils import check_integrity, download_url
 from torchvision.utils import draw_segmentation_masks
 
 __all__ = (
     "check_integrity",
+    "DatasetNotFoundError",
     "download_url",
     "download_and_extract_archive",
     "extract_archive",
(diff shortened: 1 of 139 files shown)


torchgeo/datasets/utils.py
```diff
@@ -46,6 +48,49 @@
...
+    .. versionadded:: 0.6
+    """
+
+    def __init__(self, dataset: Dataset[object]) -> None:
+        """Intstantiate a new DatasetNotFoundError instance.
+
+        Args:
+            dataset: The dataset that was requested.
+        """
+        msg = "Dataset not found"
```
adamjstewart, line 63: Lot of possible alternatives here. Could even add the name of the dataset to the error message if that makes things more clear.

ashnair1 (approved):

calebrob6 (approved):

PR #1719, opened 2023-11-08
Add SSL4EO-L Benchmark to docs

nilsleh:
I think we forgot to add the SSL4EO Benchmark Dataset/Datamodule to the Docs. Do you want to put entries for all possible constillations in the csv file?

### DIFF
--- a/docs/api/datamodules.rst
+++ b/docs/api/datamodules.rst
@@ -131,8 +131,8 @@ SSL4EO
 .. autoclass:: SSL4EOLDataModule
 .. autoclass:: SSL4EOS12DataModule
 
-SSL4EOLBenchmarkDataModule
-^^^^^^^^^^^^^^^^^^^^^^^^^^
+SSL4EO-L Benchmark
+^^^^^^^^^^^^^^^^^^
 
 .. autoclass:: SSL4EOLBenchmarkDataModule
 
--- a/docs/api/datamodules.rst
+++ b/docs/api/datamodules.rst
@@ -131,6 +131,11 @@ SSL4EO
 .. autoclass:: SSL4EOLDataModule
 .. autoclass:: SSL4EOS12DataModule
 
+SSL4EOLBenchmarkDataModule
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. autoclass:: SSL4EOLBenchmarkDataModule
+
 SustainBench Crop Yield
 ^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


adamjstewart:
Good catch! Should we also add this to `non_geo_datasets.csv`?

Yes, we should have an entry for every family of datasets in both the docs and the tables.

docs/api/non_geo_datasets.csv
```diff
@@ -37,6 +37,8 @@ Dataset,Task,Source,# Samples,# Classes,Size (px),Resolution (m),Bands
 `SpaceNet`_,I,WorldView-2/3 Planet Lab Dove,"1,889--28,728",2,102--900,0.5--4,MSI
 `SSL4EO`_-L,T,Landsat,1M,-,264x264,30,MSI
 `SSL4EO`_-S12,T,Sentinel-1/2,1M,-,264x264,10,"SAR, MSI"
+`SSL4EOLBenchmark`_-TM-TOA & CDL, S, Lansat, 25k,18,264x264,30,MSI
```
nilsleh, line 40: @adamjstewart If the row entry looks okay to you, I will add the other benchmark dataset combinations

adamjstewart: Let's just add a single row for "Landsat & NLCD & CDL"

docs/api/datamodules.rst
```diff
@@ -131,6 +131,11 @@ SSL4EO
 .. autoclass:: SSL4EOLDataModule
 .. autoclass:: SSL4EOS12DataModule
 
+SSL4EOLBenchmarkDataModule
```
adamjstewart, line 134: Should be human-readable name, not class name. Like "SSL4EO-L Benchmark". Same in `datasets.rst` and in `non_geo_datasets.csv`

docs/api/non_geo_datasets.csv
nilsleh: maybe one for NLCD and one for CLD because they have different number of classes?

adamjstewart: Sure

docs/api/non_geo_datasets.csv
```diff
@@ -37,6 +37,8 @@ Dataset,Task,Source,# Samples,# Classes,Size (px),Resolution (m),Bands
 `SpaceNet`_,I,WorldView-2/3 Planet Lab Dove,"1,889--28,728",2,102--900,0.5--4,MSI
 `SSL4EO`_-L,T,Landsat,1M,-,264x264,30,MSI
 `SSL4EO`_-S12,T,Sentinel-1/2,1M,-,264x264,10,"SAR, MSI"
+`SSL4EO-L Benchmark`_, S, Lansat & CDL, 25k,18,264x264,30,MSI
+`SSL4EO-L Benchmark`_, S, Lansat & NLCD, 25k,14,264x264,30,MSI
```
adamjstewart, line 41: NLCD has 17 classes and CDL has 134 classes (although we only evaluated on a subset in our paper). Also, I would remove the spaces around commas and capitalize the K.

nilsleh: the config files that we used say 14 classes for nlcd and 18 for cdl

adamjstewart (approved):

PR #1731, opened 2023-11-18
Added ``pathlib`` support and resolved conflicts.

pioneerHitesh:
added ``pathlib`` support , removed ``checked_instance_type`` function

### DIFF
--- a/requirements/style.txt
+++ b/requirements/style.txt
@@ -1,3 +1,3 @@
 # style
 mypy==1.10.1
-ruff==0.4.10
+ruff==0.5.0


adamjstewart:
Let's wait for a response on https://github.com/pytorch/vision/issues/8120 before we do too much. We might be able to get them to support pathlib in the next release.

pioneerHitesh:
> Let's wait for a response on https://github.com/pytorch/vision/issues/8120 before we do too much. We might be able to get them to support pathlib in the next release.

Can you please check your LinkedIn.

pioneerHitesh:
> Based on the current state of [pytorch/vision#8120](https://github.com/pytorch/vision/issues/8120), I think the easiest thing to do for now would be to define functions in `torchgeo.datasets.utils` that are simply wrappers around the torchvision functions. These functions would accept both str and pathlib, then we would cast to str for torchvision. Then we wouldn't need to cast in every single dataset file. Thoughts?

It seems like they have added support for pathlib in `datasets/utils.py`. The following [PR](https://github.com/pytorch/vision/pull/8200) is for the same. So we may not need to define extra functions for casting pathlib to str.

adamjstewart:
Fantastic, let's wait until that makes it into a stable release and then rebase this PR.

adamjstewart:
I believe [torchvision 0.18](https://github.com/pytorch/vision/releases/tag/v0.18.0) includes full pathlib support. If you're still interested, now would be a good time to try to revive this. It's probably easiest to create a new PR since most files have changed since last year.

As a first pass, can you change the type hints (without changing the code or type casting) just to see how much of pathlib we already support? Torchvision only supports `str | Path[str]`, so let's do the same in TorchGeo.

pioneerHitesh:
@adamjstewart That's great, but do i have to create a new PR?

adamjstewart:
You don't have to, although it's easier to create a new PR than to resolve all of the merge conflicts.

pioneerHitesh:
@adamjstewart There are no stubs available for some modules, so mypy is not able to perform type checking resulting in ``[import-untyped]`` errors. Should i just suppress those errors or anything else needs to be done?

adamjstewart:
Can you push your changes to GitHub so I can see what you mean? I don't fully understand.

pioneerHitesh:
@adamjstewart I haven't made any changes , on cloning the repo and running ``mypy  *`` gives ``error: Skipping analyzing "rasterio.crs": module is installed, but missing library stubs or py.typed marker  [import-untyped]``  in the ``datasets`` folder.

PR #1737, opened 2023-11-21
RGBBandsMissingError

isaaccorley:
When a dataset's plot method is used and the dataset is missing RGB bands we raise a ValueError. In our trainers we use try/except to catch this ValueError and skip plotting, however any other ValueError encountered during plotting is also silently skipped. 

This PR adds the following:
- `RGBBandsMissingError` and updates the dataset plot methods to use this for a consistent error message. Dataset tests are also updated to catch this
- Trainers ONLY skip the plotting if this error occurs. All other ValueErrors are raised


adamjstewart:
Love this solution!

> The error in our object detection trainer tests was caused by the use of torch.rand to generated dummy bounding box data. This results in some cases where xmax < xmin and/or ymax < ymin which results in the bounding box plotting throwing a ValueError

Can you move this part of the fix into a separate PR so we can backport it to 0.5.2? The rest of this PR introduces a new error class and will have to be in 0.6.0.

torchgeo/datasets/utils.py
```diff
@@ -91,6 +92,18 @@ def __init__(self, dataset: Dataset[object]) -> None:
...
 
+class RGBBandsMissingError(ValueError):
+    """Raised when a dataset is missing RGB bands for plotting.
+
+    .. versionadded:: 0.5.2
+    """
+
+    def __init__(self) -> None:
+        """Instantiate a RGBBandsMissingError instance."""
+        msg = "Dataset doesn't contain some of the RGB bands"
```
adamjstewart, line 103: Wonder if it's possible to make this message more specific, like print the bands the user requested and which bands are RGB.

adamjstewart: Also, can we use "does not" instead of "doesn't"?

isaaccorley: I may punt on making it based on the specific bands for a future PR.

adamjstewart (approved):
Can always extend the error message in the future if needed. Would be good to check that no other datasets were added since this PR was first branched that need this added. Otherwise, should be good to merge.

PR #1739, opened 2023-11-22
ObjectDetectionTask: increase test coverage for torchvision 0.14+

isaaccorley:
The try/except in the ObjectDetection trainer's validation step was suppressing the following errors which drives this PR. These errors are caused by the use of `torch.rand` to generate dummy bounding box data which results in some cases where `xmax < xmin` and/or `ymax < ymin `which results in the bounding box plotting throwing a ValueError

```python
FAILED tests/trainers/test_detection.py::TestObjectDetectionTask::test_trainer[True-faster-rcnn-nasa_marine_debris] - ValueError: Boxes need to be in (xmin, ymin, xmax, ymax) format. Use torchvision.ops.box_convert to convert them
FAILED tests/trainers/test_detection.py::TestObjectDetectionTask::test_trainer[True-fcos-nasa_marine_debris] - ValueError: Boxes need to be in (xmin, ymin, xmax, ymax) format. Use torchvision.ops.box_convert to convert them
FAILED tests/trainers/test_detection.py::TestObjectDetectionTask::test_trainer[True-retinanet-nasa_marine_debris] - ValueError: Boxes need to be in (xmin, ymin, xmax, ymax) format. Use torchvision.ops.box_convert to convert them
```

Additionally, this PR adds a missing test for covering the case where `fig` returned by the plot method is `None` because the dataset has no plot method.

### DIFF
--- a/tests/trainers/test_detection.py
+++ b/tests/trainers/test_detection.py
@@ -58,12 +58,8 @@ class ObjectDetectionTestModel(Module):
             return output
 
 
-def plot_no_rgb(*args: Any, **kwargs: Any) -> None:
+def plot(*args: Any, **kwargs: Any) -> None:
     raise ValueError
-
-
-def no_plot_method(*args: Any, **kwargs: Any) -> None:
-    return None
 
 
 class TestObjectDetectionTask:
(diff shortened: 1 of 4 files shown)


adamjstewart (commented):
Curious why we never encountered this issue before, but it fails consistently every time in your PR that bumps torch/torchvision versions. Did something change?

There's a lot of code duplication here. Can we:

1. Move `plot_no_rgb` and `no_plot_method` to `tests/trainers/conftest.py` so they can be shared?
2. Combine `test_no_rgb` and `test_no_plot_method` into a single parameterized `test_plot` method?

isaaccorley:
@adamjstewart, I may not have time to finish this in the next few days since I'm traveling. Not sure if you want to finish it up.

adamjstewart:
> Did something change?

Something did indeed change: https://github.com/pytorch/vision/pull/6123

So we used to be generating (10, 4) random floats. The chances that any particular tuple satisfies xmax > xmin AND ymax > ymin is 0.25. The chances that all 10 tuples satisfy this is $0.25^{10} = 9.5^{-7}$. This explains why the tests basically always skipped that line in torchvision 0.14+.

tests/trainers/test_detection.py
```diff
@@ -45,20 +45,27 @@ def forward(self, images: Any, targets: Any = None) -> Any:
         else:  # eval mode
             output = []
             for i in range(batch_size):
+                boxes = torch.zeros(10, 4, dtype=torch.float)
+                # Create xmax, ymax larger than 0.0
+                boxes[:, 2:4] = torch.FloatTensor(10, 2).uniform_(0.1, 0.9)
                 output.append(
                     {
-                        "boxes": torch.rand(10, 4),
+                        "boxes": boxes,
```
adamjstewart, line 53: This change alone is sufficient to increase coverage. Let's do the rest in a separate PR.

adamjstewart (approved):
Thanks for finding this bug! Glad it was only in our testing code and not in the core library.

Issue #978, opened 2022-12-26
Remove tests/datamodules


adamjstewart:
### Summary

I would like to remove (almost) all tests in `tests/datamodules` and replace them with new tests in `tests/trainers` that actually ensure our datamodules and trainers are compatible.

### Rationale

The current tests simply ensure that the data loaders don't crash. They don't actually test that the datamodules are compatible with our trainers.

### Implementation

The bulk of these have already been converted in #329. The remaining tests are:

- [ ] chesapeake (tests invalid arguments)
- [ ] fair1m (requires new trainer: #840)
- [x] inria (#975)
- [x] loveda (#966)
- [ ] nasa_marine_debris
- [ ] oscd
- [ ] potsdam
- [ ] usavars
- [ ] vaihingen
- [ ] xview2

### Alternatives

We may end up keeping some of these that test invalid arguments. The important thing is not that we don't test datamodules standalone, but that we test them with trainers whenever possible.

PR #1748, opened 2023-11-30
Add Digital typhoon dataset

nilsleh:
This PR adds the [Digital Typhoon Dataset](http://agora.ex.nii.ac.jp/digital-typhoon/dataset/).

The implementation allows the following features:
- create an input sequence of single channel images concatenated along channel dimension for nowcasting task (predicting label of last image in that sequence)
- filter samples by min or max feature values
- datamodule that lets you split by storm id (disjoint sets over the time domain) or over the time domain (disjoint set of storm ids)

TODO:
- Target Normalization for regression task

Sample Image:


calebrob6:
This is really cool! I wonder if there is any generalization between this and the Cyclone dataset

nilsleh:
> This is really cool! I wonder if there is any generalization between this and the Cyclone dataset

Stay tuned:)

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,428 @@
...
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+
+        Raises:
+            AssertionError: if ``task`` argument is invalid
+            RuntimeError: if ``download=False`` and data is not found, or checksums
+                don't match
+        """
+        try:
+            import h5py  # noqa: F401
```
nilsleh, line 122: @adamjstewart this is still being done in several datasets (chabud, skippd, so2sat, zuericrop) but `requirements/datasets.txt`lists h5py as a dependency. However, when I don't do the import in the init but at the top of the file, the docs fail with `no module h5py`

nilsleh: besides this and the pesky mypy errors, this should be ready for review

adamjstewart: `requirements/required.txt` is required deps that must always be installed, `requirements/datasets.txt` is optional deps that may or may not be installed.

nilsleh: ah duuh, thanks!

nilsleh:
@adamjstewart not sure how I can fix the read the docs error, do I need to add the TypedDict to init?

adamjstewart:
This RtD error means that:

* It's trying to document the data module class
* And add a link to where the return type is documented
* But the `SampleSequenceDict` class itself doesn't appear in the docs

Some options:

1. Make `split_dataset` a hidden method so it doesn't appear in the docs
2. Use `Any` instead
3. Add `SampleSequenceDict` to the docs

I'm leaning towards 1. What are your thoughts?

nilsleh:
Thanks, yeah option 1 makes sense, because the `TypedDict` is nicer for understanding the code.

docs/api/non_geo_datasets.csv
```diff
@@ -7,6 +7,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `Cloud Cover Detection`_,S,Sentinel-2,"CC-BY-4.0","22,728",2,512x512,10,MSI
 `COWC`_,"C, R","CSUAV AFRL, ISPRS, LINZ, AGRC","AGPL-3.0-only","388,435",2,256x256,0.15,RGB
 `Kenya Crop Type`_,S,Sentinel-2,"CC-BY-SA-4.0","4,688",7,"3,035x2,016",10,MSI
+`Digitial Typhoon Analysis`_,"C, R",Himawari,"CC-BY-SA-4.0","189,364",,512,5km,Infrared
```
adamjstewart, line 10: Shouldn't this be:
```suggestion
`Digitial Typhoon Analysis`_,"C, R",Himawari,"CC-BY-4.0","189,364",,512,5km,Infrared 
```
according to http://agora.ex.nii.ac.jp/digital-typhoon/dataset/?

adamjstewart: It's a classification dataset but there are no # classes?

tests/data/digital_typhoon/data.py
```diff
@@ -0,0 +1,115 @@
...
+
+# Define the root directory
+root = "WP/"
+IMAGE_SIZE = 32
+NUM_TYHOON_IDS = 5
+NUM_IMAGES_PER_ID = 4
+CHUNK_SIZE = 2**12
+
+# Define the root directory
+root = "./WP"
```
adamjstewart, line 20: You defined this twice. I wouldn't include `/` since that's not true on Windows

tests/data/digital_typhoon/data.py
```diff
@@ -0,0 +1,115 @@
...
+
+# If the root directory exists, remove it
+if os.path.exists(root):
+    shutil.rmtree(root)
+
+# Create the root directory if it doesn't exist
+os.makedirs(root)
```
adamjstewart, line 27: Could remove this line, the line below will do this

tests/data/digital_typhoon/data.py
```diff
@@ -0,0 +1,115 @@
...
+os.makedirs(os.path.join(root, "image"))
+os.makedirs(os.path.join(root, "metadata"))
+
+# For each typhoon_id
+all_dfs = []
+for typhoon_id in range(NUM_TYHOON_IDS):
+    # Create a directory under 'root/image/typhoon_id/'
+    os.makedirs(os.path.join(root, "image", str(typhoon_id)), exist_ok=True)
+
+    # Create dummy .hf files
```
adamjstewart, line 39:
```suggestion
    # Create dummy .h5 files
```

tests/datamodules/test_digital_typhoon.py
adamjstewart, line 1: Would love to avoid this file if it isn't needed for test coverage: #978

torchgeo/datamodules/digital_typhoon.py
```diff
@@ -0,0 +1,102 @@
...
+        super().__init__(DigitalTyphoonAnalysis, batch_size, num_workers, **kwargs)
+
+        assert (
+            split_by in self.valid_split_types
+        ), f"Please choose from {self.valid_split_types}"
+        self.split_by = split_by
+
+    def _split_dataset(
+        self, dataset: DigitalTyphoonAnalysis
+    ) -> tuple[list[_SampleSequenceDict], ...]:
```
adamjstewart, line 46:
```suggestion
    ) -> tuple[list[_SampleSequenceDict], list[_SampleSequenceDict]]:
```
Tuple will always be length 2

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,435 @@
...
+
+class _SampleSequenceDict(TypedDict):
+    """Sample sequence dictionary."""
+
+    id: str
+    seq_id: list[int]
+
+
+class DigitalTyphoonAnalysis(NonGeoDataset):
+    """Digital Tyhphoon Dataset for Analysis Task.
```
adamjstewart, line 30:
```suggestion
    """Digital Typhoon Dataset for Analysis Task.
```

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,435 @@
...
+
+    This dataset contains typhoon-centered images, derived from hourly infrared channel
+    images captured by meteorological satellites. It incorporates data from multiple
+    generations of the Himawari weather satellite, dating back to 1978. These images
```
adamjstewart, line 34: Fun fact: ひまわり means "sunflower" in Japanese

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,435 @@
...
+    <http://agora.ex.nii.ac.jp/digital-typhoon/dataset/>`_
+    for more information about the dataset.
+
+    Dataset features:
+
+    * infrared channel images from the Himawari weather satellite (512x512 px)
+      at 5km spatial resolution
+    * auxiliary features such as wind speed, pressure, and more that can be used
+      for regression or classification tasks
+    * 1,099 typhoons and 189,364 image
```
adamjstewart, line 49:
```suggestion
    * 1,099 typhoons and 189,364 images
```

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,435 @@
...
+
+
+class _SampleSequenceDict(TypedDict):
+    """Sample sequence dictionary."""
+
+    id: str
+    seq_id: list[int]
+
+
+class DigitalTyphoonAnalysis(NonGeoDataset):
```
adamjstewart, line 29: Let's call it `DigitalTyphoon` to match the name of the dataset on the website.

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,435 @@
...
+        max_feature_value: Optional[dict[str, float]] = None,
+        transforms: Optional[Callable[[dict[str, Tensor]], dict[str, Tensor]]] = None,
+        download: bool = False,
+        checksum: bool = False,
+    ) -> None:
+        """Initialize a new Digital Typhoon Analysis dataset instance.
+
+        Args:
+            root: root directory where dataset can be found
+            task: whether to load "regression" or "classification" labels
```
adamjstewart, line 112: COWC has separate classes for classification and regression. But those datasets are marketed as two datasets. Not sure what is best here.

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,435 @@
...
+            features: which auxiliary features to return
+            target: which auxiliary features to use as targets
```
adamjstewart, line 114: Why is features plural but target is singular?

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,435 @@
...
+            max_feature_value: maximum value for each feature
+            transforms: a function/transform that takes input sample and its target as
+                entry and returns a transformed version
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+
+        Raises:
+            AssertionError: if ``task`` argument is invalid
+            RuntimeError: if ``download=False`` and data is not found, or checksums
+                don't match
```
adamjstewart, line 126:
```suggestion
            DatasetNotFoundError: If dataset is not found and *download* is False. 
```

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,435 @@
...
+        self.root = root
+        self.transforms = transforms
+        self.download = download
+        self.checksum = checksum
+        self.sequence_length = sequence_length
+
+        self.min_feature_value = min_feature_value
+        self.max_feature_value = max_feature_value
+
+        assert task in self.valid_tasks, f"Please choose one of {self.valid_tasks}"
```
adamjstewart, line 143: Error message could be made more clear as to which argument was wrong, but not required

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,435 @@
...
+        self.task = task
+
+        assert set(features).issubset(set(self.valid_features))
```
adamjstewart, line 146: I usually use `<=` instead of `.issubset(...)` but up to you

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,435 @@
...
+        if self.min_feature_value is not None:
+            for feature, min_value in self.min_feature_value.items():
+                self.aux_df = self.aux_df[self.aux_df[feature] >= min_value]
+
+        # Filter aux_df according to max_target_value
+        if self.max_feature_value is not None:
+            for feature, max_value in self.max_feature_value.items():
+                self.aux_df = self.aux_df[self.aux_df[feature] <= max_value]
+
+        def get_subsequences(df: pd.DataFrame, k: int) -> list[dict[str, list[int]]]:
```
adamjstewart, line 205: Can this be hidden? Don't want people to rely on it if not necessary

adamjstewart (commented):
This is clearly a very complicated dataset. I didn't really review the pandas stuff in detail. Since this is related to your ongoing time-series support work, I'll let you decide the best format for this dataset API.

docs/api/non_geo_datasets.csv
nilsleh: It can be both a regression and classification task and for different tasks (analysis, reanalysis and forecasting) https://github.com/kitamoto-lab/benchmarks the number of classes would also vary depending on the target variable

adamjstewart: Base class makes sense depending on how different they are. We can have a list or range for # classes, doesn't have to be a single number. We do that for some other columns.

torchgeo/datasets/digital_typhoon.py
nilsleh: under section 4.1 in their paper https://arxiv.org/pdf/2311.02665.pdf they list the difference inference tasks possible with this dataset, this implementation only covers the "Analysis" task so that is why I made that name choice

torchgeo/datasets/digital_typhoon.py
nilsleh: The idea here is that either you can:
a) take the wind speeds for example and regard them as continuous regression targets
b) or split the wind speed range into bins with each bin being a different class

adamjstewart:
Don't think I've ever used target normalization before, but if you have a random train/val/test split, you can either:

1. Use a fixed seed so that it's the same split every time, calculate stats only on train
2. Generate a random split, save it to disk, distribute on HF and combine with the dataset

1 is much easier, 2 is more formal.

nilsleh:
In case 1 the normalization would only be available throughtthe Datamodule and you would have to implement it in the on after batch transform and it would not be available for the dataset class

Case 2 is not possible I think, because the target range will change depending on the args for `min_feature_value` and `max_feature_value`

adamjstewart:
In both cases you can manually compute the normalization, then copy-n-paste it and store it in the dataset class. You don't need to compute it on the fly during training. The great thing about data modules is that they don't change too much. If there are parameters that select what features you are using, this is similar to which bands are used in the So2SatDataModule.

nilsleh:
But in case 2 I actually need to compute it on the fly because the regression target range changes based on the range restriction.

adamjstewart:
How is that different from case 1? The only thing that changes is whether the split is recorded on disk or not.

nilsleh:
It's not different from case 1. Since the target range can change and there are no defined train/test splits on the dataset class level, the target normalization needs to happen in the datamodule. But that would imply that the normalization needs to move to datamodule instead of dataset (where it is currently). So I just wanted to inquire about that :)

adamjstewart:
Gotcha. Yeah I would definitely move all transforms/data aug from the dataset to the data module to match our other datasets. Lack of a pre-defined train/val/test split shouldn't matter. Allowing the user to specify min/max feature values does matter. Do we need that? Can't we just compute that based on train and store it permanently with no option to override? Is that designed to serve the same purpose as mean/std?

nilsleh:
These cyclone dataset usually have a very imbalanced target distribution (there are few images of high hurricane categories) and the min feature values would allow the user to basically only select a dataset with only hurricane category images and not images that are just clouds with wind speed of 0 for example. I basically rewrote the Tropical Cylcone dataset locally to have that functionality because it makes running experiments a lot easier, so I thought I would put that option in this dataset as well.

adamjstewart:
Up to you I guess. You can use no normalization by default and allow the user to subclass and override the normalization values. Then the user is responsible for calculating mean/std themselves based on split/min/max. 

Or you can just subtract min and divide by (max - min). Is there any reason why this wouldn't be a good idea?

tests/datamodules/test_digital_typhoon.py
```diff
@@ -0,0 +1,75 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
adamjstewart, line 1: Is this file needed for coverage or can we delete it and rely on the trainer tests? I've been trying to slowly get rid of `tests/datamodules` since it doesn't actually do a good job of ensuring that the datamodules in fact work.

tests/datasets/test_digital_typhoon.py
```diff
@@ -0,0 +1,95 @@
...
+            (3, {'wind': 0}, {'pressure': 1500}),
+            (3, {'pressure': 0}, {'wind': 100}),
+        ]
+    )
+    def dataset(
+        self, monkeypatch: MonkeyPatch, tmp_path: Path, request: SubRequest
+    ) -> DigitalTyphoonAnalysis:
+        sequence_length, min_features, max_features = request.param
+        monkeypatch.setattr(
+            torchgeo.datasets.digital_typhoon, 'download_url', download_url
```
adamjstewart, line 37: See #2232, you can now get rid of this line and the above `download_url` function.

tests/datasets/test_digital_typhoon.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def test_already_downloaded(self, dataset: DigitalTyphoonAnalysis) -> None:
+        DigitalTyphoonAnalysis(root=dataset.root)
+
+    def test_not_yet_extracted(self, tmp_path: Path) -> None:
+        root = os.path.join('tests', 'data', 'digital_typhoon')
+        filenames = ['WP.tar.gzaa', 'WP.tar.gzab']
+        for filename in filenames:
+            shutil.copyfile(
+                os.path.join(root, filename), os.path.join(str(tmp_path), filename)
```
adamjstewart, line 80: It shouldn't be necessary to convert `tmp_path` into a string. In fact, you could use this instead:
```suggestion
                os.path.join(root, filename), tmp_path / filename
```

tests/datasets/test_digital_typhoon.py
```diff
@@ -0,0 +1,95 @@
...
+
+        url = os.path.join('tests', 'data', 'digital_typhoon', 'WP.tar.gz{0}')
+        monkeypatch.setattr(DigitalTyphoonAnalysis, 'url', url)
+
+        md5sums = {
+            'aa': '5b2fed45d9719e77a482ccd4ae1b02e5',
+            'ab': '5b2fed45d9719e77a482ccd4ae1b02e5',
+        }
+        monkeypatch.setattr(DigitalTyphoonAnalysis, 'md5sums', md5sums)
+        root = str(tmp_path)
```
adamjstewart, line 48:
```suggestion
        root = tmp_path
```
Same in other functions. TorchGeo now has full support for pathlib.Path in all datasets.

torchgeo/datamodules/digital_typhoon.py
```diff
@@ -0,0 +1,104 @@
...
+from ..datasets import DigitalTyphoonAnalysis
+from ..datasets.digital_typhoon import _SampleSequenceDict
+from .geo import NonGeoDataModule
+from .utils import group_shuffle_split
+
+
+class DigitalTyphoonAnalysisDataModule(NonGeoDataModule):
+    """Digital Typhoon Analysis Data Module."""
+
+    valid_split_types: ClassVar[list[str]] = ['time', 'typhoon_id']
```
adamjstewart, line 20:
```suggestion
    valid_split_types = ('time', 'typhoon_id')
```
No need to use a mutable type, tuples are faster and more memory efficient.

torchgeo/datamodules/digital_typhoon.py
```diff
@@ -0,0 +1,104 @@
...
+        Args:
+            sample_sequences: List of sample sequence dictionaries to be split
+
+        Returns:
+            a tuple of the subset datasets
+        """
+        if self.split_by == 'time':
+            sequences = list(enumerate(sample_sequences))
+
+            sorted_sequences = sorted(sequences, key=lambda x: x[1]['seq_id'])
```
adamjstewart, line 61: A comment explaining how this sort key was chosen or what is happening here may be good to add.

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,454 @@
...
+
+    If you use this dataset in your research, please cite the following papers:
+
+    * https://doi.org/10.20783/DIAS.664
+
+    .. versionadded:: 0.6
+    """
+
+    valid_tasks: ClassVar[list[str]] = ['classification', 'regression']
+    aux_file_name: ClassVar[str] = 'aux_data.csv'
```
adamjstewart, line 65: Let's convert lists to tuples. Strings aren't mutable, so no need to declare them as class variables. Same below. The only thing that should need `ClassVar` is `dict` when dicts are unavoidable.
```suggestion
    valid_tasks = ('classification', 'regression')
    aux_file_name = 'aux_data.csv'
```

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,454 @@
...
+        'dir30',
+        'long30',
+        'short30',
+        'landfall',
+        'intp',
+    ]
+
+    url: ClassVar[str] = (
+        'https://huggingface.co/datasets/torchgeo/digital_typhoon/resolve/main/WP.tar.gz{0}'
+    )
```
adamjstewart, line 89: Use a git commit hash for reproducibility:
```suggestion
    url = 'https://hf.co/datasets/torchgeo/digital_typhoon/resolve/cf2f9ef89168d31cb09e42993d35b068688fe0df/WP.tar.gz{0}'
```

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,454 @@
...
+    max_input_clamp = 300.0
+
+    data_root = 'WP'
+
+    def __init__(
+        self,
+        root: Path = 'data',
+        task: str = 'regression',
+        features: Sequence[str] = ['wind'],
+        targets: list[str] = ['wind'],
```
adamjstewart, line 106: Does it have to be a list, or will any Sequence work?

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,454 @@
...
+        sequence_length: int = 3,
+        min_feature_value: dict[str, float] | None = None,
+        max_feature_value: dict[str, float] | None = None,
```
adamjstewart, line 109: I wonder if these should have defaults, or if that should be left up to the user. Do the dataset authors use or prescribe any specific ranges?

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,454 @@
...
+            min_feature_value: minimum value for each feature
+            max_feature_value: maximum value for each feature
+            transforms: a function/transform that takes input sample and its target as
+                entry and returns a transformed version
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+
+        Raises:
+            AssertionError: if ``task`` argument is invalid
+            DatasetNotFoundError: If dataset is not found and *download* is False.
```
adamjstewart, line 131:
```
            DependencyNotFoundError: If h5py is not installed.                           
```

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,454 @@
...
+            sequence_length: length of the sequence to return
+            min_feature_value: minimum value for each feature
+            max_feature_value: maximum value for each feature
+            transforms: a function/transform that takes input sample and its target as
+                entry and returns a transformed version
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+
+        Raises:
+            AssertionError: if ``task`` argument is invalid
```
adamjstewart, line 130: Or a couple other arguments:
```suggestion
            AssertionError: If any arguments are invalid.
```

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,454 @@
...
+                    self.root,
+                    self.data_root,
+                    'metadata',
+                    str(sample_df.iloc[-1]['id']) + '.csv',
+                ),
+                sample_df.iloc[-1]['image_path'],
+            )
+        )
+
+        # torchgeo expects a single label
```
adamjstewart, line 272: We can fix this if needed. I'm trying to add support for multi-label classification in #2219

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,454 @@
...
+                image tensor
+            """
+            h5py = lazy_import('h5py')
+
+            full_path = os.path.join(self.root, self.data_root, 'image', id, filepath)
+            with h5py.File(full_path, 'r') as h5f:
+                # tensor with added channel dimension
+                tensor = torch.from_numpy(h5f['Infrared'][:]).unsqueeze(0)
+
+                # follow normalization procedure
```
adamjstewart, line 316: Wondering if this should be in the datamodule instead. Possibly also true for other stuff.

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,454 @@
...
+        Returns:
+            features for image
+        """
+        feature_df = pd.read_csv(filepath)
+        feature_df = feature_df[feature_df['file_1'] == image_path]
+        feature_dict = {
+            name: torch.tensor(feature_df[name].item()).float()
+            for name in self.features
+        }
+        # normalize the targets for regression
```
adamjstewart, line 349: Same here

docs/api/non_geo_datasets.csv
```diff
@@ -10,6 +10,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `Kenya Crop Type`_,S,Sentinel-2,"CC-BY-SA-4.0","4,688",7,"3,035x2,016",10,MSI
 `DeepGlobe Land Cover`_,S,DigitalGlobe +Vivid,-,803,7,"2,448x2,448",0.5,RGB
 `DFC2022`_,S,Aerial,"CC-BY-4.0","3,981",15,"2,000x2,000",0.5,RGB
+`Digital Typhoon Analysis`_,"C, R",Himawari,"CC-BY-4.0","189,364",8,512,5km,Infrared
```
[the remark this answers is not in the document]
adamjstewart: P.S. I confirmed the license from https://search.diasjp.net/en/dataset/DigitalTyphoon

adamjstewart (changes requested):

torchgeo/datasets/digital_typhoon.py
nilsleh: By default all values are included.

torchgeo/datasets/digital_typhoon.py
nilsleh: I think it has to be part of the dataset because the normalization values can change. In the datamodule it will become quiet messy, and overall I think it's much nicer if normalization happens in the dataset and any augmentations in the datamodule.

tests/datamodules/test_digital_typhoon.py
nilsleh: I just added it to test the splitting mechanism explicitly.

torchgeo/datasets/digital_typhoon.py
adamjstewart: Yes, I'm wondering if we should have defaults that strip out, e.g., negative wind speeds.

torchgeo/datasets/digital_typhoon.py
adamjstewart: Only complaint is that the user can't perform their own normalization, or if they do it is now duplicated. Definitely something worth discussing more broadly though (not just in your PR, for all datasets).

torchgeo/datasets/digital_typhoon.py
adamjstewart: I guess it depends on how different the other tasks are. If it's possible to write a single dataset for all 3 tasks, I would want to name the dataset DigitalTyphoon and expand it later.

torchgeo/datasets/digital_typhoon.py
```diff
@@ -0,0 +1,458 @@
...
+
+    * https://doi.org/10.20783/DIAS.664
+
+    .. versionadded:: 0.6
+    """
+
+    valid_tasks = ('classification', 'regression')
+    aux_file_name = 'aux_data.csv'
+
+    valid_features: ClassVar[list[str]] = [
```
adamjstewart, line 67: Can we make this a tuple?

adamjstewart:
Inspired by this rename I wasted 2 hrs of my life writing https://github.com/adamjstewart/dotfiles/blob/master/bin/git-sed

adamjstewart (approved):
Just need to get minimum tests passing now.

adamjstewart (approved):

PR #1756, opened 2023-12-04
Updates to SustainBenchCropYield dataset

### LINKED ISSUE #1754: SustainBenchCropYield reads all the data on __getitem__
Every call to `__getitem__` will result in the entire dataset being loaded from file -- https://github.com/microsoft/torchgeo/blob/main/torchgeo/datasets/sustainbench_crop_yield.py#L140.

We should check if the files are mmapped (and release mmapped version if not), or add caching to make this dataset faster.

calebrob6:
Fixes #1754 -- the whole dataset is only 291Mb so we can safely load all of this in the constructor then just access it in `__getitem__` (this also only takes a few seconds at most).

This PR also updates the dataset to download the "soybeans_updated.zip" file from https://drive.google.com/drive/folders/1hsp2PlXAgcQ0pbx_vvPKHZcj_Am3rWx4 which appears to be the version of the dataset that the benchmark results reported here https://sustainlab-group.github.io/sustainbench/leaderboard/#crop-yield-prediction are based off of.

@chrisyeh96 in case you are interested!

### DIFF
--- a/torchgeo/datasets/sustainbench_crop_yield.py
+++ b/torchgeo/datasets/sustainbench_crop_yield.py
@@ -146,9 +146,7 @@ class SustainBenchCropYield(NonGeoDataset):
         Returns:
             data and label at that index
         """
-        sample: dict[str, Tensor] = {
-            "image": self.images[index],
-        }
+        sample: dict[str, Tensor] = {"image": self.images[index]}
         sample.update(self.features[index])
 
         if self.transforms is not None:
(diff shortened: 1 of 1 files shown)


torchgeo/datasets/sustainbench_crop_yield.py
```diff
@@ -211,7 +180,7 @@ def _download(self) -> None:
         download_url(
             self.url,
             self.root,
-            filename=self.dir + ".zip",
+            filename=self.dir,
```
calebrob6, line 183:
```suggestion
            filename=self.dir + ".zip",
```

nilsleh (approved):

adamjstewart (approved):

PR #1759, opened 2023-12-06
Update labeler.yml for labeler v5

isaaccorley:


adamjstewart:
Let me know if you want me to hack on this. I'm not sure whether the modified `.github/labeler.yml` in this PR is used or its using the one from main instead. So we may have to try something we are confident in, merge, then see if it worked. Wish they had a validator or something.

adamjstewart (approved):
It's definitely not reading the PR config file, only the config file from main, which makes sense for security reasons. I'll merge and see if these changes worked.

isaaccorley:
Yup I see now it needed the permissions. Nice work

PR #1760, opened 2023-12-06
Add change detection task

robmarkcole:
Add change detection task with 3 models implemented initially. Almost identical to the Segmentation trainer. Points to discuss:

1) Metrics - currently using [MulticlassAccuracy](https://lightning.ai/docs/torchmetrics/stable/classification/accuracy.html#torchmetrics.classification.MulticlassAccuracy) which is not ideal, since most of the time in binary change detection we want the metrics for the change class only. UPDATE: use [ClasswiseWrapper](https://lightning.ai/docs/torchmetrics/stable/wrappers/classwise_wrapper.html)
2) Use FocalJaccardLoss loss as default? Suggested by @isaaccorley 
3) I am using 4 GPU, resulting in the error: "It is recommended to use `self.log('val_loss', ..., sync_dist=True)` when logging on epoch level in distributed setting to accumulate the metric across devices." Apply this change?


adamjstewart:
Where are we at with this PR? We're not planning 0.6 just yet, but I would love to get this in soon. There are a lot of people using TorchGeo for change detection who could benefit from this.

robmarkcole:
@isaaccorley has some improvements to contribute

PR #1762, opened 2023-12-06
Check if labeler is working again

adamjstewart:


adamjstewart:
@isaaccorley it's working again now, thanks for the help!

PR #1769, opened 2023-12-12
LEVIR-CD+ bug fixes

isaaccorley:
This PR fixes the following bugs in the LEVIR-CD+ dataset and datamodule.

- The plot method plots the ground truth in a different cmap than a predicted mask
- Cleaned up some image normalization code which basically replicates `torchgeo.datasets.utils.percentile_normalization`
- Added `val_augs` and `test_augs` to the LEVIR-CD+ datamodule so random cropping isn't used for computing evaluation metrics

New example plot:

output

### DIFF
--- a/tests/datamodules/test_levircd.py
+++ b/tests/datamodules/test_levircd.py
@@ -86,8 +86,6 @@ class TestLEVIRCDPlusDataModule:
         batch = next(iter(datamodule.val_dataloader()))
         batch = datamodule.on_after_batch_transfer(batch, 0)
         if datamodule.val_split_pct > 0.0:
-            for k, v in batch.items():
-                print(k, v.shape, v.dtype)
             assert (
                 batch["image1"].shape[-2:] == batch["mask"].shape[-2:] == (1024, 1024)
             )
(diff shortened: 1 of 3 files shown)


isaaccorley:
I'm actually not sure how the LEVIR-CD+ datamodule tests were passing because it's random cropping 256x256 patches from 1x1 size dummy images. Added in a resize transform in the test to fix this.

tests/datamodules/test_levircd.py
```diff
@@ -31,7 +55,12 @@ def datamodule(
 
         root = str(tmp_path)
         dm = LEVIRCDPlusDataModule(
-            root=root, download=True, num_workers=0, checksum=True, val_split_pct=0.5
+            root=root,
+            download=True,
+            num_workers=0,
+            checksum=True,
+            val_split_pct=0.5,
```
robmarkcole, line 62: 50% seems very high..?

isaaccorley:
Closing because https://github.com/microsoft/torchgeo/pull/1770 includes these changes

adamjstewart:
Can you submit a PR against the releases/v0.5 branch to at least fix the mask plotting? Then we can backport it to 0.5.2.

PR #1770, opened 2023-12-13
LEVIR-CD Dataset and Datamodule

isaaccorley:
This PR adds the LEVIR-CD dataset and datamodule.

This dataset is just the prior version of LEVIR-CD+ with fewer images. It also has a slightly different directory structure. 

Also refactored the datasets to inherit from a general base LEVIRCD class


nilsleh (changes requested):
I think the docs also need to be updated with the new addition datasets.

tests/datamodules/test_levircd.py
```diff
@@ -26,12 +50,17 @@ def datamodule(
         monkeypatch.setattr(torchgeo.datasets.utils, "download_url", download_url)
         md5 = "1adf156f628aa32fb2e8fe6cada16c04"
```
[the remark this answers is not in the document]
isaaccorley: Added

torchgeo/datasets/levircd.py
```diff
@@ -262,3 +184,202 @@ def get_rgb(img: Tensor) -> "np.typing.NDArray[np.uint8]":
...
+    * masks are single-channel pngs where no change = 0, change = 255
+
+    Dataset classes:
+
+    1. no change
+    2. change
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://arxiv.org/abs/2107.09244
```
nilsleh, line 333: I am not sure about this, but does this need some kind of versionchanged?

nilsleh (commented):
Otherwise LGTM

tests/data/levircd/levircdplus/data.py
```diff
@@ -0,0 +1,61 @@
...
+import os
+import shutil
+
+import numpy as np
+from PIL import Image
+
+np.random.seed(0)
+
+
+def create_image(path: str) -> None:
```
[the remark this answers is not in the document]
isaaccorley: It's just a test data creation script per dataset. We don't have any shared utils for this.

nilsleh (approved):

docs/api/datamodules.rst
```diff
@@ -85,6 +85,16 @@ LandCover.ai
 
 .. autoclass:: LandCoverAIDataModule
 
+LEVIR-CD
```
adamjstewart, line 88: I'd prob just have a single header with 2 versions

docs/api/datasets.rst
```diff
@@ -267,6 +267,12 @@ LandCover.ai
 
 .. autoclass:: LandCoverAI
 
+LEVIR-CD
```
adamjstewart, line 270: Same here

docs/api/non_geo_datasets.csv
```diff
@@ -18,6 +18,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `IDTReeS`_,"OD,C",Aerial,"CC-BY-4.0",591,33,200x200,0.1--1,RGB
 `Inria Aerial Image Labeling`_,S,Aerial,-,360,2,"5,000x5,000",0.3,RGB
 `LandCover.ai`_,S,Aerial,"CC-BY-NC-SA-4.0","10,674",5,512x512,0.25--0.5,RGB
+`LEVIR-CD`_,CD,Google Earth,-,637,2,"1,024x1,024",0.5,RGB
 `LEVIR-CD+`_,CD,Google Earth,-,985,2,"1,024x1,024",0.5,RGB
```
adamjstewart, line 22: Could prob just combine

tests/datamodules/test_levircd.py
adamjstewart, line 1: This file will eventually be removed once we have a trainer for CD

torchgeo/datasets/levircd.py
```diff
@@ -262,3 +184,204 @@ def get_rgb(img: Tensor) -> "np.typing.NDArray[np.uint8]":
...
+
+    Dataset format:
+
+    * images are three-channel pngs
+    * masks are single-channel pngs where no change = 0, change = 255
+
+    Dataset classes:
+
+    1. no change
+    2. change
```
adamjstewart, line 234: Does our `__getitem__` remap from 0, 255 to 0, 1? Should it? We should label the classes with the returned value, not the stored value.

torchgeo/datasets/levircd.py
```diff
@@ -262,3 +184,204 @@ def get_rgb(img: Tensor) -> "np.typing.NDArray[np.uint8]":
...
+                filename=self.splits[split]["filename"],
+                md5=self.splits[split]["md5"] if self.checksum else None,
+            )
+
+
+class LEVIRCDPlus(LEVIRCDBase):
+    """LEVIR-CD+ dataset.
+
+    The `LEVIR-CD+ <https://github.com/S2Looking/Dataset>`__
+    dataset is a dataset for building change detection.
```
adamjstewart, line 312: We should change all:

> The _foo_ dataset is a dataset for _bar_

to:

> _Foo_ is a _bar_ dataset

It just reads better.

torchgeo/datasets/levircd.py
```diff
@@ -262,3 +184,204 @@ def get_rgb(img: Tensor) -> "np.typing.NDArray[np.uint8]":
...
+    Dataset classes:
+
+    1. no change
+    2. change
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://arxiv.org/abs/2107.09244
+
+    .. versionchanged:: 0.6
```
adamjstewart, line 335: Not true

torchgeo/datasets/levircd.py
```diff
@@ -174,31 +131,6 @@ def _load_target(self, path: str) -> Tensor:
...
-            return
-
-        download_and_extract_archive(
-            self.url,
-            self.root,
-            filename=self.filename,
-            md5=self.md5 if self.checksum else None,
-        )
-
     def plot(
```
adamjstewart, line 134: Example plot for both versions?

torchgeo/datamodules/levircd.py
```diff
@@ -10,12 +10,54 @@
...
+            _RandomNCrop(self.patch_size, batch_size),
+            data_keys=["image1", "image2", "mask"],
+        )
+        self.val_aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
+            data_keys=["image1", "image2", "mask"],
+        )
+        self.test_aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
+            data_keys=["image1", "image2", "mask"],
```
adamjstewart, line 57: If you set `self.aug` you don't need to duplicate this

torchgeo/datasets/levircd.py
isaaccorley: Why would this not be true? We already had a LEVIRCDPlus class and this PR changed it. There are no breaking changes to it.

adamjstewart: Ah, I read this as versionadded the first time. Still, we should say _what_ changed. Or did the API not change?

isaaccorley: I can add that in another PR. I just made a base class so I could override a couple methods for loading the files since they have different directory setups. The LEVIRCDPlus itself hasn't changed though other than inheriting from the new base class

adamjstewart: In that case I would remove versionchanged, there are no significant API differences that users need to be aware of.

PR #1772, opened 2023-12-13
ExtractPatches augmentation

isaaccorley:
When using the `_RandomNCrop` transform during training as an augmentation we need to be able to perform evaluation on the val and test sets on non overlapping patches of the same crop size.

While kornia.contrib contains an `ExtractPatches` transform, it unfortunately is not an augmentation therefore it a) doesn't work on masks in `K.AugmentationSequential` and b) has no option to combine the resulting patch dimension back into the batch dimensions for inference e.g. `b p c h w -> (b p) c h w` which breaks any downstream processing after the augmentation.

This PR adds the `_ExtractPatches` transform which can be used in the `test_aug` and `val_aug` augmentations in datamodules for evaluation on patches of images which aren't geotiffs (jpg, png, etc.), e.g. LEVIR-CD. 

Example:
```python
from torchgeo.transforms import AugmentationSequential
from torchgeo.transforms.transforms import _ExtractPatches`

augs = AugmentationSequential(
   K.Normalize(mean=0.0, std=255.0),
   _ExtractPatches(window_size=256, stride=256, padding=0, keepdim=True),
   same_on_batch=True
   data_keys=["image1", "image2", "mask"]
)
```
One thing that's important is to use the `same_on_batch=True` option in `AugmentationSequential` which will allow the batch dimension to be combined.

I've opened another PR https://github.com/kornia/kornia/pull/2685 in kornia but not sure how long that will take to get merged.

### DIFF
--- a/tests/transforms/test_transforms.py
+++ b/tests/transforms/test_transforms.py
@@ -7,6 +7,7 @@ import torch
 from torch import Tensor
 
 from torchgeo.transforms import indices, transforms
+from torchgeo.transforms.transforms import _ExtractPatches
 
 # Kornia is very particular about its boxes:
 #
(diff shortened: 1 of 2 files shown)


isaaccorley:
@robmarkcole

nilsleh (approved):

torchgeo/transforms/transforms.py
```diff
@@ -179,3 +180,77 @@ def forward(
             "input_size": out[0]["input_size"],
             "output_size": out[0]["output_size"],
         }
+
+
+class _ExtractPatches(K.GeometricAugmentationBase2D):
```
nilsleh, line 185: Only missing a test

nilsleh (changes requested):

torchgeo/transforms/transforms.py
isaaccorley: I don't see any tests for _RandomNCrop btw. Not sure how coverage is passing

nilsleh (approved):

tests/transforms/test_transforms.py
```diff
@@ -201,3 +202,56 @@ def test_sequential_transforms_augmentations(
...
     assert_matching(output, expected)
+
+
+def test_extract_patches() -> None:
+    b, c, h, w = 2, 3, 64, 64
+    p = 32
+    s = p
+    num_patches = ((h - p + s) // s) * ((w - p + s) // s)
+
+    # test default settings (when stride is not defined, s=p)
```
adamjstewart, line 213: Lot of duplicated code here, can these be refactored into parametrized pytest args?

PR #1777, opened 2023-12-15
Remov Argmax Computation for torchmetrics in Classification and Segmentation

nilsleh:
Torchmetrics automatically applies the argmax on logits already if the datatype is float (so the logits) for the Classification Metrics https://lightning.ai/docs/torchmetrics/stable/classification/accuracy.html#multiclassaccuracy.

However, some metrics like [Calibration Error](https://lightning.ai/docs/torchmetrics/stable/classification/calibration_error.html#multiclasscalibrationerror) expect the logits. So in order to overwrite the configuration of metrics without having to change the train, val , test steps, this PR removes the argmax computation.

### DIFF
--- a/torchgeo/trainers/classification.py
+++ b/torchgeo/trainers/classification.py
@@ -196,7 +196,7 @@ class ClassificationTask(BaseTask):
             and hasattr(self.logger.experiment, "add_figure")
         ):
             datamodule = self.trainer.datamodule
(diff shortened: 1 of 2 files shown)


calebrob6 (approved):

adamjstewart (approved):
You learn something new every day.

PR #1781, opened 2023-12-19
Fix for pre-commit

ashnair1:
Type ignores added in 135f564837655ebf0146325e973d9336a998073e were not reflected in the pre-commit environment since `scikit-image` was not declared a dependency. This caused the `mypy` hook to fail with the following message:

```
pyupgrade................................................................Passed
isort....................................................................Passed
black....................................................................Passed
flake8...................................................................Passed
pydocstyle...............................................................Passed
- hook id: mypy
- exit code: 1

torchgeo/datasets/vhr10.py:476: error: Unused "type: ignore" comment  [unused-ignore]
torchgeo/datasets/vhr10.py:528: error: Unused "type: ignore" comment  [unused-ignore]
Found 2 errors in 1 file (checked 292 source files)
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,5 +34,5 @@ repos:
         hooks:
         -   id: mypy
             args: [--strict, --ignore-missing-imports, --show-error-codes]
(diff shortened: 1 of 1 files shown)


adamjstewart:
Can we also add pyvista to solve the vhr10 issue?

adamjstewart (approved):

PR #1793, opened 2024-01-02
pytest: don't cast warnings to errors

adamjstewart:
Previously, all warnings were cast to errors when running pytest. This helped uncover the use of deprecated features, allowing us to stay on top of deprecations and ensure support for future versions of dependencies.

However, this also meant that minor platform-specific or version-specific warning messages would cause pytest to fail. Although it's maintainable in CI, it's impossible to support all the platforms and versions our developers are using.

This PR disables the casting of warning messages to errors. We'll need to be careful to make sure the release branch is relatively warning-free before making a release, but this will make life much easier for our developers.

@yichiac

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -197,9 +197,6 @@ match_dir = "(datamodules|datasets|losses|models|samplers|torchgeo|trainers|tran
 addopts = "-m 'not slow'"
 # https://docs.pytest.org/en/latest/how-to/capture-warnings.html
 filterwarnings = [
-    # Treat all warnings as errors
-    "error",
-
     # Warnings raised by dependencies of dependencies, out of our control
     # https://github.com/Cadene/pretrained-models.pytorch/issues/221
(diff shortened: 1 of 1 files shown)


adamjstewart:
Related to https://github.com/pytest-dev/pytest/issues/10613

calebrob6:
This makes life easier for developers as they don't need to use the CI versions of all of our dependencies?

adamjstewart:
The versions in `requirements/*.txt` change every day. But even using those exact same versions is no guarantee that you won't have warnings. For example, PyTorch raises some warnings on MPS when it falls back on the CPU, but these warnings don't occur in CI because the hardware is emulated and PyTorch isn't installed with MPS support. @yichiac encountered an issue recently where torchvision installed by Spack would issue a warning if png/jpeg support wasn't also installed. All of these warnings get promoted to errors and the tests fail locally.

calebrob6 (approved):

calebrob6:
Makes sense!

PR #1740, opened 2023-11-23
Added the NCCM datamodule

shreyakannan1205:
The purpose of this PR is to added the datamodule for the Northeastern China Crop Map dataset.

PR #1800, opened 2024-01-09
Create Sentinel-2 Test Data for NCCM Datamodule

yichiac:
This PR created a set of test data for Sentinel-2 and NCCM to ensure two datasets overlap. The new test data can be used for the development of NCCM datamodule in [#1740](https://github.com/microsoft/torchgeo/pull/1740). 

@shreyakannan1205


calebrob6 (dismissed):
LGTM

adamjstewart:
Alternative is to copy the geographic metadata from the existing S2 data to NCCM, CDL, and all other datasets we want to combine with S2.

adamjstewart:
But I'm not particularly concerned about having too much fake data if you just want to stick with this approach.

tests/data/nccm/data.py
```diff
@@ -24,15 +24,8 @@ def create_file(path: str, dtype: str):
...
-        "transform": Affine(
-            8.983152841195208e-05,
-            0.0,
-            115.483402043364,
-            0.0,
-            -8.983152841195208e-05,
-            53.531397320113605,
-        ),
+        "crs": CRS.from_epsg(32616),
+        "transform": Affine(10, 0.0, 399960.0, 0.0, -10, 4500000.0),
```
adamjstewart, line 28: If you copy the CRS/transform from the existing S2 fake data, you no longer need to create new fake data for S2. So you can get rid of all the new files in this PR and only update existing files.

PR #1806, opened 2024-01-13
README: add podcast episode

adamjstewart:
Not sure why we didn't already have a link to this. I just rewatched it and it's still largely up to date.

Also changed our paper link to the arXiv version, which is easier to read for free. We still suggest citing the ACM version though.

@robmarkcole

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,8 +37,8 @@ For [conda](https://docs.conda.io/) and [spack](https://spack.io/) installation 
 
 You can find the documentation for TorchGeo on [ReadTheDocs](https://torchgeo.readthedocs.io). This includes API documentation, contributing instructions, and several [tutorials](https://torchgeo.readthedocs.io/en/stable/tutorials/getting_started.html). For more details, check out our [paper](https://arxiv.org/abs/2111.08872), [podcast episode](https://www.youtube.com/watch?v=ET8Hb_HqNJQ), [tutorial](https://www.youtube.com/watch?v=R_FhY8aq708), and [blog post](https://pytorch.org/blog/geospatial-deep-learning-with-torchgeo/).
 
-[![podcast episode](https://img.youtube.com/vi/ET8Hb_HqNJQ/0.jpg)](https://www.youtube.com/watch?v=ET8Hb_HqNJQ =45%x)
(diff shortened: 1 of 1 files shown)


robmarkcole:
It is possible to embed the YouTube in the readme, if of interest

robmarkcole:
See end of https://github.com/robmarkcole/HASS-amazon-rekognition

adamjstewart:
Up to @calebrob6 if he wants his face on the README

robmarkcole:
It will display the video thumbnail,which can be whatever you like

nilsleh (dismissed):

calebrob6:
I don't mind my face there, but I'd rather not have _just_ my face there -- if we have a few more videos with different people then I'm down to include the thumbnails.

I have had this bad assumption that tutorial videos aren't that useful as I don't ever specifically search/click through to youtube tutorials for programming topics (strongly prefer articles/notebooks). However, Qiusheng Wu's videos are great and get a ton of traffic and I've gotten several email questions about this video https://www.youtube.com/watch?v=R_FhY8aq708

adamjstewart:
How does this look? https://github.com/adamjstewart/torchgeo/tree/docs/readme?tab=readme-ov-file#documentation

Unfortunately GitHub README files don't support embeddings, but could take screenshots like: 
Screenshot 2024-01-22 at 12 14 39
and use those images instead so it's more clear that it's clickable.

calebrob6:
I can actually see it if I click on the README.md file itself

https://github.com/adamjstewart/torchgeo/blob/docs/readme/README.md

image

adamjstewart:
I don't yet have any videos with my face to share. I'm not worried about getting credit, I'll prob give an updated talk with @robmarkcole someday (maybe after we finish the next big exciting project and add some cool new features). Once we do that, we can add that video as well. I do think pictures of the videos are better than no pictures, although I really wish we could use the full embedded view that makes it clear it's clickable. Just need to decide where to store the images.

robmarkcole:
Another option is to make a completely new video, just a quick intro to torchgeo, and use that

adamjstewart:
I might have time for that in about a month 😆 

Could also change the thumbnail to remove Caleb's face, but I personally think it looks good 😉. Shows that the library is developed by humans, not companies. As long as Caleb doesn't get all the blame when something is broken.

calebrob6 (approved):

calebrob6:
I need to get my camera setup sorted

PR #1810, opened 2024-01-16
Fix the default folds and assertion for checking valid folds in PASTIS

calebrob6:
The unique folds in the `metadata.geojson` file in the PASTIS dataset are {1,2,3,4,5}.


adamjstewart (approved):
Good catch

PR #1811, opened 2024-01-17
Clarify RasterDataset documentation for is_image and dtype

dmeaux:
The documentation for the `is_image` and `dtype` properties of `RasterDataset` were vague in regards to what to do for surface models, like DEMs. The changes proposed attempt to clarify how one should go about treating surface models for different circumstances when creating a custom RasterDataset.


docs/tutorials/custom_raster_dataset.ipynb
```diff
@@ -329,7 +329,11 @@
     "\n",
     "### `is_image`\n",
     "\n",
-    "If your data only contains image files, as is the case with Sentinel-2, use `is_image = True`. If your data only contains segmentation masks, use `is_image = False` instead.\n",
+    "If your dataset only contains source data, such as image files, like Sentinel-2, or a digital surface, like a Digital Elevation Model, Digital Surface Model, Digital Terrain Model, or a raster of temperature values, use `is_image = True`. If your dataset only contains target data, such as a segmentation mask, like land use or land cover classification, use `is_image = False` instead.\n",
```
adamjstewart, line 332: Let's leave out the part about DEM/DSM/DTM/temp for now until we decide whether it should be image or mask

docs/tutorials/custom_raster_dataset.ipynb
```diff
@@ -329,7 +329,11 @@
...
+    "\n",
+    "### `dtype`\n",
+    "\n",
+    "Defaults to float32 for `is_image == True` and long for `is_image == False`. This is what is usually wanted for 99% of datasets but can be overridden for pixel-wise regression masks (where the target should be float32). Uint16 and uint32 are automatically cast to int32 and int64, respectively, because numpy supports the former but torch does not.\n",
```
adamjstewart, line 336: Let's remove the last sentence. This conversion happens before `dtype` is used and only matters for raw files.

torchgeo/datasets/geo.py
```diff
@@ -342,7 +346,11 @@ class RasterDataset(GeoDataset):
     #: ``start`` and ``stop`` groups.
     date_format = "%Y%m%d"
 
-    #: True if dataset contains imagery, False if dataset contains mask
+    #: True if the dataset contains source data, such as imagery. False if the dataset
+    #: contains target data, such as a mask. This is the same as Kornia. When multiple
```
adamjstewart, line 350: > This is the same as Kornia

It isn't clear _what_ is the same here. The "image" and "mask" keys are the same, but you don't mention those until later.

torchgeo/datasets/geo.py
```diff
@@ -361,6 +369,12 @@ class RasterDataset(GeoDataset):
     def dtype(self) -> torch.dtype:
         """The dtype of the dataset (overrides the dtype of the data file via a cast).
 
+        Defaults to float32 for is_image = True and long for is_image = False. This is
+        what we usually want for 99% of datasets but can be overridden for pixel-wise 
+        regression masks (where it should be float32). Uint16 and uint32 are
+        automatically cast to int32 and int64, respectively, because numpy supports
+        the former but torch does not.
```
adamjstewart, line 376: Remove last sentence here

adamjstewart (changes requested):

torchgeo/datasets/geo.py
dmeaux: How about this:
> True if the dataset contains source data, such as imagery. False if the dataset contains target data, such as a mask. In order to maintain consistency with Kornia, if ``is_image`` is set to ``True``, ``__getitem__`` uses ``image`` as the key for the data; otherwise, ``__getitem__`` uses ``mask`` as the key.
?

torchgeo/datasets/geo.py
```diff
@@ -342,7 +346,11 @@ class RasterDataset(GeoDataset):
     #: ``start`` and ``stop`` groups.
     date_format = "%Y%m%d"
 
-    #: True if dataset contains imagery, False if dataset contains mask
+    #: True if the dataset contains source data, such as imagery. False if the dataset
+    #: contains target data, such as a mask. This is the same as Kornia. When multiple
+    #: datasets with different keys are combined and the same key is used for multiple
+    #: datasets, for example 2 "image" and 1 "mask", the channels will be stacked so
+    #: that there's still a single value for that key.
```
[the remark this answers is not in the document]
dmeaux: Does this work:
> The default behavior when multiple datasets with different keys are combined and the same  key is used for multiple datasets, for example 2 "image" and 1 "mask", is that the channels will be stacked so that there's still a single value for that key. However, this behavior can be changed by specifying adifferent ``collate_fn``.

? Or is the documentation now too long for you?

torchgeo/datasets/geo.py
dmeaux: Done.

torchgeo/datasets/geo.py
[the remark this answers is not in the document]
dmeaux: I committed the change for your review.

torchgeo/datasets/geo.py
dmeaux: I committed the change for your review.

adamjstewart:
You can also include the CSV changes if you want: "image -> DEM"

adamjstewart (approved):
I handled the last remaining change requests and improved the formatting a bit, hope you don't mind. Wanted to squeeze this into the next release (coming soon!)

PR #1813, opened 2024-01-18
Add EuroCrops dataset

favyen2:
This adds the EuroCrops dataset (https://www.eurocrops.tum.de/index.html).

There are currently a few issues that I wasn't sure how to deal with

- The integer label does not directly correspond to a property of the vector features, but instead some mapping is needed. Currently I replaced `VectorDataset.label_name` with `label_fn: Callable[[fiona.Feature], int]` but we could do it another way instead.
- The vector labels are provided by year but VectorDataset does not seem to support timestamps. Should we add a way to get timestamp in VectorDataset, or wrap `__getitem__`?
- There are more than 255 classes in EuroCrops so `rasterio.features.rasterize` produces uint16 output instead of uint8. pytorch doesn't support uint16 so VectorDataset needs updates for this to work.
- EuroCrops has a class hierarchy (Level 3, 4, 5, 6) where higher levels are more fine-grained (see https://www.eurocrops.tum.de/taxonomy.html). Currently `EuroCrops(...)` accepts an `hcat_level` argument specifying the degree of granularity desired. Another option would be to return multiple masks, one for each level. A complicating factor is that some labels are not annotated at the leaves of the hierarchy but instead at intermediate nodes, presumably when the fine-grained label is not known, so if user requests Level 6 they may get some labels as "common_soft_wheat" and some as "winter_common_soft_wheat" (where "winter_common_soft_wheat" is a child of "common_soft_wheat" in the hierarchy). How do we want to deal with this? These labels could be leveraged by adding up the probabilities of the descendant categories when computing cross entropy loss, but I'm not sure if this training mode is or should be supported.

Here is example output in Portugal paired with Sentinel-2 imagery via intersection dataset:

im1
im2


torchgeo/datasets/geo.py
```diff
@@ -706,7 +706,16 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
             # with the default fill value and dtype used by rasterize
             masks = np.zeros((round(height), round(width)), dtype=np.uint8)
 
-        sample = {"mask": torch.tensor(masks), "crs": self.crs, "bbox": query}
+        # Convert masks from numpy to torch.
```
favyen2, line 709: I realized that this functionality is already present in https://github.com/microsoft/torchgeo/blob/1c20678a1001805412595be9759dae18a5d956df/torchgeo/datasets/geo.py#L547. Should I move the code to a separate function?

favyen2:
@microsoft-github-policy-service agree company="Allen Institute for AI"

torchgeo/datasets/geo.py
adamjstewart: Yes, this makes sense as a separate function. It can go in `torchgeo/datasets/utils.py`.

adamjstewart:
> The vector labels are provided by year but VectorDataset does not seem to support timestamps. Should we add a way to get timestamp in VectorDataset, or wrap `__getitem__`?

Assuming the year is found in the filename, this will be implemented by #1814

torchgeo/datasets/eurocrops.py
```diff
@@ -0,0 +1,285 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""EuroCrops dataset."""
+
+import csv
```
adamjstewart, line 6: We also have a pandas dependency if you find that easier to work with, but csv is also fine for simple things.

torchgeo/datasets/eurocrops.py
```diff
@@ -0,0 +1,285 @@
...
+
+
+class EuroCrops(VectorDataset):
+    """EuroCrops Dataset (Version 9).
+
+    The `EuroCrops <https://www.eurocrops.tum.de/index.html>`__ dataset combines "all
+    publicly available self-declared crop reporting datasets from countries of the
+    European Union" into a unified format. The dataset is released under CC BY 4.0 Deed.
+
+    The dataset consists of shapefiles containing a total of X polygons. Each polygon
```
adamjstewart, line 59: I assume X is a placeholder you plan on filling in

torchgeo/datasets/eurocrops.py
```diff
@@ -0,0 +1,285 @@
...
+        ("PT.zip", 2021, "3dba9c89c559b34d57acd286505bcb66"),
+        ("SE_2021.zip", 2021, "cab164c1c400fce56f7f1873bc966858"),
+        ("SI_2021.zip", 2021, "6b2dde6ba9d09c3ef8145ea520576228"),
+        ("SK_2021.zip", 2021, "c7762b4073869673edc08502e7b22f01"),
+        # Year is unknown for Romania portion (ny = no year).
+        # ("RO_ny", ??),
+    ]
+
+    # Color palette to choose from.
+    # There are hundreds of classes so we pick color via modulo of class index.
```
adamjstewart, line 101: Does the dataset offer any recommended cmap? I know CDL does but don't know much about EuroCrops

torchgeo/datasets/eurocrops.py
```diff
@@ -0,0 +1,285 @@
...
+        (171, 108, 40, 255),
+        (184, 217, 235, 255),
+        (108, 159, 184, 255),
+    ]
+
+    def __init__(
+        self,
+        paths: Union[str, Iterable[str]] = "data",
+        crs: Optional[CRS] = None,
+        res: float = 0.00001,
```
adamjstewart, line 125: This may need to be changed depending on the units of the CRS. A default with a similar resolution as Sentinel-2 would make sense to me.

torchgeo/datasets/eurocrops.py
```diff
@@ -0,0 +1,285 @@
...
+        transforms: Optional[Callable[[dict[str, Any]], dict[str, Any]]] = None,
+        download: bool = False,
+        checksum: bool = False,
+        hcat_level: int = 6,
+    ) -> None:
+        """Initialize a new Dataset instance.
```
adamjstewart, line 131:
```suggestion
        """Initialize a new EuroCrops instance.
```

torchgeo/datasets/eurocrops.py
```diff
@@ -0,0 +1,285 @@
...
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+            hcat_level: use labels at this level of the
+                `HCATv2 hierarchy <https://www.eurocrops.tum.de/taxonomy.html>`__.
+                Valid levels are 3, 4, 5, and 6.
+
+        Raises:
+            DatasetNotFoundError: If dataset is not found and *download* is False.
+
+        .. versionchanged:: 0.5
+           *root* was renamed to *paths*.
```
adamjstewart, line 150: You can remove this. This dataset was not added until 0.6, so the `root` parameter was never renamed for this dataset.

torchgeo/datasets/eurocrops.py
```diff
@@ -0,0 +1,285 @@
...
+            download_and_extract_archive(
+                self.base_url + fname, self.paths, md5=md5 if self.checksum else None
+            )
+
+    def _load_classes(self) -> None:
+        """Load classes from the HCAT CSV file.
+
+        The classes are loaded corresponding to the hcat_level specified by the user.
+        """
+        assert isinstance(self.paths, str)
```
adamjstewart, line 216: Seems like `self.paths` can never be a list of paths, is that correct? Not sure if it's possible to relax this constraint or if we should just change the type of the initial parameter.

torchgeo/datasets/eurocrops.py
```diff
@@ -0,0 +1,285 @@
...
+            sample: a sample returned by :meth:`VectorDataset.__getitem__`
+            show_titles: flag indicating whether to show titles above each panel
+            suptitle: optional string to use as a suptitle
+
+        Returns:
+            a matplotlib Figure with the rendered sample
+
+        .. versionchanged:: 0.3
+            Method now takes a sample dict, not a Tensor. Additionally, it is possible
+            to show subplot titles and/or use a custom suptitle.
```
adamjstewart, line 256: You can remove this, the dataset didn't exist in version 0.3

torchgeo/datasets/geo.py
```diff
@@ -605,8 +605,8 @@ def __init__(
             res: resolution of the dataset in units of CRS
             transforms: a function/transform that takes input sample and its target as
                 entry and returns a transformed version
-            label_name: name of the dataset property that has the label to be
-                rasterized into the mask
+            label_fn: function to extract the label to be rasterized into the mask from
+                the dataset Feature.
```
adamjstewart, line 609: Will think about this. I'm not opposed, but want to make sure lambdas aren't too complicated for remote sensing researchers with limited Python skills.

adamjstewart (commented):
This looks great! Thanks for getting this done so quickly.

A couple remaining things:

1. Needs to be added to the documentation (the last 2 bullet points of https://torchgeo.readthedocs.io/en/stable/user/contributing.html#datasets)
2. We might need some way to select a subset of classes so we can train a model that predicts one of (maize, rice, soybean, wheat, other). This could also be done in the Lightning data module (next step). @yichiac have we done this yet for any of our other datasets? This is similar to the `classes` parameter of the CDL and NLCD datasets.

P.S. Curious to know how you're finding TorchGeo so far. VectorDataset is one of our underdeveloped features due to lack of example datasets, so the caveats you ran into are somewhat expected.

torchgeo/datasets/eurocrops.py
favyen2: Ah planned to fill it in before committing, fixed now!

torchgeo/datasets/eurocrops.py
favyen2: The closest I found is this but it just visualizes the classes on a color scale:
https://github.com/maja601/EuroCrops/blob/main/webinar_demo/visualization.ipynb

torchgeo/datasets/eurocrops.py
favyen2: The files range in CRS from WGS-84 to EPSG:31287 so I have set a default CRS (WGS-84), with res equivalent to 10 m/pixel at equator.

favyen2:
> Needs to be added to the documentation (the last 2 bullet points of https://torchgeo.readthedocs.io/en/stable/user/contributing.html#datasets)

Thanks this helps a lot. I updated the relevant docs files.

> We might need some way to select a subset of classes so we can train a model that predicts one of (maize, rice, soybean, wheat, other).

I added a classes argument now. If a feature's class code isn't in the specified classes list, but is a descendant of one of the specified classes (based on the EuroCrops label hierarchy), then it will match to that closest ancestor.

> P.S. Curious to know how you're finding TorchGeo so far.

I've only been working in a small portion of the codebase (e.g. haven't tried training any models yet), but it's quite nice how convenient it is to intersect different datasets, and also to use one or more GeoTIFFs / shapefiles as labels. And the code is very high quality.

On the other hand I'm used to datasets that define windows / grid cells and then label them (like the [microsoft/satellite-imagery-labeling-tool](https://github.com/microsoft/satellite-imagery-labeling-tool) workflow I think). This way the set of inputs/outputs that model can be trained with is well-defined. I'm not sure how it would handle e.g. shapefiles that label all the points of some category in a city, but the bbox of the shapefile may include regions outside the city that are not labeled; you wouldn't want the detector to be trained on those regions as negative examples.

adamjstewart:
> I'm not sure how it would handle e.g. shapefiles that label all the points of some category in a city, but the bbox of the shapefile may include regions outside the city that are not labeled; you wouldn't want the detector to be trained on those regions as negative examples.

Each sampler accepts an `roi` as input that you could use to narrow down your region of interest to only the city.

adamjstewart:
With respect to the "year" info, how is that exposed in the dataset? Is it only found in the download URL or also in the filename? If the former, I have ideas. If the latter, this is now supported.

favyen2:
> With respect to the "year" info, how is that exposed in the dataset? Is it only found in the download URL or also in the filename? If the former, I have ideas. If the latter, this is now supported.

Mm I saw the PR with VectorDataset date extraction you mentioned was merged, I added filename_regex/date_format so it can extract the year from the filename. I tried it out with some Sentinel-2 images and it created index and found the appropriate image correctly.

I need to update test to cover `_get_class_index`.

torchgeo/datasets/eurocrops.py
favyen2: Yes it is similar to `cdl.py` I think.

torchgeo/datasets/geo.py
favyen2: Let me know if this should be changed. Perhaps alternative where we have function like `VectorDataset.get_class_index(feat: fiona.Feature)` where default implementation returns `feat["properties"][self.label_name]` is better, then it can be overloaded to derive the feature differently without needing to overload the entire `__getitem__`.

adamjstewart: Yeah, let's go with that alternative. Then, most datasets won't require this, we only need to think about it for particularly complicated datasets like EuroCrops.

torchgeo/datasets/eurocrops.py
adamjstewart: Hmm, I don't love that either, but it's at least slightly more "official". Let's go with something similar to what they do. I'm afraid of cycling through a list because it means multiple classes will be mapped to the same color. The benefit of a floating point color scale is that it works even when the hcat level changes.

favyen2: OK I've switched to using continuous color map.

torchgeo/datasets/geo.py
```diff
@@ -725,13 +719,29 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
...
+        masks = array_to_tensor(masks)
+
+        sample = {"mask": masks, "crs": self.crs, "bbox": query}
 
         if self.transforms is not None:
             sample = self.transforms(sample)
 
         return sample
 
+    def _get_label(self, feature: Any) -> int:
```
adamjstewart, line 732: Let's make this function public so it shows up in the docs. Also, shouldn't this be:
```suggestion
    def _get_label(self, feature: Feature) -> int:
```

torchgeo/datasets/eurocrops.py
```diff
@@ -0,0 +1,266 @@
...
+    # Name of the column containing HCAT code in CSV file.
+    hcat_code_column = "HCAT2_code"
+
+    label_name = "EC_hcat_c"
+
+    filename_glob = "*_EC*.shp"
+
+    # Override variables to automatically extract timestamp.
+    filename_regex = r"""
+        ^(?P<prefix>[a-zA-Z]+(?:_[a-zA-Z]+)?)
```
adamjstewart, line 53: Do the prefix and suffix have any meaning? Country code or something? Is there a webpage or file that documents this syntax?

adamjstewart: I would personally be fine with:
```suggestion
        ^(?P<prefix>[A-Z_]+)
```
if that covers everything. I don't expect to find 4-digit sequences in the filename other than the year.

favyen2: > Do the prefix and suffix have any meaning? Country code or something? Is there a webpage or file that documents this syntax?

The EC21 stands for EuroCrops 2021 but some files just have EC instead.

It looks like the first part is either country or country + region. I updated the regex accordingly.

I didn't find any documentation about the filename format. But it should be "country{_region,}_YYYY_EC{21,}.shp".

torchgeo/datasets/geo.py
favyen2: fiona.Feature doesn't exist in the min req version. I tried Union[Feature, dict[str, Any]] but it fails because the fiona.Feature doesn't exist. I tried to define Feature in a variable way (import fiona.Feature, but catch ImportError and define a placeholder) but it also doesn't work because the different definition of Feature would be caught as a re-definition.

favyen2: In the old fiona version, features are represented as dicts.

adamjstewart: It looks like `fiona.Feature` was introduced in fiona 1.9, which is about a year old now. I'm okay with bumping the minimum supported version of fiona to 1.9 if that makes this easier to annotate. You would modify `pyproject.toml` and `requirements/min-reqs.old`.

Alternatively, you can do `"fiona.Feature"` to prevent evaluation during run-time.

torchgeo/datasets/utils.py
```diff
@@ -819,3 +819,29 @@ def path_is_vsi(path: str) -> bool:
...
+    to raise errors since there is no suitable torch dtype.
+
+    The returned tensor is a copy.
+
+    Args:
+        array: a :class:`numpy.ndarray`.
+
+    Returns:
+        A :class:`torch.Tensor` with the same dtype as array unless array is uint16 or
+        uint32, in which case an int32 or int64 Tensor is returned, respectively.
```
adamjstewart, line 841: Can you add:
```rst
.. versionadded:: 0.6
```
to the end? That will document which versions of TorchGeo support this.

torchgeo/datasets/geo.py
```diff
@@ -725,13 +719,29 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
...
         return sample
 
+    def get_label(self, feature: Any) -> int:
+        """Get label value to use for rendering a feature.
+
+        Args:
+            feature: the :class:`fiona.Feature` from which to extract the label.
+
+        Returns:
+            the integer label, or 0 if the feature should not be rendered.
```
adamjstewart, line 739: Can you add:
```rst
.. versionadded:: 0.6
```

torchgeo/datasets/eurocrops.py
```diff
@@ -0,0 +1,267 @@
...
+    The `EuroCrops <https://www.eurocrops.tum.de/index.html>`__ dataset combines "all
+    publicly available self-declared crop reporting datasets from countries of the
+    European Union" into a unified format. The dataset is released under CC BY 4.0 Deed.
+
+    The dataset consists of shapefiles containing a total of 22M polygons. Each polygon
+    is tagged with a "EC_hcat_n" attribute indicating the harmonized crop name grown
+    within the polygon in the year associated with the shapefile.
+
+    If you use this dataset in your research, please follow the citation guidelines at
+    https://github.com/maja601/EuroCrops#reference.
```
adamjstewart, line 37: Add:
```rst
.. versionadded:: 0.6
```

torchgeo/datasets/geo.py
adamjstewart: To fix the RtD tests, you can add `fiona.Feature` to https://github.com/microsoft/torchgeo/blob/v0.5.1/docs/conf.py#L59 since it isn't actually documented.

favyen2: Ah ok perfect, I used `"fiona.Feature"`.

favyen2:
Thanks, made these updates!

I didn't understand why the docs build is failing, I just saw one warning:
> cannot import name 'LayerType' from 'timm.layers' (/home/docs/checkouts/readthedocs.org/user_builds/torchgeo/envs/1813/lib/python3.11/site-packages/timm/layers/__init__.py)

torchgeo/datasets/eurocrops.py
```diff
@@ -0,0 +1,270 @@
...
+        assert isinstance(self.paths, str)
+
+        filepath = os.path.join(self.paths, self.hcat_fname)
+        if not check_integrity(filepath, self.hcat_md5 if self.checksum else None):
+            return False
+
+        for fname, md5 in self.zenodo_files:
+            filepath = os.path.join(self.paths, fname)
+            if not check_integrity(filepath, md5 if self.checksum else None):
+                return False
```
adamjstewart, line 150: This is the only remaining line not covered by unit tests

favyen2: Fixed!

adamjstewart:
Docs were failing due to #1824, I fixed this in #1827, updating your branch should solve the issue.

docs/api/geo_datasets.csv
```diff
@@ -9,6 +9,7 @@ Dataset,Type,Source,License,Size (px),Resolution (m)
 `EnviroAtlas`_,"Imagery, Masks","NAIP, NLCD, OpenStreetMap","CC-BY-4.0",-,1
 `Esri2020`_,Masks,Sentinel-2,"CC-BY-4.0",-,10
 `EU-DEM`_,Masks,"Aster, SRTM, Russian Topomaps","CSCDA-ESA",-,25
+`EuroCrops`_,Geometries,EU Countries,"CC-BY-SA-4.0",-,-
```
adamjstewart, line 12: According to https://www.eurocrops.tum.de/index.html it should be:
```suggestion
`EuroCrops`_,Geometries,EU Countries,"CC-BY-4.0",-,-
```
right?

adamjstewart: Ah, but the GitHub and Zenodo say CC-BY-SA-4.0. That's confusing. Let's go with what you have.

torchgeo/datasets/eurocrops.py
```diff
@@ -0,0 +1,270 @@
...
+    download_url,
+)
+
+
+class EuroCrops(VectorDataset):
+    """EuroCrops Dataset (Version 9).
+
+    The `EuroCrops <https://www.eurocrops.tum.de/index.html>`__ dataset combines "all
+    publicly available self-declared crop reporting datasets from countries of the
+    European Union" into a unified format. The dataset is released under CC BY 4.0 Deed.
```
adamjstewart, line 31: Here we say CC-BY-4.0 (because that's what the website says), but our docs say CC-BY-SA-4.0 (because that's what GitHub/Zenodo use). Not sure which to trust, but maybe we should be consistent.

adamjstewart (dismissed):
This looks awesome, great work! I'm running out of things to nitpick, so I think we can merge this soon. The license thing confuses me a bit. Let me reach out to the authors to clarify.

docs/conf.py
```diff
@@ -67,6 +67,8 @@
     ("py:class", "torchvision.models._api.WeightsEnum"),
     ("py:class", "torchvision.models.resnet.ResNet"),
     ("py:class", "torchvision.models.swin_transformer.SwinTransformer"),
+    ("py:class", "fiona.Feature"),
+    ("py:class", "fiona.model.Feature"),
```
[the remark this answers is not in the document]
favyen2: Since we don't refer to Feature in the codebase yet I changed our type annotations to use `fiona.model.Feature` which obviates the need for including `fiona.Feature` here. And it is alphabetical now.

adamjstewart (approved):

PR #1814, opened 2024-01-19
VectorDataset: extract mint/maxt from filename

oddeirikigland:
Opening PR after a discussion on Slack: https://torchgeo.slack.com/archives/C02DD9Q76SE/p1705589213060449


adamjstewart (changes requested):
Can you copy https://github.com/microsoft/torchgeo/blob/v0.5.1/torchgeo/datasets/geo.py#L323-L339 from RasterDataset to VectorDataset? You can remove these lines: https://github.com/microsoft/torchgeo/blob/v0.5.1/torchgeo/datasets/geo.py#L329-L333

We'll also need to add tests for this. I can point you to where these would be, but let me know if you need help.

oddeirikigland:

oddeirikigland:
> @oddeirikigland please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

oddeirikigland:
> Can you copy https://github.com/microsoft/torchgeo/blob/v0.5.1/torchgeo/datasets/geo.py#L323-L339 from RasterDataset to VectorDataset? You can remove these lines: https://github.com/microsoft/torchgeo/blob/v0.5.1/torchgeo/datasets/geo.py#L329-L333
> 
> We'll also need to add tests for this. I can point you to where these would be, but let me know if you need help.

Added the docstring. I'll need some guidance for the tests, where is the similar test for RasterDataset?

torchgeo/datasets/geo.py
```diff
@@ -640,6 +654,10 @@ def __init__(
             else:
                 mint = 0
                 maxt = sys.maxsize
+                match = re.match(filename_regex, os.path.basename(filepath))
```
adamjstewart, line 657: This needs a:
```python
if match:
```
right?

adamjstewart: Let's move the re.match outside of the try-except, similar to RasterDataset. Then we can skip files that don't match the regex.

adamjstewart:
For testing, you'll want to edit `tests/data/vector/data.py` such that the generated fake file includes a year in the filename. Then edit `CustomVectorDataset` in `tests/data/test_geo.py` to add a `filename_regex` and `date_format` to match.

torchgeo/datasets/geo.py
oddeirikigland: yep, fixed

oddeirikigland:
> For testing, you'll want to edit `tests/data/vector/data.py` such that the generated fake file includes a year in the filename. Then edit `CustomVectorDataset` in `tests/data/test_geo.py` to add a `filename_regex` and `date_format` to match.

added test

adamjstewart (approved):
Looks pretty straightforward to me. Hopefully we'll get some builtin datasets that use this feature in the future. Not sure what kind of data your group is working with but would love to have it in TorchGeo if it's public. #1813 might actually use this feature too.

PR #1816, opened 2024-01-22
SSL4EO-S12: use the uint8 compressed version

adamjstewart:
10x smaller file sizes, 10x faster to download, 10x faster I/O, same accuracy (according to Appendix 3A of https://arxiv.org/abs/2211.07044).

@wangyi111 is working on uploading these to HF, but for now this makes downloading much more reliable.


adamjstewart:
Alright, let's close this then. I'll let you open a PR once you upload lossless uint8 data to HF.

PR #1820, opened 2024-01-23
README: add color lexing to bibtex

adamjstewart:
Looks nicer with color: https://github.com/adamjstewart/torchgeo/tree/readme/bib?tab=readme-ov-file#citation


calebrob6:
Weird, I'm not seeing anything different:

image

calebrob6:
I think "```bibtex" is correct, but it doesn't get rendered differently in markdown (but does in my editor after specifying bibtex)

image

adamjstewart:
That's what I tried first but it didn't seem to work, then it did start to work. I think it's a browser caching thing. Let's try bibtex again.

adamjstewart:
Still seems buggy. It renders with colors for a second during a refresh and then goes back to black-and-white.

calebrob6:
I agree / observed the same

adamjstewart:
It works in this example: https://github.com/awesome-spectral-indices/awesome-spectral-indices?tab=readme-ov-file#citation

I say we merge and assume it will work.

calebrob6 (approved):

PR #1839, opened 2024-01-31
Fix deterministic group_shuffle_split

nilsleh:
Sets are unordered and therefore, repeated calls were yielding different train and val sizes for Cyclone dataset.


torchgeo/datamodules/utils.py
```diff
@@ -197,8 +197,9 @@ def group_shuffle_split(
         )
 
     generator = np.random.default_rng(seed=random_state)
+    ordered_list = sorted(list(group_vals))
```
adamjstewart, line 200: Let's move this up to line 188 and do it all in one go:
```python
group_vals = sorted(list(set(groups)))
```

adamjstewart:
Let's find a way to test this so the same bug doesn't happen again.

isaaccorley:
@nilsleh Would love to get this in the 0.5.2 release!

nilsleh:
> @nilsleh Would love to get this in the 0.5.2 release!

Thanks for the reminder :) Added a unit test to check repeated calls to the function.

nilsleh:
@adamjstewart and @isaaccorley not sure how to simulate repeated calls to the function after restarting script/kernel, so I thought separate processes might be a way to go, but actually not sure

adamjstewart:
> @adamjstewart and @isaaccorley not sure how to simulate repeated calls to the function after restarting script/kernel, so I thought separate processes might be a way to go, but actually not sure

This feels like overkill. Depending on the size of our fake dataset, can we just run the test once, print the order, then hardcode that in the test code? As long as it is always the same, it's deterministic.

tests/datamodules/test_utils.py
```diff
@@ -43,16 +46,21 @@ def test_group_shuffle_split() -> None:
...
+    test_cases = [(None, 0.2, 42), (0.8, None, 42)]
+
+    for train_size, test_size, random_state in test_cases:
+        train_indices1, test_indices1 = group_shuffle_split(
+            groups,
+            train_size=train_size,
+            test_size=test_size,
+            random_state=random_state,
+        )
+        # Check that the results are the same as the saved results
```
adamjstewart, line 61:
```suggestion
        # Check that the results are the same as expected
```
We are no longer "saving" them per se

adamjstewart (dismissed):
Much better, thanks!

tests/datamodules/test_utils.py
```diff
@@ -43,16 +46,21 @@ def test_group_shuffle_split() -> None:
...
+    test_cases = [(None, 0.2, 42), (0.8, None, 42)]
+
+    for train_size, test_size, random_state in test_cases:
+        train_indices1, test_indices1 = group_shuffle_split(
+            groups,
+            train_size=train_size,
+            test_size=test_size,
+            random_state=random_state,
+        )
+        # Check that the results are the same as expcected
```
adamjstewart, line 61:
```suggestion
        # Check that the results are the same as expected
```

adamjstewart (approved):

PR #1844, opened 2024-02-03
CDL: add 2023 checksum

yichiac:
The USDA releases CDL data every year in January of the following year.


adamjstewart (approved):
Thanks for keeping it up-to-date!

PR #1849, opened 2024-02-04
Debug coverage

adamjstewart:


kdarkhan:
@adamjstewart, you might need to override commit sha as I did [here](https://github.com/google/mdbook-i18n-helpers/blob/ac041d23a15df294fc043c257a43fb44168924bc/.github/workflows/coverage-report.yml#L103-L109) if you have coverage reports running against PRs. If not done, the checkout action works with `main` branch and not able to recognize PR specific changes.

adamjstewart:
Will try that if I get stuck. There are still a few steps missing anyway, no artifacts were found in the workflow.

adamjstewart:
Just wish there was an easier way to debug these changes without first merging them, since the workflow changes don't run unless they are in the main branch 😅

kdarkhan:
Agreed, this is very annoying. I was testing all my changes against my local fork. I disabled majority of the jobs to speed up the process. I was force-pushing to my main without a PR since the new workflow picks up stuff from main.

adamjstewart:
Got stuck, opened a discussion in https://github.com/orgs/community/discussions/106300

It's definitely [supported](https://github.com/actions/download-artifact?tab=readme-ov-file#download-artifacts-from-other-workflow-runs-or-repositories), I just don't know how to download artifacts from a different workflow.

adamjstewart:
@kdarkhan how did you figure out that `override_commit` and `override_pr` were required? I haven't yet figured out how to extract these from `github.event` but I think I'm getting close.

kdarkhan:
I read the docs from https://github.com/codecov/codecov-action/ and inspected the logs from workflow runs. It is possible that not all overrides are needed as `checkout` action sets the correct hash before the coverage upload, and I was trying many different approaches. Do note that upgrading `codecov-action` to v4 broken my implementation https://github.com/google/mdbook-i18n-helpers/issues/173. I did not have a chance to fix that yet.

PR #1869, opened 2024-02-08
Add eurocrops data module.

favyen2:
It is based on NAIPChesapeakeDataModule which splits bounding box of dataset into 1/2 train, 1/4 val, and 1/4 test.

This may not be the best way to train an actual model. I think it is more natural to either split by country, or to randomly assign each large grid cell (e.g. 4096x4096 pixel) to train/val/test and then sample within those grid cells. But I wasn't sure how to split by country since VectorDataset automatically detects all the files, or to assign large grid cell since there's no sampler that can take multiple large bounding boxes and sample patches within them.


yichiac:
Hi @favyen2, could you use the new splitting function [`random_grid_cell_assignment` ](https://torchgeo.readthedocs.io/en/latest/api/datasets.html#torchgeo.datasets.random_grid_cell_assignment) similar to CDL #1889  and NCCM #1902 for EuroCrops? It would be more realistic to split datasets into more grids during training.

tests/conf/eurocrops_sentinel2.yaml
```diff
@@ -0,0 +1,18 @@
+model:
+  class_path: SemanticSegmentationTask
+  init_args:
+    loss: "ce"
+    model: "deeplabv3+"
+    backbone: "resnet34"
```
adamjstewart, line 6: resnet18 would be smaller and the tests would run faster

tests/conf/eurocrops_sentinel2.yaml
```diff
@@ -0,0 +1,18 @@
...
+    num_filters: 1
+    ignore_index: 4
+data:
+  class_path: EuroCropsSentinel2DataModule
+  init_args:
+    batch_size: 2
+    patch_size: 16
+  dict_kwargs:
+    eurocrops_paths: "tests/data/eurocrops"
+    sentinel2_paths: "tests/data/sentinel2"
```
adamjstewart, line 18: Can you add an EOL char before the EOF char?

tests/data/eurocrops/data.py
```diff
@@ -12,15 +12,19 @@
 from rasterio.crs import CRS
 from shapely.geometry import Polygon, mapping
 
-SIZE = 100
+SIZE = 1000
```
adamjstewart, line 15: 128 should be sufficient if we go with an 8x8 grid

tests/trainers/test_segmentation.py
```diff
@@ -81,6 +82,13 @@ class TestSemanticSegmentationTask:
     def test_trainer(
         self, monkeypatch: MonkeyPatch, name: str, fast_dev_run: bool
     ) -> None:
+        if name == "eurocrops_sentinel2":
```
adamjstewart, line 85: This shouldn't be necessary

torchgeo/datamodules/eurocrops.py
```diff
@@ -0,0 +1,115 @@
...
+        Args:
+            stage: Either 'fit', 'validate', 'test', or 'predict'.
+        """
+        self.sentinel2 = Sentinel2(**self.sentinel2_kwargs)
+        self.eurocrops = EuroCrops(**self.eurocrops_kwargs)
+        self.dataset = self.sentinel2 & self.eurocrops
+
+        roi = self.dataset.bounds
+        midx = roi.minx + (roi.maxx - roi.minx) / 2
+        midy = roi.miny + (roi.maxy - roi.miny) / 2
```
adamjstewart, line 81: I think @yichiac wants to go with random grid cell assignment instead

torchgeo/datamodules/eurocrops.py
```diff
@@ -0,0 +1,115 @@
...
+from ..datasets import BoundingBox, EuroCrops, Sentinel2
+from ..samplers import GridGeoSampler, RandomBatchGeoSampler
+from ..transforms import AugmentationSequential
+from .geo import GeoDataModule
+
+
+class EuroCropsSentinel2DataModule(GeoDataModule):
+    """LightningDataModule implementation for the EuroCrops and Sentinel2 datasets.
+
+    Uses the train/val/test splits from the dataset.
```
adamjstewart, line 20: Missing:
```rst
.. versionadded:: 0.6
```

torchgeo/datamodules/eurocrops.py
```diff
@@ -0,0 +1,115 @@
...
+        """Run NCCM plot method.
+
+        Args:
+            *args: Arguments passed to plot method.
+            **kwargs: Keyword arguments passed to plot method.
+
+        Returns:
+            A matplotlib Figure with the image, ground truth, and predictions.
+
+        .. versionadded:: 0.4
```
adamjstewart, line 113: This plot method wasn't added in 0.4 because this entire class won't exist until 0.6. You can remove this line

torchgeo/datamodules/eurocrops.py
```diff
@@ -0,0 +1,115 @@
...
+        if stage in ["test"]:
+            test_roi = BoundingBox(
+                roi.minx, roi.maxx, midy, roi.maxy, roi.mint, roi.maxt
+            )
+            self.test_sampler = GridGeoSampler(
+                self.dataset, self.patch_size, self.patch_size, test_roi
+            )
+
+    def plot(self, *args: Any, **kwargs: Any) -> Figure:
+        """Run NCCM plot method.
```
adamjstewart, line 104:
```suggestion
        """Run EuroCrops plot method.
```

torchgeo/datasets/geo.py
```diff
@@ -722,7 +722,7 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
         # Use array_to_tensor since rasterize may return uint16/uint32 arrays.
         masks = array_to_tensor(masks)
 
-        sample = {"mask": masks, "crs": self.crs, "bbox": query}
+        sample = {"mask": masks.long(), "crs": self.crs, "bbox": query}
```
adamjstewart, line 725: Why is this needed?

adamjstewart (changes requested):
Let's rename everything from eurocrops_sentinel2 to sentinel2_eurocrops

tests/data/eurocrops/data.py
favyen2: It is in projection units, 100 would be too small since it would by 10x10.

favyen2:
I am not able to get this to work without some changes from #1889 like setting Sentinel-2 test data size to 128 instead of 36, so I will wait for that to be merged first since that will make it easier.

adamjstewart:
#1889 has now been merged, feel free to rebase and copy-n-paste whatever you want from that data module.

torchgeo/datasets/geo.py
favyen2: Otherwise I got error where masks byte tensor is used in cross entropy loss which expects long tensor.

RasterDataset has a dtype function that forces it to be either long tensor or float32 tensor depending on whether the raster represents class labels or image.

adamjstewart: Can we add a similar dtype function to VectorDataset? By default, it will always be long (there shouldn't be any image vector datasets).

tests/data/eurocrops/data.py
favyen2: I added a comment about it.

tests/conf/sentinel2_eurocrops.yaml
```diff
@@ -0,0 +1,18 @@
+model:
+  class_path: SemanticSegmentationTask
+  init_args:
+    loss: "ce"
+    model: "deeplabv3+"
+    backbone: "resnet18"
+    in_channels: 13
+    num_classes: 3
+    num_filters: 1
+    ignore_index: 4
```
adamjstewart, line 10: If there are only 3 classes, ignoring index 4 doesn't mean much.

torchgeo/datamodules/sentinel2_eurocrops.py
```diff
@@ -0,0 +1,111 @@
...
+        batch_size: int = 64,
+        patch_size: Union[int, tuple[int, int]] = 256,
+        length: Optional[int] = None,
+        num_workers: int = 0,
+        **kwargs: Any,
+    ) -> None:
+        """Initialize a new Sentinel2EuroCropsDataModule instance.
+
+        The dataset is split into train, val, and test by identifying all grid cells
+        of size chunk_size with nonzero labels, and randomly assigning those to splits.
```
adamjstewart, line 37: chunk_size is not defined

torchgeo/datamodules/sentinel2_eurocrops.py
```diff
@@ -0,0 +1,111 @@
...
+            EuroCrops,
+            batch_size,
+            patch_size,
+            length,
+            num_workers,
+            **self.eurocrops_kwargs,
+        )
+
+        self.aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std), data_keys=["image", "mask"]
```
adamjstewart, line 69: Can you copy all of `train_aug` and `aug` from the other `torchgeo/datamodules/sentinel2_*.py` datamodules?

adamjstewart (approved):

adamjstewart (approved):

PR #1870, opened 2024-02-08
Resolve NCCM checksum error and add years args

### LINKED ISSUE #1854: NCCM checksum error
### Description

NCCM dataset with download=True, checksum=True fails with checksum error, because the zip file seems to be created by the server on-the-fly.

### Steps to reproduce

Code:

```python
from torchgeo.datasets import NCCM
NCCM(download=True, checksum=True, paths="./data/nccm/")
```

Error:

```
Downloading https://figshare.com/ndownloader/articles/13090442/versions/1 to ./data/nccm/13090442.zip
RuntimeError: File not found or corrupted.
```

It seems the zip file changes each time so the md5 is different.

```
$ md5sum file1.zip
eb29f8e5b6f41ed2c32ba5ed2b0bdc90  file1.zip
$ md5sum file2.zip
179c53feddaa81c609b0cd0433db669a  file2.zip
$ head -c 14 file1.zip | hd
00000000  50 4b 03 04 0a 00 08 08  00 00 c9 8b 45 58        |PK..........EX|
0000000e
$ head -c 14 file2.zip | hd
00000000  50 4b 03 04 0a 00 08 08  00 00 a0 8c 45 58        |PK..........EX|
0000000e
```
According to https://stackoverflow.com/questions/67902342/how-to-read-zip-file-header-manually the last four bytes there indicate file modification time. (Other bytes are different too though.) So it is possible the server is generating the zip files on the fly.

### Version

0.6.0.dev0 (a52f88f614400513cb6ef432c7caa7a8c46d878f)

yichiac:
This PR resolves #1854 (the dynamic md5 issue) by adding download links and checksums for 3 mask files. Since we are changing to downloading 3 mask files separately, we no longer need to extract a zip file. Thus, the test files generated by `data.py` are also updated.

This PR also adds years args, enabling users to download masks for specific years.

The changes would be related to #1740, so tagging @shreyakannan1205


torchgeo/datasets/nccm.py
```diff
@@ -75,6 +87,7 @@ def __init__(
         paths: Union[str, Iterable[str]] = "data",
         crs: Optional[CRS] = None,
         res: Optional[float] = None,
+        years: list[int] = [2017, 2018, 2019],
```
adamjstewart, line 90: I would personally default to only 2019 to match CDL/NLCD behavior

torchgeo/datasets/nccm.py
```diff
@@ -132,11 +151,10 @@ def _verify(self) -> None:
         if self.files:
             return
 
-        # Check if the zip file has already been downloaded
+        # Check if the mask files have already been downloaded
```
adamjstewart, line 154: Isn't this already checked by `self.files` above?

torchgeo/datasets/nccm.py
```diff
@@ -55,12 +55,24 @@ class NCCM(RasterDataset):
...
     filename_regex = r"CDL(?P<year>\d{4})_clip"
     filename_glob = "CDL*.*"
-    zipfile_glob = "13090442.zip"
 
     date_format = "%Y"
     is_image = False
-    url = "https://figshare.com/ndownloader/articles/13090442/versions/1"
-    md5 = "eae952f1b346d7e649d027e8139a76f5"
+    urls = {
+        2017: "https://figshare.com/ndownloader/files/25070582",
```
adamjstewart, line 62: Can we sort these newest to oldest to match CDL? Maybe NLCD should be changed to. The idea is that newer versions will be added to the top, so you don't have to scroll as far to add them. But this is just the style used by Spack, we don't have to stick with it.

adamjstewart (commented):
Thanks for the quick fix!

adamjstewart (approved):

PR #1881, opened 2024-02-14
Add methods for extracting true footprint for sampling valid data only

### LINKED ISSUE #1330: How to avoid nodata-only patches
### Summary

Large scenes are often rotated due to CRS like so:

Landsat scene

When using our GeoSamplers, we often sample from the nodata regions around the edges.

### Rationale

Sampling these nodata-only patches results in slower I/O and slower GPU training, despite these patches contributing nothing to the model.

### Implementation

There are two places where we could potentially improve performance.

### I/O

Best case scenario would be to avoid sampling these regions entirely. In #449, someone tried adding a check to the sampler that actually loads the patch, checks if it's entirely nodata pixels, and skips it if so. Unfortunately, this doesn't seem to work in parallel, and is slow since it needs to load each patch twice.

If there was a way to load the image in its native CRS (i.e., a square with no nodata pixels in the orientation it was taken in), this would solve all of our problems. I don't know of a way to do this.

### GPU

This is actually easier to solve. We could add a feature to our data module base classes that removes all nodata-only images from each mini-batch inside `transfer_batch_to_device` or another step. This would result in variable batch sizes, but I don't think that's an issue.

### Alternatives

_No response_

### Additional information

This is a highly requested feature:

* #449 
* #548 
* #1170

adriantre:
Fix #1330 

RasterDatasets may contain nodata regions due to projecting all file to the same CRS, and due to eventual inherit nodata regions in the images.
When IntersectionDataset joins this with VectorDataset, this may yield 
1. false positive samples (bad for learning)
2. empty negative samples (may be bad for learning)

The solution can be summarised as:
- [ ] In RasterDataset, when opening each file, extract footprint and add to rtree index object
- [ ] In IntersectionDataset._merge_dataset_indices copy over the footprint to the new index. 
- [ ] In the same method, could optimise by minimizing bbox to cover only actual intersection of valid data.
- [ ] In RandomGeoSampler.__iter__, use this footprint to validate that sample bbox actually overlaps, and don't yield until a valid box is found. 
- [ ] Enable the same for GridGeoSampler **(probably other PR)**
- [ ] Remove label mask for eventual nodata-regions that  outside regions in boundary. (As the criteria above is `overlaps` and not `contains`, corners of the resulting sample may still contain nodata, while the label mask still may cover this.) **(probably other PR)**
- [ ] Add ability to balance positive and negative samples. The VectorData can be intersected with the raster valid data footprint in the GeoSampler to facilitate ballancing positives and negatives. Right now torchgeo gives the user no control of this. **(probably other PR)**

Useful resources:
Rasterio nodata masks:
https://rasterio.readthedocs.io/en/latest/topics/masks.html#nodata-masks

Extract valid data footprint as vector
https://gist.github.com/sgillies/9713809

Reproject valid data footprint with rasterio
https://geopandas.org/en/stable/docs/user_guide/reproject_fiona.html#rasterio-example

### DIFF
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -38,6 +38,7 @@ from .utils import (
     disambiguate_timestamp,
     merge_samples,
     path_is_vsi,
+    valid_data_footprint_from_datasource,
 )
 
 
@@ -426,6 +427,8 @@ class RasterDataset(GeoDataset):
                         if res is None:
                             res = src.res[0]
 
+                        valid_footprint = valid_data_footprint_from_datasource(src, crs)
+
                         with WarpedVRT(src, crs=crs) as vrt:
                             minx, miny, maxx, maxy = vrt.bounds
                 except rasterio.errors.RasterioIOError:
@@ -444,7 +447,12 @@ class RasterDataset(GeoDataset):
                         _, maxt = disambiguate_timestamp(stop, self.date_format)
 
                     coords = (minx, maxx, miny, maxy, mint, maxt)
-                    self.index.insert(i, coords, filepath)
+
+                    self.index.insert(
+                        i,
+                        coords,
+                        {"filepath": filepath, "valid_footprint": valid_footprint},
+                    )
                     i += 1
 
         if i == 0:
@@ -480,7 +488,7 @@ class RasterDataset(GeoDataset):
             IndexError: if query is not found in the index
         """
         hits = self.index.intersection(tuple(query), objects=True)
-        filepaths = cast(list[str], [hit.object for hit in hits])
+        filepaths = cast(list[str], [hit.object["filepath"] for hit in hits])
 
         if not filepaths:
             raise IndexError(
@@ -933,7 +941,7 @@ class IntersectionDataset(GeoDataset):
             for hit2 in ds2.index.intersection(hit1.bounds, objects=True):
                 box1 = BoundingBox(*hit1.bounds)
                 box2 = BoundingBox(*hit2.bounds)
-                self.index.insert(i, tuple(box1 & box2))
+                self.index.insert(i, tuple(box1 & box2), hit1.object["valid_footprint"])
                 i += 1
 
         if i == 0:
--- a/torchgeo/datasets/utils.py
+++ b/torchgeo/datasets/utils.py
@@ -20,8 +20,12 @@ from datetime import datetime, timedelta
 from typing import Any, cast, overload
 
 import numpy as np
+import pyproj
 import rasterio
 import torch
+from rasterio.features import shapes, sieve
+from rasterio.warp import transform_geom
+from shapely.geometry import Polygon
 from torch import Tensor
 from torch.utils.data import Dataset
 from torchvision.datasets.utils import check_integrity, download_url
@@ -847,3 +851,20 @@ def array_to_tensor(array: np.typing.NDArray[Any]) -> Tensor:
     elif array.dtype == np.uint32:
         array = array.astype(np.int64)
     return torch.tensor(array)
+
+
+def valid_data_footprint_from_datasource(
+    src: rasterio.io.DatasetReader, destination_crs: pyproj.crs.crs.CRS
+) -> Polygon:
+    # Read valid/nodata-mask
+    mask = src.read_masks()
+    # Close holes
+    sieved_mask = sieve(mask, 500)
+    # extract polygon for valid data values
+    geom = next(g for g, v in shapes(sieved_mask, transform=src.transform) if v > 0)
+
+    # Transform to the common CRS chosen in RasterDataset
+    polygon_dict = transform_geom(src_crs=src.crs, dst_crs=destination_crs, geom=geom)
+    # Create a Shapely Polygon object
+    polygon = Polygon(polygon_dict["coordinates"][0])
+    return polygon
--- a/torchgeo/samplers/single.py
+++ b/torchgeo/samplers/single.py
@@ -13,7 +13,7 @@ from torch.utils.data import Sampler
 
 from ..datasets import BoundingBox, GeoDataset
 from .constants import Units
-from .utils import _to_tuple, get_random_bounding_box, tile_to_chips
+from .utils import _to_tuple, get_random_bounding_box_check_valid_overlap, tile_to_chips
 
 
 class GeoSampler(Sampler[BoundingBox], abc.ABC):
@@ -142,10 +142,9 @@ class RandomGeoSampler(GeoSampler):
             hit = self.hits[idx]
             bounds = BoundingBox(*hit.bounds)
 
-            # Choose a random index within that tile
-            bounding_box = get_random_bounding_box(bounds, self.size, self.res)
-
-            yield bounding_box
+            yield get_random_bounding_box_check_valid_overlap(
+                bounds, self.size, self.res, hit.object
+            )
 
     def __len__(self) -> int:
         """Return the number of samples in a single epoch.
--- a/torchgeo/samplers/utils.py
+++ b/torchgeo/samplers/utils.py
@@ -6,6 +6,7 @@
 import math
 from typing import Optional, Union, overload
 
+import shapely
 import torch
 
 from ..datasets import BoundingBox
@@ -75,6 +76,43 @@ def get_random_bounding_box(
 
     query = BoundingBox(minx, maxx, miny, maxy, mint, maxt)
     return query
+
+
+def get_random_bounding_box_check_valid_overlap(
+    bounds: BoundingBox,
+    size: Union[tuple[float, float], float],
+    res: float,
+    valid_footprint: shapely.geometry.Polygon,
+) -> BoundingBox:
+    """Returns a random bounding box within a given bounding box.
+
+    Extends `get_random_bounding_box`by guaranteeing that
+    the bounding box overlaps spatially with the valid_footprint of the raster.
+
+    The ``size`` argument can either be:
+
+        * a single ``float`` - in which case the same value is used for the height and
+          width dimension
+        * a ``tuple`` of two floats - in which case, the first *float* is used for the
+          height dimension, and the second *float* for the width dimension
+
+    Args:
+        bounds: the larger bounding box to sample from
+        size: the size of the bounding box to sample
+        valid_footprint: a Polygon in the common CRS of the originating RasterDataset
+
+    Returns:
+        randomly sampled bounding box from the extent of the input
+    """
+    # We should be able to trust that torchgeo/rtree
+    # can guarantee that there are valid pixels within the bounds
+    while True:
+        bounding_box = get_random_bounding_box(bounds, size, res)
+        bbox = shapely.geometry.box(
+            bounding_box.minx, bounding_box.miny, bounding_box.maxx, bounding_box.maxy
+        )
+        if shapely.overlaps(bbox, valid_footprint):
+            return bounding_box
 
 
 def tile_to_chips(


torchgeo/datasets/utils.py
```diff
@@ -847,3 +851,20 @@ def array_to_tensor(array: np.typing.NDArray[Any]) -> Tensor:
     elif array.dtype == np.uint32:
         array = array.astype(np.int64)
     return torch.tensor(array)
+
+
+def valid_data_footprint_from_datasource(
+    src: rasterio.io.DatasetReader, destination_crs: pyproj.crs.crs.CRS
+) -> Polygon:
+    # Read valid/nodata-mask
+    mask = src.read_masks()
```
johnnv1, line 860: `read_masks()` returns `C, H, W` shape it has more than one band/mask

probably needs to do something like

```python
if len(masks) == 3:
    mask = np.logical_or(*masks[:, ...])
```

torchgeo/datasets/utils.py
```diff
@@ -847,3 +851,20 @@ def array_to_tensor(array: np.typing.NDArray[Any]) -> Tensor:
...
+    # Close holes
+    sieved_mask = sieve(mask, 500)
```
johnnv1, line 862: we didn't have to know the minimum `size` accepted for any raster... Maybe it can have a `factor` to compute this based on the mask shape?

torchgeo/datasets/utils.py
```diff
@@ -847,3 +851,20 @@ def array_to_tensor(array: np.typing.NDArray[Any]) -> Tensor:
...
+    # extract polygon for valid data values
+    geom = next(g for g, v in shapes(sieved_mask, transform=src.transform) if v > 0)
```
johnnv1, line 864: instead of the first polygon, maybe we should go for the bigger polygon? Or if we aim to handle everything, maybe it should be a join between the geometries

torchgeo/samplers/single.py
```diff
@@ -142,10 +142,9 @@ def __iter__(self) -> Iterator[BoundingBox]:
             hit = self.hits[idx]
             bounds = BoundingBox(*hit.bounds)
 
-            # Choose a random index within that tile
-            bounding_box = get_random_bounding_box(bounds, self.size, self.res)
-
-            yield bounding_box
+            yield get_random_bounding_box_check_valid_overlap(
```
johnnv1, line 145: maybe first add it as an option in the API, and pass the max_retries too

torchgeo/datasets/utils.py
adriantre: We should choose to only use the mask for one spectral band, and it should probably work well, like `mask=src.read_masks(1)`. But in case there is a spatial shift between the bands, passing all of them to `sieve` will (to my understanding) find the valid mask across **all** bands.

adriantre: So I think your proposal is covered by `sieve`?

adamjstewart: 1 spectral band may not be enough, especially in the case of Landsat 7 SLC-off. Would be curious to know how far things shift. If it's only a few pixels, then that's within the margin of error of this method.

torchgeo/datasets/utils.py
adriantre: The target here is a polygon with no holes. Probably 500 is never too big (22 x 22 pixels). Could increase it, too.

If there are more (bigger) holes left, we could close them using shapely after converting to vector. What do you think?

torchgeo/datasets/utils.py
adriantre: Looking at this again I see that we indeed need the logical_or to cover shifts between spectral bands. Will push soon!

(My last comment about `sieve` (or `shape`) solving this was not correct)

torchgeo/datasets/utils.py
adriantre: I am now pushing a suggestion where we keep all polygons (that are not nodata) and make a multipolygon. This will then support "complex" footprints, or disjunct mosaics

torchgeo/samplers/single.py
adriantre: Pushing a suggestion now

torchgeo/datasets/geo.py
```diff
@@ -569,11 +579,21 @@ def _load_warp_file(self, filepath: str) -> DatasetReader:
...
         if src.crs != self.crs:
-            vrt = WarpedVRT(src, crs=self.crs)
+            valid_nodatavals = [val for val in src.nodatavals if val is not None]
+            if not valid_nodatavals:  # Case for Sentinel2 L1C
+                nodata = 0.0  # Is it safe to assume? For Sentinel2 L1C it is 0.0
+            else:
+                # Get the first valid nodata value.
+                # Usualy the same value for all bands
+                nodata = valid_nodatavals[0]
+            vrt = WarpedVRT(src, nodata=nodata, crs=self.crs)
```
adriantre, line 596: Sentinel-2 has as far I can see no value set for nodata. I looked everywhere. Even enabling [alpha-layer](https://gdal.org/drivers/raster/sentinel2.html#open-options) and opening the [MSK_QUALIT-file] in the Sentinel-2 gdal driver, and looking through the (https://sentinels.copernicus.eu/web/sentinel/technical-guides/sentinel-2-msi/level-1c/masks) I found nothing.

adriantre: This change will set the nodata-value. Some datasets have other nodata-values, and we should probably let the user overwrite this, for example in their RasterDataset.

adriantre: Currently, the nodata is only overridden for the warped datasources. The non-warped (already correct CRS) are opened as is, but would also need to have the nodata overridden.

torchgeo/samplers/single.py
```diff
@@ -142,10 +156,16 @@ def __iter__(self) -> Iterator[BoundingBox]:
             hit = self.hits[idx]
             bounds = BoundingBox(*hit.bounds)
```
adriantre, line 157: Only the first hit (file) is chosen. Currently I use this file to extract the footprint. But the sample is read from the merged raster, and the footprint for this one file might not cover the other.

Should be fairly easy to combine the footprints for all hits and send that to `get_random_bounding_box_check_valid_overlap`. 

Let me know if my assumptions are correct, and I will implement it.

torchgeo/datasets/utils.py
johnnv1: > I am now pushing a suggestion where we keep all polygons (that are not nodata) and make a multipolygon. This will then support "complex" footprints, or disjunct mosaics

Nice, I would give it as a suggestion because I see some cases like this on a daily basis 😄

torchgeo/datasets/utils.py
adriantre: Hmm, you are probably on to something. I'm struggling to decide what effect it might have if we set size too big or too small

torchgeo/samplers/single.py
adriantre: The footprint for a hit is static. So this could be joined and cropped during init of IntersectionDataset

adriantre: I assume I have understood this correctly, and added this functionality in `_merge_dataset_indices`.

torchgeo/datasets/utils.py
johnnv1: I thought of something like:  closing holes that are bigger than the window size, we can still be getting some cases retrieving samples with just nodata ... considering the multi-polygon thing here.

One example is if we beforehand masks Sentinel-2 clouds as no data when closing the holes considering a size bigger than the window size we still can get random samples inside/within this nodata regions

adriantre: Sounds smart!

adriantre: One thing is that the desired patch_size to be used by sampler is not available at this point in the code. This happens on RasterDataset init, separate from the Sampler init.

adamjstewart:
I'm trying to get back up to speed on this. Can you rebase? Also, would be good to test with the new I/O Bench dataset I added. See https://torchgeo.readthedocs.io/en/latest/user/contributing.html#i-o-benchmarking for instructions.

adriantre:
> I'm trying to get back up to speed on this. Can you rebase? Also, would be good to test with the new I/O Bench dataset I added. See https://torchgeo.readthedocs.io/en/latest/user/contributing.html#i-o-benchmarking for instructions.

I rebased now, have a look. I'll look into your new dataset!

adriantre:
I did a quick test.

My initial implementation is slower on your benchmark dataset `io_raw`. 
I assume it is for two reasons, as `setup` and `dataloader_next` is the bad apples:
1. Computing the true_footprints takes about 2 seconds. (This can be precomputed and stored/cached)
2. My `RandomGeoSampler` is naive, as it just tries until it finds a random box that overlaps with true_footprint, which may result in thousands of re-tries. 

Having `patch_size=128` and `num_workers = 0`:
6 seconds slower in total.
Reduces the number of no-data only samples from 1206 to 0.

What threw me off a little was that it seems like a lot of the nodata pixels for some reason get value`1.`, is this correct?
This can e.g. be observed in `RasterDataset.__getitem__` under `if self.is_image`.

GPU available: True (mps), used: True
TPU available: False, using: 0 TPU cores
IPU available: False, using: 0 IPUs
HPU available: False, using: 0 HPUs

torchgeo/datasets/utils.py
adriantre: `np.logical_or.reduce(mask, axis=0).astype('uint8')` now accept any number of bands.

adamjstewart:
So if I'm understanding correctly, the current state of this PR is that the naïve implementation of this is actually slower than loading all nodata pixels? In that case, we may want to push this to the 0.7.0 milestone so we have more time to ensure that everything is bug-free.

May be worth looking into how Raster Vision does this, they've been around longer than us: https://github.com/microsoft/torchgeo/issues/1330#issuecomment-2284444522

P.S. Need to resolve merge conflicts.

adriantre:
If the only goal is low speed per epoch, then yes. But if the goal is low speed for the model to converge, this might still be faster.

In my opinion, this PR is the first step towards #1883 which is really important for faster overall training.

adamjstewart:
I guess I would need to see additional tests to verify if it is actually faster for model convergence. Unfortunately that's even harder to time since it's inherently stochastic. If we can't also make it faster to load data, it seems unlikely to be merged.

adriantre:
> There's no mechanism for on-the-fly filtering based on nodata content in RV GeoDatasets. However, RV also provides a way to pre-chip a dataset and then treat it as a non-geospatial dataset. And that does allow filtering based on nodata percentage via a [nodata_threshold option](https://docs.rastervision.io/en/stable/api_reference/_generated/rastervision.core.rv_pipeline.chip_options.ChipOptions.html#rastervision.core.rv_pipeline.chip_options.ChipOptions.nodata_threshold).

*from https://github.com/microsoft/torchgeo/issues/1330#issuecomment-2284444522*

This "pre-chipping" is my idea for how this could be solved, but my idea still keeps it as a GeoDataset. 

> Alternatives
One good alternative/evolution that I see is replacing the rtree-index created in IntersectionDataset. Instead of the current rtree-index, add the vectorized raster-footprints and/or vector data (actual features/shapes from e.g. shapefiles) to GeoPandas GeoDataFrames, and use their implementation of rtree to rapidly find areas where they intersect. The temporal dimension would not be part of the index (not supported), but can be added as a filter once spatial matches are found. At the end of [this article](https://geoffboeing.com/2016/10/r-tree-spatial-index-python/) they split their polygons [raster-footprints] into small grid cells [of patch size] and can rapidly retrieve cells [samples] where there is overlap with the other data set [labels], or no overlap. The underlying rtree-indices in RasterDatasets and VectorDatasets will be the same, and still be used to read the data upon sampling.

*from first post in #1883*

I would need an indication that this is a good/promising solution before starting to dive into it.

adamjstewart:
Both of these would be pretty big core changes and would need a lot of justification. Losing explicit temporal sampling would be unfortunate, as we're trying to improve time series support in the next 6 months. But the ability to check for overlap of arbitrary polygons (not just boxes) would be very nice. I am worried about the decrease in performance though, especially as you approach millions of images. A formal mathematical R-tree will always be significantly faster than a hacky R-tree that supports arbitrary shapes. Basically, I don't know the best way to do this, and I don't want you to sink tens of hours into something that may never be accepted, but I also wish someone would solve this problem and don't have the time/need myself.

adamjstewart:
Note that we are considering porting from RTree to Geopandas or STAC, which support complex shapely Geometries and would allow us to make use of this true footprint. May be worth revisiting this after the port is finished.

adamjstewart:
#2747 is now merged if you want to revive this PR. We can now store arbitrary Polygons directly in our GeoDataFrame. Sampling will come later.

adamjstewart:
Superseded by #2903

PR #1884, opened 2024-02-15
Additional Satlas pretrained models

piperwolters:
Documentation, satellite-specific transforms, and weights for additional Satlas single-image RGB & multispectral Swin-v2 models. Tests pass.

Few things:
- Renamed the existing model weights to include "SI", for single-image input. torchgeo doesn't support multi-image input yet, but there are multi-image Satlas models so hopefully someday those could be included.
- There are a couple custom transforms, for Landsat and multispectral Sentinel-2. Maybe there's a cleaner way to do this?

### DIFF
--- a/docs/api/naip_pretrained_weights.csv
+++ b/docs/api/naip_pretrained_weights.csv
@@ -1,2 +1,2 @@
 Weight,Channels,Source,Citation,License
-Swin_V2_B_Weights.NAIP_RGB_SI_SATLAS,3,'link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"ODC-BY"
+Swin_V2_B_Weights.NAIP_RGB_SI_SATLAS,3,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"ODC-BY"
--- a/docs/api/naip_pretrained_weights.csv
+++ b/docs/api/naip_pretrained_weights.csv
@@ -1,2 +1,2 @@
 Weight,Channels,Source,Citation,License
-Swin_V2_B_Weights.NAIP_RGB_SATLAS,3,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"Apache-2.0"
+Swin_V2_B_Weights.NAIP_RGB_SI_SATLAS,3,'link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"ODC-BY"
--- a/docs/api/sentinel1_pretrained_weights.csv
+++ b/docs/api/sentinel1_pretrained_weights.csv
@@ -1,2 +1,3 @@
 Weight,Channels,Source,Citation,License
 ResNet50_Weights.SENTINEL1_ALL_MOCO, 2,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,"CC-BY-4.0"
+Swin_V2_B_Weights.SENTINEL1_SI_SATLAS,2,'link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"ODC-BY"
--- a/docs/api/sentinel1_pretrained_weights.csv
+++ b/docs/api/sentinel1_pretrained_weights.csv
@@ -1,3 +1,3 @@
 Weight,Channels,Source,Citation,License
 ResNet50_Weights.SENTINEL1_ALL_MOCO, 2,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,"CC-BY-4.0"
-Swin_V2_B_Weights.SENTINEL1_SI_SATLAS,2,'link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"ODC-BY"
+Swin_V2_B_Weights.SENTINEL1_SI_SATLAS,2,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"ODC-BY"
--- a/docs/api/sentinel2_pretrained_weights.csv
+++ b/docs/api/sentinel2_pretrained_weights.csv
@@ -8,4 +8,5 @@ ResNet50_Weights.SENTINEL2_RGB_MOCO, 3,`link <https://github.com/zhu-xlab/SSL4EO
 ResNet50_Weights.SENTINEL2_RGB_SECO, 3,`link <https://github.com/ServiceNow/seasonal-contrast>`__,`link <https://arxiv.org/abs/2103.16607>`__,"Apache-2.0",87.81,,,
 ViTSmall16_Weights.SENTINEL2_ALL_DINO,13,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,"CC-BY-4.0",90.5,99.0,62.2,
 ViTSmall16_Weights.SENTINEL2_ALL_MOCO,13,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,"CC-BY-4.0",89.9,98.6,61.6,
-Swin_V2_B_Weights.SENTINEL2_RGB_SATLAS,3,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"Apache-2.0",,,,
+Swin_V2_B_Weights.SENTINEL2_RGB_SI_SATLAS,3,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"ODC-BY",,,,
+Swin_V2_B_Weights.SENTINEL2_MS_SI_SATLAS,9,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"ODC-BY",,,,
--- a/torchgeo/models/swin.py
+++ b/torchgeo/models/swin.py
@@ -101,6 +101,7 @@ class Swin_V2_B_Weights(WeightsEnum):  # type: ignore[misc]
             "model": "swin_v2_b",
             "publication": "https://arxiv.org/abs/2211.15660",
             "repo": "https://github.com/allenai/satlas",
+            "bands": ["VH", "VV"],
         },
     )
 
@@ -109,10 +110,23 @@ class Swin_V2_B_Weights(WeightsEnum):  # type: ignore[misc]
         transforms=_landsat_satlas_transforms,
         meta={
             "dataset": "Satlas",
-            "in_chans": 2,
+            "in_chans": 11,
             "model": "swin_v2_b",
             "publication": "https://arxiv.org/abs/2211.15660",
             "repo": "https://github.com/allenai/satlas",
+            "bands": [
+                "B01",
+                "B02",
+                "B03",
+                "B04",
+                "B05",
+                "B06",
+                "B07",
+                "B08",
+                "B09",
+                "B10",
+                "B11",
+            ],  # noqa: E501
         },
     )
 
--- a/torchgeo/models/swin.py
+++ b/torchgeo/models/swin.py
@@ -8,6 +8,7 @@ from typing import Any, Optional
 import kornia.augmentation as K
 import torch
 import torchvision
+from kornia.contrib import Lambda
 from torchvision.models import SwinTransformer
 from torchvision.models._api import Weights, WeightsEnum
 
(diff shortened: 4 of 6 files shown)


docs/api/naip_pretrained_weights.csv
```diff
@@ -1,2 +1,2 @@
 Weight,Channels,Source,Citation,License
-Swin_V2_B_Weights.NAIP_RGB_SATLAS,3,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"Apache-2.0"
\ No newline at end of file
+Swin_V2_B_Weights.NAIP_RGB_SI_SATLAS,3,'link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"ODC-BY"
```
isaaccorley, line 2: I think the ' character needs to be this character -> `

docs/api/sentinel1_pretrained_weights.csv
```diff
@@ -1,2 +1,3 @@
 Weight,Channels,Source,Citation,License
 ResNet50_Weights.SENTINEL1_ALL_MOCO, 2,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,"CC-BY-4.0"
+Swin_V2_B_Weights.SENTINEL1_SI_SATLAS,2,'link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"ODC-BY"
```
isaaccorley, line 3: Same here

torchgeo/models/swin.py
```diff
@@ -16,14 +16,48 @@
...
+    def forward(self, x):
+        # Divide the first 3 channels by 255
+        x[:, :3, :, :] = x[:, :3, :, :] / 255.0
+        # Divide the last 6 channels by 8160 and clip to 0-1
+        x[:, -6:, :, :] = torch.clamp(x[:, -6:, :, :] / 8160.0, 0.0, 1.0)
+        return x
+
+
+# Custom transform for Landsat multispectral model inputs.
+class LandsatTransform(torch.nn.Module):
```
isaaccorley, line 33: We try to decompose custom transforms into kornia transforms. See `resnet.py` for some examples.

This can be

```python
K.Normalize(mean=torch.tensor(4000), std=torch.tensor(1))
K.Normalize(mean=torch.tensor(0), std=torch.tensor(16320))
```

torchgeo/models/swin.py
```diff
@@ -16,14 +16,48 @@
 __all__ = ["Swin_V2_B_Weights"]
 
 
+# Custom transform for Sentinel-2 multispectral model inputs; Uses 9 bands.
+class MSS2Transform(torch.nn.Module):
```
isaaccorley, line 20: Converting this one to kornia might be more complicated. Let me draft something up

isaaccorley (changes requested):

torchgeo/models/swin.py
```diff
@@ -16,14 +16,40 @@
...
+        # Divide the last 6 channels by 8160 and clip to 0-1
+        x[:, -6:, :, :] = torch.clamp(x[:, -6:, :, :] / 8160.0, 0.0, 1.0)
+        return x
+
+
 # https://github.com/allenai/satlas/blob/bcaa968da5395f675d067613e02613a344e81415/satlas/cmd/model/train.py#L42 # noqa: E501
-# All Satlas imagery is uint8 and normalized to the range (0, 1) by dividing by 255
+# Satlas Sentinel-1 and RGB Sentinel-2 and NAIP imagery is uint8 and is normalized to (0, 1) by dividing by 255. # noqa: E501
 _satlas_transforms = AugmentationSequential(
     K.CenterCrop(256),
```
isaaccorley, line 35: No need to center crop unless you think the input shape should be fixed to 256x256

torchgeo/models/swin.py
```diff
@@ -16,14 +16,40 @@
...
+# Satlas multispectral Sentinel-2 imagery divides first 3 bands by 255 and the following 6 bands by 8160, both clipped to (0, 1). # noqa: E501
+_sentinel2_ms_satlas_transforms = AugmentationSequential(
+    K.CenterCrop(256), MSS2Transform(), data_keys=["image"]
+)
+
+# Satlas Landsat imagery is 16-bit, normalized by clipping some pixel N with (N-4000)/16320 to (0, 1). # noqa: E501
+_landsat_satlas_transforms = AugmentationSequential(
+    K.CenterCrop(256),
+    K.Normalize(mean=torch.tensor(4000), std=torch.tensor(1)),
+    K.Normalize(mean=torch.tensor(0), std=torch.tensor(16320)),
```
isaaccorley, line 49: You can implement the previous clipping by adding the following to the end of the AugmentationSequential for example:

```python
from kornia.contrib import Lambda

Lambda(lambda x: torch.clamp(x, min=0.0, max=1.0))
```

torchgeo/models/swin.py
isaaccorley: This should do the trick:

```python
_std = torch.tensor([255.0, 255.0, 255.0, 8160.0, 8160.0, 8160.0, 8160.0, 8160.0, 8160.0)] # not sure if there are additional bands between the first 3 and last 6, but can add 1.0 to not modify them
_mean = torch.zeros_like(_std)
_sentinel2_ms_satlas_transforms = AugmentationSequential(
   K.Normalize(mean=_mean, std=_std),
   Lambda(lambda x: torch.clamp(x, min=0.0, max=1.0))
   data_keys=["image"]
)

**Review by isaaccorley — CHANGES_REQUESTED** [review]

**Comment by piperwolters** [comment]
@microsoft-github-policy-service agree [company="Allen Institute for AI"]

**Comment by piperwolters** [comment]
@microsoft-github-policy-service agree [company="Allen Institute for AI"]

**Comment by piperwolters** [comment]
@microsoft-github-policy-service agree company=Allen Institute for AI

**Comment by piperwolters** [comment]
@microsoft-github-policy-service agree company=(Allen Institute for AI)

File: torchgeo/models/swin.py [hunk 7]
```diff
@@ -63,6 +79,43 @@ class Swin_V2_B_Weights(WeightsEnum):  # type: ignore[misc]
...
+            "repo": "https://github.com/allenai/satlas",
+        },
+    )
+
+    LANDSAT_SI_SATLAS = Weights(
+        url="https://huggingface.co/allenai/satlas-pretrain/resolve/main/landsat_swinb_si.pth",  # noqa: E501
+        transforms=_landsat_satlas_transforms,
+        meta={
+            "dataset": "Satlas",
+            "in_chans": 2,
```
isaaccorley, line 112: Could you define the bands here as well just for clarity?

torchgeo/models/swin.py
```diff
@@ -63,6 +79,43 @@ class Swin_V2_B_Weights(WeightsEnum):  # type: ignore[misc]
...
+            "bands": ["B02", "B03", "B04", "B05", "B06", "B07", "B08", "B11", "B12"],
+        },
+    )
+
+    SENTINEL1_SI_SATLAS = Weights(
+        url="https://huggingface.co/allenai/satlas-pretrain/resolve/main/sentinel1_swinb_si.pth",  # noqa: E501
+        transforms=_satlas_transforms,
+        meta={
+            "dataset": "Satlas",
+            "in_chans": 2,
```
isaaccorley, line 100: Could you define the bands here as well just for clarity?

isaaccorley (changes requested):
Minor change but otherwise LGTM. @adamjstewart not sure if the CLA bot is broken or what's happening here

piperwolters:
@microsoft-github-policy-service agree company="Allen Institute for AI"

calebrob6:
Looks like that last one did it 🤷‍♂️

torchgeo/models/swin.py
```diff
@@ -8,19 +8,35 @@
...
 # https://github.com/allenai/satlas/blob/bcaa968da5395f675d067613e02613a344e81415/satlas/cmd/model/train.py#L42 # noqa: E501
-# All Satlas imagery is uint8 and normalized to the range (0, 1) by dividing by 255
+# Satlas Sentinel-1 and RGB Sentinel-2 and NAIP imagery is uint8 and is normalized to (0, 1) by dividing by 255. # noqa: E501
 _satlas_transforms = AugmentationSequential(
-    K.CenterCrop(256),
-    K.Normalize(mean=torch.tensor(0), std=torch.tensor(255)),
+    K.Normalize(mean=torch.tensor(0), std=torch.tensor(255)), data_keys=["image"]
+)
+
+# Satlas multispectral Sentinel-2 imagery divides first 3 bands by 255 and the following 6 bands by 8160, both clipped to (0, 1). # noqa: E501
```
calebrob6, line 25: Why do we divide B02, B03 and B04 by 255 here and the rest by 8160? I think RGB S2 values are much larger than 255.

calebrob6: Ahh I found the details here https://github.com/allenai/satlas/blob/main/Normalization.md#sentinel-2-images. Satlas uses the TCI product for RGB which is already [0,255]. We should definitely link to this / explain it as this will be a bottleneck for getting useful embeddings out of the model!

calebrob6 (dismissed):

torchgeo/models/swin.py
calebrob6: Nice, thanks!

calebrob6 (approved):

isaaccorley (approved):

calebrob6:
Thanks Piper!!

piperwolters:
Woohoo! Thank you guys

docs/api/naip_pretrained_weights.csv
```diff
@@ -1,2 +1,2 @@
 Weight,Channels,Source,Citation,License
-Swin_V2_B_Weights.NAIP_RGB_SATLAS,3,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"Apache-2.0"
\ No newline at end of file
+Swin_V2_B_Weights.NAIP_RGB_SI_SATLAS,3,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"ODC-BY"
```
adamjstewart, line 2: Thanks for correcting this license

docs/api/sentinel2_pretrained_weights.csv
```diff
@@ -8,4 +8,5 @@ ResNet50_Weights.SENTINEL2_RGB_MOCO, 3,`link <https://github.com/zhu-xlab/SSL4EO
 ResNet50_Weights.SENTINEL2_RGB_SECO, 3,`link <https://github.com/ServiceNow/seasonal-contrast>`__,`link <https://arxiv.org/abs/2103.16607>`__,"Apache-2.0",87.81,,,
 ViTSmall16_Weights.SENTINEL2_ALL_DINO,13,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,"CC-BY-4.0",90.5,99.0,62.2,
 ViTSmall16_Weights.SENTINEL2_ALL_MOCO,13,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,"CC-BY-4.0",89.9,98.6,61.6,
-Swin_V2_B_Weights.SENTINEL2_RGB_SATLAS,3,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"Apache-2.0",,,,
\ No newline at end of file
+Swin_V2_B_Weights.SENTINEL2_RGB_SI_SATLAS,3,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"ODC-BY",,,,
+Swin_V2_B_Weights.SENTINEL2_MS_SI_SATLAS,9,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"ODC-BY",,,,
```
adamjstewart, line 12: We usually use `ALL` instead of `MS` if all bands are used. Not sure what the best naming is for consistency/clarity.

torchgeo/models/swin.py
```diff
@@ -8,19 +8,39 @@
...
 from ..transforms import AugmentationSequential
 
 __all__ = ["Swin_V2_B_Weights"]
 
-
 # https://github.com/allenai/satlas/blob/bcaa968da5395f675d067613e02613a344e81415/satlas/cmd/model/train.py#L42 # noqa: E501
-# All Satlas imagery is uint8 and normalized to the range (0, 1) by dividing by 255
+# Satlas uses the TCI product for Sentinel-2 RGB, which is in the range (0, 255).
+# See details:  https://github.com/allenai/satlas/blob/main/Normalization.md#sentinel-2-images.  # noqa: E501
+# Satlas Sentinel-1 and RGB Sentinel-2 and NAIP imagery is uint8 and is normalized to (0, 1) by dividing by 255. # noqa: E501
```
adamjstewart, line 22: Probably shouldn't use `# noqa` unless it's absolutely necessary (long URL).

torchgeo/models/swin.py
```diff
@@ -8,19 +8,39 @@
...
-    K.Normalize(mean=torch.tensor(0), std=torch.tensor(255)),
+    K.Normalize(mean=torch.tensor(0), std=torch.tensor(255)), data_keys=["image"]
+)
+
+# Satlas uses the TCI product for Sentinel-2 RGB, which is in the range (0, 255).
+# See details:  https://github.com/allenai/satlas/blob/main/Normalization.md#sentinel-2-images.  # noqa: E501
+# Satlas Sentinel-2 multispectral imagery has first 3 bands divided by 255 and the following 6 bands by 8160, both clipped to (0, 1). # noqa: E501
+_std = torch.tensor(
+    [255.0, 255.0, 255.0, 8160.0, 8160.0, 8160.0, 8160.0, 8160.0, 8160.0]
+)  # noqa: E501
```
adamjstewart, line 32: No need for `# noqa` here

torchgeo/models/swin.py
```diff
@@ -39,8 +59,8 @@ class Swin_V2_B_Weights(WeightsEnum):  # type: ignore[misc]
     .. versionadded:: 0.6
     """
 
-    NAIP_RGB_SATLAS = Weights(
-        url="https://huggingface.co/torchgeo/swin_v2_b_naip_rgb_satlas/resolve/main/swin_v2_b_naip_rgb_satlas-685f45bd.pth",  # noqa: E501
+    NAIP_RGB_SI_SATLAS = Weights(
+        url="https://huggingface.co/allenai/satlas-pretrain/resolve/main/aerial_swinb_si.pth",  # noqa: E501
```
adamjstewart, line 63: We usually include the first 8+ characters of the sha256sum in the suffix of the URL so that functions like https://pytorch.org/docs/stable/hub.html#torch.hub.load_state_dict_from_url can automatically verify the download checksum.

torchgeo/models/swin.py
```diff
@@ -63,6 +83,57 @@ class Swin_V2_B_Weights(WeightsEnum):  # type: ignore[misc]
...
+                "B03",
+                "B04",
+                "B05",
+                "B06",
+                "B07",
+                "B08",
+                "B09",
+                "B10",
+                "B11",
+            ],  # noqa: E501
```
adamjstewart, line 133: `# noqa` isn't being used here

torchgeo/transforms/transforms.py
```diff
@@ -53,7 +53,7 @@ def __init__(
             else:
                 keys.append(key)
 
-        self.augs = K.AugmentationSequential(*args, data_keys=keys, **kwargs)
+        self.augs = K.AugmentationSequential(*args, data_keys=keys, **kwargs)  # type: ignore[arg-type] # noqa: E501
```
adamjstewart, line 56: Why is arg-type ignored here?

piperwolters: We use kornia.contrib.Lambda for a few operations in the transforms functions, so the linter was getting upset about that. But there isn't another way to do those operations.

docs/api/sentinel2_pretrained_weights.csv
adamjstewart: Ah, good point. Yeah, that's fine for now. Someday I want to design a better naming scheme, but it's basically impossible since a model _could_ be trained on any subset of bands in any order. As long as it's unique I don't think the name matters that much, the enum contains the important info anyway.

torchgeo/transforms/transforms.py
adamjstewart: It should be relatively easy to create a custom transform that subclasses the right class instead of using a Lambda. But let me report this typing bug to our Kornia contacts, should be really simple to fix.

adamjstewart:
Another comment on this. Our unit tests do not actually attempt to download the model. We have separate tests for this that are only run before each release. If you run:
```console
> pytest -m slow tests/models/test_swin.py
```
you'll see that these models fail due to:
```
RuntimeError: Attempting to deserialize object on a CUDA device but torch.cuda.is_available() is False. If you are running on a CPU-only machine, please use torch.load with map_location=torch.device('cpu') to map your storages to the CPU.
```
We should either change the weights to be for the CPU (this is what we did for all other model weights in TorchGeo) or set the `map_location` for all models.

adamjstewart:
Also see #1916, I'm changing the URLs for all of our models to ensure that the md5 is reproducible even if you make this kind of change.

adamjstewart:
@piperwolters did you ever get a chance to apply the change requests here?

adamjstewart:
Soft reminder @piperwolters 😅

docs/api/naip_pretrained_weights.csv
adamjstewart: Wait, but https://huggingface.co/allenai/satlas-pretrain has an Apache-2.0 license?

Issue #1830, opened 2024-01-29
Migrate from Radiant MLHub to Source Cooperative


adamjstewart:
### Summary

We currently have several datasets from the recently defunct Radiant MLHub that we need to switch to Source Cooperative if we want to be able to automatically download them.

- [x] AgriFieldNet: #2180
- [x] Benin Cashews: #2116
- [x] Cloud Cover: #2117
- [x] CV4A Kenya Crop Type: #2090
- [x] Tropical Cyclone: #2068
- [x] NASA Marine Debris: #2183
- [x] Rwanda Field Boundary: #2118
- [x] South Africa Crop Type: #2181
- [ ] SpaceNet: #2203
- [x] Western USA Live Fuel Moisture: #2206

### Rationale

Downloads are currently broken, and many of these datasets have completely changed their file structure.

### Implementation

See https://github.com/microsoft/torchgeo/pull/2068 for an example that converts Tropical Cyclone. The resulting implementation is actually significantly simpler than the original code.

If you would like to volunteer to convert a particular dataset, please comment on this issue to say that you're working on this.

### Alternatives

We can also rehost most datasets (depending on license) on Hugging Face.

Haimantika:
Hi @adamjstewart I am interested to contribute to this, but I am fairly new and will need more guidance. What is a good place for me to start?

adamjstewart:
Hi @Haimantika, thanks for volunteering!

Let's pick a single dataset, maybe `torchgeo/datasets/benin_cashews.py`, and try to convert it to the new syntax. Once that's working, we can repeat for the other 7 datasets and remove all mention of radiant-mlhub.

This is the new dataset website: https://beta.source.coop/technoserve/cashews-benin/

If you create an account, log in, and click generate credentials, you'll see that the Azure URI is https://radiantearth.blob.core.windows.net/mlhub/technoserve-cashew-benin

We'll add a new dependency on [azure-storage-blob](https://pypi.org/project/azure-storage-blob/) in `pyproject.toml`, `requirements/datasets.txt`, and `requirements/min-reqs.old`. I can help determine the minimum supported version.

We'll probably add something similar to `download_radiant_mlhub_dataset` but for Azure blobs in `torchgeo/datasets/utils.py`. This can then be imported in `torchgeo/datasets/benin_cashews.py` and used in `_download`.

Let me know if anything is unclear. The first dataset is going to be a bit of work, but once we have one working, the rest should be easy.

[...]

PR #1887, opened 2024-02-16
Datasets: add download_azure_container utility function

Haimantika:
Needed for #1830


Haimantika:
> @Haimantika please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

pyproject.toml
```diff
@@ -93,6 +93,8 @@ datasets = [
     "pyvista>=0.34.2",
     # radiant-mlhub 0.3+ required for newer tqdm support required by lightning
     "radiant-mlhub>=0.3",
+    # source-cooperative 
+    "source-cooperative>=beta",
```
adamjstewart, line 97: Tests are failing because this isn't a real Python library

Haimantika: Ohh yes! should be cooperative, and version is 0.1.12. Fixing right away

pyproject.toml
```diff
@@ -93,6 +93,8 @@ datasets = [
     "pyvista>=0.34.2",
     # radiant-mlhub 0.3+ required for newer tqdm support required by lightning
     "radiant-mlhub>=0.3",
+    # source-cooperative 
+    "cooperative>=0.1.12",
```
adamjstewart, line 97: https://pypi.org/project/cooperative/ isn't related to what you're trying to do

yichiac: Source Cooperative uses azcopy for downloading; maybe [azure-storage-blob](https://pypi.org/project/azure-storage-blob/) will be useful.

adamjstewart:
@Haimantika are you still working on this? @darkblue-b may be willing to help finish this PR.

Haimantika:
> @Haimantika are you still working on this? @darkblue-b may be willing to help finish this PR.

Hi, I have tried wrapping my head around it, but I couldn’t get past. Apologies for not finishing it :(

adamjstewart:
Do you mind if @darkblue-b takes it from here? Can he push directly to your PR branch?

Haimantika:
> Do you mind if @darkblue-b takes it from here? Can he push directly to your PR branch?

Yes please. I will follow the changes to learn :)
Thank you again for giving me the opportunity to try!

adamjstewart:
I tested that the TropicalCyclone dataset can be downloaded using:
```python
split = "train"
account_url = "https://radiantearth.blob.core.windows.net"
container_name = "mlhub"
name_starts_with = f"nasa-tropical-storm-challenge/{split}"
```
However, much of the dataset format has changed, so this dataset will require a large rewrite. I'm going to add tests and merge this just as a new utility function, then we can open separate PRs for each dataset.

torchgeo/datasets/utils.py
```diff
@@ -266,6 +266,41 @@ def download_radiant_mlhub_collection(
...
+    try:
+        from azure.storage.blob import ContainerClient
+    except ImportError:
+        raise ImportError(
+            "azure-storage-blob is not installed and is required "
+            "to download this dataset"
+        )
+
+    client = ContainerClient(*args, **kwargs)
+    for blob in client.list_blob_names(name_starts_with=name_starts_with):
```
adamjstewart, line 297: Could wrap this with multiprocessing to speed things up, tqdm to add a progress bar, and various other improvements. There is also an asynchronous version of the API, although I don't think that will benefit us here.

torchgeo/datasets/utils.py
```diff
@@ -266,6 +266,41 @@ def download_radiant_mlhub_collection(
     collection.download(output_dir=download_root, api_key=api_key)
 
 
+def download_azure_container(
+    root: str, name_starts_with: str, *args: Any, **kwargs: Any
+) -> None:
+    """Download a container from Azure blob storage.
+
```
calebrob6, line 273: I would just add some examples of how to actually instantiate ContainerClient here

adamjstewart: Would it help if we made `account_url` and `container_name` explicit parameters instead of putting them in args/kwargs? They are required by ContainerClient, but I didn't explicitly list them to keep the code as simple as possible. Either way, this function currently doesn't show up in the docs, as it's only intended for use in TorchGeo, not use in other programs.

torchgeo/datasets/utils.py
adamjstewart: TODO: compare download speeds to the CLI version and see if multiprocessing is required to make this comparable.

tests/datasets/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+            for file in files:
+                name = os.path.join(root, file).replace(prefix + os.sep, "", 1)
+                if name.startswith(name_starts_with):
+                    yield name
+
+    def download_blob(self, blob: str) -> BinaryIO:
+        path = os.path.join(self.account_url, self.container_name, blob)
+        # TODO: filehandle leak
+        f = open(path, "rb", buffering=0)
+        return f
```
adamjstewart, line 29: I don't really know what to do about this. On macOS, we're only allowed to open 256 files at a time. I don't _think_ any of our text data will exceed this, but I don't know how else to code this to mimic the actual behavior.

torchgeo/datasets/utils.py
adamjstewart: At first glance, this is orders of magnitude slower than `azcopy sync`. I'll try parallelizing it, but if it's still noticeably slower, I suggest we rely on the CLI tool instead. Downside is that it can't be pip install'ed, but being slow is a bigger sin.

adamjstewart:
Unfortunately I think manual downloads with `azure.storage.blob` are simply too slow. I think we'll have to give up and use `azcopy sync` instead.

PR #1889, opened 2024-02-17
Add Sentinel2_CDL Datamodule

yichiac:
This PR adds the datamodule for CDL + Sentinel-2 images.


yichiac:
Are there any changes required to merge this datamodule? 
Tagging @adamjstewart for review.

torchgeo/datamodules/cdlsentinel2.py
```diff
@@ -0,0 +1,107 @@
...
+            length: Length of each training epoch.
+            num_workers: Number of workers for parallel data loading.
+            **kwargs: Additional keyword arguments passed to
+                :class:`~torchgeo.datasets.CDL`.
+        """
+        cdl_signature = "cdl_"
+        sentinel2_signature = "sentinel2_"
+        self.cdl_kwargs = {}
+        self.sentinel2_kwargs = {}
+        for key, val in kwargs.items():
```
calebrob6, line 47: Can you add some comments to describe what this logic is doing?

torchgeo/datamodules/cdlsentinel2.py
```diff
@@ -0,0 +1,107 @@
...
+        """Set up datasets and samplers.
+
+        Args:
+            stage: Either 'fit', 'validate', 'test', or 'predict'.
+        """
+        self.sentinel2 = Sentinel2(**self.sentinel2_kwargs)
+        self.cdl = CDL(**self.cdl_kwargs)
+        self.dataset = self.sentinel2 & self.cdl
+
+        generator = torch.Generator().manual_seed(0)
```
calebrob6, line 89: Can you confirm that you get the same random splits over different initializations?

torchgeo/datamodules/cdlsentinel2.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def __init__(
+        self,
+        batch_size: int = 64,
+        patch_size: Union[int, tuple[int, int]] = 16,
+        length: Optional[int] = None,
+        num_workers: int = 0,
+        **kwargs: Any,
+    ) -> None:
+        """Initialize a new CDLSentinel2DataModule instance.
```
calebrob6, line 33: I think you will need to document how to use this datamodule -- it looks like this is a wrapper around Sentinel2 and CDL data. So maybe give an example in the constructor of how to instantiate it?

tests/conf/cdlsentinel2.yaml
```diff
@@ -0,0 +1,18 @@
...
+    num_filters: 1
+    ignore_index: 0
+data:
+  class_path: CDLSentinel2DataModule
+  init_args:
+    batch_size: 2
+    patch_size: 16
+  dict_kwargs:
+    cdl_paths: "tests/data/cdl"
+    sentinel2_paths: "tests/data/sentinel2"
```
calebrob6, line 18: Can you add a newline (before Adam asks 😄 )

calebrob6 (changes requested):

calebrob6:
Also need to add this to docs

torchgeo/datamodules/cdlsentinel2.py
adamjstewart: We have the same logic in all of our A & B GeoDataModules. We could extract it out into a helper function now that it's becoming common. Would that help with the documentation?

torchgeo/datamodules/cdlsentinel2.py
adamjstewart: You do not: #1899

torchgeo/datamodules/cdlsentinel2.py
adamjstewart: It's the same as NAIPChesapeakeDataModule. If we add docs to this, we should add docs to that too.

adamjstewart:
Needs same changes as NCCM.

torchgeo/datamodules/cdlsentinel2.py
yichiac: Since #1899 is fixed by #1908, we can get the same random splits over different initializations.

yichiac:
I need to investigate why increasing the number of grids causes errors. I tried to increase the test data sizes for CDL and Sentinel2 by 10x larger. But I can only get `grid_size` up to 3 without getting the error. When `grid_size`==4, the error shows:

`IndexError: query: BoundingBox(minx=399960.0, maxx=400120.0, miny=4496400.0, maxy=4496560.0, mint=1649798921.0, maxt=1649798921.999999) not found in index with bounds: BoundingBox(minx=400860.0, maxx=403560.0, miny=4496400.0, maxy=4499100.0, mint=1649798921.0, maxt=1649798921.999999)`

Ideally, if Sentinel2 size is 360x360, it should be fine being divided by 4x4 grids.

torchgeo/datamodules/cdlsentinel2.py
yichiac: I added the additional description. I can add docs to this datamodule if we need to add them.

yichiac:
The error was fixed by changing `self.dataset` to `self.test_dataset` in `datamodules/cdlsentinel2.py`.
```
if stage in ["test"]:
    self.test_sampler = GridGeoSampler(
        self.test_dataset, self.patch_size, self.patch_size
    )
```

torchgeo/datamodules/sentinel2cdl.py
```diff
@@ -0,0 +1,110 @@
...
+            elif key.startswith(sentinel2_signature):
+                # If so, extract the key-value pair to the Sentinel-2 dictionary
+                self.sentinel2_kwargs[key[len(sentinel2_signature) :]] = val
+
+        super().__init__(
+            CDL, batch_size, patch_size, length, num_workers, **self.cdl_kwargs
+        )
+
+        self.train_aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
```
adamjstewart, line 65: Note that we're currently dividing by 255 in order to normalize our S2 images. Not sure if this is correct or not, you should check.

torchgeo/datamodules/sentinel2cdl.py
```diff
@@ -0,0 +1,110 @@
...
+        """
+        self.sentinel2 = Sentinel2(**self.sentinel2_kwargs)
+        self.cdl = CDL(**self.cdl_kwargs)
+        self.dataset = self.sentinel2 & self.cdl
+
+        generator = torch.Generator().manual_seed(0)
+
+        (self.train_dataset, self.val_dataset, self.test_dataset) = (
+            random_grid_cell_assignment(
+                self.dataset, [0.7, 0.10, 0.20], grid_size=8, generator=generator
```
adamjstewart, line 91: Is 70-10-20 what we're using for all datasets? Just want to make sure we're being consistent.

adamjstewart (approved):
Mostly looks good now. Look through the commits I added and make sure they all make sense. Couple final comments worth reviewing before we meet. I'll try to get the other PRs in similar shape so we can merge after our meeting today

adamjstewart (dismissed):

torchgeo/datamodules/sentinel2cdl.py
yichiac: [This paper](https://www.mdpi.com/2072-4292/12/2/207#secAdot1-remotesensing-12-00207) shows this splitting can reduce the performance-inflating effect. This method can be used for most Image + Mask datamodules.

yichiac:
@adamjstewart Thanks for the review!

torchgeo/datamodules/sentinel2cdl.py
adamjstewart: That paper is over 4 years old, not sure if I would trust it too much

adamjstewart (dismissed):

torchgeo/datamodules/sentinel2_cdl.py
```diff
@@ -0,0 +1,121 @@
...
+    def __init__(
+        self,
+        batch_size: int = 64,
+        patch_size: Union[int, tuple[int, int]] = 64,
+        length: Optional[int] = None,
+        num_workers: int = 0,
+        **kwargs: Any,
+    ) -> None:
+        """Initialize a new Sentinel2CDLDataModule instance.
+
```
calebrob6, line 35: Can you give an example of how to instantiate this class? It is non-obvious from the current docstring.

torchgeo/datamodules/sentinel2_cdl.py
```diff
@@ -0,0 +1,121 @@
...
+        """Run CDL plot method.
+
+        Args:
+            *args: Arguments passed to plot method.
+            **kwargs: Keyword arguments passed to plot method.
+
+        Returns:
+            A matplotlib Figure with the image, ground truth, and predictions.
+        """
+        return self.cdl.plot(*args, **kwargs)
```
calebrob6, line 121: Does this plot work when using this with a SemanticSegmentation trainer?

calebrob6 (commented):
The code looks fine, but I want to make sure that you have actually tried to train a model with this!

torchgeo/datamodules/sentinel2_cdl.py
yichiac: The datamodule can be initialized like this:
```
datamodule = Sentinel2CDLDataModule(
    crs="epsg:3857",
    batch_size=64,
    patch_size=224,
    cdl_paths="data/cdl/",
    sentinel2_paths="data/sentinel2/",
)
```

torchgeo/datamodules/sentinel2_cdl.py
adamjstewart: Yes, it's also tested in CI. It simply overrides the default that calls `self.dataset.plot`

adamjstewart (approved):
There are still a few details we need to work out for our paper experiments w.r.t. augmentations, but we can do that in a mass PR once all data modules are merged.

I agree with @calebrob6 that the documentation could always be improved, but this is also true for our existing datamodules, none of which give examples for instantiation. The `foo_` prefix stuff is already used in some of our existing datamodules like NAIPChesapeake with no explanation. Let's make a tutorial for these someday that gives example usage. For now, I'll force merge this so we aren't holding up all other datamodule PRs.

PR #1897, opened 2024-02-21
Added custom semantic segmentation trainer tutorial

calebrob6:
This is a tutorial notebook that shows users how to override a custom semantic segmentation class for training on LandCoverAI.


calebrob6:
@adamjstewart I think there is a bug (or at least some weird behavior going on) here

I can create a new class that extends SemanticSegmentationTask:

```

    # any keywords we add here between *args and **kwargs will be found in self.hparams
        super().__init__(*args, **kwargs)  # pass args and kwargs to the parent class
```

Then instantiate and everything works great
```
task = CustomSemanticSegmentationTask(model="unet", tmax=100, ...)
```

However when I go to load from file:
```
task = CustomSemanticSegmentationTask.load_from_checkpoint("lightning_logs/version_3/checkpoints/epoch=0-step=117.ckpt")
```

I get an error:
```
TypeError: SemanticSegmentationTask.__init__() got an unexpected keyword argument 'ignore'
```

I can add `del kwargs["ignore"]` before `super()....` in the constructor of `CustomSemanticSegmentationTask` but this feels like a bad hack.

(@isaaccorley in case you've seen this)

robmarkcole:
Note that https://www.reviewnb.com/ is free for open source use - would enable previewing the notebook

docs/tutorials/custom_segmentation_trainer.py
```diff
@@ -0,0 +1,194 @@
...
+)
+
+# validate that the task's hyperparameters are as expected
+task.hparams
+
+# +
+accelerator = "gpu" if torch.cuda.is_available() else "cpu"
+
+trainer = pl.Trainer(
+    accelerator=accelerator,
```
robmarkcole, line 174: Noting this is not required on Lightning 2.0 upwards

docs/tutorials/custom_segmentation_trainer.py
```diff
@@ -0,0 +1,194 @@
...
+import warnings
+
+warnings.filterwarnings("ignore", category=UserWarning, module="torch.nn.functional")
+
+
+# -
+
+# ## Custom SemanticSegmentationTask
+#
+# Now, we create a `CustomSemanticSegmentationTask` class that inhierits from `SemanticSegmentationTask` and that overrides a few methods:
```
robmarkcole, line 68: typo: inhierits

docs/tutorials/custom_segmentation_trainer.py
```diff
@@ -0,0 +1,194 @@
...
+    max_epochs=300,
+    log_every_n_steps=50,
+)
+# -
+
+trainer.fit(task, dm)
+
+# ## Test model
+#
+# Finally, we test the model on the test set and visualize the results.
```
robmarkcole, line 188: `visualize the results` - would make me expect some plots, it is a table of metrics however

calebrob6:
@robmarkcole thanks for the review! Was that easy enough to do (vs reviewnb)?

On my side, I just have to run `jupytext --sync custom_segmentation_trainer.ipynb` whenever I change either file and commit both (which seems simple enough). We could have CI that makes sure that all py and ipynb are in sync.

docs/tutorials/custom_segmentation_trainer.py
calebrob6: Yep -- not finished here. Need to get some agreement on:
1. How to review
2. How to deal with long running notebooks in CI. If I run this notebook for 250 epochs (7 hours on my machine) then I get 85.95 test mIoU which is better than the results reported in https://arxiv.org/pdf/2005.02264.pdf. I think fake datasets are less interesting.
3. How to deal with the above issue I pointed out where the `ignore` parameter is being saved as a hparam, then breaking the subclass constructor.

robmarkcole: I think it would be fine to train for 1 batch as this notebook is about demonstrating the technique, in which case a CPU can be used

robmarkcole:
Yep `easy enough`

isaaccorley:
> @adamjstewart I think there is a bug (or at least some weird behavior going on) here
> 
> 
> 
> I can create a new class that extends SemanticSegmentationTask:
> 
> 
> 
> ```
> 
> class CustomSemanticSegmentationTask(SemanticSegmentationTask):
> 
> 
> 
>     # any keywords we add here between *args and **kwargs will be found in self.hparams
> 
>     def __init__(self, *args, tmax=50, eta_min=1e-6, **kwargs) -> None:
> 
>         super().__init__(*args, **kwargs)  # pass args and kwargs to the parent class
> 
> ```
> 
> 
> 
> Then instantiate and everything works great
> 
> ```
> 
> task = CustomSemanticSegmentationTask(model="unet", tmax=100, ...)
> 
> ```
> 
> 
> 
> However when I go to load from file:
> 
> ```
> 
> task = CustomSemanticSegmentationTask.load_from_checkpoint("lightning_logs/version_3/checkpoints/epoch=0-step=117.ckpt")
> 
> ```
> 
> 
> 
> I get an error:
> 
> ```
> 
> TypeError: SemanticSegmentationTask.__init__() got an unexpected keyword argument 'ignore'
> 
> ```
> 
> 
> 
> I can add `del kwargs["ignore"]` before `super()....` in the constructor of `CustomSemanticSegmentationTask` but this feels like a bad hack.
> 
> 
> 
> (@isaaccorley in case you've seen this)
> 
> 
> 
> 

I don't see an ignore param in your custom class. Did you modify the class code you used after training that checkpoint? If so, one thing you can do is to just load the checkpoint, delete that param in the hparam dict and then save the checkpoint which would fix the error.

calebrob6:
One workaround that I'm checking is just adding `"ignore"` to the list of args ignored in `BaseTask`.

isaaccorley:
> One workaround that I'm checking is just adding `"ignore"` to the list of args ignored in `save_hyperparameters` in `BaseTask`.

I think this makes the most sense since it's not an actual hparam.

calebrob6:
The way that I've come up with to check whether an .ipynb is in sync with the corresponding .py is ` diff <(jupytext --output - --to py custom_segmentation_trainer.ipynb) custom_segmentation_trainer.py`. This should only output something like:
```
3a4
> #     formats: ipynb,py
```

calebrob6:
I don't know why notebook test is being cancelled (maybe because it is trying to run the LandCoverAI split script?).

adamjstewart:
> I don't know why notebook test is being cancelled (maybe because it is trying to run the LandCoverAI split script?).

Tests being canceled means the job either ran out of time, space, or memory. Here, my guess would be space. We want to use the smallest datasets possible, such as EuroSAT100. Might be worth creating a LandCoverAI100 or something like that.

adamjstewart:
Haven't yet had time to evaluate jupytext to decide whether or not it's what we should use. @nilsleh what did you end up using for lightning-uq-box?

calebrob6:
I really don't like the idea of making custom datasets/datamodules just to have pretty CI -- it is a large overhead for something that makes the tutorial less cool.

adamjstewart:
And I really don't like having tutorials that take 30+ minutes to download a dataset and train a model for hundreds of epochs, or tutorials that can't be tested in CI because they involve more data than our runners can store. There's always a tradeoff. You can also find a smaller dataset instead of making your own.

docs/tutorials/custom_segmentation_trainer.py
```diff
@@ -0,0 +1,197 @@
...
+        else:
+            lr = optimizers.param_groups[0]["lr"]
+        self.logger.experiment.add_scalar("lr", lr, self.current_epoch)  # type: ignore
+
+
+# ## Train model
+#
+# The remainder of the turial is straightforward and follows the typical [PyTorch Lightning](https://lightning.ai/) training routine. We instantiate a `DataModule` for the LandCover.AI dataset, instantiate a `CustomSemanticSegmentationTask` with a U-Net and ResNet-50 backbone, then train the model using a Lightning trainer.
+
+dm = LandCoverAIDataModule(root="data/", batch_size=64, num_workers=8, download=True)
```
calebrob6, line 163:
```suggestion
dm = LandCoverAIDataModule(root="data/", batch_size=2, num_workers=8, download=True)
```

calebrob6:
Luckily none of that happens here ;). LandCover.ai is 1.5GB (this will take 30+ minutes to download if your download speed is < 0.83 MB/s), training happens for 1 batch (and I can reduce the batch size to make this faster).

I'm saying that we shouldn't be catching bugs with overly sanitized examples -- if LandCoverAI breaks or Lightning training breaks then our other tests will break. If the example notebooks **are** catching bugs then we should ask ourselves why. Downloading LandCoverAI and running this now and per release doesn't seem to be a big burden.

nilsleh:
> Haven't yet had time to evaluate jupytext to decide whether or not it's what we should use. @nilsleh what did you end up using for lightning-uq-box?

So I tried jupytext, but for my tutorials I couldn't get the jupytext scripts to execute and be displayed on the documentation. I went back to notebooks, and the notebooks are now run when the documentation builds. However, I don't need to download any data and model fitting is fast, since it's just toy problems.

adamjstewart:
> How about-- is it possible to change LandCoverAI datamodule to use the test data that we already have spent time creating for this notebook (then comments saying if you _actually_ want to play with data, then do this other thing)?

This is also undesirable because the notebook by default will train and display predictions on random noise. I would much rather have a tiny dataset with real data. I'm happy to make this myself (maybe next week).

calebrob6:
I don't think it needs to display predictions -- as if we're only training for a batch for "making CI pretty" reasons then it will display noise regardless.

adamjstewart:
We can train for multiple epochs in the tutorial, but use `fast_dev_run` in CI. The trainers tutorial does this with nbmake variable mocking. This also means we could use the synthetic dataset during CI and the real dataset during the tutorial. My only hope is that we don't train for longer than it takes to explain what we're doing in a live tutorial session. If we use this notebook in a tutorial and end up sitting there watching it awkwardly for an hour then it gets tedious.

calebrob6:
@adamjstewart -- checking in on this, do we want to force a LandCoverAI100 Datamodule for this?

adamjstewart:
If the tests are failing due to OOM then we'll have to do something like that. I think it will demonstrate how quick and easy it is to get a working model using a pre-trained backbone and a tiny dataset. We could even add a pre-trained U-Net model.

calebrob6:
Yeah I'd love to follow this up with a "how to use pre-trained models" -- just need to figure this one out first

calebrob6:
Adding LandCoverAI100 in https://github.com/microsoft/torchgeo/pull/2262

robmarkcole:
As a nice addition, you could show how to export a la https://qgis-plugin-deepness.readthedocs.io/en/latest/example/example_segmentation_landcover.html

calebrob6:
Easy link to the notebook -- https://github.com/microsoft/torchgeo/blob/tutorial/week1/docs/tutorials/custom_segmentation_trainer.ipynb

calebrob6:
Tests are failing on importing `LandCoverAI100DataModule` -- is this because we install torchgeo in first cell?

adamjstewart:
Yes, see #2300 for the fix.

sfalkena:
Fix is implemented in #2306

calebrob6:
Nice, thanks @sfalkena! What are the odds of finding this bug at nearly the same time 🙂

.github/workflows/tutorials.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
       - name: Install pip dependencies
         if: steps.cache.outputs.cache-hit != 'true'
         run: |
-          pip install -r requirements/required.txt -r requirements/docs.txt -r requirements/tests.txt planetary_computer pystac .
+          pip install -r requirements/required.txt -r requirements/docs.txt -r requirements/tests.txt tensorboard planetary_computer pystac .
```
calebrob6, line 36: tensorboard is used by the trainers notebook, if we install it here it should get cached and not take up actual test time

docs/tutorials/pretrained_weights.ipynb
```diff
@@ -473,9 +473,6 @@
  ],
  "metadata": {
   "accelerator": "GPU",
-  "colab": {
```
calebrob6, line 476: I ran https://github.com/kynan/nbstripout to get rid of extra metadata

.github/workflows/tutorials.yaml
adamjstewart: Can you sort these alphabetically? Also needs to be added to `.github/workflows/release.yaml`. 

Honestly, I wonder if we shouldn't add all 3 deps to the docs or tests extra. I would like to be able to pin these and test specific versions like we do with all other deps.

I kind of want to get rid of planetary_computer/pystac as there are fine tutorials for how to use these and torchgeo has never focused on raw data downloads. But I think things like tensorboard integration are more important to demonstrate, as they directly relate to how to use our trainers.

Can we separate this out into its own PR since it's unrelated to this new tutorial?

docs/tutorials/pretrained_weights.ipynb
adamjstewart: I didn't make as much progress as I had hoped with #2070. If we stick with notebooks, ideally we would use some kind of auto-formatter built into ruff instead of nbstripout. But regardless, we should add a test to CI so there isn't as much back and forth with these files. I would ignore this for now until we make a final decision on how to manage these notebooks.

torchgeo/trainers/base.py
```diff
@@ -35,6 +35,12 @@ def __init__(self, ignore: Sequence[str] | str | None = None) -> None:
             ignore: Arguments to skip when saving hyperparameters.
         """
         super().__init__()
+        if ignore is None:
+            ignore = ['ignore']
+        elif isinstance(ignore, str):
+            ignore = [ignore, 'ignore']
+        else:
+            ignore = [*list(ignore), 'ignore']
```
adamjstewart, line 43: I think this deserves its own PR, I don't fully understand why this is needed.

calebrob6: This is described in the first few comments of the PR

docs/tutorials/pretrained_weights.ipynb
calebrob6: Does this matter?

calebrob6:
Okay this now depends on https://github.com/microsoft/torchgeo/pull/2314

isaaccorley (dismissed):
Lgtm

docs/tutorials/custom_segmentation_trainer.ipynb
```diff
@@ -0,0 +1,514 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "dm = LandCoverAI100DataModule(root='data', batch_size=10, num_workers=2, download=True)\n",
+    "\n",
+    "# You can use the following for actual training runs\n",
+    "# from torchgeo.datamodules import LandCoverAIDataModule\n",
+    "# dm = LandCoverAIDataModule(root='data/', batch_size=64, num_workers=8, download=True)"
```
adamjstewart, line 217:
```suggestion
    "# dm = LandCoverAIDataModule(root='data', batch_size=64, num_workers=8, download=True)"
```

docs/tutorials/custom_segmentation_trainer.ipynb
```diff
@@ -0,0 +1,514 @@
...
+       "\"tmax\":            50"
+      ]
+     },
+     "execution_count": null,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "# validate that the task's hyperparameters are as expected\n",
```
adamjstewart, line 268: At some point we transitioned from markdown to code comments. I would prefer markdown.

docs/tutorials/custom_segmentation_trainer.ipynb
```diff
@@ -0,0 +1,514 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
```
adamjstewart, line 245: I usually strip outputs from the notebooks before uploading them so the outputs don't change every time someone submits a PR. I know you prefer to include the output, and I'm trying to find ways to make both of us happy. Ideally, I want to store only `.py` files and generate the `.ipynb` on-the-fly but haven't figured out a way to do that yet.

docs/tutorials/custom_segmentation_trainer.ipynb
```diff
@@ -0,0 +1,514 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# You can load directly from a saved checkpoint with `.load_from_checkpoint(...)`\n",
+    "# Note that you can also just call `trainer.test(task, dm)` if you've already trained\n",
+    "# the model in the current notebook session.\n",
+    "\n",
+    "task = CustomSemanticSegmentationTask.load_from_checkpoint(\n",
```
adamjstewart, line 406: I know this is like the whole purpose of #2317, but also completely unnecessary to demonstrate a custom trainer. If you remove this, we could backport this to 0.6.1 so it makes it into the stable docs immediately.

docs/tutorials/custom_segmentation_trainer.ipynb
```diff
@@ -0,0 +1,514 @@
...
+       "└───────────────────────────┴───────────────────────────┘\n"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    },
+    {
+     "data": {
+      "text/plain": [
+       "[{'test_loss': 17.85824203491211,\n",
```
adamjstewart, line 472: Output metrics are pretty bad. At least with our EuroSAT100 examples, the accuracies aren't horrible (much higher than random guessing). Maybe if we start with a pre-trained model (even one created solely for the purpose of this tutorial) we can get better results? Would also be nice to include example plots at the end to visualize how well the model works. Again, I know this is something you wish we didn't have to do and we could instead train for minutes/hours, but we want to keep the tutorial simple and easy to walkthrough.

adamjstewart (changes requested):

docs/tutorials/custom_segmentation_trainer.ipynb
calebrob6: A few things:
- Integrating pre-trained models is outside the scope of this tutorial. If we get to a point where we post benchmark models for different datasets then we can revisit.
- Performance on this task is close to meaningless -- it is semantic segmentation over 15 samples training for a single look on 70 samples. It may be the 15 random samples in the test set are all of the same class.
- It'd be cute to maximize performance under the constraints of CPU only Github Action CI runners but I don't think it adds any value to the tutorial and eats up CI time.
- It would be nice to have pretty plots, but again, you are very unlikely to get anything coherent from training for a single 70 sample epoch. Also, plotting should be covered by other tutorials.

docs/tutorials/custom_segmentation_trainer.ipynb
calebrob6: I'm not in a rush to get this into the stable docs, and think it is a nice use case.

docs/tutorials/custom_segmentation_trainer.ipynb
calebrob6: TorchVision, Kornia, Sklearn, Lightning, etc. all have notebook output. If you don't show the notebook output then you should just have .py files?

https://pytorch.org/vision/stable/auto_examples/transforms/plot_transforms_getting_started.html#sphx-glr-auto-examples-transforms-plot-transforms-getting-started-py

https://kornia.github.io/tutorials/nbs/image_matching_lightglue.html

docs/tutorials/custom_segmentation_trainer.ipynb
calebrob6: The point is to show people how to build custom classes that extend our base trainer classes and why they would do that (to have different metrics, to change optimizer schedule, ....).

calebrob6:
@adamjstewart is this good to go?

adamjstewart:
Need to give this another detailed review, but there are a lot of lines of code that are commented out that we should clean up. Also, this PR depends on multiple new features that won't exist until 0.7.0/1.0.0, so we probably can't use it for AGU (unless we backport some of the changes without advertising them).

PR #1898, opened 2024-02-22
Correct behavior of ignore_index for JaccardLoss

### LINKED ISSUE #1891: Add `ignore_index` support for Jaccard Loss
### Summary

Currently, the `SemanticSegmentationTask` recognises the `ignore_index` parameter only when cross entropy and focal loss is used. However, `smp.losses.JaccardLoss` implicitly supports this option via its `classes` parameter, which is currently set to `self.hparams["num_classes"]`. The FR is to adapt what is passed as argument to this parameter so that `ignore_index` can work with all currently supported losses.

### Rationale

The Jaccard index is a common semantic segmentation metric which also makes for a decent loss function. Because of the way it is defined, it is important to ignore overly dominant classes (e.g., the background when classifying building rooftops); otherwise performance can be hindered significantly.

### Implementation

Change the `classes` argument of `smp.losses.JaccardLoss` in `SemanticSegmentationTask.configure_losses` from `self.hparams["num_classes"]` to `list(set(list(range(self.hparams["num_classes"]))).difference(set([ignore_index])))`, assuming that `ignore_index` is not `None`.

### Alternatives

_No response_

### Additional information

_No response_

### LINKED ISSUE #1896: trainers.segmentation JaccardLoss receiving num_classes, should be a List[int]?
### Description

From [the trainer](https://github.com/microsoft/torchgeo/blob/8defbe4358b488f1eaa0712094c07d3f802f7523/torchgeo/trainers/segmentation.py#L113C1-L116C14):
```python
        elif loss == "jaccard":
            self.criterion = smp.losses.JaccardLoss(
                mode="multiclass", classes=self.hparams["num_classes"]
            )
```
`num_classes` is an int, Yet the docs for [JaccardLoss](https://smp.readthedocs.io/en/latest/losses.html#jaccardloss) state this is expected to be a _list of int_, presumably the indices of the output channels to use in the calc. Checking if this is a bug

### Steps to reproduce

NA

### Version

main

calebrob6:
Fixes #1896 
Fixes #1891

### DIFF
--- a/torchgeo/trainers/segmentation.py
+++ b/torchgeo/trainers/segmentation.py
@@ -104,13 +104,14 @@ class SemanticSegmentationTask(BaseTask):
                 ignore_index=ignore_value, weight=self.hparams["class_weights"]
             )
         elif loss == "jaccard":
-            # JaccardLoss requires a list of classes to use instead of a class index to ignore.
-            class_set = [
+            # JaccardLoss requires a list of classes to use instead of a class
+            # index to ignore.
+            classes = [
                 i for i in range(self.hparams["num_classes"]) if i != ignore_index
             ]
 
             self.criterion = smp.losses.JaccardLoss(
-                mode="multiclass", classes=class_set
+                mode="multiclass", classes=classes
             )
         elif loss == "focal":
             self.criterion = smp.losses.FocalLoss(
--- a/torchgeo/trainers/segmentation.py
+++ b/torchgeo/trainers/segmentation.py
@@ -110,9 +110,7 @@ class SemanticSegmentationTask(BaseTask):
                 i for i in range(self.hparams["num_classes"]) if i != ignore_index
             ]
 
-            self.criterion = smp.losses.JaccardLoss(
-                mode="multiclass", classes=classes
-            )
+            self.criterion = smp.losses.JaccardLoss(mode="multiclass", classes=classes)
         elif loss == "focal":
             self.criterion = smp.losses.FocalLoss(
                 "multiclass", ignore_index=ignore_index, normalized=True
--- a/torchgeo/trainers/segmentation.py
+++ b/torchgeo/trainers/segmentation.py
@@ -70,9 +70,6 @@ class SemanticSegmentationTask(BaseTask):
             freeze_decoder: Freeze the decoder network to linear probe
                 the segmentation head.
 
-        Warns:
-            UserWarning: When loss='jaccard' and ignore_index is specified.
-
         .. versionchanged:: 0.3
            *ignore_zeros* was renamed to *ignore_index*.
 
@@ -87,13 +84,10 @@ class SemanticSegmentationTask(BaseTask):
            The *weights* parameter now supports WeightEnums and checkpoint paths.
            *learning_rate* and *learning_rate_schedule_patience* were renamed to
            *lr* and *patience*.
-        """
-        if ignore_index is not None and loss == "jaccard":
-            warnings.warn(
-                "ignore_index has no effect on training when loss='jaccard'",
-                UserWarning,
-            )
 
+        .. versionchanged:: 0.6
+            The *ignore_index* parameter now works for jaccard loss.
+        """
         self.weights = weights
         super().__init__(ignore="weights")
 
--- a/torchgeo/trainers/segmentation.py
+++ b/torchgeo/trainers/segmentation.py
@@ -104,9 +104,11 @@ class SemanticSegmentationTask(BaseTask):
                 ignore_index=ignore_value, weight=self.hparams["class_weights"]
             )
         elif loss == "jaccard":
-            class_set = set(range(self.hparams["num_classes"]))
-            if ignore_index is not None:
-                class_set.remove(ignore_index)
+            class_set = [
+                i
+                for i in range(self.hparams["num_classes"])
+                if i != ignore_index
+            ]
 
             self.criterion = smp.losses.JaccardLoss(
                 mode="multiclass", classes=class_set
--- a/torchgeo/trainers/segmentation.py
+++ b/torchgeo/trainers/segmentation.py
@@ -4,7 +4,6 @@
 """Trainers for semantic segmentation."""
 
 import os
-import warnings
 from typing import Any, Optional, Union
 
 import matplotlib.pyplot as plt
--- a/torchgeo/trainers/segmentation.py
+++ b/torchgeo/trainers/segmentation.py
@@ -105,9 +105,7 @@ class SemanticSegmentationTask(BaseTask):
             )
         elif loss == "jaccard":
             class_set = [
-                i
-                for i in range(self.hparams["num_classes"])
-                if i != ignore_index
+                i for i in range(self.hparams["num_classes"]) if i != ignore_index
             ]
 
             self.criterion = smp.losses.JaccardLoss(
--- a/torchgeo/trainers/segmentation.py
+++ b/torchgeo/trainers/segmentation.py
@@ -111,8 +111,12 @@ class SemanticSegmentationTask(BaseTask):
                 ignore_index=ignore_value, weight=self.hparams["class_weights"]
             )
         elif loss == "jaccard":
+            class_set = set(range(self.hparams["num_classes"]))
+            if ignore_index is not None:
+                class_set.remove(ignore_index)
+
             self.criterion = smp.losses.JaccardLoss(
-                mode="multiclass", classes=self.hparams["num_classes"]
+                mode="multiclass", classes=class_set
             )
         elif loss == "focal":
             self.criterion = smp.losses.FocalLoss(
--- a/torchgeo/trainers/segmentation.py
+++ b/torchgeo/trainers/segmentation.py
@@ -104,6 +104,7 @@ class SemanticSegmentationTask(BaseTask):
                 ignore_index=ignore_value, weight=self.hparams["class_weights"]
             )
         elif loss == "jaccard":
+            # JaccardLoss requires a list of classes to use instead of a class index to ignore.
             class_set = [
                 i for i in range(self.hparams["num_classes"]) if i != ignore_index
             ]
--- a/tests/conf/chesapeake_cvpr_5.yaml
+++ b/tests/conf/chesapeake_cvpr_5.yaml
@@ -1,7 +1,7 @@
 model:
   class_path: SemanticSegmentationTask
   init_args:
-    loss: "ce"
+    loss: "jaccard"
     model: "unet"
     backbone: "resnet50"
     in_channels: 4
--- a/tests/conf/chesapeake_cvpr_5.yaml
+++ b/tests/conf/chesapeake_cvpr_5.yaml
@@ -7,7 +7,7 @@ model:
     in_channels: 4
     num_classes: 5
     num_filters: 1
-    ignore_index: null
+    ignore_index: 0
 data:
   class_path: ChesapeakeCVPRDataModule
   init_args:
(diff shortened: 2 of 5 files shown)


DimitrisMantas:
Hi, I'm the author of #1891.

Thanks for taking the time to fix the issue, and, most importantly not roast the horrendous one-line fix I originally proposed.

Perhaps we should update the trainer docatring and initialization to reflect the reposed changes?

calebrob6:
Haha no roasting from me (don't throw stones in a glass house and all that 😉)! We appreciate you taking the time to open an issue.

Good point -- we need to remove the UserWarning that we were previously throwing.

DimitrisMantas:
Also, I've noticed that smp converts `classes` to a `Tensor` like so:

```python
if classes is not None:
    assert mode != BINARY_MODE, "Masking classes is not supported with mode=binary" 
    classes = to_tensor(classes, dtype=torch.long)
```

where:

```python
def to_tensor(x, dtype=None) -> torch.Tensor:
    if isinstance(x, torch.Tensor):
        if dtype is not None:
            x = x.type(dtype)
        return x
    if isinstance(x, np.ndarray):
        x = torch.from_numpy(x)
        if dtype is not None:
            x = x.type(dtype)
        return x
    if isinstance(x, (list, tuple)):
        x = np.array(x)
        x = torch.from_numpy(x)
        if dtype is not None:
            x = x.type(dtype)
        return x
```

See https://github.com/qubvel/segmentation_models.pytorch/blob/6db76a1106426ac5b55f39fba68168f3bccae7f8/segmentation_models_pytorch/losses/jaccard.py#L45 and https://github.com/qubvel/segmentation_models.pytorch/blob/6db76a1106426ac5b55f39fba68168f3bccae7f8/segmentation_models_pytorch/losses/_functional.py#L18.

Since the class set is not a checked instance, `x` does not get transformed to the format smp expects.

I think it's OK in the end since `classes` ends up being used to index a loss array, but maybe we should still be good citizens and convert the set to a tuple?

isaaccorley (approved):

torchgeo/trainers/segmentation.py
```diff
@@ -111,8 +104,12 @@ def configure_losses(self) -> None:
                 ignore_index=ignore_value, weight=self.hparams["class_weights"]
             )
         elif loss == "jaccard":
+            class_set = set(range(self.hparams["num_classes"]))
```
robmarkcole, line 107: This will just include all classes? I've suggested an alternative approach at https://github.com/microsoft/torchgeo/issues/1896#issuecomment-1958965741

robmarkcole:
As commented I think this approach is incorrect. What is required is for Jaccardloss to correctly implement ignore_index, see how it is done in https://smp.readthedocs.io/en/latest/_modules/segmentation_models_pytorch/losses/focal.html#FocalLoss

torchgeo/trainers/segmentation.py
DimitrisMantas: But according to the smp docs, `classes` should be a list of the integer class labels to include in loss calculations.

In this context, I think using this approach is fine; the ignored index is removed from the set if it is not null.

torchgeo/trainers/segmentation.py
```diff
@@ -111,8 +104,12 @@ def configure_losses(self) -> None:
...
+            if ignore_index is not None:
+                class_set.remove(ignore_index)
```
calebrob6, line 109: @robmarkcole -- if the user sets `ignore_index` then it is removed from the `class_set` here and the remaining classes are  used in JaccardLoss.

robmarkcole: But `ignore_index` is a pixel value in the mask (e.g. 255) to ignore, unless I am mistaken?

robmarkcole: Correct, I've seen nodata as -9999, 0, 255 etc

DimitrisMantas: Just to be clear, my reasoning for opening this issue in the first place was that maybe you have a background class in the target masks that you don't care about explicitly learning, and so you ignore it in the loss even if you end up outputting it.

For instance, suppose I don't want to include the background class of LamdCoverAI in the loss. Then, I would just pass the corresponding index to `ignore_index`, and regardless of how many classes I chose to output, said class wouldn't be included in loss calculations...

calebrob6: I don't actually know how that is handled by JaccardLoss or others. Let me check it out and come back.

calebrob6: I'm back -- `JaccardLoss` assumes contiguous class values:

It sets `num_classes` to the size of the predicted mask's class dimension -- https://github.com/qubvel/segmentation_models.pytorch/blob/master/segmentation_models_pytorch/losses/jaccard.py#L69

Then it converts `y_true` to a 1-hot encoding with this number of classes -- https://github.com/qubvel/segmentation_models.pytorch/blob/master/segmentation_models_pytorch/losses/jaccard.py#L80

I confirmed this will throw an error if you have a mask with values like 255 but predictions with only 4 channels.

isaaccorley: Seems like ignore_index is the way to go so you would transform the integer mask to some values that you want to ignore in the loss. Too bad JaccardLoss doesn't have this as an input but I'm currently modifying it in torchseg.

This PR is probably okay as a workaround for now though

calebrob6: @DimitrisMantas, yep totally understand. We're now thinking about the case where you have a mask where the values **aren't** {0, ..., C-1}. This happens often in practice (for example, NLCD land cover data have values 11, 12, 21, ... for whatever reason -- https://www.mrlc.gov/data/legends/national-land-cover-database-class-legend-and-description). This is probably out of scope for this particular PR as this case isn't support by smp's JaccardLoss anyway.

isaaccorley:
I think it's just missing a test when ignore_index is not None

torchgeo/trainers/segmentation.py
robmarkcole: If `ignore_index` is a large value like 255, on calling `class_set.remove(ignore_index)` a KeyError will be raised. Overall I think assuming continuous values will cause some issues and I think it would be better to wait for correct ignore_index handling in the loss. In my own code for now: 
```python
        if ignore_index:
            raise Exception("JaccardLoss does not observe ignore_index and one has been configured.")
        return smp.losses.JaccardLoss(mode="multiclass")
```

DimitrisMantas: We could use `discard` instead of `remove` to handle the case where `ignore_index` is not included in `range(self.hparams["num_classes"])`. But I agree that this does not the underlying issue of non-contiguous class labels...

Another way would be to go with the [`smp.FocalLoss`](https://github.com/qubvel/segmentation_models.pytorch/blob/6db76a1106426ac5b55f39fba68168f3bccae7f8/segmentation_models_pytorch/losses/focal.py#L78) approach:

```python
loss = 0

# Filter anchors with -1 label from loss computation
if self.ignore_index is not None:
	not_ignored = y_true != self.ignore_index

for cls in range(num_classes):
	cls_y_true = (y_true == cls).long()
	cls_y_pred = y_pred[:, cls, ...]

	if self.ignore_index is not None:
		cls_y_true = cls_y_true[not_ignored]
		cls_y_pred = cls_y_pred[not_ignored]

	loss += self.focal_loss_fn(cls_y_pred, cls_y_true)
```

However, there is a slight difference between this and the proposed approach. Namely, I believe `class_set` is effectively an index to the output channels, whereas `not_ignored` specifies certain target classes to ignore. Here is a relevant [test](https://gist.github.com/DimitrisMantas/658e16bf9c651588474550a1ad5d76ca).

calebrob6:
Okay this is working up to the point that JaccardLoss supports.

With CrossEntropyLoss you can do something like this:
```
y_pred = torch.rand(1, 4, 256, 256)
y_pred = nn.functional.softmax(y_pred, dim=1)

y_true = torch.randint(0, 4, size=(1, 256, 256))
y_true[0,0,0] = 255

loss =nn.CrossEntropyLoss(ignore_index=255)
loss(y_pred, y_true)
```
and it will work as expected (although it will error if you _don't_ ignore index here and it will error if you have values like 254 and 255 as you can only ignore one of them).

This is not supported by `JaccardLoss` as it assumes the class values range from 0 to C-1 where C is the number of classes.

I would recommend we merge this, then open a new issue that combines some of the points from above and from Robin's issue:
- Users might want to ignore multiple classes
- Users might want to ignore a single non-contiguous class value with JaccardLoss
- Users might want to use masks that have non-contiguous class values (although I think just remapping to [0, C-1] should be on the users)

adamjstewart:
> Users might want to ignore multiple classes

Why not use `classes` instead of `ignore_index` for all loss functions? Seems like they all support it. This would offer maximum flexibility.

> Users might want to ignore a single non-contiguous class value with JaccardLoss
> Users might want to use masks that have non-contiguous class values 

This isn't really how softmax works. If you want to be able to predict a class like 255, you have to have 256+ output classes.

> (although I think just remapping to [0, C-1] should be on the users)

I think this is the answer to the contiguous problem. ML simply doesn't support non-contiguous output classes. Our NLCD/CDL datasets already map to contiguous values.

calebrob6:
To do this with CrossEntropyLoss you would have to set the class weights to the ignored classes to 0 (https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html). This seems excessive.

adamjstewart:
Unless we use https://smp.readthedocs.io/en/latest/losses.html#softcrossentropyloss instead.

DimitrisMantas:
Would it perhaps be an idea to create our own versions of these losses in the future using torchmetrics?

robmarkcole:
Note that losses need to be differetiable, and not all torchmetrics are. When I asked the maintainers about this specifically they did not recommend it

torchgeo/trainers/segmentation.py
adamjstewart: Sets scare me (unsorted, non-deterministic) and the type hints require a list. Let's use:
```suggestion
            classes = list(range(self.hparams["num_classes"]))
```

adamjstewart:
Why aren't we using torchmetrics? Do they not support semantic segmentation?

isaaccorley:
I'll be updating the losses in torchseg to support both ignore_index and classes. @adamjstewart You may want to take a look at the library so we can move over to it sooner rather than later.

torchgeo/trainers/segmentation.py
```diff
@@ -111,8 +104,13 @@ def configure_losses(self) -> None:
                 ignore_index=ignore_value, weight=self.hparams["class_weights"]
             )
         elif loss == "jaccard":
+            # JaccardLoss requires a list of classes to use instead of a class index to ignore.
+            class_set = [
```
adamjstewart, line 108:
```suggestion
            classes = [
```
Not a set. I actually liked your previous solution better because it didn't involve any for-loops in Python.

calebrob6: You gotta pick one

adamjstewart (approved):
Wondering if we should backport this to 0.5.2. Although we were passing an int instead of list[int], it seems like it was still working somehow? So it's just a new feature, not a bug fix?

torchgeo/trainers/segmentation.py
adamjstewart: I picked the previous solution, just with a list instead of a set 😆

adamjstewart (approved):

PR #1902, opened 2024-02-23
NCCM: Add new data module

shreyakannan1205:
The purpose of this PR is to add NCCM's data module


torchgeo/trainers/segmentation.py
```diff
@@ -254,6 +254,7 @@ def validation_step(
             batch["prediction"] = y_hat.argmax(dim=1)
             for key in ["image", "mask", "prediction"]:
                 batch[key] = batch[key].cpu()
+
```
yichiac, line 257: This change seems irrelevant

torchgeo/samplers/batch.py
```diff
@@ -122,6 +122,7 @@ def __init__(
                     self.length += 1
                 self.hits.append(hit)
                 areas.append(bounds.area)
+
```
yichiac, line 125: This change seems irrelevant

yichiac (changes requested):
Are these irrelevant changes?

torchgeo/datamodules/nccm.py
```diff
@@ -0,0 +1,120 @@
...
+from ..samplers import GridGeoSampler, RandomBatchGeoSampler
+from ..samplers.utils import _to_tuple
+from ..transforms import AugmentationSequential
+from .geo import GeoDataModule
+
+
+class NCCMSentinel2DataModule(GeoDataModule):
+    """LightningDataModule implementation for the NCCM and Sentinel2 datasets.
+
+    Uses the train/val/test splits from the dataset.
```
adamjstewart, line 23:
```suggestion
    Uses the train/val/test splits from the dataset.

```

torchgeo/datamodules/nccm.py
```diff
@@ -0,0 +1,120 @@
...
+        self.val_aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
+            K.CenterCrop(self.patch_size),
+            data_keys=["image", "mask"],
+        )
+        self.test_aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
+            K.CenterCrop(self.patch_size),
+            data_keys=["image", "mask"],
+        )
```
adamjstewart, line 79:
```suggestion
        self.aug = AugmentationSequential(
            K.Normalize(mean=self.mean, std=self.std),
            K.CenterCrop(self.patch_size),
            data_keys=["image", "mask"],
        )
```

torchgeo/samplers/batch.py
adamjstewart, line 125:
```suggestion
```

torchgeo/trainers/segmentation.py
adamjstewart, line 257:
```suggestion
```

adamjstewart:
Now that #1878 is merged, the data module correctly detects that NCCM and Sentinel-2 don't actually overlap (specifically in time). You have two options to get the tests to pass:

1. Change the timestamp of the NCCM or Sentinel-2 fake data
2. Add more NCCM or Sentinel-2 fake data at different times

All of these options are equally valid I think.

adamjstewart (dismissed):
Code is fine, just need to get the tests to pass now.

torchgeo/datamodules/nccm.py
```diff
@@ -0,0 +1,116 @@
...
+        Args:
+            stage: Either 'fit', 'validate', 'test', or 'predict'.
+        """
+        self.sentinel2 = Sentinel2(**self.sentinel2_kwargs)
+        self.nccm = NCCM(**self.nccm_kwargs)
+        generator = torch.Generator().manual_seed(0)
+
+        self.dataset = self.sentinel2 & self.nccm
+        (self.train_dataset, self.val_dataset, self.test_dataset) = (
+            random_grid_cell_assignment(self.dataset, [0.8, 0.1, 0.1], 2, generator)
```
adamjstewart, line 89: 2 seems great for testing, but I'm not sure if it seems great for actual usage... Are you sure this is what you want? For example, with CDL, this would mean we're splitting the US into NE, SE, SW, NW quadrants and randomly assigning them to train/val/test.

yichiac: `grid_size` > 2 seems more reasonable for splitting in the actual case. I think we have three options :
1. Set a parameter for  `grid_size` so that users can define it themselves.
2. Create larger test data patches so that our pre-defined `grid_size`, say `grid_size`=8, can be executed.
3. Do both 1 and 2.

yichiac:
> Now that #1878 is merged, the data module correctly detects that NCCM and Sentinel-2 don't actually overlap (specifically in time). You have two options to get the tests to pass:
> 
> 1. Change the timestamp of the NCCM or Sentinel-2 fake data
> 2. Add more NCCM or Sentinel-2 fake data at different times
> 
> All of these options are equally valid I think.

@shreyakannan1205 The Sentinel-2 test data is 2022, and the NCCM data only has 2017-2019. This is the reason why the test error shows `no spatiotemporal intersection`. I checked that those NCCM and Sentinel-2 test data are spatially intersected, so we only need to fix the temporal intersection. As @adamjstewart suggested, you can decide which option to get the tests passed.

shreyakannan1205:
Hi @adamjstewart @yichiac  , I added the 2022 fake data to NCCM. Locally, my test cases pass, but when I push it to git, it is throwing an error. I tried making a few changes, but I am not too sure on what the issue is. It would be really helpful if you could look into it and let me know on how to fix it. Thanks!

adamjstewart:
The error message you're seeing means that the datasets have no area of intersection.

Why does it pass locally but not in CI? Because #1878 has been merged into main but not into your branch. Before running tests in CI, git will merge your PR into the main branch to make sure that your changes don't break main.

Easiest way to test locally would be to rebase your branch or merge in upstream main. You'll note that this PR contains 111 commits. This is because it's also including all of your commits from your NCCM PR which were done on the main branch. I would highly recommend overwriting your main branch with the upstream main branch to avoid this.

PR #1903, opened 2024-02-23
Add DOFA model

ShadowXZT:
Adding support for the DOFA, a unified foundation model for Remote sensing data


ShadowXZT:
@microsoft-github-policy-service agree company="TUM"

microsoft-github-policy-service[bot] ***@***.***>
于2024年2月23日周五 14:58写道：

ShadowXZT:
@microsoft-github-policy-service agree company="TUM"

torchgeo/models/dofa.py
```diff
@@ -0,0 +1,541 @@
...
+            d_model=input_dim,
+            nhead=num_heads,
+            activation="gelu",
+            norm_first=False,
+            batch_first=False,
+            dropout=False,
+        )
+        self.transformer_encoder = nn.TransformerEncoder(
+            encoder_layer, num_layers=num_layers, enable_nested_tensor=False
+        )  # type: ignore[no-untyped-call]
```
adamjstewart, line 87: Missing type hints, added in https://github.com/pytorch/pytorch/pull/120550

torchgeo/models/dofa.py
```diff
@@ -0,0 +1,541 @@
...
+    Returns:
+        A DOFA small 16 model.
+    """
+    model = OFAViT(
+        patch_size=16,
+        embed_dim=384,
+        depth=12,
+        num_heads=6,
+        mlp_ratio=4,
+        norm_layer=partial(nn.LayerNorm, eps=1e-6),  # type: ignore[arg-type]
```
adamjstewart, line 446: mypy doesn't yet support partials: https://github.com/python/mypy/issues/1484

torchgeo/models/dofa.py
```diff
@@ -0,0 +1,544 @@
...
+        """
+        y = self.w1(x)
+        y = self.nonlin1(y)
+        y = self.w2(y)
+        y = self.nonlin2(y)
+        out: Tensor = x + y
+        return out
+
+
+class Dynamic_MLP_OFA(nn.Module):
```
adamjstewart, line 148: I would like to rename this without the underscores. Maybe `DOFAMLP`?

torchgeo/models/dofa.py
```diff
@@ -0,0 +1,544 @@
...
+        self,
+        wv_planes: int,
+        inter_dim: int = 128,
+        kernel_size: int = 3,
+        embed_dim: int = 1024,
+    ) -> None:
+        """Initialize a new Dynamic_MLP_OFA instance.
+
+        Args:
+            wv_planes: Dimensions of dynamic weight generator.
```
adamjstewart, line 161: No idea what this is

torchgeo/models/dofa.py
```diff
@@ -0,0 +1,544 @@
...
+        self.weight_generator = TransformerWeightGenerator(
+            wv_planes, self._num_kernel, embed_dim
+        )
+        self.scaler = 0.01
+
+        self.fclayer = FCResLayer(wv_planes)
+
+        self._init_weights()
+
+    def _get_weights(self, waves: Tensor) -> Tensor:
```
adamjstewart, line 184: Probably going to get rid of this helper function, it doesn't do much and is only 1 line of code

torchgeo/models/dofa.py
```diff
@@ -0,0 +1,544 @@
...
+            img_feat, weights, bias=bias, stride=self.kernel_size, padding=1, dilation=1
+        )
+
+        x = dynamic_out
+        x = x.flatten(2).transpose(1, 2)
+
+        return x, waves
+
+
+class OFAViT(nn.Module):
```
adamjstewart, line 246: Should we rename this to DOFA?

torchgeo/models/dofa.py
```diff
@@ -0,0 +1,544 @@
...
+        self.head = (
+            nn.Linear(embed_dim, num_classes) if num_classes > 0 else nn.Identity()
+        )
+
+    def forward_features(self, x: Tensor, wave_list: list[float]) -> Tensor:
+        """Forward pass of the feature embedding layer.
+
+        Args:
+            x: Input mini-batch.
+            wave_list: List of wavelengths.
```
adamjstewart, line 335: Can we rename this to `wavelengths`? I think it's more clear. The type hints already make it clear that it's a list.

torchgeo/models/dofa.py
ShadowXZT: yes, DOFA sounds better

torchgeo/models/dofa.py
ShadowXZT: ok

tests/models/test_dofa.py
```diff
@@ -0,0 +1,130 @@
...
+        "wavelengths",
+        [
+            # Gaofen
+            [0.443, 0.565, 0.763, 0.765, 0.910],
+            # NAIP
+            [0.640, 0.560, 0.480],
+            [0.480, 0.560, 0.640, 0.810],
+            # Sentinel-1
+            [5.405],
+            [5.405, 5.405],
```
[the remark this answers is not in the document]
ShadowXZT: Should we add a comment here: The central wavelength should be in \mu m unit?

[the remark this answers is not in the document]
adamjstewart: Technically it only has be to in μm to use the pre-trained weights, you could use any units with the model itself, correct?

torchgeo/models/dofa.py
adamjstewart: Which sounds more accurate, DynamicMLPEncoder or DynamicMLPEmbedding?

torchgeo/models/dofa.py
adamjstewart: So it's the number of bands for the satellite?

tests/models/test_dofa.py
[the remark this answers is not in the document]
adamjstewart: Exact numbers don't matter for unit tests. I don't really like that we randomly chose to use 3.75, I think we should change that. I added units to the function docstring.

adamjstewart:
https://github.com/pytorch/pytorch_sphinx_theme/pull/192 will fix the table display. If it doesn't get merged, I'll add the hack locally

PR #1908, opened 2024-02-25
Implement deterministic GeoDataset

### LINKED ISSUE #1899: `random_bbox_assignment` non-deterministic behavior
### Description

The `random_bbox_assignment` dataset splitter (as well as all other spatial index-based splitters I suppose) is not deterministic between different program executions.

That is, splitting a dataset multiple times during the same runtime (e.g., by calling `GeoDataModule.setup()` multiple times in a row, always results in the same splits, which is expected. However, these splits are different when restarting the current kernel/program/script.

### Steps to reproduce

1. Go to the random `random_bbox_assignment` function definition and add the following line under `hits = list(dataset.index.intersection(dataset.index.bounds, objects=True))`: `x = [hit.object for hit in hits]`.
2. Insert a breakpoint in the following line (i.e., where `hits` is permuted).
3. Initialize the `L7IrishDataModule`.
4. Call `setup("fit")` on the data module and while debugging.
5. Catch the breakpoint.
6. Save the value of `x` somewhere.
7. Terminate the program.
8. Do Steps 3-6 again.
9. Notice that the value of `x` has changed.

### Version

0.5.1

DimitrisMantas:
Fixes #1899.

### DIFF
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -76,6 +76,15 @@ class TestGeoDataset:
     def dataset(self) -> GeoDataset:
         return CustomGeoDataset()
 
+    @pytest.fixture(scope="class")
+    def files(self) -> list[str]:
+        """Example list of files in the expected output order.
+
+        Used to test the titular property.
+        """
+        files = ["file://file1.tif", "file://file2.tif", "file://file3.tif"]
+        return files
+
     def test_getitem(self, dataset: GeoDataset) -> None:
         query = BoundingBox(0, 1, 2, 3, 4, 5)
         assert dataset[query] == {"index": query}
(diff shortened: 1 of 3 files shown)


DimitrisMantas:

DimitrisMantas:
OK, looking at this again about an hour later, I'd probably remove the implicit requirement for the file list to be ordered since the main thing we need to preserve is consistency.

So, maybe the order test can go, and so can the related fixture.

DimitrisMantas:
On an unrelated note, `black` took way too long to run on my machine. Perhaps we should consider migrating to [ruff](https://github.com/astral-sh/ruff) at some point?

tests/datasets/test_geo.py
```diff
@@ -76,6 +76,15 @@ class TestGeoDataset:
     def dataset(self) -> GeoDataset:
         return CustomGeoDataset()
 
+    @pytest.fixture(scope="class")
+    def files(self) -> list[str]:
```
adamjstewart, line 80: This definitely feels like overkill for what could just be a class variable.

DimitrisMantas: It's currently used in only one test, so maybe just move it there even?

adamjstewart (approved):
Looks perfect, thanks again for catching this bug and fixing it!

PR #1914, opened 2024-02-27
Document default trainer metrics

### LINKED ISSUE #1874: Document significance of macro vs micro averaging
### Issue

In the segmentation trainer, MulticlassAccuracy MulticlassJaccardIndex uses ` average="micro"` rather than the torchmetrics default macro. 

From slack: Macro weights per patch (what is the average accuracy of each patch?), whereas micro weights per pixel (what is the average accuracy of each pixel across the entire dataset?). These two metrics very greatly when the number of nodata pixels varies between patches. If one patch is mostly nodata pixels, and another patch is mostly data pixels, then macro will take the average of the two patches, while micro will sum up all pixels

Macro might be okay for some applications (such as pre-chipped datasets) but is definitely not what you want for others (such as geospatial datasets)

Further discussion: proposal: add both macro and micro and just name them appropriately OA vs AA - Overall (micro) Accuracy (OA). Average (macro) Accuracy (AA)

### Fix

Refine the description above and document at https://torchgeo.readthedocs.io/en/stable/api/trainers.html#torchgeo.trainers.SemanticSegmentationTask.configure_metrics

jdilger:
Expanded description of averaging reduction method for Segmentation metrics. Let me know if you'd like me to change anything, or if I should add the details on macro accuracy. 

Closes #1874

### DIFF
--- a/torchgeo/trainers/classification.py
+++ b/torchgeo/trainers/classification.py
@@ -96,7 +96,23 @@ class ClassificationTask(BaseTask):
             raise ValueError(f"Loss type '{loss}' is not valid.")
 
     def configure_metrics(self) -> None:
-        """Initialize the performance metrics."""
+        """Initialize the performance metrics.
+
+        * Multiclass Overall Accuracy: Ratio of correctly classified pixels. Uses
+          'micro' averaging.
+        * Multiclass Average Accuracy: Ratio of correctly classified classes. Uses
+          'macro' averaging.
+        * Multiclass Jaccard Index (IoU): Per-class overlap between predicted and
+          actual classes. Uses 'macro' averaging.
+        * Multiclass F1 Score: The harmonic mean of precision and recall. Uses
+          'micro' averaging.
+
+        .. note::
+            * 'Micro' averaging suits overall performance evaluation but may not reflect
+              minority class accuracy.
+            * 'Macro' averaging, gives equal weight to each class, useful for
+              balanced performance assessment across imbalanced classes.
+        """
         metrics = MetricCollection(
             {
                 "OverallAccuracy": MulticlassAccuracy(
@@ -252,7 +268,21 @@ class MultiLabelClassificationTask(ClassificationTask):
     """Multi-label image classification."""
 
     def configure_metrics(self) -> None:
-        """Initialize the performance metrics."""
+        """Initialize the performance metrics.
+
+        * Multiclass Overall Accuracy: Ratio of correctly classified pixels. Uses
+          'micro' averaging.
+        * Multiclass Average Accuracy: Ratio of correctly classified classes. Uses
+          'macro' averaging.
+        * Multiclass F1 Score: The harmonic mean of precision and recall. Uses
+          'micro' averaging.
+
+        .. note::
+            * 'Micro' averaging suits overall performance evaluation but may not
+            reflect minority class accuracy.
+            * 'Macro' averaging, gives equal weight to each class, useful for
+              balanced performance assessment across imbalanced classes.
+        """
         metrics = MetricCollection(
             {
                 "OverallAccuracy": MultilabelAccuracy(
--- a/torchgeo/trainers/classification.py
+++ b/torchgeo/trainers/classification.py
@@ -279,7 +279,7 @@ class MultiLabelClassificationTask(ClassificationTask):
 
         .. note::
             * 'Micro' averaging suits overall performance evaluation but may not
-            reflect minority class accuracy.
+              reflect minority class accuracy.
             * 'Macro' averaging, gives equal weight to each class, useful for
               balanced performance assessment across imbalanced classes.
         """
--- a/torchgeo/trainers/classification.py
+++ b/torchgeo/trainers/classification.py
@@ -98,20 +98,20 @@ class ClassificationTask(BaseTask):
     def configure_metrics(self) -> None:
         """Initialize the performance metrics.
 
-        * Multiclass Overall Accuracy: Ratio of correctly classified pixels. Uses
-          'micro' averaging. Higher values are better.
-        * Multiclass Average Accuracy: Ratio of correctly classified classes. Uses
-          'macro' averaging. Higher values are better.
+        * Multiclass Overall Accuracy (OA): Ratio of correctly classified pixels.
+          Uses 'micro' averaging. Higher values are better.
+        * Multiclass Average Accuracy (AA): Ratio of correctly classified classes.
+          Uses 'macro' averaging. Higher values are better.
         * Multiclass Jaccard Index (IoU): Per-class overlap between predicted and
           actual classes. Uses 'macro' averaging. Higher valuers are better.
-        * Multiclass F1 Score: The harmonic mean of precision and recall. Uses
-          'micro' averaging. Higher values are better.
+        * Multiclass F1 Score: The harmonic mean of precision and recall.
+          Uses 'micro' averaging. Higher values are better.
 
         .. note::
-            * 'Micro' averaging suits overall performance evaluation but may not reflect
-              minority class accuracy.
-            * 'Macro' averaging gives equal weight to each class, and is useful for
-              balanced performance assessment across imbalanced classes.
+           * 'Micro' averaging suits overall performance evaluation but may not reflect
+             minority class accuracy.
+           * 'Macro' averaging gives equal weight to each class, and is useful for
+             balanced performance assessment across imbalanced classes.
         """
         metrics = MetricCollection(
             {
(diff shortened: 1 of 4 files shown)


adamjstewart (commented):
Looks fantastic, maybe @robmarkcole can review?

Should we do this for all of our LightningModules? Basically just document the default metrics we use.

robmarkcole:
I feel this is a particularly nuanced issue so have expanded slightly:
```
"""
Sets up segmentation accuracy metrics:

- Multiclass Pixel Accuracy: Ratio of correctly classified pixels.
- Multiclass Jaccard Index (IoU): Per-pixel overlap between predicted and
  actual segments.

Uses 'micro' averaging, aggregating pixel predictions across classes. Each
pixel is weighted equally, favoring classes with more pixels. This may skew
performance metrics towards majority classes due to class imbalance.

Note:
- 'Micro' averaging suits overall performance evaluation but may not reflect
  minority class accuracy.
- 'Macro' averaging, not used here, gives equal weight to each class, useful
  for balanced performance assessment across imbalanced classes.
"""
```

adamjstewart:
@jdilger do you think you can produce a similar level of detailed descriptions for the accuracy metrics used in other trainers? If not maybe @robmarkcole and I can help.

jdilger:
Thanks @robmarkcole,  the more nuanced version looks great.  @adamjstewart Sure thing. I can take a first pass at it.

adamjstewart:
We're likely going to release v0.5.2 tomorrow or Saturday, so if you can add these today that would be great. If not, we may merge this anyway and then we can add more details for more trainers in a follow-up PR. Thanks again for your help!

robmarkcole:
LGTM. Only a minor suggestion for consistency

Sets up segmentation accuracy metrics >> Initialize the metrics

jdilger:
I have time to work on this today, but I went ahead and pushed the suggested change from robmarkcole. 

I noticed that the metrics setup for the classification and regression trainers are set up with dicts while segmentation and detection use lists. I'm not sure it needs to be addressed now, but defining some standard names might be helpful. For instance, in the segmentation trainer it's using `[MulticlassAccuracy(..., average="micro",)]`  but in the classification trainer it's setup as `{'OverallAccuracy':MulticlassAccuracy(..., average="micro",)}`

adamjstewart:
There are some formatting issues with lists that are causing the documentation to fail to build. We should be using the https://sublime-and-sphinx-guide.readthedocs.io/en/latest/lists.html#unordered-lists format and indenting lines that stretch across multiple lines. Also, we could be using a note: https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html.

adamjstewart:
Also I'm happy to handle the formatting stuff myself if you want to focus on adding _something_ to the other trainers. For better or for worse, I know most of the quirks of Sphinx/rST now.

jdilger:
@adamjstewart I think I figured it out, but do double-check  😄 . If there are still issues I'm happy to let you take it over. I don't have too much experience with Sphinx.

Side note: In `contributing.rst` to install the docs is written as `pip install .[docs]` which didn't work for me. I had to use `pip install '.[docs]'`. I don't know if it's a typo or related to the OS I'm using (Mac).

adamjstewart:
Hmm, quotes shouldn't be required for pip. What error message did you get? What shell are you using, bash?

jdilger:
Ah, you're  100% right. I'm using zsh which gave the error `zsh: no matches found: .[docs]`

torchgeo/trainers/segmentation.py
```diff
@@ -122,7 +122,22 @@ def configure_losses(self) -> None:
             )
 
     def configure_metrics(self) -> None:
-        """Initialize the performance metrics."""
+        """Initialize the performance metrics.
+
+        * Multiclass Pixel Accuracy: Ratio of correctly classified pixels.
+        * Multiclass Jaccard Index (IoU): Per-pixel overlap between predicted and
+          actual segments.
```
adamjstewart, line 129: Might also be worth documenting that higher is better. This isn't true for all of our metrics.

adamjstewart (commented):
Formatting looks good now. If we can do this for all other trainers by tomorrow, great. If not, we'll still merge and add more details after the release.

torchgeo/trainers/segmentation.py
jdilger: Yeah, that would be helpful for quick reference and those unfamiliar with the metrics. Maybe something like `{Higher/Lower} values indicate better agreement.`?

jdilger:
Added a first pass for the other configuration metrics. The formatting should be correct but please double-check the content :).

torchgeo/trainers/classification.py
```diff
@@ -96,7 +96,23 @@ def configure_losses(self) -> None:
...
+          'macro' averaging.
+        * Multiclass Jaccard Index (IoU): Per-class overlap between predicted and
+          actual classes. Uses 'macro' averaging.
+        * Multiclass F1 Score: The harmonic mean of precision and recall. Uses
+          'micro' averaging.
+
+        .. note::
+            * 'Micro' averaging suits overall performance evaluation but may not reflect
+              minority class accuracy.
+            * 'Macro' averaging, gives equal weight to each class, useful for
```
adamjstewart, line 113:
```suggestion
            * 'Macro' averaging gives equal weight to each class, and is useful for
```

torchgeo/trainers/segmentation.py
adamjstewart: I would be fine with that or even something as simple as `{Higher/Lower} values are better.`

jdilger: Sounds good. I'll add that in.

torchgeo/trainers/segmentation.py
```diff
@@ -122,7 +122,23 @@ def configure_losses(self) -> None:
...
+        """Initialize the performance metrics.
+
+        * Multiclass Pixel Accuracy: Ratio of correctly classified pixels. Uses
+          'micro' averaging. Higher values are better.
+        * Multiclass Jaccard Index (IoU): Per-pixel overlap between predicted and
+          actual segments.Uses 'macro' averaging. Higher values are better.
+
+        Uses 'micro' averaging, aggregating pixel predictions across classes. Each
+        pixel is weighted equally, favoring classes with more pixels. This may skew
+        performance metrics towards majority classes due to class imbalance.
```
adamjstewart, line 134: We don't have this for any other trainer and it's kinda covered in the note below, maybe we can remove it?

adamjstewart:
> Ah, you're 100% right. I'm using zsh which gave the error `zsh: no matches found: .[docs]`

It looks like zsh uses square brackets for pattern matching. Feel free to open a PR to add quotes around all use of square brackets in Installation and Contributing.

torchgeo/trainers/segmentation.py
jdilger: Agreed, I think the note below + the updated description covers it all. I'm afk for the day but feel free to remove it if you're inclined.

adamjstewart (approved):
@robmarkcole last chance to review

robmarkcole:
For the classification metrics, are they per pixel or per image?

adamjstewart:
Classification is per-image. I'm not actually sure if micro and macro are different under that context.

robmarkcole:
I think we expect them to be different for imbalanced datasets. The metrics will not compute at the pixel level but at the image level - the terms total true positives, false negatives, and false positives should be used for classification

torchgeo/trainers/regression.py
```diff
@@ -97,7 +97,21 @@ def configure_losses(self) -> None:
             )
 
     def configure_metrics(self) -> None:
-        """Initialize the performance metrics."""
+        """Initialize the performance metrics.
+
+        * Root MeanSquared Error: The squared root of the mean squared error. Lower
```
robmarkcole, line 102: could be clearer: The square root of the average of the squared differences between the predicted and actual values

torchgeo/trainers/regression.py
```diff
@@ -97,7 +97,21 @@ def configure_losses(self) -> None:
...
+          values are better.
+        * Mean Squared Error: The ratio of the sum of the average of squared errors to
```
robmarkcole, line 104: The average of the squared differences between the predicted and actual values

torchgeo/trainers/regression.py
```diff
@@ -97,7 +97,21 @@ def configure_losses(self) -> None:
...
+          the number of samples. Lower values are better.
+        * Mean Absolute Error: The ratio of the sum of the absolute value between
```
robmarkcole, line 106: The average of the absolute differences between the predicted and actual values

adamjstewart (approved):

adamjstewart:
Want to squeeze this in to the next release, but feel free to open another PR to update these docs or change the default metrics we use after the release!

PR #1915, opened 2024-02-28
add $azcopy stubs for radiantEarth repos

darkblue-b:


adamjstewart:
Related to #1830 and #1887. We spoke on Slack, I think you should just take over #1887 and finish it up. We would love to get this into the 0.5.2 release expected later this week, so I'm willing to help both of you finish that PR.

darkblue-b:
ref to https://github.com/microsoft/torchgeo/issues/1830

yichiac (changes requested):
Could you add $azcopy download support for [AgriFieldNet](https://github.com/microsoft/torchgeo/blob/main/torchgeo/datasets/agrifieldnet.py) and [South America Crop Type](https://github.com/microsoft/torchgeo/blob/main/torchgeo/datasets/south_africa_crop_type.py) datasets? Those datasets also require download from Source Cooperative (Radiant Earth).

PR #1916, opened 2024-02-28
HuggingFace: use stable download URLs

adamjstewart:
See https://discuss.huggingface.co/t/permalink-for-repositories/42691 for motivation.

Replacing huggingface.co with hf.co is entirely optional, just wanted to make the URL a little shorter.

Tested the model downloads with:
```console
> pytest -m slow tests/models/
```
Didn't test the dataset downloads.

### DIFF
--- a/torchgeo/datasets/chabud.py
+++ b/torchgeo/datasets/chabud.py
@@ -68,7 +68,7 @@ class ChaBuD(NonGeoDataset):
     ]
     rgb_bands = ["B04", "B03", "B02"]
     folds = {"train": [1, 2, 3, 4], "val": [0]}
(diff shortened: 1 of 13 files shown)


nilsleh (approved):

isaaccorley:
ChaBuD and models/swin.py didn't make it into 0.5.2 because they didn't exist yet. Will need to make sure these are added in 0.6.0

adamjstewart:
It's merged into main, which will by definition be in 0.6.0. Releases/v0.5 is frozen in time, it is no longer in the main branch. See https://nvie.com/posts/a-successful-git-branching-model/

PR #1918, opened 2024-02-28
xView2: deterministic dataset order

adamjstewart:
This is the only other place I found where we're using a set. @RitwikGupta maybe you can confirm that dataset order is not deterministic (even with setting a random seed) and confirm that this fixes it?


isaaccorley:
I think for this case this isn't as big of an issue because the train/test sets are in separate folders. But if you did random split the train set into a train/val set you would encounter issues with determinism due to the set operation.

isaaccorley (approved):

adamjstewart:
You still have issues with determinism even without random split. The order you iterate over the dataset changes, so you can't reproduce training results.

isaaccorley:
Train dataloaders random shuffle by default on each epoch, but yeah if you want to guarantee that that randomness is identical per seed, this definitely needs to be fixed.

PR #1924, opened 2024-03-03
Docs: fix trainer metric definitions

### LINKED ISSUE #1923: Errors & improvements in Metrics descriptions
### Description

- In segmentation trainer, configure_metrics, Multiclass Jaccard Index (IoU) is described as using macro but it is using micro
- In classification, MultiLabelClassificationTask, OA mentions pixels, but it should refer to image level metrics such as true and false positive
- Regression metrics cannot take macro/micro as args, perhaps we remove the Note to avoid confusion
- Detection, MeanAveragePrecision computes many metrics. Run the example at https://github.com/Lightning-AI/torchmetrics/blob/master/examples/detection_map.py returns

```
0s
result
{'map': tensor(0.6000),
 'map_50': tensor(1.),
 'map_75': tensor(1.),
 'map_small': tensor(-1.),
 'map_medium': tensor(-1.),
 'map_large': tensor(0.6000),
 'mar_1': tensor(0.6000),
 'mar_10': tensor(0.6000),
 'mar_100': tensor(0.6000),
 'mar_small': tensor(-1.),
 'mar_medium': tensor(-1.),
 'mar_large': tensor(0.6000),
 'map_per_class': tensor(-1.),
 'mar_100_per_class': tensor(-1.),
 'classes': tensor(0, dtype=torch.int32)}
```

- Note, outside the scope of this issue, but we do not pass [dist_sync_on_step](https://lightning.ai/docs/torchmetrics/stable/pages/overview.html#advanced-metric-settings) - we could and default to False, but allow setting true for distributed training with Lightning

CC @jdilger

### Steps to reproduce

NA

### Version

dev

adamjstewart:
Closes #1923 

We are nowhere near the limit of how much detail we could have. We can always start adding mathematical equations.

Alternative: why don't we just list the metrics and link to the torchmetrics docs? If a metric isn't properly explained, we can always improve the torchmetrics docs.

@robmarkcole @jdilger

### DIFF
--- a/torchgeo/trainers/detection.py
+++ b/torchgeo/trainers/detection.py
@@ -206,12 +206,11 @@ class ObjectDetectionTask(BaseTask):
     def configure_metrics(self) -> None:
         """Initialize the performance metrics.
 
-        * Mean Average Precision (mAP): Precision is the number of true positives
-          (as defined by intersection over union) divided by the number of true
-          positives + false positives. Uses 'macro' averaging. Higher values are better.
-        * Mean Average Recall (mAR): Recall is the number of true positives
-          (as defined by intersection over union) divived by the number of true
-          positives + false negatives. Uses 'macro' averaging. Higher values are better.
+        * :class:`~torchmetrics.detection.mean_ap.MeanAveragePrecision`: Mean average
+          precision (mAP) and mean average recall (mAR). Precision is the number of
+          true positives divided by the number of true positives + false positives.
+          Recall is the number of true positives divived by the number of true positives
+          + false negatives. Uses 'macro' averaging. Higher values are better.
 
         .. note::
            * 'Micro' averaging suits overall performance evaluation but may not
@@ -219,7 +218,7 @@ class ObjectDetectionTask(BaseTask):
            * 'Macro' averaging gives equal weight to each class, and is useful for
              balanced performance assessment across imbalanced classes.
         """
-        metrics = MetricCollection([MeanAveragePrecision()])
+        metrics = MetricCollection([MeanAveragePrecision(average="macro")])
         self.val_metrics = metrics.clone(prefix="val_")
         self.test_metrics = metrics.clone(prefix="test_")
 
--- a/torchgeo/trainers/detection.py
+++ b/torchgeo/trainers/detection.py
@@ -206,10 +206,12 @@ class ObjectDetectionTask(BaseTask):
     def configure_metrics(self) -> None:
         """Initialize the performance metrics.
 
-        * Mean Average Precision (mAP): Computes the Mean-Average-Precision (mAP) and
-          Mean-Average-Recall (mAR) for object detection. Prediction is based on the
-          intersection over union (IoU) between the predicted bounding boxes and the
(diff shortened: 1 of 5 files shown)


robmarkcole:
I think what should be listed here is perhaps a short description which is specifically summarising the meaning of the metrics in the imagery context - then users can lookup torch metrics for more specifics

adamjstewart:
I'm just afraid that that short description will continue to blow up as people have questions unless we direct them to the torchmetrics docs for details.

torchgeo/trainers/detection.py
```diff
@@ -206,18 +206,19 @@ def configure_models(self) -> None:
...
+          + false negatives. Uses 'macro' averaging. Higher values are better.
 
         .. note::
            * 'Micro' averaging suits overall performance evaluation but may not
              reflect minority class accuracy.
            * 'Macro' averaging gives equal weight to each class, and is useful for
              balanced performance assessment across imbalanced classes.
         """
-        metrics = MetricCollection([MeanAveragePrecision()])
+        metrics = MetricCollection([MeanAveragePrecision(average="macro")])
```
adamjstewart, line 221: This is the default, but it's better to be explicit about this since the default may change and we're documenting it due to confusion.

adamjstewart:
@robmarkcole better or worse? https://torchgeo--1924.org.readthedocs.build/en/1924/api/trainers.html

torchgeo/trainers/detection.py
robmarkcole: I typically see mAP@50 in papers, we are reporting this? The torchmetrics docs state `map score is calculated with @[ IoU=self.iou_thresholds | area=all | max_dets=max_detection_thresholds ].` but I don't know what that value is by default

adamjstewart: I think we are reporting all of the metrics you displayed in #1923, although I don't know the details of what those all mean.

robmarkcole: Yes I'm wondering if its best to return them all and let the user choose which are significant, or return a documented subset. Probably the first is preferable, but adding more docs over time could be helpful

robmarkcole:
The docs rendering looks great, and the links to torchmetrics docs are a nice addition

robmarkcole:
I did just notice we are not passing `ignore_index` to e.g. MulticlassFBetaScore

adamjstewart:
> I did just notice we are not passing `ignore_index` to e.g. MulticlassFBetaScore

The only place we use `MulticlassFBetaScore` is `ClassificationTask`, which doesn't have an `ignore_index` parameter. So I'm confused...

robmarkcole:
I realised that and deleted comment after :)

torchgeo/trainers/detection.py
adamjstewart: I would much rather improve the torchmetrics docs than our docs.

isaaccorley: I agree, I think we should link to the torchmetrics docs instead of copy/pasting their definitions here which seems like just we are just adding boilerplate to our docs when users can go look at all the torchmetrics definitions.

isaaccorley (approved):

PR #1931, opened 2024-03-03
Resolve multiple SimCLR and MoCo lightly warnings

adamjstewart:
This PR consists of 3 commits:

1. Configure the model first
2. Specify the feature dimension of the memory bank
3. Only test one warning at a time

1 is only needed for 2. The alternative would be to make `hidden_dim` and `output_dim` non-optional. This is already the case in MoCo, but not yet the case in SimCLR. I'm open to this change, let me know what you think. I kinda think 1 is the right way to go regardless, but I wonder if we should make them non-optional too just so SimCLR and MoCo match. Can't remember why I did it this way in the first place...

2 is needed to resolve the following warning:
```
lib/python3.12/site-packages/lightly/models/modules/memory_bank.py:88: UserWarning: Memory bank size 'size=10' does not specify feature dimension. It is recommended to set the feature dimension with 'size=(n, dim)' when creating the memory bank. Distributed training might fail if the feature dimension is not set.
```
We might want to consider backporting this for distributed training support.

### DIFF
--- a/torchgeo/trainers/base.py
+++ b/torchgeo/trainers/base.py
@@ -36,19 +36,19 @@ class BaseTask(LightningModule, ABC):
         """
         super().__init__()
         self.save_hyperparameters(ignore=ignore)
+        self.configure_models()
         self.configure_losses()
         self.configure_metrics()
-        self.configure_models()
+
+    @abstractmethod
+    def configure_models(self) -> None:
+        """Initialize the model."""
 
     def configure_losses(self) -> None:
         """Initialize the loss criterion."""
 
     def configure_metrics(self) -> None:
         """Initialize the performance metrics."""
-
-    @abstractmethod
-    def configure_models(self) -> None:
-        """Initialize the model."""
 
     def configure_optimizers(
         self,
--- a/torchgeo/trainers/moco.py
+++ b/torchgeo/trainers/moco.py
@@ -226,14 +226,6 @@ class MoCoTask(BaseTask):
         self.augmentation1 = augmentation1 or aug1
         self.augmentation2 = augmentation2 or aug2
 
-    def configure_losses(self) -> None:
-        """Initialize the loss criterion."""
-        self.criterion = NTXentLoss(
-            self.hparams["temperature"],
-            self.hparams["memory_bank_size"],
-            self.hparams["gather_distributed"],
-        )
-
     def configure_models(self) -> None:
         """Initialize the model."""
         model: str = self.hparams["model"]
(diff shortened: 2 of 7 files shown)


adamjstewart:
Bleh, either need to bump lightly min version or add some if-statements/try-except-statements.

adamjstewart:
We would have to use 1.4.25+, which is < 2 months old. If-statement it is.

torchgeo/trainers/base.py
```diff
@@ -36,20 +36,20 @@ def __init__(self, ignore: Optional[Union[Sequence[str], str]] = None) -> None:
         """
         super().__init__()
         self.save_hyperparameters(ignore=ignore)
+        self.configure_models()
         self.configure_losses()
         self.configure_metrics()
-        self.configure_models()
+
+    @abstractmethod
+    def configure_models(self) -> None:
```
calebrob6, line 44: Why is all this changed?

torchgeo/trainers/moco.py
```diff
@@ -282,6 +274,14 @@ def configure_models(self) -> None:
         # Initialize moving average of output
         self.avg_output_std = 0.0
 
+    def configure_losses(self) -> None:
```
calebrob6, line 277: Why are you moving these around in the file?

calebrob6:
Seems fine, but I'm worried this doesn't actually work. Is there any way to do an integration test of the SSL training?

torchgeo/trainers/base.py
adamjstewart: See the PR description above

adamjstewart:
Which part are you concerned about? There aren't really any changes other than the order of operations. We have unit tests that make sure that the code runs without crashing, but integration tests that assert that the model is actually learning are quite a bit harder.

calebrob6:
My concerns aren't relevant to this PR. Do you know if anyone has actually tried to use the SSL trainers?

torchgeo/trainers/base.py
calebrob6: Yes I read the PR description but that doesn't explain why you are moving the configure_models method around in the file. As far as I can tell that is unnecessary?

torchgeo/trainers/simclr.py
```diff
@@ -176,13 +166,16 @@ def configure_models(self) -> None:
 
         # Create projection head
         input_dim = self.backbone.num_features
-        if hidden_dim is None:
-            hidden_dim = input_dim
-        if output_dim is None:
-            output_dim = input_dim
+        if self.hparams["hidden_dim"] is None:
```
calebrob6, line 169: Just for my understanding, all these changes are so you can do this?

calebrob6:
Did we actually use the SSL trainers (I vaguely recall we didn't because of bugs)?

torchgeo/trainers/simclr.py
adamjstewart: Yes

adamjstewart:
We did

torchgeo/trainers/base.py
calebrob6: Okay got it, seems fine!

calebrob6 (approved):

calebrob6:
Oh great!

PR #1932, opened 2024-03-03
load_state_dict: monkeypatch during testing

adamjstewart:
This was fun. Depending on the order in which these two tests are run, a second warning will be raised because the hack to fc or conv1 remained in effect. If either test was run on its own, there would be no warning. Proper solution is to monkeypatch instead of editing the original object.

### DIFF
--- a/tests/trainers/test_utils.py
+++ b/tests/trainers/test_utils.py
@@ -9,6 +9,7 @@ import pytest
 import timm
 import torch
 import torch.nn as nn
+from pytest import MonkeyPatch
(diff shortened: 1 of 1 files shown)


isaaccorley:
At some point I'm going to refactor this because timm has a `reset_classifier` method which can be used to automatically change whatever the head is. This is also a better solution because not all timm models have a `conv1` and `fc` attribute.

adamjstewart:
Can we merge in the meantime?

isaaccorley (approved):

PR #1933, opened 2024-03-04
BaseTask: explicitly pass batch size to logger

adamjstewart:
Follow-up to #1928. After we squashed those warning messages, new ones popped up in other trainers.

I think I better understand the issue now. The logger would like to know the batch size. Normally, it infers this from the output of the data loader. However, we output a dictionary instead of a list of tensors, so it doesn't know how to infer this. The solution is to explicitly tell it what our batch size is.

### DIFF
--- a/torchgeo/trainers/classification.py
+++ b/torchgeo/trainers/classification.py
@@ -178,11 +178,12 @@ class ClassificationTask(BaseTask):
         """
         x = batch["image"]
         y = batch["label"]
+        batch_size = x.shape[0]
         y_hat = self(x)
         loss: Tensor = self.criterion(y_hat, y)
-        self.log("train_loss", loss)
+        self.log("train_loss", loss, batch_size=batch_size)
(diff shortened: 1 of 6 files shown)


torchgeo/trainers/classification.py
```diff
@@ -338,12 +342,13 @@ def validation_step(
         """
         x = batch["image"]
         y = batch["label"]
+        batch_size = x.shape[0]
```
nilsleh, line 345: do we have to store batch size as a variable, or can it just be directly in the `self.log` call?

adamjstewart: It can be directly used, but we're using it in multiple places and it's a little more clear to give it a name

nilsleh: ah yeah true, I had only looked at the first BYOL task which only uses it once.

nilsleh (approved):

PR #1959, opened 2024-03-25
Add South America Soybean DataModule

cookie-kyu:


torchgeo/datamodules/sentinel2_south_america_soybean.py
```diff
@@ -0,0 +1,108 @@
...
+            SouthAmericaSoybean,
+            batch_size=batch_size,
+            patch_size=patch_size,
+            length=length,
+            num_workers=num_workers,
+            **kwargs,
+        )
+
+        self.aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std), data_keys=["image", "mask"]
```
adamjstewart, line 65: Can you also copy `self.train_aug` from the other Sentinel2 data modules?

adamjstewart (changes requested):
One last change and this looks good to go!

tests/data/south_america_soybean/data.py
adamjstewart, line 1: You need to re-run this script and add the changed files to the PR in order for the tests to pass.

adamjstewart (approved):

PR #1965, opened 2024-03-27
Add an initial version of WMSDataset

### LINKED ISSUE #1963: Add a WMS Dataset
### Summary

I've written a currently pretty much untested WMSDataset, do you want to incorporate?

### Rationale

I want to use RGB aerial imagery which I can download but since my local government provides a WMS why should I clutter up my hard drive, and they make downloading large areas very hard.

### Implementation

You can see the code at [this gist](https://gist.github.com/ian-gla/49bd29c30b62a4f83b4e53cd15b1b091) as far as I can tell it works so far.

### Alternatives

_No response_

### Additional information

_No response_

ianturton:
Fixes #1963

### DIFF
--- a/requirements/required.txt
+++ b/requirements/required.txt
@@ -9,6 +9,7 @@ lightly==1.5.2
 lightning[pytorch-extra]==2.2.1
 matplotlib==3.8.3
 numpy==1.26.4
+owsLib==0.30.0
 pandas==2.2.1
 pillow==10.2.0
 pyproj==3.6.1
--- a/requirements/required.txt
+++ b/requirements/required.txt
@@ -9,7 +9,6 @@ lightly==1.5.2
 lightning[pytorch-extra]==2.2.1
 matplotlib==3.8.3
 numpy==1.26.4
-owsLib==0.30.0
 pandas==2.2.1
 pillow==10.2.0
 pyproj==3.6.1
--- a/tests/datasets/test_wms.py
+++ b/tests/datasets/test_wms.py
@@ -2,13 +2,11 @@
 # Licensed under the MIT License.
 
 import pytest
-
-from torchgeo.datasets import (
-    WMSDataset,
-)
-
+from torchgeo.datasets import WMSDataset
 import requests
-SERVICE_URL = 'https://mesonet.agron.iastate.edu/cgi-bin/wms/nexrad/n0r-t.cgi?'
+
+
+SERVICE_URL = "https://mesonet.agron.iastate.edu/cgi-bin/wms/nexrad/n0r-t.cgi?"
 
 
 class TestWMSDataset:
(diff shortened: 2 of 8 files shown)


ianturton:

torchgeo/datasets/wms.py
```diff
@@ -0,0 +1,145 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# Author: Ian Turton, Glasgow University ian.turton@gla.ac.uk
```
adamjstewart, line 3: We don't normally edit the license header. Your contribution will still be included in the git history though.

torchgeo/datasets/wms.py
```diff
@@ -0,0 +1,145 @@
...
+
+"""A simple class to fetch WMS images."""
+from io import BytesIO
+from typing import Any, Callable, Optional, Union
+
+import torchvision.transforms as transforms
+from owslib.map.wms111 import ContentMetadata
+from owslib.wms import WebMapService
```
adamjstewart, line 11: Imports should be done lazily inside of the class so that the rest of TorchGeo still works even if owslib isn't installed. If you grep for any of our other optional dataset dependencies, you should see example code for this you can copy-n-paste.

torchgeo/datasets/wms.py
```diff
@@ -0,0 +1,145 @@
...
+                except CRSError:
+                    pass
+
+    def getlayer(self) -> ContentMetadata:
+        """Return the selected layer object."""
+        return self._layer
+
+    def layers(self) -> list[str]:
+        """Return a list of availiable layers."""
+        return self._layers
```
adamjstewart, line 109: Getters/setters are generally considered to be bad practice in Python, especially if you don't actually need to do anything special. In this case, if we want the ability to access the attribute, let's just make it a public attribute.

torchgeo/datasets/wms.py
```diff
@@ -0,0 +1,145 @@
...
+            srs="epsg:" + str(self.crs.to_epsg()),
+            bbox=(query.minx, query.miny, query.maxx, query.maxy),
+            # TODO fix size
+            size=(500, 500),
+            format=self._format,
+            transparent=True,
+        )
+        sample = {"crs": self.crs, "bbox": query}
+
+        transform = transforms.Compose([transforms.ToTensor()])
```
adamjstewart, line 134: We don't normally use torchvision transforms. What is the output from wms, a numpy array? Let's just convert that directly to PyTorch without relying on PIL

adamjstewart (changes requested):
This looks like a great start! I commented on some of the immediately obvious software-y things. There are still a lot of technical questions (what happens if I change the CRS or res, how does WMS handle that?) but I'll save those for later. Thanks again for the PR!

torchgeo/datasets/wms.py
```diff
@@ -0,0 +1,145 @@
...
+from rasterio.coords import BoundingBox
+from rasterio.crs import CRS
+from rasterio.errors import CRSError
+from rtree.index import Index, Property
+
+from torchgeo.datasets import GeoDataset
+
+
+class WMSDataset(GeoDataset):
+    """Allow models to fetch images from a WMS (at a good resolution)."""
```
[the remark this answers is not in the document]
adamjstewart: Also, can you add:
```rst
.. versionadded:: 0.6
```
to the docstring? This will document what version of TorchGeo is required to use this feature.

ianturton:
I can't work out what's going on with `mypy`:

```
(torchgeo) ➜  torchgeo git:(wms) pwd                                                
/home/iturton/code/torchgeo
(torchgeo) ➜  torchgeo git:(wms) mypy tests/datasets/test_wms.py 
Success: no issues found in 1 source file
(torchgeo) ➜  torchgeo git:(wms) mypy torchgeo/datasets/wms.py 
Success: no issues found in 1 source file
(torchgeo) ➜  torchgeo git:(wms) pre-commit run --all-files
pyupgrade................................................................Passed
isort....................................................................Passed
black....................................................................Passed
flake8...................................................................Passed
pydocstyle...............................................................Passed
- hook id: mypy
- exit code: 1

tests/datasets/test_wms.py:5: error: Library stubs not installed for "requests"  [import-untyped]
tests/datasets/test_wms.py:5: note: Hint: "python3 -m pip install types-requests"
tests/datasets/test_wms.py:5: note: (or run "mypy --install-types" to install all missing stub packages)
tests/datasets/test_wms.py:5: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 313 source files)
```

and yes `types-requests` is installed

ashnair1:
pre-commit creates and uses a separate isolated environment while running hooks and not your existing environment. To solve, this, you just need to add `types-requests` to the `additional_dependencies` under mypy hook in the `.pre-commit-config.yaml`. The pre-commit environment will get updated and the hooks will pass.

tests/datasets/test_wms.py
```diff
@@ -0,0 +1,61 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import pytest
+import requests
```
adamjstewart, line 5: We don't currently have a dependency on requests. Is this needed?

tests/datasets/test_wms.py
```diff
@@ -0,0 +1,61 @@
...
+
+from torchgeo.datasets import WMSDataset
+
+SERVICE_URL = "https://mesonet.agron.iastate.edu/cgi-bin/wms/nexrad/n0r-t.cgi?"
+
+
+def service_ok(url: str, timeout: int = 5) -> bool:
+    try:
+        resp = requests.head(url, allow_redirects=True, timeout=timeout)
+        ok = bool(resp.ok)
```
ashnair1, line 15: The `ok` property of `Response` always returns a bool so casting is not necessary.

tests/datasets/test_wms.py
ianturton: possibly not, let me see as it would solve a bunch of my other issues

requirements/required.txt
```diff
@@ -9,6 +9,7 @@ lightly==1.5.2
 lightning[pytorch-extra]==2.2.1
 matplotlib==3.8.3
 numpy==1.26.4
+owsLib==0.30.0
```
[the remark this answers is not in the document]
adamjstewart: Also need to add to `requirements/min-reqs.old` in order to get the minimum tests to pass.

adamjstewart:
We recently updated the repo to use Python 3.10+. Some of your code uses Python 3.9 syntax. You can fix this with pyupgrade: https://torchgeo.readthedocs.io/en/stable/user/contributing.html#linters

torchgeo/datasets/wms.py
ianturton: It's a JPEG or PNG image can I convert that directly to a tensor?

isaaccorley: T.ToTensor() also automatically divides by 255 which we should leave up to the user to decide. You can convert the image to a tensor using `torch.from_numpy(np.array(Image.open(path).astype(np.float32)))`. You can break this up over multiple lines so it's not too verbose

torchgeo/datasets/wms.py
isaaccorley: > @ianturton you can check `torchgeo.datasets.chabud` as an example. See how we check if h5py is available in the constructor and then import h5py in the method that uses it.

adamjstewart:
@ianturton is this still a work in progress? We're preparing for a new release in the next couple weeks. Trying to merge or close as many remaining PRs as possible.

PR #1966, opened 2024-03-27
Drop Python 3.9 support

adamjstewart:
It's that time of year again!

We've been following [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) for a while now to decide when to drop support for older Python versions. However, NEP 29 is now superseded by [SPEC 0](https://scientific-python.org/specs/spec-0000/). Although NEP 29 adopted a 3.5 year support strategy, SPEC 0 now adopts a shorter 3 year support strategy. Since a new Python minor release happens once a year, this means support for the 3 most recent Python minor versions. However, PyTorch is slow to adopt the latest Python version after it gets released, so I propose waiting until it is fully supported before dropping the oldest version.

The biggest change in Python 3.10 is support for `X | Y` as a replacement for `typing.Union[X, Y]` and `X | None` as a replacement for `Optional[X]`. This means that `typing` no longer needs to be imported by most files! It's also an excuse to bump the minimum supported version for many of our dependencies.

### DIFF
--- a/.github/workflows/style.yaml
+++ b/.github/workflows/style.yaml
@@ -158,7 +158,7 @@ jobs:
     - name: List pip dependencies
       run: pip list
     - name: Run pyupgrade checks
-      run: pyupgrade --py39-plus $(find . -path ./docs/src -prune -o -name "*.py" -print)
+      run: pyupgrade --py310-plus $(find . -path ./docs/src -prune -o -name "*.py" -print)
 concurrency:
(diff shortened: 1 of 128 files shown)


adamjstewart:
Can also use the new match-case statement introduced in [3.10](https://docs.python.org/3/whatsnew/3.10.html) in places like `DataModule.setup`.

isaaccorley (approved):
I double checked each change and looks good

calebrob6 (approved):

PR #477, opened 2022-03-22
RandomGeoSampler: several bug fixes

adamjstewart:
This PR includes the following fixes:

- [x] Sample from tiles with larger areas more often (fixes #408)
- [x] Allow `size=0` queries to handle point data
- [x] Allow `size=(height, width)` queries for entire tile at a time
- [x] Use PyTorch's built-in random sampling for easier seed setting

The first requires #375 so has to be in 0.3.0. The rest could theoretically be backported to 0.2.2 but I'm not really planning on a 0.2.2 release at the moment.

PR #1968, opened 2024-03-28
RandomGeoSampler: fix performance regression

adamjstewart:
Fixes a bug I introduced all the way back in #477. Thanks @yichiac for noticing this!

We want to try to sample from a pixel-aligned grid whenever possible to avoid resampling in the average case. With the changes in this PR, we should revert back to the behavior we documented in our paper, where preprocessing greatly improves sampling performance. GridGeoSampler was not affected by this bug.

I still want to create an I/O benchmarking script similar to what we used in our original paper to ensure that this indeed works as advertised, so will keep as a draft for now.


torchgeo/samplers/utils.py
```diff
@@ -55,17 +55,16 @@ def get_random_bounding_box(
     """
     t_size = _to_tuple(size)
 
-    width = (bounds.maxx - bounds.minx - t_size[1]) // res
-    height = (bounds.maxy - bounds.miny - t_size[0]) // res
+    # May be negative if bounding box is smaller than patch size
+    width = (bounds.maxx - bounds.minx - t_size[1]) / res
+    height = (bounds.maxy - bounds.miny - t_size[0]) / res
```
adamjstewart, line 60: This is the width/height in pixel units. It no longer needs to be an integer, float is fine too. We cast to integer elsewhere.

torchgeo/samplers/utils.py
```diff
@@ -55,17 +55,16 @@ def get_random_bounding_box(
...
 
     minx = bounds.minx
     miny = bounds.miny
 
-    # random.randrange crashes for inputs <= 0
```
adamjstewart, line 64: This is a dead comment, we no longer use `random.randrange` and the input can no longer be negative

torchgeo/samplers/utils.py
```diff
@@ -55,17 +55,16 @@ def get_random_bounding_box(
...
-    if width > 0:
```
adamjstewart, line 65: No longer need to guard for negative numbers, it doesn't matter if the sample starts outside the bounds of the image for bounding boxes smaller than the patch size.

torchgeo/samplers/utils.py
```diff
@@ -55,17 +55,16 @@ def get_random_bounding_box(
...
-        minx += torch.rand(1).item() * width * res
-    if height > 0:
-        miny += torch.rand(1).item() * height * res
+    # Use an integer multiple of res to avoid resampling
+    minx += int(torch.rand(1).item() * width) * res
```
adamjstewart, line 66: This is the only real important line

adamjstewart:
Evaluation using #1972:

| | raw (random) | raw (grid) | preprocessed (random) | preprocessed (grid) |
| --- | --- | --- | --- | --- |
| before | 17.223 | 10.974 | 15.685 | 4.6075 |
| after | 17.360 | 11.032 | 9.6127 | 4.6673 |

A 60% speedup for RandomGeoSampler when data is preprocessed (same CRS, res, TAP, COG).

isaaccorley (approved):

PR #1970, opened 2024-03-29
Add South Africa Crop Type Dataloader

GeorgeHuber:


GeorgeHuber:
@adamjstewart For some reason the data seems to be loading incorrectly and throwing an error in the tests but the dataset works fine with test data locally. Do you have any ideas for why this could be occurring? I was thinking it could be something related to PyTorch Lightning but am not too familiar with it.

adamjstewart:
If the tests pass locally but fail in CI, the reason is usually that your local copy was branched off of main a while ago and upstream main has since changed. In CI, your branch is merged with the current main to make sure that CI won't start breaking only after your PR has been merged. I suggest updating your local branch to experiment locally, or just trusting that CI is right.

GeorgeHuber:
> If the tests pass locally but fail in CI, the reason is usually that your local copy was branched off of main a while ago and upstream main has since changed. In CI, your branch is merged with the current main to make sure that CI won't start breaking only after your PR has been merged. I suggest updating your local branch to experiment locally, or just trusting that CI is right.

Thanks, I seem to have successfully merged upstream and am now seeing the same thing locally as in CI. My datamodule seems to be failing when splitting the dataset with 
`generator = torch.Generator().manual_seed(0)
  (self.train_dataset, self.val_dataset, self.test_dataset) = (
            random_bbox_assignment(dataset, [0.8, 0.1, 0.1], generator)
        )
`.
The resultant datasets do not include all bands for every bbox for some reason. Could this have something to do with the data being structured with a different file for every band?

adamjstewart:
> The resultant datasets do not include all bands for every bbox for some reason.

For datasets where each band is stored in a separate file, only a single band will be in the R-tree index.

adamjstewart:
@yichiac this error message looks familiar, did we already solve something like this in a different PR?

torchgeo/datasets/south_africa_crop_type.py
```diff
@@ -62,10 +62,9 @@ class SouthAfricaCropType(RasterDataset):
 
     filename_regex = r"""
         ^(?P<field_id>[0-9]*)
-        _(?P<date>[0-9]{4}_[0-9]{2}_[0-9]{2})
+        _(?P<timestamp>[0-9]{4}_07_[0-9]{2})
```
adamjstewart, line 65: Have you confirmed that all locations have a July image? Also, what is the purpose of avoiding the `<date>` name? Although we don't currently support time-series, we plan to in the near future.

torchgeo/datasets/south_africa_crop_type.py
```diff
@@ -62,10 +62,9 @@ class SouthAfricaCropType(RasterDataset):
...
         _(?P<band>(B[0-9A-Z]{2} | VH | VV))
         _10m"""
-    date_format = "%Y_%m_%d"
```
adamjstewart, line 68: Why was this change required?

torchgeo/datasets/south_africa_crop_type.py
```diff
@@ -101,7 +100,7 @@ def __init__(
         paths: str | Iterable[str] = "data",
         crs: CRS | None = None,
         classes: list[int] = list(cmap.keys()),
-        bands: list[str] = all_bands,
+        bands: list[str] = s2_bands,
```
adamjstewart, line 103: I'll let you decide which default value makes the most sense. We can override this in CI if needed in the YAML file.

torchgeo/datasets/south_africa_crop_type.py
```diff
@@ -153,7 +152,6 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
             raise IndexError(
                 f"query: {query} not found in index with bounds: {self.bounds}"
             )
-
```
adamjstewart, line 156: Seems unrelated, I actually like the empty line here

torchgeo/datasets/south_africa_crop_type.py
GeorgeHuber: Wrapping most of your comments together, I have found the bug around a design issue with the dataset as it pertains to the dataloader. Since time series datasets are not currently supported by torchgeo and each field has multiple dates of imagery available, the call to split the dataset into val, test, train splits includes uneven distributions of the original dataset.

For example: In the test data we have two fields 12 & 66 each with a set of imagery from July and another from May. The dataset split with a certain seed includes only may imagery in the test which then contains no valid chips. I can get this to work by changing the regex to only pick up files with the dates we're looking so we are guaranteed valid imagery is in each split but this has limitations as we will still include multiple dates in July for a single field when we only use one which will slow down training. I think a better solution would be to sample the datasets by field ID though I am not sure how to do this. Additionally since S1 and S2 data are in different folders, the R-tree creates an entry for both S1 and S2 for each field leading to similar problems with split. I think this functionality would either have to be added to the base torchgeo-geodataset or this dataset will only work with either s1 or s2 but not both at the same time. This should be fine for our research purposes though.

adamjstewart: > Additionally since S1 and S2 data are in different folders, the R-tree creates an entry for both S1 and S2 for each field leading to similar problems with split. I think this functionality would either have to be added to the base torchgeo-geodataset or this dataset will only work with either s1 or s2 but not both at the same time.

The "right" way to handle the case of both S1 and S2 is to use `IntersectionDataset` like I did in #1972. But like you said, for our research, we only currently care about S2. So we can leave support for both up to future developers.

torchgeo/datasets/south_africa_crop_type.py
```diff
@@ -60,11 +61,17 @@ class SouthAfricaCropType(RasterDataset):
     .. versionadded:: 0.6
     """
 
-    filename_regex = r"""
+    s1_regex = r"""
         ^(?P<field_id>[0-9]*)
-        _(?P<date>[0-9]{4}_[0-9]{2}_[0-9]{2})
-        _(?P<band>(B[0-9A-Z]{2} | VH | VV))
+        _(?P<date>[0-9]{4}_07_[0-9]{2})
+        _(?P<band>( VH || "VV" ))
```
adamjstewart, line 67: Why is VV in quotes?

adamjstewart (approved):

PR #1972, opened 2024-04-01
I/O Bench: add new dataset

### LINKED ISSUE #190: GeoDataset performance
We have `benchmark.py` for evaluating the performance of different sampler methods over a `GeoDataset`. Currently `benchmark.py` works with a ZipDataset of Landsat 8 imagery and CDL labels. Generally, we want TorchGeo to be able to sample arbitrary patches of data from GeoDatasets as quickly as possible.

There are several GeoDatasets cases that we need to consider:
- When both _reprojection_ and _resampling_ need to be done
- When only _resampling_ needs to be done
- When neither needs to be done

This issue is to track the performance of these cases.

adamjstewart:
Our [paper](https://arxiv.org/abs/2111.08872) includes an I/O performance benchmark consisting of 114 Landsat 8 Level-2 scenes and 1 CDL 2019 mask. Since its publication, we have not been benchmarking I/O during PRs or between releases. This has led to difficulty in properly evaluating proposed changes (#1881) and massive performance bugs that have gone undetected for years (#1968).

This PR adds a dataset to properly test this. It consists of a single Landsat 9 scene and CDL mask with the following splits:

* original: the original files as downloaded from USGS Earth Explorer and USDA CropScape
* raw: the same files with compression and with CDL clipped to the bounds of the Landsat scene
* preprocessed: the same files with compression, reprojected to the same CRS, as COGs, with TAP

These roughly correspond to the same categories as used in our paper, with a few changes:

* Only a single Landsat scene
* CDL is clipped to the bounds of Landsat

These changes are made to make the dataset as small as possible so it can be quickly downloaded and fit on any system. I believe that the dataset is still useful even at such a small size. This should be considered to be version 1 of the dataset, with many future changes to evaluate a broader set of conditions.

- [x] Add dataset
- [x] Add data module
- [x] Add trainer
- [x] Add tests
- [x] Add documentation

Sample image over Champaign County, IL, USA:

iobench

---

This PR also represents an interesting and novel dataset design. Note that instead of writing a custom `RasterDataset` and overriding `__getitem__` to handle both Landsat and CDL files, I'm actually subclassing from `IntersectionDataset`, instantiating Landsat and CDL classes, and computing the intersection. This is significantly easier and could be used to simplify our implementations of other raster datasets with both images and masks: AgriFieldNet, Chesapeake, EnviroAtlas, GlobBiomass, L7 Irish, L8 Biome, LandCover.ai, etc.

FYI @yichiac

Closes #190


adamjstewart:
Trying to resist the urge to write a custom profiler that automatically formats the output we care about as a Markdown table: https://lightning.ai/docs/pytorch/stable/tuning/profiler_expert.html

PR #1978, opened 2024-04-03
Remove AugPipe

ashnair1:
The next version of kornia (`0.7.3`) will contain 2 fixes (kornia/kornia#2846, kornia/kornia#2856) that will allow us to drop AugPipe and have detection only depend on kornia's `AugmentationSequential`.

### DIFF
--- a/tests/datasets/test_vhr10.py
+++ b/tests/datasets/test_vhr10.py
@@ -85,7 +85,7 @@ class TestVHR10:
                 x['prediction_labels'] = x['class']
                 x['prediction_boxes'] = x['bbox_xyxy']
                 x['prediction_scores'] = torch.Tensor([scores[i]])
-                if 'masks' in x:
+                if 'mask' in x:
                     x['prediction_masks'] = x['mask']
                     dataset.plot(x, show_feats='masks')
                     plt.close()
--- a/tests/datasets/test_vhr10.py
+++ b/tests/datasets/test_vhr10.py
@@ -83,7 +83,7 @@ class TestVHR10:
             for i in range(3):
                 x = dataset[i]
                 x['prediction_labels'] = x['label']
-                x['prediction_boxes'] = x['bbox_xyxy']
+                x['prediction_bbox_xyxy'] = x['bbox_xyxy']
                 x['prediction_scores'] = torch.Tensor([scores[i]])
                 if 'mask' in x:
                     x['prediction_masks'] = x['mask']
--- a/tests/datasets/test_vhr10.py
+++ b/tests/datasets/test_vhr10.py
@@ -41,7 +41,7 @@ class TestVHR10:
             assert isinstance(x, dict)
             assert isinstance(x['image'], torch.Tensor)
             if dataset.split == 'positive':
-                assert isinstance(x['class'], torch.Tensor)
+                assert isinstance(x['label'], torch.Tensor)
                 assert isinstance(x['bbox_xyxy'], torch.Tensor)
                 if 'mask' in x:
                     assert isinstance(x['mask'], torch.Tensor)
@@ -82,7 +82,7 @@ class TestVHR10:
             scores = [0.7, 0.3, 0.7]
             for i in range(3):
                 x = dataset[i]
-                x['prediction_labels'] = x['class']
+                x['prediction_labels'] = x['label']
                 x['prediction_boxes'] = x['bbox_xyxy']
                 x['prediction_scores'] = torch.Tensor([scores[i]])
                 if 'mask' in x:
--- a/tests/datasets/test_vhr10.py
+++ b/tests/datasets/test_vhr10.py
@@ -41,10 +41,10 @@ class TestVHR10:
             assert isinstance(x, dict)
             assert isinstance(x['image'], torch.Tensor)
             if dataset.split == 'positive':
-                assert isinstance(x['labels'], torch.Tensor)
-                assert isinstance(x['boxes'], torch.Tensor)
-                if 'masks' in x:
-                    assert isinstance(x['masks'], torch.Tensor)
+                assert isinstance(x['class'], torch.Tensor)
+                assert isinstance(x['bbox_xyxy'], torch.Tensor)
+                if 'mask' in x:
+                    assert isinstance(x['mask'], torch.Tensor)
 
     def test_len(self, dataset: VHR10) -> None:
         if dataset.split == 'positive':
@@ -82,10 +82,10 @@ class TestVHR10:
             scores = [0.7, 0.3, 0.7]
             for i in range(3):
                 x = dataset[i]
-                x['prediction_labels'] = x['labels']
-                x['prediction_boxes'] = x['boxes']
+                x['prediction_labels'] = x['class']
+                x['prediction_boxes'] = x['bbox_xyxy']
                 x['prediction_scores'] = torch.Tensor([scores[i]])
                 if 'masks' in x:
-                    x['prediction_masks'] = x['masks']
+                    x['prediction_masks'] = x['mask']
                     dataset.plot(x, show_feats='masks')
                     plt.close()
--- a/tests/transforms/test_transforms.py
+++ b/tests/transforms/test_transforms.py
@@ -6,7 +6,7 @@ import pytest
 import torch
 from torch import Tensor
 
-from torchgeo.transforms import indices, transforms
+from torchgeo.transforms import indices
 from torchgeo.transforms.transforms import _ExtractPatches
 
 # Kornia is very particular about its boxes:
@@ -203,12 +203,12 @@ def test_extract_patches() -> None:
         'image': torch.randn(size=(b, c, h, w)),
         'mask': torch.randint(low=0, high=2, size=(b, h, w)),
     }
-    train_transforms = transforms.AugmentationSequential(
-        _ExtractPatches(window_size=p), same_on_batch=True, data_keys=['image', 'mask']
+    train_transforms = K.AugmentationSequential(
+        _ExtractPatches(window_size=p), same_on_batch=True, data_keys=None
     )
     output = train_transforms(batch)
-    assert batch['image'].shape == (b * num_patches, c, p, p)
-    assert batch['mask'].shape == (b * num_patches, p, p)
+    assert output['image'].shape == (b * num_patches, c, p, p)
+    assert output['mask'].shape == (b * num_patches, 1, p, p)
 
     # Test different stride
     s = 16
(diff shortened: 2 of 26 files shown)


torchgeo/datamodules/nasa_marine_debris.py
```diff
@@ -45,12 +44,10 @@ def __init__(
...
 
-        self.aug = AugPipe(
-            AugmentationSequential(
-                K.Normalize(mean=self.mean, std=self.std), data_keys=["image", "boxes"]
-            ),
-            batch_size,
+        self.aug = K.AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std), data_keys=None, keepdim=True
         )
+        self.aug.keepdim = True  # type: ignore[attr-defined]
```
ashnair1, line 50: For now, the attribute has to be set manually. Refer to kornia/kornia#2848

adamjstewart:
How close are we to also removing our custom AugmentationSequential wrapper?

ashnair1:
> How close are we to also removing our custom AugmentationSequential wrapper?

Pretty close actually. If `kornia==0.7.3` is released soon, we should be able to fully remove it by next release.

torchgeo/datamodules/nasa_marine_debris.py
adamjstewart: Let's add a comment here like we did in the other PR, I won't remember this even a week from now.

adamjstewart: Same thing every other location we need this. Hopefully the upstream bug gets resolved soon and we can remove all these comments.

robmarkcole:
Noting https://github.com/kornia/kornia/releases/tag/v0.7.3

adamjstewart:
@ashnair1 is this waiting on kornia 0.7.4 for something?

ashnair1:
Yeah #2147 needs to be merged before this and that needs kornia 0.7.4.

robmarkcole:
Note: 0.7.4 will be released by the end of September

calebrob6:
This is good to go forward now

adamjstewart:
Drop in coverage is due to some lines in our `AugmentationSequential` wrapper that are no longer hit. Can we just delete this wrapper now, or are we using it somewhere else?

ashnair1:
> Drop in coverage is due to some lines in our `AugmentationSequential` wrapper that are no longer hit. Can we just delete this wrapper now, or are we using it somewhere else? Ideally we would add some simple tests and deprecate it so as not to confuse too many people in the next release.

No, the `AugmentationSequential` wrapper can be removed in this PR. I just needed to update tests to see if they were working as expected.

Seems `_ExtractPatches` can be removed as well. I've switched to using kornia's [ExtractTensorPatches](https://kornia.readthedocs.io/en/latest/contrib.html#kornia.contrib.ExtractTensorPatches) instead.

torchgeo/transforms/transforms.py
```diff
@@ -7,107 +7,8 @@
...
-from einops import rearrange
-from kornia.contrib import extract_tensor_patches
 from kornia.geometry import crop_by_indices
-from kornia.geometry.boxes import Boxes
 from torch import Tensor
-from torch.nn.modules import Module
-
-
-# TODO: contribute these to Kornia and delete this file
-class AugmentationSequential(Module):
```
adamjstewart, line 19: Maybe we can have a simple:
```python
from kornia.augmentation import AugmentationSequential

__all__ = ['AugmentationSequential']
```
This will maintain backwards compatibility. Even better would be to find a way to deprecate it like in #2396. Maybe something like:
```python
@deprecated('Use kornia.augmentation.AugmentationSequential instead')
class AugmentationSequential(K.AugmentationSequential):
    ...
```
I'm just worried that this class has been here since TorchGeo 0.1 and although it isn't documented, it's used in all of our builtin datamodules and all of our tutorials.

isaaccorley:
We need to keep `_ExtractPatches` because `kornia.contrib.ExtractTensorPatches` doesn't inherit from the GeometricAugmentation so it won't apply to both image and masks when passing the batch dict.

tests/transforms/test_transforms.py
```diff
@@ -215,12 +203,12 @@ def test_extract_patches() -> None:
         'image': torch.randn(size=(b, c, h, w)),
         'mask': torch.randint(low=0, high=2, size=(b, h, w)),
     }
-    train_transforms = transforms.AugmentationSequential(
-        _ExtractPatches(window_size=p), same_on_batch=True, data_keys=['image', 'mask']
-    )
-    output = train_transforms(batch)
-    assert batch['image'].shape == (b * num_patches, c, p, p)
-    assert batch['mask'].shape == (b * num_patches, p, p)
+    train_transforms = ExtractTensorPatches(p, s)
```
isaaccorley, line 206: This should be used inside of kornia's AugmentationSequential so we can know that it works inside the datamodules. Our datamodules don't apply to image and mask in separate calls.

isaaccorley (dismissed):

ashnair1:
Scoping this back to just removing AugPipe. `AugmentationSequential` can't be removed yet as `_ExtractPatches` doesn't work correctly with kornia's `AugmentationSequential`.

isaaccorley:
@adamjstewart Do we want to remove _ExtractPatches and I can make a PR in kornia for an ExtractTensorPatches augmentation? Pinging @edgarriba for awareness.

edgarriba:
@isaaccorley sounds good to me. Feel free to join our discourse to chat about it: https://discord.gg/Vf3QgRq2 @johnnv1

adamjstewart:
I would like to remove _everything_ in `torchgeo/transforms/transforms.py` sooner rather than later. So yes, anything anyone is willing to upstream I would be happy to help review. In particular, I'm also interested in upstreaming RandomNCrop. We noticed a bug in it when using data with a time dimension so we just need to fix that.

isaaccorley:
Then yes sorry @ashnair1, let's remove AugmentationSequential and _ExtractPatches

adamjstewart:
Or deprecate. It may take a while for upstream to make a new release.

ashnair1:
> Or deprecate. It may take a while for upstream to make a new release.

I think it might be better to keep the current scope of the PR i.e. removal of `AugPipe`. Deprecation can be done in #2396 right?

torchgeo/trainers/detection.py
```diff
@@ -238,8 +238,12 @@ def training_step(
         """
         x = batch['image']
         batch_size = x.shape[0]
+        # Get bbox key as it can be one of {"bbox", "bbox_xyxy", "bbox_xywh"}
```
adamjstewart, line 241: Are our models compatible with all 3, or do we need to assert that a specific format is used?

tests/datasets/test_vhr10.py
```diff
@@ -41,10 +41,10 @@ def test_getitem(self, dataset: VHR10) -> None:
             assert isinstance(x, dict)
             assert isinstance(x['image'], torch.Tensor)
             if dataset.split == 'positive':
-                assert isinstance(x['labels'], torch.Tensor)
-                assert isinstance(x['boxes'], torch.Tensor)
-                if 'masks' in x:
-                    assert isinstance(x['masks'], torch.Tensor)
+                assert isinstance(x['class'], torch.Tensor)
```
adamjstewart, line 44: Why the switch from labels to class? I don't think either are standard in Kornia

tests/datasets/test_vhr10.py
```diff
@@ -82,10 +82,10 @@ def test_plot(self, dataset: VHR10) -> None:
             scores = [0.7, 0.3, 0.7]
             for i in range(3):
                 x = dataset[i]
-                x['prediction_labels'] = x['labels']
-                x['prediction_boxes'] = x['boxes']
+                x['prediction_labels'] = x['class']
+                x['prediction_boxes'] = x['bbox_xyxy']
```
adamjstewart, line 86: I think we should also change the names of the `prediction_*` keys to match the new input keys.

torchgeo/trainers/detection.py
ashnair1: Fixed. Our detection models only work with `bbox_xyxy`.

tests/datasets/test_vhr10.py
ashnair1: Class is a valid DataKey in kornia. Refer to https://github.com/kornia/kornia/blame/beb400c1de2afd0cd102511ed64fc6a7a5b4ba19/kornia/constants.py#L137

But I had also added Label as an alias

tests/datasets/test_vhr10.py
ashnair1: So in this case, since our current detection models only accept boxes in `xyxy` format, the key would be `prediction_bbox_xyxy` right?

adamjstewart: Correct

tests/datasets/test_vhr10.py
adamjstewart: Ah, didn't realize that. Let's make all datasets consistent.

torchgeo/datasets/zuericrop.py
```diff
@@ -11,9 +11,14 @@
 from matplotlib.figure import Figure
 from torch import Tensor
 
-from .errors import DatasetNotFoundError, RGBBandsMissingError
-from .geo import NonGeoDataset
-from .utils import Path, download_url, lazy_import, percentile_normalization
+from torchgeo.datasets.errors import DatasetNotFoundError, RGBBandsMissingError
+from torchgeo.datasets.geo import NonGeoDataset
+from torchgeo.datasets.utils import (
```
adamjstewart, line 16: I prefer relative imports for `torchgeo/` things

adamjstewart (approved):
One minor undo and then I think this is ready to merge!

adamjstewart (dismissed):

adamjstewart:
Drop in coverage is for `AugmentationSequential`: https://app.codecov.io/gh/microsoft/torchgeo/pull/1978/blob/torchgeo/transforms/transforms.py

Can you add a couple tests?

adamjstewart:
Alternatively, convert our AugmentationSequential to an alias of Kornia's so you don't need to add any tests. Or I can do that in my PR, but my PR can't be merged without this PR.

ashnair1:
Making our `AugmentationSequential` an alias for kornia's won't work as `_ExtractPatches` won't work with it. It would have to be removed. Refer to https://github.com/microsoft/torchgeo/pull/1978#issuecomment-2564361152.

adamjstewart:
Ah right. So we have to figure out how to fix or replace `_ExtractPatches` before either of our PRs can be merged 😢

adamjstewart:
> `_ExtractPatches` doesn't work correctly with kornia's `AugmentationSequential`.

Can you clarify this? I can hack this to work. Would like to push forward on this PR, either by adding more tests or fixing compatibility.

ashnair1:
> > `_ExtractPatches` doesn't work correctly with kornia's `AugmentationSequential`.
> 
> Can you clarify this? I can hack this to work. Would like to push forward on this PR, either by adding more tests or fixing compatibility.

Turns out the test was incorrect. Our AugmentationSequential was changing the input batch itself and that's what we were testing. This should be good to go now.

adamjstewart:
Drop in coverage is for our `AugmentationSequential`. In `torchgeo/transforms/transforms.py`, can you replace our `AugmentationSequential` with:
```python
from kornia.augmentation import AugmentationSequential

# Only include import redirects
__all__ = ('AugmentationSequential',)
```
I'll figure out the deprecation stuff in another PR.

adamjstewart:
Well this is funny. The documentation tests are failing because it's trying to document Kornia's `AugmentationSequential` but we're missing an image file. In `docs/api/transforms.rst`, let's use:
```rst
.. automodule:: torchgeo.transforms
   :exclude-members: AugmentationSequential
```

adamjstewart (approved):
Thanks again for all your work on wrangling the sample dicts to have a consistent format! I'll start work on deprecation and type hints.

PR #1994, opened 2024-04-13
Use ruff

### LINKED ISSUE #1986: Use ruff
### Summary

I'm starting to seriously consider using [ruff](https://github.com/astral-sh/ruff) to replace the following style tools we currently use:

* black
* flake8
* isort
* pydocstyle
* pyupgrade

It doesn't yet handle static type analysis (mypy), but the devs [would like to](https://github.com/astral-sh/ruff/issues/3893)

### Rationale

A few reasons:

* It supports pyproject.toml, unlike flake8/pre-commit/pyupgrade
* Fewer dependencies to track and manage
* It's a single tool to run instead of multiple (although pre-commit slightly alleviates this)
* It's fast (only really matters for isort which contains a [bug](https://github.com/PyCQA/isort/issues/1912) where gitignore is not ignored)
* My favorite style tool, pydocstyle, is now [deprecated](https://github.com/PyCQA/pydocstyle/pull/658) and contains known [bugs](https://github.com/PyCQA/pydocstyle/issues/459) 
* It's rapidly [gaining popularity](https://github.com/astral-sh/ruff?tab=readme-ov-file#whos-using-ruff), including recent adoption by [PyTorch](https://github.com/pytorch/pytorch/pull/99785)

### Implementation

Easy to implement, only need to decide on the exact configuration details. Just curious to see how many files change and what new issues get flagged.

### Alternatives

A lot of users have suggested this to me. I'm unaware of any popular alternatives.

### Additional information

Only downside is that ruff is built in rust. This doesn't matter for pip/conda users, but rust is non-trivial to build from source. But kornia also requires rust now, so we can't really avoid that anyway. Seems like rust is the future of Python, and will become harder to avoid in the future.

adamjstewart:
- [x] Configure ruff
- [x] Apply format changes
- [x] Fix linter errors
- [x] Update requirements files
- [x] Update GitHub Actions
- [x] Update docs
- [x] Update pre-commit
- [x] Update VS Code

Closes #1986


.devcontainer/devcontainer.json
```diff
@@ -20,21 +20,16 @@
   "customizations": {
     "vscode": {
       "extensions": [
+        "GitHub.codespaces",
+        "charliermarsh.ruff",
         "ms-python.python",
-        "ms-python.isort",
-        "ms-python.black-formatter",
-        "ms-python.vscode-pylance",
```
adamjstewart, line 26: Got rid of pylance since we aren't currently using it anywhere else. Also, haven't tested this file since I don't know how to use it.

.pre-commit-config.yaml
adamjstewart, line 1: @ashnair1 can you test this?

I'm wondering if we should remove mypy from pre-commit since it's so slow. 

We could also add the [pre-commit](https://github.com/pre-commit/action) action so that fixes are automatically applied to PRs.

docs/tutorials/pretrained_weights.ipynb
```diff
@@ -79,8 +79,8 @@
     "from lightning.pytorch import Trainer\n",
     "\n",
     "from torchgeo.datamodules import EuroSAT100DataModule\n",
-    "from torchgeo.trainers import ClassificationTask\n",
-    "from torchgeo.models import ResNet18_Weights"
+    "from torchgeo.models import ResNet18_Weights\n",
+    "from torchgeo.trainers import ClassificationTask"
```
adamjstewart, line 83: Seems like isort and pyupgrade did not have Jupyter notebook support

experiments/ssl4eo/landsat/plot_landsat_bands.py
```diff
@@ -46,9 +46,7 @@
 df = df.iloc[::-1]
 
 fig, ax = plt.subplots(figsize=(5.5, args.fig_height))
-ax1, ax2 = fig.subplots(
-    nrows=1, ncols=2, gridspec_kw={"width_ratios": [3, 1]}
-)  # type: ignore[misc]
+ax1, ax2 = fig.subplots(nrows=1, ncols=2, gridspec_kw={"width_ratios": [3, 1]})  # type: ignore[misc]
```
adamjstewart, line 49: https://docs.astral.sh/ruff/formatter/black/#pragma-comments-are-ignored-when-computing-line-width

requirements/min-reqs.old
```diff
@@ -33,20 +33,6 @@ scikit-image==0.19.0
...
-# style
-black[jupyter]==21.8b0
-flake8==3.8.0
-isort[colors]==5.8.0
-pydocstyle[toml]==6.1.0
-pyupgrade==2.8.0
-
 # tests
-mypy==0.900
-nbmake==1.3.3
```
adamjstewart, line 50: We don't actually run any docs/style tests with the min version, no need to install those deps.

requirements/tests.txt
```diff
@@ -1,5 +1,4 @@
 # tests
-mypy==1.9.0
```
adamjstewart, line 2: I moved "mypy" under "style" in CI but forgot to move it in the requirements files.

torchgeo/datasets/cms_mangrove_canopy.py
```diff
@@ -205,17 +205,13 @@ def __init__(
         assert isinstance(country, str), "Country argument must be a str."
         assert (
             country in self.all_countries
-        ), "You have selected an invalid country, please choose one of {}".format(
-            self.all_countries
-        )
+        ), f"You have selected an invalid country, please choose one of {self.all_countries}"
```
adamjstewart, line 208: https://github.com/asottile/pyupgrade?tab=readme-ov-file#f-strings

torchgeo/datasets/cropharvest.py
```diff
@@ -293,12 +293,12 @@ def _extract(self) -> None:
         features_path = os.path.join(self.root, self.file_dict["features"]["filename"])
         extract_archive(features_path)
 
-    def plot(self, sample: dict[str, Tensor], subtitle: str | None = None) -> Figure:
+    def plot(self, sample: dict[str, Tensor], suptitle: str | None = None) -> Figure:
```
adamjstewart, line 296: Actual bug caught by pydocstyle!

torchgeo/datasets/fair1m.py
```diff
@@ -239,8 +239,10 @@ def __init__(
 
         Args:
             root: root directory where dataset can be found
+            split: one of "train", "val", or "test"
             transforms: a function/transform that takes input sample and its target as
                 entry and returns a transformed version
+            download: if True, download dataset and store it in the root directory
```
adamjstewart, line 245: Lots of missing parameters in the docs

torchgeo/datasets/millionaid.py
```diff
@@ -290,7 +291,7 @@ def _load_files(self, root: str) -> list[dict[str, Any]]:
 
         images = imgs_no_subcat + imgs_subcat
 
-        files = [dict(image=img, label=l) for img, l in zip(images, labels)]
+        files = [dict(image=image, label=label) for image, label in zip(images, labels)]
```
adamjstewart, line 294: Linter complained about l being ambiguous (l vs. 1)

torchgeo/models/fcsiam.py
```diff
@@ -134,35 +134,8 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         """Initialize a new FCSiamConc model.
 
         Args:
-            encoder_name: Name of the classification model that will
```
adamjstewart, line 137: Only change in this PR that might be controversial. I think it's better to just point users to the upstream docs and say "ditto".

isaaccorley (approved):
LGTM

.github/workflows/style.yaml
```diff
@@ -132,33 +57,10 @@ jobs:
...
-      if: steps.cache.outputs.cache-hit != 'true'
+    - name: Run ruff checks
       run: |
-        pip install -r requirements/style.txt
-        pip cache purge
-    - name: List pip dependencies
-      run: pip list
-    - name: Run pyupgrade checks
-      run: pyupgrade --py310-plus $(find . -path ./docs/src -prune -o -name "*.py" -print)
+        ruff check --output-format=github --no-fix .
```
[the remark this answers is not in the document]
adamjstewart: Nope, it displays them for me: https://github.com/microsoft/torchgeo/actions/runs/8675768435/job/23789552352

adamjstewart:
I actually plan to expand the number of rules (stricter pydocstyle) and make some other formatting changes (double -> single quotes) but will save that for another PR. So far, this PR simply replaces our current tools with a single, faster, less buggy tool.

adamjstewart:
I've tried rerunning the minimum tests 5 or 6 times but no luck. Pretty sure codecov is dead: #1995

.pre-commit-config.yaml
ashnair1: Tested the file and it works fine. 

The slowness observed in the mypy hook occurs when there is a change in the additional dependencies as the isolated environment (where pre-commit runs the hooks) needs to be recreated. Once it's done, subsequent runs will be faster.

Incorporating the `pre-commit` action is a good idea. That way, new contributors won't have to worry about style. From a CI perspective, since we already have a Github action for checking `mypy`, we wouldn't need a mypy hook anymore in the pre-commit config anymore.

On the flip side, having the mypy hook in the pre-commit config is useful for testing mypy when you don't have the latest versions of the libraries installed locally :)

isaaccorley:
Shall we merge w/o codecov passing?

ashnair1 (approved):

calebrob6:
How fast is this now?

adamjstewart:
### Before

```console
> time begin; black .; flake8; isort .; pydocstyle; pyupgrade --py310-plus $(find . -path ./docs/src -prune -o -name "*.py" -print); end
...
Executed in    7.18 secs    fish           external
   usr time   12.33 secs    1.30 millis   12.33 secs
   sys time    2.74 secs    9.48 millis    2.73 secs
```

### After

```console
> time begin; ruff check; ruff format; end
...
Executed in   83.13 millis    fish           external
   usr time   48.57 millis    0.17 millis   48.41 millis
   sys time   83.33 millis    2.75 millis   80.58 millis
```

PR #1997, opened 2024-04-15
QuakeSet dataset

isaaccorley:
This PR adds the QuakeSet dataset from the paper ["QuakeSet: A Dataset and Low-Resource Models to Monitor Earthquakes through Sentinel-1", Cambrin et al.](https://arxiv.org/abs/2403.18116). This is also the dataset that will be used for the new ECML-PKDD 2024 Seismic Analysis

Sample plots (plotting (VV, VH, VV / VH) as RGB false color image:

image1
 and Monitoring Challenge (SMAC).

image2

@DarthReca @nilsleh


adamjstewart:
Drop in coverage is real, you're missing coverage for ImportError

docs/api/non_geo_datasets.csv
```diff
@@ -29,6 +29,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `PASTIS`_,I,Sentinel-1/2,"CC-BY-4.0","2,433",19,128x128xT,10,MSI
 `PatternNet`_,C,Google Earth,-,"30,400",38,256x256,0.06--5,RGB
 `Potsdam`_,S,Aerial,-,38,6,"6,000x6,000",0.05,MSI
+`QuakeSet`_,C,Sentinel-1,"OpenRAIL","3,327",2,512x512,5,SAR
```
nilsleh, line 32: do we also need to add "R" for Regression?

torchgeo/datamodules/quakeset.py
```diff
@@ -0,0 +1,42 @@
...
+from .geo import NonGeoDataModule
+
+
+class QuakeSetDataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the QuakeSet dataset.
+
+    .. versionadded:: 0.6
+    """
+
+    mean = torch.tensor(0.0)
```
nilsleh, line 22: would it help if we actually compute some reasonable normalization statistics for the challenge based on the train set? or should we leave that to participants to decide how they scale data?

torchgeo/datasets/quakeset.py
```diff
@@ -0,0 +1,289 @@
...
+    for the Seismic Monitoring and Analysis (SMAC) ECML-PKDD 2024 Discovery Challenge.
+
+    Dataset features:
+
+    * Sentinel-1 SAR imagery
+    * before/pre/post imagery of areas affected by earthquakes
+    * 2 multispectral bands (VV/VH)
+    * 3,327 pairs of pre and post images with 5 m per pixel resolution (512x512 px)
+    * 2 classification labels (unaffected / affected by earthquake)
+    * earthquake magnitudes for each sample
```
nilsleh, line 34: I guess overall question, do we want to have Classification and Regression supported out of the box?

nilsleh: The magnitude is returned by default? But maybe we could also test it with a RegressionTask then?

isaaccorley: I think the issue is that they both can't be called "label" in the sample dict. So the only workaround would be to support only one or the other

torchgeo/datamodules/quakeset.py
isaaccorley: We can do this but I wasn't sure if that's the best way to normalize Sentinel-1 bands

isaaccorley:
> Drop in coverage is real, you're missing coverage for ImportError

It only took a few hours to get reported 🫠

torchgeo/datasets/quakeset.py
adamjstewart: This is an interesting observation, will add a note in #985

isaaccorley: I think we should leave it as a classification task for now and just return the magnitude as additional data if the user wants to use it.

nilsleh (approved):

torchgeo/datasets/quakeset.py
```diff
@@ -0,0 +1,289 @@
...
+    """QuakeSet dataset.
+
+    `QuakeSet <https://huggingface.co/datasets/DarthReca/quakeset>`__
+    is a dataset for Earthquake Change Detection and Magnitude Estimation and is used
+    for the Seismic Monitoring and Analysis (SMAC) ECML-PKDD 2024 Discovery Challenge.
+
+    Dataset features:
+
+    * Sentinel-1 SAR imagery
+    * before/pre/post imagery of areas affected by earthquakes
```
adamjstewart, line 30: What is the difference between before and pre?

torchgeo/datasets/quakeset.py
```diff
@@ -0,0 +1,289 @@
...
+    * 2 multispectral bands (VV/VH)
```
adamjstewart, line 31:
```suggestion
    * 2 SAR bands (VV/VH)
```

torchgeo/datasets/quakeset.py
```diff
@@ -0,0 +1,289 @@
...
+        image = sample["image"].permute((1, 2, 0)).numpy()
+        label = cast(int, sample["label"].item())
+        label_class = self.classes[label]
+
+        # Create false color image for pre image
+        vv = percentile_normalization(image[..., 0]) + 1e-16
+        vh = percentile_normalization(image[..., 1]) + 1e-16
+        pre_fci = np.stack([vv, vh, vv / vh], axis=-1).clip(0, 1)
+
+        # Create false color image for post image
```
adamjstewart, line 253: What about before?

torchgeo/datasets/quakeset.py
isaaccorley: The before+pre = unaffected area, pre+post = earthquake affected area. You can think of before+pre as a hard negative because it's of the same location of the earthquake but a "change" pair prior to the event.

torchgeo/datasets/quakeset.py
isaaccorley: Pre/post has different meanings here, it's just the images at t=1 and t=2. Maybe I can just rename as image1, image2?

torchgeo/datasets/quakeset.py
adamjstewart: Let's add this to the docstring in some form.

torchgeo/datasets/quakeset.py
isaaccorley: Changed to `fci1`, `fci2`

torchgeo/datasets/quakeset.py
isaaccorley: done

torchgeo/datasets/quakeset.py
isaaccorley: done

torchgeo/datasets/quakeset.py
```diff
@@ -0,0 +1,291 @@
...
+
+       * `h5py <https://pypi.org/project/h5py/>`_ to load the dataset
+
+    .. versionadded:: 0.6
+    """
+
+    filename = "earthquakes.h5"
+    url = "https://hf.co/datasets/DarthReca/quakeset/resolve/bead1d25fb9979dbf703f9ede3e8b349f73b29f7/earthquakes.h5"
+    md5 = "76fc7c76b7ca56f4844d852e175e1560"
+    splits = {"train": "train", "val": "validation", "test": "test"}
```
adamjstewart, line 63: Why not use "validation" instead of "val"? Then we don't need a dict to map.

torchgeo/datasets/quakeset.py
```diff
@@ -0,0 +1,291 @@
...
+                        data.append(
+                            dict(
+                                key=k,
+                                patch=patch,
+                                images=("before", "pre"),
+                                label=0,
+                                magnitude=0.0,
+                            )
+                        )
+        f.close()
```
adamjstewart, line 176: Can we use `with h5py.File(...) as f:`?

torchgeo/datasets/quakeset.py
```diff
@@ -0,0 +1,291 @@
...
+            the sample keys, patches, images, labels, and magnitudes
+        """
+        import h5py
+
+        f = h5py.File(self.filepath)
+
+        data = []
+        for k in sorted(f.keys()):
+            if f[k].attrs["split"] != self.splits[self.split]:
+                continue
```
adamjstewart, line 149: What is this needed for?

torchgeo/datasets/quakeset.py
isaaccorley: The dataset h5 file attributes uses "validation". All other torchgeo dataset splits use "val".

torchgeo/datasets/quakeset.py
isaaccorley: Seems like that just adds more indentation

torchgeo/datasets/quakeset.py
isaaccorley: To skip samples that aren't part of the split

torchgeo/datasets/quakeset.py
adamjstewart: > All other torchgeo dataset splits use "val".

This is by coincidence, not rule. So2Sat also uses "validation".

torchgeo/datasets/quakeset.py
adamjstewart: It also helps prevent file handle leaks if the process dies inside the loop.

adamjstewart: Indentation can be reduced by inverting the if-statement on line 152, or by saving the dict to a variable before appending.

torchgeo/datasets/quakeset.py
isaaccorley: Ahh okay will change

isaaccorley:
@adamjstewart Seems like the classification task tests don't like "validation" as a split but I'm not seeing where it's being fixed for so2sat

isaaccorley:
We should probably rethink the splits because in torchgeo.datamodules.geo we are hardcoding the setup method to use `train`, `val`, `test`. Removing the one line in QuakeSet dataset requires adding a setup method to the QuakeSet datamodule with 10+ lines. Going to revert back

torchgeo/datasets/quakeset.py
```diff
@@ -0,0 +1,289 @@
...
+            root: root directory where dataset can be found
+            split: one of "train", "val", or "test"
+            transforms: a function/transform that takes input sample and its target as
+                entry and returns a transformed version
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+
+        Raises:
+            AssertionError: If ``split`` argument is invalid.
+            DatasetNotFoundError: If dataset is not found and *download* is False.
```
adamjstewart, line 86: ImportError: If h5py is not installed.

isaaccorley:
Any chance we could get this merged today? The competition has already started.

adamjstewart (approved):

PR #2003, opened 2024-04-16
Fix plotting in datamodules when dataset is a Subset

calebrob6:
We have a `.plot(...)` method in our `BaseDataModule` that routes plot requests from the trainers through to the dataset object's `.plot(...)` method. This currently only works if `trainer.dataset` or `trainer.val_dataset` has a plot method. However, when you call `torch.utils.data.random_split` to get a train/val split, your dataset objects are `Subsets` which don't have a plot method.

### DIFF
--- a/torchgeo/datamodules/geo.py
+++ b/torchgeo/datamodules/geo.py
@@ -11,7 +11,7 @@ import torch
 from lightning.pytorch import LightningDataModule
 from matplotlib.figure import Figure
 from torch import Tensor
-from torch.utils.data import DataLoader, Dataset, default_collate, Subset
+from torch.utils.data import DataLoader, Dataset, Subset, default_collate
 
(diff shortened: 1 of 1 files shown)


isaaccorley (approved):

calebrob6:
I can confirm that with this change my plots show up in tensorboard :+1:

adamjstewart:
Very clever, don't know why I never thought of this.

Do any of our builtin datamodules suffer from this issue? Just want to make sure we have 100% test coverage.

robmarkcole:
I've def experienced this issue, likely OSCD or levir

adamjstewart:
Ah, I just noticed codecov actually works for this PR since it doesn't come from a fork.

adamjstewart (approved):

Issue #1289, opened 2023-04-25
DataModules: rename split parameters


adamjstewart:
### Summary

We should rename the split parameters we use in data modules.

### Rationale

Currently, our modules use `val_split_pct` and `test_split_pct`. However, our splitting functions only accept integer lengths or ratios, none of them actually accept percentages.

### Implementation

We should use "length" to be consistent with `torch.utils.data.random_split`. We should either use `lengths` for methods that accept a variable number of splits or `val_length`/`test_length` for methods that differentiate between the two.

### Alternatives

Open to suggestions for other names, but want to follow existing standards when possible.

isaaccorley:
`torchgeo.datamodules.utils.dataset_split` accepts percentages. Also `GeoDataModule` already has a `length` argument.

adamjstewart:
`torchgeo.datamodules.utils.dataset_split` was removed in #2005.

My suggestion is to replace all `*_split_pct` parameters with a single `split_lengths` parameter that gets passed to `torch.utils.data.random_split`. We can also rename `length` to `sampler_length` for clarity if we want to. This would be a backwards incompatible change though.

adamjstewart:
Going to ignore this for now, we have bigger backwards-incompatible changes to tackle than this.

PR #2005, opened 2024-04-17
Remove torchgeo.datamodules.dataset_split

adamjstewart:
This function was added as a wrapper around `torch.utils.data.random_split` with support for percentages. However, this feature was added upstream, and the minimum supported version of PyTorch now supports it. This wrapper can now be removed.

There are no backwards incompatible changes since this function was never documented.

### DIFF
--- a/torchgeo/datamodules/deepglobelandcover.py
+++ b/torchgeo/datamodules/deepglobelandcover.py
@@ -6,13 +6,13 @@
 from typing import Any
 
 import kornia.augmentation as K
+from torch.utils.data import dataset_split
 
 from ..datasets import DeepGlobeLandCover
 from ..samplers.utils import _to_tuple
 from ..transforms import AugmentationSequential
 from ..transforms.transforms import _RandomNCrop
 from .geo import NonGeoDataModule
-from .utils import dataset_split
 
 
 class DeepGlobeLandCoverDataModule(NonGeoDataModule):
--- a/torchgeo/datamodules/deepglobelandcover.py
+++ b/torchgeo/datamodules/deepglobelandcover.py
@@ -6,7 +6,7 @@
 from typing import Any
 
 import kornia.augmentation as K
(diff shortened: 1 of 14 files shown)


adamjstewart:
Could be made simpler by allowing the user to input `lengths` directly, so they can choose how to define the split (percentages, numbers, etc.). Wouldn't be backwards compatible, but would also resolve #1289. However, many GeoDataModules already accept a `length` parameter, so I'm not sure what to call this. Maybe `split_lengths`?

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -59,8 +60,9 @@ def setup(self, stage: str) -> None:
         """
         if stage in ["fit", "validate"]:
             self.dataset = DeepGlobeLandCover(split="train", **self.kwargs)
-            self.train_dataset, self.val_dataset = dataset_split(
-                self.dataset, self.val_split_pct
+            generator = torch.Generator().manual_seed(0)
```
isaaccorley, line 63: Should we be setting the random seed manually? Seems like this would conflict if a user is already setting a global random seed on purpose to evaluate random runs

adamjstewart: In the case a user runs `torchgeo fit` followed by `torchgeo validate`, we need to ensure that the same split is chosen both times. This is even more important for datamodules like NASA Marine Debris that choose train/val/test all at once. That's not to say that we can't or shouldn't add a `seed` parameter to all datamodules. I would be open to a PR that does this (so far this PR just preserves the current behavior).

isaaccorley: Ahh that's right we were already setting the seed previously. Carry on

adamjstewart:
Let's merge this as is. We can think about renaming (lengths) or adding (random seed) parameters in a separate PR.

PR #2015, opened 2024-04-20
RasterDataset: add control over resampling algorithm

### LINKED ISSUE #2012: Overrideable resample property for IntersectionDataset
### Summary

`IntersectionDataset` performs `nearest` resampling by default which does not always give a desirable result. This FR is to add the ability to override this behaviour

### Rationale

_No response_

### Implementation

Add an overrideable `resample` property (similar to `dtype`) that changes resampling to bilinear for float32 and nearest for long. Then pass this into `rasterio.merge.merge` [here](https://github.com/microsoft/torchgeo/blob/8a0cc1ffc41cac7d3d2c3940789bde8757b297be/torchgeo/datasets/geo.py#L558)

### Alternatives

_No response_

### Additional information

_No response_

adamjstewart:
By default, `rasterio.merge.merge` uses "nearest" as its resampling algorithm. This is fast and works well for masks where you don't want to interpolate two categorical classes, but results in resampling artifacts for floating point images. This PR adds an attribute to control the resampling algorithm used. It defaults to "bilinear" for float (usually images) and "nearest" for int (usually masks). 

I chose to base this on `dtype` instead of `is_image` because there are times when you want to interpolate floating point pixelwise regression masks.

@robmarkcole can you upload before and after plots from the example you showed me?

Closes #2012


torchgeo/datasets/geo.py
```diff
@@ -384,6 +385,23 @@ def dtype(self) -> torch.dtype:
...
+
+        Defaults to cubic for float dtypes and nearest for int dtypes.
+
+        Returns:
+            The resampling method to use.
+
+        .. versionadded:: 0.6
+        """
+        # Based on torch.is_floating_point
+        if self.dtype in [torch.float64, torch.float32, torch.float16, torch.bfloat16]:
```
adamjstewart, line 400: `torch.double == torch.float64`, and the tests are designed to confirm that both work.

adamjstewart:
Doesn't seem to have as big of an impact on [I/O performance](https://torchgeo.readthedocs.io/en/latest/user/contributing.html#i-o-benchmarking) as I expected:

|  | raw (random) | raw (grid) | preprocessed (random) | preprocessed (grid) |
| -------------: | -------------: | ---: | ---: | ---: |
| before | 17.350 | 10.984 | 9.9158 | 4.6496 |
| after | 18.706 | 12.371 | 9.5933 | 4.5972 |

adamjstewart:
This approach doesn't work well for datasets like L7 Irish or L8 Biome where a single `RasterDataset` is used for both image and mask. I think the solution is to instead subclass `IntersectionDataset` like I did in #1972.

DimitrisMantas:
I think cubic interpolation is not such a good idea because the corresponding convolution kernel has negative weights, meaning that the output data range is not guaranteed.

This can really mess up any normalization or scaling applied to the data before the resampling step.

Here’s a minimal example of what I’m talking about:

```python
import numpy as np
import rasterio

dst_data = data = np.random.rand(512, 512)

dst: rasterio.io.DatasetWriter
with rasterio.open(
    "temp.tif", mode="w", count=1, dtype=np.float32, width=512, height=512
) as dst:
    dst.write(dst_data, indexes=1)

src: rasterio.io.DatasetReader
with rasterio.open("temp.tif") as src:
    src_data = src.read(
        out_shape=(src.count, int(src.height * 2), int(src.width * 2)),
        resampling=rasterio.enums.Resampling.cubic,
    )
    assert (
        src_data.min() >= 0 and src_data.max() < 1
    ), "The cubic kernel has negative weights!"

```

This can also cause issues with no-data values; there may be huge inaccuracies in the final product depending on where they land in relation to the kernel.

I think bilinear interpolation is fine for 99.9% of cases.

robmarkcole:
This pair shows the original issue, addressed in this PR

image

adamjstewart:
You can play around with other enums and see which looks best: https://rasterio.readthedocs.io/en/stable/api/rasterio.enums.html#rasterio.enums.Resampling

I agree with @DimitrisMantas that we should pick a safe/simple/fast default.

DimitrisMantas:
I think this is more or less the one-liner for each method:

- Nearest/Mode: Good for masks; not suitable for continuous fields (i.e., images)
- Bilinear: Suitable for continuous data; works best with smooth-ish fields (e.g., DEMs & DSMs)
-  Cubic/Cubic Spline/Lanczos: These are all the same as far as we are concerned; they are a bit unpredictable.
- Average: It sounds sounds a bit like bilinear, but it's probably a kernel.
- Gauss: Produces a probably undesirable smoothing effect when upsampling images.
- Min/Max/etc.: They are probably useful in certain regression tasks, but I can't think of a potential application at the moment.

DimitrisMantas:
By the way, Lanczos is technically the best of the bunch...

isaaccorley:
Agree with ^. It should be NN resampling for masks and Bilinear for all else by default. If a user wants something specific they can override.

Moritz-Langer:
Hello,
I was stumbling upon this PR because of some errors created by the default resampling from a RasterDataset and the process of overriding it. This might be of help for torchgeo in general, depending on the design choices on how to interact with the module.

I was hoping to override resampling like follows:
```

raster_path = "path_to.tif"

raster_dataset = RasterDataset(raster_path)
raster_dataset.resampling = "nearest"
```

However this results in `AttributeError: property 'resampling' of 'RasterDataset' object has no setter`.

Out of curiosity, how should I interact with it?
What I ended up doing is the following: 

```
class CustomRasterDataset(GeoDataset):
    @property
    def resampling(self) -> Resampling:
        """
        Resampling algorithm used when reading input files.
        Returns user-specified method if set, otherwise defaults based on dtype.
        """
        # 1. Check if user set a specific value
        if self._resampling is not None:
            return self._resampling
        
        # 2. Fallback to default behavior (Original Logic)
        if self.dtype in [torch.float64, torch.float32, torch.float16, torch.bfloat16]:
            return Resampling.bilinear
        else:
            return Resampling.nearest

    @resampling.setter
    def resampling(self, value: str | Resampling):
        """Setter for resampling method."""
        # Convert string to Resampling Enum if necessary
        if isinstance(value, str):
            try:
                # getattr(Resampling, "nearest") -> Resampling.nearest
                value = getattr(Resampling, value.lower())
            except AttributeError:
                raise ValueError(f"Invalid resampling method: {value}. Must be 'nearest', 'bilinear', etc.")
        
        if not isinstance(value, Resampling):
             raise ValueError("Value must be a rasterio.enums.Resampling member or valid string.")

        # Store the value
        self._resampling = value
        print(f"Resampling set to: {self._resampling}")

    def __init__(
        self,
        crs: CRS | None = None,
        res: float | None = None,
        bands: Sequence[str] | None = None,
        cache: bool = True,
        resampling: Resampling = None # <--- Add this argument
    ) -> None:
        
        # 1. Initialize internal variable BEFORE calling super().__init__ if possible,
        # or rely on default until set.
        self._resampling = resampling

# Init and other logic

```

adamjstewart:
Hi @Moritz-Langer!

`RasterDataset` is an abstract base class and is not intended for direct use. I don't _think_ we have any docs or tutorials suggesting the above usage, but I could be wrong.

The correct way to change resampling algorithms is via a simple subclass:
```python

class CustomRasterDataset(GeoDataset):
    resampling = "nearest"

raster_path = "path_to.tif"
raster_dataset = CustomRasterDataset(raster_path)
```
This is semi-documented in https://torchgeo.readthedocs.io/en/stable/tutorials/custom_raster_dataset.html, but could probably be improved. Feel free to submit a PR if you have any ideas how to make this more clear in the docs!

Moritz-Langer:
Thank you for the instant reply, I will add this example in the docs and submit a PR.

PR #2018, opened 2024-04-22
Prettier formatter

### LINKED ISSUE #2004: Add YAML formatter
### Summary

We should add a YAML formatter to CI

### Rationale

Indentation is inconsistent, and I frequently have to remind people to add EOL chars before EOF chars.

### Implementation

Bunch of options:

* [yamlfmt](https://github.com/google/yamlfmt)
* [yamlfix](https://github.com/lyz-code/yamlfix)
* [prettier](https://github.com/prettier/prettier)
* [pretty-yaml](https://github.com/mk-fg/pretty-yaml)
* and many more

### Alternatives

_No response_

### Additional information

_No response_

Domejko:
Closes #2004


Domejko:

tests/data/agb_live_woody_density/Aboveground_Live_Woody_Biomass_Density.geojson
```diff
@@ -1 +1,32 @@
-{"type": "FeatureCollection", "name": "Aboveground_Live_Woody_Biomass_Density", "crs": {"type": "name", "properties": {"name": "urn:ogc:def:crs:OGC:1.3:CRS84"}}, "features": [{"type": "Feature", "properties": {"tile_id": "00N_000E", "Mg_px_1_download": "tests/data/agb_live_woody_density/00N_000E.tif", "ObjectId": 1, "Shape__Area": 1245542622548.87, "Shape__Length": 4464169.76558139}, "geometry": {"type": "Polygon", "coordinates": [[[0.0, 0.0], [10.0, 0.0], [10.0, -10.0], [0.0, -10.0], [0.0, 0.0]]]}}]}
```
adamjstewart, line 1: These files are auto-generated. If we update these in this PR, running `data.py` again will change the file. Is there an option to ignore all `tests/data/*` directories?

adamjstewart (commented):
Vast majority of these changes look good to me. I left comments on the remaining ones that concern me. We'll also need to mention prettier in `docs/user/contributing.rst` and add CI in `.github/workflows/style.yaml`, but I can do those if you don't feel comfortable. I wish prettier supported rst and ipynb, but it already supports a lot more formats than I expected, so I'm happy to see that.

.github/SECURITY.md
```diff
@@ -4,27 +4,27 @@
 
 Microsoft takes the security of our software products and services seriously, which includes all source code repositories managed through our GitHub organizations, which include [Microsoft](https://github.com/Microsoft), [Azure](https://github.com/Azure), [DotNet](https://github.com/dotnet), [AspNet](https://github.com/aspnet), [Xamarin](https://github.com/xamarin), and [our GitHub organizations](https://opensource.microsoft.com/).
 
-If you believe you have found a security vulnerability in any Microsoft-owned repository that meets [Microsoft's definition of a security vulnerability](https://docs.microsoft.com/en-us/previous-versions/tn-archive/cc751383(v=technet.10)), please report it to us as described below.
+If you believe you have found a security vulnerability in any Microsoft-owned repository that meets [Microsoft's definition of a security vulnerability](<https://docs.microsoft.com/en-us/previous-versions/tn-archive/cc751383(v=technet.10)>), please report it to us as described below.
```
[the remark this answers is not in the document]
Domejko: Me neither, especially that it's done only on one link.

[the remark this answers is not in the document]
adamjstewart: I'm guessing it's because the link contains parentheses, but I don't know where to find evidence of this.

experiments/ssl4eo/landsat/conf/ssl4eo_benchmark_etm_sr_cdl.yaml
```diff
@@ -22,4 +22,24 @@ data:
     root: "data/ssl4eo_benchmark"
     sensor: "etm_sr"
     product: "cdl"
-    classes: [0, 1, 5, 24, 36, 37, 61, 111, 121, 122, 131, 141, 142, 143, 152, 176, 190, 195]
+    classes:
```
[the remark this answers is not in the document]
Domejko: Unfortunately this is the way how Prettier is handling lists. This topic was discussed with authors already 5-6 years ago and there are still no changes, but there are 3 solutions:
- adding `// prettier-ignore` on the lines which would be quite troublesome 
- adding [prettier-plugin-multiline-arrays](https://github.com/electrovir/prettier-plugin-multiline-arrays) plugin to Prettier, which is a better solution, but it requires installing this plugin
- last one is to leave it as it is if there is not too many cases with such a long arrays that could be hard to read

package-lock.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "torchgeo",
+  "version": "0.5.2",
```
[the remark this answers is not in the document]
Domejko: no from what I have checked they are not necessary. I will remove then `"version"` field

tests/data/agb_live_woody_density/Aboveground_Live_Woody_Biomass_Density.geojson
Domejko: Yes we can specify in `.prettierignore` what we would like to ignore.

Domejko:
> We'll also need to mention prettier in `docs/user/contributing.rst` and add CI in `.github/workflows/style.yaml`

In `docs/user/contributing.rst` I assume that you would like to just mention that Prettier is used in CI to keep the files format consistent. 
As it comes to `.github/workflows/style.yaml` it have been already added to CI but due to the size and number of changes in this PR you could missed it. Let me know are you satisfied with this implementation in CI or does some changes have to be made.

package-lock.json
```diff
@@ -0,0 +1,30 @@
...
+    "": {
+      "name": "torchgeo",
+      "version": "0.5.2",
+      "license": "MIT",
+      "dependencies": {
+        "prettier": "^3.2.5"
+      }
+    },
+    "node_modules/prettier": {
+      "version": "3.2.5",
```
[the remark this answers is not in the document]
Domejko: Added `npm` to `.github/dependabot.yml` with a weekly interval.

experiments/ssl4eo/landsat/conf/ssl4eo_benchmark_etm_sr_cdl.yaml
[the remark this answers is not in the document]
adamjstewart: If you format it this way manually, will prettier reformat it back to square brackets?

.github/workflows/style.yaml
```diff
@@ -2,65 +2,84 @@ name: style
...
+          ruff check --output-format=github --no-fix .
+          ruff format --diff
+  prettier:
+    name: prettier
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone repo
+        uses: actions/checkout@v4.1.2
+      - name: Set up nodejs
+        uses: actions/setup-node@v4
```
adamjstewart, line 71:
```suggestion
        uses: actions/setup-node@v4.0.2
```

.github/workflows/style.yaml
```diff
@@ -2,65 +2,84 @@ name: style
...
+        with:
+          node-version: "20"
+          cache: "npm"
```
adamjstewart, line 74: Does this actually help? I noticed that cache for setup-python didn't make installation any faster.

adamjstewart:
> In `docs/user/contributing.rst` I assume that you would like to just mention that Prettier is used in CI to keep the files format consistent.

And also how to use prettier yourself. Basically, add it to the "Linters" section and follow a similar structure as ruff/mypy.

experiments/ssl4eo/landsat/conf/ssl4eo_benchmark_etm_sr_cdl.yaml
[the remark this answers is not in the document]
Domejko: Yes, every time it will go through CI it will be set to one-liner. Only solution to keep them in multi-line format as I mention is to use `prettier-ignore` above those lines or install a plugin.

.github/workflows/style.yaml
Domejko: Since it's already quite short I doubt that it will give any significant speedup, but haven't testes it either way, I just did noticed that you have used cache for python so I thought that on this one you will also wanna have it.

Some change will make using `npm ci` with cache instead of `npm install --save-dev --save-exact prettier`, in `npm` documentation it suggested that it have cut time by half but it was case where 150 dependencies have been installed.

.github/workflows/style.yaml
Domejko: Added

experiments/ssl4eo/landsat/conf/ssl4eo_benchmark_etm_sr_cdl.yaml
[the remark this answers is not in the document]
adamjstewart: Are you sure? `tests/conf/seco_byol_1.yaml` is formatted as a list without square brackets.

.github/workflows/style.yaml
```diff
@@ -2,65 +2,84 @@ name: style
...
+      - name: Installing prettier
+        run: |
+          npm ci
+          npm cache clean --force
+      - name: List npm dependencies
+        run: npm ls --all
+      - name: Run prettier formatting
+        run: npx prettier . --write
```
adamjstewart, line 82:
```suggestion
        run: npx prettier . --check
```
Write is used to format, check is used to confirm that everything looks good. The former is used by users, but the latter is used in CI.

adamjstewart:
We should also add a pre-commit hook: https://prettier.io/docs/en/precommit#option-2-pre-commithttpsgithubcompre-commitpre-commit

experiments/ssl4eo/landsat/conf/ssl4eo_benchmark_etm_sr_cdl.yaml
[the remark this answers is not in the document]
Domejko: I think we had a misunderstanding but now I exactly get what you mean.

Yes if the file will be initially formatted like:
```yaml
classes:
   - 1
   - 2
   - 3
```
It will stay that way. I assume that `experiments/ssl4eo/landsat/conf/ssl4eo_benchmark_etm_sr_cdl.yaml` was formatted this way due to maximum line length.

I see that it's one of the flaws of Prettier that it will not unify such cases because if it will be initially formatted in this way:
```yaml
classes:
   - 1
   - 2
   - 3
```
it will stay that way, and if it will be in such format:
```yaml
[0, 11, 21, 22, 31, 41, 42, 43, 52, 71, 81, 82, 90, 95]
```
it will also gonna be keep that way unless it will exceed max line length which can be adjusted in `.toml` file then it will look like `experiments/ssl4eo/landsat/conf/ssl4eo_benchmark_etm_sr_cdl.yaml`

.github/workflows/style.yaml
Domejko: True, my mistake.

package-lock.json
adamjstewart, line 1: Are we missing a `package.json` file? CI seems to be failing without it.

.prettierignore
```diff
@@ -0,0 +1,2 @@
+# Ignore artifacts:
+tests/data/*
```
adamjstewart, line 2:
```suggestion
tests/data/*/**
```
This will keep changes to `tests/data/README.md` but skip all the sub dirs. This is also what we use in `.gitattributes`.

P.S. You still need to undo changes to these files in this PR.

package-lock.json
adamjstewart: Can these files be moved into `requirements/`? Then we'll have less clutter in the root directory.

Domejko: As I read `npm ci` documentation it was saying that it *must* have `package-lock.json` and didn't mention `package.json` so I was hoping that it will work with out it. I will add then `package.json`.

Domejko: `package.json` have been added and files have been moved to `requirements/` folder

Domejko:
I did revert changes made in `tests/data` directory. Added Prettier as a hook in `.pre-commit-config.yaml` with exception regex for folders inside `tests/data`. Also updated `docs/user/contributing.rst` on how to install and use Prettier.

All requested changes have been made, so if there are no additional remarks, changes/implements I think it's good to go.

adamjstewart:
Current test failure was fixed in main, if you update your branch it should pass.

Domejko:
Yes it already have passed one in [Reverted changes in data folder](https://github.com/microsoft/torchgeo/pull/2018/commits/c03e0e96a42362d2ac2d7bcf03fa3095df03af02) but I have used `git add .` and my `.idea` folder have been added so I've added `.idea` to `.gitignore` removed `.idea` form PR and next test have failed.

adamjstewart:
Does your branch include https://github.com/microsoft/torchgeo/pull/2024? That should prevent CI from installing pytest from the cache on macOS.

Domejko:
No it doesn't. I will update it now.

.github/dependabot.yml
```diff
@@ -29,3 +29,9 @@ updates:
         versions: ">=6"
       # segmentation-models-pytorch pins timm, must update in unison
       - dependency-name: "timm"
+      # torchvision pins torch, must update in unison
+      - dependency-name: "torch"
```
adamjstewart, line 33: These two lines can be removed

.github/dependabot.yml
```diff
@@ -29,3 +29,9 @@ updates:
...
+  - package-ecosystem: "npm"
+    directory: "/"
```
adamjstewart, line 35:
```suggestion
    directory: "/requirements"
```

.github/workflows/labeler.yml
```diff
@@ -1,16 +1,16 @@
...
     runs-on: ubuntu-latest
     steps:
-    - name: Clone repo
-      uses: actions/checkout@v4.1.4
-    - name: Add label
-      uses: actions/labeler@v5.0.0
-      with:
-        sync-labels: true
+      - name: Clone repo
+        uses: actions/checkout@v4.1.2
```
adamjstewart, line 12:
```suggestion
        uses: actions/checkout@v4.1.4
```

.github/workflows/release.yaml
```diff
@@ -2,92 +2,92 @@ name: release
...
-        pip cache purge
-    - name: List pip dependencies
-      run: pip list
-    - name: Run pytest checks
-      run: |
-        pytest --cov=torchgeo --cov-report=xml --durations=10
-        python -m torchgeo --help
-        torchgeo --help
+      - name: Clone repo
+        uses: actions/checkout@v4.1.2
```
adamjstewart, line 15:
```suggestion
        uses: actions/checkout@v4.1.4
```

.github/workflows/release.yaml
```diff
@@ -2,92 +2,92 @@ name: release
...
-        pip cache purge
-    - name: List pip dependencies
-      run: pip list
-    - name: Run integration checks
-      run: |
-        pytest -m slow --durations=10
-        python -m torchgeo --help
-        torchgeo --help
+      - name: Clone repo
+        uses: actions/checkout@v4.1.2
```
adamjstewart, line 43:
```suggestion
        uses: actions/checkout@v4.1.4
```

.github/workflows/release.yaml
```diff
@@ -2,92 +2,92 @@ name: release
...
-      if: steps.cache.outputs.cache-hit != 'true'
-      run: |
-        pip install .[docs,tests] planetary_computer pystac
-        pip cache purge
-    - name: List pip dependencies
-      run: pip list
-    - name: Run notebook checks
-      run: pytest --nbmake --durations=10 docs/tutorials
+      - name: Clone repo
+        uses: actions/checkout@v4.1.2
```
adamjstewart, line 71:
```suggestion
        uses: actions/checkout@v4.1.4
```

.github/workflows/style.yaml
```diff
@@ -2,65 +2,85 @@ name: style
...
-      if: steps.cache.outputs.cache-hit != 'true'
-      run: |
-        pip install -r requirements/required.txt -r requirements/datasets.txt -r requirements/style.txt -r requirements/tests.txt
-        pip cache purge
-    - name: List pip dependencies
-      run: pip list
-    - name: Run mypy checks
-      run: mypy .
+      - name: Clone repo
+        uses: actions/checkout@v4.1.2
```
adamjstewart, line 17:
```suggestion
        uses: actions/checkout@v4.1.4
```

.github/workflows/style.yaml
```diff
@@ -2,65 +2,85 @@ name: style
...
-        pip install -r requirements/style.txt
-        pip cache purge
-    - name: List pip dependencies
-      run: pip list
-    - name: Run ruff checks
-      run: |
-        ruff check --output-format=github --no-fix .
-        ruff format --diff
+      - name: Clone repo
+        uses: actions/checkout@v4.1.2
```
adamjstewart, line 42:
```suggestion
        uses: actions/checkout@v4.1.4
```

.github/workflows/style.yaml
```diff
@@ -2,65 +2,85 @@ name: style
...
+      - name: Run ruff checks
+        run: |
+          ruff check --output-format=github --no-fix .
+          ruff format --diff
+  prettier:
+    name: prettier
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone repo
+        uses: actions/checkout@v4.1.2
```
adamjstewart, line 69:
```suggestion
        uses: actions/checkout@v4.1.4
```

.github/workflows/tests.yaml
```diff
@@ -17,88 +17,88 @@ jobs:
...
-    - name: Run pytest checks
-      run: |
-        pytest --cov=torchgeo --cov-report=xml --durations=10
-        python3 -m torchgeo --help
-    - name: Report coverage
-      uses: codecov/codecov-action@v4.3.0
-      with:
-        token: ${{ secrets.CODECOV_TOKEN }}
+      - name: Clone repo
+        uses: actions/checkout@v4.1.2
```
adamjstewart, line 23:
```suggestion
        uses: actions/checkout@v4.1.4
```

.github/workflows/tests.yaml
```diff
@@ -17,88 +17,88 @@ jobs:
...
```
adamjstewart, line 70:
```suggestion
        uses: actions/checkout@v4.1.4
```

.github/workflows/tutorials.yaml
```diff
@@ -2,40 +2,40 @@ name: tutorials
...
-      if: steps.cache.outputs.cache-hit != 'true'
-      run: |
-        pip install -r requirements/required.txt -r requirements/docs.txt -r requirements/tests.txt planetary_computer pystac
-        pip cache purge
-    - name: List pip dependencies
-      run: pip list
-    - name: Run notebook checks
-      run: pytest --nbmake --durations=10 docs/tutorials
+      - name: Clone repo
+        uses: actions/checkout@v4.1.2
```
adamjstewart, line 19:
```suggestion
        uses: actions/checkout@v4.1.4
```

.gitignore
```diff
@@ -43,6 +43,8 @@ share/python-wheels/
 .installed.cfg
 *.egg
 MANIFEST
+requirements/node_modules
+.idea
```
adamjstewart, line 47: I prefer not to add text editor- or OS-specific ignores since there are thousands and it isn't clear which to add. Users should add those themselves in their global gitignore.

.pre-commit-config.yaml
```diff
@@ -33,3 +33,12 @@ repos:
...
         exclude: (build|data|dist|logo|logs|output)/
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v3.1.0
+    hooks:
+      - id: prettier
+        language: node
+        language_version: 22.0.0
+        args:
+          - --write
+        exclude: ^data\/[^\/]+\/$
```
adamjstewart, line 44:
```suggestion
        exclude: ^tests/data/[^/]+/$
```

.prettierignore
```diff
@@ -0,0 +1,2 @@
+# Ignore artifacts:
+tests/data/*/**
```
adamjstewart, line 2: Can you add an EOL char before the EOF char?

docs/user/contributing.rst
```diff
@@ -110,6 +111,23 @@ Mypy won't fix your code for you, but will warn you about potential issues with
 
 If you've never used mypy before or aren't familiar with `Python type hints <https://docs.python.org/3/library/typing.html>`_, this check can be particularly daunting. Don't hesitate to ask for help with resolving any of these warnings on your pull request.
 
+Prettier is a code formatter that helps to ensure consistent code style across a project. It supports various languages. Follow these steps to install Prettier module:
```
adamjstewart, line 114:
```suggestion
Prettier is a code formatter that helps to ensure consistent code style across a project. It supports various languages. Follow these steps to install Prettier:
```

docs/user/contributing.rst
```diff
@@ -110,6 +111,23 @@ Mypy won't fix your code for you, but will warn you about potential issues with
...
+
+1. Install Node.js: Prettier is a Node.js module, so you need to have Node.js installed on your system. You can download and install Node.js from `Node.js official website <https://nodejs.org/en>`_.
```
adamjstewart, line 116:
```suggestion
1. Install Node.js: Prettier is a Node.js module, so you need to have Node.js installed on your system. You can download and install Node.js from the `Node.js official website <https://nodejs.org/en>`_.
```

requirements/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "torchgeo",
```
adamjstewart, line 2: I'm curious how many of these values are required. Here we're only using this file to track dependencies. We aren't going to publish torchgeo on npm.

adamjstewart (commented):
Looks like there were a bunch of main branch merges that undid other PRs.

experiments/ssl4eo/landsat/conf/ssl4eo_benchmark_etm_sr_cdl.yaml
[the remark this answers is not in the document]
adamjstewart: In that case, can we format these files like:
```yaml
classes:
   - 1
   - 2
   - 3
```

.github/SECURITY.md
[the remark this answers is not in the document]
adamjstewart: Only documentation I found on this is https://www.markdownguide.org/basic-syntax/#link-best-practices. Can you try %28 and %29 instead of ( and )? Then the angle brackets shouldn't be required.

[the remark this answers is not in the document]
adamjstewart: Other suggestions and an overview of the problem can be found here: https://meta.stackexchange.com/questions/13501/links-to-urls-containing-parentheses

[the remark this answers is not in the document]
adamjstewart:
```suggestion
If you believe you have found a security vulnerability in any Microsoft-owned repository that meets [Microsoft's definition of a security vulnerability](https://docs.microsoft.com/en-us/previous-versions/tn-archive/cc751383%28v=technet.10%29), please report it to us as described below.
```

experiments/ssl4eo/landsat/conf/ssl4eo_benchmark_etm_sr_cdl.yaml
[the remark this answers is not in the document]
Domejko: Yes but it all will have to be done manually, so if it will get some new files and changes formatted this way:
```yaml
[0, 11, 21, 22, 31, 41, 42, 43, 52, 71, 81, 82, 90, 95]
```
it will stay that way.

I already started manually changing files.

.github/SECURITY.md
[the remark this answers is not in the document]
Domejko: This solution have worked.

.github/workflows/style.yaml
```diff
@@ -2,65 +2,85 @@ name: style
...
+        with:
+          node-version: "20"
+          cache: "npm"
+          cache-dependency-path: "requirements/"
+      - name: Installing prettier
+        run: |
+          npm ci --prefix requirements/
+          npm cache clean --force
+      - name: List npm dependencies
+        run: npm ls --all
```
adamjstewart, line 81: Currently this is showing 0 installed packages.

.github/workflows/style.yaml
```diff
@@ -2,65 +2,85 @@ name: style
...
+        uses: actions/checkout@v4.1.4
+      - name: Set up nodejs
+        uses: actions/setup-node@v4.0.2
+        with:
+          node-version: "20"
+          cache: "npm"
+          cache-dependency-path: "requirements/"
+      - name: Installing prettier
+        run: |
+          npm ci --prefix requirements/
```
adamjstewart, line 78: I'm pretty sure `--prefix` controls the directory where packages are installed, not the directory where package.json can be found.

.github/workflows/style.yaml
```diff
@@ -2,65 +2,85 @@ name: style
...
+          npm cache clean --force
+      - name: List npm dependencies
+        run: npm ls --all
+      - name: Run prettier formatting
+        run: npx prettier . --check
```
adamjstewart, line 83: It looks like the install isn't working, we get the following when we run npx prettier:
```
npm WARN exec The following package was not found and will be installed: prettier@3.2.5
```

.github/workflows/style.yaml
```diff
@@ -2,65 +2,85 @@ name: style
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone repo
+        uses: actions/checkout@v4.1.4
+      - name: Set up nodejs
+        uses: actions/setup-node@v4.0.2
+        with:
+          node-version: "20"
+          cache: "npm"
+          cache-dependency-path: "requirements/"
```
adamjstewart, line 75:
```suggestion
          cache-dependency-path: "requirements/package-lock.json"
```

adamjstewart (commented):
All formatted files look good now, just need to fix a few problems in CI.

Domejko:
That's strange, when I was testing this on my fork I got
```command
Run npx prettier . --check
Checking formatting...
All matched files use Prettier code style!
```
and here again it can't find it and need to install it.

adamjstewart:
What if you run it as `prettier` instead of `npx prettier`? Does that help?

Domejko:
Unfortunately it have to be run through `npx`. The issue is that `npm ci` need to be run in directory where JSON package files are and there is no flag to provide nested path to them. So when we run it in `requirements/` it install `node_modules` there and as we run check command from root it doesn't see modules there. Strange is this that it worked at first on my fork. 

We might try to use `npm ci` with `--prefix` flag and absolute path to project root because when I tried with simple `../` it didn't work. But I think that using `npm install` instead with path to `requirements/` will be the cleanest solution. Installation process with `npm ci` takes 393ms so even if it doubles it will be unnoticeable.

adamjstewart:
I guess in order of priorities, I have:

1. It MUST install a stable version from a lock file, it CAN'T just install the latest version
2. I would prefer that we store these files in `requirements` if we can, but we could put them in the root if that isn't possible
3. I would prefer that we don't install the same thing in CI twice

We could change the working-directory for every step of the pipeline so that installation and running happen in requirements and we pass it `..` to run on the whole project. Kinda ugly, but probably works.

Domejko:
I have used and tested `npm install requirements/` and everything works fine. I had to add to `package.json` `"name": "torchgeo"` since without it git wasn't happy and we could not clone prettier. It looks like this: 
```yaml
      - name: Installing prettier
        run: |
          npm install requirements/
          npm cache clean --force
      - name: List npm dependencies
        run: npm ls --all
      - name: Run prettier formatting
        run: npx prettier . --check
```


Wanted to ask also are you sure you don't want to put `node_modules` to `.gitignore` ? Because when contributors will start to use it not all of them will have it set up as global ignore. It's not IDE specific.

adamjstewart:
`node_modules` should be in `.gitignore`, IDE-specific things should not.

requirements/package.json
```diff
@@ -0,0 +1,7 @@
...
+  "private": "true",
+  "dependencies": {
+    "prettier": "^3.2.5"
```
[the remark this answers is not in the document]
Domejko: We should change in `.pre-commit-config.yaml` flag from `--write` to `--check` since it have version only up to `v3.1.0` because repo have been archived at 11th of April. But I suggest keeping it as a reminder for those who would forget to run `npx prettier . --write` before sending PR.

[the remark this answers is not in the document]
adamjstewart: 3.2.5 is the newest version. I want to know what the oldest version that works for us is. For example, if our formatting relies on features introduced in 3.1, we would say `>=3.1` instead of `^3.2.5`. It's not super critical since it's only used in CI though.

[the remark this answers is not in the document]
Domejko: Yes and they are working already on 4.0.0 since last year so I assume it might come out in near future .Potentially we can use `>=3.0.0` because there are no breaking changes in 3.1.0 and 3.2.0 and as I checked release notes those small changes don't affect formats that we are using.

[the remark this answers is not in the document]
adamjstewart: There are no _documented_ breaking changes. That doesn't mean there aren't any breaking changes 😉 

It would be good to test with 3.0.0, 2.X, etc. and see which versions result in the same formatting. But this is optional.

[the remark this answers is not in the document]
Domejko: True, I will check are there any differences between 3.0.0 and 3.2.5. Below 3.0.0 I think that this would be problematic:
```md
Ignore .gitignored files by default ([#14731](https://github.com/prettier/prettier/pull/14731) by [@fisker](https://github.com/fisker))

Prettier ignores files ignored by .gitignore by default. If you want the old behavior(only ignore files ignored by .prettierignore), use

prettier . --write --ignore-path=.prettierignore
``` 
That was introduced in 3.0.0 and going below would cause formatting all env files IDE specific folders etc.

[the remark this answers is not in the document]
Domejko: I have just checked 3.0.0 vs 3.2.5 and after formatting files with 3.0.0 and then checking it with 3.2.5 all files remained unchanged, so we can use `>=3.0.0`. In that case I will leave `--write` flag in `.pre-commit-config.yaml` and update `package.json`.

There is also one thing, when we run `npm install requirements/` it creates `package.json` and lock in root directory so those two will also have to be added to `.gitignore` as `/*.json` to ignore them only in root.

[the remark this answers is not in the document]
adamjstewart: Could we run `npm install` from inside the `requirements` directory to prevent that? Is `npm install <path>` a common strategy? Maybe we need to move these config files to the root after all...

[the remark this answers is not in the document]
Domejko: Is it common I don't know since I haven't worked on many python/node projects, the one that I'm contributing to recently have them in root directory but it's a web app. At the moment I'm not on my working station so I can't check but I assume that running 'npm install' inside `requirements` folder would cause creating `node_modules` also in that directory but we could use `--prefix` for that. But also in that case this step need to be mentioned in `contributing.rst`.

I think that simplest solution would be best in that case it's either staying with `npm install requirements/` and `/*.json` in `.gitignore,` **or** moving `package.json` and lock too root and using `npm ci` instead.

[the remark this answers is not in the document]
adamjstewart: The files in `requirements` aren't used by contributors, they're only used in CI. The simplest solution is certainly to put everything in root, it just clutters root with stuff that isn't usually needed.

.github/workflows/style.yaml
```diff
@@ -2,65 +2,85 @@ name: style
...
+        uses: actions/checkout@v4.1.4
+      - name: Set up nodejs
+        uses: actions/setup-node@v4.0.2
+        with:
+          node-version: "20"
+          cache: "npm"
+          cache-dependency-path: "requirements/package-lock.json"
+      - name: Installing prettier
+        run: |
+          npm install requirements/
```
adamjstewart, line 78: Shouldn't we be using `npm ci` instead of `npm install`?

docs/user/contributing.rst
```diff
@@ -110,6 +111,23 @@ Mypy won't fix your code for you, but will warn you about potential issues with
...
 If you've never used mypy before or aren't familiar with `Python type hints <https://docs.python.org/3/library/typing.html>`_, this check can be particularly daunting. Don't hesitate to ask for help with resolving any of these warnings on your pull request.
 
+Prettier is a code formatter that helps to ensure consistent code style across a project. It supports various languages. Follow these steps to install Prettier:
+
+1. Install Node.js: Prettier is a Node.js module, so you need to have Node.js installed on your system. You can download and install Node.js from the `Node.js official website <https://nodejs.org/en>`_.
+2. Install Prettier: Use the following command to install the Prettier module in your project:
+
+.. code-block:: console
+
+   $ npm install requirements/
```
adamjstewart, line 121:
```suggestion
   $ npm install prettier
```
Users shouldn't know about `requirements`

requirements/package.json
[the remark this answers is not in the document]
Domejko: But contributors will be installing it by them self and if they won't be instructed to do it the same way as in CI then again `package.json` will be created in root directory and either it will get send to PR or will need to be placed in `.gitignore`. So they would need to be instructed to install it the same way as it is in CI, other way such change wouldn't make sense.

That's why the most optimal solution is current one, where we ignore `/*.json` and install from `requirements/`.

Other solution is to remove part where contributor install and use `prettier` and obligate them to use `pre-commit`. But it's short term if we would like to use newer versions since hook for `prettier` is already archived and I don't know dose version 4.0.0 will have new hook.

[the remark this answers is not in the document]
adamjstewart: I don't see any issues with adding `node_modules` to `.gitignore`. It should ignore that directory regardless of if it's in the root or in `requirements/`.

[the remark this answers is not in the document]
Domejko: But not `node_modules` where an issue but adding `/*.json` to `.gitignore`.



> There is also one thing, when we run `npm install requirements/` it creates `package.json` and lock in root directory so those two will also have to be added to `.gitignore` as `/*.json` to ignore them only in root.

`node_modules` are already added since we discussed that.

.github/workflows/style.yaml
Domejko: No, not when we have `package.json` and lock in `requirements/`. We discussed that before that there is no way to provide a path for `ci` to work.

requirements/package.json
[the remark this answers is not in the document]
adamjstewart: Ah sorry, I misunderstood. So if we replace `npm install requirements` with `npm install prettier` in the docs, we shouldn't have that issue, right?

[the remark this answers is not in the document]
Domejko: I dug some for and found a solution. We can use `npm install prettier --no-save` the `--no-save` flag will tell npm to not create `package.json` and lock files and will make a "clean" install. It would work with `npm install requirements/` as well.

.github/workflows/style.yaml
adamjstewart: That was just an assumption on my part. I don't think `--prefix` specifies the location where the json files are found, it specifies the installation directory. I would assume that `npm install` and `npm ci` support the same features, don't they?

requirements/package.json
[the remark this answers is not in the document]
adamjstewart: If it makes things easier, I'm not that opposed to keeping `package.json` in the root, although I would definitely prefer to keep the lock file in `requirements`.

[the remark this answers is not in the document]
Domejko: **Contributors:**
If we will set up `npm install prettier --no-save` or `npm install requirements/ --no-save` as default installation command we don't have issue with them creating `.json` file in root directory. It will just create `node_modules` and install latest `Prettier` version. So in `.gitignore` will be only included `node_modules`.

**CI:**
At the moment we use there `npm install requirements/` which reads from `.json` files we have set up there. Everything is working fine, CI commands are clean, we get to list dependencies and run check.
If we would like to use `npm ci` we would **have to** place both `.json` files in root directory since `npm ci` must have existing `package-lock.json` and `package-lock.json` need to have existing `package.json` to work.

.github/workflows/style.yaml
Domejko: No `npm install` and `npm ci` have different features. The key difference I have mentioned in conversation above.

requirements/package.json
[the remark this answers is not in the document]
Domejko: We might also try using `npm install prettier --no-save` in CI, if it will work there then we might not need `.json` files in `requirements/` either.

.gitignore
```diff
@@ -82,6 +82,10 @@ db.sqlite3-journal
 instance/
 .webassets-cache
 
+# Node stuff:
```
adamjstewart, line 85: Let's move this section up just below TorchGeo, most everything else in that file is from the default Python .gitignore, so I want to keep it easy to update.

requirements/package.json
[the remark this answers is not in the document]
adamjstewart: We don't want this in CI. In CI, we want to enforce that a specific version of prettier is being installed, and we want dependabot to update this version in its own PRs. This prevents contributor PRs from failing due to new version updates that are unrelated to their PRs.

For contributors, I don't care too much either way. It's in `.gitignore`, so it shouldn't be an issue. But if you want to add that, I'm fine with adding that.

docs/user/contributing.rst
Domejko: So here we will use `npm install prettier --no-save` and will keep `/*.json` in `.gitignore` for case is someone would mess up command or something like that. You agree with this ?

adamjstewart: Yep, sounds good to me

adamjstewart (approved):
Everything looks good to me now!

Thanks for all of your hard work on this. I have exactly 0 years of experience with node/npm, so I'm glad I have the chance to learn a new tool.

I'll merge this and make it a required test in CI. We may play around with the settings in the future, but for now they all seem like good defaults.

Domejko:
My pleasure, I'm happy that I could contribute to this project and too me it was also a very good experience, I have learn new tools that for sure might come in handy in the future.

PR #2034, opened 2024-04-27
L7 Irish: convert to IntersectionDataset

adamjstewart:
Converts L7 Irish to an IntersectionDataset following the design proposed in #1972.

This is needed for #2015 so that images and masks can independently control their resampling algorithm.

~Only remaining bug is that the length of the dataset jumped from 206 to 438, which implies that some images in the dataset overlap.~

Will do the same thing to L8 Biome after this is merged.

@yichiac

### DIFF
--- a/tests/datasets/test_l7irish.py
+++ b/tests/datasets/test_l7irish.py
@@ -66,6 +66,7 @@ class TestL7Irish:
 
     def test_already_extracted(self, dataset: L7Irish) -> None:
         L7Irish(dataset.paths, download=True)
+        L7Irish([dataset.paths], download=True)
 
     def test_already_downloaded(self, tmp_path: Path) -> None:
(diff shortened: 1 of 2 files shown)


adamjstewart:
> Only remaining bug is that the length of the dataset jumped from 206 to 438, which implies that some images in the dataset overlap.

I've got this down to 210. Some files actually overlap in both time and space:

* `austral/p74_r92/L71074092_09220011103.TIF` and `mid_latitude_south/p74_r91/L71074091_09120011103.TIF`
* `austral/p227_r98/L71227098_09820011103.TIF` and `austral/p227_r97/L71227097_09720011103.TIF`

I actually have a different idea, let me try that instead.

adamjstewart:
Okay, now we're back down to 206 like the original dataset. Note that the above pairs of images that overlap may be double-counted. However, this is the same behavior as we had before.

PR #2038, opened 2024-05-01
SatlasPretrain: ResNet50/152 and Swin_V2_T Weights

yichiac:
This PR adds SatlasPretrain ResNet50 weights as the Enums.

@piperwolters @favyen2


adamjstewart (changes requested):
Need to apply all the same changes I requested on https://github.com/microsoft/torchgeo/pull/1884

Also, CI passes because we only use fake model weights, but using the real model weights result in test failures:
```console
> pytest -m slow tests/models/test_resnet.py
...
E           RuntimeError: Attempting to deserialize object on a CUDA device but torch.cuda.is_available() is False. If you are running on a CPU-only machine, please use torch.load with map_location=torch.device('cpu') to map your storages to the CPU.
```

docs/api/landsat_pretrained_weights.csv
```diff
@@ -29,4 +29,5 @@ ResNet50_Weights.LANDSAT_OLI_SR_MOCO,8--9,7,`link <https://github.com/microsoft/
...
 ViTSmall16_Weights.LANDSAT_OLI_SR_MOCO,8--9,7,`link <https://github.com/microsoft/torchgeo>`__,`link <https://proceedings.neurips.cc/paper_files/paper/2023/hash/bbf7ee04e2aefec136ecf60e346c2e61-Abstract-Datasets_and_Benchmarks.html>`__,"CC0-1.0",66.81,50.16,64.17,47.24
 ViTSmall16_Weights.LANDSAT_OLI_SR_SIMCLR,8--9,7,`link <https://github.com/microsoft/torchgeo>`__,`link <https://proceedings.neurips.cc/paper_files/paper/2023/hash/bbf7ee04e2aefec136ecf60e346c2e61-Abstract-Datasets_and_Benchmarks.html>`__,"CC0-1.0",65.04,48.20,62.61,45.46
-Swin_V2_B_Weights.LANDSAT_MS_SI_SATLAS,11,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,"ODC-BY",,,,
+Swin_V2_B_Weights.LANDSAT_SI_SATLAS,8--9,11,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,Apache-2.0,,,,
+Swin_V2_B_Weights.LANDSAT_MI_SATLAS,8--9,11,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,Apache-2.0,,,,
```
adamjstewart, line 33: https://github.com/allenai/satlas/issues/48

torchgeo/models/swin.py
```diff
@@ -46,75 +46,138 @@
...
         transforms=_satlas_transforms,
         meta={
-            'dataset': 'Satlas',
+            'dataset': 'SatlasPretrain',
             'in_chans': 3,
-            'model': 'swin_v2_b',
+            'model': 'swin_v2_t',
+            'publication': 'https://arxiv.org/abs/2211.15660',
+            'repo': 'https://github.com/allenai/satlas',
+            'bands': ('B02', 'B03', 'B04'),
```
adamjstewart, line 80: Can anyone confirm if the band order is RGB or BGR?

torchgeo/models/swin.py
```diff
@@ -155,6 +335,16 @@ def swin_v2_b(
     model: SwinTransformer = torchvision.models.swin_v2_b(weights=None, *args, **kwargs)
 
     if weights:
-        model.load_state_dict(weights.get_state_dict(progress=True), strict=False)
+        num_channels = weights.meta['in_chans']
+        out_channels = model.features[0][0].out_channels
+        # https://github.com/allenai/satlaspretrain_models/blob/main/satlaspretrain_models/models/backbones.py#L27
+        model.features[0][0] = torch.nn.Conv2d(
```
adamjstewart, line 341: This was completely broken for MS models until now.

adamjstewart (dismissed):

torchgeo/transforms/transforms.py
```diff
@@ -272,3 +272,54 @@ def apply_transform(
             out = rearrange(out, 'b t c h w -> (b t) c h w')
 
         return out
+
+
+class _Clamp(K.IntensityAugmentationBase2D):
```
adamjstewart, line 277: @johnnv1 is this something Kornia would be interested in? If so I can contribute it when I have a chance.

torchgeo/models/swin.py
yichiac: According to their [demo](https://github.com/allenai/satlaspretrain_models/blob/main/demo.ipynb), it looks more like RGB. I haven't found a clear definition of TCI order, but it seems to be RGB.
@favyen2 @piperwolters, please confirm.

piperwolters: It should be RGB..
ESA’s Sentinel-2 User Handbook: The TCI product is described as a true color composite using Red (Band 4), Green (Band 3), and Blue (Band 2) bands.

piperwolters: [Satlas Normalization README](https://github.com/allenai/satlas/blob/55bb52fb90a9caaa91bf5fb428e76bfd911a5364/Normalization.md#sentinel-2-images)

adamjstewart: So this is also true for MSI? It would be 4, 3, 2, 5, 6, 7, ...? @yichiac I'm curious if this is the order you used in your experiments or not. Could explain why SatlasPretrain performed worse than SSL4EO-S12 (or it could be the subset of supported bands).

adamjstewart: NAIP too?

adamjstewart:
Are there any published performance metrics for all weights? Would love to add these to the weights tables.

torchgeo/models/swin.py
yichiac: I think TCI's order is 4, 3, 2, and MSI is 2, 3, 4, 5, 6, 7, ... The starting accuracy (acc for the first few epochs) of SatlasPretrain is usually the highest.

piperwolters: NAIP is also RGB. 

I'm pretty sure the MSI is TCI, 5, 6, 7, 8, 11, 12, so 4, 3, 2, 5, 6, 7, 8, 11, 12. 
See [this Sentinel-2 example](https://github.com/allenai/satlas/blob/main/Normalization.md#sentinel-2-example) in Satlas docs.

adamjstewart: @yichiac we should redo our Satlas experiments with the correct band order.

adamjstewart (approved):
Will wait a bit more for further review but it sounds like everything is in order now.

piperwolters:
> Will wait a bit more for further review but it sounds like everything is in order now.

Looking into if we have performance metrics that could fit into the torchgeo tables. Everything else looks good to me. 
Favyen is OOO this week, so could wait for him to have a look as well and 100% confirm band order.

adamjstewart:
> Looking into if we have performance metrics that could fit into the torchgeo tables. Everything else looks good to me.

It doesn't have to be the same benchmark datasets we used, it could be different benchmarks.

> Favyen is OOO this week, so could wait for him to have a look as well and 100% confirm band order.

I'm planning on releasing TorchGeo 0.6.0 by the end of this month, so hopefully he can review by Monday.

adamjstewart:
Going to merge so we can finalize the release soon. @favyen2 can review later if he has time.

torchgeo/transforms/transforms.py
johnnv1: I think so, not sure why it will be a augmentation. But i think it's fine to add to kornia if have uses for it. Just to make sure, @shijianjian we don't have it, right?

adamjstewart: The use case is that this comes up commonly in satellite imagery pipelines where images aren't always uint8 and need to be converted to uint8 (0–255) or normalized (0–1) without out-of-bounds data. See [SeCo](https://github.com/ServiceNow/seasonal-contrast/blob/8285173ec205b64bc3e53b880344dd6c3f79fa7a/datasets/seco_dataset.py#L53) and [Satlas](https://github.com/allenai/satlas/blob/main/Normalization.md#sentinel-2-images) for examples where this is used.

PR #2045, opened 2024-05-04
CI: test optional datasets on every commit

adamjstewart:
TorchGeo has a number of optional dependencies only needed by certain datasets. Until now, we've only ensured that all tests pass without these dependencies installed on release branches. However, this makes it difficult to get full coverage without relying on ugly import monkeypatching. It's simpler and easier to test both on every commit to get full coverage. We already do this for the minimum version of our dependencies.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -7,34 +7,6 @@ on:
     branches:
       - release**
 jobs:
-  datasets:
-    name: datasets
-    runs-on: ubuntu-latest
-    steps:
-      - name: Clone repo
-        uses: actions/checkout@v4.1.4
-      - name: Set up python
-        uses: actions/setup-python@v5.1.0
-        with:
-          python-version: "3.12"
-      - name: Cache dependencies
-        uses: actions/cache@v4.0.2
-        id: cache
-        with:
-          path: ${{ env.pythonLocation }}
-          key: ${{ env.pythonLocation }}-${{ hashFiles('pyproject.toml') }}-datasets
-      - name: Install pip dependencies
-        if: steps.cache.outputs.cache-hit != 'true'
-        run: |
-          pip install .[tests]
-          pip cache purge
-      - name: List pip dependencies
-        run: pip list
-      - name: Run pytest checks
-        run: |
-          pytest --cov=torchgeo --cov-report=xml --durations=10
-          python -m torchgeo --help
-          torchgeo --help
   integration:
     name: integration
     runs-on: ubuntu-latest
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -99,6 +99,39 @@ jobs:
         uses: codecov/codecov-action@v4.3.0
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+  datasets:
+    name: datasets
+    runs-on: ubuntu-latest
+    env:
+      MPLBACKEND: Agg
+    steps:
+      - name: Clone repo
+        uses: actions/checkout@v4.1.4
+      - name: Set up python
+        uses: actions/setup-python@v5.1.0
+        with:
+          python-version: "3.12"
+      - name: Cache dependencies
+        uses: actions/cache@v4.0.2
+        id: cache
+        with:
+          path: ${{ env.pythonLocation }}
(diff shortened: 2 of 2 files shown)


tests/datamodules/test_usavars.py
```diff
@@ -14,7 +14,6 @@
 class TestUSAVarsDataModule:
     @pytest.fixture
     def datamodule(self, request: SubRequest) -> USAVarsDataModule:
-        pytest.importorskip('pandas', minversion='1.1.3')
```
adamjstewart, line 17: Unrelated to this PR, but this dep is now required

tests/trainers/test_classification.py
```diff
@@ -89,7 +89,7 @@ def test_trainer(
         self, monkeypatch: MonkeyPatch, name: str, fast_dev_run: bool
     ) -> None:
         if name.startswith('so2sat') or name == 'quakeset':
-            pytest.importorskip('h5py', minversion='3')
+            pytest.importorskip('h5py', minversion='3.6')
```
adamjstewart, line 92: We forgot to update the minimum deps when dropping Python 3.9 support

tests/trainers/test_segmentation.py
```diff
@@ -87,12 +87,16 @@ class TestSemanticSegmentationTask:
...
     ) -> None:
-        if name == 'naipchesapeake':
-            pytest.importorskip('zipfile_deflate64')
-
-        if name == 'landcoverai':
-            sha256 = 'ecec8e871faf1bbd8ca525ca95ddc1c1f5213f40afb94599884bd85f990ebd6b'
-            monkeypatch.setattr(LandCoverAI, 'sha256', sha256)
+        match name:
+            case 'chabud':
+                pytest.importorskip('h5py', minversion='3.6')
```
adamjstewart, line 92: This was missing and caused the release branch tests to fail without this fix

torchgeo/datasets/cropharvest.py
```diff
@@ -111,15 +111,7 @@ def __init__(
 
         Raises:
             DatasetNotFoundError: If dataset is not found and *download* is False.
-            ImportError: If h5py is not installed
         """
-        try:
```
adamjstewart, line 116: No need to fail immediately upon instantiation, h5py is only needed for `__getitem__`

torchgeo/datasets/idtrees.py
```diff
@@ -182,13 +181,6 @@ def __init__(
         self.num_classes = len(self.classes)
         self._verify()
 
-        try:
```
adamjstewart, line 185: It wasn't easy to test missing pyvista when laspy is also missing without this change.

tests/datasets/test_quakeset.py
```diff
@@ -27,6 +27,7 @@ class TestQuakeSet:
     def dataset(
         self, monkeypatch: MonkeyPatch, tmp_path: Path, request: SubRequest
     ) -> QuakeSet:
+        pytest.importorskip('h5py', minversion='3.6')
```
adamjstewart, line 30: This was missing and caused the tests to fail on the release branch.

tests/datasets/test_utils.py
```diff
@@ -127,10 +116,6 @@ def download_url(url: str, root: str, *args: str) -> None:
     shutil.copy(url, root)
 
 
-def test_mock_missing_module(mock_missing_module: None) -> None:
```
adamjstewart, line 130: Unit tests for our unit tests. I love it.

torchgeo/datasets/cropharvest.py
calebrob6: I disagree, without h5py instantiation won't be useful? I'd rather find out immediately that my dataset won't work.

torchgeo/datasets/idtrees.py
calebrob6: Same here -- I think the dataset should fail to instantiate if the deps aren't installed

adamjstewart: I think I have an idea that will allow us to fail-fast and still get full coverage, will open another prereq PR with just that change so that this PR can be a test-only PR.

torchgeo/datasets/cropharvest.py
adamjstewart: It does allow you to download/extract the dataset and plot images, just not load images. I get your point though. How long of a delay is there in realizing that the dataset doesn't work if we don't error until getitem?

PR #2051, opened 2024-05-07
Avoid use of __all__

adamjstewart:
In TorchGeo, we use `__all__` to prevent "unused import" errors from ruff (formerly flake8). These are only needed in `__init__.py` files to support import aliasing. However, this has led to the proliferation of `__all__` in a few other places. This PR cleans those up.

### DIFF
--- a/torchgeo/datasets/utils.py
+++ b/torchgeo/datasets/utils.py
@@ -27,26 +27,8 @@ from torch.utils.data import Dataset
 from torchvision.datasets.utils import check_integrity, download_url
 from torchvision.utils import draw_segmentation_masks
 
-__all__ = (
-    'check_integrity',
-    'DatasetNotFoundError',
-    'RGBBandsMissingError',
-    'download_url',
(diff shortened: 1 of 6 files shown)


torchgeo/datasets/utils.py
```diff
@@ -27,26 +27,8 @@
...
-    'merge_samples',
-    'unbind_samples',
-    'rasterio_loader',
-    'sort_sentinel2_bands',
-    'draw_semantic_segmentation_masks',
-    'rgb_to_mask',
-    'percentile_normalization',
-)
+# Only include import redirects
+__all__ = ('check_integrity', 'download_url')
```
adamjstewart, line 31: `torchgeo.datasets.check_integrity` is the same as `torchvision.datasets.check_integrity`, but we alias it here for consistency. We don't really need to do this. But for now, `__all__` is needed to avoid unused import errors.

isaaccorley (approved):

PR #2052, opened 2024-05-07
torch.hub: add support for DOFA and Swin models

adamjstewart:
We forgot to add support for DOFA and Swin ViT to `torch.hub`.

### DIFF
--- a/torchgeo/models/dofa.py
+++ b/torchgeo/models/dofa.py
@@ -417,7 +417,7 @@ class DOFALarge16_Weights(WeightsEnum):  # type: ignore[misc]
     )
 
 
-def dofa_small_patch16_224(**kwargs: Any) -> DOFA:
+def dofa_small_patch16_224(*args: Any, **kwargs: Any) -> DOFA:
     """Dynamic One-For-All (DOFA) small patch size 16 model.
 
     If you use this model in your research, please cite the following paper:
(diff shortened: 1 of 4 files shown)


isaaccorley (approved):

torchgeo/models/dofa.py
```diff
@@ -518,10 +524,12 @@ def dofa_huge_patch16_224(**kwargs: Any) -> DOFA:
...
 
     Args:
+        *args: Additional arguments to pass to :class:`DOFA`.
         **kwargs: Additional keywork arguments to pass to :class:`DOFA`.
 
     Returns:
         A DOFA huge 16 model.
     """
-    model = DOFA(patch_size=14, embed_dim=1280, depth=32, num_heads=16, **kwargs)
+    kwargs |= {'patch_size': 14, 'embed_dim': 1280, 'depth': 32, 'num_heads': 16}
```
calebrob6, line 533: Should we warn the user if these overwrite kwargs?

calebrob6 (approved):

torchgeo/models/dofa.py
adamjstewart: I think it's worth adding, just not in this PR. Torchvision does issue a warning as of https://github.com/pytorch/vision/pull/5618.

PR #2054, opened 2024-05-08
Datasets: improve lazy import error msg for missing deps

adamjstewart:
TorchGeo has a number of optional dependencies only needed by certain datasets. We use lazy imports to allow these dependencies to be optional. This PR improves these error messages to make it more clear how to solve the issue.

### Before

```
ImportError: scipy is not installed and is required to use this dataset
```

### After

```
torchgeo.datasets.errors.DependencyNotFoundError: scipy is not installed and is required to use this dataset. Either run:

$ pip install scipy

to install just this dependency, or:

$ pip install torchgeo[datasets]

to install all optional dataset dependencies.
```

This has a number of other benefits:

* Reduces code duplication (making it easier to update the message for all imports if we want to, e.g., add conda installation instructions)
* Allows us to easily fail fast during dataset instantiation
* Makes it easier to get 100% coverage on all datasets in #2045

Closes #2045 
Prerequisite for #2043

### DIFF
--- a/torchgeo/datasets/utils.py
+++ b/torchgeo/datasets/utils.py
@@ -794,7 +794,7 @@ def lazy_import(name: str) -> Any:
     except ModuleNotFoundError:
         # Map from import name to package name on PyPI
         name = name.split('.')[0].replace('_', '-')
-        module_to_pypi = collections.defaultdict(lambda: name)
+        module_to_pypi: dict[str, str] = collections.defaultdict(lambda: name)
         module_to_pypi |= {'cv2': 'opencv-python', 'skimage': 'scikit-image'}
         name = module_to_pypi[name]
         raise MissingDependencyError(name) from None
--- a/torchgeo/datasets/utils.py
+++ b/torchgeo/datasets/utils.py
@@ -40,13 +40,7 @@ class _rarfile:
             self.kwargs = kwargs
 
         def __enter__(self) -> Any:
-            try:
-                import rarfile
-            except ImportError:
-                raise ImportError(
-                    'rarfile is not installed and is required to extract this dataset'
-                )
-
+            rarfile = lazy_import('rarfile')
             # TODO: catch exception for when rarfile is installed but not
             # unrar/unar/bsdtar
             return rarfile.RarFile(*self.args, **self.kwargs)
@@ -160,14 +154,11 @@ def download_radiant_mlhub_dataset(
         api_key: the API key to use for all requests from the session. Can also be
             passed in via the ``MLHUB_API_KEY`` environment variable, or configured in
             ``~/.mlhub/profiles``.
-    """
-    try:
-        import radiant_mlhub
-    except ImportError:
-        raise ImportError(
-            'radiant_mlhub is not installed and is required to download this dataset'
-        )
 
+    Raises:
+        MissingDependencyError: If radiant_mlhub is not installed.
+    """
+    radiant_mlhub = lazy_import('radiant_mlhub')
     dataset = radiant_mlhub.Dataset.fetch(dataset_id, api_key=api_key)
     dataset.download(output_dir=download_root, api_key=api_key)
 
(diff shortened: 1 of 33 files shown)


torchgeo/datasets/utils.py
```diff
@@ -773,3 +764,25 @@ def array_to_tensor(array: np.typing.NDArray[Any]) -> Tensor:
...
+    """
+    try:
+        return importlib.import_module(name)
+    except ModuleNotFoundError:
+        # Map from import name to package name on PyPI
+        name = name.split('.')[0].replace('_', '-')
+        module_to_pypi: dict[str, str] = collections.defaultdict(lambda: name)
+        module_to_pypi |= {'cv2': 'opencv-python', 'skimage': 'scikit-image'}
+        name = module_to_pypi[name]
+        raise MissingDependencyError(name) from None
```
adamjstewart, line 788: TIL about `raise ... from ...`

### Without

```
> python3 -c 'from torchgeo.datasets.utils import lazy_import; lazy_import("foo")'
  File "/Users/Adam/torchgeo/torchgeo/datasets/utils.py", line 781, in lazy_import
    return importlib.import_module(name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Adam/spack/var/spack/environments/default/.spack-env/view/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1140, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'foo'

During handling of the above exception, another exception occurred:

  File "<string>", line 1, in <module>
  File "/Users/Adam/torchgeo/torchgeo/datasets/utils.py", line 788, in lazy_import
    raise MissingDependencyError(name)
torchgeo.datasets.errors.MissingDependencyError: foo is not installed and is required to use this dataset. Either run:

$ pip install foo

to install just this dependency, or:

$ pip install torchgeo[datasets]

to install all optional dataset dependencies.
```

### With

```
> python3 -c 'from torchgeo.datasets.utils import lazy_import; lazy_import("foo")'
  File "<string>", line 1, in <module>
  File "/Users/Adam/torchgeo/torchgeo/datasets/utils.py", line 788, in lazy_import
    raise MissingDependencyError(name) from None
torchgeo.datasets.errors.MissingDependencyError: foo is not installed and is required to use this dataset. Either run:

$ pip install foo

to install just this dependency, or:

$ pip install torchgeo[datasets]

to install all optional dataset dependencies.
```

torchgeo/datasets/vhr10.py
```diff
@@ -465,7 +459,7 @@ def plot(
             # Add masks
             if show_feats in {'masks', 'both'} and 'masks' in sample:
                 mask = masks[i]
-                contours = find_contours(mask, 0.5)  # type: ignore[no-untyped-call]
+                contours = skimage.measure.find_contours(mask, 0.5)
```
adamjstewart, line 462: Only downside of the lazy import function is that we lose type hints.

torchgeo/datasets/utils.py
```diff
@@ -773,3 +764,25 @@ def array_to_tensor(array: np.typing.NDArray[Any]) -> Tensor:
...
+        name: Name of module to import.
+
+    Raises:
+        MissingDependencyError: If *name* is not installed.
+
+    .. versionadded:: 0.6
+    """
+    try:
+        return importlib.import_module(name)
+    except ModuleNotFoundError:
```
adamjstewart, line 782: Apparently ImportError and ModuleNotFoundError are different:
```pycon
>>> import foo
ModuleNotFoundError: No module named 'foo'
>>> import numpy.foo
ModuleNotFoundError: No module named 'numpy.foo'
>>> from numpy import foo
ImportError: cannot import name 'foo' from 'numpy' (/Users/Adam/spack/var/spack/environments/default/.spack-env/view/lib/python3.11/site-packages/numpy/__init__.py)
```

torchgeo/datasets/utils.py
```diff
@@ -773,3 +764,25 @@ def array_to_tensor(array: np.typing.NDArray[Any]) -> Tensor:
     elif array.dtype == np.uint32:
         array = array.astype(np.int64)
     return torch.tensor(array)
+
+
+def lazy_import(name: str) -> Any:
+    """Lazy import of *name*.
+
+    Args:
+        name: Name of module to import.
```
adamjstewart, line 773: We could also check the version, but that would require an extra dep on `packaging`.

PR #2057, opened 2024-05-11
Scale-MAE model

isaaccorley:
Finally getting around to adding this.

Adds Scale-MAE model (ViT encoder only) and pretrained weights.

I've verified this reproduces KNN performance at different resolutions for UCMerced but will repeat for other datasets.

@RitwikGupta let me know if this looks good. I cleaned up some of the code a bit to work out of the box with our trainers (this required setting the res when initializing the model instead of dynamically but I think it should still be fine).

@calebrob6 lmk if you want to team up on this one.


docs/api/misc_pretrained_weights.csv
```diff
@@ -1,2 +1,3 @@
 Weight,Channels,Source,Citation,License
 ResNet50_Weights.FMOW_RGB_GASSL, 3,`link <https://github.com/sustainlab-group/geography-aware-ssl>`__,`link <https://arxiv.org/abs/2011.09980>`__,-
+ScaleMAE_ViTLarge16_Weights.FMOW_RGB_SCALEMAE, 3,`link <https://github.com/bair-climate-initiative/scale-mae>`__,`link <https://arxiv.org/abs/2212.14532>`__,"CC-BY-NC-4.0"
```
adamjstewart, line 3: Needs EOL char before EOF char

torchgeo/models/scale_mae.py
```diff
@@ -0,0 +1,255 @@
+from collections import OrderedDict
```
adamjstewart, line 1: Needs license and docstring

adamjstewart (commented):
Also see changes in https://github.com/microsoft/torchgeo/pull/2052/files

docs/api/misc_pretrained_weights.csv
adamjstewart, line 1: Two thoughts:

1. I wonder if we should move both of these to the new "Sensor-Agnostic" section because they technically work for (RGB-only) imagery from _any_ sensor
2. Since both of these have evaluation results on fMoW, can we add additional columns with those performance metrics (assuming they are comparable)? If we move them to "Sensor-Agnostic", we may need two tables, one for things evaluated on GEO-Bench and one for things evaluated on fMoW.

docs/api/models.rst
```diff
@@ -50,6 +50,12 @@ ResNet
 .. autoclass:: ResNet18_Weights
 .. autoclass:: ResNet50_Weights
 
+Scale-MAE Vision Transformer
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
adamjstewart, line 54:
```suggestion
Scale-MAE
^^^^^^^^^
```
This is how we named DOFA, which also uses a ViT backbone

torchgeo/models/scale_mae.py
```diff
@@ -0,0 +1,252 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""Pre-trained Scale-MAE Vision Transformer models."""
```
adamjstewart, line 4:
```suggestion
"""Pre-trained Scale-MAE models."""
```

torchgeo/models/scale_mae.py
```diff
@@ -0,0 +1,252 @@
...
+    omega = 1.0 / 10000**omega
+    pos = pos.reshape(-1)
+    out = torch.einsum('m,d->md', pos, omega)
+    emb_sin = torch.sin(out)
+    emb_cos = torch.cos(out)
+    emb = torch.cat([emb_sin, emb_cos], dim=1)
+    return emb
+
+
+class ScaleMAEViT(VisionTransformer):  # type: ignore[misc]
```
adamjstewart, line 95:
```suggestion
class ScaleMAE(VisionTransformer):  # type: ignore[misc]
```

torchgeo/models/scale_mae.py
```diff
@@ -0,0 +1,252 @@
...
+    return state_dict
+
+
+# https://github.com/pytorch/vision/pull/6883
+# https://github.com/pytorch/vision/pull/7107
+# Can be removed once torchvision>=0.15 is required
+Weights.__deepcopy__ = lambda *args, **kwargs: args[0]
+
+
+class ScaleMAE_ViTLarge16_Weights(WeightsEnum):  # type: ignore[misc]
```
adamjstewart, line 191:
```suggestion
class ScaleMAELarge16_Weights(WeightsEnum):  # type: ignore[misc]
```

torchgeo/models/scale_mae.py
```diff
@@ -0,0 +1,252 @@
...
+            'in_chans': 3,
+            'img_size': 224,
+            'model': 'vit_large_patch16',
+            'publication': 'https://arxiv.org/abs/2212.14532',
+            'repo': 'https://github.com/bair-climate-initiative/scale-mae',
+        },
+    )
+
+
+def scalemae_vit_large_patch16(
```
adamjstewart, line 211:
```suggestion
def scalemae_large_patch16_224(
```

adamjstewart: Or is the image size customizable?

adamjstewart: Can we add additional sizes? I know we only have pre-trained weights available for large, but we might as well add functions to instantiate other sizes like they do in the source repo.

docs/api/misc_pretrained_weights.csv
adamjstewart: We may also want to add a short summary or table of which sensor-agnostic models provide which features. For example, DOFA enables explicit dynamic spectral band support (via model arch) and implicit dynamic resolution (via training data) while Scale-MAE has no dynamic spectral band support (RGB-only) but explicit dynamic resolution support (via model arch). Not sure about GASSL, maybe only implicit dynamic resolution (via training data)? It's worth mentioning that neither have dynamic temporal resolution support (maybe Satlas does?). I'm planning on highlighting this in our release notes, so I can also write something up if needed. Something like:

"The following pre-trained models offer dynamic spatial (resolution), temporal (time span), and/or spectral (wavelength) support, either via their training data (implicit) or via their model architecture (explicit):"

| Model | Spatial | Temporal | Spectral |
| --- | --- | --- | --- |
| DOFA | implicit | - | explicit |
| GASSL | implicit | - | - |
| Scale-MAE | explicit | - | - |

We could also optionally specify the range of resolutions/time spans/wavelengths that the model was pre-trained on. Just want to give users more feedback as to which model to choose.

adamjstewart (changes requested):
Some minor renaming suggestions to make things more consistent with DOFA, and some major documentation improvement suggestions. I'm willing to help with both if needed.

torchgeo/models/scale_mae.py
isaaccorley: The image size can be changed and the positional embeddings will get interpolated.

docs/api/misc_pretrained_weights.csv
isaaccorley: Should we save this for a different PR?

torchgeo/models/scale_mae.py
adamjstewart: Updated diff to remove 244

docs/api/misc_pretrained_weights.csv
adamjstewart: I'm fine with that, just don't let me forget before the release.

isaaccorley: I'll open a PR after this one so we don't forget to finish it

adamjstewart (approved):

PR #2062, opened 2024-05-15
Fix South America Soybean Filename Glob and Regex

yichiac:
This PR fixes the filename glob and regex for the files downloaded from the source.


tests/data/south_america_soybean/SouthAmericaSoybean.zip
adamjstewart, line 1: Why do we need this zip file? The real dataset doesn't have a zip file. Let's delete it and remove it from `data.py`.

adamjstewart (dismissed):

tests/data/south_america_soybean/data.py
adamjstewart, line 1: Delete this shutil import to get the tests to pass

adamjstewart (approved):

PR #2064, opened 2024-05-16
Datasets: add CLI support

adamjstewart:
Alternative to #2043 

#2043 is difficult to test because it requires `azcopy` to _not_ be installed, but it seems like it is installed on the GitHub Actions runners.

This PR goes in a different direction by adding a generic `which` function that returns an `Executable` for interacting with _any_ command-line program, not just `azcopy`. This is a watered-down version of the same idea used by [Spack](https://github.com/spack/spack/blob/develop/lib/spack/spack/util/executable.py), and can be used as follows:
```python
from torchgeo.datasets.utils import which

azcopy = which("azcopy")
azcopy("sync", "https://radiantearth.blob.core.windows.net/mlhub/nasa-tropical-storm-challenge", ".", "--recursive=true")
```

Prerequisite for #1830 and #2068
Closes #1887
Closes #1915
Closes #2043

### DIFF
--- a/tests/datasets/azcopy
+++ b/tests/datasets/azcopy
@@ -1,0 +1,19 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""Basic mock-up of the azcopy CLI."""
+
+import argparse
+import shutil
+
+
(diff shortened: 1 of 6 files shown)


calebrob6 (commented):
Nicely done!

PR #2074, opened 2024-05-21
EuroSAT Subclass with Longitude-Based Splits

burakekim:
`SpatialSplitEuroSAT` subclass overriding the default splits of `EuroSAT`.

Splits 60/20/20 by longitude.

Kudos to @calebrob6 for creating the [spatial splits](https://huggingface.co/datasets/torchgeo/eurosat/commit/1c11c73a87b40b0485d103231a97829991b8e22f).


torchgeo/datasets/eurosat.py
```diff
@@ -318,3 +318,25 @@ class EuroSAT100(EuroSAT):
...
+
+    Attributes:
+        split_urls (dict): A dictionary mapping each split to its corresponding URL.
+        split_md5s (dict): A dictionary mapping each split to its corresponding MD5.
+    """
+    
+    split_urls = {
+        "train": "https://hf.co/datasets/torchgeo/eurosat/resolve/main/eurosat-spatial-train.txt?download=true",  # noqa: E501
+        "val": "https://hf.co/datasets/torchgeo/eurosat/resolve/main/eurosat-spatial-val.txt?download=true",  # noqa: E501
+        "test": "https://hf.co/datasets/torchgeo/eurosat/resolve/main/eurosat-spatial-test.txt?download=true",  # noqa: E501
```
adamjstewart, line 336: Replace "main" with a specific commit hash for reproducibility

torchgeo/datasets/eurosat.py
```diff
@@ -318,3 +318,25 @@ class EuroSAT100(EuroSAT):
         'val': '3e3f8b3c344182b8d126c4cc88f3f215',
         'test': 'f908f151b950f270ad18e61153579794',
     }
+
+
+class SpatialSplitEuroSAT(EuroSAT):
+    """A subclass of the EuroSAT dataset class overriding the default splits.
```
adamjstewart, line 324: The fact that it's a subclass, and the parent class, will be automatically documented. Instead, let's talk about what it is and why it's useful.

adamjstewart (changes requested):
We haven't really thought about the best way to handle multiple split definitions for a single dataset. Do we want to create a new class, or just a new parameter in the original class?

burakekim:
> We haven't really thought about the best way to handle multiple split definitions for a single dataset. Do we want to create a new class, or just a new parameter in the original class?

No preference. Adding a new parameter to the original class might keep things simpler and cut down on extra code. However, IMHO, it looks cleaner this way

adamjstewart:
> On a side note, it might make sense to give the users heads up that they need to manually set the stage in setup method

That's something that can be documented in Lightning. I don't actually expect users to use the data module without a Trainer, which handles all of this for you.

torchgeo/datasets/eurosat.py
```diff
@@ -318,3 +318,24 @@ class EuroSAT100(EuroSAT):
...
+    """Overrides the default EuroSAT dataset splits.
+
+    Splits the data into training, validation, and test sets based on longitude.
+    The splits are distributed as 60%, 20%, and 20% respectively.
+
+    .. versionadded:: 0.6
+    """
+
+    split_urls = {
+        'train': 'https://hf.co/datasets/torchgeo/eurosat/resolve/1c11c73a87b40b0485d103231a97829991b8e22f/eurosat-spatial-train.txt?download=true',  # noqa: E501
```
adamjstewart, line 333: Doesn't this need to be "raw", not "resolve"? Also, can you remove `?download=true`? I'm also not sure if `noqa` is needed now that we're using ruff.

torchgeo/datasets/eurosat.py
```diff
@@ -318,3 +318,24 @@ class EuroSAT100(EuroSAT):
         'val': '3e3f8b3c344182b8d126c4cc88f3f215',
         'test': 'f908f151b950f270ad18e61153579794',
     }
+
+
+class SpatialSplitEuroSAT(EuroSAT):
```
adamjstewart, line 323: I wonder if EuroSATSpatial or EuroSATGeo would be a better name. It's up to you guys though, it's your dataset.

Another option would be to create a `GeoDataset` for `EuroSAT` that allows users to define their own geographic split. This would provide optimal flexibility, and allow you to combine with other imagery as well.

These are both completely optional. I don't know what the best approach is. It's probably fine to merge as is, just thinking outside the box to make sure we've covered every possibility.

burakekim: Great ideas! I will go with EuroSATSpatial then; it sounds a bit more descriptive, I guess.

Regarding refactoring as GeoDataset: the flexibility of defining custom geo splits sounds intriguing and could theoretically be applied to any of the NonGeoDatasets. However, I am not exactly sure if it's worth the extra code and maintenance. Overriding the splits sounds very convenient as I see it

torchgeo/datasets/eurosat.py
burakekim: Done

torchgeo/datasets/eurosat.py
adamjstewart: > and could theoretically be applied to any of the NonGeoDatasets

Only the ones that store files with geospatial metadata. Many use .jpg or .png and don't release this metadata.

If you do go with a GeoDataset, then EuroSATGeo might be a better name. That's what we did for LandCover.ai which has both Geo- and NonGeoDataset versions.

adamjstewart:
Also needs tests. Should be as simple as adding one more class name [here](https://github.com/microsoft/torchgeo/blob/v0.5.2/tests/datasets/test_eurosat.py#L26), one config file similar to [here](https://github.com/microsoft/torchgeo/blob/v0.5.2/tests/conf/eurosat.yaml), and one new line [here](https://github.com/microsoft/torchgeo/blob/v0.5.2/tests/trainers/test_classification.py#L73).

torchgeo/datasets/eurosat.py
burakekim: Right, the ones with geotiff should be suitable only. Re GeoDataset: I would go with the current settings for now if that is fine. Would love to reformulate that way if I need higher flexibility (i.e., another set of partitions)

burakekim:
> Also needs tests. Should be as simple as adding one more class name [here](https://github.com/microsoft/torchgeo/blob/v0.5.2/tests/datasets/test_eurosat.py#L26), one config file similar to [here](https://github.com/microsoft/torchgeo/blob/v0.5.2/tests/conf/eurosat.yaml), and one new line [here](https://github.com/microsoft/torchgeo/blob/v0.5.2/tests/trainers/test_classification.py#L73).

Thanks for pointing out these! Done. Hope it is good now

torchgeo/datamodules/eurosat.py
```diff
@@ -93,3 +93,29 @@ def __init__(
         bands = kwargs.get('bands', EuroSAT.all_band_names)
         self.mean = torch.tensor([MEAN[b] for b in bands])
         self.std = torch.tensor([STD[b] for b in bands])
+
+
+class EuroSATSpatialDataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the EuroSATSpatial dataset.
+
+    Intended for tutorials and demonstrations, not for benchmarking.
```
adamjstewart, line 101: This one is not for tutorials, it is for benchmarking.

adamjstewart (changes requested):
Also needs documentation: https://github.com/microsoft/torchgeo/blob/v0.5.2/docs/api/datasets.rst#L218

adamjstewart:
Add to `torchgeo.datamodules.__init__.py` to solve the readthedocs error.

adamjstewart (dismissed):
Optional: we could put Spatial before 100 in the code and docs since users aren't really giong to use 100, it's only for demos. But it's prob fine to merge as is.

burakekim:
> Optional: we could put Spatial before 100 in the code and docs since users aren't really giong to use 100, it's only for demos. But it's prob fine to merge as is.

No preference: So I can do that!

burakekim:
Done!

adamjstewart (approved):

PR #2085, opened 2024-05-25
Sentinel: test length

adamjstewart:
Working on a PR to test the length of every dataset. Curious if we can get away with half as many S2 test images.


adamjstewart:
Nope. I don't love that the 2022 directory contains data for multiple years, but don't want to make data.py gigantic just to solve this. Let's close for now.

PR #2086, opened 2024-05-25
GlobBiomass: several bug fixes

adamjstewart:
This PR fixes the following bugs:

- [x] GlobBiomass data is from 2010, although it isn't in the filename
- [x] Mask is for pixelwise regression, use float32, not long
- [x] Length of dataset was doubled due to unspecific filename_glob

I also improved the documentation to include definitions and units of AGB and GSV.

I didn't do these in this PR, but some suggestions for future contributions to this dataset:

- [ ] Add download support, URLs are public
- [ ] Add support for file extensions other than `.tif`
- [ ] Split estimation mask and uncertainty mask into two different dict keys (backwards incompatible)
- [ ] Treat measurements like bands and allow return both
- [ ] Treat measurements AND errors like bands and allow return both
- [ ] Add data module, our first real pixelwise regression dataset

### DIFF
--- a/torchgeo/datasets/globbiomass.py
+++ b/torchgeo/datasets/globbiomass.py
@@ -15,45 +15,56 @@ from rasterio.crs import CRS
 
 from .errors import DatasetNotFoundError
 from .geo import RasterDataset
-from .utils import BoundingBox, check_integrity, extract_archive
+from .utils import BoundingBox, check_integrity, disambiguate_timestamp, extract_archive
 
 
 class GlobBiomass(RasterDataset):
     """GlobBiomass dataset.
 
-    The `GlobBiomass dataset <https://doi.pangaea.de/10.1594/PANGAEA.894711>`_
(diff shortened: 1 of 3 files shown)


isaaccorley (approved):

torchgeo/datasets/globbiomass.py
```diff
@@ -15,45 +15,56 @@
...
+    * AGB: m3/ha
+    * GSV: tons/ha (i.e., Mg/ha)
 
     Dataset features:
 
-    * estimates of AGB and GSV around the world at ~100m per pixel resolution
-      (45,000x45,0000 px)
-    * standard error maps of respective measurement at same resolution
+    * Global estimates of AGB and GSV at ~100 m per pixel resolution
+      (45,000 x 45,0000 px)
```
isaaccorley, line 42: I think there might be an extra 0 in the second 45,0000

isaaccorley (changes requested):

isaaccorley (approved):

PR #2092, opened 2024-05-28
Update VHR-10 dataset plotting

robmarkcole:
Fix plotting using percentiles

Note: I have uploaded the dataset at https://huggingface.co/datasets/satellite-image-deep-learning/VHR-10 - can transfer to torchgeo org if you want to use it here

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,15 +120,22 @@ TorchGeo includes a number of [_benchmark datasets_](https://torchgeo.readthedoc
 If you've used [torchvision](https://pytorch.org/vision) before, these datasets should seem very familiar. In this example, we'll create a dataset for the Northwestern Polytechnical University (NWPU) very-high-resolution ten-class ([VHR-10](https://github.com/chaozhong2010/VHR-10_dataset_coco)) geospatial object detection dataset. This dataset can be automatically downloaded, checksummed, and extracted, just like with torchvision.
 
 ```python
+from torch.utils.data import DataLoader
+
+from torchgeo.datamodules.utils import collate_fn_detection
 from torchgeo.datasets import VHR10
-from torchgeo.datamodules import VHR10DataModule
 
 # Initialize the dataset
 dataset = VHR10(root="...", download=True, checksum=True)
 
 # Initialize the dataloader with the custom collate function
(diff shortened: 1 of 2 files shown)


robmarkcole:
I just assumed the docs predate the datamodules but we would like to showcase this feature - personally I always use datamodules now but perhaps am in the minority there. Suggest a second opinion before I change the PR

robmarkcole:
@ashnair1  I checked the approach I suggested (`imgIds=id_)`) and this results in erroneous masks - strange as the API appears to support this approach, suggesting the dataset has some quirks

image

adamjstewart:
### Regarding the README

The README has a few different sections. The section you're editing is designed to show off our benchmark _datasets_. There is a separate section below specifically for PyTorch Lightning, including _data modules_. I would prefer to keep using a dataset instead of a data module here. 

Actually, I think we're better off choosing a different dataset (maybe EuroSAT?). VHR-10 is too complicated and makes it look like TorchGeo is hard to use. Maybe we could move VHR-10 to the Lightning section of the README since it has a pretty visualization.

### Regarding plotting

Yes, this is a common issue with plotting via dataset vs. plotting via data module. I agree with your solution, and proposed basically the same thing for all datasets in #1263. This issue is not specific to VHR-10, but affects almost all datasets/data modules. I haven't had the time/energy to fix this for all datasets, but contributions are definitely welcome here, even if it's only 1 dataset at a time.

Can we separate the README and plotting changes into two separate PRs?

adamjstewart:
> Note: I have uploaded the dataset at https://huggingface.co/datasets/satellite-image-deep-learning/VHR-10 - can transfer to torchgeo org if you want to use it here

It's easier to download tar/zip/rar files instead of raw data. But I'm happy to change the download to your org or ours. Not sure if we should rehost all rar datasets as tar/zip to make them easier to extract?

robmarkcole:
@adamjstewart this PR now just addresses the plotting.

RE dataset choice, I agree EuroSAT would be a nice choice - it also supports selecting subsets of bands so shows immediately what is possible with torchgeo

adamjstewart (approved):

PR #2097, opened 2024-05-29
RESISC45: document rarfile dependency

adamjstewart:
Copied the text from VHR-10, our only other .rar dataset.


burakekim:
@adamjstewart Did we contact the authors to ask whether we can host a copy ourselves? Since I am already on a roll(!), I can try my luck to save us from this dependency. What do you say?

adamjstewart:
Yes, let's try that! I messaged you on Slack with more suggestions.

adamjstewart:
I think we can still merge this while we wait to hear back from the authors.

PR #2101, opened 2024-05-30
Update num_classes parameter description in ObjectDetectionTask class

robmarkcole:
adds num_classes. - Number of prediction classes (including the background) to the ObjectDetectionTask docs,


robmarkcole:
Considering `in_channels` does nothing currently, is it worth mentioning? Ref https://github.com/microsoft/torchgeo/issues/1156

robmarkcole:
@adamjstewart done

PR #2112, opened 2024-06-06
AgriFieldNet: fix list formatting

adamjstewart:


adamjstewart:
Ugh, Sphinx only supports sequentially numbered ordered lists...

nilsleh (approved):

PR #2130, opened 2024-06-19
SemanticSegmentationTask: add class-wise metrics

robmarkcole:
Addresses https://github.com/microsoft/torchgeo/issues/2121 for segmentation. Mostly copied from @isaaccorley as [here](https://github.com/isaaccorley/a-change-detection-reality-check/blob/0b570f095366685ce401e62c0a1b69a2783c0fbf/src/change_detection.py#L129) - he is additionally passing `on_epoch=True` which is also adopted here


DimitrisMantas:
Given that most metrics of interest are broken (e.g., all of them when ``average="macro"`` and ``ignore_index`` is specified, ``JaccardIndex`` outputting zeros instead of NaNs for absent and ignored classes and also taking them into account in the reduction step), should we make an effort to see if and how we could fix them?

robmarkcole:
@DimitrisMantas good call on my ignoring the `ignore_index`.! In fairness they do address issues, but have a long backlog. When I made some noise they addressed https://github.com/Lightning-AI/torchmetrics/pull/2198
My opinion is it is better to work with torchmetrics to address the issues, rather than implement from scratch here. I see your comment at https://github.com/Lightning-AI/torchmetrics/issues/2535#issuecomment-2143514389 so perhaps a pragmatic approach is not to add new metrics that we have concerns about, but also to create specific issues which track these concerns

torchgeo/trainers/segmentation.py
```diff
@@ -232,7 +315,10 @@ def training_step(
         loss: Tensor = self.criterion(y_hat, y)
         self.log('train_loss', loss, batch_size=batch_size)
         self.train_metrics(y_hat, y)
-        self.log_dict(self.train_metrics, batch_size=batch_size)
+        self.log_dict(
+            {f'{k}': v for k, v in self.train_metrics.compute().items()},
```
DimitrisMantas, line 319: I think that ``compute()`` should not be called per step, since this will cause a reduction to happen multiple times per epoch and in the end the output is the average of the average of the average...

DimitrisMantas:
Sure, that makes sense; please excuse the rant haha.

torchgeo/trainers/segmentation.py
robmarkcole: Can be addressed with `on_epoch=True`?

DimitrisMantas: Not sure exactly how this works, but I found it gave wrong results for when ``on_epoch`` wasn't specified.

robmarkcole:
Applied `on_epoch=True,` to all steps for consistency - however this might make the train loss less useful as it is now reported per epoch rather than per step

isaaccorley:
I just set on_epoch=True to be explicit but I think that pytorch lightning auto sets on_epoch to be False for training and True for all else.

DimitrisMantas:
You need to set both ``on_step`` and ``on_epoch`` to get logs only per step or per epoch.

robmarkcole:
@DimitrisMantas now just performing `on_step` for train loss, so a more manageable 36 metrics now

robmarkcole:
Not sure about this failing test `ValueError: Problem with given class_path 'torchgeo.trainers.SemanticSegmentationTask'`

isaaccorley:
Must be an issue with on of the minimum versions of the package since it's passing for the other tests.

torchgeo/trainers/segmentation.py
```diff
@@ -230,9 +324,14 @@ def training_step(
         batch_size = x.shape[0]
         y_hat = self(x)
         loss: Tensor = self.criterion(y_hat, y)
-        self.log('train_loss', loss, batch_size=batch_size)
+        self.log('train_loss', loss, batch_size=batch_size, on_epoch=True, on_step=True)
         self.train_metrics(y_hat, y)
-        self.log_dict(self.train_metrics, batch_size=batch_size)
+        self.log_dict(
+            {f'{k}': v for k, v in self.train_metrics.compute().items()},
```
robmarkcole, line 330: I'm a bit unsure about this workflow - presumably `compute` is called once at the `on_epoch` and the `reset` is called automatically after? Could be safer to use the `on_stage_epoch_end` unless anyone is very confident about this

adamjstewart:
The failing tests are for the minimum supported version of our dependencies. If you install the same version of torchmetrics we're using in https://github.com/microsoft/torchgeo/blob/main/requirements/min-reqs.old, you should be able to reproduce the bug. Feel free to increase that version if needed.

robmarkcole:
@adamjstewart tests passing

adamjstewart:
The new min is pretty recent. Did you ever figure out why older versions were raising an error?

requirements/min-reqs.old
```diff
@@ -18,7 +18,7 @@ segmentation-models-pytorch==0.2.0
 shapely==1.8.0
 timm==0.4.12
 torch==1.13.0
-torchmetrics==0.10.0
+torchmetrics==1.2.0
```
adamjstewart, line 21: You'll also need to change `pyproject.toml` and `.pre-commit-config.yaml`

torchgeo/trainers/segmentation.py
```diff
@@ -250,9 +349,13 @@ def validation_step(
         batch_size = x.shape[0]
         y_hat = self(x)
         loss = self.criterion(y_hat, y)
-        self.log('val_loss', loss, batch_size=batch_size)
+        self.log('val_loss', loss, batch_size=batch_size, on_epoch=True)
```
DimitrisMantas, line 352: I believe `on_epoch=True` is the default behavior for `validation_step` and `training_step` (even though the latter is [undocumented](https://lightning.ai/docs/pytorch/stable/extensions/logging.html#logging-from-a-lightningmodule)), so maybe we could get rid of this?

torchgeo/trainers/segmentation.py
```diff
@@ -250,9 +349,13 @@ def validation_step(
...
         self.val_metrics(y_hat, y)
-        self.log_dict(self.val_metrics, batch_size=batch_size)
+        self.log_dict(
+            {f'{k}': v for k, v in self.val_metrics.compute().items()},
```
DimitrisMantas, line 355: I believe this is for the class-wise metrics but is not needed in later versions of TorchMetrics as `ClasswiseWrapper` handles the reporting correctly.

We should definitely get rid of this as `compute()` reduces the metric, but this should happen at the end of each *epoch* not *step*.

adamjstewart:
This PR does a lot of stuff:

* Add macro, micro, and classwise averaging
* Switch from MulticlassAccuracy to Accuracy, etc.
* Modify on_epoch/on_step logging frequency

Can we split these into separate PRs so it's easier to review and clearer which changes necessitate what? The first couple are likely easy to approve, the latter may require justification. I'm also worried that we'll change these only for segmentation and forget about the rest and end up with non-uniform metrics.

robmarkcole:
@adam I also tried `torchmetrics==1.1.0` and got the failures, but didn't dig into it

robmarkcole:
@adamjstewart have dropped the `on_epoch/on_step` modifications from this MR. It feels manageable to me now, but can split further if required

pyproject.toml
```diff
@@ -72,8 +72,8 @@ dependencies = [
     "timm>=0.4.12",
     # torch 1.13+ required by torchvision
     "torch>=1.13",
-    # torchmetrics 0.10+ required for binary/multiclass/multilabel classification metrics
-    "torchmetrics>=0.10",
+    # torchmetrics 1.2.0+ required for MetricCollection
```
adamjstewart, line 75: MetricCollection exists in older releases too, this isn't the reason why 1.2.0+ is required. Also, I usually chop off the .0 at the end

torchgeo/trainers/segmentation.py
```diff
@@ -190,27 +199,112 @@ def configure_metrics(self) -> None:
...
              for balanced performance assessment across imbalanced classes.
         """
         num_classes: int = self.hparams['num_classes']
         ignore_index: int | None = self.hparams['ignore_index']
-        metrics = MetricCollection(
-            [
-                MulticlassAccuracy(
+        labels: list[str] | None = self.hparams['labels']
+
+        self.train_metrics = MetricCollection(
```
adamjstewart, line 209: Can this be done in a for-loop?

pyproject.toml
robmarkcole: I could keep iterating on older versions until the test errors appear, then ref those/dig into the reason. 1.2 was released a year ago, I think it is a reasonable expectation that codebases will be updated since then? https://pypi.org/project/torchmetrics/#history

adamjstewart: I would still like to know why 1.2 is required. I would start by searching for the error message you were seeing on GitHub.

torchgeo/trainers/segmentation.py
robmarkcole: Dropping on_epoch, val defaults to on_step, as does train.

Re compute, reverted to
```
        self.val_metrics(y_hat, y)
        self.log_dict(
            self.val_metrics,
            batch_size=batch_size,
        )
```
and get 
```
ValueError: The `.compute()` return of the metric logged as 'val_Accuracy' must be a tensor. Found {'multiclassaccuracy_background': tensor(0.0009, device='cuda:0'), 'multiclassaccuracy_burned_area': tensor(0.9992, device='cuda:0')}
```

robmarkcole:
@adamjstewart I'm inclined to close this PR as I don't feel confident I understand the behaviour of torchmetrics in this implementation. Elsewhere I am using the on_stage_epoch_end hooks and feel confident I do understand the behaviour with that approach. Overall I think this should be a change we make from a place of understanding, and in smaller steps than this PR takes

torchgeo/trainers/segmentation.py
DimitrisMantas: It works as expected on v1.4.0. I remember I had this issue when before I upgraded versions on my project and got around it by basically doing `metrics(input, target)` and then looping over the `MetricCollection` object and unpacking any iterables I found. I think this allowed me to get rid of the `compute` call.

torchgeo/trainers/segmentation.py
robmarkcole: Something like the following? I think it is less readable 
```python
# Define the metrics and their parameters
metric_definitions = {
    'OverallAccuracy': (Accuracy, {'average': 'micro'}),
    'OverallF1Score': (FBetaScore, {'average': 'micro', 'beta': 1.0}),
    'OverallJaccardIndex': (JaccardIndex, {'average': 'micro'}),
    'AverageAccuracy': (Accuracy, {'average': 'macro'}),
    'AverageF1Score': (FBetaScore, {'average': 'macro', 'beta': 1.0}),
    'AverageJaccardIndex': (JaccardIndex, {'average': 'macro'}),
}

classwise_metric_definitions = {
    'Accuracy': Accuracy,
    'Precision': Precision,
    'Recall': Recall,
    'F1Score': (FBetaScore, {'beta': 1.0}),
    'JaccardIndex': JaccardIndex,
}

# Initialize the metrics dictionary
metrics_dict = {}

# Populate the metrics dictionary using the metric definitions
for name, (metric_class, params) in metric_definitions.items():
    metrics_dict[name] = metric_class(
        task='multiclass',
        num_classes=num_classes,
        multidim_average='global',
        ignore_index=ignore_index,
        **params
    )

for name, metric_class in classwise_metric_definitions.items():
    params = {}
    if isinstance(metric_class, tuple):
        metric_class, params = metric_class
    metrics_dict[name] = ClasswiseWrapper(
        metric_class(
            task='multiclass',
            num_classes=num_classes,
            average='none',
            multidim_average='global',
            ignore_index=ignore_index,
            **params
        ),
        labels=labels
    )

self.train_metrics = MetricCollection(metrics_dict, prefix='train_')
```

robmarkcole:
torchmetrics=1.1.0 test errors [here](https://github.com/microsoft/torchgeo/actions/runs/10286260844/job/28466487135?pr=2130)
```
MeanAveragePrecision(), kwargs = {'average': 'macro'}
...
ValueError: Unexpected keyword arguments: `average`
```

torchgeo/trainers/segmentation.py
adamjstewart: Something like:
```python
for average in ['macro', 'micro', 'none']:
    ...
```
May need an if-statement for 'none'.
May need

robmarkcole: Th F1 score doesn't quite fit the pattern:
```python
from torchmetrics import MetricCollection, Accuracy, FBetaScore, JaccardIndex, Precision, Recall

# Define metric classes and their parameters
metric_classes = {
    'Accuracy': Accuracy,
    'F1Score': FBetaScore,
    'JaccardIndex': JaccardIndex,
    'Precision': Precision,
    'Recall': Recall,
}

# Define a dictionary to hold the metric instances
metrics_dict = {}

# Loop through the types of averaging
for average in ['micro', 'macro']:
    for metric_name, metric_class in metric_classes.items():
        name = f"Overall{metric_name}" if average == 'micro' else f"Average{metric_name}"
        params = {
            'task': 'multiclass',
            'num_classes': num_classes,
            'average': average,
            'multidim_average': 'global',
            'ignore_index': ignore_index,
        }
        if metric_name == 'F1Score':
            params['beta'] = 1.0
        metrics_dict[name] = metric_class(**params)

# Loop through the classwise metrics
for metric_name, metric_class in metric_classes.items():
    if metric_name != 'F1Score':
        metrics_dict[metric_name] = ClasswiseWrapper(
            metric_class(
                task='multiclass',
                num_classes=num_classes,
                average='none',
                multidim_average='global',
                ignore_index=ignore_index,
            ),
            labels=labels
        )
    else:
        metrics_dict[metric_name] = ClasswiseWrapper(
            metric_class(
                task='multiclass',
                num_classes=num_classes,
                beta=1.0,
                average='none',
                multidim_average='global',
                ignore_index=ignore_index,
            ),
            labels=labels
        )

# Create the MetricCollection with the prefix 'train_'
self.train_metrics = MetricCollection(metrics_dict, prefix='train_')
```

torchgeo/trainers/segmentation.py
robmarkcole: I would be all for `v1.4.0` personally :-)

torchgeo/trainers/segmentation.py
robmarkcole: Later we could refactor into a metrics factory method

torchgeo/trainers/segmentation.py
robmarkcole: Ok looks like we support 1.4.1?? https://github.com/microsoft/torchgeo/blob/8c02cec3801672947455db92f76537072b15dd4a/requirements/required.txt#L21

robmarkcole:
Tests failing due to use of `multidim_average`

pyproject.toml
```diff
@@ -47,9 +47,9 @@ dependencies = [
     # https://github.com/microsoft/torchgeo/issues/1824
     "lightly>=1.4.5,!=1.4.26",
     # lightning 2+ required for LightningCLI args + sys.argv support
-    # lightning 2.3 contains known bugs related to YAML parsing
+    # lightning 2.3+ contains known bugs related to YAML parsing
     # https://github.com/Lightning-AI/pytorch-lightning/issues/19977
-    "lightning[pytorch-extra]>=2,!=2.3.*",
+    "lightning[pytorch-extra]>=2,<2.3",
```
robmarkcole, line 52: wierd, locally I dont have this change

torchgeo/trainers/segmentation.py
adamjstewart: https://github.com/microsoft/torchgeo/blob/main/requirements/README.md

adamjstewart: Does it work in 1.3 or 1.2? 1.4 is pretty new. We do require the latest and greatest kornia, so it's not impossible to require such a new version, just don't want to make torchgeo difficult to install for people stuck on older versions.

robmarkcole: @DimitrisMantas can you share an example implementation re https://github.com/microsoft/torchgeo/pull/2130#discussion_r1707124290

robmarkcole:
After discussion with torchmetrics devs, created https://github.com/Lightning-AI/torchmetrics/issues/2683

adamjstewart:
That's such a complicated minimal reproducible example lol.

adamjstewart:
I tried making a self-contained minimal reproducible example but couldn't get one working and gave up.

DimitrisMantas:
It just hit me that we should be a bit careful with which metrics we add to avoid unnecessary computation; class-wise accuracy and recall are the same thing and so are micro-averaged accuracy, precision, and recall.

adamjstewart:
Any sense of how much these metrics actually add to processing time? If it isn't noticeable by a human, I don't particularly care about the overhead.

DimitrisMantas:
Haven't measured it but doubt it's much.

robmarkcole:
I believe Lightning offers tools for profiling

adamjstewart:
They do, see https://torchgeo.readthedocs.io/en/latest/user/contributing.html#i-o-benchmarking

torchgeo/trainers/segmentation.py
```diff
@@ -232,7 +284,10 @@ def training_step(
         loss: Tensor = self.criterion(y_hat, y)
         self.log('train_loss', loss, batch_size=batch_size)
         self.train_metrics(y_hat, y)
-        self.log_dict(self.train_metrics, batch_size=batch_size)
+        self.log_dict(
```
robmarkcole, line 287: log_dict now supports MetricCollections https://github.com/Lightning-AI/torchmetrics/pull/2720

robmarkcole:
@adamjstewart @DimitrisMantas per this comment we should be using the `_epoch_end` hooks https://github.com/Lightning-AI/torchmetrics/issues/2683#issuecomment-2331146337

DimitrisMantas:
I see the issue, but I must be missing something because [my own code](https://github.com/DimitrisMantas/RoofSense/blob/master/training/task.py) uses the standard logging tools and metric collections work just fine.

DimitrisMantas:
Ok, so basically what the torchmetrics guys are [saying](https://github.com/Lightning-AI/torchmetrics/issues/2683#issuecomment-2331146337) is that automatic logging is not supported for metric collections?

DimitrisMantas:
@robmarkcole I can confirm the recommended approach yields consistent results.

calebrob6 (approved):

adamjstewart:
Sorry it's taken me so long to review. I was originally hung up on the hack required to support `ClasswiseWrapper` in `log_dict`, but I've gotten over that. If torchmetrics wants to make that easier in the future, great. But I also really want this feature, so let's not wait on that.

Only remaining concern is that the code required to loop over all metrics and averages actually makes the code more complicated and difficult to read than avoiding loops entirely. If we want to add new metrics in the future, it looks non-straightforward. I wonder if we can loop over averages only and still keep things simple.

I would also really like to see this done for `ClassificationTask` too so `SemanticSegmentationTask` doesn't have a different set of features or metrics.

isaaccorley:
I think it's better practice to do `on_epoch=True` for val and test metrics so that you will get a single value per metric per epoch for the validation and test set. Otherwise you will get a metric per step which is not great for comparing performance across models.

robmarkcole:
To clarify if we want `on_step=False,` or `True` for train

isaaccorley:
That's a good question, I normally set the train to only record steps and not the full epoch average because depending on how many metrics and the size of your train set this can use up a lot of memory because it's saving every step and then averaging at the end of the epoch. I think for now we can set both as true.

robmarkcole:
From memory I think if using both the loss is named `loss_step` and `loss_epoch`, trying to get a model trained to demonstrate. I assume most people will be fine from this, but it could require some explanation

isaaccorley (dismissed):
This LGTM. Any concerns with merging this @adamjstewart?

adamjstewart:
I don't think these concerns have been addressed: https://github.com/microsoft/torchgeo/pull/2130#issuecomment-2386988586

isaaccorley:
I agree, I think just making a list of the metrics instead of using loops is probably more readable and easier to add/remove individual metrics based on a users need

robmarkcole:
merging main really messed this branch up, will create a new branch/PR

calebrob6:
Coming back to this to say I just reimplemented in a private project, would love to have this upstream!

robmarkcole:
Honestly dont know when I will get around to it - the last occasion I had a few hours I ended up debugging dataset issues

pyproject.toml
```diff
@@ -74,8 +74,8 @@ dependencies = [
     "timm>=0.4.12",
     # torch 1.13+ required by torchvision
     "torch>=1.13",
-    # torchmetrics 0.10+ required for binary/multiclass/multilabel classification metrics
-    "torchmetrics>=0.10",
+    # torchmetrics 1.1.1+ required for average argument to MeanAveragePrecision
```
adamjstewart, line 77: According to https://github.com/Lightning-AI/torchmetrics/commit/63c7bbe6ac163659e9631badce296ff1c658d45e the argument didn't exist until 1.2

PR #2147, opened 2024-07-01
datamodules: Switch to kornia AugmentationSequential

ashnair1:
Switch to kornia's `AugmentationSequential` in `datamodules` submodule

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,7 @@ repos:
           - kornia>=0.6.9
           - lightning>=2.0.9
           - matplotlib>=3.8.1
-          - numpy>=1.22
+          - numpy>=1.22,<2
           - pillow>=10.3.0
           - pytest>=6.1.2
           - pyvista>=0.34.2
--- a/torchgeo/datamodules/chesapeake.py
+++ b/torchgeo/datamodules/chesapeake.py
@@ -13,7 +13,6 @@ from torch import Tensor
 
 from ..datasets import ChesapeakeCVPR
 from ..samplers import GridGeoSampler, RandomBatchGeoSampler
-from ..transforms import AugmentationSequential
 from .geo import GeoDataModule
 
 
@@ -122,8 +121,8 @@ class ChesapeakeCVPRDataModule(GeoDataModule):
         else:
             self.layers = ['naip-new', 'lc']
 
-        self.aug = AugmentationSequential(
-            K.Normalize(mean=self.mean, std=self.std), data_keys=['image', 'mask']
+        self.aug = K.AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std), data_keys=None
         )
 
     def setup(self, stage: str) -> None:
--- a/torchgeo/datamodules/inria.py
+++ b/torchgeo/datamodules/inria.py
@@ -9,7 +9,6 @@ import kornia.augmentation as K
 
 from ..datasets import InriaAerialImageLabeling
 from ..samplers.utils import _to_tuple
-from ..transforms import AugmentationSequential
 from ..transforms.transforms import _RandomNCrop
 from .geo import NonGeoDataModule
 
@@ -40,26 +39,26 @@ class InriaAerialImageLabelingDataModule(NonGeoDataModule):
             **kwargs: Additional keyword arguments passed to
                 :class:`~torchgeo.datasets.InriaAerialImageLabeling`.
         """
-        super().__init__(InriaAerialImageLabeling, 1, num_workers, **kwargs)
+        super().__init__(InriaAerialImageLabeling, batch_size, num_workers, **kwargs)
 
         self.patch_size = _to_tuple(patch_size)
 
-        self.train_aug = AugmentationSequential(
+        self.train_aug = K.AugmentationSequential(
             K.Normalize(mean=self.mean, std=self.std),
             K.RandomHorizontalFlip(p=0.5),
             K.RandomVerticalFlip(p=0.5),
             _RandomNCrop(self.patch_size, batch_size),
-            data_keys=['image', 'mask'],
+            data_keys=None,
         )
-        self.aug = AugmentationSequential(
+        self.aug = K.AugmentationSequential(
             K.Normalize(mean=self.mean, std=self.std),
             _RandomNCrop(self.patch_size, batch_size),
-            data_keys=['image', 'mask'],
+            data_keys=None,
         )
-        self.predict_aug = AugmentationSequential(
+        self.predict_aug = K.AugmentationSequential(
             K.Normalize(mean=self.mean, std=self.std),
             _RandomNCrop(self.patch_size, batch_size),
-            data_keys=['image'],
+            data_keys=None,
         )
 
     def setup(self, stage: str) -> None:
--- a/tests/datamodules/test_geo.py
+++ b/tests/datamodules/test_geo.py
@@ -32,7 +32,7 @@ class CustomGeoDataset(GeoDataset):
 
     def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
         image = torch.arange(3 * 2 * 2, dtype=torch.float).view(3, 2, 2)
-        return {"image": image, "crs": CRS.from_epsg(4326), "bbox": query}
+        return {'image': image, 'crs': CRS.from_epsg(4326), 'bbox': query}
 
     def plot(self, *args: Any, **kwargs: Any) -> Figure:
         return plt.figure()
@@ -68,7 +68,7 @@ class CustomNonGeoDataset(NonGeoDataset):
         self.length = length
 
     def __getitem__(self, index: int) -> dict[str, Tensor]:
-        return {"image": torch.arange(3 * 2 * 2, dtype=torch.float).view(3, 2, 2)}
+        return {'image': torch.arange(3 * 2 * 2, dtype=torch.float).view(3, 2, 2)}
 
     def __len__(self) -> int:
         return self.length
--- a/tests/datamodules/test_geo.py
+++ b/tests/datamodules/test_geo.py
@@ -31,8 +31,8 @@ class CustomGeoDataset(GeoDataset):
         self.res = 1
 
     def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
-        image = torch.arange(3 * 2 * 2).view(3, 2, 2)
-        return {'image': image, 'crs': CRS.from_epsg(4326), 'bbox': query}
+        image = torch.arange(3 * 2 * 2, dtype=torch.float).view(3, 2, 2)
+        return {"image": image, "crs": CRS.from_epsg(4326), "bbox": query}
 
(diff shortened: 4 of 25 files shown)


torchgeo/datasets/cyclone.py
```diff
@@ -125,10 +125,12 @@ def __getitem__(self, index: int) -> dict[str, Any]:
         )
 
         sample: dict[str, Any] = {'image': self._load_image(directory)}
-        sample.update(self._load_features(directory))
+        features = self._load_features(directory)
+        sample['label'] = features['label']
 
         if self.transforms is not None:
             sample = self.transforms(sample)
+            sample.update({x: features[x] for x in features if x != 'label'})
```
ashnair1, line 133: Assumes that transforms will only be applied on kornia keys i.e. 'image', 'mask', 'class', 'boxes' etc.

ashnair1: Would be nice if kornia's `AugmentationSequential` ignores keys are that are not relevant to it while accepting a dict input.

ashnair1:
TODO: Tutorials need to be updated once we fully switch to kornia's `AugmentationSequential`

.pre-commit-config.yaml
```diff
@@ -25,7 +25,7 @@ repos:
           - kornia>=0.6.9
           - lightning>=2.0.9
           - matplotlib>=3.8.1
-          - numpy>=1.22
+          - numpy>=1.22,<2
```
adamjstewart, line 28: Should be no longer required after #2151

torchgeo/datamodules/inria.py
```diff
@@ -40,26 +39,26 @@ def __init__(
             **kwargs: Additional keyword arguments passed to
                 :class:`~torchgeo.datasets.InriaAerialImageLabeling`.
         """
-        super().__init__(InriaAerialImageLabeling, 1, num_workers, **kwargs)
+        super().__init__(InriaAerialImageLabeling, batch_size, num_workers, **kwargs)
```
adamjstewart, line 42: Why this change?

ashnair1: Might have included that by accident in this PR. I noticed this passing of 1 to super in a couple of classes (`deepglobelandcover`, `inria`, `levircd`, `potsdam`, `vahingen`). Shouldn't we be passing `batch_size` and not 1 here?

adamjstewart: We chose 1 instead of batch size because these all use RandomNCrop which takes a single image (batch size=1) and converts it to N crops (batch size=batch size). All of these data modules are for images that are very large where we can use a single image to create an entire mini-batch instead of loading N files from disk. Not the perfect solution, happy to change this. But that's why it is the way it is.

ashnair1: Got it. Will be reverting this

torchgeo/trainers/segmentation.py
```diff
@@ -289,6 +295,8 @@ def test_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) -> None
             batch_idx: Integer displaying index of this batch.
             dataloader_idx: Index of the current dataloader.
         """
+        if 'mask' in batch and batch['mask'].shape[1] == 1:
```
adamjstewart, line 298: Generally speaking, I'm trying to reduce the number of if-statements. Things like `torch.compile` should theoretically work at the level of LightningModules, but don't when they contain if-statements. I know we don't yet have the ability to assert the dimensions of things like masks, but can we fix this at the dataset or data module level instead of at the trainer level? Related to #985 where I tried and failed to propose a standardized type and dimension for all sample keys.

ashnair1: We could fix this at the datamodule level by adding this if statement to every segmentation datasets' `on_after_batch_transfer` function

adamjstewart: Actually, I think data modules are also `nn.Module`s and could theoretically be compiled. Since they are attached to the trainer, they might even _need_ to be compilable in order to compile the trainer/lightning module. Can we fix this at the dataset level instead?

ashnair1: No. The extra channel dim is added by the augmentation. We could potentially push this check to kornia though.

adamjstewart: Will using keep_dim help? We should use that in every single transform. I actually can't believe it's false by default.

ashnair1: Testing inria now (via `pytest -v tests/trainers/segmentation.py`) and setting keepdim to True does resolve the error in trainer (`test_trainer`) but not the one in `test_no_plot_method` and `test_no_rgb`.

ashnair1: Ah no, that's because those tests use a different datamodule. Ok, we might have something here.

ashnair1: So setting `keepdim=True` does indeed work for all NonGeoDataModules. But tests are failing for most GeoDataModules.

.pre-commit-config.yaml
```diff
@@ -24,7 +24,7 @@ repos:
           - einops>=0.6.0
           - kornia>=0.6.9
           - lightning>=2.0.9
-          - matplotlib>=3.8.1
+          - matplotlib>=3.8.1,<3.9.1
```
adamjstewart, line 27: Note that this will be fixed once https://github.com/matplotlib/matplotlib/pull/28518 is merged

adamjstewart (commented):
Still confused by some of the dataset changes, but all datamodule changes now look good. I'll see if I can figure out why the tests are failing.

torchgeo/datasets/cyclone.py
adamjstewart: Not opposed to supporting both, but maybe save that for a follow-up PR if that isn't required by this PR.

adamjstewart:
> But tests are failing for most GeoDataModules.

I wish it were this simple... Note that some failing tests are for NonGeoDataModules (`ssl4eo_l_benchmark_*`). 

Also note that some tests fail depending on batch size. For example, `gid15` passes with `batch_size=1`, but fails if `batch_size=2`. But there are other datasets for which both 1 and 2+ pass. For the case of gid15, it seems like this is because the image is 1x1, so this may be a red herring.

Most of the failing tests seem to be because the ground truth mask has a channel dimension, which cross-entropy loss does not like. Let me see if I can fix a few.

torchgeo/datamodules/chesapeake.py
```diff
@@ -94,7 +91,7 @@ def __init__(
         # This is a rough estimate of how large of a patch we will need to sample in
         # EPSG:3857 in order to guarantee a large enough patch in the local CRS.
         self.original_patch_size = patch_size * 3
-        kwargs['transforms'] = _Transform(K.CenterCrop(patch_size))
+        kwargs['transforms'] = _Transform(K.CenterCrop(patch_size, keepdim=True))
```
adamjstewart, line 94: If you have time, see if you can replace `_Transform` with `K.AugmentationSequential`. It should be possible now with `keepdim=True`.

torchgeo/datasets/cyclone.py
ashnair1: I don't recall us specifying datasets only support torchvision-style transforms and thought we were moving towards kornia. 

This was done because when a batch is passed through our kornia augmentations in `on_batch_after_transfer` it can only have kornia compatible keys.

ashnair1: With kornia/kornia#2971, we should be able to pass keys and have kornia only act on the kornia compatible keys and ignore the rest.

torchgeo/datasets/cyclone.py
```diff
@@ -96,17 +96,19 @@ def __getitem__(self, index: int) -> dict[str, Any]:
...
         }
 
         image_id = self.labels.iat[index, 0]
-        sample['image'] = self._load_image(image_id)
+        sample = {'image': self._load_image(image_id)}
+        sample['label'] = features['label']
 
         if self.transforms is not None:
             sample = self.transforms(sample)
+            sample.update({x: features[x] for x in features if x != 'label'})
```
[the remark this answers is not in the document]
ashnair1: Reason for this mentioned here- https://github.com/microsoft/torchgeo/pull/2147#discussion_r1685425105

These type of fixes won't be required once kornia/kornia#2971 makes it to a release. It allows non-kornia compatible keys to be passed.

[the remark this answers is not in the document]
adamjstewart: Hmm, this kind of seems like a hack. It's an implementation detail that `TropicalCycloneDataModule` doesn't use `TropicalCyclone(..., transform=...)`, not by design. I definitely don't think we should return different things depending on whether `transforms` is used or not.

I have a better idea. In `BaseDataModule` in `transfer_batch_to_device`, drop all keys that Kornia doesn't recognize. This will allow us to drop most changes to datasets and the `transfer_batch_to_device` in `GeoDataModule`. You can loop over all keys and drop any that don't correspond to a valid Kornia key (there should be a way to get a list of valid keys from Kornia).

adamjstewart:
Is this waiting on Kornia 0.7.4? Trying to decide if this should be in TorchGeo 0.6.0 next week or if we should bump it to a later release.

robmarkcole:
Kornia 0.7.4. is out :-)

adamjstewart:
Yep, reviewing this now...

pyproject.toml
```diff
@@ -40,8 +40,8 @@ dependencies = [
     "einops>=0.3",
     # fiona 1.8.21+ required for Python 3.10 wheels
     "fiona>=1.8.21",
-    # kornia 0.7.3+ required for instance segmentation support in AugmentationSequential
-    "kornia>=0.7.3",
+    # kornia 0.7.4+ required for using native AugmentationSequential, allowing us to replace our custom implementation
```
adamjstewart, line 43:
```suggestion
    # kornia 0.7.4+ required for AugmentationSequential support for unknown keys
```
No need to mention something that will no longer exist

tests/datasets/test_chesapeake.py
```diff
@@ -212,6 +212,9 @@ def test_plot(self, dataset: ChesapeakeCVPR) -> None:
         plt.close()
         dataset.plot(x, show_titles=False)
         plt.close()
-        x['prediction'] = x['mask'][:, :, 0].clone().unsqueeze(2)
+        if x['mask'].ndim == 2:
```
adamjstewart, line 215: Can you explain this change?

adamjstewart (commented):
This looks really clean now, thanks for all the hard work on this!

I think the only other place we use our wrapper is in the transforms tutorial? Maybe we can remove that in a separate PR so we don't hold this up.

tests/datasets/test_chesapeake.py
adamjstewart: Okay, so now we either load a H x W mask, or a C x H x W mask. @nilsleh was also asking for support for multi-channel masks in our trainer. There's an open TODO. We'll figure that out someday.

adamjstewart (approved):

PR #2154, opened 2024-07-05
Added support for pathlib

pioneerHitesh:
Pathlib support has been added to resolve issue #1616.


pioneerHitesh:
@adamjstewart  I have added pathlib support for ``datasets/geo.py``. `Bytes` support is not added because of its incompatibility with ``os.path.join``.

adamjstewart:
I agree with not adding `bytes` support because torchvision didn't do it either and we would first have to convince them.

I like the use of `TypeAlias` and think that's what we should use.

At this point, I think you just need to update the rest of TorchGeo to use this new type alias. Otherwise, mypy will complain that things aren't compatible. I _think_ that will only involve `torchgeo.datasets`, but I'm not sure if there are any uses of paths in the rest of TorchGeo. Let's stick to `torchgeo.datasets` and see if we can get the tests to pass by only modifying that. Anywhere where you see `root` or `paths` should use the new alias you created.

pioneerHitesh:
@adamjstewart I have completed adding support for pathlib may you please check.

adamjstewart:
Can you resolve the conflicts? The main branch has changed quite a bit since you branched off.

adamjstewart:
No need to close, you can always use the same PR.

pioneerHitesh:
@adamjstewart  may you suggest what changes need to be done in ``cyclone.py`` to get rid of the errors

pioneerHitesh:
@adamjstewart may you please help me with the failing checks for windows-latest?

adamjstewart:
There seem to be some datasets that were missed:
```console
> grep ' root: str' *.py
chesapeake.py:        root: str = 'data',
geo.py:        root: str = 'data',
seco.py:        root: str = 'data',
seco.py:    def _load_patch(self, root: str, subdir: str) -> Tensor:
ucmerced.py:        root: str = 'data',
```
There are also many datasets containing logic like:
```python
assert isinstance(self.paths, str)
```
that should be updated to:
```python
assert isinstance(self.paths, str | pathlib.Path)
```

adamjstewart:
Btw, it's much easier for me to make changes to your PR if you create it on a separate branch as discussed in [Contributing](https://torchgeo.readthedocs.io/en/stable/user/contributing.html#git).

pioneerHitesh:
@adamjstewart I have implemented the changes. may you please take a look at it.

pioneerHitesh:
> Btw, it's much easier for me to make changes to your PR if you create it on a separate branch as discussed in [Contributing](https://torchgeo.readthedocs.io/en/stable/user/contributing.html#git).

Have done this in #2173 may you please take a look?

PR #2162, opened 2024-07-12
Update to readMe of generating testdata

sfalkena:
While generating fake rasterdata I noticed a small error in the readme that gave me an `AttributeError: 'Profile' object has no attribute 'indexes'`.


adamjstewart (approved):
Good catch!

PR #2173, opened 2024-07-17
Pathlib

pioneerHitesh:
Changed branch from ``main`` to ``pathlib`` in accordance with contributing guidelines.


adamjstewart (approved):
I think this is finally ready to merge!

First, I want to thank @pioneerHitesh for not giving up on this PR. This was a monumental effort, and I'm sure took way longer than you expected when you signed up to work on #1616 almost a year ago!

Second, I want to thank our CI for testing this PR. I wasn't convinced that older versions of torchvision would support pathlib, and was especially worried about all of our dependencies that lack type hints. But according to CI, all versions of all dependencies magically support pathlib for all purposes we use them for!

PR #2176, opened 2024-07-18
Normalization Statistics in EuroSAT

### LINKED ISSUE #2175: Normalization in Datamodules does not use correct statistics
### Description

For some datasets like EuroSAT, the normalization statistics per band are included in the datamodule. However, they are not passed in to the `self.aug` transform or others such that the image normalizations are wrong (the applied augmentations are still the default mean=0, std=255). I would expect that this also happens for other datamodules.

### Steps to reproduce

```python
from torchgeo.transforms import AugmentationSequential
from torchgeo.datamodules import EuroSATDataModule
from copy import deepcopy

ds = EuroSAT(root="/mnt/SSD2/nils/projects/vae/data", bands=["B04", "B03", "B02"])
dm = EuroSATDataModule(root="/mnt/SSD2/nils/projects/vae/data", bands=["B04", "B03", "B02"], batch_size=256)
dm.setup("fit")
train_loader = dm.train_dataloader()
orig_batch = next(iter(train_loader))


# augmentation from datamodule
print(dm.aug)
batch = dm.aug(deepcopy(orig_batch))

print("Default")
print("mean", batch["image"].mean())
print("std", batch["image"].std())

# updated augmentation
normalize_transform = AugmentationSequential(
    K.Normalize(mean=dm.mean, std=dm.std),
    data_keys=["image"]
)

batch = normalize_transform(deepcopy(orig_batch))

print("Updated")
print("mean", batch["image"].mean())
print("std", batch["image"].std())
```

```
AugmentationSequential(
  (augs): AugmentationSequential(
    (Normalize_0): Normalize(p=1.0, p_batch=1.0, same_on_batch=True, mean=0, std=255)
  )
)
Default
mean tensor(4.0712)
std tensor(1.7841)
Updated
mean tensor(0.0053)
std tensor(0.9901)
```

### Version

'0.6.0.dev0'

nilsleh:
Closes #2175 .

I searched for `self.bands = kwargs(...` because those are used when band specific normalizations are updated and EuroSAT actually seems to be the only one. All other matches have it before `__init__` or update the augmentation.

### DIFF
--- a/torchgeo/datamodules/eurosat.py
+++ b/torchgeo/datamodules/eurosat.py
@@ -94,11 +94,10 @@ class EuroSATDataModule(NonGeoDataModule):
             **kwargs: Additional keyword arguments passed to
                 :class:`~torchgeo.datasets.EuroSAT`.
         """
-        super().__init__(EuroSAT, batch_size, num_workers, **kwargs)
-
         bands = kwargs.get('bands', EuroSAT.all_band_names)
         self.mean = torch.tensor([MEAN[b] for b in bands])
         self.std = torch.tensor([STD[b] for b in bands])
+        super().__init__(EuroSAT, batch_size, num_workers, **kwargs)
 
 
 class EuroSATSpatialDataModule(NonGeoDataModule):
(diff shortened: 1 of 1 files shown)


adamjstewart (approved):
Good catch!

PR #2191, opened 2024-07-24
Add DeCUR RN50 weights

wangyi111:
Adding DeCUR RN50 weights for Sentinel-2 all bands and Sentinel-1 all bands.


torchgeo/models/resnet.py
```diff
@@ -404,6 +404,19 @@ class ResNet50_Weights(WeightsEnum):  # type: ignore[misc]
         },
     )
 
+    SENTINEL1_ALL_DECUR = Weights(
+        url='https://huggingface.co/torchgeo/decur/resolve/9328eeb90c686a88b30f8526ed757b4bc0f12027/rn50_ssl4eo-s12_sar_decur_ep100-f0e69ba2.pth',  # noqa: E501
+        transforms=_zhu_xlab_transforms,
```
adamjstewart, line 409: Just to confirm, you used the same transforms as SSL4EO-S12?

* Resize(256)
* CenterCrop(224)
* Normalize(mean=0, std=10000)

wangyi111: is this for pretraining or downstream? if for downstream it's mean/std style normalization.

adamjstewart: Downstream

wangyi111: ok then in my experiments i was using the mean/std of the downstream dataset. but I guess 0/10000 normalization could work too, haven't tested though.

adamjstewart: > is this for pretraining or downstream?

To clarify, ideally it should be for both. If we define "preprocessing" as things like normalization that are required to be the same across pre-train/fine-tune and train/test and "data augmentation" as things like random flip that should only be applied during training, then `transforms` should be the former. So the question is really what transforms should always be applied to the data to make sure it's in the right format (range, dimensions, zoom level, etc.) for the model to accept.

wangyi111: I see, like to encode good out-of-the-box representations if the raw image is similar to the pretraining dataset. I think I need to add normalization for S1. The existing `SENTINEL1_ALL_MOCO` uses `_zhu_xlab_transforms` (dividing 10000), which seems also not suitable?

adamjstewart: Correct. You're the expert on the normalizations you used for each paper, so I'll let you decide what to use. Note that `_zhu_xlab_transforms` may be poorly named in the case where you used a different set of normalizations for each paper. Feel free to change the name to `_ssl4eo_s1_transforms`, `_ssl4eo_s2_transforms`, `_decur_transforms`, etc. if they are different.

adamjstewart (dismissed):

torchgeo/models/resnet.py
```diff
@@ -13,14 +13,64 @@
 
 # https://github.com/zhu-xlab/SSL4EO-S12/blob/d2868adfada65e40910bfcedfc49bc3b20df2248/src/benchmark/transfer_classification/linear_BE_moco.py#L167
```
adamjstewart, line 14: I think these comments should be moved down to the s2 section. Maybe you can add a link pointing to where the S1 metrics are defined.

adamjstewart (approved):

PR #2199, opened 2024-07-29
Rename bbox -> bounds

ashnair1:
'bbox' in kornia augmentations is used to identify bounding boxes of instances in an image. However, in GeoDatasets, bbox refers to the geographic bounds of the data. This PR aims to make these distinct by renaming 'bbox' in GeoDatasets to 'geo_bbox'

`bounds` could also be another option.

### DIFF
--- a/tests/data/western_usa_live_fuel_moisture/data.py
+++ b/tests/data/western_usa_live_fuel_moisture/data.py
@@ -172,7 +172,7 @@ STAC = {
             'type': 'text/csv',
         },
     },
-    'bbox': [-115.8855556, 42.44111111, -115.8855556, 42.44111111],
+    'geo_bbox': [-115.8855556, 42.44111111, -115.8855556, 42.44111111],
     'collection': 'su_sar_moisture_content',
     'geometry': {'coordinates': [-115.8855556, 42.44111111], 'type': 'Point'},
     'id': 'su_sar_moisture_content_0001',
--- a/tests/data/western_usa_live_fuel_moisture/data.py
+++ b/tests/data/western_usa_live_fuel_moisture/data.py
@@ -172,7 +172,7 @@ STAC = {
             'type': 'text/csv',
         },
     },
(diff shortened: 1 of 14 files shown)


adamjstewart:
I would also be okay with removing it entirely. We currently don't use it anywhere. The original intent was that it could be used for stitching together prediction patches, but we don't do that yet.

tests/data/western_usa_live_fuel_moisture/data.py
```diff
@@ -172,7 +172,7 @@
             'type': 'text/csv',
         },
     },
-    'bbox': [-115.8855556, 42.44111111, -115.8855556, 42.44111111],
+    'bounds': [-115.8855556, 42.44111111, -115.8855556, 42.44111111],
```
adamjstewart, line 175: I'm guessing this was changed by accident?

adamjstewart (changes requested):

tests/data/western_usa_live_fuel_moisture/data.py
ashnair1: It was

adamjstewart (approved):
Looks fine to me, hoping you can convince at least one more person to review the name choice so we don't end up changing it again later. Note that rtree supports both `bounds` and `bbox` for the same objects, but they have different meanings. This has always confused me.

robmarkcole:
I think `bounds` is fine, another option could be `extent`

ashnair1:
> Looks fine to me, hoping you can convince at least one more person to review the name choice so we don't end up changing it again later. Note that rtree supports both `bounds` and `bbox` for the same objects, but they have different meanings. This has always confused me.

Looks like rtree's [bbox](https://github.com/Toblerity/rtree/blob/184270d7540ba5de12f7518f77c9b5a3e857d6c0/rtree/index.py#L1370) is just an interleaved bounds. We could also consider 'query' as another option.

adamjstewart:
Based on popular vote, I think we'll stick with `bounds`.

PR #2202, opened 2024-07-31
Add MMEarth dataset

nilsleh:
This PR adds the [MMEarth](https://vishalned.github.io/mmearth/) dataset. Implementation is tested with the 100k version under the assumption that the dataset format is identical for the other versions.

There are some expected [changes to the dataset](https://github.com/vishalned/MMEarth-train/issues/5) so still a draft. But opening a PR already for some discussion points:

- As a multi modal dataset there are various configuration and selection schemes. The dataset has arguments for modality and band selection within modalities (the current implementation is a suggestion, but open for cleaner suggestions)
- For that same reason, we don't have a classic "image" and "label" sample dictionary because there are various ways of using this dataset across classification, segmentation and pixelwise regression. Is it reasonable to expect people to write `transform` functions that would define the sample as needed for their models?
- Normalization statistics are provided with the dataset. Normally we do standardization in the DataModule, however, given the complexity of various modalities and also two normalization schemes (z-score and min-max) I think it makes the dataset more useful when supplying this functionality in the dataset already
- The most complex part is handling the different modalities because some of them require some extra handling as seen [here](https://github.com/vishalned/MMEarth-train/blob/8d6114e8e3ccb5ca5d98858e742dac24350b64fd/mmearth_dataset.py#L69)
- Additionally, need to decide what to do with the nan-values and clearly comment that
- currently this is just an implementation to support a non-geo dataset with standard index sampling, for other indexing schemes we can write an inherited class that also has a script that generates the needed parquet file and download it from HF
- probably things I am missing so feel free to add @adamjstewart @ando-shah


nilsleh:
@adamjstewart and @ando-shah I think as a basic Dataset this does the job. For any more specific requirements we can subclass this to support other model specific needs.

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,505 @@
...
+    .. note::
+
+       This dataset requires the following additional library to be installed:
+
+       * `h5py <https://pypi.org/project/h5py/>`_ to load the dataset
+
+    .. versionadded:: 0.6
+    """
+
+    ds_versions = ['MMEarth', 'MMEarth64', 'MMEarth100k']
```
adamjstewart, line 75: Use tuples instead of lists to make ruff happy

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,505 @@
...
+
+    filenames = {
```
adamjstewart, line 77:
```suggestion
    filenames: ClassVar[dict[str, str]] = {
```

adamjstewart (commented):
I added a new test to ruff

adamjstewart:
Planning to get a working draft of SatlasPretrain first and then we can modify this PR to ensure it matches and meets @ando-shah's needs. These will be some of our first multi-modal datasets, so want to make sure we're doing things consistently (at least at an external user-facing API level).

docs/api/non_geo_datasets.csv
```diff
@@ -24,6 +24,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `LoveDA`_,S,Google Earth,"CC-BY-NC-SA-4.0","5,987",7,"1,024x1,024",0.3,RGB
 `MapInWild`_,S,"Sentinel-1/2, ESA WorldCover, NOAA VIIRS DNB","CC-BY-4.0",1018,1,1920x1920,10--463.83,"SAR, MSI, 2020_Map, avg_rad"
 `Million-AID`_,C,Google Earth,-,1M,51--73,,0.5--153,RGB
+`MMEarth`_,"C, S", Google Earth,"CC-BY-4.0","100K - 1M",,"128x128 or 64x64","multiple modalities",MSI
```
adamjstewart, line 27:
```suggestion
`MMEarth`_,"C, S", Google Earth,"CC-BY-4.0","100K--1M",,"128x128 or 64x64","multiple modalities",MSI
```
This will create an en-dash for ranges.

tests/datasets/test_mmearth.py
```diff
@@ -0,0 +1,127 @@
...
+from _pytest.fixtures import SubRequest
+
+from torchgeo.datasets import DatasetNotFoundError, MMEarth
+
+pytest.importorskip('h5py', minversion='3.6')
+
+data_dir_dict = {
+    'MMEarth': 'tests/data/mmearth/data_1M_v001/',
+    'MMEarth64': 'tests/data/mmearth/data_1M_v001_64/',
+    'MMEarth100k': 'tests/data/mmearth/data_100k_v001/',
```
adamjstewart, line 20: Do these need os.path.join so it uses \ on Windows? Or does `Path` handle that internally?

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,505 @@
...
+import torch
+from torch import Tensor
+
+from .errors import DatasetNotFoundError
+from .geo import NonGeoDataset
+from .utils import Path, lazy_import
+
+
+class MMEarth(NonGeoDataset):
+    """MM-Earth dataset.
```
adamjstewart, line 21:
```suggestion
    """MMEarth dataset.
```
Same elsewhere, this is the official dataset name

torchgeo/datamodules/mmearth.py
```diff
@@ -0,0 +1,51 @@
...
+
+        Args:
+            batch_size: Size of each mini-batch.
+            num_workers: Number of workers for parallel data loading.
+            **kwargs: Additional keyword arguments passed to
+                :class:`~torchgeo.datasets.MMEarth`.
+        """
+        super().__init__(MMEarth, batch_size, num_workers, **kwargs)
+
+    def on_after_batch_transfer(
```
adamjstewart, line 39: What is the purpose of this override? If you don't want any augs, that should be much easier to override.

torchgeo/datamodules/mmearth.py
adamjstewart, line 1: I'm dubious of the value of a data module that isn't compatible with any of our trainers, but it also isn't hurting anything.

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,505 @@
...
+
+    There are three different versions of the dataset, that vary in image size
+    and the number of tiles:
+
+    * MMEarth: 128x128 px, 1.2 M tiles, 579 GB
+    * MMEarth64: 64x64 px, 1.2 M tiles, 162 GB
+    * MMEarth100k: 128x128 px, 100 K tiles, 48 GB
```
adamjstewart, line 28:
```suggestion
    * MMEarth: 128x128 px, 1.2M tiles, 579 GB
    * MMEarth64: 64x64 px, 1.2M tiles, 162 GB
    * MMEarth100k: 128x128 px, 100K tiles, 48 GB
```
these aren't units

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,505 @@
...
+
+    The dataset consists of 12 modalities:
+
+    * Aster: elevation and slope
+    * Biome: 14 terrestrial ecosystem categories
+    * ETH Canopy Height: Canopy height and standard deviation
+    * Dynamic World: 9 landcover categories
+    * Ecoregion: 846 ecoregion categories
+    * era5: Climate reanalysis data for temperature mean, min, and max of [year, month, previous month]
```
adamjstewart, line 37:
```suggestion
    * ERA5: Climate reanalysis data for temperature mean, min, and max of [year, month, previous month]
```
Since everything else is capitalized

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,505 @@
...
+    * Sentinel-2 Cloudmask: Sentinel-2 cloud mask
+    * Sentinel-2 Cloud probability: Sentinel-2 cloud probability
+    * Sentinel-2 SCL: Sentinel-2 scene classification
+
+    that are synchronized across tiles.
+
+    Dataset format:
+
+    * Dataset in single HDF5 file
+    * Json files for band statistics, splits, and tile information
```
adamjstewart, line 56:
```suggestion
    * JSON files for band statistics, splits, and tile information
```

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,505 @@
...
+
+    For additional information, as well as bash scripts to
+    download the data, please refer to the
+    `official repository <https://github.com/vishalned/MMEarth-data?tab=readme-ov-file#data-download>`_.
```
adamjstewart, line 60: If we have time, let's add download support

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,505 @@
...
+    .. note::
+
+       This dataset requires the following additional library to be installed:
+
+       * `h5py <https://pypi.org/project/h5py/>`_ to load the dataset
+
+    .. versionadded:: 0.6
+    """
+
+    ds_versions: ClassVar[list[str]] = ['MMEarth', 'MMEarth64', 'MMEarth100k']
```
adamjstewart, line 75: Let's use tuples instead of lists

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,505 @@
...
+
+    filenames: ClassVar[dict[str, str]] = {
+        'MMEarth': 'data_1M_v001',
+        'MMEarth64': 'data_1M_v001_64',
+        'MMEarth100k': 'data_100k_v001',
+    }
+    splits: ClassVar[list[str]] = ['train', 'val', 'test']
+
+    all_modalities: ClassVar[Sequence[str]] = (
```
adamjstewart, line 84: A normal tuple is fine here. Function inputs should be as broad as possible. Function outputs (and class attributes) should be as narrow as possible.

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,505 @@
...
+        modality_bands: dict[str, list[str]] | None = None,
+        split: str = 'train',
+        normalization_mode: str = 'z-score',
+        transforms: Callable[[dict[str, Tensor]], dict[str, Tensor]] | None = None,
+    ) -> None:
+        """Initialize the MM-Earth dataset.
+
+        Args:
+            root: root directory where dataset can be found
+            ds_version: one of "MMEarth", "MMEarth64", or "MMEarth100k"
```
adamjstewart, line 189: How much added complexity was it to support all 3? If it wasn't hard, great! If it's annoying, we could stick to the big version only.

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,505 @@
...
+        'biome': 255,
+        'eco_region': 65535,
+    }
+
+    norm_modes: ClassVar[list[str]] = ['z-score', 'min-max']
+
+    def __init__(
+        self,
+        root: Path = 'data',
+        ds_version: str = 'MMEarth',
```
adamjstewart, line 178: No need to prefix dataset parameters with `ds_`. I'm a little worried about calling this version because there is also a 001 version that may change as well. Not sure what a good alternative would be though.

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,505 @@
...
+
+        self.indices = self._load_indices()
+        self.band_stats = self._load_normalization_stats()
+        self.tile_info = self._load_tile_info()
+
+    def _verify(self) -> None:
+        """Verify the dataset.
+
+        Raises:
+            AssertionError: if dataset files are not found
```
adamjstewart, line 241: Not true. I usually move the errors to the `__init__` docstring since `_verify` won't be included in the docs.

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,505 @@
...
+            os.path.join(
+                self.root, self.filenames[self.ds_version], self.band_stats_filename
+            )
+        ) as f:
+            band_stats = json.load(f)
+
+        return cast(dict[str, dict[str, float]], band_stats)
+
+    def _load_tile_info(self) -> dict[str, dict[str, str]]:
+        """Load tile information."""
```
adamjstewart, line 288: Missing documentation of return value

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,505 @@
...
+                data = (data - min_val[:, None, None]) / (
+                    max_val[:, None, None] - min_val[:, None, None]
+                )
+            else:
+                data = (data - min_val) / (max_val - min_val)
+
+        return data
+
+    def __len__(self) -> int:
+        """Return the length of the dataset."""
```
adamjstewart, line 504: Missing documentation of return value

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,505 @@
...
+            AssertionError: if ``modalities`` is not a sequence
+            ValueError: if an invalid modality name is provided
+        """
+        # validate modalities
+        assert isinstance(modalities, Sequence), "'modalities' must be a sequence"
+        for modality in modalities:
+            if modality not in self.all_modalities:
+                raise ValueError(f"'{modality}' is an invalid modality name.")
+
+    def _validate_modality_bands(self, modality_bands: dict[str, list[str]]) -> None:
```
adamjstewart, line 314: Validation is good, but not if it makes the code too complex.

adamjstewart (changes requested):

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,505 @@
...
+                self.root, self.filenames[self.ds_version], self.dataset_filename
+            ),
+            'r',
+        ) as f:
+            name = f['metadata'][ds_index][0].decode('utf-8')
+            tile_info = self.tile_info[name]
+            l2a = tile_info['S2_type'] == 'l2a'
+            for modality in self.modalities:
+                data = f[modality][ds_index][:]
+                sample[modality] = self._process_modality(data, modality, l2a)
```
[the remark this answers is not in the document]
nilsleh: Okay, I think we can provide a guess on a default, but for some modalities like elevation or vegetation height, they might be used as either "image" or "mask".

torchgeo/datasets/mmearth.py
nilsleh: How about "variant", or "subset"

torchgeo/datasets/mmearth.py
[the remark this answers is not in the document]
adamjstewart: We just need a good default. Someone can customize/override/transform the dataset if they have a different use case. Check other datasets and see what defaults they used.

torchgeo/datasets/mmearth.py
adamjstewart: I like `subset`

nilsleh:
Ando requested that Sentinel 1 should be separated into ascending and descending. That turned out to be a bit tricky, since MMEarth only stores "sentinel1" as a concatenation of ascending and descending, but I think I implemented it correctly to support band selection and normalization etc

nilsleh:
As another point regarding Sentinel1:

- Sentinel 1 ascending and descending are stored separately, however, in the hdf5 file, there is just a single sentinel1 key that always has a shape of [8, 128, 128] so if one of ascending or descending or a subset of bands within the two is not available it is still stored as [1, 128,128] of nans in the dataset
- in order to not return Nans from the `__getitem__` , only non-nan data is now returned, i.e. if I request `{'sentinel1_asc': ['VV', 'VH', 'HH', 'HV']}` in the dataset init but for the particular tile only `{'sentinel1_asc': ['VV', 'VH']}` is available, then a [2, 128, 128] tensor will be returned
- additionally, the sample contains a dict with the sensor and bands actually found in this particular sample

docs/api/datasets/non_geo_datasets.csv
```diff
@@ -28,6 +28,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `LoveDA`_,S,Google Earth,"CC-BY-NC-SA-4.0","5,987",7,"1,024x1,024",0.3,RGB
 `MapInWild`_,S,"Sentinel-1/2, ESA WorldCover, NOAA VIIRS DNB","CC-BY-4.0",1018,1,1920x1920,10--463.83,"SAR, MSI, 2020_Map, avg_rad"
 `Million-AID`_,C,Google Earth,-,1M,51--73,,0.5--153,RGB
+`MMEarth`_,"C, S","Aster, Sentinel, ERA5","CC-BY-4.0","100K--1M",,"128x128 or 64x64","multiple modalities",MSI
```
adamjstewart, line 31: Can we list out the resolutions instead of just saying "multiple modalities"?

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,595 @@
...
+            'sentinel2_cloudmask',
+            'sentinel2_cloudprod',
+            'sentinel2_scl',
+        ]:
+            tensor = torch.from_numpy(data.astype(np.int32)).long()
+
+        # TODO: tensor might still contain nans, how to handle this?
+        return tensor
+
+    def _normalize_modality(
```
adamjstewart, line 552: I think this should be a datamodule transform, not hard-coded into the dataset.

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,595 @@
...
+    .. note::
+
+       This dataset requires the following additional library to be installed:
+
+       * `h5py <https://pypi.org/project/h5py/>`_ to load the dataset
+
+    .. versionadded:: 0.7
+    """
+
+    subsets: ClassVar[tuple[str, ...]] = ('MMEarth', 'MMEarth64', 'MMEarth100k')
```
adamjstewart, line 75: No need to type hint here, the type can be automatically detected

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,595 @@
...
+
+    filenames: ClassVar[dict[str, str]] = {
+        'MMEarth': 'data_1M_v001',
+        'MMEarth64': 'data_1M_v001_64',
+        'MMEarth100k': 'data_100k_v001',
+    }
+
+    all_modalities: ClassVar[tuple[str, ...]] = (
```
adamjstewart, line 83: Same here

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,595 @@
...
+        Returns:
+            processed data
+        """
+        # band selection for modality
+        indices = self._select_indices_for_modality(modality, bands)
+        data = data[indices, ...]
+
+        # See https://github.com/vishalned/MMEarth-train/blob/8d6114e8e3ccb5ca5d98858e742dac24350b64fd/mmearth_dataset.py#L69
+        if modality == 'dynamic_world':
+            # first replacs 0 with nan then assign new labels to have 0-index classes
```
adamjstewart, line 493:
```suggestion
            # first replace 0 with nan then assign new labels to have 0-index classes
```

adamjstewart (commented):
Data loader seems much more complicated than it needs to be. Does it satisfy all of Ando's requirements?

nilsleh:
> Data loader seems much more complicated than it needs to be. Does it satisfy all of Ando's requirements?

Yeah all the feedback I have gotten from him, I have incorporated. I agree it is complicated but it's also loading data from >10 modalities so not a typical single image/target dataset. He said he will try it now and give any other feedback.

torchgeo/datasets/mmearth.py
nilsleh: I think I would disagree with you here. I know that for most datasets, the normalization in torchgeo is only applied in the datamodule. I am also not a big fan of that (I think normalization and augmentation should be handled separately). Thinking ahead of training over multiple sources, normalization is unique to each data source, while things like geometric augmentations (flips etc) are applicable to all inputs. A datamodule then would have to involve all this normalization logic but for multiple cases depending on the data source, so to me this makes more sense.

docs/api/datasets/non_geo_datasets.csv
nilsleh: The paper on page 6 states ". Additionally, all pixel-level modalities were
re-projected to the Sentinel-2 10m grid, if needed, yielding a harmonized data
cube." So I guess actually everything is 10m in this dataset.

torchgeo/datasets/mmearth.py
adamjstewart: My counterarguments:

* TorchGeo is based on torchvision, which does not do normalization in the dataset, although you can specify transforms you want to apply
* Lightning intends its datamodules to be used for [splitting, transforming, normalizing, and tokenizing](https://lightning.ai/docs/pytorch/stable/data/datamodule.html#why-do-i-need-a-datamodule) the data
* Yes, normalizations are dataset-specific, but so are datamodules. We don't need generic datamodules that can be used for any dataset

I agree that multimodal datasets throw a wrench in this design, since each image needs a different normalization. I don't have a good counterargument for that, it is kind of annoying that we have to write custom transforms to handle each mode separately. I also fully agree that normalizations should be parameters so that I can specify them in my YAML file, although Lightning doesn't have a way to checkpoint these.

I don't disagree with you enough to hold up this PR, but I would like to decide on a consistent strategy for all datasets. Right now, our datasets are hundreds of lines of code, while our datamodules are only tens of lines of code. I want to try to shift as much of the complexity as possible to the datamodules if we can (for example, converting mask classes to ordinal numbers) so that the data loaders can focus on data loading and all transforms can be applied on the GPU.

docs/api/datasets/non_geo_datasets.csv
adamjstewart: Maybe @vishalned can comment (FYI, we are adding a data loader for MMEarth to TorchGeo!)

vishalned: Hey, yes exactly. every modality has been reprojected to 10m resolution. 

Also, just reading from the previous comments it seems like its a bit tricky with the nan values and the S1 bands (sorry about this). Incase the hdf5 is not useful to detect empty bands, i believe i also store some meta information regarding the bands (which are present and not) in the tile_info.json. 

Hope this helps, and let me know if you need any more clarifications

torchgeo/datasets/mmearth.py
nilsleh: Yeah I am happy to discuss this at a later stage, just as a brief answer:

- torchvision expects the normalization on the dataset level. if you take any common dataset like imagenet, mnist, cifar etc,  everyone always has to copy that same transform snippet and normalization values from some other repo and pass it to the dataset class, which is redudant and seems unnecessary
- in the documentation lightning page in their example, the normalization is defined in a `DataModule` but is actually applied inside the `Dataset` through a transform function, no `after_batch_transform` function

Thus I don't think even lightning is fully clear on this, but at the top of my head I also don't have the perfect solution for torchgeo.

ando-shah: Im happy to move the normalization function to wrappers to keep this streamlined

nilsleh: @adamjstewart would you prefer, if I write a datamodule where all the normalization logic lives?

adamjstewart: I'll let you decide. I'm okay with both. I'm just trying to find ways to keep the datasets simple.

nilsleh: Fair! I think I would keep it for now then, but open to changing it at some other point.

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,594 @@
...
+        normalization_mode: str = 'z-score',
+        transforms: Callable[[dict[str, Tensor]], dict[str, Tensor]] | None = None,
+    ) -> None:
+        """Initialize the MMEarth dataset.
+
+        Args:
+            root: root directory where dataset can be found
+            subset: one of "MMEarth", "MMEarth64", or "MMEarth100k"
+            modalities: list of modalities to load
+            modality_bands: dictionary of modality bands, see `all_modality_bands`
```
adamjstewart, line 198: At the moment, `all_modality_bands` isn't documented. Should it be?

nilsleh:
@ando-shah found a bug in my implementation. The available band names were different from the sample specific band name, for example in era5, they have the current or previous date of the sample included in the band name, so not a general name, and therefore it was not correctly retrieved.

torchgeo/datasets/mmearth.py
```diff
@@ -0,0 +1,620 @@
...
+            normalization_mode: one of "z-score" or "min-max"
+            transforms: a function/transform that takes input sample dictionary
+                and returns a transformed version
+
+        Raises:
+            AssertionError: if ``normalization_mode``, ``subset``, or ``split`` is invalid
```
adamjstewart, line 204:
```suggestion
            AssertionError: if *normalization_mode*, *subset*, or *split* is invalid
```
The Python and Sphinx docs use italics to reference parameters, trying to match them.

torchgeo/datasets/mmearth.py
nilsleh: I meant to point to the class variable `all_modality_bands`

nilsleh:
@adamjstewart @ando-shah while trying to write a subclass based on a metadata file, an annoying part is the `image_{}` and `mask_{}` naming for the sample keys, because you need to remove them again to match the meta data naming of the modalities. So since there isn't a torchgeo trainer that works out of the box on this dataset anyways, maybe we can just return the modality name itself?

nilsleh:
@admjstewart okay I think we can merge this then, and handle everything else in a subclass.

torchgeo/datasets/mmearth.py
nilsleh: mhm, I thought I could have the reference to the existing class attributes, do I need to redefine them again?

adamjstewart: The class attributes exist in the code, but not in the documentation. To make them exist in the documentation too, you need to use `#:` to describe the attribute. You don't need to define them twice.

nilsleh: Mhm, but it's all displayed in uppercase on the docs page, which will be misleading because that is not how the actual data is defined. Is there a way to control that?

adamjstewart (approved):
Still a bit more complicated than I would like, but it gets the job done.

adamjstewart:
Just realized we forgot to add a plot method to this dataset: #2377 

I'll see if any of the remote sensing seminar students are interested in adding this. If not, may ask you to add a plot after deadlines pass and you have more time.

PR #2206, opened 2024-08-02
WesternUSALiveFuelMoisture: radiant mlhub -> source coop

adamjstewart:
See #1830


torchgeo/datasets/western_usa_live_fuel_moisture.py
```diff
@@ -204,8 +201,6 @@ def __init__(
         input_features: list[str] = all_variable_names,
         transforms: Callable[[dict[str, Any]], dict[str, Any]] | None = None,
         download: bool = False,
-        api_key: str | None = None,
```
nilsleh, line 207: Does this need a versionchanged, since those arguments were removed?

adamjstewart: There is versionadded and versionchanged but no versionremoved, so I haven't been adding this to other docstrings.

nilsleh (approved):

PR #2209, opened 2024-08-06
Add GeoNRW dataset

nilsleh:
This PR adds the [GeoNRW dataset](https://ieee-dataport.org/open-access/geonrw):

- [GitHub](https://github.com/gbaier/geonrw)
- [paper](https://ieeexplore.ieee.org/document/9406194)

As it is licensed under CC-BY-4.0, I put it on [HF](https://huggingface.co/datasets/torchgeo/geonrw) for easier usage.

TODOS:
- [x]  write unit tests
- [x] automatic download and verify
- [x] check plotting because it seems off, although I followed [this](https://github.com/gbaier/geonrw/blob/main/pytorch/nrw.py)
Screenshot from 2024-08-06 14-28-02


nilsleh:
@isaaccorley Could you take a look at the plotting, and possibly find a mistake based on their code and this implementation, since there seem to be some visual discrepancies between optical and segmentation image?

torchgeo/datasets/geonrw.py
```diff
@@ -0,0 +1,334 @@
...
+from torchvision import transforms
+
+from .errors import DatasetNotFoundError
+from .geo import NonGeoDataset
+from .utils import Path, download_and_extract_archive, extract_archive
+
+
+class GeoNRW(NonGeoDataset):
+    """GeoNRW dataset.
+
```
adamjstewart, line 27: Can we add a link to the data "homepage"? https://ieee-dataport.org/open-access/geonrw or even https://github.com/gbaier/geonrw

docs/api/non_geo_datasets.csv
```diff
@@ -15,6 +15,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `FAIR1M`_,OD,Gaofen/Google Earth,"CC-BY-NC-SA-3.0","15,000",37,"1,024x1,024",0.3--0.8,RGB
 `FireRisk`_,C,NAIP Aerial,"CC-BY-NC-4.0","91,872",7,"320x320",1,RGB
 `Forest Damage`_,OD,Drone imagery,"CDLA-Permissive-1.0","1,543",4,"1,500x1,500",,RGB
+`GeoNRW`_,S,Aerial,"CC-BY-4.0","7,783",11,"1,000x1,000p",1,"RGB, DEM"
```
adamjstewart, line 18: Confirmed the license based on https://ieee-dataport.org/open-access/geonrw

adamjstewart: If p means pixels, we already have units for this column:
```suggestion
`GeoNRW`_,S,Aerial,"CC-BY-4.0","7,783",11,"1,000x1,000",1,"RGB, DEM"
```

tests/data/geonrw/data.py
```diff
@@ -0,0 +1,87 @@
...
+            Image.fromarray(seg_image).save(os.path.join(city, f'{utm_coords}_seg.tif'))
+
+
+def create_tarball(output_filename: str, source_dirs: list[str]) -> None:
+    with tarfile.open(output_filename, 'w:gz') as tar:
+        for source_dir in source_dirs:
+            tar.add(source_dir, arcname=os.path.basename(source_dir))
+
+
+def calculate_md5(filename: str) -> str:
```
adamjstewart, line 60: https://docs.python.org/3/library/hashlib.html#hashlib.file_digest works great and is much simpler. These `data.py` files don't end up in any code or tests, so it's fine to rely on newer Python features even if they aren't present in our minimum supported Python version.

tests/data/geonrw/data.py
```diff
@@ -0,0 +1,87 @@
...
+    with open(filename, 'rb') as f:
+        for chunk in iter(lambda: f.read(4096), b''):
+            hash_md5.update(chunk)
+    return hash_md5.hexdigest()
+
+
+# Main function
+def main() -> None:
+    train_cities = TRAIN_CITIES
+    test_cities = TEST_CITIES
```
adamjstewart, line 71: A lot of this `main` and `if name == main` and global variables -> local variables stuff is probably overkill

tests/data/geonrw/data.py
adamjstewart: Honestly, most of our test code doesn't even check the md5 anyway, so you might not even need to compute or monkeypatch it. I stopped doing that in most of my new dataset PRs.

tests/data/geonrw/data.py
```diff
@@ -0,0 +1,87 @@
...
+            rgb_image = np.random.randint(0, 256, (*IMAGE_SIZE, 3), dtype=np.uint8)
+            dem_image = np.random.randint(0, 256, IMAGE_SIZE, dtype=np.uint8)
+            seg_image = np.random.randint(0, len(CLASSES), IMAGE_SIZE, dtype=np.uint8)
+
+            Image.fromarray(rgb_image).save(os.path.join(city, f'{utm_coords}_rgb.jp2'))
+            Image.fromarray(dem_image).save(os.path.join(city, f'{utm_coords}_dem.tif'))
+            Image.fromarray(seg_image).save(os.path.join(city, f'{utm_coords}_seg.tif'))
+
+
+def create_tarball(output_filename: str, source_dirs: list[str]) -> None:
```
adamjstewart, line 54: shutil.make_archive is a simpler version of this and can also be very convenient

tests/data/geonrw/data.py
adamjstewart, line 1: Feel free to ignore all comments on this file. I still need to figure out if there's a convenient way to either generate fake data on the fly in our tests or at least avoid code duplication in our `data.py` files. Until then, `data.py` is the "Wild West" and anything goes.

tests/datasets/test_geonrw.py
```diff
@@ -0,0 +1,80 @@
...
+import torch
+import torch.nn as nn
+from _pytest.fixtures import SubRequest
+from pytest import MonkeyPatch
+
+import torchgeo.datasets.utils
+from torchgeo.datasets import DatasetNotFoundError, GeoNRW
+
+
+def download_url(url: str, root: str | Path, *args: str) -> None:
```
adamjstewart, line 19: See #2232

torchgeo/datamodules/geonrw.py
```diff
@@ -0,0 +1,67 @@
...
+from .utils import group_shuffle_split
+
+
+class GeoNRWDataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the GeoNRW dataset.
+
+    Implements 80/20 train/val splits based on city locations.
+    See :func:`setup` for more details.
+
+    .. versionadded: 0.6.0
```
adamjstewart, line 24:
```suggestion
    .. versionadded: 0.6
```
Should only ever be 2 digits

torchgeo/datamodules/geonrw.py
```diff
@@ -0,0 +1,67 @@
...
+from ..transforms import AugmentationSequential
+from .geo import NonGeoDataModule
+from .utils import group_shuffle_split
+
+
+class GeoNRWDataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the GeoNRW dataset.
+
+    Implements 80/20 train/val splits based on city locations.
+    See :func:`setup` for more details.
```
adamjstewart, line 22: The `setup` function docstring doesn't actually contain more details, not sure if this sentence is helpful.

torchgeo/datasets/geonrw.py
```diff
@@ -0,0 +1,334 @@
...
+    8. airport,shipyard
+    9. roads
+    10. buildings
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://ieeexplore.ieee.org/document/9406194
+
+
+    .. versionadded:: 0.6.0
```
adamjstewart, line 62:
```suggestion
    .. versionadded:: 0.6
```

torchgeo/datasets/geonrw.py
```diff
@@ -0,0 +1,334 @@
...
+    """
+
+    splits: ClassVar[list[str]] = ['train', 'test']
```
adamjstewart, line 65:
```suggestion
    splits = ('train', 'test')
```
Same below, no need to use a mutable list instead of a tuple

torchgeo/datasets/geonrw.py
```diff
@@ -0,0 +1,334 @@
...
+
```
adamjstewart, line 66: We should add a comment pointing to https://github.com/gbaier/geonrw/blob/main/pytorch/nrw.py so we know where this stuff comes from.

torchgeo/datasets/geonrw.py
```diff
@@ -0,0 +1,334 @@
...
+        'rgb': lambda utm_coords: '{}_{}_rgb.jp2'.format(*utm_coords),
+        'dem': lambda utm_coords: '{}_{}_dem.tif'.format(*utm_coords),
+        'seg': lambda utm_coords: '{}_{}_seg.tif'.format(*utm_coords),
+    }
+
+    modalities: ClassVar[list[str]] = ['rgb', 'dem', 'seg']
+
+    url: ClassVar[str] = (
+        'https://huggingface.co/datasets/torchgeo/geonrw/resolve/main/{}'
+    )
```
adamjstewart, line 158:
```suggestion
    url = 'https://hf.co/datasets/torchgeo/geonrw/resolve/3cb6bdf2a615b9e526c7dcff85fd1f20728081b7/{}'
```
Stable git commit hash for reproducibility

torchgeo/datasets/geonrw.py
```diff
@@ -0,0 +1,334 @@
...
+    def __getitem__(self, index: int) -> dict[str, Tensor]:
+        """Return an index within the dataset.
+
+        Args:
+            index: index to return
+
+        Returns:
+            data and label at that index
+        """
+        to_tensor = transforms.ToTensor()
```
adamjstewart, line 230: Is this really needed? It's just a 1-liner to use `torch.from_numpy()` right? Eventually I'm guessing they will replace transforms v1 with v2, which will be a breaking API change. Just trying to reduce our reliance on torchvision.

adamjstewart (changes requested):
Will investigate the plotting code and get back to you.

torchgeo/datamodules/geonrw.py
nilsleh: But in the setup function I can see how the train and val indices are generated from a group_shuffle_split and how that is implemented?

torchgeo/datasets/geonrw.py
nilsleh: it's the standard conversion from PIL images so I just used that. Otherwise, it would be `torch.from_numpy(np.asarray(self.readers[modality](modality_path)))`

torchgeo/datamodules/geonrw.py
adamjstewart: Yes, but if you're looking at the documentation, this link doesn't help, and if you're looking at the code, you can probably already see the implementation details. For example, see the rendered docs for https://torchgeo.readthedocs.io/en/stable/api/datamodules.html#tropical-cyclone. The link to the setup function isn't very helpful in this case.

torchgeo/datasets/geonrw.py
adamjstewart: I guess I'm okay with it, we can always change it later if the transforms API has breaking changes.

torchgeo/datasets/geonrw.py
```diff
@@ -0,0 +1,334 @@
...
+
+
+    .. versionadded:: 0.6
+    """
+
+    # Splits taken from https://github.com/gbaier/geonrw/blob/ecfcdbca8cfaaeb490a9c6916980f385b9f3941a/pytorch/nrw.py#L48
+
+    splits = ('train', 'test')
+
+    train_list: ClassVar[list[str]] = [
```
adamjstewart, line 71: Use a tuple here and elsewhere instead of a list so you can avoid ClassVar

nilsleh: But then I get mypy complaining about assignin tuples of different length `torchgeo/datasets/geonrw.py:198: error: Incompatible types in assignment (expression has type "tuple[str, ...]", variable has type "tuple[str, str, str]") [assignment]`

adamjstewart: Bleh. I encountered this a few times in #2218. You can solve this using:
```suggestion
    train_list: tuple[str, ...] = (
```
to make it clear that the length of the tuple is meaningless/variable.

torchgeo/datasets/geonrw.py
```diff
@@ -0,0 +1,334 @@
...
+
+        fig, axs = plt.subplots(
+            nrows=1, ncols=ncols, figsize=(ncols * 5, 10), sharex=True
+        )
+
+        axs[0].imshow(sample['image'].permute(1, 2, 0))
+        axs[0].axis('off')
+        axs[1].imshow(sample['dem'].squeeze(0), cmap='gray')
+        axs[1].axis('off')
+        axs[2].imshow(sample['mask'].squeeze(0), cmap=self.colormap)
```
adamjstewart, line 311: See if this helps:
```suggestion
        axs[2].imshow(sample['mask'].squeeze(0), cmap=self.colormap, vmin=0, vmax=10, interpolation='none')
```
This is what we used in L7 Irish/L8 Biome.

adamjstewart (approved):

adamjstewart:
Note to self: this dataset requires PIL built with jpeg2000 support, not just regular jpeg. We can update the Spack recipe to include this.

adamjstewart:
I successfully downloaded, loaded, and plotted the data. Seems like all TODOs are now complete.
geonrw

PR #2218, opened 2024-08-09
Ruff: enable ruff-specific rules

adamjstewart:


torchgeo/datasets/fair1m.py
```diff
@@ -175,9 +175,9 @@ class FAIR1M(NonGeoDataset):
             os.path.join('validation', 'images'),
             os.path.join('validation', 'labelXml'),
         ),
-        'test': (os.path.join('test', 'images')),
+        'test': (os.path.join('test', 'images'),),
```
adamjstewart, line 178: This was an actual bug in the dataset uncovered by mypy once I added type hints!

torchgeo/datasets/geo.py
```diff
@@ -458,7 +458,7 @@ def __init__(
                         # See if file has a color map
                         if len(self.cmap) == 0:
                             try:
-                                self.cmap = src.colormap(1)
+                                self.cmap = src.colormap(1)  # type: ignore[misc]
```
adamjstewart, line 461: Don't love ignoring this but also don't see a better solution.

PR #2219, opened 2024-08-12
Trainers: support binary, multiclass, and multilabel tasks

### LINKED ISSUE #245: Add BinarySemanticSegmentationTask to properly compute IoU
ETCI2021 is a binary segmentation task with a positive class and a background class. As such, the IoU metric should only be computed over the positive class (i.e. computing IoU), however currently the 2 class mIoU is computed.

### LINKED ISSUE #513: Regarding torchgeo BinarySemanticSegmentationTask
Hi can i know how to import BinarySemanticSegmentationTask

### LINKED ISSUE #2205: Multiclass Classification: assert num_classes >=2
### Summary

Both segmentation and object detection require that the background be included and there is currently a note on these args: `num_classes: Number of prediction classes (including the background)`. Considering every dataaset must have at least 1 class, the min value of num_classes is 2. I propose adding an assertion, to prevent people (like myself!) from forgetting this and setting `num_classes=1` for datasets with a single class. 

### Rationale

This config error has happened to me several times, and can pass silently

### Implementation

I suppose we add validation to the BaseTask init

### Alternatives

_No response_

### Additional information

_No response_

adamjstewart:
Instead of having separate trainers for binary, multiclass, and multilabel, let's create a single trainer that can handle all 3.

This applies to both Classification and Semantic Segmentation but not to our other trainers.

### Deprecation Note

`MultiLabelClassificationTask(num_classes=...)` is deprecated, use `ClassificationTask(task='multilabel', num_labels=...)` instead.

### References

These links were really helpful in making sure things are supported correctly:

* https://discuss.pytorch.org/t/single-vs-multi-label-vs-multi-class-training-best-practice/143089/3

Closes #2205 @robmarkcole
Closes #599 @kethan52 @geografif
Closes #513 @kethan52 
Closes #245 @calebrob6

### DIFF
--- a/torchgeo/trainers/classification.py
+++ b/torchgeo/trainers/classification.py
@@ -128,13 +128,13 @@ class ClassificationTask(BaseTask):
     def configure_metrics(self) -> None:
         """Initialize the performance metrics.
 
-        * :class:`~torchmetrics.classification.Accuracy`: The number of
+        * :class:`~torchmetrics.Accuracy`: The number of
           true positives divided by the dataset size. Both overall accuracy (OA)
           using 'micro' averaging and average accuracy (AA) using 'macro' averaging
           are reported. Higher values are better.
-        * :class:`~torchmetrics.classification.JaccardIndex`: Intersection
+        * :class:`~torchmetrics.JaccardIndex`: Intersection
           over union (IoU). Uses 'macro' averaging. Higher valuers are better.
-        * :class:`~torchmetrics.classification.FBetaScore`: F1 score.
+        * :class:`~torchmetrics.FBetaScore`: F1 score.
           The harmonic mean of precision and recall. Uses 'micro' averaging.
           Higher values are better.
 
--- a/torchgeo/trainers/classification.py
+++ b/torchgeo/trainers/classification.py
@@ -107,11 +107,7 @@ class ClassificationTask(BaseTask):
                 param.requires_grad = True
 
     def configure_losses(self) -> None:
-        """Initialize the loss criterion.
-
-        Raises:
-            ValueError: If *loss* is invalid.
-        """
+        """Initialize the loss criterion."""
         match self.hparams['loss']:
             case 'ce':
                 self.criterion: nn.Module = nn.CrossEntropyLoss(
--- a/torchgeo/trainers/classification.py
+++ b/torchgeo/trainers/classification.py
@@ -9,19 +9,12 @@ from typing import Any, Literal
 import kornia.augmentation as K
 import matplotlib.pyplot as plt
 import timm
-import torch
 import torch.nn as nn
 from matplotlib.figure import Figure
 from segmentation_models_pytorch.losses import FocalLoss, JaccardLoss
 from torch import Tensor
 from torchmetrics import MetricCollection
-from torchmetrics.classification import (
-    Accuracy,
-    FBetaScore,
-    JaccardIndex,
-    MultilabelAccuracy,
-    MultilabelFBetaScore,
-)
+from torchmetrics.classification import Accuracy, FBetaScore, JaccardIndex
 from torchvision.models._api import WeightsEnum
 from typing_extensions import deprecated
 
@@ -184,6 +177,12 @@ class ClassificationTask(BaseTask):
         y = batch['label']
         batch_size = x.shape[0]
         y_hat = self(x)
+
+        match self.hparams['task']:
+            case 'binary':
+                y = y.float()
+                y_hat = y_hat.squeeze(1)
+
         loss: Tensor = self.criterion(y_hat, y)
         self.log('train_loss', loss, batch_size=batch_size)
         self.train_metrics(y_hat, y)
@@ -205,6 +204,15 @@ class ClassificationTask(BaseTask):
         y = batch['label']
         batch_size = x.shape[0]
         y_hat = self(x)
+
+        match self.hparams['task']:
+            case 'binary':
+                y = y.float()
+                y_hat = y_hat.squeeze(1)
+                batch['prediction'] = (y_hat >= 0.5).long()
+            case 'multiclass':
+                batch['prediction'] = y_hat.argmax(dim=1)
+
         loss = self.criterion(y_hat, y)
         self.log('val_loss', loss, batch_size=batch_size)
         self.val_metrics(y_hat, y)
@@ -225,7 +233,6 @@ class ClassificationTask(BaseTask):
                 keepdim=True,
             )
             batch = aug(batch)
-            batch['prediction'] = y_hat.argmax(dim=-1)
             for key in ['image', 'label', 'prediction']:
                 batch[key] = batch[key].cpu()
             sample = unbind_samples(batch)[0]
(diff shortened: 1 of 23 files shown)


torchgeo/trainers/classification.py
```diff
@@ -37,7 +30,9 @@ def __init__(
         model: str = 'resnet50',
         weights: WeightsEnum | str | bool | None = None,
         in_channels: int = 3,
-        num_classes: int = 1000,
+        task: str = 'multiclass',
+        num_classes: int | None = None,
+        num_labels: int | None = None,
```
adamjstewart, line 35: The defaults here match torchmetrics. If `task='multiclass'`, only `num_classes` is used. If `task='multilabel'`, only `num_labels` is used. If `task='binary'`, both are ignored. Honestly, we could have a single `num_classes` if we want and simply use it for both.

torchgeo/trainers/classification.py
```diff
@@ -266,147 +262,3 @@ def predict_step(
         x = batch['image']
         y_hat: Tensor = self(x).softmax(dim=-1)
         return y_hat
-
-
-class MultiLabelClassificationTask(ClassificationTask):
```
adamjstewart, line 271: The _right_ thing to do would be to deprecate this first and remove it in 0.7.0. Not sure how widely used it is. Deprecation is kind of annoying because you need to change all tests to acknowledge the warning message.

adamjstewart:
I think we first need to add a multilabel semantic segmentation dataset to properly test this.

torchgeo/trainers/classification.py
```diff
@@ -179,12 +171,16 @@ def training_step(
         x = batch['image']
         y = batch['label']
         batch_size = x.shape[0]
-        y_hat = self(x)
-        loss: Tensor = self.criterion(y_hat, y)
-        self.log('train_loss', loss, batch_size=batch_size)
+        y_hat = self(x).squeeze(1)
```
adamjstewart, line 174: `task='binary'` metrics require us to squash the 1 class dimension

torchgeo/trainers/classification.py
```diff
@@ -214,12 +214,17 @@ def validation_step(
             and hasattr(self.logger, 'experiment')
             and hasattr(self.logger.experiment, 'add_figure')
         ):
-            datamodule = self.trainer.datamodule
-            batch['prediction'] = y_hat.argmax(dim=-1)
+            match self.hparams['task']:
+                case 'binary' | 'multilabel':
+                    batch['prediction'] = (y_hat >= 0.5).long()
```
adamjstewart, line 219: 0.5 is also the threshold used by torchmetrics

nilsleh: https://lightning.ai/docs/torchmetrics/stable/classification/accuracy.html#binaryaccuracy torchmetrics does this automatically, if you pass a float

adamjstewart: Yep, this part is for plotting, not for metrics.

torchgeo/trainers/segmentation.py
```diff
@@ -148,43 +150,38 @@ def configure_models(self) -> None:
...
+        """Initialize the loss criterion."""
+        ignore_index: int | None = self.hparams['ignore_index']
+        match self.hparams['loss']:
+            case 'ce':
+                ignore_value = -1000 if ignore_index is None else ignore_index
+                self.criterion: nn.Module = nn.CrossEntropyLoss(
+                    ignore_index=ignore_value, weight=self.hparams['class_weights']
+                )
+            case 'bce':
+                self.criterion = nn.BCEWithLogitsLoss()
```
adamjstewart, line 162: Is there any way we can modify this to support `ignore_index`? Maybe clever use of `pos_weight`?

calebrob6:
I like the change, however it makes me a _little_ nervous to accept this without confirming that this new class can actually train models under the three different modes (some semantic bug that the tests aren't catching). Have you tried this with a multilabel or binary problem?

torchgeo/trainers/segmentation.py
```diff
@@ -84,48 +86,48 @@ class and used with 'ce' loss.
...
-        elif model == 'fcn':
-            self.model = FCN(
-                in_channels=in_channels, classes=num_classes, num_filters=num_filters
-            )
-        else:
-            raise ValueError(
-                f"Model type '{model}' is not valid. "
-                "Currently, only supports 'unet', 'deeplabv3+' and 'fcn'."
-            )
+        match model:
```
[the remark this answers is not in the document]
adamjstewart: We don't. I want to use `Literal` type hints to more effectively document accepted values, although this makes it a little harder to raise error messages appropriately. We could still test it with invalid inputs and tell mypy to ignore the incorrect inputs if we really want better error messages here. I'm on the fence about this.

adamjstewart:
Agreed, I was hoping someone could do some quick benchmarking for me. My biggest concern is whether these loss functions and metrics agree on the threshold for binary classification (0/1 with 0.5 threshold vs. -1/1 with 0 threshold).

We have plenty of builtin binary classification/segmentation and a few multilabel classification datasets but no multilabel segmentation datasets (yet). I actually wanted to add one for multicropping (multiple crops grown in the same field during the same year) but there are no existing datasets.

I'll be busy working on Copernicus-Bench for the next few days so don't yet have time to benchmark these myself.

torchgeo/trainers/segmentation.py
isaaccorley: Probably would need to write a simple wrapper to do the masking of the loss. Something like this should work

```python
class BCE(nn.BCEWithLogitsLoss):
   def __init__(self, ignore_index, *args, **kwargs):
      super().__init__(*args, **kwargs)
      self.ignore_index = ignore_index
   
   def forward(self, input, target):
      ignore_mask = target == ignore_index
      loss = F.binary_cross_entropy_with_logits(
         input,
         target,
         self.weight,
         pos_weight=self.pos_weight,
         reduction='none'
      )
      loss = loss[~ignore_mask]

      if self.reduction = 'mean':
         loss = loss.mean()
      elif self.reduction = 'sum':
         loss = loss.sum()
      else:
         pass

      return loss
```

torchgeo/trainers/classification.py
isaaccorley: I think it's fine to be explicit

torchgeo/trainers/classification.py
```diff
@@ -179,12 +171,16 @@ def training_step(
...
         batch_size = x.shape[0]
-        y_hat = self(x)
-        loss: Tensor = self.criterion(y_hat, y)
-        self.log('train_loss', loss, batch_size=batch_size)
+        y_hat = self(x).squeeze(1)
         self.train_metrics(y_hat, y)
         self.log_dict(self.train_metrics, batch_size=batch_size)
 
+        if self.hparams['loss'] == 'bce':
+            y = y.float()
```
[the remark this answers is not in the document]
isaaccorley: yup it's for the case that you targets are probabilities like [0.2, 0.8, 0.1] etc. and not just ints [0, 1, 0]. This is common for example if you use methods like label smoothing

isaaccorley (approved):

PR #2220, opened 2024-08-12
pre-commit: avoid buggy matplotlib type hints

adamjstewart:
The `pyplot.subplots` type hints were broken in https://github.com/matplotlib/matplotlib/pull/27001 (3.9.1) and fixed in https://github.com/matplotlib/matplotlib/pull/28518 (3.9.2, not yet released). This PR prevents pre-commit from using this bad version.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
           - einops>=0.6.0
           - kornia>=0.6.9
           - lightning>=2.0.9
-          - matplotlib>=3.8.1,!=3.9.0,!=3.9.1
+          - matplotlib>=3.8.1,!=3.9.1.*
           - numpy>=1.22
           - pillow>=10.4.0
           - pytest>=6.1.2
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
           - einops>=0.6.0
           - kornia>=0.6.9
           - lightning>=2.0.9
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -24,7 +24,7 @@ repos:
           - einops>=0.6.0
           - kornia>=0.6.9
           - lightning>=2.0.9
-          - matplotlib>=3.8.1
+          - matplotlib>=3.8.1,!=3.9.1.*
```
adamjstewart, line 27: The trailing `.*` is needed because of a 3.9.1.post1 release and should skip both 3.9.1 and 3.9.1.post1 according to my interpretation of https://packaging.python.org/en/latest/specifications/version-specifiers/#version-exclusion. 3.9.1 (without a trailing `.*`) would apparently not skip 3.9.1.post1. Would be great if someone could test to confirm though.

adamjstewart:
And with perfect timing, 3.9.2 was just released. We could just bump the minimum version to that.

ashnair1:
> And with perfect timing, 3.9.2 was just released. We could just bump the minimum version to that.

Nice! Bumping min version would be best.

adamjstewart:
Will do this in the dependabot PR then. There are a few other type hints I need to make happy first.

.pre-commit-config.yaml
ashnair1: For the record, this did work 👍

Issue #1758, opened 2023-12-05
Sentinel2 Dataset behavior


calebrob6:
### Description

I have a Sentinel 2 scene with the following files (e.g. in `./test_scene/`):

```
T36KVU_20210513T073609_B01_60m.tif
T36KVU_20210513T073609_B02_10m.tif
T36KVU_20210513T073609_B03_10m.tif
T36KVU_20210513T073609_B04_10m.tif
T36KVU_20210513T073609_B05_20m.tif
T36KVU_20210513T073609_B06_20m.tif
T36KVU_20210513T073609_B07_20m.tif
T36KVU_20210513T073609_B08_10m.tif
T36KVU_20210513T073609_B09_60m.tif
T36KVU_20210513T073609_B11_20m.tif
T36KVU_20210513T073609_B12_20m.tif
T36KVU_20210513T073609_B8A_20m.tif
T36KVU_20210513T073609_TCI_10m.tif
```

I would expect any of the following to work:
```
ds = Sentinel2(
    "test_scene/",
    bands=["B01"],
)
```

```
ds = Sentinel2(
    "test_scene/",
    bands=["B01", "B02"],
)
```

```
ds = Sentinel2(
    "test_scene/",
    bands=["B01", "B02"],
    res=37
)
```

However the `filename_glob` and `filename_regex` are setup in such a way that none of the above are recognized as valid Sentinel 2 scenes.

### Steps to reproduce

see above

### Version

0.6.0.dev0

calebrob6:
Further:
```
ds = Sentinel2(
    "test_scene/",
    bands=["B01", "B02"],
    res=60,
)
```
will not throw an error, but `ds[ds.bounds]` will throw an error.

@estherrolf for visibility

calebrob6:
I think this is strange behavior as one of the points of RasterDataset is that it can resample/align different layers to the same resolution.

adamjstewart:
Was this fixed by #3043? I've also been considering only supporting .SAFE format and using GDAL's builtin SENTINEL2 driver instead of manually navigating the directory hierarchy.

PR #2222, opened 2024-08-13
GeoDataset: ignore other bands for separate files

### LINKED ISSUE #2221: GridGeoSampler resamples same image repeatedly with separate_files and multiple dates
### Description

Hi,

Whenever I try to create a dataset which uses separate files for each band and which has multiple dates available, I notice that the sampler produces a sample for every band and every date (for every file that intersects the dataset index). I could understand that we want a sample per date, but in the current approach the sampler dataset receives a BoundingBox for every band, leading to resampling the same spatial area over and over, leading to wasted compute energy during inference.

I have created a reproducable example using the Sentinel2 testfiles as an input. Right now I'm printing the sum of the tensor to show that the 2 different tensors are retrieved 4 times. Additionally, I agree with the fact that `sampler.hits` should have length 8, but `len(sampler)` should have length 2 in my opinion.

Could you tell me if this behavior is to be expected? I have been thinking on ways to go around this issue, and for sampling a dataset like Sentinel2 we could probably leverage the tile-id, but I would prefer to add a more generalizable option.

### Steps to reproduce

```
from torchgeo.samplers import GridGeoSampler
from torchgeo.datasets import stack_samples, unbind_samples

sentinel_urls = [
    "https://github.com/microsoft/torchgeo/raw/main/tests/data/sentinel2/S2A_MSIL2A_20220414T110751_N0400_R108_T26EMU_20220414T165533.SAFE/GRANULE/L2A_T26EMU_A035569_20220414T110747/IMG_DATA/R10m/T26EMU_20190414T110751_AOT_10m.jp2",
    "https://github.com/microsoft/torchgeo/raw/main/tests/data/sentinel2/S2A_MSIL2A_20220414T110751_N0400_R108_T26EMU_20220414T165533.SAFE/GRANULE/L2A_T26EMU_A035569_20220414T110747/IMG_DATA/R10m/T26EMU_20190414T110751_B02_10m.jp2",
    "https://github.com/microsoft/torchgeo/raw/main/tests/data/sentinel2/S2A_MSIL2A_20220414T110751_N0400_R108_T26EMU_20220414T165533.SAFE/GRANULE/L2A_T26EMU_A035569_20220414T110747/IMG_DATA/R10m/T26EMU_20190414T110751_B03_10m.jp2",
    "https://github.com/microsoft/torchgeo/raw/main/tests/data/sentinel2/S2A_MSIL2A_20220414T110751_N0400_R108_T26EMU_20220414T165533.SAFE/GRANULE/L2A_T26EMU_A035569_20220414T110747/IMG_DATA/R10m/T26EMU_20190414T110751_B04_10m.jp2",
    "https://github.com/microsoft/torchgeo/raw/main/tests/data/sentinel2/S2A_MSIL2A_20220414T110751_N0400_R108_T26EMU_20220414T165533.SAFE/GRANULE/L2A_T26EMU_A035569_20220414T110747/IMG_DATA/R10m/T26EMU_20190414T110751_B08_10m.jp2",
    "https://github.com/microsoft/torchgeo/raw/main/tests/data/sentinel2/S2A_MSIL2A_20220414T110751_N0400_R108_T26EMU_20220414T165533.SAFE/GRANULE/L2A_T26EMU_A035569_20220414T110747/IMG_DATA/R10m/T26EMU_20190414T110751_TCI_10m.jp2",
    "https://github.com/microsoft/torchgeo/raw/main/tests/data/sentinel2/S2A_MSIL2A_20220414T110751_N0400_R108_T26EMU_20220414T165533.SAFE/GRANULE/L2A_T26EMU_A035569_20220414T110747/IMG_DATA/R10m/T26EMU_20190414T110751_WVP_10m.jp2",
    "https://github.com/microsoft/torchgeo/raw/main/tests/data/sentinel2/S2A_MSIL2A_20220414T110751_N0400_R108_T26EMU_20220414T165533.SAFE/GRANULE/L2A_T26EMU_A035569_20220414T110747/IMG_DATA/R10m/T26EMU_20220414T110751_AOT_10m.jp2",
    "https://github.com/microsoft/torchgeo/raw/main/tests/data/sentinel2/S2A_MSIL2A_20220414T110751_N0400_R108_T26EMU_20220414T165533.SAFE/GRANULE/L2A_T26EMU_A035569_20220414T110747/IMG_DATA/R10m/T26EMU_20220414T110751_B02_10m.jp2",
    "https://github.com/microsoft/torchgeo/raw/main/tests/data/sentinel2/S2A_MSIL2A_20220414T110751_N0400_R108_T26EMU_20220414T165533.SAFE/GRANULE/L2A_T26EMU_A035569_20220414T110747/IMG_DATA/R10m/T26EMU_20220414T110751_B03_10m.jp2",
    "https://github.com/microsoft/torchgeo/raw/main/tests/data/sentinel2/S2A_MSIL2A_20220414T110751_N0400_R108_T26EMU_20220414T165533.SAFE/GRANULE/L2A_T26EMU_A035569_20220414T110747/IMG_DATA/R10m/T26EMU_20220414T110751_B04_10m.jp2",
    "https://github.com/microsoft/torchgeo/raw/main/tests/data/sentinel2/S2A_MSIL2A_20220414T110751_N0400_R108_T26EMU_20220414T165533.SAFE/GRANULE/L2A_T26EMU_A035569_20220414T110747/IMG_DATA/R10m/T26EMU_20220414T110751_B08_10m.jp2",
    "https://github.com/microsoft/torchgeo/raw/main/tests/data/sentinel2/S2A_MSIL2A_20220414T110751_N0400_R108_T26EMU_20220414T165533.SAFE/GRANULE/L2A_T26EMU_A035569_20220414T110747/IMG_DATA/R10m/T26EMU_20220414T110751_TCI_10m.jp2",
    "https://github.com/microsoft/torchgeo/raw/main/tests/data/sentinel2/S2A_MSIL2A_20220414T110751_N0400_R108_T26EMU_20220414T165533.SAFE/GRANULE/L2A_T26EMU_A035569_20220414T110747/IMG_DATA/R10m/T26EMU_20220414T110751_WVP_10m.jp2",
]

dataset = Sentinel2(
    paths = sentinel_urls,
    cache=False,
    bands=["B02", "B03", "B04", "B08"],

)

sampler = GridGeoSampler(dataset, size=(128, 128), stride=(10, 10))

# Since we specify only 4 bands, we are expecting 8 hits, 2 timestamps per band.
print(len(sampler.hits))
# Here I would expect to have 2 samples (1 per timestamp)
print(len(sampler))

dataloader = DataLoader(dataset, sampler=sampler, batch_size=1,  collate_fn=stack_samples, num_workers=8)

for batch in dataloader:
    for sample in unbind_samples(batch):
        print(sample['image'].sum())
        print(sample['image'].shape)

```

### Version

0.6.0.dev0, e973c1e3ca126f9f76573f43d8bfba910fa78efa

adamjstewart:
If the user passes a list of files including all bands, ignore everything other than the primary band so that the file doesn't end up in the index multiple times.

### Pros

The dataset length matches what you would expect, and each bounding box is not sampled multiple times. This is especially important when using a dataset splitter to avoid the same image ending up in both train and test.

### Cons

Requires the users to ensure that the `bands` parameter includes the bands in `paths`.

Fixes #2221 

@sfalkena @adriantre can you review?

### DIFF
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -295,7 +295,7 @@ class TestRasterDataset:
             ],
         ],
     )
-    @pytest.mark.filterwarnings("ignore:Could not find any relevant files")
+    @pytest.mark.filterwarnings('ignore:Could not find any relevant files')
     def test_files_separate(self, paths: str | Iterable[str]) -> None:
         assert len(Sentinel2(paths, bands=Sentinel2.rgb_bands).files) == 2
 
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -207,14 +207,27 @@ class TestGeoDataset:
 
 
 class TestRasterDataset:
+    naip_dir = os.path.join('tests', 'data', 'naip')
+    s2_dir = os.path.join(
+        'tests',
+        'data',
+        'sentinel2',
+        'S2A_MSIL2A_20220414T110751_N0400_R108_T26EMU_20220414T165533.SAFE',
+        'GRANULE',
+        'L2A_T26EMU_A035569_20220414T110747',
+        'IMG_DATA',
+        'R10m',
+    )
+
     @pytest.fixture(params=zip([['R', 'G', 'B'], None], [True, False]))
     def naip(self, request: SubRequest) -> NAIP:
-        root = os.path.join('tests', 'data', 'naip')
         bands = request.param[0]
         crs = CRS.from_epsg(4087)
         transforms = nn.Identity()
         cache = request.param[1]
-        return NAIP(root, crs=crs, bands=bands, transforms=transforms, cache=cache)
+        return NAIP(
+            self.naip_dir, crs=crs, bands=bands, transforms=transforms, cache=cache
+        )
 
     @pytest.fixture(
         params=zip(
@@ -236,34 +249,55 @@ class TestRasterDataset:
         'paths',
         [
             # Single directory
-            os.path.join('tests', 'data', 'naip'),
+            naip_dir,
             # Multiple directories
-            [
-                os.path.join('tests', 'data', 'naip'),
-                os.path.join('tests', 'data', 'naip'),
-            ],
-            # Single file
-            os.path.join('tests', 'data', 'naip', 'm_3807511_ne_18_060_20181104.tif'),
+            [naip_dir, naip_dir],
             # Multiple files
             (
-                os.path.join(
-                    'tests', 'data', 'naip', 'm_3807511_ne_18_060_20181104.tif'
-                ),
-                os.path.join(
-                    'tests', 'data', 'naip', 'm_3807511_ne_18_060_20190605.tif'
-                ),
+                os.path.join(naip_dir, 'm_3807511_ne_18_060_20181104.tif'),
+                os.path.join(naip_dir, 'm_3807511_ne_18_060_20190605.tif'),
             ),
             # Combination
-            {
-                os.path.join('tests', 'data', 'naip'),
-                os.path.join(
-                    'tests', 'data', 'naip', 'm_3807511_ne_18_060_20181104.tif'
-                ),
-            },
+            {naip_dir, os.path.join(naip_dir, 'm_3807511_ne_18_060_20181104.tif')},
         ],
     )
     def test_files(self, paths: str | Iterable[str]) -> None:
-        assert 1 <= len(NAIP(paths).files) <= 2
+        assert len(NAIP(paths).files) == 2
+
+    @pytest.mark.parametrize(
+        'paths',
+        [
+            # Single directory
+            s2_dir,
+            # Multiple directories
+            [s2_dir, s2_dir],
+            # Multiple files (single band)
+            [
+                os.path.join(s2_dir, 'T26EMU_20190414T110751_B04_10m.jp2'),
+                os.path.join(s2_dir, 'T26EMU_20220414T110751_B04_10m.jp2'),
+            ],
+            # Multiple files (multiple bands)
+            [
+                os.path.join(s2_dir, 'T26EMU_20190414T110751_B04_10m.jp2'),
+                os.path.join(s2_dir, 'T26EMU_20190414T110751_B03_10m.jp2'),
+                os.path.join(s2_dir, 'T26EMU_20190414T110751_B02_10m.jp2'),
+                os.path.join(s2_dir, 'T26EMU_20220414T110751_B04_10m.jp2'),
+                os.path.join(s2_dir, 'T26EMU_20220414T110751_B03_10m.jp2'),
+                os.path.join(s2_dir, 'T26EMU_20220414T110751_B02_10m.jp2'),
+            ],
+            # Combination
+            [
+                s2_dir,
+                os.path.join(s2_dir, 'T26EMU_20190414T110751_B04_10m.jp2'),
+                os.path.join(s2_dir, 'T26EMU_20220414T110751_B04_10m.jp2'),
+                os.path.join(s2_dir, 'T26EMU_20220414T110751_B03_10m.jp2'),
+                os.path.join(s2_dir, 'T26EMU_20220414T110751_B02_10m.jp2'),
+            ],
+        ],
+    )
+    @pytest.mark.filterwarnings("ignore:Could not find any relevant files")
+    def test_files_separate(self, paths: str | Iterable[str]) -> None:
+        assert len(Sentinel2(paths, bands=Sentinel2.rgb_bands).files) == 2
 
     def test_getitem_single_file(self, naip: NAIP) -> None:
         x = naip[naip.bounds]
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -312,7 +312,7 @@ class GeoDataset(Dataset[dict[str, Any]], abc.ABC):
                 pathname = os.path.join(path, '**', self.filename_glob)
                 files |= set(glob.iglob(pathname, recursive=True))
             elif (os.path.isfile(path) or path_is_vsi(path)) and fnmatch.fnmatch(
-                str(path), os.path.join('*', self.filename_glob)
+                str(path), f'*{self.filename_glob}'
             ):
                 files.add(path)
             elif not hasattr(self, 'download'):
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -313,7 +313,7 @@ class GeoDataset(Dataset[dict[str, Any]], abc.ABC):
                 files |= set(glob.iglob(pathname, recursive=True))
             elif (
                 os.path.isfile(path)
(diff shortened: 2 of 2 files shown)


torchgeo/datasets/geo.py
```diff
@@ -310,7 +311,11 @@ def files(self) -> list[Path]:
             if os.path.isdir(path):
                 pathname = os.path.join(path, '**', self.filename_glob)
                 files |= set(glob.iglob(pathname, recursive=True))
-            elif os.path.isfile(path) or path_is_vsi(path):
+            elif (
+                os.path.isfile(path)
+                and fnmatch.fnmatch(str(path), os.path.join('*', self.filename_glob))
```
adamjstewart, line 316: Alternative: use `fnmatch.filter` at the end. This won't result in a warning though. This could be a good thing or a bad thing.

adamjstewart: https://github.com/python/typeshed/pull/12522

sfalkena (approved):
Approving the PR, but just noting that it might be good to add a testcase with the URLs for future reference?

adamjstewart:
@adriantre is adding a bunch of new VSI tests in #1399, I'll let him test this in his PR.

adriantre:
Works as expected for zip archives too. #1399 is conflicting with this line, but I say we merge this and I incorporate it to my proposed changes.

```python
import os
import fnmatch
from pathlib import Path

archive = 'tests/data/cms_mangrove_canopy/CMS_Global_Map_Mangrove_Canopy_1665.zip!/CMS_Global_Map_Mangrove_Canopy_1665/data'

filename_glob = '**/Mangrove_agb_Angola*'

path1 = Path('zip://{archive}/Mangrove_agb_Angola.tif')
fnmatch.fnmatch(str(path1), os.path.join('*', filename_glob))
> True

path2 = Path('zip://{archive}!/Mangrove_hba95_Angola.tif')
fnmatch.fnmatch(str(path2), os.path.join('*', filename_glob))
> False
```

tests/datasets/test_geo.py
```diff
@@ -236,34 +249,55 @@ def sentinel(self, request: SubRequest) -> Sentinel2:
...
+                os.path.join(s2_dir, 'T26EMU_20190414T110751_B04_10m.jp2'),
+                os.path.join(s2_dir, 'T26EMU_20220414T110751_B04_10m.jp2'),
+                os.path.join(s2_dir, 'T26EMU_20220414T110751_B03_10m.jp2'),
+                os.path.join(s2_dir, 'T26EMU_20220414T110751_B02_10m.jp2'),
+            ],
+        ],
+    )
+    @pytest.mark.filterwarnings('ignore:Could not find any relevant files')
+    def test_files_separate(self, paths: str | Iterable[str]) -> None:
+        assert len(Sentinel2(paths, bands=Sentinel2.rgb_bands).files) == 2
```
adriantre, line 300: Could add a test that takes `bands=rgb_bands` without providing that band. This should then warn (given that `not hasattr(self, 'download')`)

adriantre: I see you provide # Multiple files (single band) which is not B02. This warns, no?

adamjstewart: `Sentinel2.filename_glob` looks for `self.bands[0]`, which in the case of `bands=rgb_bands` is B04.

adriantre: Woops! I suggest adding a test that has mismatch between the files and bands provided.

adriantre: I guess that would even raise `DatasetNotFoundError`, if nothing is found.

adriantre: How about this:

```python
@pytest.mark.parametrize(
        'paths',
        [
            # Multiple files (single band)
            [
                os.path.join(s2_dir, 'T26EMU_20190414T110751_B02_10m.jp2'),
                os.path.join(s2_dir, 'T26EMU_20220414T110751_B04_10m.jp2'),
            ],
        ],
    )
    def test_files_separate_missing_band_file(self, paths: str | Iterable[str]) -> None:
        with pytest.warns(UserWarning, match='Could not find any relevant files'):
            assert len(Sentinel2(paths, bands=Sentinel2.rgb_bands).files) == 1
```

tests/datasets/test_geo.py
```diff
@@ -236,34 +249,55 @@ def sentinel(self, request: SubRequest) -> Sentinel2:
...
+            [
+                s2_dir,
+                os.path.join(s2_dir, 'T26EMU_20190414T110751_B04_10m.jp2'),
+                os.path.join(s2_dir, 'T26EMU_20220414T110751_B04_10m.jp2'),
+                os.path.join(s2_dir, 'T26EMU_20220414T110751_B03_10m.jp2'),
+                os.path.join(s2_dir, 'T26EMU_20220414T110751_B02_10m.jp2'),
+            ],
+        ],
+    )
+    @pytest.mark.filterwarnings('ignore:Could not find any relevant files')
```
adriantre, line 298: This filterwarning is not needed.

adamjstewart: It is for `# multiple files (multiple bands)`

adriantre: Oh, sorry. So it will warn a lot if all band files are provided. We could warn a different message if `self.separate_files=True` hinting about this?

adamjstewart: We could but it just makes the function so much more complicated. Nested if-statements looking for matches to `filename_regex` but not `filename_glob`, etc.

adriantre: Yeah, it is very cumbersome.

tests/datasets/test_geo.py
adriantre: This was based on my misunderstanding that it only warns if a product has a missing band file. This is not relevant then.

sfalkena:
I had another try at this and I think I have found another example where this doesn't work and that would pledge in favour of the sampler-based approach. If I pass a list of files, only the band name is replaced from the filename in hits. However, I have the case where different bands come in different resolutions, the dataset fails to find the files. e.g where the files are something like:

```
["T43REP_20240511T053639_B01_60m.tif",
"T43REP_20240511T053639_B02_10m.tif"]
```

(for example only band 1 is inserted in the index, but in the `__get_item__` it tries to load `T43REP_20240511T053639_B02_60m.tif` which does not exist.

Right now I have created a custom dataset that puts all files in the index (for both bands in time) and a custom sampler that filters the hits based on the bounds of the hits. I will see if I can create a prototype within the geosampler in the coming days

adamjstewart:
@sfalkena Sentinel-2 is a constant thorn in my side. Can you open a separate issue/PR for that? I genuinely don't know how to handle it because the same band may appear in multiple resolutions. We'll likely need to completely override `Sentinel2.__getitem__` to search 10, 20, 60 in that order.

adamjstewart:
> Can you open a separate issue/PR for that?

Nevermind, we already have one: #1758

PR #2225, opened 2024-08-13
Fix failing Windows tests

adamjstewart:
In our Windows tests, `pytest` is failing, but the subsequent `torchgeo` command passes. On Linux and macOS, the default shell (bash) supports fail-fast (`set -x`) but this doesn't seem to work for the default shell (pwsh) on Windows. The result is that the check passes even though `pytest` fails.

On Windows, `pytest` is failing because it's finding the system `azcopy.exe`. The documentation of [shutil.which](https://docs.python.org/3/library/shutil.html#shutil.which) reveals that `PATHEXT` is used to append extensions to `shutil.which` lookups on Windows.

- [x] Use bash as the shell for all platforms to get fail-fast support
- [x] Use `azcopy.bat` to run `azcopy.py` to get `pytest` working on Windows

Let's hope the symlink doesn't add any issues for our Windows contributors.

References:

* https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell


tests/datasets/conftest.py
```diff
@@ -13,4 +13,5 @@
 def azcopy(monkeypatch: MonkeyPatch) -> Executable:
     path = os.path.dirname(os.path.realpath(__file__))
     monkeypatch.setenv('PATH', path, prepend=os.pathsep)
+    monkeypatch.setenv('PATHEXT', '.py', prepend=os.pathsep)
```
adamjstewart, line 16: Alternatively, mock `shutil.which` to return the path we want

calebrob6 (approved):
The depths that you go to in order to get tests to pass continually amazes me!! Great work!

PR #2233, opened 2024-08-19
RasterDataset: support date_str containing % character

burakekim:
A simple logic to replace %% with a placeholder string in _disambiguate_timestamp_ before parsing the date string and reverting it back afterward

### DIFF
--- a/torchgeo/datasets/utils.py
+++ b/torchgeo/datasets/utils.py
@@ -308,9 +308,13 @@ def disambiguate_timestamp(date_str: str, format: str) -> tuple[float, float]:
     Returns:
         (mint, maxt) tuple for indexing
     """
+    
+    format = format.replace('%%', 'TEMP_PERCENT_PLACEHOLDER')
+
     mint = datetime.strptime(date_str, format)
 
-    # TODO: This doesn't correctly handle literal `%%` characters in format
+    format = format.replace('TEMP_PERCENT_PLACEHOLDER', '%%')
+
     # TODO: May have issues with time zones, UTC vs. local time, and DST
     # TODO: This is really tedious, is there a better way to do this?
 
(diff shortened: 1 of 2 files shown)


adamjstewart:
Did you encounter files with % in the name?

torchgeo/datasets/utils.py
```diff
@@ -308,9 +308,12 @@ def disambiguate_timestamp(date_str: str, format: str) -> tuple[float, float]:
     Returns:
         (mint, maxt) tuple for indexing
     """
+    format = format.replace('%%', 'TEMP_PERCENT_PLACEHOLDER')
```
adamjstewart, line 311: This won't work unless `date_str` also contains `TEMP_PERCENT_PLACEHOLDER`

torchgeo/datasets/utils.py
```diff
@@ -308,9 +308,12 @@ def disambiguate_timestamp(date_str: str, format: str) -> tuple[float, float]:
...
+
     mint = datetime.strptime(date_str, format)
 
-    # TODO: This doesn't correctly handle literal `%%` characters in format
+    format = format.replace('TEMP_PERCENT_PLACEHOLDER', '%%')
```
adamjstewart, line 315: `format` isn't used after this line, no need to change it back

torchgeo/datasets/utils.py
adamjstewart: Better idea:
```python
date_str = date_str.replace('%', '')
format = format.replace('%%', '')
```

adamjstewart:
Let's add a unit test like `'2024%m', '%Y%%m'`

torchgeo/datasets/utils.py
adamjstewart: You could even do this after `strptime` and then you don't need to touch `date_str`.

adamjstewart (approved):

PR #2234, opened 2024-08-19
disambiguate_timestamp supports %%

adamjstewart:
Closes #2233 @burakekim


adamjstewart:
Nevermind, I remembered what this comment meant. In the case where a string like below is passed:
```python
>>> from datetime import datetime
>>> datetime.strptime('2024%m', '%Y%%m')
datetime.datetime(2024, 1, 1, 0, 0)
```
we will incorrectly detect that the format has month resolution (`%m`) when in reality it only has year resolution (`%Y`). I can add an xfail test for this if you want.

PR #2235, opened 2024-08-19
Add CaBuAr dataset

DarthReca:
This PR extends the ChaBuD dataset introduced in #1259. 

It is based on data presented in both [CaBuAr: California Burned Areas dataset for delineation](https://arxiv.org/abs/2401.11519) and [ChaBuD challenge](https://huggingface.co/spaces/competitions/ChaBuD-ECML-PKDD2023). Train and Validation are taken from CaBuAr, while the Test is from ChaBuD.

The files are hosted on [HuggingFace](https://huggingface.co/datasets/DarthReca/california_burned_areas). 

These are some samples taken from the dataset:

sample_0
sample_1


DarthReca:

adamjstewart:
Missing tests for the new datamodule. Usually we put these in the trainer tests, but we don't yet have a trainer for change detection (feel free to work on this if you want). For now, you'll have to create a new `tests/datamodules` test.

DarthReca:
I have added the datamodule test, but the project coverage is decreasing for some reason.

Ruff mentions some issues with the cabuar dataset file, but no changes were made, and the last run was ok.

For the datasets test, should I add another importorskip to the datamodule test?

adamjstewart:
* ruff: I just merged #2218 which adds extra checks for certain coding patterns to avoid. You can rebase/merge main to get the tests to fail locally. I can comment directly on the lines with the suggested changes if that helps.
* datasets: Copy the `pytest.importorskip` to your datamodules tests as well
* coverage: No idea what's going on here, probably a random bug. If you still see the issue after you push a couple more commits I'll investigate.

torchgeo/datasets/cabuar.py
```diff
@@ -0,0 +1,298 @@
...
+    .. note::
+
+       This dataset requires the following additional library to be installed:
+
+       * `h5py <https://pypi.org/project/h5py/>`_ to load the dataset
+
+    .. versionadded:: 0.6
+    """
+
+    all_bands = [
```
adamjstewart, line 56: Use a tuple instead of a list. Lists are mutable, and if you modify the list in one instance, it modifies the list in all instances because it's a class attribute, not an instance attribute.

torchgeo/datasets/cabuar.py
```diff
@@ -0,0 +1,298 @@
...
+        'B06',
+        'B07',
+        'B08',
+        'B8A',
+        'B09',
+        'B11',
+        'B12',
+    ]
+    rgb_bands = ['B04', 'B03', 'B02']
+    folds = {'train': [1, 2, 3, 4], 'val': [0], 'test': ['chabud']}  # type: dict[str, list[object]]
```
adamjstewart, line 71:
```suggestion
    folds: ClassVar[dict[str, list[object]]] = {'train': [1, 2, 3, 4], 'val': [0], 'test': ['chabud']}
```

torchgeo/datasets/cabuar.py
DarthReca: Changed. Typing, too

torchgeo/datasets/cabuar.py
DarthReca: Corrected according to this typing format

DarthReca:
Everything fine. Do you suggest keeping both implementations (ChaBuD and CaBuAr)?

adamjstewart:
I see no reason not to keep both. Even if the newer dataset is more useful, someone might want to reproduce the results on the previous dataset for comparison against older papers.

torchgeo/datasets/cabuar.py
```diff
@@ -0,0 +1,303 @@
...
+from .geo import NonGeoDataset
+from .utils import Path, download_url, lazy_import, percentile_normalization
+
+
+class CaBuAr(NonGeoDataset):
+    """CaBuAr dataset.
+
+    `CaBuAr <https://huggingface.co/datasets/DarthReca/california_burned_areas>`__
+    is a dataset for Change detection for Burned area Delineation and part of
+    the splits are used for the ChaBuD ECML-PKDD 2023 Discovery Challenge.
```
adamjstewart, line 26: Depending on how much code is the same, it may be possible to either subclass `ChaBuD`:
```python
from .chabud import ChaBuD

class CaBuAr(ChaBuD):
    ...
```
or create a shared abstract base class. We could either have both datasets in a single file or a different dataset in each file. It's borderline since it's more like a ChaBuD v2, but it has a unique name.

adamjstewart (changes requested):

adamjstewart:
@DarthReca do you have time to address these change requests? Trying to finalize the 0.6.0 release this week.

DarthReca:
I will start working today to resolve any issues.

tests/datamodules/test_cabuar.py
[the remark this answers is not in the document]
DarthReca: Ok, I'm trying to get this to work. If it works, I can remove the previous test, right?

tests/data/cabuar/data.py
```diff
@@ -0,0 +1,80 @@
...
+    '108ae2a9-d7d6-42f7-b89a-90bb75c23ccb_0',
+    '29413474-04b8-4bb1-8b89-fd640023d4a6_0',
+    '43f2e60a-73b4-4f33-b99e-319d892fcab4_0',
+]
+folds = (
+    [
+        random.sample(fold_mapping['train'], 1)[0],
+        random.sample(fold_mapping['train'], 1)[0],
+        random.sample(fold_mapping['train'], 1)[0],
+        random.sample(fold_mapping['train'], 1)[0],
```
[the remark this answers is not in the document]
DarthReca: Yes, clearer for sure

tests/datamodules/test_cabuar.py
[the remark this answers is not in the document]
DarthReca: The test file has been removed and works as suggested in `test_segmentation.py`

torchgeo/datasets/cabuar.py
DarthReca: Originally the data was proposed in CaBuAr, then we extended the dataset for the challenge. However, the challenge page is not working anymore.

tests/datamodules/test_cabuar.py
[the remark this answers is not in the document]
DarthReca: I am closing the issues related to the test file

tests/datasets/test_cabuar.py
```diff
@@ -0,0 +1,102 @@
...
+
+def download_url(
+    url: str, root: str | Path, filename: str, *args: str, **kwargs: str
+) -> None:
+    shutil.copy(url, os.path.join(root, filename))
+
+
+class TestCaBuAr:
+    @pytest.fixture(
+        params=zip([CaBuAr.all_bands, CaBuAr.rgb_bands], ['train', 'val', 'test'])
```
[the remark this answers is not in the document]
DarthReca: You are right. I was searching for `product` not `zip.`

torchgeo/datasets/cabuar.py
DarthReca: I can probably do the reverse if that seems appropriate: I can make ChaBuD an extension of CaBuAr since the latter deals already with more files and subsets.

adamjstewart (approved):
Everything looks fine to me now. Can you add an example figure created by the `plot` method to the PR description? This will help validate that the plot method works correctly. Feel free to make one dataset a subclass of the other if you want, but I'm also fine keeping things the way they are. Would like to merge this by the end of the day so we can prepare the release.

DarthReca:
I added some plots. If this is fine for now, everything should be ready.

PR #2236, opened 2024-08-19
Docs: emphasize multi-modal foundation models

adamjstewart:
Adds some text to help explain how to choose a set of pre-trained weights and emphasize our new multi-modal foundation models.

Preview: https://torchgeo--2236.org.readthedocs.build/en/2236/api/models.html#pretrained-weights

@wangyi111 @ShadowXZT you might be interested in reviewing this. I think GASSL, Scale-MAE, and DOFA are the only multimodal models in TorchGeo so far. Let me know if you disagree with any of the explicit/implicit definitions.


adamjstewart:
Let's merge #2038 first since both PRs edit the same files.

PR #2241, opened 2024-08-21
mypy: even more strict

adamjstewart:
Adds additional checks like `warn_unreachable` that revealed new bugs. And fixes those bugs!

I also played around with generics like:
```python
from typing import TypeVar

K = TypeVar('K')
V = TypeVar('V')

# replace all dict[Any, Any] with dict[K, V]
```
but ran into too many issues and gave up. Maybe another day.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -141,27 +141,42 @@ exclude_lines = [
     "@overload",
 ]
 
+# https://mypy.readthedocs.io/en/stable/config_file.html
 [tool.mypy]
-python_version = "3.10"
+# Import discovery
 ignore_missing_imports = true
-show_error_codes = true
-exclude = "(build|data|dist|docs/src|images|logo|logs|output)/"
+exclude = "(build|data|dist|docs/.*|images|logo|.*logs|output|requirements)/"
 
-# Strict
-warn_unused_configs = true
+# Disallow dynamic typing (TODO: work in progress)
+disallow_any_unimported = false
+disallow_any_expr = false
+disallow_any_decorated = false
+disallow_any_explicit = false
 disallow_any_generics = true
 disallow_subclassing_any = true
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -141,27 +141,42 @@ exclude_lines = [
     "@overload",
 ]
 
+# https://mypy.readthedocs.io/en/stable/config_file.html
 [tool.mypy]
-python_version = "3.10"
```
adamjstewart, line 145: This isn't particularly useful, what we really need is https://github.com/python/mypy/issues/12286

torchgeo/datasets/utils.py
```diff
@@ -473,7 +477,7 @@ def merge_samples(samples: Iterable[dict[Any, Any]]) -> dict[Any, Any]:
     return collated
 
 
-def unbind_samples(sample: dict[Any, Sequence[Any]]) -> list[dict[Any, Any]]:
+def unbind_samples(sample: MutableMapping[Any, Any]) -> list[dict[Any, Any]]:
```
adamjstewart, line 480: This should not be a sequence, this was a bug

torchgeo/transforms/transforms.py
```diff
@@ -99,7 +99,7 @@ def forward(self, batch: dict[str, Tensor]) -> dict[str, Tensor]:
 
         # Convert boxes to default [N, 4]
         if 'boxes' in batch:
-            batch['boxes'] = Boxes(batch['boxes']).to_tensor(mode='xyxy')  # type:ignore[assignment]
+            batch['boxes'] = Boxes(batch['boxes']).to_tensor(mode='xyxy')
```
adamjstewart, line 102: This bug was introduced in #1082

PR #800, opened 2022-09-28
Samplers: Add PointGeoSampler

ak3ra:
This PR attempts to implement a PointGeo Sampler. 
The `PointGeoSampler` takes in an Image dataset and a dataset containing (lat/lon) points (eg GBIF, INaturalist), finds the intersection of the two datasets and returns a sample bounding box of the image dataset which contains the lat/lon coordinates.

See #723 for design discussion

PR #2242, opened 2024-08-21
Samplers: avoid computing intersection multiple times

### LINKED ISSUE #537: Why Sampler Index intersection with roi
Why do the `Sampler`s call `self.index.intersection(tuple(self.roi), objects=True):` (line 116, single.py)?

This seems to be called in `GeoSampler` as well which they all derive but with the `dataset.index`.

adamjstewart:
This should theoretically double the speed of sampler initialization when an `roi` is passed.

Closes #537 @Modexus, sorry I didn't get to this earlier.

### DIFF
--- a/torchgeo/samplers/batch.py
+++ b/torchgeo/samplers/batch.py
@@ -32,18 +32,9 @@ class BatchGeoSampler(Sampler[list[BoundingBox]], abc.ABC):
             roi: region of interest to sample from (minx, maxx, miny, maxy, mint, maxt)
                 (defaults to the bounds of ``dataset.index``)
         """
-        if roi is None:
-            self.index = dataset.index
-            roi = BoundingBox(*self.index.bounds)
-        else:
-            self.index = Index(interleaved=False, properties=Property(dimension=3))
-            hits = dataset.index.intersection(tuple(roi), objects=True)
-            for hit in hits:
-                bbox = BoundingBox(*hit.bounds) & roi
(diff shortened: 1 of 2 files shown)


adamjstewart:
Alternatively, we could move all of the complexity of computing `hits`/`areas`/`length` to the base class to avoid code duplication as @Modexus suggested.

adamjstewart:
> Alternatively, we could move all of the complexity of computing `hits`/`areas`/`length` to the base class to avoid code duplication as @Modexus suggested.

This is tricky because datasets like `PreChippedGeoSampler` (and theoretical `PointGeoSamplers`: #800) don't have a concept of `size` or `units`, so it doesn't always make sense to add them to the base class. I'll just fix the tests and move on.

adamjstewart:
Note that this may actually be slower when an `roi` is not used. Which I think is the usual case, and makes me hesitant to merge this PR.

adamjstewart:
To summarize, the current behavior looks like:

* Average case: 1 intersection
* Worst case: 3 intersections

* Average case: 2 intersections
* Worst case: 2 intersections

Here, average case means the user did not pass in an `roi` and the bounds of the dataset are used instead, whereas worst case means an `roi` is passed. I've always cared more about the average case than the worst case, and these are both $\mathcal{O}(n)$ in the worst case anyway (albeit with different constants in front) so it doesn't make a huge difference. I think we should close this.

PR #2248, opened 2024-08-23
SatlasPretrain: add new dataset

adamjstewart:
This PR adds a data loader for the SatlasPretrain dataset.

This is a work in progress:

- [x] Basic data loader
- [x] Documentation
- [x] Tests
- [x] Switch download from requests to aws
- [x] Add all checksums
- [x] Add support for multiple images per sensor per tile
- [x] Return timestamp info
- [ ] Add support for dynamic and vector labels?
- [ ] Add support for band selection?
- [ ] Add support for time-series images?

References:

* https://github.com/allenai/satlas/blob/main/satlas/model/dataset.py
* https://github.com/allenai/satlas/blob/main/SatlasPretrain.md
* https://github.com/allenai/satlas/blob/main/Normalization.md

@favyen2 @piperwolters can you review this PR as time permits? I'm still in the process of downloading the entire dataset, so it's going to be a bit before I can actually test it myself, but wanted to open a WIP PR anyway. I have a ton of questions for you that I'll leave in-line and we can resolve as we finalize the PR. The first draft will likely have significantly limited functionality compared to your reference implementation, but that's fine for our use case. We can always expand it in the future.

@ando-shah and I are planning on heavily using your dataset for our next paper. Our specific use case requires us to only sample from tiles where all low-resolution products (S1, S2, L) are available. My current plan is to create a custom `metadata/train_lowres_matching.json` file containing the paired-down list of tiles. We can host this in our repo, or you're also welcome to include this in your `metadata.tar.gz` file once it's complete.

satlas

(yes, the original image is that washed out, the TCI product is not great)


docs/api/non_geo_datasets.csv
```diff
@@ -33,6 +33,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `ReforesTree`_,"OD, R",Aerial,"CC-BY-4.0",100,6,"4,000x4,000",0.02,RGB
 `RESISC45`_,C,Google Earth,-,"31,500",45,256x256,0.2--30,RGB
 `Rwanda Field Boundary`_,S,Planetscope,"NICFI AND CC-BY-4.0",70,2,256x256,4.7,RGB + NIR
+`SatlasPretrain`_,"C, R, S","NAIP, Landsat, Sentinel",ESA AND CC0-1.0 AND ODbL-1.0 AND CC-BY-4.0,302M,137,512,0.5--30,"SAR, MSI"
```
adamjstewart, line 36: * Task: possibly object detection and instance segmentation as well?
* License: ESA and "public domain" aren't valid SPDX license identifiers, but I'd rather list all than just say "various" or leave it blank? Depends on how the table gets formatted
* Resolution: is this correct, or was everything resampled to the nearest zoom level?

torchgeo/datasets/satlas.py
```diff
@@ -0,0 +1,700 @@
...
+
+class _Task(TypedDict, total=False):
+    BackgroundInvalid: bool
+    categories: list[str]
+    colors: list[list[int]]
+    type: str
+
+
+# https://github.com/allenai/satlas/blob/main/satlas/model/dataset.py
+TASKS: dict[str, _Task] = {
```
adamjstewart, line 32: Are y'all comfortable with us copy-n-pasting this dict from your code without license attribution? Not sure how this works from a legal perspective. So far we're only using the colors, but we may also want to use the categories. We could reformat this if needed.

torchgeo/datasets/satlas.py
```diff
@@ -0,0 +1,700 @@
...
+    deforestation to monitoring marine infrastructure. However, developing automatic
+    computer vision systems to parse these images requires a huge amount of manual
+    labeling of training data. By combining over 30 TB of satellite images with 137
+    label categories, SatlasPretrain serves as an effective pre-training dataset that
+    greatly reduces the effort needed to develop robust models for downstream satellite
+    image applications.
+
+    Reference implementation:
+
+    * https://github.com/allenai/satlas/blob/main/satlas/model/dataset.py
```
adamjstewart, line 463: Not sure if it's worth adding this in the docstring or just leaving it in the comments only. Happy to add pointers to the official codebase/data loaders for Satlas which may be preferred by some users.

torchgeo/datasets/satlas.py
```diff
@@ -0,0 +1,700 @@
...
+        ),
+        'labels': (
+            'https://ai2-public-datasets.s3.amazonaws.com/satlas/satlas-dataset-v1-labels-dynamic.tar',
+            'https://ai2-public-datasets.s3.amazonaws.com/satlas/satlas-dataset-v1-labels-static.tar',
+        ),
+        'metadata': (
+            'https://ai2-public-datasets.s3.amazonaws.com/satlas/satlas-dataset-v1-metadata.tar',
+        ),
+    }
+    # TODO
```
adamjstewart, line 504: Can anyone help me compute these MD5s? We're downloading all the non-NAIP data but likely won't be using NAIP in the near future.

torchgeo/datasets/satlas.py
```diff
@@ -0,0 +1,700 @@
...
+    md5s: ClassVar[dict[str, tuple[str, ...]]] = {
+        'landsat': (),
+        'naip': (),
+        'sentinel1': (),
+        'sentinel2': (),
+        'labels': (),
+        'metadata': (),
+    }
+
+    # NOTE: 'tci' is RGB (b04-02), not BGR (b02-04)
```
adamjstewart, line 514: Could optionally add a `bands` parameter that allows users to specify the order of spectral bands returned by the model, but so far I don't think we need this feature.

torchgeo/datasets/satlas.py
```diff
@@ -0,0 +1,700 @@
...
+        Args:
+            image: Image product.
+            col: Web Mercator column.
+            row: Web Mercator row.
+
+        Returns:
+            An image tensor.
+        """
+        bands = self.bands[image]
+        fname_glob = os.path.join(self.root, image, '*', bands[0], f'{col}_{row}.png')
```
adamjstewart, line 606: This is one pain point I noticed. It isn't clear to me how to map between entries in `train_lowres.json` to an actual file without relying on globbing (slow). Is there always a unique file for each image/band/col/row pair? Or is it possible to have multiple files from different landsat/sentinel/naip scenes at different times? If the former, it would be really helpful to have a JSON file mapping between col/row (the things in `train_lowres.json`) to the directory where this image is stored.

torchgeo/datasets/satlas.py
```diff
@@ -0,0 +1,700 @@
...
+        path_parts = list(os.path.split(glob.glob(fname_glob)[0]))
+        channels = []
+        for band in bands:
+            path_parts[-3] = band
+            with Image.open(os.path.join(*path_parts)) as img:
+                channels.append(torch.tensor(np.array(img, dtype=np.float32)))
+        return torch.cat(channels)
+
+    def _load_label(self, label: str, col: int, row: int) -> Tensor:
```
adamjstewart, line 615: TODO: add support for vector labels.

torchgeo/datasets/satlas.py
```diff
@@ -0,0 +1,700 @@
...
+            index: Index to return.
+
+        Returns:
+            Data and label at that index.
+        """
+        col, row = self.data[index]
+        sample: dict[str, Tensor] = {}
+
+        for image in self.images:
+            sample[f'image_{image}'] = self._load_image(image, col, row)
```
adamjstewart, line 584: The design decision here is to use `image_{landsat,naip,sentinel}` as the key so we can retain support for `kornia.augmentation.AugmentationSequential` auto-detecting the type of key/value pairs. Not sure how important this actually is since many augmentations like Normalize will be unique to each image, but it could be useful for augmentations like RandomCrop.

torchgeo/datasets/satlas.py
adamjstewart: > Or is it possible to have multiple files from different landsat/sentinel/naip scenes at different times?

Looks like this is the case. Not exactly sure how to interpret the `train_lowres.json` file or what a training epoch might look like since there isn't a 1-to-1 mapping between epoch index and image file.

adamjstewart:
I'm curious if anyone has ever managed to successfully download https://github.com/allenai/satlas/blob/main/satlaspretrain_urls.txt because I have been trying for weeks and the download always dies in the middle.

calebrob6:
Have you tried the aws-cli?

adamjstewart:
How do I convert these URLs to s3 equivalents?

calebrob6:
`aws s3 cp s3://ai2-public-datasets/satlas/satlas-dataset-v1-sentinel2-a.tar . --no-sign-
request`

You can also do:
`aws s3 ls s3://ai2-public-datasets/satlas/ --no-sign-request`

ando-shah:
+1  to Caleb's suggestion. Best to run in a screen / tmux - mine took a long time, and unzipping s2 took a day or so (on slower disks)!

> `aws s3 cp s3://ai2-public-datasets/satlas/satlas-dataset-v1-sentinel2-a.tar . --no-sign-request`
> 
> You can also do: `aws s3 ls s3://ai2-public-datasets/satlas/ --no-sign-request`
> 
> From working with Maxar Open Data on S3 and similar on Azure -- using `s3` and `azcopy` is much much better than `wget`ing the HTTPS URL.

adamjstewart:
Ugh, they don't have Landsat on AWS though...

torchgeo/datasets/satlas.py
favyen2: Yes that is fine with us for sure.

torchgeo/datasets/satlas.py
favyen2: The static labels are intended to be paired with an arbitrary image or time series of images from 2022 (for Sentinel-2) or 2018-2021 (for NAIP).

The dynamic labels reference a specific image/timestamp.

On each train step, it:
1. Samples a random task, weighted by the number of tiles at which labels are available for that task. Then it samples a random tile. (https://github.com/allenai/satlas/blob/55bb52fb90a9caaa91bf5fb428e76bfd911a5364/satlas/model/dataset.py#L709)
2. Loads the labels at that tile for that task. (https://github.com/allenai/satlas/blob/55bb52fb90a9caaa91bf5fb428e76bfd911a5364/satlas/model/dataset.py#L797)
3. Identifies the available images that correspond with the loaded labels, based on the validity time range of the label.
4. Samples a random subset of the images to use as inputs (1 for single-image model, 8 for Sentinel-2 multi-image model, 4 for NAIP multi-image model). For dynamic labels, there will only be one matching image though; the backbone applies shared encoder on each image and then temporal max pooling across feature maps; so it can input one image for dynamic labels but multiple images for static labels.

Due to the dataset size, we use mini epochs of a fixed duration (32768 batches), but could also train with epochs where every (task, tile) tuple is sampled. The image subset can be thought of as a data augmentation.

adamjstewart: How does 3 work? Specifically, given a col/row pair, how do you get a list of times/directories where that col/row can be found? Right now I use globbing but it's incredibly slow.

favyen2: We create a cache mapping from column/row to the list of image IDs containing that tile. This is created by reading good_images_X.json from the metadata.tar, which is a list of (col, row, imageID) tuples (`list[tuple[int, int, str]]`), and then converting it to a `dict[tuple[int, int], list[str]]`. For Sentinel-2 it is called `metadata/good_images_lowres_all.json`.

It needs to be combined with `image_times.json` to make sure the timestamp of the image is appropriate for the label though.

The `good_images_X.json` files are generated by [get_good_images.py](https://github.com/allenai/satlas/blob/main/satlas/cmd/to_dataset/get_good_images.py).

adamjstewart: Thanks, this is exactly what I was looking for!

docs/api/non_geo_datasets.csv
adamjstewart: @favyen2 can you double check these?

favyen2: The tasks are:
* Semantic segmentation
* Per-pixel regression
* Object detection (bounding box)
* Instance segmentation
* Classification

For resolution, images are up-sampled to the next higher zoom level, so in terms of the original image resolution, I think this is correct. NAIP is 60 cm/pixel, and Landsat B8 is 15 m/pixel, so to be more precise it could be `0.5--15` (if the range should include all the bands of the satellite images, then it should be `0.5--60` since Sentinel-2 has 60 m/pixel bands).

adamjstewart: Usually I just look at the most common resolution of the satellite, so Sentinel-2 is 10 m and Landsat is 30 m.

adamjstewart:
Still working on extracting files and running `get_good_images.py` on the full dataset, but this data loader seems to work on my fake data. I suggest we review/merge this as is and fix any bugs that we discover later in a separate PR.

docs/api/datasets/non_geo_datasets.csv
```diff
@@ -36,6 +36,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `ReforesTree`_,"OD, R",Aerial,"CC-BY-4.0",100,6,"4,000x4,000",0.02,RGB
 `RESISC45`_,C,Google Earth,-,"31,500",45,256x256,0.2--30,RGB
 `Rwanda Field Boundary`_,S,Planetscope,"NICFI AND CC-BY-4.0",70,2,256x256,4.7,RGB + NIR
+`SatlasPretrain`_,"C, R, S, I, OD","NAIP, Landsat, Sentinel",ESA AND CC0-1.0 AND ODbL-1.0 AND CC-BY-4.0,302M,137,512,0.6--30,"SAR, MSI"
```
nilsleh, line 39: also Sentinel2 right?

torchgeo/datasets/satlas.py
```diff
@@ -0,0 +1,770 @@
...
+class _Task(TypedDict, total=False):
+    BackgroundInvalid: bool
+    categories: list[str]
+    colors: list[list[int]]
+    type: str
+
+
+# https://github.com/allenai/satlas/blob/main/satlas/model/dataset.py
+TASKS: dict[str, _Task] = {
+    'polyline_bin_segment': {
```
nilsleh, line 33: Do you want to explicitly have these 400 lines of code here or maybe in a separate file that is part of the download? From what I can tell `TASKS` is only used in the plot function for the colors and could be loaded instead?

adamjstewart: I think it's fine, it's all copy-n-pasted anyway. Using downloadable code is bad without checksumming, although JSON might be okay. Could even shrink it if needed since we don't use all attributes.

nilsleh (approved):

PR #2249, opened 2024-08-23
Add Sample and Batch TypedDicts

adamjstewart:
See #985 for discussion.

Not sure if we'll be able to sneak this into the release or not, but it would help a lot with the uniformity of our datasets. May also have to wait until after @ashnair1's work to remove our many Kornia hacks.

### DIFF
--- a/docs/api/datasets.rst
+++ b/docs/api/datasets.rst
@@ -555,6 +555,8 @@ Utilities
 ---------
 
 .. autoclass:: BoundingBox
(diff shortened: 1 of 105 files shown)


adamjstewart:
Let's either wait for 3.10 support to end or add typing-extensions as a dependency so we can use `total=False` instead of annotating every possible key.

adamjstewart:
Closing as the main branch has diverged too much anyway.

PR #2253, opened 2024-08-25
SkyScript: add new dataset

adamjstewart:
TorchGeo's first text dataset!

skyscript

@ShadowXZT


torchgeo/datasets/skyscript.py
```diff
@@ -0,0 +1,187 @@
...
+            A dict containing image and caption at index.
+        """
+        filepath, title = self.captions.iloc[index][:2]
+
+        with Image.open(os.path.join(self.root, filepath)) as img:
+            array = np.array(img, dtype=np.float32)
+            array = rearrange(array, 'h w c -> c h w')
+            image = torch.from_numpy(array)
+
+        sample = {'image': image, 'text': title}
```
adamjstewart, line 121: Since this is our first text dataset, we have a couple of things to decide here:

* sample key: text? caption? title?
* value dtype: str? Tensor?

Also, all images are completely different sizes (ranging from 100 px to 1000 px). We should decide whether we want to crop/resize in the dataset or force the user to add a transform to do this.

nilsleh: I think "text" is the most neutral, since it could be a caption, a title, a description, or a prompt etc. In the end people need to apply a text embedding so str for now is fine, since the options for different text embeddings is large.

And I do like giving the crop/resize option at the datamodule level, to have one implementation that is "low effort" to get started.

nilsleh (approved):

torchgeo/datasets/skyscript.py
adamjstewart: I'm actually leaning towards "caption" now. For example, imagine a geo-embedded dataset with text as input and output. We need some way to distinguish between the two. But let me ask a couple NLP people and see if there is a standard w.r.t. augmentation libraries.

adamjstewart: Torchvision uses "caption" for COCO, which seems to be the gold standard dataset for image captioning: https://github.com/pytorch/vision/blob/main/torchvision/datasets/coco.py

PR #2256, opened 2024-08-27
DOFA: fix bug in patch embedding

adamjstewart:
@ando-shah discovered a minor bug (1-line) in DOFA with major consequences (performance is sensitive to band order). @ShadowXZT fixed this and is currently retraining the model. I'll update the model weights once this is done and hopefully we can sneak this in to the 0.6.0 release.


ando-shah:
This was a joint effort with @LeWaldm!

adamjstewart:
New model weights have been released. All performance metrics have significantly improved as a result of this bug fix!

PR #2270, opened 2024-09-01
GeoDataset: allow a mix of str and pathlib paths

adamjstewart:
Currently, if a user passes in both str and pathlib.Path paths, GeoDataset crashes:
```
        # Sort the output to enforce deterministic behavior.
>       return sorted(files)
E       TypeError: '<' not supported between instances of 'PosixPath' and 'str'
```
The easiest solution is to convert all values to the same type for easy comparison. This also prevents duplicates, since:
```pycon
>>> import os
>>> import pathlib
>>> x = "foo/bar"
>>> y = pathlib.Path("foo/bar")
>>> x == y
False
```

FYI @pioneerHitesh @adriantre

### DIFF
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -205,6 +205,14 @@ class TestGeoDataset:
             CustomGeoDataset(paths=paths1).files == CustomGeoDataset(paths=paths2).files
         )
 
+    def test_files_property_mix_str_and_pathlib(self, tmp_path: Path) -> None:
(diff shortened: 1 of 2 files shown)


adamjstewart:
I actually wanted to broaden things to support all of `os.PathLike`, which is how I discovered this bug. But since torchvision doesn't support `os.PathLike`, I think I'll give up on that for now.

adriantre (approved):
Looks good to me 👍

PR #2273, opened 2024-09-03
Datasets: support os.PathLike

adamjstewart:
As of #2173, all TorchGeo datasets now support both `str` and `pathlib.Path` paths. However, `pathlib.Path` is actually a subclass of `os.PathLike`, and [PEP 519](https://peps.python.org/pep-0519/#provide-specific-type-hinting-support) says to use:
```
typing.Union[str, bytes, os.PathLike]
```
to type hint all paths (we skipped `bytes` because it was a nightmare). This PR expands our support to all of `os.PathLike` (although I don't know any subclasses other than `pathlib.Path`).

@pioneerHitesh was there a reason we chose `pathlib.Path` instead of `os.PathLike`?

### DIFF
--- a/torchgeo/datasets/agrifieldnet.py
+++ b/torchgeo/datasets/agrifieldnet.py
@@ -4,7 +4,6 @@
 """AgriFieldNet India Challenge dataset."""
 
 import os
-import pathlib
 import re
 from collections.abc import Callable, Iterable, Sequence
 from typing import Any, ClassVar, cast
@@ -181,10 +180,10 @@ class AgriFieldNet(RasterDataset):
         Returns:
             data, label, and field ids at that index
         """
-        assert isinstance(self.paths, str | pathlib.Path)
+        assert isinstance(self.paths, str | os.PathLike)
 
         hits = self.index.intersection(tuple(query), objects=True)
-        filepaths = cast(list[Path], [hit.object for hit in hits])
+        filepaths = cast(list[str], [hit.object for hit in hits])
 
         if not filepaths:
             raise IndexError(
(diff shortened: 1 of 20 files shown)


torchgeo/datasets/agrifieldnet.py
```diff
@@ -181,10 +180,10 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
         Returns:
             data, label, and field ids at that index
         """
-        assert isinstance(self.paths, str | pathlib.Path)
+        assert isinstance(self.paths, str | os.PathLike)
 
         hits = self.index.intersection(tuple(query), objects=True)
-        filepaths = cast(list[Path], [hit.object for hit in hits])
+        filepaths = cast(list[str], [hit.object for hit in hits])
```
adamjstewart, line 186: As of #2270, we can reasonably expect `hit.object` to be a `str` by default.

torchgeo/datasets/utils.py
```diff
@@ -42,7 +41,7 @@
 )
 
 
-Path: TypeAlias = str | pathlib.Path
+Path: TypeAlias = str | os.PathLike[str]
```
adamjstewart, line 44: This is the actual change of interest

calebrob6 (approved):

pioneerHitesh:
@adamjstewart To the best of my knowledge there wasn't any  technical reason to choose ``pathlib.Path`` over ``os.PathLike`` .We implemented ``pathlib.Path`` support as it was mentioned in the original issue.

PR #2275, opened 2024-09-04
pytest: set default --cov, --cov-report

adamjstewart:
With this, you can now use `pytest --cov` and pytest will automatically know you want coverage specifically for `--cov=torchgeo` and `--cov-report=term-missing`. This can also easily be overridden on the command line:
```console
> pytest --cov=torchgeo.datasets.foo --cov-report=xml
```
I haven't yet found a way to enable coverage by default (without needing `--cov`) in a way that can also be overridden on the command line, but I'm pretty happy with this for now.

P.S. I also removed the 10 slowest tests reports, as we don't really pay any attention to these.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -46,7 +46,7 @@ jobs:
         run: pip list
       - name: Run pytest checks
         run: |
-          pytest --cov=torchgeo --cov-report=xml --durations=10
+          pytest --cov --cov-report=xml
           python3 -m torchgeo --help
       - name: Report coverage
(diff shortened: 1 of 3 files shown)


calebrob6:
> P.S. I also removed the 10 slowest tests reports, as we don't really pay any attention to these.

What tests were these?

adamjstewart:
You can see them on CI for the latest commit: https://github.com/microsoft/torchgeo/actions/runs/10728981926/job/29754689425

It's all model setup, especially Scale-MAE, DOFA, etc.

calebrob6 (approved):

PR #2279, opened 2024-09-04
Docs: installation-agnostic prettier usage

adamjstewart:
We shouldn't specify how users should install prettier, they can use any node package manager they want. We already have a link to the prettier docs, which explain the many different ways to install prettier.

I also prefer to specify flags (`--write`) before directories (`.`).

@Domejko

### DIFF
--- a/docs/user/contributing.rst
+++ b/docs/user/contributing.rst
@@ -111,21 +111,11 @@ Mypy won't fix your code for you, but will warn you about potential issues with 
 
 If you've never used mypy before or aren't familiar with `Python type hints <https://docs.python.org/3/library/typing.html>`_, this check can be particularly daunting. Don't hesitate to ask for help with resolving any of these warnings on your pull request.
 
-Prettier is a code formatter that helps to ensure consistent code style across a project. It supports various languages. Follow these steps to install Prettier:
-
-1. Install Node.js: Prettier is a Node.js module, so you need to have Node.js installed on your system. You can download and install Node.js from the `Node.js official website <https://nodejs.org/en>`_.
-2. Install Prettier: Use the following command to install the Prettier module in your project:
+Prettier is a code formatter that helps to ensure consistent code style across a project. It supports various languages.
 
 .. code-block:: console
 
-   $ npm install prettier --no-save
-
-
-3. Run Prettier: Use the following command to run Prettier formating:
-
-.. code-block:: console
-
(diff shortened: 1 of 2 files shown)


docs/user/contributing.rst
```diff
@@ -111,21 +111,11 @@ Mypy won't fix your code for you, but will warn you about potential issues with
...
 If you've never used mypy before or aren't familiar with `Python type hints <https://docs.python.org/3/library/typing.html>`_, this check can be particularly daunting. Don't hesitate to ask for help with resolving any of these warnings on your pull request.
 
-Prettier is a code formatter that helps to ensure consistent code style across a project. It supports various languages. Follow these steps to install Prettier:
-
-1. Install Node.js: Prettier is a Node.js module, so you need to have Node.js installed on your system. You can download and install Node.js from the `Node.js official website <https://nodejs.org/en>`_.
-2. Install Prettier: Use the following command to install the Prettier module in your project:
-
-.. code-block:: console
-
-   $ npm install prettier --no-save
```
calebrob6, line 121: Why make it harder for people to figure out how to install prettier?

calebrob6: I'd assume that a majority of our contributors are not familiar with Node.js ecosystem (as if they are trying to contribute to a python library then they know python) -- I'm certainly not. As a potential contributor I wouldn't want to have to go spend time figuring out how to install a non-python package (especially a javascript one!) just so I can make tests pass in the package that I actually want to contribute to. This block will reduce friction slightly.

adamjstewart: We already link to the prettier documentation, which includes detailed installation instructions. In all honesty, I don't expect most contributors to even need to use prettier. It's mostly just a check in CI so I don't have to tell people to add an EOL char before the EOF char. As far as I know, it's never failed on any PRs. I believe it's also automatically managed by pre-commit if people prefer that.

calebrob6: Ah so your point is something like "if installation instructions change we won't have the responsibility to update our docs"?

calebrob6 (approved):

docs/user/contributing.rst
adamjstewart: More like there are 4 different installation instructions for npm, yarn, pnpm, and bun, and I don't even use any of them.

PR #2281, opened 2024-09-04
Docs: move dataset CSV files to subdir

adamjstewart:
Figured now might be a good time to move these files like we did for the weights CSV files.


calebrob6 (approved):

calebrob6:
Sure

PR #2296, opened 2024-09-15
Add Fields Of The World field boundary delineation dataset

calebrob6:
Dataset for field boundary delineation -- https://beta.source.coop/repositories/kerner-lab/fields-of-the-world/description/

**Note**: this PR adds the pyarrow dependency as this is needed by Pandas to read parquet files.

To-Do:
- [x] Dataset
- [x] Example plots
- [x] Fake data
- [x] Tests

image

image

image


adamjstewart:
Now we just need an instance segmentation trainer.

requirements/min-reqs.old
```diff
@@ -25,6 +25,7 @@ torchvision==0.14.0
 h5py==3.6.0
 laspy==2.0.0
 opencv-python==4.5.4.58
+pyarrow==15.0.0
```
calebrob6, line 28: pyarrow<15.0.0 doesn't have an explicit dependency on numpy but fails if numpy>=2 is installed

calebrob6:
Should be good for review now

torchgeo/datasets/ftw.py
```diff
@@ -0,0 +1,363 @@
...
+            show_titles: flag indicating whether to show titles above each panel
+            suptitle: optional suptitle to use for figure
+
+        Returns:
+            a matplotlib Figure with the rendered sample
+        """
+        fig, axs = plt.subplots(nrows=1, ncols=3, figsize=(15, 5))
+
+        win_a = sample['image'][0:3].permute(1, 2, 0)
+        win_b = sample['image'][4:7].permute(1, 2, 0)
```
adamjstewart, line 328: I'm starting to want to use einops in places like this to make it more clear exactly what's happening. Optional request

adamjstewart (changes requested):

torchgeo/datasets/ftw.py
```diff
@@ -0,0 +1,363 @@
...
+
+    * images are four-channel GeoTIFFs with dimension 256x256
+    * segmentation masks (both two and three class) are single-channel GeoTIFFs
+    * instance masks are single-channel GeoTIFFs
+
+    Dataset classes:
+
+    1. background
+    2. field (two-class)
+    3. field-boundary (three-class)
```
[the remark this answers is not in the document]
calebrob6: image

Table from the paper (will be on arixv any day now). 3-class improved _object level_ metrics significantly as it causes better separation of adjacent fields, although pixel level metrics were similar between the two. I think distance transform type losses will be helpful (and actually implemented this to test out) but it was out of the scope of this paper. Hopefully others (or myself) will be able to test

torchgeo/datasets/ftw.py
calebrob6: Done

docs/api/datasets/non_geo_datasets.csv
```diff
@@ -15,6 +15,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `ETCI2021 Flood Detection`_,S,Sentinel-1,-,"66,810",2,256x256,5--20,SAR
 `EuroSAT`_,C,Sentinel-2,"MIT","27,000",10,64x64,10,MSI
 `FAIR1M`_,OD,Gaofen/Google Earth,"CC-BY-NC-SA-3.0","15,000",37,"1,024x1,024",0.3--0.8,RGB
+`Fields Of The World`_,"S,I",Sentinel-2,"Various","70795","2,3",256x256,10,MSI
```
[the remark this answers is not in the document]
calebrob6: The labels for each country dataset have a potentially different license. If you click through the links in the table shown https://beta.source.coop/repositories/kerner-lab/fields-of-the-world/ you can see the individual licenses. E.g. Austria is CC-BY-4.0 (https://beta.source.coop/repositories/kerner-lab/fields-of-the-world-austria/description/) while South Africa is CC-BY-NC-SA-4.0. The point of having the dataset download data per-country is so that the end-user can easily choose what data they want to use.

requirements/min-reqs.old
calebrob6: Hmmm does this raise the min required pandas version to 2.0 (e.g. in min-reqs.old)

adamjstewart: A good question. 2.0 is over 1 yr old, so it wouldn't be _horrible_ to make that the new min version. But we could also just throw pyarrow at any old fixed version in there for older pandas just to make sure it works. Or skip that specific test and just let pytest skip it.

calebrob6: @isaaccorley what were the issues you had with pandas>=2?

adamjstewart: I'm guessing that was an issue with numpy 2, not pandas 2?

calebrob6: Ah yep nevermind, he was mentioning "geopandas<1" and "numpy<2". I'm fine with adding pyarrow to min-reqs with a note that says to remove that when we upgrade pandas[parquet]>=2 as a min requirement.

calebrob6: Done

docs/api/datasets/non_geo_datasets.csv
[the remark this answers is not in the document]
calebrob6: FYI we will update https://beta.source.coop/repositories/kerner-lab/fields-of-the-world/ with the license for each country in the dataset, and I'll copy this information over into the constructor docstring

requirements/required.txt
```diff
@@ -9,7 +9,7 @@ lightly==1.5.12
 lightning[pytorch-extra]==2.4.0
 matplotlib==3.9.2
 numpy==2.1.1
-pandas==2.2.2
+pandas[parquet]==2.2.2
```
adamjstewart, line 12: Let's undo this change so we can test that the tests also pass even if optional dependencies like pyarrow aren't installed.

tests/datasets/test_ftw.py
```diff
@@ -0,0 +1,88 @@
...
+import matplotlib.pyplot as plt
+import pytest
+import torch
+import torch.nn as nn
+from _pytest.fixtures import SubRequest
+from pytest import MonkeyPatch
+from torch.utils.data import ConcatDataset
+
+from torchgeo.datasets import DatasetNotFoundError, FieldsOfTheWorld
+
```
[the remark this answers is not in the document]
adamjstewart: I think this is still needed if we want the tests to pass even when pyarrow isn't installed.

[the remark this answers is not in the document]
calebrob6: we need pandas parquet support for the tests to pass as the test data has parquet files

requirements/required.txt
calebrob6: I'm not understanding what exactly you are looking for here.
- We need pandas[parquet] for tests for the fields of the world dataset to pass
- If we have pandas<2 for min-reqs then we need pyarrow for the min tests to pass

adamjstewart: Correct. But TorchGeo has required dependencies and optional dependencies. Although pandas is required, `pandas[parquet]` is not. The tests should all pass regardless of whether or not optional dependencies are installed. For example, we now have 8 datasets which require h5py. But those tests don't fail if h5py isn't installed, because we use:
```python
pytest.importorskip('h5py', minversion='3.6')
```
in all of those tests. We can do the same thing for pyarrow since it's only required by one dataset.

P.S. Didn't realize this number was now up to 8, maybe h5py should be required...

tests/datasets/test_ftw.py
```diff
@@ -0,0 +1,90 @@
...
+pytest.importorskip('pyarrow', minversion='15.0.0')
```
adamjstewart, line 19: This is very new, I would use 7.0.0 because that's what pandas 2.0.0 uses, or just remove the minversion since we don't really track it.

calebrob6: Done

torchgeo/datasets/ftw.py
```diff
@@ -0,0 +1,363 @@
...
+        'rwanda',
+        'slovakia',
+        'slovenia',
+        'south_africa',
+        'spain',
+        'sweden',
+        'vietnam',
+    ]
+
+    base_url: ClassVar[str] = (
```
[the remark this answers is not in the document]
adamjstewart: Can remove this type hint

adamjstewart (approved):
Only one minor comment on unnecessary type hints, but this should be ready to go

requirements/required.txt
```diff
@@ -9,7 +9,7 @@ lightly==1.5.13
 lightning[pytorch-extra]==2.4.0
 matplotlib==3.9.2
 numpy==2.1.1
-pandas==2.2.3
+pandas[parquet]==2.2.3
```
adamjstewart, line 12: I thought we got rid of this change, did this somehow come back? Parquet isn't required, only optional.

calebrob6, line 12:
```suggestion
pandas==2.2.3
```

calebrob6: I probably messed up the rebase

adamjstewart (approved):

PR #2306, opened 2024-09-20
GH actions tutorials - Install TorchGeo from checked out repo

sfalkena:
Currently, TorchGeo is installed by the first cell of the first notebook that is run. That installs the latest release of the repository, but fails to install the current version of the repo during PR's. Therefore, nbmake fails if both a feature and a tutorial notebook get added as part of a PR at the same time. This fix installs the checked out repository instead.

Fixes issues found by #2300 and #1897


adamjstewart (approved):
I think this is the best 1-character PR I've ever seen. Can't believe we've been testing the latest release instead of the latest commit all this time. Thanks for the fix!

PR #2308, opened 2024-09-20
RasterDataset SITS support

sfalkena:
First steps towards Satellite Imagery TimeSeries support within TorchGeo (following #640). Right now I have started with the RasterDataset, but I think my methods can be applied to other GeoDatasets too. I have created a sits_dataset notebook, so feel free to run and play with an example that I prepared.

**Notes**:
- This approach expects the `filename_glob` to be generic enough to pick up on all bands when dealing with `separate_files=True`, so that all files end up in the index. This would lead to the same sampler issues from #2222. If we agree with the general setup of this PR, I will create a separate PR for those sampler fixes. (I have this internally) 
- Currently whenever `return_as_ts=False` shape [c,h,w] is returned, and when `return_as_ts=True` shape [t,c,h,w] is returned. I personally would prefer to move to the [t,c,h,w] standard, and setting `t=1` for non-ts data.
- Next to the SITS, the `__get_item__` also returns the individual dates as an array, so that any model needing the individual dates for the sample can use it.
- Currently, samplers will take the `mint` and `maxt` from the sampler index for timeseries, so that the temporal aspect of the ROI could also limit the temporal aspect. By smartly combining multiple roi's we could then make timeseries for every combination of dates I think.
- `IntersectionDataset` and `UnionDataset` take the `return_as_ts` property based on their input datasets properties.

I have currently not created any unit tests, will do so once we agree on the best structure.

### DIFF
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -12,11 +12,13 @@ import re
 import sys
 import warnings
 from collections.abc import Callable, Iterable, Sequence
+from datetime import datetime
 from typing import Any, ClassVar, cast
 
 import fiona
 import fiona.transform
 import numpy as np
+import pandas as pd
 import pyproj
 import rasterio
 import rasterio.merge
@@ -371,6 +373,9 @@ class RasterDataset(GeoDataset):
     #: True if data is stored in a separate file for each band, else False.
     separate_files = False
 
+    # Whether to return imagery as SITS or not
+    return_as_ts: bool = False
+
     #: Names of all available bands in the dataset
     all_bands: tuple[str, ...] = ()
 
@@ -485,6 +490,10 @@ class RasterDataset(GeoDataset):
                         stop = match.group('stop')
                         mint, _ = disambiguate_timestamp(start, self.date_format)
                         _, maxt = disambiguate_timestamp(stop, self.date_format)
+                    elif self.return_as_ts:
+                        warnings.warn(
+                            'return_as_ts = True, but no date could be found from filename_regex'
+                        )
 
                     coords = (minx, maxx, miny, maxy, mint, maxt)
                     self.index.insert(i, coords, filepath)
@@ -510,6 +519,26 @@ class RasterDataset(GeoDataset):
         self._crs = cast(CRS, crs)
         self._res = cast(float, res)
 
+    def _get_regex_groups_as_df(self, filepaths: list[str]) -> pd.DataFrame:
+        """Extracts the regex metadata from a list of filepaths.
+
+        Args:
+            filepaths (list): A list of filepaths.
+
+        Returns:
+            pandas.DataFrame: A DataFrame containing the extracted file metadata.
+        """
+        filename_regex = re.compile(self.filename_regex, re.VERBOSE)
+        file_metadata = []
+        for filepath in filepaths:
+            match = re.match(filename_regex, filepath)
+            if match:
+                meta = match.groupdict()
+                meta.update({'filepath': filepath})
+                file_metadata.append(meta)
+
+        return pd.DataFrame(file_metadata)
+
     def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
         """Retrieve image/mask and metadata indexed by query.
 
@@ -530,28 +559,56 @@ class RasterDataset(GeoDataset):
                 f'query: {query} not found in index with bounds: {self.bounds}'
             )
 
+        self.file_df = self._get_regex_groups_as_df(filepaths)
+
+        res = []
         if self.separate_files:
-            data_list: list[Tensor] = []
-            filename_regex = re.compile(self.filename_regex, re.VERBOSE)
-            for band in self.bands:
-                band_filepaths = []
-                for filepath in filepaths:
-                    filename = os.path.basename(filepath)
-                    directory = os.path.dirname(filepath)
-                    match = re.match(filename_regex, filename)
-                    if match:
-                        if 'band' in match.groupdict():
-                            start = match.start('band')
-                            end = match.end('band')
-                            filename = filename[:start] + band + filename[end:]
-                    filepath = os.path.join(directory, filename)
-                    band_filepaths.append(filepath)
-                data_list.append(self._merge_files(band_filepaths, query))
-            data = torch.cat(data_list)
+            if self.return_as_ts:
+                grouped = self.file_df.groupby(['date', 'band']).agg(list)
+                all_dates = np.sort(self.file_df.date.unique())
+                for date in all_dates:
+                    res_per_date = [
+                        self._merge_files(filepaths, query)
+                        for band, filepaths in grouped.sort_values('band')
+                        .loc[date, 'filepath']
+                        .items()
+                    ]
+                    res.append(
+                        torch.cat(res_per_date).unsqueeze(0)
+                        if len(res_per_date) > 0
+                        else torch.tensor([])
+                    )
+                data = torch.cat(res) if len(res) > 0 else torch.tensor([])
+            else:
+                res_per_band = [
+                    self._merge_files(filepaths, query)
+                    for band, filepaths in self.file_df.groupby(['band'])
+                    .agg(list)
+                    .loc[:, 'filepath']
+                    .items()
+                ]
+                data = (
+                    torch.cat(res_per_band)
+                    if len(res_per_band) > 0
+                    else torch.tensor([])
+                )
         else:
-            data = self._merge_files(filepaths, query, self.band_indexes)
+            if self.return_as_ts:
+                all_dates = self.file_df.date.unique()
+                grouped = self.file_df.groupby(['date']).agg(list)
+                res = [
+                    self._merge_files(grouped.loc[date, 'filepath'], query).unsqueeze(0)
+                    for date in all_dates
+                ]
+                data = torch.cat(res) if len(res) > 0 else torch.tensor([])
+            else:
+                data = self._merge_files(filepaths, query, self.band_indexes)
 
         sample = {'crs': self.crs, 'bounds': query}
+        if self.return_as_ts:
+            sample['dates'] = [
+                datetime.strptime(date, self.date_format) for date in all_dates
+            ]
 
         data = data.to(self.dtype)
         if self.is_image:
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -97,6 +97,9 @@ class GeoDataset(Dataset[dict[str, Any]], abc.ABC):
     #: a different file format than what it was originally downloaded as.
     filename_glob = '*'
 
+    # Whether to return the dataset as a Timeseries, this will add another dimension to the dataset
+    return_as_ts = False
+
     # NOTE: according to the Python docs:
     #
     # * https://docs.python.org/3/library/exceptions.html#NotImplementedError
@@ -982,6 +985,7 @@ class IntersectionDataset(GeoDataset):
             if not isinstance(ds, GeoDataset):
                 raise ValueError('IntersectionDataset only supports GeoDatasets')
 
+        self.return_as_ts = dataset1.return_as_ts or dataset2.return_as_ts
         self.crs = dataset1.crs
         self.res = dataset1.res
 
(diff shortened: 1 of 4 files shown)


sfalkena:
Asking explicit feedback from @nilsleh @adamjstewart and @hfangcat

adamjstewart:
I need to find time to sit down and brainstorm this more. May include you and a few others in the discussion. I _think_ the changes proposed here are necessary for what I'm envisioning, but I want to finalize the rest of the picture before I make you do more work. I appreciate your eagerness to work on this, but don't want you to waste too much time on one particular direction before we make a decision.

nilsleh:
@sfalkena Thanks for the contribution and jumping into the effort for time-series support. I am not sure whether you already saw #877 , where we took a stab at this a while ago (and has unfortunately been stale for a while). But I'd be happy to jump in again and brainstorm a good approach for time-series support. In that PR we continued the discussion a bit about possible different use-cases, modalities, and modeling setups worthy of considering, so I'd be interested in what you think about those, just as possible further ideas to consider.

sfalkena:
Have added a new implementation of the SITS dataset, that is a relatively straightforward extension of the current RasterDataset implementation. Added the timeseries parameter to Sentinel dataset for now, will add it to all the other datasets when we agree on the implementation.

Also, feel free to run and plot the SITS dataset by running the tutorial notebook!

adamjstewart:
Can you update the initial comment description? Most of that is out-of-date by now.

torchgeo/datasets/geo.py
```diff
@@ -415,18 +416,24 @@ def __init__(
             transforms: a function/transform that takes an input sample
                 and returns a transformed version
             cache: if True, cache file handle to speed up repeated sampling
+            time_series: if True, return imagery as time series with shape [T,C,H,W]
+                and include 'dates' key with datetime information
```
adamjstewart: Let's not refer to it as "imagery", it may be masks or other raster data. Let's just call it "data". 

I'm not sure if "dates" is the right name for the key. It is a datetime, not a date. In other places in TorchGeo, we are using x, y, t for the centroids. We could change that to lon, lat, time.

Let's also add a description like:
"if False, return a merged mosaic with shape [C,H,W]"

torchgeo/datasets/geo.py
```diff
@@ -525,34 +532,89 @@ def __getitem__(self, query: GeoSlice) -> dict[str, Any]:
...
+        if self.time_series:
+            # Group by unique datetime intervals to support timeseries
+            grouped_index = index.groupby(index.index)
+            time_steps = []
+            dates = []
+
+            for datetime_interval, group in grouped_index:
+                # Extract the center date of the interval
+                center_date = datetime_interval.mid
+                dates.append(center_date)
```
adamjstewart: I would like to make as many values to be Tensors as we can. In other places, we convert the datetime/pd.Timestamp object to a `.timestamp()` float that can be stored as a Tensor.

torchgeo/datasets/geo.py
```diff
@@ -525,34 +532,89 @@ def __getitem__(self, query: GeoSlice) -> dict[str, Any]:
...
+                    )
+
+                time_steps.append(time_step_data)
+
+            # Stack along time dimension to create [T,C,H,W]
+            data = torch.stack(time_steps, dim=0)
+
         else:
-            data = self._merge_files(index.filepath, query, self.band_indexes)
+            # Original non-timeseries behavior
```
adamjstewart: Let's remove this comment, "original" won't make sense in a few years.

torchgeo/datasets/geo.py
```diff
@@ -525,34 +532,89 @@ def __getitem__(self, query: GeoSlice) -> dict[str, Any]:
...
         data = data.to(self.dtype)
         if self.is_image:
             sample['image'] = data
         else:
-            sample['mask'] = data.squeeze(0)
+            # Try to squeeze singleband mask
+            if self.time_series:
+                sample['mask'] = data.squeeze(1)
+            else:
+                sample['mask'] = data.squeeze(0)
```
adamjstewart: @ashnair1 do you remember why we added this? In general, I don't like using [un]squeeze, as it can have unintended consequences.

torchgeo/datasets/geo.py
```diff
@@ -525,34 +532,89 @@ def __getitem__(self, query: GeoSlice) -> dict[str, Any]:
...
-                    if match:
-                        if 'band' in match.groupdict():
-                            start = match.start('band')
-                            end = match.end('band')
-                            filename = filename[:start] + band + filename[end:]
-                    filepath = os.path.join(directory, filename)
-                    band_filepaths.append(filepath)
-                data_list.append(self._merge_files(band_filepaths, query))
-            data = torch.cat(data_list)
+        if self.time_series:
```
adamjstewart, line 535: I'm worried that `__getitem__` has become massive. We should use more helper functions here. There is quite a lot of duplicated code that concerns me.

adamjstewart (changes requested):
Only reviewed `geo.py` so far, will review the rest once we finalize `geo.py`

torchgeo/datasets/geo.py
sfalkena: Good point, I have refactored it.

torchgeo/datasets/geo.py
sfalkena: Renamed "dates" to "timestamps"

torchgeo/datasets/geo.py
sfalkena: Done

torchgeo/datasets/geo.py
sfalkena: Done

torchgeo/datasets/geo.py
sfalkena: My guess was to drop the single channel dimension, but not sure why that was initially in there, I just followed what was there.

docs/tutorials/customization.rst
```diff
@@ -7,6 +7,7 @@ The following tutorials will teach you how to customize TorchGeo to meet your ne
 
 * `Custom Non-Geospatial Datasets <contribute_non_geo_dataset.ipynb>`_: How to create and contribute a new NonGeoDataset
 * `Custom Raster Datasets <custom_raster_dataset.ipynb>`_: How to create a new RasterDataset
+* `SITS Raster Datasets <sits_dataset.ipynb>`_: How to create a RasterDataset with SITS support
```
adamjstewart, line 10: Took a quick look at the tutorial. I like it and think it's useful. However, I would put it in Basic Usage. Instead of writing a new dataset, simply show something like:

1. Create a dataset using `Sentinel2(paths)`
2. Create a spatial sampler, show the dimensions it returns
3. Create a new dataset using `Sentinel2(paths, time_series=True)`
4. Create a temporal sampler, show the dimensions it returns

The tests are currently failing because `root` is undefined

tests/datasets/test_geo.py
```diff
@@ -335,20 +336,44 @@ def naip(self, request: SubRequest) -> NAIP:
...
-        return Sentinel2(root, bands=bands, transforms=transforms, cache=cache)
+        time_series = request.param[2]
+        dataset = Sentinel2(
+            root,
+            bands=bands,
+            transforms=transforms,
+            cache=cache,
+            time_series=time_series,
+        )
+        dataset.is_image = request.param[3]
```
adamjstewart, line 375: I wouldn't bother testing `is_image` here, unless this is needed to get coverage?

torchgeo/datasets/geo.py
```diff
@@ -234,7 +234,7 @@ def crs(self) -> CRS:
     def crs(self, new_crs: CRS) -> None:
         """Change the :term:`coordinate reference system (CRS)` of a GeoDataset.
 
-        If ``new_crs == self.crs``, does nothing, otherwise updates the index.
+        If ``new_crs == self.crs``, does nothing, otherwise uptimestamps the index.
```
```suggestion
        If ``new_crs == self.crs``, does nothing, otherwise updates the index.
```

torchgeo/datasets/geo.py
```diff
@@ -525,12 +532,67 @@ def __getitem__(self, query: GeoSlice) -> dict[str, Any]:
...
+
+            # Stack along time dimension to create [T,C,H,W]
+            data = torch.stack(time_steps, dim=0)
+
+        else:
+            data = self._process_files_for_group(index.filepath, query)
+
+        sample: dict[str, Any] = {'crs': self.crs, 'bounds': query}
+        if self.time_series:
+            sample['timestamps'] = torch.tensor(timestamps, dtype=torch.float)
```
adamjstewart, line 558: `timestamps` should be singular just like `image` and `mask`. Not sure if `timestamp` is best, we use `t` in other places but that is less clear.

torchgeo/datasets/geo.py
```diff
@@ -525,12 +532,67 @@ def __getitem__(self, query: GeoSlice) -> dict[str, Any]:
...
+
+        data = data.to(self.dtype)
+        if self.is_image:
+            sample['image'] = data
+        else:
+            # Try to squeeze singleband mask
+            if self.time_series:
+                sample['mask'] = data.squeeze(1)
+            else:
+                sample['mask'] = data.squeeze(0)
```
adamjstewart: Can we use -3 instead? That should work for both

torchgeo/datasets/geo.py
```diff
@@ -525,12 +532,67 @@ def __getitem__(self, query: GeoSlice) -> dict[str, Any]:
...
+
+        if self.transforms is not None:
+            sample = self.transforms(sample)
+
+        return sample
+
+    def _process_files_for_group(
```
adamjstewart: Name is too generic, this could mean anything

torchgeo/datasets/sentinel.py
```diff
@@ -352,6 +356,8 @@ def __init__(
             transforms: a function/transform that takes an input sample
                 and returns a transformed version
             cache: if True, cache file handle to speed up repeated sampling
+            time_series: if True, return imagery as time series with shape [T,C,H,W]
+                and include 'dates' key with datetime information
```
adamjstewart, line 360: Key needs to be renamed here, but let's finalize a name first.

adamjstewart (changes requested):
`__getitem__` logic is still a bit complex. I wonder if we could do something like this. The goal is to get a single data structure that is a list of lists. The inner list represents separate bands that need to be stacked. The outer list represents separate tiles that need to be stacked along a new dimension (if time_series), else merged.

Maybe that's not any better than what you're doing. If I had time, I would try to develop my own solution to see if it becomes just as complex as what you're doing. 

Let's spend more time and use more creativity here and see if there's a simpler way to handle this with fewer if-branches. Right now, it hurts my brain just to look at and try to reason about.

adamjstewart:
Another way to think about this: if we didn't merge files with the same timestamp, I _think_ the code should only require a few lines of changes. So do that first, then see how little we would need to modify to merge files with the same timestamp (possibly in a separate PR).

torchgeo/datasets/geo.py
sfalkena: My opinion is that timestamps is a good description of what it is. However, we could name it timestamp_info, or ts_info if you like?

torchgeo/datasets/geo.py
sfalkena: Good one, fixed.

torchgeo/datasets/geo.py
sfalkena: Fixed

torchgeo/datasets/geo.py
sfalkena: Renamed to `_merge_file_collection`, lmk if you have better suggestions

sfalkena:
I have resolved PR comments. I have given the merging another refactor to make it more functional. Let me know what you think. I will update the tutorial once we agree  on all comments.

torchgeo/datasets/geo.py
```diff
@@ -525,40 +532,106 @@ def __getitem__(self, query: GeoSlice) -> dict[str, Any]:
...
+            ]
+
+        # Process all bands in parallel conceptually
+        band_tensors = [
+            self._merge_files(get_band_filepaths(band), query) for band in self.bands
+        ]
+
+        return torch.cat(band_tensors, dim=0)
+
+    def _construct_band_filepath(
```
adamjstewart, line 611: FYI, we added a new `_update_filepath` function in #3043 similar to this.

tests/datasets/test_geo.py
adamjstewart: Ah now I understand, this is indeed needed for coverage.

PR #2309, opened 2024-09-20
Sampling random generator

sfalkena:
Following #2307, adding the possibility to set the seed of the random samplers by supplying a `torch.Generator`. As far as I could tell, this is the only place where no random generator was used yet, apart from the Kornia transforms (would we want that? They do not support a generator argument).


calebrob6 (approved):

tests/samplers/test_batch.py
```diff
@@ -144,6 +145,20 @@ def test_weighted_sampling(self) -> None:
...
+        ds.index.insert(0, (0, 10, 0, 10, 0, 10))
+        sampler = RandomBatchGeoSampler(ds, 1, 1, generator=torch.manual_seed(0))
+        for bbox in sampler:
+            sample1 = bbox
+            break
+
+        sampler = RandomBatchGeoSampler(ds, 1, 1, generator=torch.manual_seed(0))
+        for bbox in sampler:
+            sample2 = bbox
+            break
```
adamjstewart, line 159:
```suggestion
        sampler1 = RandomBatchGeoSampler(ds, 1, 1, generator=torch.manual_seed(0))
        sampler2 = RandomBatchGeoSampler(ds, 1, 1, generator=torch.manual_seed(0))
        sample1 = next(iter(sampler1))
        sample2 = next(iter(sampler2))
```
Same for other tests

torchgeo/samplers/batch.py
```diff
@@ -97,9 +98,11 @@ def __init__(
             roi: region of interest to sample from (minx, maxx, miny, maxy, mint, maxt)
                 (defaults to the bounds of ``dataset.index``)
             units: defines if ``size`` is in pixel or CRS units
+            generator: random number generator
```
adamjstewart, line 101: I would like to make our usage of `generator` consistent across all of TorchGeo. This includes the name, description, type, and default value. This isn't exactly consistent in our existing code, but I don't want to make it even more consistent. So let's decide on defaults for these and then make sure everything matches. Here is my suggestion:

* name: `generator`
* description: `Pseudo-random number generator (PRNG).`
* type: `Generator | None`
* default: `None`

If you agree with this, you or I can open a separate PR for existing code.

adamjstewart: Note that every function we add a new parameter to needs:
```rst
.. versionadded:: 0.7
   The *generator* parameter.
```

adamjstewart (changes requested):
Looks great! Just some minor comments on consistency and simplicity.

We should probably do the same for all datamodules, but let's do that in a separate PR.

torchgeo/samplers/batch.py
sfalkena: > I would like to make our usage of `generator` consistent across all of TorchGeo. This includes the name, description, type, and default value. This isn't exactly consistent in our existing code, but I don't want to make it even more consistent. So let's decide on defaults for these and then make sure everything matches. Here is my suggestion:
> 
> * name: `generator`
> * description: `Pseudo-random number generator (PRNG).`
> * type: `Generator | None`
> * default: `None`
> 
> If you agree with this, you or I can open a separate PR for existing code.

My docstring was loosely based on Torch: "generator ([Generator](https://pytorch.org/docs/stable/generated/torch.Generator.html#torch.Generator)) – Generator used in sampling.", but I am totally fine with your proposed structure.

torchgeo/samplers/batch.py
```diff
@@ -86,6 +88,9 @@ def __init__(
         .. versionchanged:: 0.4
            ``length`` parameter is now optional, a reasonable default will be used
 
+        .. versionadded:: 0.7
+        The *generator* parameter.
```
adamjstewart, line 92:
```suggestion
           The *generator* parameter.
```
Same for others

torchgeo/samplers/single.py
```diff
@@ -281,9 +296,12 @@ def __init__(
             roi: region of interest to sample from (minx, maxx, miny, maxy, mint, maxt)
                 (defaults to the bounds of ``dataset.index``)
             shuffle: if True, reshuffle data at every epoch
+            generator: The random number generator used in combination with shuffle.
+
```
adamjstewart, line 300: Needs versionadded

sfalkena:
What is the thing with codecov failing? When I check the dashboard states 100%.

adamjstewart:
I'm trying to figure out the codecov thing now. I'm 99.9% sure it's a glitch, not your fault. Let's rerun the tests just to make sure.

adamjstewart (approved):

adamjstewart:
> default: `None`

This is actually a bit tricky. For many uses of `torch.rand*`, `None` is fine. However, in some cases like data module splitting, we actually have to use some kind of generator. If we don't, we end up with overlapping train/val/test splits. So to clarify, I think the default value to use in the function signature is `None`, but for some cases, we need to use:
```python
if generator is None:
    generator = torch.default_generator()
```
or something like that.

adamjstewart:
Do you want to open a PR to harmonize the previous examples where we used `generator`?

PR #2315, opened 2024-09-23
Adding tensorboard to CI

calebrob6:
In our tutorial notebooks we have an initial cell does a `%pip install torchgeo` along with any optional dependencies for that specific notebook (currently planetary-computer and pystac for downloading data). One notebook uses tensorboard so `pip install`s that as well. This PR adds tensorboard to the CI setup so we don't waste valuable time (time that we could spend running random data through random models to make sure dimensions align 😃) on downloading tensorboard at test time.


adamjstewart (approved):
We can add this to pyproject.toml in a future PR.

PR #2316, opened 2024-09-24
Bugfix: wrong test assertion

sfalkena:
In my previous pull requests, I changed an assertion at the last moment and saw this somehow got merged into main (I dont know why this test passed). I believe this change was wrong, hence the PR to revert it.

### DIFF
--- a/torchgeo/samplers/single.py
+++ b/torchgeo/samplers/single.py
@@ -320,7 +320,6 @@ class PreChippedGeoSampler(GeoSampler):
             generator = partial(torch.randperm, generator=self.generator)
 
         for idx in generator(len(self)):
-            print(idx)
             yield BoundingBox(*self.hits[idx].bounds)
 
     def __len__(self) -> int:
--- a/torchgeo/samplers/single.py
+++ b/torchgeo/samplers/single.py
@@ -320,6 +320,7 @@ class PreChippedGeoSampler(GeoSampler):
             generator = partial(torch.randperm, generator=self.generator)
 
         for idx in generator(len(self)):
+            print(idx)
             yield BoundingBox(*self.hits[idx].bounds)
 
     def __len__(self) -> int:
--- a/docs/tutorials/custom_raster_dataset.ipynb
+++ b/docs/tutorials/custom_raster_dataset.ipynb
@@ -478,10 +478,9 @@
    },
    "outputs": [],
    "source": [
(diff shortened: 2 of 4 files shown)


sfalkena:
Turns out the order matters.

This fails: 
```
        sampler1 = PreChippedGeoSampler(
            ds, shuffle=True, generator=torch.manual_seed(2)
        )
        sample1 = next(iter(sampler1))
        sampler2 = PreChippedGeoSampler(
            ds, shuffle=True, generator=torch.manual_seed(2)
        )
        sample2 = next(iter(sampler2))
        assert sample1 == sample2
```
But this passes:
```
        sampler1 = PreChippedGeoSampler(
            ds, shuffle=True, generator=torch.manual_seed(2)
        )
        sampler2 = PreChippedGeoSampler(
            ds, shuffle=True, generator=torch.manual_seed(2)
        )
        sample1 = next(iter(sampler1))
        sample2 = next(iter(sampler2))
        assert sample1 == sample2
```
Don't know why though.

adamjstewart:
And that's only true for pre-chipped, not for the other samplers?

adamjstewart:
I'm not sure if `torch.manual_seed` is designed to be used to create generators. Maybe we should explicitly create one and then set the seed of that generator.

sfalkena:
Indeed I think this is where the problem originates from. We need to use `torch.Generator().manual_seed()`. The [documentation ](https://pytorch.org/docs/stable/generated/torch.manual_seed.html) of `torch.manual_seed` mentions that it returns a generator object, but diving deeper it is only used on the process level. I will change all instances accordingly.

sfalkena:
Found that a tutorial notebook was also using torch.manual_seed, so changed that as an example as well.

adamjstewart:
I think the tutorial notebook example is fine, we aren't using the generator returned by `torch.manual_seed`. But it's also fine to demonstrate the new generator parameter, so let's keep your change.

torchgeo/samplers/single.py
```diff
@@ -320,6 +320,7 @@ def __iter__(self) -> Iterator[BoundingBox]:
             generator = partial(torch.randperm, generator=self.generator)
 
         for idx in generator(len(self)):
+            print(idx)
```
adamjstewart, line 323: You can remove this now

adamjstewart (approved):
Thanks for the quick fix!

PR #2317, opened 2024-09-24
BaseTask: fix load_from_checkpoint, ignore 'ignore'

adamjstewart:
Alternative to #2314. @calebrob6 can you see if this also fixes your issue?

I think this makes more sense, as `ignore` is more of a class attribute than an instance parameter. It also makes it possible to add subclasses with additional ignores. Downside is that it's technically backwards incompatible and will have to wait until 0.7.0.

### DIFF
--- a/torchgeo/trainers/base.py
+++ b/torchgeo/trainers/base.py
@@ -19,6 +19,9 @@ class BaseTask(LightningModule, ABC):
     .. versionadded:: 0.5
     """
 
+    #: Parameters to ignore when saving hyperparameters.
+    ignore: Sequence[str] | str | None = 'weights'
+
     #: Model to train.
     model: Any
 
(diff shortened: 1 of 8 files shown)


adamjstewart:
I haven't yet found a simple way to test this, may skip the tests.

calebrob6:
🫢

calebrob6:
Yep it works! Here is the code I used to test if that's helpful:

```
import torch
from lightning.pytorch import Trainer
from lightning.pytorch.callbacks import ModelCheckpoint
from torch.optim import Adam
from torch.utils.data import DataLoader, Dataset

        super().__init__(*args, **kwargs)

    def configure_optimizers(self):
        return {
            'optimizer': Adam(self.parameters(), lr=self.hparams['lr']),
        }

class TestDataset(Dataset):
    def __init__(self, n_samples=10):
        self.n_samples = n_samples
        self.data = torch.rand((n_samples, 3, 256, 256))
        self.targets = torch.randint(0, 2, (n_samples, 256, 256))
    def __len__(self):
        return self.n_samples
    def __getitem__(self, idx):
        return {
            "image": self.data[idx],
            "mask": self.targets[idx],
        }



task = CustomSemanticSegmentationTask(model="unet", tmax=100)

dataset = TestDataset(10)
train_loader = DataLoader(dataset, batch_size=5)

checkpoint_callback = ModelCheckpoint(
    monitor='train_loss',
    dirpath='./tmp/',
    filename='my_model-{epoch:02d}',
    save_top_k=1,
    mode='min'
)

trainer = Trainer(
    max_epochs=1,
    callbacks=[checkpoint_callback],
    num_sanity_val_steps=0,
)

trainer.fit(task, train_loader)

task = CustomSemanticSegmentationTask.load_from_checkpoint("tmp/my_model-epoch=00.ckpt")
```

calebrob6 (approved):

PR #2323, opened 2024-09-27
Removing eval in model weight API

calebrob6:
Removing use of eval.

~@nilsleh -- do you remember what the use case for eval was? If not, I say we replace all instances of `get_weight` with `get_model_weights` and remove `get_weight`.~ Nevermind, I understand it now.

`get_weight(...)` allows a user to pass a string "ResNet18_Weights.LANDSAT_ETM_SR_MOCO" and get back the object `ResNet18_Weights.LANDSAT_ETM_SR_MOCO` WeightEnum. Previously we did this with `eval("ResNet18_Weights.LANDSAT_ETM_SR_MOCO")` which is _bad_. This proposed fix simply iterates through all available WeightEnums to look for matches and has the benefit of raising an error that makes sense if there is no match (vs. the `eval(...)` which would do whatever).

### DIFF
--- a/torchgeo/models/api.py
+++ b/torchgeo/models/api.py
@@ -111,12 +111,11 @@ def get_weight(name: str) -> WeightsEnum:
         The requested weight enum.
 
     Raises:
-        ValueError: if `name` doesn't point to a valid WeightsEnum 
+        ValueError: if `name` doesn't point to a valid WeightsEnum
     """
     if name in _model_weights:
         return _model_weights[name]
     else:
-        sub_weights = {}
         for name, weight_enum in _model_weights.items():
             if isinstance(name, str):
                 for sub_weight_enum in weight_enum:
--- a/torchgeo/models/api.py
+++ b/torchgeo/models/api.py
@@ -67,6 +67,11 @@ _model_weights = {
     'vit_small_patch16_224': ViTSmall16_Weights,
 }
 
+for name, weight_enum in _model_weights.items():
+    if isinstance(name, str):
+        for sub_weight_enum in weight_enum:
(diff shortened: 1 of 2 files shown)


adamjstewart:
Most of this API was copied from torchvision, let me check what their code looks like these days.

adamjstewart:
I wonder if we can use torchvision's `@register_model` decorator and remove this entire file. Timm also has a way to register models, although I don't know how compatible it is. But it would let us easily use our custom models in our trainers.

calebrob6:
I looked at the torchvision version (https://github.com/pytorch/vision/blob/main/torchvision/models/_api.py#L108) and it seems more hacky than what I've implemented here. I updated the PR description with more details.

torchgeo/models/api.py
```diff
@@ -109,8 +109,17 @@ def get_weight(name: str) -> WeightsEnum:
 
     Returns:
         The requested weight enum.
+
+    Raises:
+        ValueError: if `name` doesn't point to a valid WeightsEnum
```
adamjstewart, line 114:
```suggestion
        ValueError: If *name* is not a valid WeightsEnum.
```

adamjstewart (dismissed):
More lines of code, but also more secure. This is also better from a type checking perspective, as `eval` can return any type. We can think about how to delete this file and properly register our models in a separate PR.

adamjstewart (approved):

adamjstewart:
Published CVE for reference: https://www.cve.org/CVERecord?id=CVE-2024-49048

adamjstewart:
And published GitHub advisory: https://github.com/torchgeo/torchgeo/security/advisories/GHSA-ghq9-vc6f-8qjf

Issue #2289, opened 2024-09-09
Support model in_chans not equal to pre-trained weights in_chans


adamjstewart:
### Summary

If a user specifies `in_chans` and `weights`, and `weights.meta['in_chans']` differs from `in_chans`, the user-specified argument should take precedence and weights should be repeated, similar to how timm handles pre-trained weights.

### Rationale

When working on change detection, it is common to take two images and stack them along the channel dimension. However, this makes it impossible to use our pre-trained weights. Ideally, I would like to support something like:
```python
from torchgeo.models import ResNet50_Weights, resnet50

model = resnet50(in_chans=4, weights=ResNet50_Weights.SENTINEL1_ALL_MOCO)
```
Here, the weights have 2 channels (HH and HV), while the dataset and model will have 4 channels (HH, HV, HH, HV).

### Implementation

https://timm.fast.ai/models#Case-2:-When-the-number-of-input-channels-is-not-1 describes the implementation that timm uses. This can be imported as:
```python
from timm.models.helpers import load_pretrained
```
We should make use of this in all of our model definitions instead of `model.load_state_dict`.

### Alternatives

There is some ongoing work to add a ChangeDetectionTask that may split each image into a separate sample key. However, there will always be models that require images stacked along the channel dimension, so I don't think we can avoid supporting this use case.

keves1:
I'm interested in contributing to this, could I have this assigned to me? I looked into the `load_pretrained` method and found that it will only copy the weights if `in_chans` of the weights is 3, otherwise it uses random init for the first conv layer (see link to `adapt_input_conv()` which is called by `load_pretrained()`)

https://github.com/huggingface/pytorch-image-models/blob/ee5b1e8217134e9f016a0086b793c34abb721216/timm/models/_manipulate.py#L256-L278

So this would need to be adapted since there are a variable number of input channels for the weights we would use.

adamjstewart:
Note: this is definitely needed for our new ChangeDetectionTask.

adamjstewart:
@isaaccorley is proposing that we simply register all of our models with timm and then we no longer need any custom magic to repeat weight channels.

PR #2324, opened 2024-09-27
Support model in_chans not equal to pre-trained weights in_chans

keves1:
This PR addresses #2289.

I've made a good start on this and am creating this draft PR to get some feedback before continuing. I have created a load_pretrained function that will copy weights as done by timm, but which supports any number of in_chans of the weights, not just 3. So far I just implemented this new functionality in resnet50 and will expand that to other models if this is a good approach.

With this change, you can create a model like this, in this case the two channels will be copied into 4:
```python
model = resnet50(weights=ResNet50_Weights.SENTINEL1_ALL_MOCO, in_chans=4)
```

I created a new file torchgeo/models/utils.py where I put the load_pretrained function since I didn't see any suitable existing file. Is there a better place for it?

### DIFF
--- a/torchgeo/models/resnet.py
+++ b/torchgeo/models/resnet.py
@@ -715,7 +715,6 @@ def resnet50(
 
     if weights:
         pretrained_cfg = {}
-        pretrained_cfg['url'] = weights.url
         pretrained_cfg['first_conv'] = 'conv1'
         in_chans = kwargs.get('in_chans', 3)
         missing_keys, unexpected_keys = load_pretrained(
--- a/torchgeo/models/resnet.py
+++ b/torchgeo/models/resnet.py
@@ -17,6 +17,7 @@ from .swin import (
     _satlas_sentinel2_transforms,
     _satlas_transforms,
 )
+from .utils import load_pretrained
 
 # https://github.com/zhu-xlab/DeCUR/blob/f190e9a3895ef645c005c8c2fce287ffa5a937e3/src/transfer_classification_BE/linear_BE_resnet.py#L286
 # Normalization by channel-wise band statistics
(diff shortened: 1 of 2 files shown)


keves1:

adamjstewart:
> I have created a load_pretrained function that will copy weights as done by timm, but which supports any number of in_chans of the weights, not just 3.

Why can't we use `timm.models.helpers.load_pretrained` instead of writing our own custom code?

keves1:
> Why can't we use `timm.models.helpers.load_pretrained` instead of writing our own custom code?

Because that function will only copy the weights to additional input channels if the first convolution layer of the weights has 3 input channels. Otherwise it raises a NotImplementedError exception and randomly initializes the weights. I linked to the timm implementation in my comment on issue #2289 .

adamjstewart:
Will try to review next week, this week is quite busy. Apologies for the wait!

adamjstewart:
I recently discovered `timm.models.adapt_input_conv`, which makes it easy to change the number of channels without losing pretrained weights. See #2602 for an example. There seems to be a lot of other builtin stuff in timm we might be able to make use of. If it doesn't support everything we need, I would be willing to try to add support for it in timm.

keves1:
I actually looked into `timm.models.adapt_input_conv` (this is what is called by `timm.models.helpers.load_pretrained`) and it only works if the if the first convolution layer of the weights has 3 input channels (see my comment above). So we couldn't use `adapt_input_conv` with weights trained on multispectral or other non 3 channel imagery.

Here's an example:
```python
from timm.models import adapt_input_conv
from torchgeo.models import ResNet18_Weights

weights = ResNet18_Weights.SENTINEL2_ALL_MOCO.get_state_dict()
adapt_input_conv(26, weights['conv1.weight'])
```

```
NotImplementedError                       Traceback (most recent call last)
Cell In[17], line 1
----> 1 adapt_input_conv(26, weights['conv1.weight'])

File ~/miniconda3/envs/torchgeo-env/lib/python3.11/site-packages/timm/models/_manipulate.py:270, in adapt_input_conv(in_chans, conv_weight)
    268 elif in_chans != 3:
    269     if I != 3:
--> 270         raise NotImplementedError('Weight format not supported by conversion.')
    271     else:
    272         # NOTE this strategy should be better than random init, but there could be other combinations of
    273         # the original RGB input layer weights that'd work better for specific cases.
    274         repeat = int(math.ceil(in_chans / 3))

NotImplementedError: Weight format not supported by conversion.
```

There may be other methods in timm which we could use like you mentioned, or may need to add support.
I'm focusing on trying to get a PR ready for adding the Autoregression trainer, so I won't be working on this more in the short term.

PR #2326, opened 2024-09-28
Tests: set matplotlib backend locally too

adamjstewart:
The default matplotlib backend is interactive. In CI, this results in the tests hanging. Locally, this can result in the terminal losing focus. In CI, we worked around this by setting the `MPLBACKEND` environment variable. However, this doesn't solve the problem locally. This PR uses `matplotlib.use` instead, and works both locally and in CI.

Reference: https://matplotlib.org/stable/users/explain/figure/backends.html

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -15,8 +15,6 @@ jobs:
   latest:
     name: latest
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


calebrob6:
Nice

calebrob6 (approved):

PR #2333, opened 2024-10-04
DataModules: add configurable args to dataloader

### LINKED ISSUE #2332: Expose more arguments to dataloader
### Summary

I have noticed differences in data loading speeds by varying `persistent_workers`, `pin_memory`, and `prefetch_factor`. In the dataloader. I saw that `persistent_workers` has already been enabled when `num_workers>1` (https://github.com/microsoft/torchgeo/pull/2291). Would it be possible to let more arguments than `num_workers` be configurable in datamodules?

### Rationale

It enables faster data loading in some cases.

### Implementation

Let `pin_memory` and `prefetch_factor` be configurable like `num_workers`. See https://github.com/microsoft/torchgeo/pull/2333.

### Alternatives

Maybe not so nice with many additional arguments to datamodule. Is there any way to get around that?

### Additional information

_No response_

anton-emanuel:
Closes https://github.com/microsoft/torchgeo/issues/2332.

### DIFF
--- a/tests/datamodules/test_utils.py
+++ b/tests/datamodules/test_utils.py
@@ -6,7 +6,7 @@ import re
 import numpy as np
 import pytest
 
-from torchgeo.datamodules.utils import group_shuffle_split
+from torchgeo.datamodules.utils import group_shuffle_split, split_prefixed_kwargs
 
 
 def test_group_shuffle_split() -> None:
@@ -44,3 +44,20 @@ def test_group_shuffle_split() -> None:
 
         assert len(set(train_indices1) & set(test_indices1)) == 0
         assert len(set(groups[train_indices1])) == 2
+
+
+def test_split_prefixed_kwargs() -> None:
+    kwargs = {
+        'testprefix1_param1': 10,
+        'testprefix1_param2': 20,
+        'testprefix2_param3': 30,
+        'other_param': 40,
+    }
+
+    testprefix1_kwargs, testprefix2_kwargs, other_kwargs = split_prefixed_kwargs(
+        'testprefix1_', 'testprefix2_', **kwargs
+    )
+
+    assert testprefix1_kwargs == {'param1': 10, 'param2': 20}
+    assert testprefix2_kwargs == {'param3': 30}
+    assert other_kwargs == {'other_param': 40}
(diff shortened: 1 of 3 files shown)


adamjstewart:
How much of a speed difference do you notice?

At the moment, the datamodules are all designed to pass additional kwargs to the dataset class. So we should technically add these new args to every datamodule, not just to the base class. We _could_ design things such that all kwargs are either passed to the data loader or the dataset class, but I'm worried users might get confused by this. Of course, they don't need to understand the feature if they don't want to use it.

I guess my concern is, why stop at only these variables?

anton-emanuel:
I changed it to use kwargs now

anton-emanuel:
@microsoft-github-policy-service agree company="Kongsberg Satellite Services"

DimitrisMantas:
I think it woul be better for kwargs to be a JSON-like dict, something like {"dataset": ..., "dataloader": ...}. We could even turn this into a typed dict or a dataclass so that it can be clearly documented and tested

adamjstewart:
> I feel like I have seen this solution somewhere in TorchGeo, but might have been another library

Yes, we do something similar in our combined GeoDatasets (NAIP + Chesapeake, Sentinel-2 + CDL/EuroCrops/NCCM/SAS).

> I think it woul be better for kwargs to be a JSON-like dict

The downside of this is that it would be backwards-incompatible and affect all datamodules.

I'm still thinking about this idea. I wish there was a way to predict a good default value. I'm curious if any other maintainers have experimented with these variables and how important they are to tune. In the meantime, anyone can do this themselves by overriding the respective methods like so:
```python
from torchgeo.datamodules import TropicalCycloneDataModule

class MyTropicalCycloneDataModule(TropicalCycloneDataModule):
    def _dataloader_factory(self, split):
        dataset = self._valid_attribute(f'{split}_dataset', 'dataset')                   
        batch_size = self._valid_attribute(f'{split}_batch_size', 'batch_size')          
        return DataLoader(                                                               
            dataset=dataset,                                                             
            batch_size=batch_size,                                                       
            shuffle=split == 'train',                                                    
            num_workers=self.num_workers,                                                
            collate_fn=self.collate_fn,                                                  
            persistent_workers=self.num_workers > 0,  
            ...                                   
        )
```

calebrob6:
I think this is still important to expose

adamjstewart:
@evgeniialappo can you add this to your I/O benchmarking experiments?

PR #2340, opened 2024-10-08
Docs: scikit-eo is a new alternative library

### LINKED ISSUE #2335: Add scikit-eo to list of alternatives
### Issue

[scikit-eo](https://github.com/yotarazona/scikit-eo) now qualifies to appear on our list of [TorchGeo alternatives](https://torchgeo.readthedocs.io/en/stable/user/alternatives.html).

### Fix

We should compute the info we need to add scikit-eo.

@yotarazona may be able to help submit a PR or at least provide stats on # datasets/weights/etc.

Nowosad:
@adamjstewart -- I was asked by @yotarazona to help on the issue #2335 

Closes #2335

### DIFF
--- a/docs/user/metrics/features.csv
+++ b/docs/user/metrics/features.csv
@@ -7,3 +7,4 @@ Library,ML Backend,I/O Backend,Spatial Backend,Transform Backend,Datasets,Weight
 `DeepForest`_,PyTorch,"GDAL, OpenCV, pandas, pillow, scipy",R-tree,Albumentations,0,3,❌,❌,❌,❌
 `SITS`_,R Torch,GDAL,-,tidyverse,22,0,❌,✅,✅,✅
 `TerraTorch`_,PyTorch,"GDAL, h5py, pandas, xarray",R-tree,Albumentations,16,1,✅,✅,❌,🚧
+`scikit-eo`_,"scikit-learn, TensorFlow*","pandas, scipy, numpy","geopandas, rasterio",numpy,22,0,❌,❌,❌,🚧
--- a/docs/user/metrics/features.csv
+++ b/docs/user/metrics/features.csv
@@ -7,4 +7,4 @@ Library,ML Backend,I/O Backend,Spatial Backend,Transform Backend,Datasets,Weight
 `DeepForest`_,PyTorch,"GDAL, OpenCV, pandas, pillow, scipy",R-tree,Albumentations,0,3,❌,❌,❌,❌
 `SITS`_,R Torch,GDAL,-,tidyverse,22,0,❌,✅,✅,✅
 `TerraTorch`_,PyTorch,"GDAL, h5py, pandas, xarray",R-tree,Albumentations,16,1,✅,✅,❌,🚧
-`scikit-eo`_,"scikit-learn, TensorFlow*","pandas, scipy, numpy","geopandas, rasterio",numpy,22,0,❌,❌,❌,🚧
+`scikit-eo`_,"scikit-learn, TensorFlow","pandas, scipy, numpy, rasterio","geopandas",numpy,0,0,❌,❌,❌,🚧
--- a/docs/user/metrics/github.csv
+++ b/docs/user/metrics/github.csv
@@ -7,3 +7,4 @@ Library,Contributors,Forks,Watchers,Stars,Issues,PRs,Releases,Commits,Core SLOCs
 `DeepForest`_,17,172,17,474,413,301,44,864,"3,357","1,794",86%,MIT
 `SITS`_,14,76,28,451,622,583,44,"6,244","24,284","8,697",94%,GPL-2.0
 `TerraTorch`_,9,10,9,121,46,92,2,243,"10,101",583,44%,Apache-2.0
+`scikit-eo`_,6,17,8,132,20,11,15,496,,,Apache-2.0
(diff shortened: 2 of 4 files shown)


adamjstewart:
Thanks, can you also update github.csv and downloads.csv? https://torchgeo.readthedocs.io/en/stable/user/alternatives.html covers some of the details of how we calculate SLOCs and downloads. I'm also happy to help, you've already done the hard parts for me.

Nowosad:
Hi @adamjstewart, I just added more info on the package, except three cells in the github.csv file, which I do not know the answers for. Maybe you or @yotarazona can help there?

Nowosad:

docs/user/metrics/features.csv
```diff
@@ -7,3 +7,4 @@ Library,ML Backend,I/O Backend,Spatial Backend,Transform Backend,Datasets,Weight
 `DeepForest`_,PyTorch,"GDAL, OpenCV, pandas, pillow, scipy",R-tree,Albumentations,0,3,❌,❌,❌,❌
 `SITS`_,R Torch,GDAL,-,tidyverse,22,0,❌,✅,✅,✅
 `TerraTorch`_,PyTorch,"GDAL, h5py, pandas, xarray",R-tree,Albumentations,16,1,✅,✅,❌,🚧
+`scikit-eo`_,"scikit-learn, TensorFlow*","pandas, scipy, numpy","geopandas, rasterio",numpy,22,0,❌,❌,❌,🚧
```
adamjstewart, line 10: Tell me more about the TensorFlow integration. Do all versions support TF? How tightly is TF integrated into the library design (like is it a core component or just used in some demo notebooks)?

Usually rasterio is used for I/O. By "spatial backend" we just mean the library used to track the bounding boxes of all data for sampling purposes, although I'm fine with a wider interpretation.

Also, where can I find the list of supported datasets?

Nowosad: @yotarazona should be able to answer here

docs/user/metrics/github.csv
```diff
@@ -7,3 +7,4 @@ Library,Contributors,Forks,Watchers,Stars,Issues,PRs,Releases,Commits,Core SLOCs
 `DeepForest`_,17,172,17,474,413,301,44,864,"3,357","1,794",86%,MIT
 `SITS`_,14,76,28,451,622,583,44,"6,244","24,284","8,697",94%,GPL-2.0
 `TerraTorch`_,9,10,9,121,46,92,2,243,"10,101",583,44%,Apache-2.0
+`scikit-eo`_,6,17,8,132,20,11,15,496,,,Apache-2.0
```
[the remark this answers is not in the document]
Nowosad: @yotarazona should be able to answer here as well

docs/user/metrics/features.csv
yotarazona: Hi @adamjstewart and @Nowosad 

Scikit-eo uses TensorFlow only in the ```deeplearning.py``` implementation (https://github.com/yotarazona/scikit-eo/blob/main/scikeo/deeplearning.py) and only on a notebook: https://github.com/yotarazona/scikit-eo/blob/main/examples/notebooks/11_Deep_Learning_Classification_FullyConnected.ipynb. 
The ```deeplearning.py``` implementation was developed since the first version, but since version v0.2.29, the documentation was changed according to the latest updates of TensorFlow (i.e., the way to load the tensorflow library and keras). 

Regarding the I/O, got it, so rasterio is used as a I/O.
Regarding the "Spatial Backend", we put geopandas and raster because there is an implementation that allows to extract multivariate values from the spatial intersection of a satellite image and a vector spatial data (point feature for now). In any case, it can be left as “no support” in case it does not comply with the “Backend spatial” stipulations.

Right now we do not have a list of datasets but in the example notebooks you will find download links to the datasets used in each demo example. In most examples we use two datasets, a satellite image and a vector data. Need a list of datasets? We can insert a list in the README.md presentation of the examples.

adamjstewart: * ML Backend: Got it. In that case, it seems like TF is still supported, so you could probably remove the asterisk if you want
* I/O Backend: In this case I think it makes sense to list rasterio as an I/O backend and geopandas as a spatial backend
* Datasets: If I understand correctly, scikit-eo does not provide any builtin data loaders for existing datasets, it simply provides the model training code (and demonstrates how to create your own data loaders in the documentation). In that case, I would put 0 for # of datasets.

adamjstewart:
You'll also need to add:
```rst
.. _scikit-eo: https://github.com/yotarazona/scikit-eo
```
to the end of `docs/user/alternatives.rst` in order for the link to render correctly.

docs/user/metrics/github.csv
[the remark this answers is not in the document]
yotarazona: @adamjstewart and @Nowosad, the "shp" argument has been changed to "vector" since some versions ago. Now you should not get errors. About the DeepLearning test error, it's weird but I don't get any error :).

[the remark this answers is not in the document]
adamjstewart: I would highly recommend running pytest in CI, it's the only way to prevent bugs like this from appearing. Let me know if you need any help setting up CI, and feel free to use TorchGeo's CI as an example.

adamjstewart:
@Nowosad I'm working on the 0.6.1 release today if you want to try to finish this. You should now have all the data you need.

Nowosad:
Hi @adamjstewart -- thanks for all the help. I just updated the files based on your suggestions.

adamjstewart (approved):

adamjstewart:
Thanks for the PR @Nowosad! And best of luck to the future of scikit-eo!!

PR #2346, opened 2024-10-11
DOFA: save parameters as instance attributes

nilsleh:
Adds following class attributes to DOFA:
-  `embed_dim` 
- `depth`

### DIFF
--- a/torchgeo/models/dofa.py
+++ b/torchgeo/models/dofa.py
@@ -83,7 +83,11 @@ class TransformerWeightGenerator(nn.Module):
         self.wt_num = 128
         self.weight_tokens = nn.Parameter(torch.empty([self.wt_num, input_dim]))
         self.bias_token = nn.Parameter(torch.empty([1, input_dim]))
+        self.input_dim = input_dim
+        self.output_dim = output_dim
         self.embed_dim = embed_dim
+        self.num_heads = num_heads
+        self.num_layers = num_layers
 
         # timm's trunc_normal_(std=.02) is effectively normal_(std=0.02) as cutoff is
         # too big (2.)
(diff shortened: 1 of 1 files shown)


adamjstewart:
Should we just save all parameters in case someone wants to reference them later?

FYI @xiong-zhitong

nilsleh:
Yeah, I was wondering why not all parameters were stored, but will include all params now.

adamjstewart:
Can we move the parameter saving to the top of `__init__`? This is how we do it for all other classes (e.g., datasets). This lets us immediately check for valid args, assign backup values in case of None, and use the attrs later in `__init__` (not that I think these apply to DOFA, just want to be consistent).

torchgeo/models/dofa.py
```diff
@@ -267,6 +274,15 @@ def __init__(
...
+        self.patch_size = patch_size
+        self.drop_rate = drop_rate
+        self.embed_dim = embed_dim
+        self.depth = depth
+        self.num_heads = num_heads
+        self.num_classes = num_classes
+        self.mlp_ratio = mlp_ratio
+
         self.dynamic_embed_dim = dynamic_embed_dim
         self.global_pool = global_pool
```
adamjstewart, line 287: Let's move these up and keep them in the same order as `__init__`

torchgeo/models/dofa.py
```diff
@@ -267,8 +274,17 @@ def __init__(
...
+        self.depth = depth
+        self.num_heads = num_heads
         self.dynamic_embed_dim = dynamic_embed_dim
+        self.num_classes = num_classes
         self.global_pool = global_pool
+        self.mlp_ratio = mlp_ratio
+
         if self.global_pool:
             norm_layer = norm_layer
             embed_dim = embed_dim
```
adamjstewart, line 290: Do these lines do anything? We can probably delete them (could be in a separate PR)

adamjstewart (approved):

torchgeo/models/dofa.py
nilsleh: It does not look like it, but did not want to touch other parts beyond saving the attributes.

PR #2350, opened 2024-10-16
Add CaFFe Dataset

nilsleh:
This PR adds the [CaFFe (CAlving Fronts and where to Find thEm) dataset](https://doi.pangaea.de/10.1594/PANGAEA.940950) and accompanying `DataModule` for calving front and land scape zone segmentation.

Implementation for the chipped dataset based on [this script](https://github.com/Nora-Go/Calving_Fronts_and_Where_to_Find_Them/blob/de8e8c36292470fd4037501155db2f5d4aa8ef13/data_processing/data_preprocessing.py), which I uploaded to [Huggingface](https://huggingface.co/datasets/torchgeo/glacier_calving_front/tree/main).

Dataset features:
- 13,090 train, 2,241 validation, and 3,761 test images
- varying spatial resolution of 6-20m
- paired binary calving front segmentation masks
- paired multi-class land cover segmentation masks

 Dataset format:
- images are single-channel pngs with dimension 512x512
- segmentation masks are single-channel pngs

TODOs:
- [x] Check class enumeration and plotting for correct class/colors 

Updated Random Sample Plot:
glacier_test


torchgeo/datamodules/glacier_calving_front.py
```diff
@@ -0,0 +1,67 @@
...
+
+        self.aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
+            K.Resize(size),
+            data_keys=['image', 'mask'],
+        )
+
+        self.size = size
+
+    def setup(self, stage: str) -> None:
```
adamjstewart, line 57: Looks the same as the base class, could prob be removed

torchgeo/datasets/glacier_calving_front.py
```diff
@@ -0,0 +1,249 @@
...
+
+    zipfilename = 'glacier_calving_data.zip'
+
+    data_dir = 'glacier_calving_data'
+
+    image_dir = 'sar_images'
+
+    mask_dirs = ('fronts', 'zones')
+
+    url = 'https://huggingface.co/datasets/torchgeo/glacier_calving_front/resolve/main/glacier_calving_data.zip'
```
adamjstewart, line 65: Replace `main` with the git commit hash for stability/reproducibility.

nilsleh:
Hi @Nora-Go,

we would like to add your dataset to torchgeo but I have a couple of questions:

- the imagery and labels come as `.png` files and I did not find any accompanying metadata files that would give information about the geo-location etc of the imagery and patches, which would be extremely useful
- I am not precisely sure about the encoding of the land zone classes. In your repo, the classes are converted to labels [here](https://github.com/Nora-Go/Calving_Fronts_and_Where_to_Find_Them/blob/de8e8c36292470fd4037501155db2f5d4aa8ef13/validate_or_test.py#L67) and [this comment](https://github.com/Nora-Go/Calving_Fronts_and_Where_to_Find_Them/blob/de8e8c36292470fd4037501155db2f5d4aa8ef13/models/zones_segmentation_model.py#L10) seems to suggest that for the class label pixel values of [0, 64, 127, 255], the labels are [Glacier, Rock, Ocean/Ice Melange, NA] in that same order, where NA is essentially a background class. Is that correct?

Thanks in advance!

torchgeo/datasets/glacier_calving_front.py
```diff
@@ -0,0 +1,249 @@
...
+
+    md5 = '56e39e33f88a9842f48c513083e3c50a'
+
+    px_class_values: ClassVar[dict[int, str]] = {
+        0: 'background',
+        64: 'ocean',
+        127: 'rock',
+        254: 'glacier',
```
adamjstewart, line 73: Either in the dataset or the datamodule, we need to map these to ordinal numbers, correct? I've been meaning to add a transform for this since it comes up so often.

nilsleh: I marked as a draft, because I am not 100% sure here, that's why I asked the question to the author. But yeas you are correct, missing code for mapping it to ordinal and will do that once I get the answer :)

Nora-Go:
First of all - cool! :D happy that the access to the dataset gets easier and that you take the time! Thank you :)

>     * the imagery and labels come as `.png` files and I did not find any accompanying metadata files that would give information about the geo-location etc of the imagery and patches, which would be extremely useful. Thus, I am wondering where I might be able to find this information.
> 
This information is not publicly available so far. If you want I can send you the geo tiffs - those have the information and you can extract them to include them here.

>     * I am not precisely sure about the encoding of the land zone classes. In your repo, the classes are converted to labels [here](https://github.com/Nora-Go/Calving_Fronts_and_Where_to_Find_Them/blob/de8e8c36292470fd4037501155db2f5d4aa8ef13/validate_or_test.py#L67) and [this comment](https://github.com/Nora-Go/Calving_Fronts_and_Where_to_Find_Them/blob/de8e8c36292470fd4037501155db2f5d4aa8ef13/models/zones_segmentation_model.py#L10) seems to suggest that for the class label pixel values of [0, 64, 127, 255], the labels are [Glacier, Rock, Ocean/Ice Melange, NA] in that same order, where NA is essentially a background class. Is that correct?
> 
It is 0: NA (no information available - e.g. you don't see anything in the image, but in reality it would be one of the following), 64: Rock outcrop, 127: Glacier, and 254: Ocean/Ice Melange

I have some questions myself:
Do you have to provide the patches instead of the full images? And a predefined val split (using the one that I used for the baseline)? At the moment I'm actually using patches of size 512 x 512 and that works better. And the val split I used so far would not work if you wanted to take advantage of the time series information :)

nilsleh:
@Nora-Go thank you for the reply. Ah too bad, that metadata is not available right away. I won't have the time to make a deep dive into the dataset, I was just looking for a interesting task with good labels for an evaluation.

However, if you upload the geo tiffs somewhere (could also be the Hugginface repo) someone can add it at a later time, this time as a `GeoDataset` where patch sampling then happens on the fly with a `GeoSampler`. Since, I was interested in just a "benchmark" evaluation, a predefined split is better for consistency and comparability.  I can rerun your patching script to include the dataset as a 512x512 version if you prefer that?

If you are interested in adding any of these parts yourself, and should you have questions about that, feel free to reach out.

Nora-Go:
@nilsleh ah I see. Ok. Do you need the meta data already for using it as a benchmark? If yes, I can try to extract what you need. Otherwise I'll see how I'll deal with the geo tiffs for a future GeoDataset.

For using the dataset as a benchmark, do you want to compare against my baseline (then the 256 x 256 is good) or do you want to compare against the state-of-the-art? The state-of-the-art (https://ieeexplore.ieee.org/abstract/document/10440599) uses a mixture between 256x256 and 512x512.

Let me know if you have any further questions or need help handling the dataset :)

nilsleh:
Having latitude, longitude, and time available as metadata would already be really helpful for further downstream evaluation, so that would be great already in the patched dataset version, for example as a csv or json file with png filenames mapping to that information (but other formats work as well of course).

I think it's fine to go with 512x512 for now as the evaluation would be more an internal comparison of models, and if you are using the 512x512 version in your research now as well, this might be more up to date.

Also in the linked paper, I see that you name the dataset `CaFFe`, should we use that name as well here?

Nora-Go:
Yes, I guess 512x512 would be better :) and yes, it would be great if you could use the name CaFFe (which stands for "CAlving Fronts and where to Find thEm")! Thank you!

I'll provide you with a csv - just give me a little time :) (I'm just waiting for a response from a collaborator)

torchgeo/datasets/glacier_calving_front.py
```diff
@@ -62,15 +63,15 @@ class GlacierCalvingFront(NonGeoDataset):
...
 
     px_class_values: ClassVar[dict[int, str]] = {
-        0: 'background',
-        64: 'ocean',
-        127: 'rock',
-        254: 'glacier',
+        0: 'N/A',
+        64: 'rock',
+        127: 'glacier',
+        255: 'ocean/ice melange',
```
Nora-Go: ocean/ice melange is 254 not 255 :)

nilsleh: thanks!

docs/api/datasets/non_geo_datasets.csv
```diff
@@ -4,6 +4,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `BigEarthNet`_,C,Sentinel-1/2,"CDLA-Permissive-1.0","590,326",19--43,120x120,10,"SAR, MSI"
 `BioMassters`_,R,Sentinel-1/2 and Lidar,"CC-BY-4.0",,,256x256, 10, "SAR, MSI"
 `CaBuAr`_,CD,Sentinel-2,"OpenRAIL",424,2,512x512,20,MSI
+`CaFFe`_,S,"Sentinel-1, TerraSAR-X, TanDEM-X, ENVISAT, European Remote Sensing Satellite 1&2, ALOS PALSAR, and RADARSAT-1","CC-BY-4.0","19092","2 or 4","512x512",6-20,"SAR"
```
adamjstewart, line 7:
```suggestion
`CaFFe`_,S,"Sentinel-1, TerraSAR-X, TanDEM-X, ENVISAT, ERS-1/2, ALOS PALSAR, and RADARSAT-1","CC-BY-4.0","19092","2 or 4","512x512",6-20,"SAR"
```

adamjstewart: Confirmed the license

torchgeo/datamodules/caffe.py
```diff
@@ -0,0 +1,55 @@
...
+class CaFFeDataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the CaFFe dataset.
+
+    Implements the default splits that come with the dataset.
+
+    .. versionadded:: 0.7
+    """
+
+    mean = torch.Tensor([0.5517])
+    std = torch.Tensor([11.8478])
```
adamjstewart, line 25: Same normalization for all channels?

torchgeo/datamodules/caffe.py
```diff
@@ -0,0 +1,55 @@
...
+            num_workers: Number of workers for parallel data loading.
+            size: resize images of input size 512x512 to size x size
+            **kwargs: Additional keyword arguments passed to
+                :class:`~torchgeo.datasets.CaFFe`.
+        """
+        super().__init__(CaFFe, batch_size, num_workers, **kwargs)
+
+        self.train_aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
+            K.Resize(size),
```
adamjstewart, line 43: Tell me more about this resize. What are the original image sizes? Are they consistent? If not, I don't think this will work since it's after collation.

torchgeo/datamodules/caffe.py
```diff
@@ -0,0 +1,55 @@
...
+            data_keys=['image', 'mask'],
+        )
+
+        self.aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
+            K.Resize(size),
+            data_keys=['image', 'mask'],
+        )
+
+        self.size = size
```
adamjstewart, line 55: I would personally put this at the top

torchgeo/datasets/caffe.py
```diff
@@ -0,0 +1,293 @@
...
+    * paired multi-class land cover segmentation masks
+
+    Dataset format:
+
+    * images are single-channel pngs with dimension 512x512
+    * segmentation masks are single-channel pngs
+
+    Dataset classes:
+
+    1. N/A
```
adamjstewart, line 44: Should we start numbering at 0?

torchgeo/datasets/caffe.py
```diff
@@ -0,0 +1,293 @@
...
+
+        img = (
+            read_tensor(
+                os.path.join(
+                    self.root, self.data_dir, self.image_dir, self.split, img_filename
+                )
+            )
+            .unsqueeze(0)
+            .float()
+        )
```
adamjstewart, line 161: These long lines could be condensed quite a bit using something like:
```suggestion
        path = os.path.join(self.root, self.data_dir, self.image_dir, self.split, img_filename)
        img = read_tensor(path).unsqueeze(0).float()
```

torchgeo/datasets/caffe.py
```diff
@@ -0,0 +1,293 @@
...
+        zone_mask = read_tensor(
+            os.path.join(
+                self.root, self.data_dir, self.mask_dirs[1], self.split, zones_filename
+            )
+        ).long()
+
+        zone_mask = self.ordinal_map_zones[zone_mask]
+        front_mask = self.ordinal_map_fronts[front_mask]
+
+        sample = {'image': img, 'mask_front': front_mask, 'mask_zones': zone_mask}
```
adamjstewart, line 176: For our trainers, I was actually hoping to have a single 'mask' key with multiple channels instead of splitting them. Both are compatible with Kornia, but the latter is more likely going to be easy to do in a single generic trainer. This is also how [geodl](https://github.com/maxwell-geospatial/geodl) works (although they require all masks to have a channel dimension, even if there is only a single channel).

adamjstewart (commented):
All minor comments, this is mostly ready to go

nilsleh:
> Hi :) just a clarification on your sample here: The area in the left upper corner is ocean and should have a different color in the zones label. I guess that comes from the solved bug with ocean == 254 and not 255 :)
 Thanks, added the updated plot at the top :)

torchgeo/datamodules/caffe.py
nilsleh: The original size of the dataset patches is 512x512, however, due to memory constraints or specific architecture requirements, it's often necessary to resize to something like 256, or 224. Therefore, I think it is nice to provide this option out of the box, so people don't have to overwrite it.

torchgeo/datasets/caffe.py
nilsleh: This is tricky with pytorch and torchmetrics I think because usually the convention is either [batch_size, num_classes, height, width] with one-hot encoding, or [batch_size, height, width] but then class indices. With a single mask you'd try to do both.

torchgeo/datamodules/caffe.py
nilsleh: It is just a single channel input image.

adamjstewart (approved):

Nora-Go:
and here is the meta data:
[meta_data.csv](https://github.com/user-attachments/files/17519964/meta_data.csv)

let me know whether something is unclear :)

nilsleh:
> and here is the meta data:

@Nora-Go Thanks a lot for providing that. Would it be possible to integrate that into your [data_processing.py script](https://github.com/Nora-Go/Calving_Fronts_and_Where_to_Find_Them/blob/de8e8c36292470fd4037501155db2f5d4aa8ef13/data_processing/data_preprocessing.py) namely that as the png files are turned into 512x512 patches, the respective bounds for that patch and the other metadata are collected in a csv? Then we could upload it to huggingface as well and output that information alongside the data sample.

Nora-Go:
@nilsleh at the moment, I'm unfortunately time-wise constrained. So, I won't be able to do it any time soon - sorry :/

nilsleh:
> This information is not publicly available so far. If you want I can send you the geo tiffs - those have the information and you can extract them to include them here.

Hi @Nora-Go, I was looking at the dataset a bit more and am now wondering whether you could send me the original GeoTiff files? There is some other metadata information usually included in tif files that would be very helpful.

nilsleh:
@Nora-Go Awesome, yeah that's fine! I would probably put it on [Huggingface](https://huggingface.co/datasets/torchgeo/caffe) afterwards just so it's not lost :)

Nora-Go:
Sure :) give me a few minutes

PR #2367, opened 2024-10-24
Add SpaceNet6 datamodule

nilsleh:
This PR adds a DataModule for SpaceNet6. And makes the following changes:
- Introduce `SpaceNetBaseDataModule`
- Add dummy test data for `SpaceNet1`, it exists in the actual dataset, but not in the dummy data
- Add dummy data for `SpaceNet6`
- Resctructure `tests/data/spacenet/` to generate test data for each version

Other Notes:
- The "test" data in `SpaceNet` comes without labels, so in the datamodules, I have configured them to be "predict" data.
- One could also restructure the `data.py` scripts, because there are shared functions that can be utilized across versions.
- it would be nice to have geospatial train/val/test splits generated from the `train` data, and not random splits


tests/datasets/test_spacenet.py
adamjstewart, line 1: Now that we have fake data for SpaceNet 6, should we also test that here? Should be able to parametrize the fixture such that you don't need to add any extra lines of code.

adamjstewart (commented):
Knowing full well how different each SpaceNet version is, I won't try to ask you to combine the data.py to avoid code duplication. Hopefully someday we can have data modules for all versions.

tests/datasets/test_spacenet.py
nilsleh: Was a bit more involved, but I think it works now, except for mypy...

adamjstewart: Easy fix, it's an instance, not a class

adamjstewart (approved):

PR #2368, opened 2024-10-24
Add FTW DataModule

### LINKED ISSUE #2365: Add FTW Datamodule
### Summary

@calebrob6 I would like to use the FTW datasets for some experiments, and saw that there is a `DataModule` in the [FTW-repo](https://github.com/fieldsoftheworld/ftw-baselines/blob/main/src/ftw/datamodules.py) which essentially looks already like a torchgeo datamodule. Wondering if it is possible to add this here as well? I can also open a PR for this.

nilsleh:
Closes #2365 by adding a FTW Datamodule.


calebrob6 (dismissed):
LGTM -- only minor comment is can we have something in the docstring that indicates that you can pass different country sets through to the Dataset object? (also can you verify that this part works as expected -- e.g. if I want to train/test/val on austria and south africa can I do this?)

nilsleh:
> LGTM -- only minor comment is can we have something in the docstring that indicates that you can pass different country sets through to the Dataset object? 

I now added different country arguments for each split, like you have in FTW-repo.

nilsleh:
Is there a good geolocation split that we could make the default, just so that you have something more interesting out of the box?

tests/trainers/test_segmentation.py
```diff
@@ -92,6 +93,8 @@ def test_trainer(
                     'ecec8e871faf1bbd8ca525ca95ddc1c1f5213f40afb94599884bd85f990ebd6b'
                 )
                 monkeypatch.setattr(LandCoverAI, 'sha256', sha256)
+            case 'ftw':
```
adamjstewart, line 96: Maybe move this one up to keep it in alphabetical order

torchgeo/datamodules/ftw.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class FieldsOfTheWorldDataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the FTW dataset.
+
+    .. versionadded:: 0.7
+    """
+
+    mean = torch.tensor([0, 0, 0, 0, 0, 0, 0, 0])
+    std = torch.tensor([3000, 3000, 3000, 3000, 3000, 3000, 3000, 3000])
```
adamjstewart, line 23:
```suggestion
    mean = torch.tensor(0)
    std = torch.tensor(3000)
```
Much easier if all channels are the same

torchgeo/datamodules/ftw.py
```diff
@@ -0,0 +1,90 @@
...
+        self.aug = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std), data_keys=['image', 'mask']
+        )
+
+    def setup(self, stage: str) -> None:
+        """Set up datasets.
+
+        Called at the beginning of fit, validate, test, or predict. During distributed
+        training, this method is called from every process across all the nodes. Setting
+        state here is recommended.
```
adamjstewart, line 75: I don't usually copy this paragraph, it's only in the base class so people know when to use it.

adamjstewart (dismissed):
Only minor comments

adamjstewart (approved):

PR #2370, opened 2024-10-25
Add CROMA pretrained model

nilsleh:
This PR adds the CROMA model based on the implementation in the [repo](https://github.com/antofuller/CROMA/blob/main/use_croma.py), that can load the pretrained weights and return the modality embeddings.

@antofuller Thank you open-sourcing your code and putting the weights on HF. Tagging you in case you want to comment on this PR with any comments, questions, etc.

```python
import torch
from torchgeo.models import CROMALarge_Weights, croma_large

model = croma_large(CROMALarge_Weights.CROMA_VIT)

x_sar = torch.randn(1, 2, 120, 120)
x_opt = torch.randn(1, 12, 120, 120)

out = model(x_sar, x_opt)
```


torchgeo/models/croma.py
```diff
@@ -0,0 +1,649 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# Code based on https://github.com/antofuller/CROMA under MIT License
```
adamjstewart, line 4: Need to check but I don't think we can just copy-n-paste without also copying the entire license (including who owns the copyright). I wish we had a legal team to answer questions like this...

nilsleh:
Hi @antofuller, we are having to copy quiet a few individual blocks, like Attention and ViT, from your code base, and want to give proper attribution. So we would just like to check if that is okay with you. Alternatively (or preferred if possible), I was also wondering whether those blocks are so standard that they could be imported from a library like `torchvision` or `timm` or were they all written from scratch specifically for CROMA?

antofuller:
> Hi @antofuller, we are having to copy quiet a few individual blocks, like Attention and ViT, from your code base, and want to give proper attribution. So we would just like to check if that is okay with you. Alternatively (or preferred if possible), I was also wondering whether those blocks are so standard that they could be imported from a library like `torchvision` or `timm` or were they all written from scratch specifically for CROMA?

Hey! Thanks for doing this :)

These blocks are not standard unfortunately. I wrote them myself — they were inspired by lucidrains various transformer repos. It might be possible to adapt my code to timm's blocks but I'm swamped right now. I hope just copying from my codebase is ok for you folks.

Thanks again, and let me know if you have further questions.

torchgeo/models/croma.py
```diff
@@ -123,17 +123,16 @@ def forward(
         """
         return_dict: dict[str, Tensor] = {}
 
-        device = x_sar.device if x_sar is not None else x_optical.device
-        attn_bias = self.attn_bias.to(device)
```
adamjstewart: @nilsleh check if this is valid. I think if the model is transferred to a device, so will all of its attributes, so this line isn't needed. This line doesn't work in the case where both x_sar and x_optical are None, and adding:
```python
assert x_sar or x_optical
```
wasn't sufficient to convince mypy.

adamjstewart (commented):
Need to run the slow tests locally to confirm the remote weights load without missing/extra keys.

EDIT: Slow tests pass, will check again after updates

torchgeo/models/croma.py
nilsleh: will have to test this live and potentially open another PR to fix it.

torchgeo/models/croma.py
```diff
@@ -0,0 +1,644 @@
...
+
+
+class CROMABase_Weights(WeightsEnum):  # type: ignore[misc]
+    """CROMA base model weights.
+
+    .. versionadded:: 0.7
+    """
+
+    CROMA_VIT = Weights(
+        url='https://hf.co/torchgeo/croma/resolve/17591bcf0711940cc71167d6618d74f60847b43b/CROMA_base- 0238d814.pt',
```
adamjstewart, line 506:
```suggestion
        url='https://hf.co/torchgeo/croma/resolve/17591bcf0711940cc71167d6618d74f60847b43b/CROMA_base-0238d814.pt',
```

adamjstewart: Oh wait, the file you uploaded also contains this space. Can you rename the uploaded file and update the commit hashes in the URL?

adamjstewart:
Weights also need to be documented in docs/api/weights/agnostic.csv (assuming based on code). Also I think there's a table in models.rst for that section listing explicit/implicit dynamic features.

docs/api/weights/agnostic.csv
```diff
@@ -1,4 +1,5 @@
 Weight,Source,Citation,License,Spatial,Temporal,Spectral,m-bigearthnet,m-forestnet,m-brick-kiln,m-pv4ger,m-so2sat,m-eurosat,m-pv4ger-seg,m-nz-cattle,m-NeonTree,m-cashew-plant,m-SA-crop,m-chesapeake
+CROMA,`link <https://github.com/antofuller/CROMA>`__,`link <https://arxiv.org/abs/2311.00566>`__,CC-B>-4.0,implicit,-,implicit,,,,,,,,,,,,
```
adamjstewart:
```suggestion
CROMA,`link <https://github.com/antofuller/CROMA>`__,`link <https://arxiv.org/abs/2311.00566>`__,CC-BY-4.0,implicit,-,implicit,,,,,,,,,,,,
```

adamjstewart (approved):

adamjstewart:
We forgot to add this to `hubconf.py`. Can you open a PR or remind me to?

PR #2371, opened 2024-10-28
Add BigEarthNetv2 dataset

ando-shah:
Basic BENv2 dataset added. `_verify(..)` and checksums missing


torchgeo/datasets/bigearthnetv2.py
```diff
@@ -0,0 +1,664 @@
...
+import torch
+from matplotlib.figure import Figure
+from rasterio.enums import Resampling
+from torch import Tensor
+import pandas as pd
+from torchvision import transforms
+from torchgeo.datasets.geo import NonGeoDataset
+from torchgeo.datasets.utils import download_url, extract_archive, sort_sentinel2_bands
+
+class BigEarthNetv2(NonGeoDataset):
```
nilsleh, line 23: I think between the existing `BigEarthNet` version and version 2, there is a lot of shared code. The `__init__` signature is the same as well, so I would think it's beneficial to inherit from `BigEarthNet` instead of `NonGeoDataset` so that a lot of the code does not have to be written twice, and instead only overwrite the relevant parts.

nilsleh (changes requested):

torchgeo/datasets/bigearthnetv2.py
adamjstewart: Can someone point me to a summary of the changes between v1 and v2? Basically I'm curious if this is just a newer version of the same dataset (slightly more images or better labels) or a completely new iteration. We could either do:
```python
BigEarthNet(version=1)
BigEarthNet(version=2)
```
in the case of the former or:
```python
BigEarthNet()
BigEarthNetv2()
```
in the case of the latter.

ando-shah:



Ando Shah

torchgeo/datasets/bigearthnetv2.py
ando-shah: Yes you're right, I've basically used v1 as the template to write v2. 
v2 is the same dataset with the following changes:

- Better labels
- Better spatial splits
- Fewer images - problematic images removed

The [reBEN paper](http://arxiv.org/abs/2407.03653) details these differences

adamjstewart: Gotcha. In that case I'm kind of leaning towards adding a `version` parameter to the existing dataset. We could default to 2. This would require the fewest code modifications.

torchgeo/datasets/bigearthnetv2.py
```diff
@@ -0,0 +1,664 @@
...
+
+        Returns:
+            list of dicts of s1 and s2 folder paths
+        """
+        filename = self.splits_metadata[self.split]["filename"]
+        dir_s1 = self.metadata_locs["s1"]["directory"]
+        dir_s2 = self.metadata_locs["s2"]["directory"]
+        dir_maps = self.metadata_locs["maps"]["directory"]
+
+        self.metadata = pd.read_parquet(os.path.join(self.root, filename))
```
nilsleh, line 366: Just noticed, that this is missing a filter based on the split for the appropriate set.

torchgeo/datasets/bigearthnetv2.py
```diff
@@ -0,0 +1,664 @@
...
+        num_classes: int = 19,
+        transforms: Optional[Callable[[dict[str, Tensor]], dict[str, Tensor]]] = None,
+        download: bool = False,
+        checksum: bool = False,
+    ) -> None:
+        """Initialize a new BigEarthNet dataset instance.
+
+        Args:
+            root: root directory where dataset can be found
+            split: train/val/test split to load
```
nilsleh, line 301: And in v2, the val split is called "validation"

torchgeo/datasets/bigearthnetv2.py
adamjstewart: We have a similar discussion on naming in #2394

adamjstewart:
Closed by #2531

PR #2387, opened 2024-11-02
Add Annual NLCD dataset

### LINKED ISSUE #2386: Add Annual NLCD Datasets
### Summary

MRLC has released the annual NLCD data products (1985-2023). See [here](https://www.mrlc.gov/data?f%5B0%5D=category%3ALand%20Cover&f%5B1%5D=project_tax_term_term_parents_tax_term_name%3AAnnual%20NLCD).

The download URL looks like this:

```python
year = 2023 # any year between 1985 and 2023
url = f"https://s3-us-west-2.amazonaws.com/mrlc/Annual_NLCD_LndCov_{year}_CU_C1V0.tif"
```

The current torchgeo NLCD dataset look like this:
https://torchgeo.readthedocs.io/en/stable/api/datasets.html#nlcd

```python
url = 'https://s3-us-west-2.amazonaws.com/mrlc/nlcd_{}_land_cover_l48_20210604.zip'

md5s: ClassVar[dict[int, str]] = {
    2001: '538166a4d783204764e3df3b221fc4cd',
    2006: '67454e7874a00294adb9442374d0c309',
    2011: 'ea524c835d173658eeb6fa3c8e6b917b',
    2016: '452726f6e3bd3f70d8ca2476723d238a',
    2019: '82851c3f8105763b01c83b4a9e6f3961',
}
```

Should we update the NLCD class to use the annual NLCD data products or create another NLCD class like `AnnualNLCD`?

giswqs:
This PR updates the existing `NLCD` dataset to use the [Annual NLCD](https://www.usgs.gov/centers/eros/science/annual-national-land-cover-database#overview) data products. Fix #2386 

See this [notebook](https://colab.research.google.com/drive/1rMCIAh0WPgRx3XJd63ayd0vU7iDLsl45?usp=sharing) for the md5s verification.


giswqs:
The PR has passed all checks and is ready for reivew

adamjstewart:
I thought about this in more detail.

At first I was hesitant to override the old class since Annual NLCD turns out to be quite different from Legacy NLCD and this is somewhat backwards incompatible (not reproducible). In which case maybe we should have a new `AnnualNLCD` class. But then I realized "legacy" implies no one should use the old version anymore. Instead of deprecating the old class and eventually renaming the new class, it's probably better to just update the existing class.

My next thought was whether we should have subclasses for the different product categories:

1. Land Cover (LndCov)
2. Land Cover Change (LndChg)
3. Land Cover Confidence (LndCnf)
4. Fractional Impervious Surface (FctImp)
5. Impervious Descriptor (ImpDsc)
6. Spectral Change Day of Year (SpcChg)

So this would then become `NLCDLndCov`, and NLCD would simply become an ABC. But I think it makes more sense to me to have all of these products combined into a single dataset.

I'm going to ask others to chime in on this change and see if they agree with me before merging. TorchGeo is still 0.X so we're okay with backwards incompatible changes right now. I'm more interested in what data products people are interested in and how they will use them so we can design the ideal data loader.

giswqs:
@adamjstewart I agree with you. I will wait for others to chime in before making further changes.

isaaccorley:
I think for the sake of not having to maintain both and confuse users on which one to pick, we can just use the Annual so I will +1 to this PR.

PR #2388, opened 2024-11-04
Remove Chesapeake7 and Chesapeake13

giswqs:
`Chesapeake7` and `Chesapeake13` are not found in the datasets. They should be removed.


adamjstewart (approved):
Ah yes, we recently removed these, good find.

Issue #2303, opened 2024-09-18
Support Datamodule and from the FLAIR project for semantic segmentation


rbavery:
### Summary

I just stumbled across FLAIR, a high res (.2 meter) semantic segmentation dataset of 19 land cover categories. Full description here https://ignf.github.io/FLAIR/#FLAIR1 That link also references a U-Net model baseline trained on FLAIR

It is described in an [OGC Report on ML Engineering](http://ogc.pages.ogc.org/T19-Machine_Learning/documents/D003/document.html#_1c2d42df-b2b4-409a-a30d-20fddb5b0965) as "a comprehensive and high-quality collection of labeled satellite imagery aimed at advancing land cover classification and geospatial analysis tasks". It's maintained by the French National Institute of Geographic and Forest Information (IGN).

### Rationale

I'm interested in composing a list of high quality, challenging datasets for benchmarking semantic segmentation and object detection models and at a glance FLAIR seems like one of them. It seems like the rigor of maintenance and description of the FLAIR dataset is high compared to other datasets and so I would like to see this available in torchgeo. 

Also, I think I recall seeing that torchgeo would like to offer models that are inference-ready, not just pretrained with self-superivison but fine-tuned to address popular tasks in remote sensing. This U-Net model seems like a good start, but I can raise a separate issue for adding the model. 

### Implementation

I haven't contributed to torchgeo before but I would check out other PRs that added datamodules and pretrained models and follow that example.

### Additional information

This comment is on my mind: https://github.com/Clay-foundation/model/discussions/269#discussioncomment-9704291

I'd love for more challenging datasets to be front and center when benchmarking models rather than Eurosat. I'd also like for us to use datasets that we have a solid understanding of the geographic and class distribution and I like that FLAIR lays this out on their site. 

I might be a bit slow to implement this but would like to submit a PR when I have time if this sounds like a good idea.

adamjstewart:
Yes, we would love to have FLAIR in TorchGeo!

If you would like to take a stab at this, see https://torchgeo.readthedocs.io/en/stable/user/contributing.html#datasets for a list of files you'll need to modify. We have dozens of other semantic segmentation datasets you can base your code on. Let us know if you have any trouble with the testing and we would be happy to help!

adamjstewart:
I thought we solved all the problems in #2292?

adamjstewart:
The images are already pre-chipped. Any reason you don't want to make a NonGeoDataset?

[...]

PR #2394, opened 2024-11-05
FLAIR#2 Dataset and Datamodule Integration

MathiasBaumgartinger:
# FLAIR#2 dataset

The `FLAIR #2 <https://github.com/IGNF/FLAIR-2>` dataset is an extensive dataset from the French National Institute  of Geographical and Forest Information (IGN) that provides a unique and rich resource for large-scale geospatial analysis. 
The  dataset is sampled countrywide and is composed of over 20 billion annotated pixels of very high resolution aerial    imagery at 0.2 m spatial resolution, acquired over three years and different months (spatio-temporal domains). 

The FLAIR2 dataset is a dataset for semantic segmentation of aerial images. It contains aerial images, sentinel-2 images and masks for 13 classes. 
    The dataset is split into a training and test set.

    Dataset features:

    * over 20 billion annotated pixels
    * aerial imagery
        * 5x512x512
        * 0.2m spatial resolution
        * 5 channels (RGB-NIR-Elevation)
    * Sentinel-2 imagery
        * 10-20m spatial resolution
        * 10 spectral bands
        * snow/cloud masks (with 0-100 probability)
        * multiple time steps (T)
        * Tx10xWxH, T, W, H are variable
    * label (masks)
        * 512x512
        * 13 classes

    Dataset classes:

    0: "building",
    1: "pervious surface",
    2: "impervious surface",
    3: "bare soil",
    4: "water",
    5: "coniferous",
    6: "deciduous",
    7: "brushwood",
    8: "vineyard",
    9: "herbaceous vegetation",
    10: "agricultural land",
    11: "plowed land",
    12: "other"  

    * https://doi.org/10.48550/arXiv.2310.13336

image

### `NonGeoDataset`, `__init()__`

After discussions following https://github.com/microsoft/torchgeo/issues/2303, we decided that *at least* until faulty mask data are fixed the flair2 ds will be of type `NonGeoDataset`. Other than with common `NonGeoDatasets`, FLAIR2 exposes a `use_toy` and `use_sentinel` argument. The `use_toy`-flag will instead use the toy data which is a small subset of data. The `use_sentinel` argument on the other hand decides whether a sample includes the augmented sentinel data provided by the maintainers of FLAIR2.

### `_verify`, `_download`, `_extract`

As each of the splits/sample-types (i.e. `[train, test], [aerial, sentinel, labels]` are contained in a individual zip download, download and extraction has to happen multiple times. On the other hand, the toy dataset is contained in a singular zip. Furthermore, to map the super-patches of the sentinel data to the actual input image, a `flair-2_centroids_sp_to_patch.json` is required, which has to be equally has to be downloaded as an individual zip.

### `_load_image`, `_load_sentinel`, `_load_target`

For storage reasons, the elevation (5th band) of the image is stored as a uint. The original height thus is multiplied by 5. We decided to divide the height by 5 to get the original height, to make the trained model more usable for other data. See Questions please.

As mentioned previously, additional metadata has to be used to get from the `sentinel.npy` to the actual area. Initially for debugging reasons, we implemented to return not the cropped image but the original data and the cropping-slices (i.e. indices). Consequently, the images can be plot in a more meaningful matter. Otherwise, the resolution is so low that one can hardly recognize features. This was crucial for debugging to find the correct logic (classic y, x instead of x, y ordering mistake). We do not know if this is smart for "production code". See Questions please.
Moreover, the dimensions of the sentinel data $T \times C=10 \times W \times H$ vary both $T$ and $W$, $H$. This is problematic for the `datamodule`. We have not done extensive research, but the varying dimensions seem to bug the module. Disabling the `use_sentinel`-flag will make the module work.

The labels include values from 1 to 19. The datapaper clearly mentions grouping classes $> 13$ into one class `other` due to underrepresentation. We followed this suggestion. Furthermore, rescaling from 0 to 12 was applied. See Questions please.

## Questions

- Do you consider the Elevation rescaling as distortion of the dataset? Shall I exclude it? The argument for it would be easier re-usability on new datasets.
> For storage optimization reasons, this elevation information is multiplied by a factor of 5 and encoded as a 8bit unsigned integer datatype. 

- How shall we load/provide sentinel data? As cropped data or any other way. I do not see the current implementation as fit for production.
  - Also, how do we want to plot it? The small red rectangle in the example plot above is the actual region. The low resolution is quite observable there.

- Shall we rescale the Classes to start from 0? Shall we group the classes as suggested in the datapaper?

- Check integrity in `download_url` does not seem to work (in unit-tests), why?
  - I have to call an own `check_integrity` call otherwise it passes, even if md5s do not match.
  
- The github actions on the forked repo produce a magic ruff error (https://github.com/MathiasBaumgartinger/torchgeo/actions/runs/11687694109/job/32556175383#step:7:1265). Can you help me resolve this mystery? 

## TODOs/FIXMEs

- [x] Extend tests for toy datasets and apply md5 check 
- [x] Find correct band for plotting sentinel
- [ ] Datamodule cannot handle sentinel data yet


nilsleh (changes requested):

docs/api/datasets/non_geo_datasets.csv
```diff
@@ -1,62 +1,64 @@
-Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
```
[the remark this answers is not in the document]
adamjstewart: Probably line endings

MathiasBaumgartinger:
Hi! Sorry for the late response.

First of all, thanks for the review. I will get to work on the proposed changes ;-)

docs/api/datasets/non_geo_datasets.csv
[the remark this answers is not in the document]
MathiasBaumgartinger: Should I try to change to the original line endings, or does it not really matter?

torchgeo/datamodules/flair2.py
```diff
@@ -0,0 +1,74 @@
...
+                Should be a multiple of 32 for most segmentation architectures.
+            val_split_pct: Percentage of the dataset to use as a validation set.
+            num_workers: Number of workers for parallel data loading.
+            augs: Optional augmentations to apply to the dataset.
+            **kwargs: Additional keyword arguments passed to
+                :class:`~torchgeo.datasets.Potsdam2D`.
+        """
+        super().__init__(FLAIR2, batch_size, num_workers, **kwargs)
+
+        self.patch_size = _to_tuple(patch_size)
```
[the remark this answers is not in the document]
MathiasBaumgartinger: I am by no means an expert, but have been reading some papers in the domain the last year. I suppose, the increased feature dimensions provided by the hyperspectral bands of sentinel-2 have shown to increase model performance.

Furthermore, in the datapaper [\[p. 6\]](https://arxiv.org/pdf/2305.14467) of FLAIR2 the authors argue: 

> The interest of the extended spatial information provided by the Sentinel-2 super-patches is particularly visible in the last two rows of Figure 4. Indeed, the location on a beach or on a lake is difficult to determine from the aerial image alone, and could easily be confused with the sea for example in the last row. 

The model leveraging the additional information outperform a standard baseline model:

| Model        | mIoU  | Building | Pervious Surface | Impervious Surface | Bare Soil | Water  | Coniferous | Deciduous | Brushwood | Vineyard | Herbaceous Vegetation | Agricultural Land | Plowed Land |
|--------------|-------|----------|------------------|---------------------|-----------|--------|------------|-----------|-----------|----------|------------------------|-------------------|-------------|
| U-Net        | 0.547 | 0.8009   | 0.4727           | 0.6988              | 0.3076    | 0.7985 | 0.5758     | 0.7014    | 0.2392    | 0.6012   | 0.4653                | 0.5449            | **0.3583**  |
| U-T&T        | 0.5594| 0.8285   | 0.4980           | 0.7204              | 0.2982    | **0.8009** | 0.6041 | 0.7189    | **0.2541** | 0.6580   | 0.4684                | 0.5478            | 0.3157      |
| U-T&T best   | **0.5758** | **0.8368** | **0.4995**     | **0.7446**          | **0.3959** | 0.7952 | **0.6339** | **0.7239** | 0.2485 | **0.6678** | **0.4750**        | **0.5513**        | 0.3381      |

torchgeo/datamodules/flair2.py
```diff
@@ -0,0 +1,74 @@
...
+from ..datasets import FLAIR2
+from ..samplers.utils import _to_tuple
+from ..transforms import AugmentationSequential
+from .geo import NonGeoDataModule
+
+
+class FLAIR2DataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the FLAIR2 dataset.
+
+    Uses the train/test splits from the dataset.
```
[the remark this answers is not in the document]
MathiasBaumgartinger: Maybe I misunderstood a concept here, but are they? The dataset has two splits: `train`, `test` reflected in the directory structure. Train and validation are created randomly from the `train` sub-directories. However, if the current stage/split is `test`, another sub-directory is loaded.

`self.dataset = FLAIR2(split="train", **self.kwargs)`
vs.
`self.test_dataset = FLAIR2(split="test", **self.kwargs)` 

Loaded like FLAIR2 dataset line 333fff.
```py
    os.path.join(
            self.root, 
            self.dir_names[self.split]["images"],
            "**", self.globs["images"]), recursive=True)
```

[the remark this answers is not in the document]
nilsleh: Sorry, not entirely sure what I saw there when I made that comment, it all looks correct.

torchgeo/datamodules/flair2.py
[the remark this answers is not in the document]
nilsleh: Thanks for pointing that out, I should have looked at the paper myself... In that case, if we want to support a resize, it would be more involved, if we wont to resize the sentinel image as well but keeping the resolution difference. Or still just provide the Sentinel 2 scene as context.

MathiasBaumgartinger:
Apart from the sentinel data, I think everything is on track now. Let me know whether you have a preferred way of me handling this.

JacobJeppesen:
I was just testing the pull request, and there was an issue with the download, where it gets a 404 when trying to download `https://storage.gra.cloud.ovh.net/v1/AUTH_366279ce616242ebb14161b7991a8461/defi-ia/flair_data_2/flair-2_centroids_sp_to_patch.zip`. I think reason stems from inconsistent naming, where the zip file is named `flair_2_centroids_sp_to_patch.zip`, with underscore after `flair`, and the json file inside is named `flair-2_centroids_sp_to_patch.json`, with hyphen after `flair`. I.e., the correct download url is `https://storage.gra.cloud.ovh.net/v1/AUTH_366279ce616242ebb14161b7991a8461/defi-ia/flair_data_2/flair_2_centroids_sp_to_patch.zip`.

Great work though! Awesome to see FLAIR being implemented :smiley:

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,659 @@
...
+        to_extract: list[str] = []
+
+        # Check if dataset files (by checking glob) are present already
+        for train_or_test, dir_name in self.dir_names[self.split].items(): 
+            downloaded_path = os.path.join(self.root, dir_name)
+            if not os.path.isdir(downloaded_path):
+                to_extract.append(dir_name)
+                continue
+            
+            files_glob = os.path.join(downloaded_path, "**", self.globs[train_or_test])
```
JacobJeppesen, line 440: When instantiating the datamodule, it will extract the `flair_sen_train` zip-file on every run. I traced it here, and I believe it's becuase when going through when `dir_name="flair_sen_train"`, I get `files_glob="/home/jhj/dataset_test/FLAIR2/flair_sen_train/**/SEN2_*{0}.npy"`, with the curly braces at the end. This makes the `glob.glob()` fail in the next line. I tried manually setting it to `files_glob="/home/jhj/dataset_test/FLAIR2/flair_sen_train/**/SEN2_*.npy"`, which fixed the issue.

JacobJeppesen: Seems like it's the same with `flair_2_sen_test`, which also gets the `{0}` in `files_glob`, and extracts the zip-file on every run.

JacobJeppesen:
Apologize if the comments were a bit scattered, and not really a proper review. I was just trying to run the code, and commented on the go with any issues I encountered :slightly_smiling_face:

MathiasBaumgartinger:
> I was just testing the pull request, and there was an issue with the download, where it gets a 404 when trying to download `https://storage.gra.cloud.ovh.net/v1/AUTH_366279ce616242ebb14161b7991a8461/defi-ia/flair_data_2/flair-2_centroids_sp_to_patch.zip`. I think reason stems from inconsistent naming, where the zip file is named `flair_2_centroids_sp_to_patch.zip`, with underscore after `flair`, and the json file inside is named `flair-2_centroids_sp_to_patch.json`, with hyphen after `flair`. I.e., the correct download url is `https://storage.gra.cloud.ovh.net/v1/AUTH_366279ce616242ebb14161b7991a8461/defi-ia/flair_data_2/flair_2_centroids_sp_to_patch.zip`.
> 
> Great work though! Awesome to see FLAIR being implemented 😃

Weird; I was sure I tried this. Thanks for letting me know.

torchgeo/datasets/flair2.py
MathiasBaumgartinger: > When instantiating the datamodule, it will extract the `flair_sen_train` zip-file on every run. I traced it here, and I believe it's becuase when going through when `dir_name="flair_sen_train"`, I get `files_glob="/home/jhj/dataset_test/FLAIR2/flair_sen_train/**/SEN2_*{0}.npy"`, with the curly braces at the end. This makes the `glob.glob()` fail in the next line. I tried manually setting it to `files_glob="/home/jhj/dataset_test/FLAIR2/flair_sen_train/**/SEN2_*.npy"`, which fixed the issue.

I see why the unexpected behavior appears. For clarification: the format string is necessary, as there are two files inside the directory: `mask` and `data` files. To be able to respectively get only the corresponding mask or data file, I had to format the glob.

Would you be willing to share with me your code snippet so I can debug this real quick?

JacobJeppesen: I just manually removed the `{0}` in the `files_glob` string before running `glob.glob()`:
```
files_glob = os.path.join(downloaded_path, "**", self.globs[train_or_test])
if "flair_sen_train" in files_glob:
    files_glob="/home/jhj/dataset_test/FLAIR2/flair_sen_train/**/SEN2_*.npy"
if not glob.glob(files_glob, recursive=True):
    to_extract.append(dir_name)
```
I.e., hard-coded `files_glob` so I could check if it stopped extracting the zip-file on every run. Not really a solution, but just wanted to see if I was the right place in the code.

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,659 @@
...
+        
+        masks = sorted(glob.glob(os.path.join(
+            self.root, 
+            self.dir_names[self.split]["masks"],
+            "**", self.globs["masks"]), recursive=True))
+        
+        files = [
+            dict(image=image, sentinel=sentinel, mask=mask)
+            for image, sentinel, mask in zip(images, sentinels, masks)
+        ]
```
JacobJeppesen, line 362: It seems like this part filters out most of the aerial images. I think it's because each Sentinel-2 image covers multiple aerial images. E.g., `flair_aerial_train/D004_2021/Z1_NN` contains 100 aerial images, and the corresponding `flair_sen_train/D004_2021/Z1_NN` contains one .npy file. So the association between Sentinel-2 and the aerial images should probably be done on a per-folder basis instead of per-file basis. E.g., something like:
```
        sentinel_lookup = {"/".join(s["data"].split("/")[-4:-2]): s for s in sentinels}
        files = [
            dict(image=image,
                 sentinel=sentinel_lookup["/".join(image.split("/")[-4:-2])],
                 mask=mask)
            for image, mask in zip(images, masks)
        ]
```

MathiasBaumgartinger: Oh wow! What a severe mistake. Thanks for clarifying. Your approach seems very valid.

JacobJeppesen: That happens sometimes :slightly_smiling_face: 

I just tried training a model on the aerial data now with the code posted above, and all seems to be working. I haven't looked closely at the Sentinel-2 data, although they looked correct when I sampled a couple of elements in the `files` list in the debugger.

JacobJeppesen:
It got all the correct aerial imagery files now and the download and extraction is working. However, I'm getting a much more annoying bug now: `IReadBlock failed at X offset ...` random error during training.

I've had this issue before with Rasterio, and it was due to concurrency issues. This shouldn't really happen here, as the dataloader just iterates through a list of files, and each sample is a separate tif file. I can see there's also been a discussion here (https://github.com/microsoft/torchgeo/discussions/594), so perhaps @adamjstewart or @calebrob6 knows what's going on. My own guess is that it could be multiple workers having GDAL do simultaneous scans of the same directory when opening individual files in said directory. I.e., there's often many image files in each sub-folder in the dataset, and GDAL will scan the sub-folder for metadata when opening a file inside it, which could be the culprit. For now at least, I've added the code below in `datasets/flair2.py` and is testing to see if it solves the issue. 

This has been added below the imports in the top: 
``` 
# Set GDAL to avoid scanning read directories to avoid IReadBlock errors in Rasterio when using num_workers>0 in the 
# dataloader. We should not have any concurrency issues for this dataset, as each worker should read individual tif files,
# but it seems like it might occasionally happen when multiple workers scan the same directory simultaneously. 
os.environ["GDAL_DISABLE_READDIR_ON_OPEN"] = "EMPTY_DIR"
```

I've also minimized the time the datareader is open in Rasterio by closing it before doing the tensor operations. This shouldn't really solve the issue here, but added it anyways for good practice. In the `_load_image()` function:
```
with rasterio.open(path) as f:
    array: np.typing.NDArray[np.int_] = f.read()
tensor = torch.from_numpy(array).float()
if "B05" in self.bands:
    # Height channel will always be the last dimension
    tensor[-1] = torch.div(tensor[-1], 5)

```
and in the `_load_target()` function:
```
with rasterio.open(path) as f:
    array: np.typing.NDArray[np.int_] = f.read(1)
tensor = torch.from_numpy(array).long()
# According to datapaper, the dataset contains classes beyond 13
# however, those are grouped into a single "other" class
# Rescale the classes to be in the range [0, 12] by subtracting 1
torch.clamp(tensor - 1, 0, len(self.classes) - 1, out=tensor)
``` 

Not sure if it fixes it, but I'll train a handful of models and try it out. The issue coming from simultaneous directory scans is a bit of a guess.

docs/api/datasets/non_geo_datasets.csv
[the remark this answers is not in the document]
adamjstewart: Change to the original line endings please, that will make `git diff` and `git blame` much more useful

adamjstewart:
Never had this multiprocessing issue before, and I don't see anything in the code that looks sus. Can you reproduce this issue with the unit tests? If not I can also try downloading the dataset and reproducing locally.

JacobJeppesen:
Finally got it debugged, and it was a system error. Download, extract, and training on the aerial data seems to all work as it should. Sorry about the noise! 

Completely unrelated to this PR, but it was a corruption in one of the zpools in ZFS. So if you start using that at some point, and you get weird errors that looks like disk errors, but checking the disk says everything is fine, then try checking the status on your zpools. Spent too much time identifying this :expressionless:

adamjstewart:
Can you resolve the merge conflicts so we can run CI on this PR?

MathiasBaumgartinger:
I think everything should be on track so far. I am getting a ruff error for unsorted tuples in the `[datasets/datamodules]/__init.py__` files. I explicitly tried sorting them using with the vscode sort ascending command ([8779f89](https://github.com/microsoft/torchgeo/pull/2394/commits/8779f89ebfb953c151ec2b1644d133d9e431e7d8)), but apparently this did work :man_shrugging:. Let me know if there is anything else I can do for you =)

adamjstewart:
If you use ruff 0.8.0 it will sort `__all__` correctly. You can also copy-n-paste the file from the latest version and add the new line.

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,762 @@
...
+    }
+
+    @staticmethod
+    def per_band_statistics(
+        split: str, bands: Sequence[str] = all_bands
+    ) -> tuple[list[float], ...]:
+        """Get statistics (min, max, means, stdvs) for each used band in order.
+
+        Args:
+            split (str): Split for which to get statistics (currently only for train)
```
nilsleh, line 197: For the docstring notion in torchgeo, we do not include the type in the docstring again, only function arguments.

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,762 @@
...
+            array: np.typing.NDArray[np.int_] = f.read()
+            tensor = torch.from_numpy(array).float()
+            if 'B05' in self.bands:
+                # Height channel will always be the last dimension
+                tensor[-1] = torch.div(tensor[-1], 5)
+
+        return tensor
+
+    def _load_sentinel(self, path: Path) -> Tensor:
+        # FIXME: should this really be returned as a tuple?
```
nilsleh, line 443: This can be removed, right?

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,762 @@
...
+            self._extract(candidate)
+
+    def _download(self, url: str, suffix: str = '.zip') -> None:
+        """Download the dataset."""
+        download_url(
+            os.path.join(self.url_prefix, f'{url}{suffix}'),
+            self.root,
+            md5=self.md5s.get(url, None) if self.checksum else None,
+        )
+        # FIXME: Why is download_url not checking integrity (tests run through)?
```
nilsleh, line 537: Is this fixed?

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,762 @@
...
+        self._extract()
+
+    def _download(self, url: str, suffix: str = '.zip') -> None:
+        """Download the dataset."""
+        download_url(
+            os.path.join(self.url_prefix, f'{url}{suffix}'),
+            self.root,
+            md5=self.md5s.get(url, None) if self.checksum else None,
+        )
+        # FIXME: Why is download_url not checking integrity (tests run through)?
```
nilsleh, line 754: how about this FIXME?

torchgeo/datamodules/flair2.py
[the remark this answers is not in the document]
nilsleh: I think that's a good idea. Could either be included here, or in a separate PR.

torchgeo/datamodules/flair2.py
```diff
@@ -0,0 +1,79 @@
+"""This module contains the FLAIR2DataModule class for loading the FLAIR2 dataset.
```
nilsleh, line 1: I think this also needs the microsoft heading

tests/data/flair2/FLAIR2/md5s.txt
```diff
@@ -0,0 +1,8 @@
+/home/mathias/Dev/forks/torchgeo/tests/data/flair2/FLAIR2/flair_2_labels_test.zip: b13c4a3cb7ebb5cadddc36474bb386f9
```
nilsleh, line 1: maybe remove the personal directory from this text file.

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,762 @@
...
+            a matplotlib Figure with the rendered sample
+        """
+
+        def normalize_plot(tensor: Tensor) -> Tensor:
+            """Normalize the plot."""
+            return (tensor - tensor.min()) / (tensor.max() - tensor.min())
+
+        rgb_indices = [self.all_bands.index(band) for band in self.rgb_bands]
+        # Check if RGB bands are present in self.bands
+        if not all([band in self.bands for band in self.rgb_bands]):
```
nilsleh, line 572: The Codecoverage is indicating that the RGB Band Missing is not being hit, so I think you just need to add a separate plot test similar to https://github.com/microsoft/torchgeo/blob/2f3e8fda8f1d0d585ac480696434e5c69427cf8a/tests/datasets/test_eurosat.py#L110 for example.

torchgeo/datasets/flair2.py
nilsleh: I think this will also resolve the failing docs test

torchgeo/datasets/flair2.py
MathiasBaumgartinger: I left this here as somewhat of a reminder, because it has the same behavior as the parent class. I.e. currently all tests pass even with wrong md5s.

torchgeo/datamodules/flair2.py
```diff
@@ -0,0 +1,79 @@
...
+        super().__init__(self.ds_class, batch_size, num_workers, **kwargs)
+
+        self.patch_size = _to_tuple(patch_size)
+        self.val_split_pct = val_split_pct
+
+        self.aug: AugmentationSequential = AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std), data_keys=['image', 'mask']
+        )
+
+        self.augs = augs if augs is not None else self.aug
```
JacobJeppesen, line 64: Is `self.augs` intended to act on the data somewhere? My immediate thought was that it would be related to the augmentations part of the base datamodule:https://github.com/microsoft/torchgeo/blob/fedf99375535f801565856cd774bfa9e5a251d55/torchgeo/datamodules/geo.py#L70-L79 But it's not in there, and setting it to an arbitrary value doesn't seem to do anything. Or perhaps I'm missing something?

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,762 @@
...
+            Tensor: the loaded image
+        """
+        with rasterio.open(path) as f:
+            array: np.typing.NDArray[np.int_] = f.read()
+            tensor = torch.from_numpy(array).float()
+            if 'B05' in self.bands:
+                # Height channel will always be the last dimension
+                tensor[-1] = torch.div(tensor[-1], 5)
+
+        return tensor
```
JacobJeppesen, line 440: Should the bands perhaps be extracted here based on `self.bands`? E.g., something like
```
    def _load_image(self, path: Path) -> Tensor:
        """Load a single image.
        Args:
            path: path to the image
        Returns:
            Tensor: the loaded image
        """
        with rasterio.open(path) as f:
            array: np.typing.NDArray[np.int_] = f.read()
            tensor = torch.from_numpy(array).float()
            if 'B05' in self.bands:
                # Height channel will always be the last dimension
                tensor[-1] = torch.div(tensor[-1], 5)

        # Extract the bands to be used. E.g., self.bands=("B01", "B02", "B03") will extract the RGB bands.
        tensor = tensor[[int(band[-2:]) - 1 for band in self.bands]]

        return tensor
``` 
Then when a user has defined n bands in `self.bands`, the returned sample will only contain those bands, instead of all five. Perhaps `self.bands` should also be renamed to `self.aerial_bands`, and a `self.sentinel_bands` should be added(?)

torchgeo/datamodules/flair2.py
nilsleh: `self.aug` is applied [here](https://github.com/microsoft/torchgeo/blob/fedf99375535f801565856cd774bfa9e5a251d55/torchgeo/datamodules/geo.py#L140) if the split specific augmentations are not specified

JacobJeppesen: Yes, that was also my understanding. But I couldn't find `self.augs` anywhere. I.e., if the user instantiates the datamodule now with augmentations provided through the `augs` parameter, they won't have an effect, and the default augmentations will still be used. It could also be a typo, and the intention was `self.aug = augs if augs is not None else self.aug`.

MathiasBaumgartinger:
Thanks everyone for reviewing. Tried to apply all suggested changes. 

As for the checks, according to the log we face the following error: `/home/docs/checkouts/readthedocs.org/user_builds/torchgeo/checkouts/2394/docs/api/datasets.rst:194: ERROR: "csv-table" widths do not match the number of columns in table (10).`
Truly, I have messed something up here (i.e. something with line endings, separations and quotation marks). However, I resolved this problem and the error in the check seems to persist.

docs/api/datasets/non_geo_datasets.csv
```diff
@@ -18,6 +18,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `FAIR1M`_,OD,Gaofen/Google Earth,"CC-BY-NC-SA-3.0","15,000",37,"1,024x1,024",0.3--0.8,RGB
 `Fields Of The World`_,"S,I",Sentinel-2,"Various","70795","2,3",256x256,10,MSI
 `FireRisk`_,C,NAIP Aerial,"CC-BY-NC-4.0","91,872",7,"320x320",1,RGB
+`FLAIR2`_,S,"IGN, Sentinel-2",OPENLICENSE-2.0,7741,13--18,512x512,0.2--20,"RGB+NIR+NDSM, MSI",
```
adamjstewart, line 21:
```suggestion
`FLAIR2`_,S,"IGN, Sentinel-2",OPENLICENSE-2.0,7741,13--18,512x512,0.2--20,"RGB+NIR+NDSM, MSI"
```
This should fix the doc tests

tests/data/flair2/data.py
adamjstewart, line 1: We seem to be creating a **lot** of fake test, is it possible to reduce this number and still test things appropriately?

torchgeo/datamodules/flair2.py
```diff
@@ -0,0 +1,85 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+#
+# FLAIR dataset is released under open license 2.0
+# ..... https://www.etalab.gouv.fr/wp-content/uploads/2018/11/open-licence.pdf
+# ..... https://ignf.github.io/FLAIR/#FLAIR2
```
adamjstewart, line 6: No need for this, it's already included in the docs. The important thing is that this file is MIT licensed like the rest of TorchGeo.

docs/api/datasets/non_geo_datasets.csv
adamjstewart: The correct SPDX license identifier is actually `etalab-2.0`, not `OPENLICENSE-2.0`. There are apparently a lot of "open license" licenses: https://spdx.org/licenses/

tests/datamodules/test_flair2.py
adamjstewart, line 1: We can delete this file and instead test the dataset using a `tests/conf/flair2.yaml` file and 1 line of code in `tests/trainers/test_segmentation.py`. We are actively trying to get rid of `tests/datamodules` since it doesn't test compatibility.

tests/datasets/test_flair2.py
```diff
@@ -0,0 +1,150 @@
...
+        params=[
+            (split, init_class)
+            for split in ['train', 'test']
+            for init_class in [FLAIR2, FLAIR2Toy]
+        ]
+    )
+    def dataset(
+        self, monkeypatch: MonkeyPatch, tmp_path: Path, request: SubRequest
+    ) -> FLAIR2:
+        md5s = {
```
adamjstewart, line 34: Could also just skip `checksum=True` and not bother monkeypatching any MD5s

torchgeo/datamodules/flair2.py
```diff
@@ -0,0 +1,85 @@
...
+
+"""This module contains the FLAIR2DataModule class for loading the FLAIR2 dataset.
+
+The FLAIR dataset is released under open license 2.0:
+- https://www.etalab.gouv.fr/wp-content/uploads/2018/11/open-licence.pdf
+- https://ignf.github.io/FLAIR/#FLAIR2
+
+Code for loading dataset licensed under the MIT License.
```
adamjstewart, line 14: No need for this either, we don't document this for our other hundred datasets

torchgeo/datamodules/flair2.py
```diff
@@ -0,0 +1,85 @@
...
+    Uses the train/test splits from the dataset.
+    """
+
+    def __init__(
+        self,
+        batch_size: int = 64,
+        patch_size: tuple[int, int] | int = 64,
+        val_split_pct: float = 0.2,
+        num_workers: int = 0,
+        use_toy: bool = False,
```
adamjstewart, line 41: I'd rather have a separate `FLAIR2ToyDataModule` class. It can be a subclass.

torchgeo/datamodules/flair2.py
```diff
@@ -0,0 +1,85 @@
...
+            batch_size: Size of each mini-batch.
+            patch_size: Size of each patch, either ``size`` or ``(height, width)``.
+                Should be a multiple of 32 for most segmentation architectures.
+            val_split_pct: Percentage of the dataset to use as a validation set.
+            num_workers: Number of workers for parallel data loading.
+            use_toy: Whether to use the toy version of the dataset.
+            augs: Optional augmentations to apply to the dataset.
+            **kwargs: Additional keyword arguments passed to :class:`~torchgeo.datasets.FLAIR2`.
+
+            ..versionadded:: 0.7
```
adamjstewart, line 57:
```suggestion
        .. versionadded:: 0.7
```

torchgeo/datamodules/flair2.py
```diff
@@ -0,0 +1,85 @@
...
+        """Initialize a new FLAIR2DataModule instance.
+
+        Args:
+            batch_size: Size of each mini-batch.
+            patch_size: Size of each patch, either ``size`` or ``(height, width)``.
+                Should be a multiple of 32 for most segmentation architectures.
+            val_split_pct: Percentage of the dataset to use as a validation set.
+            num_workers: Number of workers for parallel data loading.
+            use_toy: Whether to use the toy version of the dataset.
+            augs: Optional augmentations to apply to the dataset.
```
adamjstewart, line 54: We have an ongoing debate on how best to make augmentations customizable. For now, I would avoid adding this until that debate settles down.

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,800 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+#
+# FLAIR dataset is realeasd under open license 2.0
+# ..... https://www.etalab.gouv.fr/wp-content/uploads/2018/11/open-licence.pdf
+# ..... https://ignf.github.io/FLAIR/#FLAIR2
```
adamjstewart, line 6: Wouldn't bother with this, it's already documented in the docs

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,800 @@
...
+
+class FLAIR2(NonGeoDataset):
+    """FLAIR #2 (The French Land cover from Aerospace ImageRy) dataset.
+
+    The `FLAIR #2 <https://github.com/IGNF/FLAIR-2>` dataset is an extensive dataset from the French National Institute
+    of Geographical and Forest Information (IGN) that provides a unique and rich resource for large-scale geospatial analysis.
+    The  dataset is sampled countrywide and is composed of over 20 billion annotated pixels of very high resolution aerial
+    imagery at 0.2 m spatial resolution, acquired over three years and different months (spatio-temporal domains).
+
+    The FLAIR2 dataset is a dataset for semantic segmentation of aerial images. It contains aerial images, sentinel-2 images and masks for 13 classes.
```
adamjstewart, line 38: All lines should be wrapped to 88 chars

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,800 @@
...
+
+from .errors import DatasetNotFoundError, RGBBandsMissingError
+from .geo import NonGeoDataset
+from .utils import Path, check_integrity, download_url, extract_archive
+
+
+class FLAIR2(NonGeoDataset):
+    """FLAIR #2 (The French Land cover from Aerospace ImageRy) dataset.
+
+    The `FLAIR #2 <https://github.com/IGNF/FLAIR-2>` dataset is an extensive dataset from the French National Institute
```
adamjstewart, line 33:
```suggestion
    The `FLAIR #2 <https://github.com/IGNF/FLAIR-2>`_ dataset is an extensive dataset from the French National Institute
```

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,800 @@
...
+        * snow/cloud masks (with 0-100 probability)
+        * multiple time steps (T)
+        * Tx10xWxH, T, W, H are variable
+    * label (masks)
+        * 512x512
+        * 13 classes
+
+    Dataset classes:
+
+    0: "building",
```
adamjstewart, line 60: I think this has to be 0. instead of 0: in order for the list to be formatted correctly, but let's see once the docs build

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,800 @@
...
+    7: "brushwood",
+    8: "vineyard",
+    9: "herbaceous vegetation",
+    10: "agricultural land",
+    11: "plowed land",
+    12: "other"
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://doi.org/10.48550/arXiv.2310.13336
```
adamjstewart, line 76: Should probably cite the FLAIR #2 paper (https://arxiv.org/abs/2305.14467) instead of the FLAIR paper

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,800 @@
...
+    """
+
+    splits: ClassVar[Sequence[str]] = ('train', 'test')
```
adamjstewart, line 79: No need to annotate this, a tuple is fine

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,800 @@
...
+
+    url_prefix: ClassVar[str] = (
```
adamjstewart, line 81: Same here, it's implied that it's a class var

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,800 @@
...
+            'olive',  # brushwood
+            'purple',  # vineyard
+            'lime',  # herbaceous vegetation
+            'yellow',  # agricultural land
+            'orange',  # plowed land
+            'red',  # other
+        ]
+    )
+
+    statistics: ClassVar[dict[str, dict[str, dict[str, float]]]] = {
```
adamjstewart, line 176: Normalization should happen in the datamodule, not in the dataset

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,800 @@
...
+        """Initialize a new FLAIR2 dataset instance.
+
+        Args:
+            root: root directory where dataset can be found
+            split: which split to load, one of 'train' or 'test'
+            aerial_bands: which bands to load (B01, B02, B03, B04, B05)
+            transforms: optional transforms to apply to sample
+            download: whether to download the dataset if it is not found
+            checksum: whether to verify the dataset using checksums
+            use_sentinel: whether to use sentinel data in the dataset # FIXME: sentinel does not work with dataloader due to varying dimensions
```
adamjstewart, line 261: How much do they vary? Can we resize or crop to a consistent dimension?

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,800 @@
...
+            sentinel_bands: which bands to load from sentinel data (B01, B02, ..., B10)
+
+        Raises:
+            DatasetNotFoundError
+
+        ..versionadded:: 0.7
```
adamjstewart, line 267:
```suggestion
        .. versionadded:: 0.7
```

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,800 @@
...
+            split: which split to load, one of 'train' or 'test'
+            aerial_bands: which bands to load (B01, B02, B03, B04, B05)
+            transforms: optional transforms to apply to sample
+            download: whether to download the dataset if it is not found
+            checksum: whether to verify the dataset using checksums
+            use_sentinel: whether to use sentinel data in the dataset # FIXME: sentinel does not work with dataloader due to varying dimensions
+            sentinel_bands: which bands to load from sentinel data (B01, B02, ..., B10)
+
+        Raises:
+            DatasetNotFoundError
```
adamjstewart, line 265: Needs a description for when this error is raised

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,800 @@
...
+        )
+        # FIXME: Why is download_url not checking integrity (tests run through)?
+        assert check_integrity(
+            os.path.join(self.root, f'{url}{suffix}'),
+            self.md5s.get(url, None) if self.checksum else None,
+        )
+
+    def _extract(self, file_path: str) -> None:
+        """Extract the dataset."""
+        assert isinstance(self.root, str | os.PathLike)
```
adamjstewart, line 576: No need to assert this, it's already checked by mypy

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,800 @@
...
+        Args:
+            sample: a sample return by :meth:`__getitem__`
+            show_titles: flag indicating whether to show titles above each panel
+            suptitle: optional suptitle to use for figure
+
+        Returns:
+            a matplotlib Figure with the rendered sample
+        """
+
+        def normalize_plot(tensor: Tensor) -> Tensor:
```
adamjstewart, line 597: I would instead use `.utils.percentile_normalization`

torchgeo/datasets/flair2.py
```diff
@@ -0,0 +1,800 @@
...
+
+        # Obtain mask and predictions if available
+        mask = sample['mask'].numpy().astype('uint8').squeeze()
+
+        showing_predictions = 'prediction' in sample
+        predictions = None
+        if showing_predictions:
+            predictions = sample['prediction'].numpy().astype('uint8').squeeze()
+
+        # Remove none available plots
```
adamjstewart, line 639:
```suggestion
        # Remove non-available plots
```

adamjstewart (commented):
Does anyone have general opinions on whether we should call this:
```python
FLAIR2()
```
or:
```python
FLAIR(version=2)
```

JacobJeppesen:
> Does anyone have general opinions on whether we should call this:
> 
> ```python
> FLAIR2()
> ```
> 
> or:
> 
> ```python
> FLAIR(version=2)
> ```

I think it'd make sense to use `FLAIR(version=2)`, as it seems like each new version is a superset of the previous version. Most users will probably use the latest version, so if they are individual datasets, `FLAIR1()` and `FLAIR2()` might end up as somewhat unused datasets once `FLAIR3()` is released. As I understand https://github.com/microsoft/torchgeo/issues/2303#issuecomment-2390977539, once version 3 is released, version 1 and 2 data can be directly loaded from version 3 by filtering the files. So the lowest complexity solution would probably be a `FLAIR()` dataset, where version 3 is the full dataset, version 2 is reduced coverage by filtering files/area, and version 1 is the same reduced coverage, but only aerial.

MathiasBaumgartinger:
> > Does anyone have general opinions on whether we should call this:
> > ```python
> > FLAIR2()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > or:
> > ```python
> > FLAIR(version=2)
> > ```
> 
> I think it'd make sense to use `FLAIR(version=2)`, as it seems like each new version is a superset of the previous version. Most users will probably use the latest version, so if they are individual datasets, `FLAIR1()` and `FLAIR2()` might end up as somewhat unused datasets once `FLAIR3()` is released. As I understand [#2303 (comment)](https://github.com/microsoft/torchgeo/issues/2303#issuecomment-2390977539), once version 3 is released, version 1 and 2 data can be directly loaded from version 3 by filtering the files. So the lowest complexity solution would probably be a `FLAIR()` dataset, where version 3 is the full dataset, version 2 is reduced coverage by filtering files/area, and version 1 is the same reduced coverage, but only aerial.

I am in contact with @agarioud. If I do understand him correctly, I doubt that new datasets will have the *requirement* to be backward-capable. 

> Unfortunately there is no direct compatibility with FLAIR#1/#2 apart the aerial images as we reworked the supervision (land cover and now LPIS)

But I agree. Adding versioning to `FLAIR()` will probably result in less dead code. From a design perspective: 
1. We have a parent dataset `FLAIR` handling all the common logic. By passing a specific version (default will always be latest) a new child class is initialized which overrides version specific logic. 
2. We have a parent datamodule `FLAIRModule`. Which passes a version down to datasets. I.e. probably only a single `FLAIRModule` is necessary (no inheritance).
3. Points 1 and 2 will be applied for `FLAIRToy` and `FLAIRToyModule` too (probably both will inherit from the corresponding non-toy classes.)

agarioud:
The upcoming FLAIR-INC dataset will include all data from FLAIR#1 (aerial images) and FLAIR#2 (which added Sentinel-2 data that were previously NPY files covering larger extents but will now have the same spatial extent as the aerial patches and be in TIFF format). Additionally, it will introduce five new modalities and a second supervision dataset. 
If needed, a toy dataset can already be shared with @MathiasBaumgartinger 

Therefore, I also believe a FLAIR() dataset would probably be more efficient ?

adamjstewart:
It's not really a matter of compatibility. Basically, we either use:
```python
class FLAIR(NonGeoDataset, abc.ABC):
    # shared base class

class FLAIR2(FLAIR):
    # override specific stuff
```
or:
```python
class FLAIR(NonGeoDataset):
    def __init__(self, version=2, ...):
        if version == 2:
            # override specific stuff
```
It's more a question of whether the devs think of this as a new version of an existing dataset or a new dataset. We usually go with the former because it offers a bit better reproducibility (if the default `version` gets changed, then reproducibility is broken). See our `EuroSAT` dataset for an example of this. I think the only cases where we use `version` instead are in `MoCoTask` and `SimCLRTask`. I'm fine with both solutions, just want to make sure we're all on the same page.

tests/datamodules/test_flair2.py
MathiasBaumgartinger: Do the tests in `tests/trainers/test_segmentation.py` actually work already? For me, 35 fail and 14 pass

tests/datasets/test_flair2.py
MathiasBaumgartinger: I think I will leave it for now. In fact this is the reason i detected [torchgeo/datasets/flair2.py#L567-L571](https://github.com/MathiasBaumgartinger/torchgeo/blob/2289568a5f850255a3b74d13d649b3daf81e3875/torchgeo/datasets/flair2.py#L567-L571)

tests/datamodules/test_flair2.py
adamjstewart: They work in CI. What error messages are you seeing? Note that PyTorch doesn't yet fully support Python 3.13 so some of the checkpointing tests will fail.

tests/datasets/test_flair2.py
adamjstewart: The reason that download_url isn't checking the MD5 is because we monkeypatch it to avoid downloading anything.

adamjstewart:
Is this still a WIP? Would love to include this in the 0.7.0 release next week.

adamjstewart:
Not finalized yet but hopefully by the end of this week.

adamjstewart:
We're planning 0.9 this week, is this still a WIP?

adamjstewart:
#3181 (FLAIR-HUB) is pretty close to being merged, do you still want to revive this PR or should we close? I'm completely fine with adding FLAIR#1 or FLAIR#2 for reproducibility even though FLAIR-HUB is the successor.

PR #2396, opened 2024-11-07
Deprecate AugmentationSequential wrapper

adamjstewart:
Thoughts on whether we should deprecate this or just remove it? It's been deprecated since 0.4, but also we've been using it continuously since then. We're moving towards a 1.0 release in the future after which we'll need to start formally deprecating things anyway.


adamjstewart:
It looks like NASA Marine Debris and VHR-10 data modules still use the deprecated wrapper?

ashnair1:
> Thoughts on whether we should deprecate this or just remove it? It's been deprecated since 0.4, but also we've been using it continuously since then. We're moving towards a 1.0 release in the future after which we'll need to start formally deprecating things anyway.

I prefer just removing it tbh. But I would like to verify kornia's augmentations work as expected for the detection datasets before that.

> It looks like NASA Marine Debris and VHR-10 data modules still use the deprecated wrapper?

Yeah, I was planning to have the object detection datasets switch to Kornia's `AugmentationSequential` in #1978

tests/transforms/test_transforms.py
```diff
@@ -82,7 +82,8 @@ def test_augmentation_sequential_gray(batch_gray: dict[str, Tensor]) -> None:
         'bbox_xyxy': torch.tensor([[0.0, 0.0, 2.0, 2.0]], dtype=torch.float),
         'labels': torch.tensor([[0, 1]]),
     }
-    augs = K.AugmentationSequential(K.RandomHorizontalFlip(p=1.0), data_keys=None)
+    with pytest.deprecated_call():
+        augs = AugmentationSequential(K.RandomHorizontalFlip(p=1.0), data_keys=None)
```
adamjstewart, line 86: These tests were designed for our wrapper, no need to test the upstream implementation. Should we test our deprecated wrapper or just remove these tests?

ashnair1: I think we can go ahead and remove them. Don't see a reason to maintain them now.

tests/transforms/test_transforms.py
```diff
@@ -6,7 +6,7 @@
 import torch
 from torch import Tensor
 
-from torchgeo.transforms import indices
+from torchgeo.transforms import AugmentationSequential, indices
 from torchgeo.transforms.transforms import _ExtractPatches
 
 # Kornia is very particular about its boxes:
```
ashnair1, line 12: Just wanted to add a clarifying note here.

The format in which kornia expects boxes to be depends on the key used.

1. If `bbox` key is used, it expects boxes to be in `vertices_plus` mode i.e. `(N, 4, 2)` shaped tensor 
2. If `bbox_xyxy` key is used, it expects boxes to be in `xyxy_plus` mode i.e. `(N, 4)` shaped tensor -> (`xmin, ymin, xmax, ymax`)
3. If `bbox_xywh` key is used, it expects boxes to be in `xywh` mode i.e. `(N, 4)` shaped tensor -> (`xmin, ymin, width, height`)


Refer to:  
1.  [Box key - Box mode mapping](https://github.com/kornia/kornia/blob/10b59b5e1751117519497b9c52c2281a2535fc35/kornia/augmentation/container/augment.py#L609)
2. [Box formats](https://github.com/kornia/kornia/blob/10b59b5e1751117519497b9c52c2281a2535fc35/kornia/geometry/boxes.py#L442)

adamjstewart: Yep, this comment was added out of frustration back when the format seemed to change with every release. It has since settled down, so this comment is likely no longer necessary.

ashnair1 (approved):

calebrob6 (approved):
Slight preference for just removing it as we will definitely want to remove it later

PR #2402, opened 2024-11-10
TreeSatAI: Add new dataset

adamjstewart:
TODO:

- [x] Implement dataset
- [x] Implement data module
- [x] Add docs
- [x] Add tests
- [ ] Add support for regression?
- [ ] Add support for semantic segmentation?
- [ ] Add support for keypoint detection?
- [ ] Add support for 200m images?
- [ ] Add time series extension?

@ando-shah @nilsleh 

@cgava25 Thanks for making your dataset public and for the excellent documentation! Hopefully this PR will make your dataset even easier to use for the ML community. Let me know if you have any suggestions for ways in which the data loaders should be modified.

treesatai


torchgeo/datasets/treesatai.py
```diff
@@ -0,0 +1,294 @@
...
+
+    def __len__(self) -> int:
+        """Return the number of data points in the dataset.
+
+        Returns:
+            Length of the dataset.
+        """
+        return len(self.files)
+
+    def __getitem__(self, index: int) -> dict[str, Tensor]:
```
nilsleh, line 177: This looks like it returns the multilabel? Shouldn't we support the different tasks possible with the data, so multiclass and multilabel?

adamjstewart: Multiclass would just be the majority class. I think it's less interesting and unique than multilabel. We can't easily support both without an init parameter.

I would like to add semantic segmentation someday but it requires us to rasterize the GeoJSON files which was more work than I wanted to do on a Sunday.

There are also keypoint detection labels but we don't have a trainer for this and I'm less familiar with how that works.

adamjstewart: Actually it's a bit more complicated than that. There is:

* 20 species classes (single labels)
* 12 age classes (single labels)
* 15 genus classes (multi labels)

The first two can be found in the filename (and are majority class). The latter is found in the labels JSON file I'm currently using. It isn't super clear in the paper which task they are reporting accuracies for.

adamjstewart: I'm also not sure whether this is "soft" multilabel classification or multivariate regression. We're basically estimating the percent of an image that corresponds to different classes.

nilsleh: Are the labels representing, what fraction of the image is covered by a tree class? or to what confidence degree a tree from that class is present in the image? Intuitively from the labels it would appear that it something like the former, since multilabel confidences do not sum to 1, they would be independent confidences per label right?

adamjstewart: Yes, the former.

adamjstewart: Our `MultiLabelClassificationTask` doesn't work because the metrics require binary labels, so I implemented it as a `RegressionTask` for now. Could convert the percentages to binary labels if we want support for multi-label classification.

ando-shah (approved):
Works as advertised. Would be nice to have timeseries support as a follow-on

PR #2405, opened 2024-11-12
Add configurable metadata file and advanced filtering options for BioMassters dataset

PedroConrado:
This PR adds a parameter to allow passing different metadata files for the BioMassters dataset and implements new metadata-based filtering options. If the default metadata file is used, the behavior remains unchanged, ensuring backward compatibility. The additional filters, controlled by new parameters (max_cloud_percentage, max_red_mean, and include_corrupt), enable more refined dataset selection without impacting the default usage scenario.

### DIFF
--- a/torchgeo/datasets/biomassters.py
+++ b/torchgeo/datasets/biomassters.py
@@ -53,7 +53,7 @@ class BioMassters(NonGeoDataset):
     valid_splits = ('train', 'test')
     valid_sensors = ('S1', 'S2')
 
-    metadata_filename = 'The_BioMassters_-_features_metadata.csv.csv'
+    default_metadata_filename = 'The_BioMassters_-_features_metadata.csv.csv'
 
     def __init__(
         self,
@@ -61,6 +61,10 @@ class BioMassters(NonGeoDataset):
         split: str = 'train',
         sensors: Sequence[str] = ['S1', 'S2'],
         as_time_series: bool = False,
+        metadata_filename: str = default_metadata_filename,
+        max_cloud_percentage: float | None = None,
+        max_red_mean: float | None = None,
+        include_corrupt: bool = True,
     ) -> None:
         """Initialize a new instance of BioMassters dataset.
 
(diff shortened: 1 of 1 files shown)


torchgeo/datasets/biomassters.py
```diff
@@ -74,6 +78,10 @@ def __init__(
                 Sentinel 2 ('S1', 'S2')
             as_time_series: whether or not to return all available
                 time-steps or just a single one for a given target location
+            metadata_filename: metadata file to be used
+            max_cloud_percentage: maximum allowed cloud percentage for images
+            max_red_mean: maximum allowed red_mean value for images
+            include_corrupt: whether to include images marked as corrupted
```
adamjstewart, line 85: Depending on what we decide, we need to add something like the following to the docstring:
```rst
.. versionadded:: 0.7
   The *metadata_filename*, *max_cloud_percentage*, ... parameters.
```

PedroConrado: thank you for the reply. I agree with the filtering on the data loader, but I didn't find any datamodule for this one.

PedroConrado: thank you for the reply. I agree with the filtering on the data loader, but I didn't find any datamodule for this one.

nilsleh:
> I'm generally _okay_ with making the metadata file configurable (I did the same thing for SatlasPretrain so I could override the defaults).
> 
> I'm generally _opposed_ to adding too much preprocessing/filtering to the data loader. Personally I would prefer to see this kind of stuff in the data module.
> 
> @nilsleh wrote the original data loader so I'll let him review. Thanks for the contribution!

Trying to think whether it would make more sense to have a dedicated function, that can be overwritten to do the filtering. Otherwise there might be many more key words and a "default" is difficult to define.

PedroConrado:
> > I'm generally _okay_ with making the metadata file configurable (I did the same thing for SatlasPretrain so I could override the defaults).
> > I'm generally _opposed_ to adding too much preprocessing/filtering to the data loader. Personally I would prefer to see this kind of stuff in the data module.
> > @nilsleh wrote the original data loader so I'll let him review. Thanks for the contribution!
> 
> Trying to think whether it would make more sense to have a dedicated function, that can be overwritten to do the filtering. Otherwise there might be many more key words and a "default" is difficult to define.

I implemented it this way because I needed specific filters, but I understand that a more general approach could be beneficial. I’m open to making adjustments if needed.

adamjstewart:
Is this still a WIP? Would love to include this in the 0.7.0 release next week.

torchgeo/datasets/biomassters.py
adamjstewart: You can add a datamodule if you want 🙂

adamjstewart:
No response from PR author and no updates in 2+ years, PR appears to be abandoned, closing.

Issue #2382, opened 2024-10-31
Time Series Support


adamjstewart:
This issue is designed to track progress towards full time series support in TorchGeo. If you are interested in helping with this effort, please join the `#time-series` channel on [Slack](https://join.slack.com/t/torchgeo/shared_invite/zt-22rse667m-eqtCeNW0yI000Tl4B~2PIw) and attend our weekly meetings.

#### Required changes

- [x] Replace rtree with geopandas: #2747
- [x] Replace fiona with geopandas: #2962
- [x] Spatial-only intersection: #2837
- [x] Xarray support: #509, #1490, #2945
- [x] Collation functions for padding: #2921
- [x] Convert change detection datasets: #2862, #2863, #2878, #2879, #2906, #3028
- [x] Return time series: #2308, #3183
- [ ] Convert time series datasets: #3042, #2921, #3368, #3369, #3371, #3372
- [ ] Splitters: guarantee non-overlapping spatial and/or temporal splits: @hkristen 

#### Desired changes

- [x] Replace BoundingBox with slicing: #2804
- [x] Support Polygon ROIs in all splitters: #2835
- [x] Deprecate BoundingBox: #2847
- [x] Store true image footprint: #2903
- [ ] Properly handle vector datasets: @isaaccorley
- [ ] Cache geopandas index: @isaaccorley
- [ ] STACDataset: see #stac channel
- [ ] ERA5
- [ ] CMIP6
- [ ] Support integer indexing?: @adamjstewart
- [ ] Unify GeoDataset and NonGeoDataset?: #1353
- [ ] Intersection/Union multiple queries?: #2408
- [ ] Tuple/MultiQuery dataset?: #877

#### Required changes

- [x] Support Polygon ROIs in all samplers: #2812
- [x] Separate spatial and temporal samplers: #3552
- [x] Ensure samplers do not sample outside Polygon bounds: #3552
- [x] Visualization: visualize an entire epoch with polygons and points: #3552

#### Desired changes

- [ ] Support weighted samplers?: #757
- [ ] Pre-computation?: run sample_points once per epoch for debugging?: #2300

### Transforms

- [x] B x T x C x H x W <-> B x TC x H x W: #2477 (TODO: image + mask)
- [ ] Temporal embeddings: #2826
- [ ] Temporal up/downsampling: https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.interpolate.html
- [ ] UTC time <-> local time
[...]

PR #2408, opened 2024-11-13
Multi-bbox timeseries querying

sfalkena:
This is a PR that is part of TorchGeo Timeseries support (#2382 - Return time series). The goal is to provide an interface that allows for the following options for querying: 

`sample = dataset[bbox]`
`sample = dataset[[bbox1, bbox2, bbox3]]`
`sample = dataset[([dataset1_bbox1], [dataset2_bbox1, dataset2_bbox2])]`

The idea is that anything within a single bbox get's merged into a single raster, and that any query with multiple bboxes will stack along the time dimension. Querying with a tuple of (iterable) bboxes would split the subqueries to different datasets.

As long as this PR is in draft mode, I will keep `test.ipynb` for anyone interested in helping to try out the new method. 

**Current status:** The first two ways of querying has been implemented for the RasterDataset and tested with a small example.

What has changed:
- I've refactored quite a bit to reduce the complexity of  `__init__` and `__get_item__` methods. `try_set_metadata, _get_bounds, _compile_and_check_filename_regex` and `_populate_index` are examples of this.
- Existing functionality to merge everything within a bbox has been moved in `__merge_single_bbox`, where the biggest difference is that I keep track of a dataframe for all regex metadata instead of `lists` and `dicts`. The main reason is that it allows to group filepaths per band easily, as well of keeping track of which dates went into which merged raster. More about that later.
- Multi-bbox queries are stacked across a new dimension (t, c, h, w) by `__merge_query`. We need to agree if we want to go for timedimension 1 with non-temporal datasets.
- Apart from the tensor with timeseries imagery, I am returning sample['dates'] in datetime format too. This is a list of dates that went into every single bbox. So in the multi-bbox case this will be `[[bbox1_t1, bbox1_t2, ...], [bbox2_t1]]`. I chose datetime format for now since that worked well for me in practice, but this can be converted to any format by the transforms. I was thinking that instead of a list of dates, maybe we could return a daterange or something, but that mostly depends on the downstream use.
- The `filename_glob` has been relaxed, so that all files/bands end up in the dataset index.
- A new class variable `nodata_value` has been added, and `drop_nodata` has been added to the `init` of the class. Setting the value to True will ignore any merged raster that contains nodata values. This came from using the class in practice with Sentinel2, and seeing that some timestamps contained black (parts of) imagery, since some sentinel tiles are not square. In theory, this could become a separate PR, but I chose to add it here, because the effect of nodata pixels becomes more pronounced with timeseries.

What is still left to do:
- Implement method 3, query with tuple (or raise an error if trying to index a "single" dataset with a tuple).
- Implement/copy querying strategy to other datasets.
- Pass all pre-commit checks.

### DIFF
--- a/test.ipynb
+++ b/test.ipynb
@@ -95,21 +95,1079 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 5,
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import torch\n",
+    "import plotly.express as px\n",
+    "\n",
+    "def plot(\n",
+    "    sample: dict,\n",
+    "    indices_to_plot,\n",
+    "    show = False,\n",
+    "    **kwargs,\n",
+    "):\n",
+    "    \"\"\"Plots the image data from the given sample.\n",
+    "\n",
+    "    Args:\n",
+    "        sample (dict): A dictionary containing the image data returned by self.__get_item__. Should contain the key \"image\".\n",
+    "        indices_to_plot (list, optional): A list of indices to plot. If not provided, the method will use the RGB bands defined in `self.rgb_bands`.\n",
+    "        show (bool, optional): Whether to display the plot. Defaults to False.\n",
+    "        **kwargs (dict): Additional keyword arguments to be passed to `px.imshow`.\n",
+    "\n",
+    "    Returns:\n",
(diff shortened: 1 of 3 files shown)


sfalkena:
Closing this PR as this idea is not relevant towards current TS support

PR #2410, opened 2024-11-14
HySpecNet-11k: add new dataset

adamjstewart:
Adds a data loader for the [HySpecNet-11k](https://doi.org/10.5061/dryad.fttdz08zh) dataset.

- [x] Add dataset
- [x] Add data module
- [x] Add docs
- [x] Add tests

enmap

@ando-shah @nilsleh


ando-shah:
There could be a `split=None` mode which picks up images from all splits. This would be useful for pretraining to maximize sample size

adamjstewart:
Easiest way to implement that would probably be:
```python
ds = HySpecNet11k(split='train') + HySpecNet11k(split='val') + HySpecNet11k(split='test')
```

torchgeo/datasets/hyspecnet.py
```diff
@@ -0,0 +1,229 @@
...
+        transforms: Callable[[dict[str, Tensor]], dict[str, Tensor]] | None = None,
+        download: bool = False,
+        checksum: bool = False,
+    ) -> None:
+        """Initialize a new HySpecNet11k instance.
+
+        Args:
+            root: Root directory where dataset can be found.
+            split: One of 'train', 'val', or 'test'.
+            strategy: Either 'easy' for patchwise splitting or 'hard' for tilewise
```
nilsleh, line 120: What is the intuition between "easy" and hard" as the splitting strategy?

torchgeo/datasets/hyspecnet.py
```diff
@@ -0,0 +1,229 @@
...
+    166,
+]
+valid_channels_ids = [c + 1 for c in range(224) if c not in invalid_channels]
+
+
+class HySpecNet11k(NonGeoDataset):
+    """HySpecNet-11k dataset.
+
+    `HySpecNet-11k <https://doi.org/10.5061/dryad.fttdz08zh>`__ is a large-scale
+    hyperspectral benchmark dataset made up of 11,483 nonoverlapping image patches
```
nilsleh, line 53: From first inspection it's not clear what this dataset is a "benchmark" for. In the paper they look at compression I believe. Mabye we can state that it can be useful for pretraining (as done with the config task tests) or others. Or just include that it is made up of "unlabeled nonoverlapping image patches".

torchgeo/datasets/hyspecnet.py
adamjstewart: A great question. I'm just copying the naming used by the dataset authors. It doesn't make a lot of sense to me either.

nilsleh: Maybe we can switch that, and change the docs to say "this is referred to as 'easy' " in the original?

adamjstewart: It would just be extra code. Do you think the current documentation (both in `__init__` and in the class docstring) are insufficient?

torchgeo/datasets/hyspecnet.py
adamjstewart: Yeah, I also didn't know what to put in the CSV file for task. We don't currently have an SSL task in that table.

torchgeo/datasets/hyspecnet.py
nilsleh: It's well explained there, just seems unintuitive to call it "easy" and "hard" but fine to keep it then with the orig code.

Issue #2006, opened 2024-04-17
Change documentation theme


adamjstewart:
### Issue

We've been using [pytorch-sphinx-theme](https://github.com/pytorch/pytorch_sphinx_theme) since the start of the project, but it contains a number of known issues and is barely maintained. We should consider switching to a modern theme with fewer bugs and better support.

### Fix

I was originally planning on using [trojanzoo-sphinx-theme](https://github.com/ain-soph/trojanzoo_sphinx_theme) but it looks like this is also unmaintained now. I don't think I have the skills or expertise to maintain my own fork of either. We should look into popular themes and choose one that looks cool. Can people propose themes and link to examples on this issue?

adamjstewart:
PyTorch Lightning uses a fork of pytorch-sphinx-theme called [lightning-sphinx-theme](https://github.com/Lightning-Universe/lightning_sphinx_theme). However, this fork seems to have some of the same problems (no PyPI releases).

adamjstewart:
Sure! You should probably start by proposing a few themes that don't look too different from our current theme or are popular in ML. Once we narrow it down, you can open a few example PRs that change the theme and we can see what they look like.

Thakor-Yashpal:
Thanks for the suggestions! I'm leaning towards Furo (with careful customization) or Material for Sphinx. Furo's design is a good starting point, and I'm confident I can make it match the Py Torch theme's overall feel. I'll start working on customizing Furo, and if I run into any challenges, I'll get back to you with more advice.

adamjstewart:
https://github.com/pydata/pydata-sphinx-theme is pretty nice

Thakor-Yashpal:
> https://github.com/pydata/pydata-sphinx-theme is pretty nice

Thanks for sharing! The PyData Sphinx Theme looks pretty nice; I’ll try this out!

ain-soph:
Hi, this is the author from `trojanzoo_sphinx_theme`. Just saw this issue today, and I'm actually still maintaining the theme actively (I don't see any issue proposed in GitHub though).  
I tried compiled another internal project using the theme today with `Sphinx 8.2.3`, and it still works fine without any warning.

PyTorch has updated their theme to another looking, which I personally don't like the appearance. So I won't change the outlook to keep consistent. There are still some javascript/css bug fixes in my theme compared with pytorch official ones. I proposed those PRs, but no response from maintainers (4 years now)

adamjstewart:
[...]

PR #2414, opened 2024-11-15
issue #2006

Thakor-Yashpal:


adamjstewart:
I think you need to install furo in order for the docs to build. You can add furo to `requirements/docs.txt` and `pyproject.toml` and try again. Later we'll figure out the minimum version that actually works.

P.S. You don't need to create a new PR each time, you can just push more commits to the same PR.

Thakor-Yashpal (commented):
Is this correct or incorrect? Did I make any mistakes?

adamjstewart:
Looks like our sphinx version is too old, you'll need to update that too.

Thakor-Yashpal:
It's okay if it takes time to work on it; I'm working on it.

adamjstewart:
Let me know if you need any more help.

Thakor-Yashpal:
I'd like to give it another try.

PR #2416, opened 2024-11-18
Custom disaster-based train/test splits for xView2 dataset

burakekim:
cc: @calebrob6 

`XView2DistShift` is a subclass of `XView2` designed to modify the original train/test splits. Similar to EuroSATSpatial #2074, this class enables domain adaptation and out-of-distribution (OOD) detection experiments.

From the docstring: 
> This class allows for the selection of particular disasters to be used as the training set (in-domain) and test set (out-of-domain). The dataset can be split according to the disaster names specified by the user, enabling the model to train on one disaster type and evaluate on a different, out-of-domain disaster. The goal is to test the generalization ability of models trained on one disaster to perform on another.


adamjstewart:
We decided on EuroSAT _Spatial_ before, why switch to XView2 _Dist Shift_ now? Will there be any corresponding citations for these new splits?

It would be nice to move more of the shared code in the XView2 base class so that the only thing that needs to be changed in this subclass is the URLs. How different are these datasets?

burakekim:
> We decided on EuroSAT _Spatial_ before, why switch to XView2 _Dist Shift_ now?

EuroSAT split

calebrob6:
Great dataset. highly recommend

burakekim:
This is how it works: 

```root = "/data/dataood/xview2"

id_ood_disaster =  [{'disaster_name': 'hurricane-matthew', 'pre-post': 'post'},
  {'disaster_name': 'mexico-earthquake', 'pre-post': 'post'}]

xview2 = XView2DistShift(root=root, 
                         split="test",
                         id_ood_disaster=id_ood_disaster)

> ID sample len: 311, OOD sample len: 159
```

All the existing methods are revised to make XView2DistShift work as it should. I cannot seem to find a way to prune further (unless I upstream a method or two to `XView2` but that is too much of refactoring).

If it looks good, I can go ahead with unit tests. 

@adamjstewart, just to loop you in: As you may have noticed, we are upstreaming some modifications to existing datasets to make them suitable for assessing models under controlled domain shifts. This unlocks a whole new research dimension in TG that users can easily explore, including but not limited to robustness, generalization ability, anomaly detection, and novelty detection!

burakekim:
@adamjstewart is this good to go?

.gitignore
```diff
@@ -148,3 +148,4 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+xbdood.ipynb
```
adamjstewart, line 151: What is this?

torchgeo/datasets/xview.py
```diff
@@ -282,3 +282,195 @@ def plot(
...
+        'palu-tsunami',
+        'hurricane-florence',
+        'hurricane-michael',
+        'midwest-flooding',
+    ]
+
+    def __init__(
+        self,
+        root: str = 'data',
+        split: str = 'train',
```
adamjstewart, line 316: Not required, but I'm trying to convert more of our datasets to use proper type hints:
```suggestion
        split: Literal['train', 'test'] = 'train',
```
Where `Literal` is from `typing`. Not all str are valid, only these two.

torchgeo/datasets/xview.py
```diff
@@ -282,3 +282,195 @@ def plot(
...
+        id_ood_disaster: list[dict[str, str]] = [
+            {'disaster_name': 'hurricane-matthew', 'pre-post': 'post'},
+            {'disaster_name': 'mexico-earthquake', 'pre-post': 'post'},
+        ],
+        transforms: Callable[[dict[str, torch.Tensor]], dict[str, torch.Tensor]]
```
adamjstewart, line 321:
```suggestion
        transforms: Callable[[dict[str, Tensor]], dict[str, Tensor]]
```

torchgeo/datasets/xview.py
```diff
@@ -282,3 +282,195 @@ def plot(
...
+        'hurricane-florence',
+        'hurricane-michael',
+        'midwest-flooding',
+    ]
+
+    def __init__(
+        self,
+        root: str = 'data',
+        split: str = 'train',
+        id_ood_disaster: list[dict[str, str]] = [
```
adamjstewart, line 317: Do tuples or other Sequences also work?

torchgeo/datasets/xview.py
```diff
@@ -282,3 +282,195 @@ def plot(
...
+            root: Root directory where the dataset is located.
+            split: One of "train" or "test".
+            id_ood_disaster: List containing in-distribution and out-of-distribution disaster names.
+            transforms: a function/transform that takes input sample and its target as
+                entry and returns a transformed version
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+
+        Raises:
+            AssertionError: If *split* is invalid.
+            ValueError: If a disaster name in `id_ood_disaster` is not one of the valid disasters.
```
adamjstewart, line 337:
```suggestion
            ValueError: If a disaster name in *id_ood_disaster* is not one of the valid disasters.
```
Same formatting, but makes it more clear that it's in italics, not teletype.

torchgeo/datasets/xview.py
```diff
@@ -282,3 +282,195 @@ def plot(
...
+            self._load_split_files_by_disaster_and_type(
+                self.all_files, id_ood_disaster[0], id_ood_disaster[1]
+            )
+        )
+
+        train_size, test_size = self.get_id_ood_sizes()
+        print(f'ID sample len: {train_size}, OOD sample len: {test_size}')
+
+    def __getitem__(self, index: int) -> dict[str, torch.Tensor]:
+        """Get an item from the dataset at the given index."""
```
adamjstewart, line 371: Need to document Args and Returns

torchgeo/datasets/xview.py
```diff
@@ -282,3 +282,195 @@ def plot(
...
+        mask[(mask == 3) | (mask == 4)] = 0  # major/destroyed → background
+
+        sample = {'image': image, 'mask': mask}
+        if self.transforms:
+            sample = self.transforms(sample)
+
+        return sample
+
+    def __len__(self) -> int:
+        """Return the total number of samples in the dataset."""
```
adamjstewart, line 392: Need to document Returns

torchgeo/datasets/xview.py
```diff
@@ -282,3 +282,195 @@ def plot(
...
+        """Return the number of samples in the train and test splits."""
+        return (len(self.split_files['train']), len(self.split_files['test']))
+
+    def _initialize_files(self, root: str) -> list[dict[str, str]]:
+        """Initialize the dataset by loading file paths and computing basenames with sample numbers."""
+        all_files = []
+        for split in self.metadata.keys():
+            image_root = os.path.join(root, split, 'images')
+            mask_root = os.path.join(root, split, 'targets')
+            images = glob.glob(os.path.join(image_root, '*.png'))
```
adamjstewart, line 409: These should be sorted in order to ensure reproducibility

torchgeo/datasets/xview.py
```diff
@@ -282,3 +282,195 @@ def plot(
...
+                all_files.append(file_info)
+        return all_files
+
+    def _load_split_files_by_disaster_and_type(
+        self,
+        files: list[dict[str, str]],
+        id_disaster: dict[str, str],
+        ood_disaster: dict[str, str],
+    ) -> dict[str, list[dict[str, str]]]:
+        """Return the filepaths for the train (ID) and test (OOD) sets based on disaster name and pre-post disaster type.
```
adamjstewart, line 434: Some of these lines look longer than 88 chars

torchgeo/datasets/xview.py
```diff
@@ -282,3 +282,195 @@ def plot(
...
+
+        return sample
+
+    def __len__(self) -> int:
+        """Return the total number of samples in the dataset."""
+        return (
+            len(self.split_files['train'])
+            if self.split == 'train'
+            else len(self.split_files['test'])
+        )
```
adamjstewart, line 397: This is rather complicated. Do we need to store metadata for both train and test when only one is chosen?

torchgeo/datasets/xview.py
```diff
@@ -282,3 +282,195 @@ def plot(
...
+
+class XView2DistShift(XView2):
+    """A subclass of the XView2 dataset designed to reformat the original train/test splits.
+
+    This class allows for the selection of particular disasters to be used as the
+    training set (in-domain) and test set (out-of-domain). The dataset can be split
+    according to the disaster names specified by the user, enabling the model to train
+    on one disaster type and evaluate on a different, out-of-domain disaster. The goal
+    is to test the generalization ability of models trained on one disaster to perform
+    on others.
```
adamjstewart, line 295: Needs:
```rst
.. versionadded:: 0.8
```

adamjstewart (changes requested):

.gitignore
burakekim: That is for me to sanity check -- not here to stay

torchgeo/datasets/xview.py
burakekim: Liked it! Will use this

torchgeo/datasets/xview.py
burakekim: No. Currently we support lists only

burakekim:
Better now. Thanks for the review @adamjstewart!

torchgeo/datasets/xview.py
calebrob6: why not?
```
return len(self.split_files[self.split])
```

torchgeo/datasets/xview.py
```diff
@@ -282,3 +282,210 @@ def plot(
...
+    def __getitem__(self, index: int) -> dict[str, torch.Tensor]:
+        """Return an index within the dataset.
+
+        Args:
+            index: index to return
+
+        Returns:
+            data and label at that index
+        """
+        file_info = (
```
calebrob6, line 381: Same here -- does `self.split_files[self.split][index]` work?

torchgeo/datasets/xview.py
burakekim: This works! Thanks

torchgeo/datasets/xview.py
burakekim: Works -- thanks!

.gitignore
```diff
@@ -147,4 +147,4 @@ venv.bak/
 dmypy.json
 
 # Pyre type checker
-.pyre/
+.pyre/
```
adamjstewart, line 150: Can you undo changes to this file? You removed an EOL char.

torchgeo/datasets/xview.py
```diff
@@ -282,3 +282,202 @@ def plot(
...
+        'socal-fire',
+        'hurricane-matthew',
+        'mexico-earthquake',
+        'guatemala-volcano',
+        'santa-rosa-wildfire',
+        'palu-tsunami',
+        'hurricane-florence',
+        'hurricane-michael',
+        'midwest-flooding',
+    ]
```
adamjstewart, line 313:
```suggestion
    valid_disasters = (
        'hurricane-harvey',
        'socal-fire',
        'hurricane-matthew',
        'mexico-earthquake',
        'guatemala-volcano',
        'santa-rosa-wildfire',
        'palu-tsunami',
        'hurricane-florence',
        'hurricane-michael',
        'midwest-flooding',
    )
```

adamjstewart (changes requested):

torchgeo/datasets/xview.py
adamjstewart: I don't see anything in the code that suggests that a tuple wouldn't work. If both a tuple and list are valid, I would use `Sequence` for the type hints and tuple for the default value.

tests/datasets/test_xview.py
```diff
@@ -92,3 +92,94 @@ def test_plot(self, dataset: XView2) -> None:
         x['prediction'] = x['mask'][0].clone()
         dataset.plot(x)
         plt.close()
+
+
+class TestXView2DistShift:
+    @pytest.fixture(params=['train', 'test'])
+    def dataset(self, monkeypatch: MonkeyPatch, request: SubRequest) -> XView2DistShift:
+        monkeypatch.setattr(
```
adamjstewart, line 100: Might not even need to monkeypatch this if you remove `checksum=True`, I never bother with this

tests/datasets/test_xview.py
```diff
@@ -92,3 +92,94 @@ def test_plot(self, dataset: XView2) -> None:
...
+        root = os.path.join('tests', 'data', 'xview2')
+        split = request.param
+        transforms = nn.Identity()
+
+        return XView2DistShift(
+            root=root,
+            split=split,
+            id_ood_disaster=[
+                {'disaster_name': 'hurricane-harvey', 'pre-post': 'post'},
+                {'disaster_name': 'hurricane-harvey', 'pre-post': 'post'},
```
adamjstewart, line 125: Is this duplication required for some reason?

tests/datasets/test_xview.py
```diff
@@ -92,3 +92,94 @@ def test_plot(self, dataset: XView2) -> None:
...
+
+    def test_getitem(self, dataset: XView2DistShift) -> None:
+        x = dataset[0]
+        assert isinstance(x, dict)
+        assert isinstance(x['image'], torch.Tensor)
+        assert isinstance(x['mask'], torch.Tensor)
+        assert set(torch.unique(x['mask']).tolist()).issubset({0, 1})  # binary mask
+
+    def test_len(self, dataset: XView2DistShift) -> None:
+        assert len(dataset) > 0
```
adamjstewart, line 139: Let's test a specific length to ensure it behaves as expected. Can use an if-statement if it's different for train and test.

tests/datasets/test_xview.py
```diff
@@ -92,3 +92,94 @@ def test_plot(self, dataset: XView2) -> None:
...
+                    {'disaster_name': 'hurricane-harvey', 'pre-post': 'post'},
+                ],
+            )
+
+    def test_missing_disaster_name_key(self) -> None:
+        with pytest.raises(
+            ValueError, match="Each disaster entry must contain a 'disaster_name' key."
+        ):
+            XView2DistShift(
+                root='tests/data/xview2',
```
adamjstewart, line 156:
```suggestion
                root=os.path.join('tests', 'data', 'xview2'),
```
Windows

torchgeo/datasets/xview.py
```diff
@@ -282,3 +282,213 @@ def plot(
...
+        ),
+        transforms: Callable[[dict[str, Tensor]], dict[str, Tensor]] | None = None,
+        checksum: bool = False,
+    ) -> None:
+        """Initialize the XView2DistShift dataset instance.
+
+        Args:
+            root: Root directory where the dataset is located.
+            split: One of "train" or "test".
+            id_ood_disaster: List containing in-distribution and out-of-distribution disaster names.
```
adamjstewart, line 331: Needs more description. Clarify that both `disaster_name` and `pre-post` are required. Also explain what they mean, I have no idea

torchgeo/datasets/xview.py
```diff
@@ -282,3 +282,213 @@ def plot(
...
+            ValueError: If a disaster name in *id_ood_disaster* is not one of the valid disasters.
+            DatasetNotFoundError: If dataset is not found.
+        """
+        assert split in ['train', 'test'], "Split must be either 'train' or 'test'."
+        assert len(id_ood_disaster) == 2, (
+            'id_ood_disaster must contain exactly two items'
+        )
+
+        for disaster in id_ood_disaster:
+            if 'disaster_name' not in disaster:
```
adamjstewart, line 347: This check is duplicated below...

torchgeo/datasets/xview.py
```diff
@@ -282,3 +282,213 @@ def plot(
...
+            root: Root directory where the dataset is located.
+            split: One of "train" or "test".
+            id_ood_disaster: List containing in-distribution and out-of-distribution disaster names.
+            transforms: a function/transform that takes input sample and its target as
+                entry and returns a transformed version
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+
+        Raises:
+            AssertionError: If *split* is invalid.
+            ValueError: If a disaster name in *id_ood_disaster* is not one of the valid disasters.
```
adamjstewart, line 338: Seems like many other undocumented cases where these are raised

torchgeo/datasets/xview.py
```diff
@@ -282,3 +282,213 @@ def plot(
...
+
+        self._verify()
+
+        # Load all files and compute basenames and disasters only once
+        self.all_files = self._initialize_files(root)
+
+        # Split logic by disaster and pre-post type
+        self.split_files: dict[str, list[dict[str, str]]] = (
+            self._load_split_files_by_disaster_and_type(
+                self.all_files, id_ood_disaster[0], id_ood_disaster[1]
```
adamjstewart, line 376: If I pass in a dict of length 100, are the later 98 arguments ignored? Seems like a dict is a bad choice for this. Why not use 4 different parameters so they can be more clearly documented and type checked?

adamjstewart (changes requested):
Still a lot of concerns about how id/ood arguments are handled.

calebrob6:
@adamjstewart I think this one would be great to have and isn't a large lift. I'm going to take it over.

PR #2395, opened 2024-11-06
Update utils.py

lvxiaojie111:
fix the bug！
when running code:
dataset = naip & chesapeake  
from https://torchgeo.readthedocs.io/en/stable/tutorials/getting_started.html;  
it generates an err: 
ValueError: Bounding boxes BoundingBox(minx=440055.0, maxx=446085.0, miny=4295910.600000001, maxy=4303429.2, mint=1541260800.0, maxt=1541347199.999999) and BoundingBox(minx=402172.2959946942, maxx=532531.001762003, miny=4244077.01658443, maxy=4414067.523828775, mint=1356969600.0, maxt=1388505599.999999) do not overlap.

This new request  fix the err .

PR #2419, opened 2024-11-20
Test tutorials on all OSes

adamjstewart:
We are getting reports in #2395 that our tutorials do not work on Windows


adamjstewart:
Not necessary

PR #2422, opened 2024-11-21
Add ChangeDetectionTask

keves1:
This PR is to add a change detection trainer as mentioned in #2382.

Key points/items to discuss:
- I used the OSCD dataset to test with and modified this dataset to use a temporal dimension.
- With the added temporal dimension, Kornia’s `AugmentationSequential` doesn’t work, but can be combined with `VideoSequential` to support the temporal dimension (see Kornia [docs](https://kornia.readthedocs.io/en/latest/augmentation.container.html)). I overrode `self.aug` in the `OSCDDataModule` to do this but not sure if this should be incorporated into the `BaseDataModule` instead.
- `VideoSequential` adds a temporal dimension to the mask. Not sure if there is a way to avoid this, or if this is desirable, but I added an if statement to the `AugmentationSequential` wrapper to check for and remove this added dimension.
- The OSCDDataModule applies `_RandomNCrop` augmentation, but this does not work for time series data. I'm not sure how to modify `_RandomNCrop` to fix this and would appreciate some help/guidance.
- There are a few tests that I need to still make pass.

cc @robmarkcole


torchgeo/trainers/change.py
```diff
@@ -0,0 +1,303 @@
...
+)
+from torchmetrics.wrappers import ClasswiseWrapper
+from torchvision.models._api import WeightsEnum
+
+from ..models import FCSiamConc, FCSiamDiff, get_weight
+from . import utils
+from .base import BaseTask
+
+
+class FocalJaccardLoss(nn.Module):
```
robmarkcole, line 28: Should go in `torchgeo/losses/` ?

tests/conf/oscd.yaml
```diff
@@ -0,0 +1,17 @@
+model:
+  class_path: ChangeDetectionTask
+  init_args:
+    loss: 'ce'
+    model: 'unet'
+    backbone: 'resnet18'
+    in_channels: 13
```
robmarkcole, line 7: If stacking the channels should this be 2 * 13?

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,303 @@
...
+                model does not support pretrained weights. Pretrained ViT weight enums
+                are not supported yet.
+            in_channels: Number of input channels to model.
+            num_classes: Number of prediction classes.
+            class_weights: Optional rescaling weight given to each
+                class and used with 'ce' loss.
+            labels: Optional labels to use for classes in metrics
+                e.g. ["background", "change"]
+            loss: Name of the loss function, currently supports
+                'ce', 'jaccard', 'focal' or 'focal-jaccard' loss.
```
robmarkcole, line 95: and ce-jaccard ?

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,303 @@
...
+            mode="multiclass", normalized=True)
+        self.jaccard_loss = smp.losses.JaccardLoss(mode="multiclass")
+
+    def forward(self, preds: torch.Tensor, targets: torch.Tensor) -> torch.Tensor:
+        """Compute the loss."""
+        return cast(torch.Tensor, self.focal_loss(preds, targets) + self.jaccard_loss(preds, targets))
+
+
+class XEntJaccardLoss(nn.Module):
+    """XEntJaccardLoss."""
```
robmarkcole, line 44: Is there a paper ref for this?

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,303 @@
...
+
+    def configure_losses(self) -> None:
+        """Initialize the loss criterion.
+
+        Raises:
+            ValueError: If *loss* is invalid.
+        """
+        loss: str = self.hparams["loss"]
+        ignore_index = self.hparams["ignore_index"]
+        if loss == "ce":
```
robmarkcole, line 124: Noting that for change detection BCE may be more appropriate

https://github.com/microsoft/torchgeo/issues/245

robmarkcole:
I wonder if we should limit the scope to change between two timesteps and binary change - then we can use binary metrics and provide a template for the plot methods. I say this because this is the most common change detection task by a mile. Might also simplify the augmentations approach? Treating as a video sequence seems overkill.
Also I understand there is support for multitemporal coming later.

tests/conf/oscd.yaml
robmarkcole: OK and it is not necessary to config T?

adamjstewart:
> I wonder if we should limit the scope to change between two timesteps 

I'm personally okay with this, although @hfangcat has a recent work using multiple pre-event images that would be nice to support someday (could be a subclass if necessary).

> and binary change 

Again, this would probably be fine as a starting point, although I would someday like to make all trainers support binary/multiclass/multilabel, e.g., #2219.

> provide a template for the plot methods. 

Could also do this in the datasets (at least for benchmark NonGeoDatasets). We're also trying to remove explicit plotting in the trainers: #2184

> I say this because this is the most common change detection task by a mile. 

Agreed.

> Might also simplify the augmentations approach? Treating as a video sequence seems overkill.

I actually like the video augmentations, but let me loop in the Kornia folks to get their opinion: @edgarriba @johnnv1

> Also I understand there is support for multitemporal coming later.

Correct, see #2382 for the big picture (I think I also sent you a recording of my presented plan).

adamjstewart:
> VideoSequential adds a temporal dimension to the mask. Not sure if there is a way to avoid this

Can you try `keepdim=True`?

> I added an if statement to the AugmentationSequential wrapper to check for and remove this added dimension.

@ashnair1 would this work directly with `K.AugmentationSequential` now? We are trying to phase out our `AugmentationSequential` wrapper now that upstream supports (almost?) everything we need.

torchgeo/trainers/change.py
keves1: Yes, I missed that, thanks.

torchgeo/trainers/change.py
keves1: Not that I'm aware of, and I didn't find one from a quick search. This loss function was already in there with what I started from.

keves1:
I will go ahead and make changes for this to be for binary change and two timesteps, sounds like a good starting point.

> Can you try keepdim=True?

I tried this and it didn't get rid of the other dimension. I also looked into `extra_args` but didn't see any options to help with this.

> Could also do this in the datasets (at least for benchmark NonGeoDatasets). We're also trying to remove explicit plotting in the trainers

I was going to add plotting in the trainer, but would you rather not then? What would this look like in the dataset?

tests/conf/oscd.yaml
keves1: `in_channels` is multiplied by 2 in `configure_models` when initializing Unet.

torchgeo/trainers/change.py
keves1: That makes sense, I'll put this and `XEntJaccardLoss` there.

robmarkcole:
Perhaps there should even be a base class ChangeDetection and subclasses for BinaryChangeDetection etc?

adamjstewart:
That's exactly what I'm trying to _undo_ in #2219.

adamjstewart:
> I was going to add plotting in the trainer, but would you rather not then?

We can copy-n-paste the `validation_step` plotting stuff used by other trainers, but that's probably going to disappear soon (I think we're just waiting on testing in #2184.

> What would this look like in the dataset?

See `OSCD.plot()`

keves1:
I've updated this to now support only binary change with two timesteps.
To get `test_weight_file` in `test_change.py` to work with the two images stacked on the channel dimension for Unet, I modified the pytest fixture `model()` in `conftest.py` to use timm to create the model instead of torchvision, so that an `in_channels` parameter can be passed.

I still haven't been able to figure out how to make `transforms.transforms._RandomNCrop` work with the added temporal dimension. It seems to have something to do with `_NCropGenerator` not properly handling the temporal dimension but I really don't understand what is going on there.

tests/trainers/test_change.py
```diff
@@ -0,0 +1,169 @@
...
+    @pytest.fixture
+    def mocked_weights(
+        self,
+        tmp_path: Path,
+        monkeypatch: MonkeyPatch,
+        weights: WeightsEnum,
+        load_state_dict_from_url: None,
+    ) -> WeightsEnum:
+        path = tmp_path / f'{weights}.pth'
+        # multiply in_chans by 2 since images are concatenated
```
adamjstewart, line 83: How hard would it be to do late fusion, so pass each image through the encoder separately, then concatenate them, then pass them through the decoder? This would make it easier to use pre-trained models.

tests/trainers/test_change.py
```diff
@@ -0,0 +1,169 @@
...
+            weights.meta['model'], in_chans=weights.meta['in_chans'] * 2
+        )
+        torch.save(model.state_dict(), path)
+        try:
+            monkeypatch.setattr(weights.value, 'url', str(path))
+        except AttributeError:
+            monkeypatch.setattr(weights, 'url', str(path))
+        return weights
+
+    @pytest.mark.parametrize('model', [6], indirect=True)
```
adamjstewart, line 94: Remind me what `[6]` means here?

torchgeo/datamodules/oscd.py
```diff
@@ -86,9 +86,11 @@ def __init__(
...
 
         self.aug = AugmentationSequential(
-            K.Normalize(mean=self.mean, std=self.std),
-            _RandomNCrop(self.patch_size, batch_size),
-            data_keys=['image1', 'image2', 'mask'],
+            K.VideoSequential(
+                K.Normalize(mean=self.mean, std=self.std),
+                _RandomNCrop(self.patch_size, batch_size),
+            ),
+            data_keys=['image', 'mask'],
```
adamjstewart, line 93: Shouldn't this be `data_keys=None` so that Kornia automatically detects the key name from the dict keys?

torchgeo/datasets/oscd.py
```diff
@@ -240,7 +242,7 @@ def _load_target(self, path: Path) -> Tensor:
             array: np.typing.NDArray[np.int_] = np.array(img.convert('L'))
             tensor = torch.from_numpy(array)
             tensor = torch.clamp(tensor, min=0, max=1)
-            tensor = tensor.to(torch.long)
+            tensor = tensor.to(torch.float)
```
adamjstewart, line 245: Why would the target be a float?

torchgeo/losses/__init__.py
```diff
@@ -3,6 +3,8 @@
 
 """TorchGeo losses."""
 
+from .focaljaccard import BinaryFocalJaccardLoss
 from .qr import QRLoss, RQLoss
+from .xentjaccard import BinaryXEntJaccardLoss
 
-__all__ = ('QRLoss', 'RQLoss')
+__all__ = ('QRLoss', 'RQLoss', 'BinaryFocalJaccardLoss', 'BinaryXEntJaccardLoss')
```
adamjstewart, line 10: Never seen these losses before, are they standard in the literature? If not, I don't think we should add them.

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,233 @@
...
+            loss: Name of the loss function, currently supports
+                'bce', 'jaccard', 'focal', 'focal-jaccard' or 'bce-jaccard' loss.
+            lr: Learning rate for optimizer.
+            patience: Patience for learning rate scheduler.
+            freeze_backbone: Freeze the backbone network to fine-tune the
+                decoder and segmentation head.
+            freeze_decoder: Freeze the decoder network to linear probe
+                the segmentation head.
+
+        .. versionadded: 0.6
```
adamjstewart, line 65:
```suggestion
        .. versionadded: 0.7
```

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,233 @@
...
+        if self.hparams['freeze_backbone'] and model in ['unet']:
+            for param in self.model.encoder.parameters():
+                param.requires_grad = False
+
+        # Freeze decoder
+        if self.hparams['freeze_decoder'] and model in ['unet']:
+            for param in self.model.decoder.parameters():
+                param.requires_grad = False
+
+    def _shared_step(self, batch: Any, batch_idx: int, stage: str) -> Tensor:
```
adamjstewart, line 162: Needs a docstring

torchgeo/transforms/transforms.py
```diff
@@ -102,7 +102,13 @@ def forward(self, batch: dict[str, Any]) -> dict[str, Any]:
             batch['boxes'] = Boxes(batch['boxes']).to_tensor(mode='xyxy')
 
         # Torchmetrics does not support masks with a channel dimension
-        if 'mask' in batch and batch['mask'].shape[1] == 1:
+        # Kornia adds a temporal dimension to mask when passed through VideoSequential.
```
adamjstewart, line 105: We should no longer be using our `AugmentationSequential` wrapper, let's use the Kornia version upstream.

adamjstewart (commented):
Can you resolve the merge conflicts so we can run the tests?

tests/trainers/test_change.py
keves1: It's definitely possible, although I think we would need a custom Unet implementation in torchgeo/models to do this. It would simplify using the pretrained weights but is late fusion a common enough approach that many people would find this useful?

tests/trainers/test_change.py
keves1: Number of input channels (2 3-channel images stacked)

torchgeo/datamodules/oscd.py
keves1: Yes, I updated this when resolving the merge conflict.

torchgeo/datasets/oscd.py
keves1: The loss function `BCEWithLogitsLoss` expects the target to be a float.

torchgeo/trainers/change.py
keves1: I will add this.

torchgeo/transforms/transforms.py
keves1: Got it, I've now switched it to `K.AugmentationSequential`.

keves1:
I'm going to need some help figuring out how to get `transforms.transforms._RandomNCrop` to work with the added temporal dimension (this is used by the OSCD dataset). I've delved into this a few times but with my lack of familiarity with Kornia I haven't been able to track down the source of the issue. You can see the issue by running `tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer`.

Also, disregard my earlier comments about Kornia `VideoSequential` adding a dimension to the mask, this seems to have resolved with the latest Kornia version.

torchgeo/datasets/oscd.py
adamjstewart: Why do we have to use BCEWithLogitsLoss? Can we use BCELoss instead?

keves1: Both BCELoss and BCEWithLogitsLoss require float targets. Here's a brief explanation I found as to why: https://discuss.pytorch.org/t/inconsistency-between-loss-functions-input-types/138942. Is there any issue with the target being converted to a float here?

adamjstewart: Ah, I see. For our binary classification datasets, we convert the label to a float in `MultiLabelClassificationTask` instead of in the dataset. I would kind of like our datasets to be consistent (int for classification and float for regression). Let's change it in `ChangeDetectionTask` instead.

keves1: That makes sense, I'll make this change.

keves1:
I see that in the automated pytest checks (in tests / minimum) there was a syntax error, but I don't see this issue locally. How do I resolve this?

keves1:
I noticed I also need to update tests/datasets/test_oscd.py. And I think we are removing tests/datamodules/test_oscd.py per #978 now that we have a change detection task, right?

Also, how do you usually run prettier locally? I see there is an issue in the prettier check here but it isn't installed in the devcontainer unless I'm mistaken.

keves1:
> How hard would it be to do late fusion, so pass each image through the encoder separately, then concatenate them, then pass them through the decoder?

@adamjstewart I think this is a remaining question to resolve on this PR, and I realized that this is already what `torchgeo.models.FCSiamConc` is doing, right? And `FCSiamConc` is one of the model options for this trainer. So maybe it makes sense to also have an option that concatenates the images before passing through the model (keeping the Unet how it is).

adamjstewart:
Worth considering using https://github.com/Z-Zheng/pytorch-change-models

isaaccorley:
@keves1 Could I jump in and help get this over the finish line?

keves1:
@isaaccorley some help would be great, though I also want to do what I can. I see you made some changes to transforms, so that took care of the issue with _RandomNCrop? Is all that is left resolving the merge conflicts or do you see other changes needed too?

tests/trainers/test_change.py
isaaccorley: This is what `torchgeo.models.FCSiamConc` does already.

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,239 @@
...
+            loss: Name of the loss function, currently supports
+                'bce', 'jaccard', or 'focal' loss.
+            lr: Learning rate for optimizer.
+            patience: Patience for learning rate scheduler.
+            freeze_backbone: Freeze the backbone network to fine-tune the
+                decoder and segmentation head.
+            freeze_decoder: Freeze the decoder network to linear probe
+                the segmentation head.
+
+        .. versionadded: 0.7
```
isaaccorley, line 65: This should be 0.8 right @adamjstewart?

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,239 @@
...
+
+        if model == 'unet':
+            self.model = smp.Unet(
+                encoder_name=backbone,
+                encoder_weights='imagenet' if weights is True else None,
+                in_channels=in_channels * 2,  # images are concatenated
+                classes=num_classes,
+            )
+        elif model == 'fcsiamdiff':
+            self.model = FCSiamDiff(
```
isaaccorley, line 122: Need to pass the `encoder_name=backbone` here as well

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,239 @@
...
+                in_channels=in_channels,
+                classes=num_classes,
+                encoder_weights='imagenet' if weights is True else None,
+            )
+        elif model == 'fcsiamconc':
+            self.model = FCSiamConc(
```
isaaccorley, line 128: Need to pass the `encoder_name=backbone` here as well

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,239 @@
...
+            batch_idx: Integer displaying index of this batch._
+            stage: The current stage.
+
+        Returns:
+            The loss tensor.
+        """
+        model: str = self.hparams['model']
+        x = batch['image']
+        y = batch['mask']
+        y = y.unsqueeze(dim=1)  # channel dim for binary loss functions/metrics
```
isaaccorley, line 172:
```suggestion
        y = rearrange(y, "b h w -> b () h w")  # channel dim for binary loss functions/metrics
```

isaaccorley: Will need to import `from einops import rearrange` though

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,239 @@
...
+        if model == 'unet':
+            x = x.flatten(start_dim=1, end_dim=2)
```
isaaccorley, line 174: nit: This is a bit non-intuitive for me, I'd recommend using `einops.rearrange` here

```suggestion
            x = rearrange(x, "b t c h w -> b (t c) h w")
```

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,239 @@
...
+        y_hat = self(x)
+
+        loss: Tensor = self.criterion(y_hat, y.to(torch.float))
+        self.log(f'{stage}_loss', loss)
+
+        # Retrieve the correct metrics based on the stage
+        metrics = getattr(self, f'{stage}_metrics', None)
+        if metrics:
```
isaaccorley, line 182: Why would metrics be `None`?

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,239 @@
...
+        Args:
+            batch: The output of your DataLoader.
+            batch_idx: Integer displaying index of this batch.
+            dataloader_idx: Index of the current dataloader.
+
+        Returns:
+            Output predicted class.
+        """
+        model: str = self.hparams['model']
+        threshold = 0.5
```
isaaccorley, line 233: We should make this a configurable as an arg to `ChangeDetectionTask` that defaults to  0.5 instead of hardcoding.

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,239 @@
...
+        x = batch['image']
+        if model == 'unet':
+            x = x.flatten(start_dim=1, end_dim=2)
```
isaaccorley, line 236:
```suggestion
            x = rearrange(x, "b t c h w -> b (t c) h w")
```

torchgeo/trainers/__init__.py
```diff
@@ -14,8 +15,10 @@
 from .simclr import SimCLRTask
 
 __all__ = (
+    # Supervised
```
isaaccorley, line 18: BYOLTask and MoCoTask aren't supervised. Probably can just remove this comment

isaaccorley (changes requested):
Will need to rebase to get the new augmentation changes for OSCD

isaaccorley (changes requested):
Will need to rebase to get the new augmentation changes for OSCD

isaaccorley:
> @isaaccorley some help would be great, though I also want to do what I can. I see you made some changes to transforms, so that took care of the issue with _RandomNCrop? Is all that is left resolving the merge conflicts or do you see other changes needed too?

Left some review comments but lgtm once those are changed.

keves1:
@isaaccorley I rebased to bring in your augmentation changes. It looks like `_ExtractPatches` doesn't work with the added temporal dimension (the images have shape BTCHW) - you can see the issue by running the tests in `test_change.py`. Do you know how to fix that?

I'll work on your review comments.

isaaccorley:
I'll take a look and se what can be done about running with the temporal dim.

torchgeo/trainers/change.py
keves1: done

torchgeo/trainers/change.py
keves1: done

torchgeo/trainers/change.py
keves1: done

torchgeo/trainers/change.py
keves1: This is cleaner, thanks. Made the change.

torchgeo/trainers/change.py
keves1: metrics wouldn't be `None` the way the code is currently implemented, but isn't it good to check for good measure since `getattr` could return `None`?

torchgeo/trainers/change.py
keves1: done

torchgeo/trainers/change.py
keves1: done

torchgeo/trainers/__init__.py
keves1: done

keves1:
I'm trying to get test coverage for `predict_step` in `ChangeDetectionTask`, but it looks like this is not being tested because there isn't a predict dataset defined in the OSCD datamodule. Is it valid to just define `self.dataset = OSCD(split='test')` in `setup` of the OSCD datamodule?

isaaccorley:
Yes I think that should be sufficient

isaaccorley:
> @isaaccorley  I'm trying to get test coverage for `predict_step` in `ChangeDetectionTask`, but it looks like this is not being tested because there isn't a predict dataset defined in the OSCD datamodule. Is it valid to just define `self.dataset = OSCD(split='test')` in `setup` of the OSCD datamodule?

Yes I think that should be sufficient.

keves1:
@isaaccorley and @adamjstewart I think this is ready. Let me know what I should put in `ChangeDetectionTask` for `versionadded`.

isaaccorley (approved):
This lgtm. I believe it should be version 0.8 since this would be a new feature and that is the next milestone.

torchgeo/datamodules/oscd.py
```diff
@@ -105,6 +108,7 @@ def setup(self, stage: str) -> None:
         Args:
             stage: Either 'fit', 'validate', 'test', or 'predict'.
         """
+        self.dataset = OSCD(split='test', **self.kwargs)
```
adamjstewart, line 111: Why is this line needed?

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,255 @@
...
+        backbone: str = 'resnet50',
+        weights: WeightsEnum | str | bool | None = None,
+        in_channels: int = 3,
+        pos_weight: Tensor | None = None,
+        loss: str = 'bce',
+        lr: float = 1e-3,
+        patience: int = 10,
+        freeze_backbone: bool = False,
+        freeze_decoder: bool = False,
+        threshold: float = 0.5,
```
adamjstewart, line 42: Technically we don't expose the threshold in `SemanticSegmentationTask(task='binary')`, but this is probably a good idea

torchgeo/trainers/change.py
adamjstewart: Yes, this should be 0.8

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,255 @@
...
+    """Change Detection. Currently supports binary change between two timesteps."""
+
+    def __init__(
+        self,
+        model: str = 'unet',
+        backbone: str = 'resnet50',
+        weights: WeightsEnum | str | bool | None = None,
+        in_channels: int = 3,
+        pos_weight: Tensor | None = None,
+        loss: str = 'bce',
```
adamjstewart, line 37: Not all str are valid:
```suggestion
        loss: Literal['bce', 'jaccard', 'focal'] = 'bce',
```
where `Literal` is from `typing`

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,255 @@
...
+from . import utils
+from .base import BaseTask
+
+
+class ChangeDetectionTask(BaseTask):
+    """Change Detection. Currently supports binary change between two timesteps."""
+
+    def __init__(
+        self,
+        model: str = 'unet',
```
adamjstewart, line 32: Not all str are valid:
```suggestion
        model: Literal['unet', 'fcsiamdiff', 'fcsiamconc'] = 'unet',
```
where `Literal` is from `typing`

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,255 @@
...
+                "Currently, supports 'bce', 'jaccard', or 'focal' loss."
+            )
+
+    def configure_metrics(self) -> None:
+        """Initialize the performance metrics."""
+        metrics = MetricCollection(
+            {
+                'accuracy': BinaryAccuracy(),
+                'jaccard': BinaryJaccardIndex(),
+                'f1': BinaryF1Score(),
```
adamjstewart, line 100: Do we have any non-binary change detection datasets in TorchGeo? I imagine that some datasets exist with at least 3 classes (e.g., no change, deforestation, reforestation). It would be cool to support those too, identical to how we handled this in SemanticSegmentationTask which now supports binary, multiclass, and multilabel.

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,255 @@
...
+        super().__init__()
+
+    def configure_losses(self) -> None:
+        """Initialize the loss criterion.
+
+        Raises:
+            ValueError: If *loss* is invalid.
+        """
+        loss: str = self.hparams['loss']
+        if loss == 'bce':
```
adamjstewart, line 82: I would change this to a match-case statement, similar to what we did in Classification/SemanticSegmentationTask

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,255 @@
...
+            model: Name of the model to use.
+            backbone: Name of the `timm
+                <https://smp.readthedocs.io/en/latest/encoders_timm.html>`__ or `smp
+                <https://smp.readthedocs.io/en/latest/encoders.html>`__ backbone to use.
+            weights: Initial model weights. Either a weight enum, the string
+                representation of a weight enum, True for ImageNet weights, False or
+                None for random weights, or the path to a saved model state dict. FCN
+                model does not support pretrained weights. Pretrained ViT weight enums
+                are not supported yet.
+            in_channels: Number of input channels to model.
```
adamjstewart, line 56: Should we clarify that this is the number of channels per image, not the number of input channels to the model?

adamjstewart (changes requested):
SemanticSegmentationTask, on which this trainer is based, changed a lot in #2560, #2219, and #2690. We should update this PR with those same changes.

keves1:
> SemanticSegmentationTask, on which this trainer is based, changed a lot in #2560, #2219, and #2690. We should update this PR with those same changes.

I could add the denormalization for plotting, but as far as adding support for multiclass and multilabel, we specifically decided that this PR would just support binary change detection, so I'd like to keep it like that and someone else can extend it to those cases later.

torchgeo/datamodules/oscd.py
keves1: This is needed so that there is a dataset that can be used for testing `predict_step` in the trainer. Alternatively we could do:
```python
if stage in ['predict']:
    self.predict_dataset = OSCD(split='test', **self.kwargs)
```

torchgeo/trainers/change.py
keves1: Looking at how this is implemented in Classification/SemanticSegmentationTask, it looks like there is no checking for invalid strings, is that correct? Just wondering if I'm missing something or if that is intentional since that would lead to an obscure error.

adamjstewart:
@keves1 yes, let's add denormalizing and switch from if-statements to match-statements. 

I'll let @hkristen take a stab at non-binary change detection in a follow-up PR after this is merged. I know @hfangcat is also interested in change detection for 3+ images, I'm not sure if that would be a feature to add to ChangeDetectionTask or SemanticSegmentationTask. I would like to merge this PR soon as I think it's almost ready, we can iterate on additional features later.

torchgeo/datamodules/oscd.py
adamjstewart: I would rather monkeypatch during testing than create a fake predict dataset for OSCD.

torchgeo/trainers/change.py
keves1: Done.

torchgeo/trainers/change.py
keves1: I agree, I changed that.

keves1:
@adamjstewart I added denormalizing and switched to match statements, as well as made the other changes you suggested. I think it's ready now.

keves1:
Looks like the way I did the monkeypatch to test `predict_step` didn't work (it didn't run `predict_step`). What would be the right way to do this?

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,286 @@
...
+            patience: Patience for learning rate scheduler.
+            freeze_backbone: Freeze the backbone network to fine-tune the
+                decoder and segmentation head.
+            freeze_decoder: Freeze the decoder network to linear probe
+                the segmentation head.
+            threshold: Decision threshold used to convert predicted probabilities into binary
+                class labels. Predictions greater than or equal to the threshold are assigned
+                class 1, and those below are assigned class 0. Must be between 0 and 1.
+
+        .. versionadded: 0.8
```
adamjstewart, line 74: Can you move this from `__init__` to the class-level docstring?

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,286 @@
...
+        """
+        self.weights = weights
+        super().__init__()
+
+    def configure_losses(self) -> None:
+        """Initialize the loss criterion.
+
+        Raises:
+            ValueError: If *loss* is invalid.
```
adamjstewart, line 83: Can remove this

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,286 @@
...
+        )
+        self.train_metrics = metrics.clone(prefix='train_')
+        self.val_metrics = metrics.clone(prefix='val_')
+        self.test_metrics = metrics.clone(prefix='test_')
+
+    def configure_models(self) -> None:
+        """Initialize the model.
+
+        Raises:
+            ValueError: If *model* is invalid.
```
adamjstewart, line 112: Can remove this

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,286 @@
...
+        Returns:
+            Output predicted class.
+        """
+        model: str = self.hparams['model']
+        threshold: float = self.hparams['threshold']
+        x = batch['image']
+        if model == 'unet':
+            x = rearrange(x, 'b t c h w -> b (t c) h w')
+        y_hat: Tensor = self(x)
+        y_hat_hard = (nn.functional.sigmoid(y_hat) >= threshold).int()
```
adamjstewart, line 285: Can also do `y_hat.sigmoid()` like we did in `SemanticSegmentationTask`

adamjstewart: In other trainers, we return the soft prediction, not the thresholded output.

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,286 @@
...
+                and hasattr(self.logger.experiment, 'add_figure')
+            ):
+                datamodule = self.trainer.datamodule
+                aug = K.AugmentationSequential(
+                    K.VideoSequential(K.Denormalize(datamodule.mean, datamodule.std)),
+                    data_keys=None,
+                    keepdim=True,
+                )
+                batch = aug(batch)
+                batch['prediction'] = y_hat.argmax(dim=-1)
```
adamjstewart, line 215: Shouldn't this be `y_hat >= threshold`, not argmax? See `SemanticSegmentationTask`

adamjstewart (changes requested):

torchgeo/trainers/change.py
keves1: You are right that it should be using the threshold, not argmax. But in `SemanticSegmentationTask` (and here), shouldn't sigmoid be applied to `y_hat` before thresholding? So it would be `batch['prediction'] = (y_hat.sigmoid() >= 0.5).long()`.

tests/trainers/test_change.py
```diff
@@ -0,0 +1,235 @@
...
+        except MisconfigurationException:
+            pass
+        try:
+            main(['predict', *args])
+        except MisconfigurationException:
+            pass
+
+    def test_predict(self, fast_dev_run: bool) -> None:
+        datamodule = PredictChangeDetectionDataModule(
+            root='tests/data/oscd',
```
adamjstewart, line 87:
```suggestion
            root=os.path.join('tests', 'data', 'oscd'),
```
Windows, same below

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,274 @@
...
+from ..datasets import RGBBandsMissingError, unbind_samples
+from ..models import FCSiamConc, FCSiamDiff, get_weight
+from . import utils
+from .base import BaseTask
+
+
+class ChangeDetectionTask(BaseTask):
+    """Change Detection. Currently supports binary change between two timesteps.
+
+    .. versionadded: 0.8
```
adamjstewart, line 34:
```suggestion
    .. versionadded:: 0.8
```

torchgeo/trainers/change.py
```diff
@@ -0,0 +1,274 @@
...
+        Args:
+            model: Name of the model to use.
+            backbone: Name of the `timm
+                <https://smp.readthedocs.io/en/latest/encoders_timm.html>`__ or `smp
+                <https://smp.readthedocs.io/en/latest/encoders.html>`__ backbone to use.
+            weights: Initial model weights. Either a weight enum, the string
+                representation of a weight enum, True for ImageNet weights, False or
+                None for random weights, or the path to a saved model state dict. FCN
+                model does not support pretrained weights. Pretrained ViT weight enums
+                are not supported yet.
```
adamjstewart, line 61: @isaaccorley is this still true? I thought SMP now supports ViT backbones. Or is this something we need to change in our weight loader?

adamjstewart (changes requested):

torchgeo/trainers/change.py
adamjstewart: I think you are right, good catch! Want to submit a separate PR to fix ClassificationTask and SemanticSegmentationTask? Then we can backport it to 0.7.1.

adamjstewart:
Very close to being ready to merge, excited to finally get this in!!!

torchgeo/trainers/change.py
isaaccorley: This is not true anymore but will need to do some testing to make sure it works out of the box in our trainers.

adamjstewart: Cool, let's remove this sentence then. I'll open a separate PR to remove it from other trainers.

torchgeo/trainers/change.py
keves1: > Want to submit a separate PR to fix ClassificationTask and SemanticSegmentationTask?

I'll let someone else do that. But I've changed it here.

keves1:
@adamjstewart Let me know if this looks ready now!

adamjstewart (approved):
Thanks for the hard work on this, so happy to finally merge!

adamjstewart:
@hkristen now that this is merged, you can take a stab at adding multiclass/multilabel support if you want. See https://github.com/microsoft/torchgeo/pull/2219 for what this change looked like for SemanticSegmentationTask (code should be almost identical). Also see https://torchgeo.readthedocs.io/en/latest/user/contributing.html for general contributing guidelines. Happy to help with any questions you have or tests you encounter trouble with.

keves1:
> Thanks for the hard work on this, so happy to finally merge!

You're welcome, thanks for your help, and glad it could be merged!

adamjstewart:
@hkristen I just realized we actually do need multiclass change detection support for the BRIGHT and xView2 datasets. So this is slightly higher priority if you still have time to work on this. If not, let me know and I can try to find time.

hkristen:
@adamjstewart I can start working on it next week and should have a PR ready the week after. If you need it earlier, feel free to jump on it :=)

adamjstewart:
@hkristen have you made any progress on this?

hkristen:
Took a bit longer than expected but now the first version it is finally here #2890

torchgeo/trainers/change.py
adamjstewart: It may be None in `predict_step`

PR #2427, opened 2024-11-26
CI: skip PyVista tests on Windows due to segfault

adamjstewart:
@banesullivan we are seeing segfaults on Windows starting yesterday. The exact error can be seen here: https://github.com/microsoft/torchgeo/actions/runs/12012402908/job/33530984354

We did not see this error 3 days ago with the exact same versions of Python and PyVista: https://github.com/microsoft/torchgeo/actions/runs/11986395218/job/33419398537

We also don't see this issue on macOS or Linux.

We see a completely different issue for our minimum supported version of PyVista: https://github.com/microsoft/torchgeo/actions/runs/12012402908/job/33530981513

This last log is probably most useful for debugging to figure out exactly what changed.

### DIFF
--- a/requirements/min-reqs.old
+++ b/requirements/min-reqs.old
@@ -27,7 +27,6 @@ laspy==2.0.0
 opencv-python==4.5.4.58
 pycocotools==2.0.7
 pyarrow==15.0.0  # Remove when we upgrade min version of pandas to `pandas[parquet]>=2`
-pyvista==0.34.2
 scikit-image==0.19.0
 scipy==1.7.2
 


adamjstewart:
We may actually remove our pyvista dependency in #2428 but it would still be nice to get to the bottom of this. I'm curious what changed (maybe an apt-get update?) that caused the tests to fail. This would also be nice to solve in the 0.6.2 branch.

nilsleh (approved):

banesullivan:
Not sure what is causing that segfault, but I noticed you are installing VTK 9.4.0. See https://github.com/microsoft/torchgeo/actions/runs/12012402908/job/33530984354#step:6:164

We're going to issue a patch release to PyVista that will limit the VTK version later today. See https://github.com/pyvista/pyvista/pull/6872

adamjstewart:
Ah, looks like VTK 9.4 was released on PyPI just yesterday, that explains why we started having issues yesterday. Let's wait for the patch release instead of skipping tests. Thanks for the quick response!

banesullivan:
the segfault on widows with 9.4 may be related to https://github.com/pyvista/setup-headless-display-action/issues/5#issuecomment-2501583084

PR #2428, opened 2024-11-26
IDTReeS: remove support for plotting lidar point cloud

adamjstewart:
I don't think this feature (`IDTReeS.plot_las`) is widely used and it requires multiple dependencies (`pyvista`, `pyvista/setup-headless-display-action`) just to test. Visualization has always been orthogonal to TorchGeo's goals. We would like the bare minimum plotting capable in matplotlib. Users can always use PyVista themselves outside of TorchGeo.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -34,8 +34,6 @@ jobs:
           path: ${{ env.pythonLocation }}
           key: ${{ runner.os }}-${{ runner.arch }}-Python-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('requirements/required.txt', 'requirements/datasets.txt', 'requirements/tests.txt') }}
         if: ${{ runner.os != 'macOS' }}
-      - name: Setup headless display for pyvista
-        uses: pyvista/setup-headless-display-action@v3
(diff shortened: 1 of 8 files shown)


adamjstewart:
I wonder how people would feel about also removing our dependency on `laspy`, IDTReeS is the only dataset that uses it right now and none of our trainers know what to do with point clouds. Are we planning on adding more lidar datasets in the future?

adamjstewart:
We can also plot the 3D point cloud in matplotlib if this is a useful feature, PyVista is just overkill for our needs.

calebrob6:
I agree -- I'm wondering if we can make this a tutorial or put this snippet somewhere though so people can copy if they want?

adamjstewart:
> and would like to remove laspy

I'm actually now hesitating about this. I had one audience member at AGU who was particularly interested in lidar support. It's similar to SAR support, where a lot of domain expertise is required to do things correctly and we won't have that until someone contributes it, but I would love to expand our lidar support if we can.

> I'm wondering if we can make this a tutorial or put this snippet somewhere though so people can copy if they want?

I would suggest here: https://tutorial.pyvista.org/

PR #2432, opened 2024-11-27
EuroSAT: redistribute split files on Hugging Face

### LINKED ISSUE #2424: EuroSAT Dataset 403 Download Error
### Description

I'm trying to download the EuroSAT dataset, and I'm getting a 403 Forbidden HTTP error.

Image

### Steps to reproduce

1. Ensure you have a directory where the EuroSAT dataset is not already downloaded
2. Run the following line of code to download the EuroSAT dataset: `dataset = EuroSAT("path/to/data", download=True)` with `path/to/data` set to the empty directory from step 1
3. You should see the .zip file download, but then you'll get the following error before the .txt files download:

Image

### Version

0.6.1

adamjstewart:
Fixes #2424 

Thanks @isaiahlg for reporting the issue and @tameTNT for finding a copy of the files!

Tested using:
```python
from torchgeo.datasets import EuroSAT, EuroSATSpatial, EuroSAT100

EuroSAT('data/eurosat', download=True, checksum=True)
EuroSATSpatial('data/eurosatspatial', download=True, checksum=True)
EuroSAT100('data/eurosat100', download=True, checksum=True)
```


tests/datasets/test_eurosat.py
```diff
@@ -84,14 +60,14 @@ def test_add(self, dataset: EuroSAT) -> None:
         assert len(ds) == 4
 
     def test_already_downloaded(self, dataset: EuroSAT, tmp_path: Path) -> None:
-        EuroSAT(root=tmp_path, download=True)
+        type(dataset)(tmp_path)
```
adamjstewart, line 63: Testing got a bit ugly. We don't actually _have_ to test EuroSATSpatial or EuroSAT100, we could just test EuroSAT and get 100% coverage.

calebrob6 (approved):

Issue #2418, opened 2024-11-19
Add additional tutorials


adamjstewart:
### Issue

We are preparing for a TorchGeo tutorial at AGU and need to greatly expand our existing list of tutorials. This issue lists the tutorials that still need to be added and tracks progress towards completion.

General requirements:

* Accessible: tutorials should require no prior knowledge of ML or RS
* End-to-end: complete training and inference pipelines
* Well-tested: all tutorials must be tested in CI to ensure they remain up-to-date
* Resource-efficient: CI necessitates toy datasets that can be quickly downloaded
* Linter-approved: all notebooks should pass our ruff style checks

### Fix

The current plan is to completely rewrite all of our existing tutorials and organize them as follows:

- [x] Getting Started: general overview of DL/RS/TorchGeo, links to all other tutorial sections #2439
  - [x] Introduction to PyTorch: datasets, train-val-test splits, PyTorch, training, evaluation #2440
  - [x] Introduction to Geospatial Data: challenges of RS data, CRS, projections, resolution #2446
  - [x] Introduction to TorchGeo: types of datasets, purpose of samplers #2457
- [x] Basic Usage: targeted towards the ML crowd, more focused on training and evaluation #2439
  - [ ] Datasets: NonGeo/Geo/Raster/Vector/Intersection/Union, dataset splitters #2455
  - [ ] Samplers: GeoSampler, ROI, train vs evaluation samplers #2455
  - [x] Transforms: how to perform preprocessing, data augmentation, spectral indices, etc.
  - [x] Models: how to use models from timm, torchvision, and SMP, how to load pre-trained models, torch.hub, etc.
  - [x] Lightning: purpose of data modules and trainers, examples for classification, regression, etc.
  - [x] CLI: command-line interface and experimentation, reproducibility and best practices #2479
- [x] Case Studies: end-to-end workflows, targeted towards the RS crowd, more focused on inference #2439
  - [ ] Land cover mapping: for agriculture #2449
  - [ ] Change detection: for building damage assessment?
  - [ ] Instance segmentation: for field boundary detection? @burakekim 
  - [x] Hydrology: #960
  - [ ] Time series: after #2382 is complete
  - [ ] Any other ideas?
- [x] Customization: how to write your own datasets and contribute them back #2439
  - [x] NonGeoDataset #2451
  - [x] RasterDataset
  - [x] Data modules: Geo, NonGeo #2452
  - [ ] Transforms
  - [ ] Models: @nilsleh
[...]

PR #2440, opened 2024-12-02
Add Introduction to PyTorch tutorial

adamjstewart:
The first of many tutorials introduced for #2418 

This tutorial is designed for remote sensing folks with little to no background in deep learning or PyTorch. It provides a foundation upon which all later tutorials will build.

### DIFF
--- a/docs/tutorials/pytorch.ipynb
+++ b/docs/tutorials/pytorch.ipynb
@@ -1,0 +1,507 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "45973fd5-6259-4e03-9501-02ee96f3f870",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Copyright (c) Microsoft Corporation. All rights reserved.\n",
+    "# Licensed under the MIT License."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "9478ed9a",
+   "metadata": {
+    "id": "NdrXRgjU7Zih"
+   },
+   "source": [
+    "# Introduction to PyTorch\n",
+    "\n",
+    "_Written by: Adam J. Stewart_\n",
+    "\n",
+    "In this tutorial, we introduce the basics of deep learning with PyTorch. Understanding deep learning terminology and the training and evaluation pipeline in PyTorch is essential to using TorchGeo."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "34f10e9f",
+   "metadata": {
+    "id": "lCqHTGRYBZcz"
+   },
+   "source": [
+    "## Setup\n",
+    "\n",
+    "First, we install TorchGeo and all of its dependencies, including PyTorch."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "019092f0",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%pip install torchgeo"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "4db9f791",
+   "metadata": {
+    "id": "dV0NLHfGBMWl"
+   },
+   "source": [
+    "## Imports\n",
+    "\n",
+    "Next, we import PyTorch, TorchGeo, and any other libraries we need. We also manually set the random seed to ensure the reproducibility of our experiments."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "3d92b0f1",
+   "metadata": {
+    "id": "entire-albania"
+   },
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "import tempfile\n",
+    "\n",
+    "import kornia.augmentation as K\n",
+    "import torch\n",
+    "from torch import nn, optim\n",
+    "from torch.utils.data import DataLoader\n",
+    "\n",
+    "from torchgeo.datasets import EuroSAT100\n",
+    "from torchgeo.models import ResNet18_Weights, resnet18\n",
+    "\n",
+    "torch.manual_seed(0)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "9d13c2db-e5d4-4d83-846b-a2c32774bb44",
+   "metadata": {},
+   "source": [
+    "## Definitions\n",
+    "\n",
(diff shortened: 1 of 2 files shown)


docs/tutorials/pytorch.ipynb
```diff
@@ -0,0 +1,507 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
```
adamjstewart, line 4: Would like to start formatting the copyright in code cells, as tools like `ruff` can only check code cells: https://github.com/astral-sh/ruff/issues/12802

docs/tutorials/pytorch.ipynb
```diff
@@ -0,0 +1,507 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "9478ed9a",
+   "metadata": {
+    "id": "NdrXRgjU7Zih"
+   },
+   "source": [
+    "# Introduction to PyTorch\n",
+    "\n",
+    "_Written by: Adam J. Stewart_\n",
```
adamjstewart, line 23: Would like to start giving people credit for the tutorials they write, especially for the case study tutorials. This is intended to be the first person who wrote the tutorial, we wouldn't append other names except in the case of major revisions.

docs/tutorials/pytorch.ipynb
```diff
@@ -0,0 +1,507 @@
...
+    "In machine learning, we typically iterate over our datasets multiple times. Each full pass through the dataset is called an *epoch*. The following hyperparameter controls the number of epoch for which we train our model, and can be modified to train the model for longer:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "eb5dc7e8-6cb3-4457-83ad-7fa5aef8ea0c",
+   "metadata": {
+    "nbmake": {
+     "mock": {
```
adamjstewart, line 414: The tutorial runs for 100 epochs, but we mock it to only run for a single epoch during testing.

docs/tutorials/pytorch.ipynb
```diff
@@ -0,0 +1,507 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "7cd0bd25-e19a-4b26-94a1-fe9a544e8afd",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "evaluate(test_dataloader)"
```
adamjstewart, line 460: Final accuracies are 65% on validation and 85% on testing. Pretty good for only 100 images. We are nowhere near plateauing, and could run for longer to get higher accuracies, but I think this proves the point.

docs/tutorials/pytorch.ipynb
adamjstewart: This new tutorial only takes 9.69s to test.

calebrob6:
Would not put "self driving cars" as an application of DL

docs/tutorials/pytorch.ipynb
calebrob6: great idea

adamjstewart:
Why not?

docs/tutorials/pytorch.ipynb
```diff
@@ -0,0 +1,507 @@
...
+   "id": "7f26e4b8",
+   "metadata": {
+    "id": "5rLknZxrBEMz"
+   },
+   "source": [
+    "## Datasets\n",
+    "\n",
+    "In order to learn by example, we first need examples. In machine learning, we construct datasets of the form:\n",
+    "\n",
+    "$$\\mathcal{D} = \\left\\{\\left(x^{(i)}, y^{(i)}\\right)\\right\\}_{i=1}^N$$\n",
```
nilsleh, line 118: These math notations are barely visible when looking at the docs page

docs/tutorials/pytorch.ipynb
```diff
@@ -0,0 +1,507 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "f89e20ae-d3b6-4f05-a83f-f7034dd9862f",
+   "metadata": {},
+   "source": [
+    "In machine learning, we not only want to train a model, but also evaluate its performance on unseen data. Oftentimes, our dataset is split into three separate subsets:\n",
+    "\n",
+    "* train: for training the model *parameters*\n",
+    "* val: for validating the model *hyperparameters*\n",
+    "* test: for testing the model *performance*\n",
```
nilsleh, line 164: Validation is also to assess model performance right? Strictly speaking, testing data should only be run once at the very end to "deploy" the model, although almost everyone is indirectly tuning on test set performance as well.

docs/tutorials/pytorch.ipynb
```diff
@@ -0,0 +1,507 @@
...
+   "id": "a12b9e9b-26cc-43f4-a517-31b805862df5",
+   "metadata": {},
+   "source": [
+    "## Loss Function\n",
+    "\n",
+    "If $y$ is our expected output (also called \"ground truth\") and $\\hat{y}$ is our predicted output, our goal is to minimize the difference between $y$ and $\\hat{y}$. This difference is referred to as *error* or *loss*, and the loss function tells us how big of a mistake we made. For regression tasks, a simple mean squared error is sufficient:\n",
+    "\n",
+    "$$\\mathcal{L}(y, \\hat{y}) = \\left(y - \\hat{y}\\right)^2$$\n",
+    "\n",
+    "For classification tasks, such as EuroSAT, we instead use a negative log-likelihood:\n",
```
nilsleh, line 266: MSE is also log-likelihood of a Gaussian with unit variance.

docs/tutorials/pytorch.ipynb
adamjstewart: I don't think it looks too bad, the characters are of similar size to the normal text. During the tutorial, I'll likely zoom in so that everyone can see.

docs/tutorials/pytorch.ipynb
adamjstewart: Technically, you're only supposed to use validation for hyperparameter tuning (and early stopping). You do evaluate performance on val, but only to decide which set of hyperparams is best. I think I've used it as such in the rest of the tutorial.

It may be worth noting that sometimes test isn't public, but I didn't go into that much detail here.

docs/tutorials/pytorch.ipynb
adamjstewart: Interesting, didn't know that. I think it's easier to explain MSE as is, although it would be nice to tie into NLL/CE more smoothly.

PR #2449, opened 2024-12-05
Add Crop Type Mapping tutorial

burakekim:
Adding a new tutorial according to https://github.com/microsoft/torchgeo/issues/2418 

This tutorial demonstrates how to combine Sentinel-2 and <del>CDL</del>  EuroCrops datasets using the <del>Sentinel2CDLDataModule</del> `Sentinel2EuroCropsDataModule`. It covers training a semantic segmentation model, along with evaluation and inference steps.


nilsleh:
@burakekim Can you add the tutorial to the table of content in the rst file? Then one can view it in the CI docs to review as well :)

docs/index.rst
```diff
@@ -35,6 +35,16 @@ torchgeo
 
 .. toctree::
    :maxdepth: 1
+   :caption: Tutorials
```
adamjstewart, line 38: The files are no longer directly linked in `docs/index.rst`, you should instead update `docs/tutorials/case_studies.rst`

burakekim: after N commits (where N is the number of commits that is unfashionably high), I think I solved it

burakekim:
The initial draft that is somewhat end-to-end is now out:
- downloads a Sentinel-2 patch with rasterio's windowed reading
- prepares EuroCrops for Slovakia 
- visualizes the Sentinel-2 patch and its corresponding EuroCrops mask (with matplotlib and on a dynamic map)
- trains and loads a dummy model for qualitative and quantitative evaluation 

There are quite a few things I want to correct and improve:
- train on GPU for a reasonable number of epochs, with proper dataloader and Trainer hyperparameters
- maybe host the pretrained model + Sentinel-2 patch on HF? 
- use a bigger Sentinel-2 patch for training and possibly download another patch for inference OR use some sort of opportunistic sampling (can we do that with `GridSampler`?) for proper evaluation that tones down potential spatial autocorrelation
- EuroCrops has over 300 labels, but each country has its own distinct subset. The [number of classes Slovakia has](https://github.com/maja601/EuroCrops/wiki/Slovakia) is still high. Shall we just turn this into a binary crop classification? 
- addressing the question above comes down to what we want to do with the trained model, i.e., does it add value to have N classes where N is greater than 2? 
- there is a skimage dependency to visualize S2 with percentile normalization; and folium, pyproj, shapely for plotting the Sentinel-2 and EuroCrops bounds on a dynamic map -- or is it fine to download 3rd party libraries for individual case studies?

docs/tutorials/case_studies.rst
```diff
@@ -4,11 +4,12 @@ Case Studies
 The following case studies present end-to-end workflows for common use cases of geospatial machine learning:
 
 * `Earth Surface Water <earth_surface_water.ipynb>`_: A workflow for mapping surface water, including lakes and rivers
-
+* `Land Cover <land_cover.ipynb>` : An end-to-end workflow for mapping land cover using `VectorDataset` and Sentinel-2, includes data curation, training, evaluation, and inference
```
adamjstewart, line 7:
```suggestion
* `Land Cover <land_cover.ipynb>`_: An end-to-end workflow for mapping land cover using `VectorDataset` and Sentinel-2, includes data curation, training, evaluation, and inference
```

adamjstewart:
Still need to actually look at the code, but here are responses to your TODOs:

> train on GPU for a reasonable number of epochs, with proper dataloader and Trainer hyperparameters

Note that this needs to run in CI, preferably in seconds, not days. We can monkeypatch certain hyperparams to make this faster, but it shouldn't _require_ a GPU.

> maybe host the pretrained model + Sentinel-2 patch on HF?

Happy to do this if it makes the above faster while still getting good results.

> use a bigger Sentinel-2 patch for training and possibly download another patch for inference or use some sort of opportunistic sampling (can we do that with GridSampler?) for proper evaluation that tones down potential spatial autocorrelation

Avoid big data, this needs to run in CI where we have very limited storage, don't want to wait 10 min to download data during a tutorial. Not sure what you mean by opportunistic sampling, but there are various GeoDataset splitting methods that you can use to chop a tile into east/west splits, grids, etc.

> EuroCrops has over 300 labels, but each country has its own distinct subset. The number of classes Slovakia has is still high. Shall we just turn this into a binary crop classification? addressing the question above comes down to what we want to do with the trained model, i.e., does it add value to form multi-class classification?

I think both add value. Basically, we should have _some_ kind of binary semantic segmentation application, and _some_ kind of multiclass semantic segmentation application. They don't both have to be for agriculture though. For binary, something like building mapping may make more sense.

Also, tasks involving agriculture benefit greatly from time-series data. I'm planning on extending this tutorial for time series once we add support for it. So don't worry too much about the details right now, they will change in the future. This will also make the big data problem even worse, so keep the images small for now.

> there is a skimage dependency to visualize Sentinel-2 with percentile normalization; and folium, pyproj, shapely for plotting the Sentinel-2 and EuroCrops bounds on a dynamic map -- or is it fine to download 3rd party libraries for individual case studies?

Would prefer to avoid any additional dependencies if we can. Any reason we can't plot a static map with matplotlib? `eurocrops.plot(sample)` and `sentinel2.plot(sample)` should get you pretty far. If we do need to add additional deps, they need to be installed in `.github/workflows/tutorials.yaml` and `.github/workflows/release.yaml` like we did with `planetary_computer`. But I'm trying to get rid of those too, since they aren't absolutely necessary and aren't tracked by dependabot like our formal deps.

> P.S. In the next iteration, I am thinking of renaming the case study to Crop Type Classification. That would describe the task better

I agree with the rename. Both "Crop Classification" and "Crop Type Mapping" are common names. I think the latter may actually be even more common, and more technically correct. A computer vision person may argue that this is semantic segmentation, not classification. Of course, semantic segmentation is just pixelwise classification, so the distinction isn't too important.

burakekim:
Re: Using a dynamic map of overlaid mask and Sentinel-2 AOIs: I think it _just looks nice_, which is not a legit reason to keep it. I will discard it later

Re: Naming: Changed to `crop type mapping`, although I do not think the `mapping` wording perfectly fits the task -- not in the technical sense, but in the practical sense. We do not often say, "In this tutorial, we map crop types[...]" but rather, "In this tutorial, we classify crop types[...]" Still, I am fine with both

Re: CPU-friendly pipeline: With the current setup, each epoch takes ~8 minutes. I was targeting a training+test of 30mins on my laptop, which gives me a budget of 3 epochs. However, this way, the prediction does not look pretty at all for 10-class classification. We can do one or both: increase the tutorial time budget or lower the number of classes from 10 to, say, 3 (three most occuring classes in our AOI). What are your priorities? 

Re: Uploading the weights and Sentinel-2 patch on HF: This does not seem to offer a great deal of convenience unless we have a particular focus on allowing users to have the pretrained weights and Sentinel-2 patch and skip straight to inference --- which I do not think is the case because that bypasses the end-to-end showcase of TG abilities, which is the key feature of this tutorial

Re: Multi-class setting: I first listed all the crop type classes that fall under our AOI, got the 10 crop type classes with the highest occurrence, and set that as the `classes` argument

Thinking about it, we could consider a method for the users to choose the top-N crop types based on their occurrences for their AOIs. Otherwise, for larger AOIs (ours is fairly small and initially had 30 classes and for example, [239 for Estonia](https://github.com/maja601/EuroCrops/wiki/Estonia)), the number of crop types might make the task challenging for those with specific experimental needs. This would be another PR, though

adamjstewart:
> With the current setup, each epoch takes ~8 minutes. I was targeting a training+test of 30mins on my laptop, which gives me a budget of 3 epochs.

Change this from minutes to seconds and then this might get merged. Imagine presenting a tutorial in person and clicking "run" and telling people to just wait 30 min and it will be ready. It isn't hard to hack this to be fast in CI, but it should also be fast in person as well. This is where pre-trained models can help.

Exact number of crop types that is appropriate depends on the region, but 10 sounds okay to me.

burakekim:
> This is where pre-trained models can help.

The current setup includes pre-trained model and still takes 8mins/epoch. I can make the AOI much smaller but might make the prediction look less pretty and force us to use less classes. Do you have any tricks in mind to go from minutes to seconds?

adamjstewart:
How large is the current AOI? If we use a pre-trained decoder, it doesn't matter how small it is, the predictions will still be good.

For CI, we can use `fast_dev_run=True` like we do in the `pretrained_weights.ipynb` and `trainers.ipynb` tutorials. This only happens during testing, so it would still be slow during the actual tutorial presentation.

burakekim:
> How large is the current AOI?

It is 1640x1531 pixels. I use the RGB bands of Sentinel-2 with the `ResNet18_Weights.SENTINEL2_RGB_MOCO` weights, where both the encoder and decoder are frozen.

This is the setup I thought would be the fastest, but it is still slow, likely due to one or both of the following reasons:

- My laptop is old:  Each epoch takes 3 minutes to complete, but the next epoch starts only after 10 minutes. I/O operations like checkpoint saving might be consuming considerable wall-time. If you prefer, you can run the notebook on your machine and we see how long each epoch it takes

- I have not experimented much with the `patch_size` and `length` arguments. There might be a configuration that could improve speed, which I am unaware of

adamjstewart:
> It is 1640x1531 pixels.

Okay, that's not that big, we should be able to make this work.

> both the encoder and decoder are frozen.

I don't think we should freeze the decoder (unless we have a pretrained decoder), even if it makes things faster. I don't want to teach people something useless.

> My laptop is old

Can you try on Google Colab/Lightning Studios? Is that any better? That's the expected run environment, so as long as it is fast enough there, I'm happy.

burakekim:
> Can you try on Google Colab

https://colab.research.google.com/drive/119g5jC1WxxeTuYLHcCECpYENx6UpIgHF?usp=sharing

Yep, it was my laptop. On Colab it takes <1 minute each epoch. I am working on the hyperparameters to get an acceptable prediction

burakekim:
Limiting to 3 classes (2 crops + background) and using class weights did the trick. Now, the prediction is more meaningful. One thing I need to look into is whether to clip the part of the Sentinel-2 that was used for training when performing deployment

P.S., it is the same Colab link as above

adamjstewart:
Could modify https://medium.com/@sulemanhamdani10/supervised-wheat-classification-using-pytorchs-torchgeo-combining-satellite-imagery-and-python-fc7f95c82e to use time series.

robmarkcole:
@adamjstewart is this basically replaced by https://github.com/torchgeo/torchgeo/pull/3706 ?

PR #2450, opened 2024-12-05
Add MMFlood dataset

lccol:
This PR adds the MMFlood dataset from the paper ["MMFlood: A Multimodal Dataset for Flood Delineation From Satellite Imagery"](https://ieeexplore.ieee.org/abstract/document/9882096). This is a Sentinel-1 + DEM dataset for Image Segmentation.

Original tif files are of variable resolution. Max height in pixels is 2147, max width in pixels is 2313 (which are the ones reported in the docs). The dataset also includes hydrography information, but it is not available for all acquisitions (currently the implemented class does not read such tif files).

Example with False Color representation
immagine


lccol:

tests/datamodules/test_mmflood.py
```diff
@@ -0,0 +1,72 @@
+import os
```
nilsleh, line 1: For testing the datamodule, we usually add a config.yaml file in `tests/conf`and then test the datamodule directlry with one of the trainers. So the steps are:
- create a mlflood.yaml file with the specific parameters
- add this yaml file to the list of files to be tested in `tests/trainers/test_segmentation`

then this whole datamodule test file is should not be necessary

torchgeo/datamodules/mmflood.py
```diff
@@ -0,0 +1,101 @@
...
+            extra_args={
+                DataKey.MASK: {'resample': Resample.NEAREST, 'align_corners': None}
+            },
+        )
+
+        self.aug = K.AugmentationSequential(
+            K.Normalize(avg, self.std), keepdim=True, data_keys=None
+        )
+
+        return
```
nilsleh, line 78: don't need a return statement here

torchgeo/datasets/mmflood.py
```diff
@@ -0,0 +1,422 @@
...
+
+    Dataset classes:
+
+    0. no flood
+    1. flood
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://doi.org/10.1109/ACCESS.2022.3205419
+    """
```
nilsleh, line 46: versionadded tag missing here

torchgeo/datasets/mmflood.py
```diff
@@ -0,0 +1,422 @@
...
+        # Verify integrity of the dataset, initializing
+        # self.image_files, self.label_files, self.dem_files attributes
+        self._verify()
+        self.metadata_df = self._load_metadata()
+        self.folders = self._load_folders(check_folders=True)
+        paths = [x['s1_raw'] for x in self.folders]
+
+        # Build the index
+        super().__init__(paths=paths, crs=crs, transforms=transforms, cache=cache)
+        return
```
nilsleh, line 130: The return statements for functions that do not return anything can be removed

torchgeo/datasets/mmflood.py
```diff
@@ -0,0 +1,422 @@
...
+
+        res_folders = [
+            {'s1_raw': img_path, 'mask': mask_path, 'dem': dem_path}
+            for img_path, mask_path, dem_path in zip(image_files, mask_files, dem_files)
+        ]
+
+        if not check_folders:
+            return res_folders
+
+        # Verify image, dem and mask pairings
```
nilsleh, line 243: Could this verification instead be part of  `self._verify()` ? So first check the existence of the necessary files and then also compute this check?

torchgeo/datamodules/mmflood.py
```diff
@@ -0,0 +1,101 @@
...
+from kornia.constants import DataKey, Resample
+
+from ..datasets import MMFlood
+from ..samplers import GridGeoSampler, RandomBatchGeoSampler
+from ..samplers.utils import _to_tuple
+from .geo import GeoDataModule
+
+
+class MMFloodDataModule(GeoDataModule):
+    """LightningDataModule implementation for the MMFlood dataset."""
```
nilsleh, line 19: need a `versionadded` tag here

torchgeo/datasets/mmflood.py
```diff
@@ -0,0 +1,422 @@
...
+from .geo import RasterDataset
+from .utils import BoundingBox, Path, download_url, extract_archive
+
+
+class MMFlood(RasterDataset):
+    """MMFlood dataset.
+
+    `MMFlood <https://huggingface.co/datasets/links-ads/mmflood>`__ dataset is a multimodal flood delineation dataset.
+
+    Dataset features:
```
nilsleh, line 30: Do you know how many different tiles there are for this dataset?  I think this would be useful as an indication of the dataset size. Is it one tile per flood event?

nilsleh (changes requested):
Thanks so much for the contribution, this is great! I have made a first pass of comments below. If you have questions about anything feel free to comment.

torchgeo/datasets/mmflood.py
lccol: Thanks for all your comments! I'll fix all the other parts as soon as possible.

About this, I don't know the exact number of unique satellite tiles that were downloaded to create this dataset... The authors of this dataset may have merged different tiles to obtain the 1748 smaller tif files (mentioned as acquisitions in the docstring) that are present in the dataset. Maybe @edornd has more details about this

edornd: Hi @nilsleh, unfortunately we don't have the records of the individual tiles anymore since we built a mosaic on the spot when downloading the individual AoIs from Copernicus EMS. I confirm that there are 1748 individual AoI with varying sizes. Roughly, each AoI corresponds to a single S1 tile, since they're smaller, altough I remember that some overlapped. Each tile has an acquisition date tag though, so it is theoretically possible to backtrack the unique S1 tile ID.
When we trained for our benchmarks, we tiled these images into overlapping 512x512 chips obtaining 6182 chips for training, and 560 chips for validation (the test we tiled online at inference time), although these numbers are arbitrary and depend on implementation.

nilsleh: Hi @edornd and @lccol, thanks for the reply. That's fine! I think the 1748 number is information enough for documentation purposes. So I think I would just add a bullet point to the class docstring along the lines of "1748 tiles of varying pixel dimensions".

torchgeo/datamodules/mmflood.py
```diff
@@ -0,0 +1,101 @@
...
+class MMFloodDataModule(GeoDataModule):
+    """LightningDataModule implementation for the MMFlood dataset.
+
+    .. versionadded:: 0.7
+    """
+
+    # Computed over train set
+    mean = torch.tensor([0.1785585, 0.03574104, 168.45529])
+    median = torch.tensor([0.116051525, 0.025692634, 86.0])
+    std = torch.tensor([2.405442, 0.22719479, 242.74359])
```
adamjstewart, line 27: Do these work both `include_dem` True and False? Seems like the length should change

adamjstewart (commented):
Mostly looks good now, thanks for the hard work! 

Only other comment I would make is that the recommended approach for these "curated" geospatial datasets (`RasterDataset`s containing both images and masks) is to create a dummy dataset for the images, a dummy dataset for the masks, and an `IntersectionDataset` that combines them. This usually lets you completely skip the `__init__` and `__getitem__` since it will inherit from `RasterDataset`. See `L7Irish` and `L8Biome` for examples of these. Up to you whether or not you want to do this since you're almost done, but it could make the code a bit cleaner.

torchgeo/datasets/mmflood.py
```diff
@@ -0,0 +1,458 @@
...
+
+        Args:
+            index: indexes to return
+            query: (minx, maxx, miny, maxy, mint, maxt) coordinates to index
+
+        Returns:
+            the merged image
+        """
+        image = self._load_tif(index, modality='s1_raw', query=query).float()
+        if self.include_dem:
```
nilsleh, line 284: In the docstring you mentioned that the DEM is not available for all regions, but is this also explicitly handled? Or what happens if the DEM is not available. Do I get a tensor with a zero padded channel dimension? 

Or maybe as an additional check, if you have the full dataset downloaded, can you instantiate a datamodule or dataloader over the dataset an iterate through the length of the dataset without errors?

lccol: DEM is available for all the regions. Hydrography maps are missing for many of the tiles (738 missing). They are downloaded but not handled by this class. Maybe should I remove them from the docstring? Or introduce a flag `include_hydro` in the constructor and load only a portion of the dataset (1012 tiles out of 1748 total)?

nilsleh: Ah sorry, misread. If you think that hydrography should be useful to users, then I think a flag to `include_hydro` with default `False` would be a good thing. And then just add in the docstring like you mentioned here, that if someone wants to use hydro they will only have a subset of the data available.

lccol:
Thank you @adamjstewart for your comments. You managed all of your comments, including the conversion of `MMFlood` to `RasterDataset` class, similarly to `L7Irish` and `L8Biome`. I have just two questions:
 - I noticed that few entries have missing values, leading to NaNs. I was thinking of either putting all pixel values and mask values to 0 (option 1) or add a new entry `missing` (option 2) in the `dict` returned by the `__getitem__`. It will contain a Tensor of the same shape as the mask, with True in case of missing values in the image, False otherwise. I tried to check but I haven't found any other dataset which have NaN values in them. Probably this last option is kind of unusual compared to all the other datasets within the library... What do you think?
 - from some tests, I found that some of the tiles are partially overlapped and thus different tif files are merged. This should be fine I guess, since both images and masks are merged in a consistent manner (same order for the reverse painter algorithm). However, the dataset uses the tags in the tif files to store the timestamp of each tile. Is there a way to create the RTree with the temporal information stored in the tags? From my understanding, `RasterDataset` parses dates directly from the filename...

adamjstewart:
* I definitely prefer option 1. Our trainers support `ignore_index` which can be used to ignore these values during performance computation.
* You mean the timestamp isn't stored in the filename, it's stored in some kind of metadata? How do you access this metadata? It's possible to override `__init__` and extract the appropriate metadata yourself, it's just really ugly.

lccol:
> * I definitely prefer option 1. Our trainers support `ignore_index` which can be used to ignore these values during performance computation.

Ok. I should have implemented it using the `ignore_index`. 

> * You mean the timestamp isn't stored in the filename, it's stored in some kind of metadata? How do you access this metadata? It's possible to override `__init__` and extract the appropriate metadata yourself, it's just really ugly.

Yes, that is correct. To access this timestamp I should do as follows

```
with rasterio.open(path, “r”) as src:
        timestamp = sec.tags()[“event_date”]
```

I haven’t checked properly, but I think that should not be a big issue the current way it is implemented (i.e. do not retrieve the timestamp for each date), since I believe most of the overlapping tiles refers to the same date…

torchgeo/datasets/mmflood.py
```diff
@@ -0,0 +1,565 @@
...
+            # Check pairings of all files
+            self._verify_tif_pairings()
+            return
+        if not self.download:
+            raise DatasetNotFoundError(self)
+        self._download()
+        self._merge_tar_files()
+        self._extract()
+
+    def _verify_tif_pairings(self) -> None:
```
adamjstewart, line 448: We don't normally go to this level of detail in `_verify`, this could be removed if you want to keep the class simple

adamjstewart (commented):
More comments to try to make the code simpler

torchgeo/datasets/mmflood.py
```diff
@@ -0,0 +1,565 @@
...
+            assert len(images) == len(masks) == len(dems)
+            for filename in sorted(images.keys()):
+                image_files.append(images[filename])
+                mask_files.append(masks[filename])
+                dem_files.append(dems[filename])
+                hydro_files.append(hydros.get(filename, None))
+
+        paths['image'] = cast(list[str | None], image_files)
+        paths['mask'] = cast(list[str | None], mask_files)
+        paths['dem'] = cast(list[str | None], dem_files)
```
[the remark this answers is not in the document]
lccol: I removed this function entirely since it was used only by the _verify_tif_pairings (which I removed)

torchgeo/datasets/mmflood.py
```diff
@@ -0,0 +1,565 @@
...
+
+        Returns:
+            sample of image, mask and metadata at that index
+
+        Raises:
+            IndexError: if query is not found in the index
+        """
+        data = super().__getitem__(query)
+        missing_data = data['image'].isnan().any(dim=0)
+        # Set all pixel values of invalid areas to 0, all mask values to 255
```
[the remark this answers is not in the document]
lccol: Currently, 0 is used for no flood, 1 is used for flood in segmentation masks. Invalid pixels due to missing data are set to `_ignore_index` (currently 255). Should I set it to some other value, e.g., -1 such that the trainers ignore such values in performance and loss computation?

torchgeo/datasets/mmflood.py
```diff
@@ -0,0 +1,565 @@
...
+class MMFloodHydro(MMFloodComponent):
+    """Hydrography map component for MMFlood dataset."""
+
+    @property
+    def content(self) -> str:
+        """The subfolder containing hydrography data."""
+        return 'hydro'
+
+
+class MMFloodIntersection(IntersectionDataset):
```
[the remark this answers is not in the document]
lccol: I simplified the code and removed `MMFloodS1`, `MMFloodDEM` and just kept the base `MMFloodComponent` class and `MMFloodIntersection`. I also removed the code which forced the ordering between S1, DEM and Hydro.

`MMFloodIntersection` is needed due to the `_merge_dataset_indices` method, which forces the index to be the one of dataset 2 (hydro is always the last element and is the most restrictive one). I cannot use the `&` operator since some of the S1 acquisitions are overlapped, leading to a higher number of entries compared to the original dataset.

```
print(f's1: {len(s1)}')
print(f'dem: {len(dem)}')
print(f'hydro: {len(hydro)}')
print(f'mask: {len(mask)}')

s1: 1748
dem: 1748
hydro: 1012
mask: 1748

print(len(s1 & hydro))
3979
```

torchgeo/datasets/mmflood.py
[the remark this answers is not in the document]
adamjstewart: Gotcha, I think that's okay, the important thing is that the model only has to predict 2 classes, and those classes are sequential.

tests/datasets/test_mmflood.py
```diff
@@ -0,0 +1,112 @@
...
+
+class TestMMFlood:
+    @pytest.fixture(
+        params=product([True, False], [True, False], ['train', 'val', 'test'])
+    )
+    def dataset(
+        self, monkeypatch: MonkeyPatch, tmp_path: Path, request: SubRequest
+    ) -> MMFlood:
+        dataset_root = os.path.join('tests', 'data', 'mmflood/')
+        url = os.path.join(dataset_root)
```
adamjstewart, line 33:
```suggestion
        url = os.path.join('tests', 'data', 'mmflood') + os.sep
```
Use `os.sep` to make it platform-independent

adamjstewart (approved):
Looks good now, will give @nilsleh a chance to review

torchgeo/datasets/mmflood.py
```diff
@@ -0,0 +1,382 @@
...
+            include_hydro: If True, hydrography data is concatenated as last channel.
+                Only a smaller subset of the original dataset is loaded in this case.
+            transforms: a function/transform that takes input sample and its target as
+                entry and returns a transformed version
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+            cache: if True, cache file handle to speed up repeated sampling
+
+        Raises:
+            DatasetNotFoundError: If dataset is not found and *download* is False.
```
nilsleh, line 176: Also raises AssertionError if the split is invalid

torchgeo/datasets/mmflood.py
```diff
@@ -0,0 +1,382 @@
...
+        self.is_image = content != 'mask'
+        paths = []
+        for s in subfolders:
+            paths += glob(os.path.join(root, '**', f'{s}*-*', self.content, '*.tif'))
+        paths = sorted(paths)
+        super().__init__(paths, crs, res, transforms=transforms, cache=cache)
+
+
+class MMFloodIntersection(IntersectionDataset):
+    """Intersection dataset used to merge two or more MMFloodComponents."""
```
nilsleh, line 62: Do these helper classes also need a versionadded? @adamjstewart

adamjstewart: Not really, they won't appear in the documentation anyway.

nilsleh (approved):

adamjstewart (approved):
Thanks!

adamjstewart:
Very confused why the ETCI tests are failing, they aren't failing on main...

adamjstewart:
Looks like it's finally falling on main too. Will debug tomorrow night.

PR #2460, opened 2024-12-09
Error happened at variable image_files_names

decembboy:
there were some error at variable image_file_names so i fix them please check


adamjstewart:
Fixed by #2469

PR #2467, opened 2024-12-12
Docs: remove mathcal

adamjstewart:
`\mathcal{}` does not render correctly on Google Colab: https://github.com/googlecolab/colabtools/issues/3192

Let's just remove it so the tutorials are widely accessible.


calebrob6:
I slightly disagree as I think \mathcal makes these more readable and most places should render this correctly, but considering live tutorials will likely use Colab this change will make our tutorials look more polished. No strong preference either way.

calebrob6:
You can setup a ModelCheckpoint to save last then point it to `/tmp/eurosat100/lightning_logs/version_0/checkpoints/last.ckpt`

adamjstewart:
Yeah, I really wish we could fix Colab, but I don't think we have enough power to do that. These were challenging during the tutorial because they all display as square boxes, so no formatting is better than an empty box.

PR #2473, opened 2024-12-16
GlobalBuildingMap: add new dataset

adamjstewart:
Data isn't yet public, but will be public once the paper is accepted. Once it is public, we will likely rehost on HF to support automatic downloads.

gbm


calebrob6 (approved):

calebrob6:
Super excited for this one :)

adamjstewart:
I am too... We're going to try to resubmit this week. Fingers crossed that it gets accepted this time around so we can finally release the data.

I'll merge this for now and add download support in the next release.

PR #2475, opened 2024-12-17
Update with plot class for eddmaps,gbif,mm_earth,inaturalist,western_usa_live_fuel_mositure

preethatr07:


adamjstewart:
Can you open a separate PR for each dataset? Also, the `plot` function signature should match all of our other datasets:
```python
    def plot(
        self,
        sample: dict[str, Any],
        show_titles: bool = True,
        suptitle: str | None = None,
    ) -> Figure:

        Args:
            sample: a sample returned by :meth:`__getitem__`
            show_titles: flag indicating whether to show titles above each panel
            suptitle: optional string to use as a suptitle

        Returns:
            a matplotlib Figure with the rendered sample

        .. versionadded:: 0.7
        """
```
So we are plotting a single sample returned by `__getitem__`, we aren't plotting the entire dataset or creating a new sample in the plot function.

PR #2476, opened 2024-12-17
Data viusalisation project idea

preethatr07:
I have created a folder project-idea which contains the project plan document and its relevant files.


adamjstewart:
I don't think this is relevant, these files don't get installed when you `pip install torchgeo` so no one will ever use them.

PR #2477, opened 2024-12-18
Temporal transform: rearrange temporal and channel dimensions

yichiac:
This PR is to add a temporal transform: rearrange temporal and channel dimensions, as mentioned in https://github.com/microsoft/torchgeo/issues/2382.


adamjstewart:
Would be better to have a single PR for each transform instead of multiple in 1 PR.

yichiac:
Not sure if the class names are good. Tagging @adamjstewart for review.

adamjstewart (changes requested):
My vote would be for a single class that simply wraps around `einops.rearrange` like so:
```python
class Rearrange(GeometricAugmentationBase2D):
    def __init__(self, *args: Any, **kwargs: Any) -> None:
        super().__init__(p=1)
        self.flags = {'args': args, 'kwargs': kwargs}

    def apply_transform(
      self, input: Tensor, params: Dict[str, Tensor], flags: Dict[str, Any], transform: Optional[Tensor] = None
   ) -> Tensor:
        return rearrange(input, *flags['args'], **flags['kwargs'])
```
Note that `GeometricAugmentationBase2D` has additional functions that need to be overridden, see https://kornia.readthedocs.io/en/latest/augmentation.base.html#custom-augmentation-classes for details.

torchgeo/transforms/temporal.py
```diff
@@ -0,0 +1,60 @@
...
+
+from typing import Any
+
+from einops import rearrange
+from kornia.augmentation._3d.geometric.base import GeometricAugmentationBase3D
+from torch import Tensor
+
+
+class Rearrange(GeometricAugmentationBase3D):
+    """Rearrange tensor between [B, T, C, H, W] and [B, T*C, H, W]."""
```
adamjstewart, line 14:
```suggestion
    """Rearrange tensor dimensions.
        
    Examples:
        To insert a time dimension::
        
            Rearrange('b (t c) h w -> b t c h w', c=1)
           
        To collapse the time dimension::
        
            Rearrange('b t c h w -> b (t c) h w')
    """
```
This is both:

* more generic (we can rearrange any dimensions, not just time)
* more specific (here are examples of how to use it)

Let's see how the formatting looks.

torchgeo/transforms/temporal.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Initialize a Rearrange instance.
+
+        Args:
+            *args: Positional arguments for einops.rearrange
+            **kwargs: Keyword arguments for einops.rearrange
```
adamjstewart, line 21:
```suggestion
            *args: Positional arguments for :func:`einops.rearrange`.
            **kwargs: Keyword arguments for :func:`einops.rearrange`.
```
This should link to the einops docs directly. You will also need to add https://einops.rocks/ to `intersphinx_mapping` in `docs/conf.py`.

torchgeo/transforms/temporal.py
```diff
@@ -0,0 +1,60 @@
...
+
+        Args:
+            input: the input tensor
+            params: generated parameters
+            flags: static parameters
+
+        Returns:
+            the transformation
+        """
+        out = self.identity_matrix(input)
```
adamjstewart, line 59: Not 100% sure what is correct here. Ideally, we would support reversing this transform, although that becomes quite tricky in the general case.

adamjstewart (commented):
Mostly minor documentation changes, this is pretty close to being done now!

adamjstewart:
Can you add the import to `torchgeo/transforms/__init__.py`?

adamjstewart:
To fix the doc tests, you can add `kornia.augmentation._3d.geometric.base.GeometricAugmentationBase3D` to `nitpick_ignore` in `docs/conf.py`.

adamjstewart (approved):
This is ready to the best of my knowledge, but will give others a few days to review before merging on Monday.

torchgeo/transforms/temporal.py
isaaccorley: Probably better if someone explicitly just defines another `Rearrange` transform in their list of transforms. I've never personally used the inverse unless it's for denormalizing.

isaaccorley (approved):
LGTM. I will definitely use this transform.

adamjstewart:
Was just thinking about this. Does this transform actually work? Let's say I have an image (B x C x H x W) and a mask (B x H x W) and I want to insert a time dimension. Since the image and mask have a different number of dimensions, I'm not sure if I could do this with a single Rearrange aug.

isaaccorley:
> Was just thinking about this. Does this transform actually work? Let's say I have an image (B x C x H x W) and a mask (B x H x W) and I want to insert a time dimension. Since the image and mask have a different number of dimensions, I'm not sure if I could do this with a single Rearrange aug.

No they would have to have the same dimensions, although I think that we could have it just only apply it to the image and skip the mask if that makese sense

adamjstewart:
Yeah, I don't think we'll ever have matching dims for image and mask, and I'm not even sure what would happen for bbox, etc. We may need to rethink this design...

isaaccorley:
> Yeah, I don't think we'll ever have matching dims for image and mask, and I'm not even sure what would happen for bbox, etc. We may need to rethink this design...

I think be default you have to override certain augs for the bbox to be transformed otherwise it's just a no-op on the bbox. I remember having to do this for `_ExtractPatches` so let me look

adamjstewart:
Maybe we could accept a dict as input?
```python
{
    'image': 'b (t c) h w -> b t c h w',
    'mask': 'b h w -> b t h w',
}
```
Does the Kornia augmentation have access to which key it is currently being applied to?

isaaccorley:
Yes this makes sense to me. Or maybe having `image`, `mask`, args and mask defaults to `None` but I'm in favor of either.

PR #2479, opened 2024-12-19
Add TorchGeo CLI tutorial

adamjstewart:
Basically copied the docs from the README to a tutorial.

### DIFF
--- a/docs/tutorials/basic_usage.rst
+++ b/docs/tutorials/basic_usage.rst
@@ -7,6 +7,7 @@ The following tutorials introduce the basic concepts and components of TorchGeo:
 * `Indices <indices.ipynb>`_: Spectral indices
 * `Pretrained Weights <pretrained_weights.ipynb>`_: Models and pretrained weights
 * `Lightning Trainers <trainers.ipynb>`_: PyTorch Lightning data modules and trainers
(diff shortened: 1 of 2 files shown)


calebrob6:
`torchgeo validate` and `torchgeo test` won't use the trained model by default

adamjstewart:
Yeah, I'm not sure if there's an easy way to know the name of the trained model checkpoint to use. Maybe we can specify the name to use? I wish `--best` worked.

calebrob6:
You can setup a ModelCheckpoint to save last then point it to `/tmp/eurosat100/lightning_logs/version_0/checkpoints/last.ckpt`

adamjstewart:
Hmm, I'll look into it, but I'm hoping to not make the config.yaml too complicated.

calebrob6:
Then we need to be really clear that "torchgeo test" and "torchgeo validate" are just going to return results from a random initialized model (which is.... not helpful). I think a reader would expect "torchgeo test" to test the trained model as the tutorial is written currently. We should also show users how to _actually_ test their model too because that's going to be what people want to do with "test" or "validate" lol

adamjstewart:
Fixed

calebrob6 (approved):

PR #2480, opened 2024-12-19
Fixed the bandorder of EuroSAT

### LINKED ISSUE #2458: Wrong band order in EuroSAT dataset
### Description

Hello,

During the loading of the EuroSAT dataset, the band order with `[..., B08, B8A, B09, ...]` as described [here](https://github.com/microsoft/torchgeo/blob/main/torchgeo/datasets/eurosat.py#L78) is assumed. However, in the original EuroSAT repository, [this](https://github.com/phelber/EuroSAT/issues/7) issue was raised regarding the band order and accepted by the original authors, which describes the Bandorder as `[..., B08, B09, ..., B12, B8A]`. I was therefore wondering if this might be a mistake in `torchgeo` that assumes a different band order or if there is any additional information available that shows that the original repository's assumption is incorrect.

Thanks!

### Steps to reproduce

No code needed

### Version

0.6.1

lhackel-tub:
Fixes applied to dataset, datamodule and within the tutorial to align with [the issue raised in the EuroSAT git repository](https://github.com/phelber/EuroSAT/issues/7).

Fixes #2458

I did not check other Sentinel-2 datasets.


lhackel-tub:

adamjstewart (approved):
Looks great, thanks!

torchgeo/datamodules/eurosat.py
```diff
@@ -67,11 +67,11 @@
...
-    'B8A': 404.91978886,
-    'B09': 4.77584468,
-    'B10': 1002.58768311,
-    'B11': 761.30323499,
-    'B12': 1231.58581042,
+    'B09': 404.91978886,
+    'B10': 4.77584468,
+    'B11': 1002.58768311,
+    'B12': 761.30323499,
+    'B8A': 1231.58581042,
```
calebrob6, line 74: These don't need to be moved around, this is a dictionary

calebrob6:
I don't think this is correct because we calculate the mean and stdev in whatever order the channels are seen in the files

torchgeo/datamodules/eurosat.py
adamjstewart: True, but it's better to keep them in the same order as everything else.

adamjstewart:
> I don't think this is correct because we calculate the mean and stdev in whatever order the channels are seen in the files

Correct, and since we're changing the order of the band names in the files, we also need to change the order of the mean/std dev we apply.

PR #2499, opened 2025-01-03
EuroCrops: handle Nones in get_label

### LINKED ISSUE #2497: EuroCrops get_label cannot handle features with None attributes
### Description

The `get_label` method of the EuroCrops dataset is attempting to render features with `None` values, which raises an error. The attribute `EC_hcat_c` is set to `None` for some countries in the dataset, and this is not handled in the source code.

`hcat_code = feature['properties'][self.label_name] # Here, self.label_name is "EC_hcat_c"
`

```
TypeError                                 Traceback (most recent call last)
Cell In[7], [line 1](vscode-notebook-cell:?execution_count=7&line=1)
----> [1](vscode-notebook-cell:?execution_count=7&line=1) eurocrops_sk[small_bbox]

File ~\Desktop\torchgeo\torchgeo\datasets\geo.py:759, in VectorDataset.__getitem__(self, query)
    [753](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/burak/Desktop/torchgeo/docs/tutorials/~/Desktop/torchgeo/torchgeo/datasets/geo.py:753)         for feature in src.filter(bbox=(minx, miny, maxx, maxy)):
    [754](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/burak/Desktop/torchgeo/docs/tutorials/~/Desktop/torchgeo/torchgeo/datasets/geo.py:754)             # Warp geometries to requested CRS
    [755](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/burak/Desktop/torchgeo/docs/tutorials/~/Desktop/torchgeo/torchgeo/datasets/geo.py:755)             # print("feature['properties']", feature['properties'])
    [756](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/burak/Desktop/torchgeo/docs/tutorials/~/Desktop/torchgeo/torchgeo/datasets/geo.py:756)             shape = fiona.transform.transform_geom(
    [757](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/burak/Desktop/torchgeo/docs/tutorials/~/Desktop/torchgeo/torchgeo/datasets/geo.py:757)                 src.crs, self.crs.to_dict(), feature['geometry']
    [758](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/burak/Desktop/torchgeo/docs/tutorials/~/Desktop/torchgeo/torchgeo/datasets/geo.py:758)             )
--> [759](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/burak/Desktop/torchgeo/docs/tutorials/~/Desktop/torchgeo/torchgeo/datasets/geo.py:759)             label = self.get_label(feature)
    [760](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/burak/Desktop/torchgeo/docs/tutorials/~/Desktop/torchgeo/torchgeo/datasets/geo.py:760)             shapes.append((shape, label))
    [762](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/burak/Desktop/torchgeo/docs/tutorials/~/Desktop/torchgeo/torchgeo/datasets/geo.py:762) # Rasterize geometries

File ~\Desktop\torchgeo\torchgeo\datasets\eurocrops.py:215, in EuroCrops.get_label(self, feature)
    [213](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/burak/Desktop/torchgeo/docs/tutorials/~/Desktop/torchgeo/torchgeo/datasets/eurocrops.py:213)     return self.class_map[hcat_code]
    [214](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/burak/Desktop/torchgeo/docs/tutorials/~/Desktop/torchgeo/torchgeo/datasets/eurocrops.py:214) print("hcat_code", hcat_code)
--> [215](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/burak/Desktop/torchgeo/docs/tutorials/~/Desktop/torchgeo/torchgeo/datasets/eurocrops.py:215) hcat_code_list = list(hcat_code)
    [216](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/burak/Desktop/torchgeo/docs/tutorials/~/Desktop/torchgeo/torchgeo/datasets/eurocrops.py:216) if all(c == '0' for c in hcat_code_list):
    [217](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/burak/Desktop/torchgeo/docs/tutorials/~/Desktop/torchgeo/torchgeo/datasets/eurocrops.py:217)     break

TypeError: 'NoneType' object is not iterable
```

### Steps to reproduce

```
from torchgeo.datasets import EuroCrops

# Limit to a country with relatively small label size
# It needs to have None values inside, like Slovakia (scroll down): https://github.com/maja601/EuroCrops/wiki/Slovakia
class EuroCropsSK(EuroCrops):
    base_url = 'https://zenodo.org/records/8229128/files/'

    hcat_fname = 'HCAT2.csv'
    hcat_md5 = 'b323e8de3d8d507bd0550968925b6906'
    hcat_code_column = 'HCAT2_code'

    label_name = 'EC_hcat_c'

    filename_glob = '*_EC*.shp'

    filename_regex = r"""
        ^(?P<country>[A-Z]{2})
        (_(?P<region>[A-Z]+))?
        _
        (?P<date>\d{4})
        _
        (?P<suffix>EC(?:21)?)
        \.shp$
    """
    date_format = '%Y'
    zenodo_files: tuple[tuple[str, str], ...] = (
        ('SK_2021.zip', 'c7762b4073869673edc08502e7b22f01'),
    )


# Download the labels
eurocrops_sk = EuroCropsSK('./eurocrops_sk', download=True)

# Split the bounds into two and take the small one for fast debugging
small_bbox, big_bbox = eurocrops_sk.bounds.split(proportion=0.1, horizontal=True)

# Just curious
print(small_bbox.area)

# Trigger __getitem__
eurocrops_sk[small_bbox]
```

### Version

'0.7.0.dev0'

burakekim:
Closes #2497 

Now we return 0 if hcat_code is None, not rendering the feature.


adamjstewart:
Nice catch!

It looks like HCATv2 actually has a class for this: 
```csv
"not_known_and_other,3399000000"
```
I wonder if we should use this instead:
```python
hcat_code = feature['properties'][self.label_name] or '3399000000'
```

burakekim:
> It looks like HCATv2 actually has a class for this:

Under [Harmonisation with HCAT](https://github.com/maja601/EuroCrops?tab=readme-ov-file#hamonisation-with-hcat-), it seems `EC_hcat_c` is already the label itself, and that is the variable that is assigned as `None`. If I understand correctly, `EC_hcat_c` with `None` should have been discarded from the features since they are not assigned any value, including `not_known_and_other`.

I am in favor of removing the print statement that notifies the user each time we hit `None`, as it becomes very verbose during training.

favyen2:
> do you remember why you added 0? Just to appease mypy?

Do you mean for the HCAT code or for the label returned by the `get_label` function?

I don't think I ran into cases where HCAT code was None or 0 so it may be in a new version of the data or in a geographic area that I did not test.

For returning 0 from `get_label`, the function needs to return some integer for cases where the HCAT code is not in the list of codes that the user provides (i.e. the list they care about). Currently it returns the background class 0, which is also used for pixels that are not covered by any polygon. There may be cases where the user wants to use a different code for background class but for this use case I think it would work for them to add a catch-all class to the list of classes passed to the constructor since it is hierarchical (e.g. `classes=["1234", "5678", "0000"]`, now there is a separate "other" category that will be labeled as 3 after rasterization distinct from the background class 0).

adamjstewart:
Should we use 3399000000 instead of 0? I can also ask the EuroCrops folks if necessary.

favyen2:
As far as I know, `get_label` converts the HCAT code into a class integer suitable for training the segmentation model. So after the conversion these returned class indices should correspond to output neurons from the neural model rather than the original codes.

favyen2:
Oh, I guess you are saying that some users might still find it useful to train on the feature, mapping it to an "other" category rather than to 0 which matches the background category.

I'm not sure the answer to this, it may depend on why there are these weird features in the dataset that don't have a code. I guess the safest would be to mark them invalid but otherwise without digging more into the data I think both are reasonable solutions.

adamjstewart:
> Oh, I guess you are saying that some users might still find it useful to train on the feature, mapping it to an "other" category (which first setting the code to 3399000000 and then mapping it based on the user-provided category list would enable) rather than being limited to mapping it to 0 which matches the background category.

I'm actually saying the exact opposite. I'm wondering if the background class, None, and "not_known_and_other" should all be mapped to the same value. Otherwise, users will have to add all three to `ignore_index` during training.

burakekim:
I think it would be worth reaching out to EuroCrops people to ask whether `None` and `not_known_and_other` serve any specific purpose across the entire dataset. It is best to find out if there is any reason behind this design choice

favyen2:
> I'm actually saying the exact opposite. I'm wondering if the background class, None, and "not_known_and_other" should all be mapped to the same value. Otherwise, users will have to add all three to ignore_index during training.

I think there would be cases where users want to ignore None / not_known_and_other but not background, to try to have the model distinguish areas that are not crop fields.

burakekim:
> some of the None fields either represent duplicated field parcels

> These are duplicate field parcels with either the same or different crop labels, since most people only wanted one label per parcel

Does this mean some fields have multiple labels, and only one is used while the rest are mapped to None? If so, it is better not to map None to any other label.

> Regarding not_known_and_other these are predominately either artificial built-up areas or water bodies.

This feels more like background to me.

adamjstewart:
I guess I don't have strong opinions at the moment. We should make a decision here and get this merged though.

burakekim:
There does not seem to be `background` class in HCAT. I am guessing `not_known_and_other` is best treated as background, and I suggest we keep this as a class (not map to 0) to help model differentiate between what we are interested in and what we are not (assuming the task is formed from this perspective)

Shortly, this is what the current PR does: keep the `not_known_and_other` as -background- class and do not render `None`. I say we can merge this as is

burakekim:
On a side note:
HCAT here for convenience: [HCAT2.csv](https://github.com/user-attachments/files/18587301/HCAT2.csv). Here, there is another class with a rather-vague description: `unmaintained - 3308000000`

Currently, we treat this as a separate class. If this class is what I speculate it is (the source the annotation is received is no longer maintained?), it might be best to map to `None`. Once they release the new HCAT, it would be nice to sync with the EuroCrops folks to understand their decision making process

calebrob6 (approved):

calebrob6:
LGTM

adamjstewart (approved):
Alright, I'm fine with this. The user can always write a custom transform to combine classes if needed. Let's give them the maximum granularity and let them decide how to handle all of these missing/unknown/etc. classes.

PR #2508, opened 2025-01-09
Add DL4GAMAlps dataset

dcodrut:
This PR adds a a Multi-modal Dataset for Glacier Mapping (Segmentation) in the European Alps.

The dataset consists of Sentinel-2 images from 2015 (mainly), 2016 and 2017, and binary segmentation masks for glaciers, based on an inventory built by glaciology experts ([Paul et al. 2020](https://doi.org/10.1594/PANGAEA.909133)).
Given that glacier ice is not always visible in the images, due to seasonal snow, shadow/cloud cover and, most importantly, debris cover, the dataset also includes additional features that can help in the segmentation task.

Here's a sample extracted with the plot function implemented in the dataset:
image

A preprint is available [here](https://doi.org/10.22541/essoar.173557607.70204641/v1) that describes in detail how the dataset was constructed.
For a shorter description check also: https://huggingface.co/datasets/dcodrut/dl4gam_alps.


torchgeo/datasets/glacier_mapping_alps.py
```diff
@@ -0,0 +1,419 @@
...
+            split: one of "train", "val", or "test"
+            cv_iter: one of 1, 2, 3, 4, 5 (for the five-fold geographical cross-validation scheme)
+            version: one of "small" or "large" (controls the sampling overlap)
+            bands: the Sentinel-2 bands to use as input (default: RGB + NIR + SWIR)
+            extra_features: additional features to include (default: None; see the class attribute for the available)
+            transforms: a function/transform that takes input sample and its target as entry and returns a transformed
+             version
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+        """
```
nilsleh, line 150: Missing the "Raises" statement, something like:

```
Raises:
            AssertionError: If ``split`` or ``bands`` arguments are invalid.
            DependencyNotFoundError: If xarray is not installed.
```

torchgeo/datasets/glacier_mapping_alps.py
```diff
@@ -0,0 +1,419 @@
...
+        self.cv_iter = cv_iter
+        self.version = version
+        self.bands = bands
+        self.extra_features = extra_features
+        self.transforms = transforms
+        self.download = download
+        self.checksum = checksum
+
+        # sanity checks
+        assert split in (
```
nilsleh, line 162: you could move the valid options as class attributes above the init as something like `valid_split` etc, then the assert statements become a bit more concise.

torchgeo/datasets/glacier_mapping_alps.py
```diff
@@ -0,0 +1,419 @@
...
+        # get the corresponding urls and checksums
+        self.url_dataset = self.download_metadata[label]['url']
+        self.md5_dataset = self.download_metadata[label]['checksum']
+        self.url_csv_splits = self.download_metadata['splits_csv']['url']
+        self.md5_csv_splits = self.download_metadata['splits_csv']['checksum']
+
+        self._verify()
+        self._prepare_files()
+
+    def __len__(self) -> int:
```
nilsleh, line 199: The syntax for the return doc string should start with "Returns":

```
Returns:
    the number of patches in the dataset
```

for example.

torchgeo/datasets/glacier_mapping_alps.py
```diff
@@ -0,0 +1,419 @@
...
+        """The length of the dataset.
+
+        It returns the number of patches in the dataset.
+        """
+        return len(self.fp_patches)
+
+    def __getitem__(self, index: int) -> dict[str, Tensor]:
```
nilsleh, line 206: The doc string should also include "Args" and "Returns" explicitly, for the formatting check for example: https://github.com/microsoft/torchgeo/blob/5cca8e7336fc582ddc81a86e05d50b505ff74a07/torchgeo/datasets/caffe.py#L142

torchgeo/datasets/glacier_mapping_alps.py
```diff
@@ -0,0 +1,419 @@
...
+import os
+from collections.abc import Callable, Sequence
+from pathlib import Path
+from typing import ClassVar
+
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+import torch
+import xarray as xr
```
nilsleh, line 15: The unit tests are failing atm, because xarray is not a dependency of torchgeo (yet). Although that might change sooner or later (@adamjstewart) might have more insights yet. Until then, to get the unit tests to pass, `xarray` has to be handled as an additional dependency that the user has to install themself. For an idea of how this is done, you can check https://github.com/microsoft/torchgeo/blob/5cca8e7336fc582ddc81a86e05d50b505ff74a07/torchgeo/datasets/chabud.py#L50 for example. So you already include the note, but then need to remove the actual import statement in the file header towards a lazy import like here: https://github.com/microsoft/torchgeo/blob/5cca8e7336fc582ddc81a86e05d50b505ff74a07/torchgeo/datasets/chabud.py#L172

 This also needs needs to be handled in the pytest file like here: https://github.com/microsoft/torchgeo/blob/5cca8e7336fc582ddc81a86e05d50b505ff74a07/tests/datasets/test_chabud.py#L16 and added as a requirement in the file `torchgeo/requirements/datasets.txt`.

nilsleh (changes requested):
Thanks a lot for the contribution :) I hope this first round of review is able to resolve the `import xarray` error that is preventing the unit tests from running. Once they are working, will take another look for anything else!

torchgeo/datasets/glacier_mapping_alps.py
dcodrut: Thanks! Out of curiosity, shouldn't this be checked automatically by one of those linting tools? (I added the pre-commit hooks)

dcodrut:

torchgeo/datasets/glacier_mapping_alps.py
dcodrut: Now I understood how this is supposed to work, makes sense. Thanks! Fixed it now but it lead to a conflict in `datasets.txt`. Note that I added the versions I'm currently using. 

Xarray requires also netcdf4 to be installed to be able to work with netCDF files. There is also [h5netcdf](https://github.com/h5netcdf/h5netcdf) (see the second Tip [here](https://docs.xarray.dev/en/stable/user-guide/io.html#netcdf)). I don't have experience with this but perhaps for torchgeo it is relevant to check whether the later is indeed faster and in that case the dependency for my dataset could also be replaced.

dcodrut:
> Thanks a lot for the contribution :) I hope this first round of review is able to resolve the `import xarray` error that is preventing the unit tests from running. Once they are working, will take another look for anything else!

I hope it will be somehow useful. Happy to contribute and learn new things on the way :)
 Thanks for the comments and please let me know if there are additional things I should change.

torchgeo/datasets/glacier_mapping_alps.py
adamjstewart: As far as I know, there are no linting tools capable of checking this. It's a runtime error, not a type error. Ruff has some support for checking documentation of Args, but not Raises.

adamjstewart:
I'm completely fine with adding xarray and netcdf4 as optional dependencies. We've been talking about adding xarray support for a long time. We're still not sure if it will be directly through xarray or something like rioxarray, but that doesn't matter for this PR.

adamjstewart:
Can you resolve the merge conflicts so the tests run?

adamjstewart:
Can you resolve the merge conflicts?

torchgeo/datasets/glacier_mapping_alps.py
```diff
@@ -0,0 +1,427 @@
...
+from .geo import NonGeoDataset
+from .utils import (
+    download_and_extract_archive,
+    download_url,
+    extract_archive,
+    lazy_import,
+)
+
+
+class GlacierMappingAlps(NonGeoDataset):
```
adamjstewart, line 28: Should we rename the dataset to DL4GAM or DL4GAMAlps? Or is DL4GAM the project/model and GlacierMappingAlps is the name of the dataset? It's up to you, it's your dataset.

torchgeo/datasets/glacier_mapping_alps.py
```diff
@@ -0,0 +1,427 @@
...
+    r"""A Multi-modal Dataset for Glacier Mapping (Segmentation) in the European Alps.
+
+    The dataset consists of Sentinel-2 images from 2015 (mainly), 2016 and 2017, and binary segmentation masks for
```
adamjstewart, line 31: Would be nice if we could limit all lines to 88 chars, but if ruff doesn't complain it isn't technically required

torchgeo/datasets/glacier_mapping_alps.py
```diff
@@ -0,0 +1,427 @@
...
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://doi.org/10.22541/essoar.173557607.70204641/v1
+
+    .. note::
+
+        This dataset requires the following additional libraries to be installed:
+
+        * `xarray <https://docs.xarray.dev/en/stable/getting-started-guide/installing.html>`_
+        * `netcdf4 <https://unidata.github.io/netcdf4-python/>`_
```
adamjstewart, line 79: I usually just link to the PyPI homepage

torchgeo/datasets/glacier_mapping_alps.py
```diff
@@ -0,0 +1,427 @@
...
+        'dhdt',  # surface elevation change
+        'v',  # surface velocity
+    )
+    valid_splits = ('train', 'val', 'test')
+    valid_versions = ('small', 'large')
+    valid_cv_iters = (1, 2, 3, 4, 5)
+
+    def __init__(
+        self,
+        root: str | os.PathLike[str] = 'data',
```
adamjstewart, line 134: Let's use `.utils.Path` as the type hint for consistency with other datasets

torchgeo/datasets/glacier_mapping_alps.py
```diff
@@ -0,0 +1,427 @@
...
+
+        Args:
+            root: root directory where dataset can be found
+            split: one of "train", "val", or "test"
+            cv_iter: one of 1, 2, 3, 4, 5 (for the five-fold geographical cross-validation scheme)
+            version: one of "small" or "large" (controls the sampling overlap)
+            bands: the Sentinel-2 bands to use as input (default: RGB + NIR + SWIR)
+            extra_features: additional features to include (default: None; see the class attribute for the available)
+            transforms: a function/transform that takes input sample and its target as entry and returns a transformed
+             version
```
adamjstewart, line 154:
```suggestion
                version
```

torchgeo/datasets/glacier_mapping_alps.py
```diff
@@ -0,0 +1,427 @@
...
+        transforms: Callable[[dict[str, Tensor]], dict[str, Tensor]] | None = None,
+        download: bool = False,
+        checksum: bool = False,
+    ) -> None:
+        """Initialize the dataset.
+
+        Args:
+            root: root directory where dataset can be found
+            split: one of "train", "val", or "test"
+            cv_iter: one of 1, 2, 3, 4, 5 (for the five-fold geographical cross-validation scheme)
```
adamjstewart, line 149: I wonder if this should be in a datamodule instead of in the dataset. We don't see K-fold cross validation very often in deep learning, but I would love to be able to reproduce your paper

torchgeo/datasets/glacier_mapping_alps.py
```diff
@@ -0,0 +1,427 @@
...
+            version: one of "small" or "large" (controls the sampling overlap)
+            bands: the Sentinel-2 bands to use as input (default: RGB + NIR + SWIR)
+            extra_features: additional features to include (default: None; see the class attribute for the available)
+            transforms: a function/transform that takes input sample and its target as entry and returns a transformed
+             version
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+
+        Raises:
+            AssertionError: if the ``split``, ``cv_iter``, ``version``, ``bands`` or ``extra_features`` are invalid
```
adamjstewart, line 159: I usually use `*foo*` instead of backticks because that's what the Python docs uses. Could also be less specific and just say "if any parameters are invalid".

torchgeo/datasets/glacier_mapping_alps.py
```diff
@@ -0,0 +1,427 @@
...
+    def __len__(self) -> int:
+        """The length of the dataset.
+
+        Returns:
+            the number of patches in the dataset
+        """
+        return len(self.fp_patches)
+
+    def __getitem__(self, index: int) -> dict[str, Tensor]:
+        """It loads the netcdf file for the given index and returns the sample as a dict.
```
adamjstewart, line 213:
```suggestion
        """Load the NetCDF file for the given index and return the sample as a dict.
```

torchgeo/datasets/glacier_mapping_alps.py
```diff
@@ -0,0 +1,427 @@
...
+        Returns:
+            dict: a dictionary containing the sample with the following:
+
+                * the Sentinel-2 image (selected bands)
+                * the glacier mask (binary mask with all the glaciers in the current patch)
+                * the debris mask
+                * the cloud and shadow mask
+                * the additional features (DEM, derived features, etc.) if required
+        """
+        xr = lazy_import('xarray')
```
adamjstewart, line 227: For other datasets, we added a `lazy_import` to the `__init__` just so that the dataset would fail more quickly.

torchgeo/datasets/glacier_mapping_alps.py
```diff
@@ -0,0 +1,427 @@
...
+
+    def __init__(
+        self,
+        root: str | os.PathLike[str] = 'data',
+        split: str = 'train',
+        cv_iter: int = 1,
+        version: str = 'small',
+        bands: Sequence[str] = rgb_nir_swir_bands,
+        extra_features: Sequence[str] | None = None,
+        transforms: Callable[[dict[str, Tensor]], dict[str, Tensor]] | None = None,
```
adamjstewart, line 140: These aren't currently being used in `__getitem__`

adamjstewart (approved):
Code looks fantastic, just a few minor formatting comments. Thanks for taking the time to contribute your dataset to TorchGeo, let's hope it makes it easier for people to actually use and do cool science with!

requirements/datasets.txt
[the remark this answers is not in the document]
dcodrut: I found the minimum versions, manually. Do you use any tool for this?

torchgeo/datasets/glacier_mapping_alps.py
dcodrut: Was not sure when I named it. Thanks for the suggestion, I think the second (DL4GAMAlps) makes more sense, and it's also consistent with the HF dataset.

torchgeo/datasets/glacier_mapping_alps.py
dcodrut: It was like this initially but I then changed it because it was overlapping with the other `Path` that I use (from `pathlib`). Not sure how to handle this ...?

torchgeo/datasets/glacier_mapping_alps.py
dcodrut: I was also not sure how to handle this. I saw in the other datasets that the train-val-split is included in the dataset definition so I added the cross validation iteration also here which basically controls the train-val-split indices, with all the necessary information being read from the same [csv file](https://huggingface.co/datasets/dcodrut/dl4gam_alps/blob/main/data/map_all_splits_all_folds.csv).

In the paper there is also a table with the results from each cross-validation iteration. Therefore if you want to reproduce the results, you can just train 5 different models, evaluate them independently and compare the results. 

Since you said that you don't see k-fold cross-validation often in DL: I needed it for a different reason, not for validation reasons. I needed the model inferences on the entire dataset but this implies using the training samples also. To avoid this, I do a cross-validation split, to ensure that every sample in the dataset is used for testing exactly once, and at the end I collect the testing results.

torchgeo/datasets/glacier_mapping_alps.py
dcodrut: Thanks, good catch

requirements/datasets.txt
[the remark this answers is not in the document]
adamjstewart: Nah, I do it manually too, you can use binary search to speed it up though.

torchgeo/datasets/glacier_mapping_alps.py
adamjstewart: You can import pathlib instead of pathlib.Path.

torchgeo/datasets/glacier_mapping_alps.py
adamjstewart: Interesting. Yeah, if the CV split basically just specifies which specific train-test split to use and it's all documented in CSV files, let's keep it in the dataset.

requirements/datasets.txt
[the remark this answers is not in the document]
dcodrut: yep, that's what I did

torchgeo/datasets/dl4gam_alps.py
```diff
@@ -0,0 +1,450 @@
...
+                'Debris Mask',
+                'Clouds and Shadows Mask',
+            ]:
+                q_lim_clip = 0.025
+                img = np.clip(
+                    img, np.quantile(img, q_lim_clip), np.quantile(img, 1 - q_lim_clip)
+                )
+
+            vmin, vmax = np.min(img), np.max(img)
+            # smale the images to [0,1]
```
nilsleh, line 431: "scale" instead of "smale" typo?

nilsleh (approved):

torchgeo/datasets/dl4gam_alps.py
dcodrut: Fixed it but for some reason the tests are failing now. I merged the main branch, has to do with this I imagine. Could you please have a look?

adamjstewart: Not sure why the jobs were canceled (this usually happens when a commit is replaced by a newer commit), but I reran them now.

dcodrut: Thanks, looks fine now! 
Yeah, I noticed that but then shouldn't the new commit trigger again the entire test suite?

nilsleh (dismissed):

torchgeo/datasets/dl4gam_alps.py
adamjstewart: It should. You can also close and reopen a PR to retrigger the tests.

docs/api/datasets/non_geo_datasets.csv
```diff
@@ -21,6 +21,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `Forest Damage`_,OD,Drone imagery,"CDLA-Permissive-1.0","1,543",4,"1,500x1,500",,RGB
 `GeoNRW`_,S,Aerial,"CC-BY-4.0","7,783",11,"1,000x1,000",1,"RGB, DEM"
 `GID-15`_,S,Gaofen-2,-,150,15,"6,800x7,200",3,RGB
+`DL4GAM Alps`_,S,"Sentinel-2","CC-BY-4.0","2,251 or 11,440","2","256x256","10","MSI"
```
adamjstewart, line 24: Confirmed the license

pyproject.toml
```diff
@@ -87,6 +87,8 @@ datasets = [
     "h5py>=3.6",
     # laspy 2+ required for laspy.read
     "laspy>=2",
+    # netcdf4 1.5.4+ required for xarray.open_dataset with engine="netcdf4"
+    "netcdf4>=1.5.4",
```
adamjstewart, line 91:
```suggestion
    # netcdf4 1.5.8+ required for Python 3.10 wheels
    "netcdf4>=1.5.8",
```
Older versions might work, but are annoying to test in CI due to lack of wheels.

pyproject.toml
```diff
@@ -97,6 +99,8 @@ datasets = [
     "scikit-image>=0.19",
     # scipy 1.7.2+ required for Python 3.10 wheels
     "scipy>=1.7.2",
+    # xarray 2023.9+ required for xarray.open_dataset
```
adamjstewart, line 102: Pretty sure xarray.open_dataset existed before 2023, what error do you see for older versions?

requirements/datasets.txt
adamjstewart, line 1: Also need to update `requirements/min-reqs.old` with the minimum supported versions so we can test those.

tests/datasets/test_dl4gam_alps.py
```diff
@@ -0,0 +1,112 @@
...
+import pytest
+import torch
+import torch.nn as nn
+from _pytest.fixtures import SubRequest
+from pytest import MonkeyPatch
+
+from torchgeo.datasets import DatasetNotFoundError, DL4GAMAlps
+
+pytest.importorskip('xarray', minversion='2023.9')
+pytest.importorskip('netCDF4', minversion='1.5.4')
```
adamjstewart, line 17:
```suggestion
pytest.importorskip('netCDF4', minversion='1.5.8')
```

torchgeo/datasets/dl4gam_alps.py
```diff
@@ -0,0 +1,450 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""DL4GAMAlps Dataset."""
```
adamjstewart, line 4: Are there any plans to release DL4GAM datasets for other locations? If so, we could rename most of these files to `dl4gam.py` and then have region-specific datasets grouped together.

torchgeo/datasets/dl4gam_alps.py
```diff
@@ -0,0 +1,450 @@
...
+        'B6',
+        'B7',
+        'B8',
+        'B8A',
+        'B9',
+        'B10',
+        'B11',
+        'B12',
+    )
+    rgb_nir_swir_bands = ('B4', 'B3', 'B2', 'B8', 'B11')  # the subset used in our paper
```
adamjstewart, line 129:
```suggestion
    rgb_nir_swir_bands = ('B4', 'B3', 'B2', 'B8', 'B11')  # the subset used in the paper
```

torchgeo/datasets/dl4gam_alps.py
```diff
@@ -0,0 +1,450 @@
...
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://doi.org/10.22541/essoar.173557607.70204641/v1
+
+    .. note::
+
+        This dataset requires the following additional libraries to be installed:
+
+        * `xarray <https://pypi.org/project/xarray/>`_
+        * `netcdf4 <https://pypi.org/project/netCDF4/>`_
```
adamjstewart, line 89: Technically h5netcdf is also a valid engine

torchgeo/datasets/dl4gam_alps.py
```diff
@@ -0,0 +1,450 @@
...
+                # impute the missing values with the mean
+                # or zero (for dh/dt and surface velocity)
+                v_fill = 0.0 if feature in ('dhdt', 'v') else np.nanmean(vals)
+                vals[np.isnan(vals)] = v_fill
+
+                sample[feature] = vals
+
+        # convert to torch tensors
+        for k, v in sample.items():
+            sample[k] = torch.from_numpy(v)
```
adamjstewart, line 278: We try to avoid dynamic typing, could we convert to torch tensors earlier or have two different dicts?

torchgeo/datasets/dl4gam_alps.py
```diff
@@ -0,0 +1,450 @@
...
+            show_titles: flag indicating whether to show titles above each panel
+            suptitle: optional string to use as a suptitle
+            clip_extrema: flag indicating whether to clip the lowest/highest 2.5% of the
+                values for contrast enhancement
+
+        Returns:
+            a matplotlib Figure with the rendered sample
+        """
+        # we expect the RGB bands to be present
+        assert {'B4', 'B3', 'B2'}.issubset(set(self.bands)), (
```
adamjstewart, line 355: See `RGBBandsMissingError`. Also need to add `Raises:` to the docstring.

torchgeo/datasets/dl4gam_alps.py
dcodrut: Thanks for the suggestion. Yeah, that's the plan although at the moment the focus is on improving the current one (e.g.  by including other modalities).

pyproject.toml
dcodrut: That's correct. However when I tried the previous version (2023.8), the tests failed because of a numpy-related error. I assumed that the minimum version of xarray has to be in line with the other tochgeo packages but I'm actually not sure how the logic of these minimum versions works. 

This is the error I get with 2023.8:

```tests/datasets/test_dl4gam.py:None (tests/datasets/test_dl4gam.py)
tests/datasets/test_dl4gam.py:16: in <module>
    pytest.importorskip('xarray', minversion='2023.9')
../../../miniforge3/envs/torchgeo/lib/python3.12/site-packages/xarray/__init__.py:1: in <module>
    from xarray import testing, tutorial
../../../miniforge3/envs/torchgeo/lib/python3.12/site-packages/xarray/testing.py:10: in <module>
    from xarray.core import duck_array_ops, formatting, utils
../../../miniforge3/envs/torchgeo/lib/python3.12/site-packages/xarray/core/duck_array_ops.py:36: in <module>
    from xarray.core import dask_array_ops, dtypes, nputils
../../../miniforge3/envs/torchgeo/lib/python3.12/site-packages/xarray/core/dask_array_ops.py:3: in <module>
    from xarray.core import dtypes, nputils
../../../miniforge3/envs/torchgeo/lib/python3.12/site-packages/xarray/core/dtypes.py:43: in <module>
    (np.bytes_, np.unicode_),  # numpy promotes to unicode
../../../miniforge3/envs/torchgeo/lib/python3.12/site-packages/numpy/__init__.py:400: in __getattr__
    raise AttributeError(
E   AttributeError: `np.unicode_` was removed in the NumPy 2.0 release. Use `np.str_` instead.
ERROR conda.cli.main_run:execute(125): `conda run python /snap/pycharm-professional/443/plugins/python-ce/helpers/pycharm/_jb_pytest_runner.py --target tests/datasets/test_dl4gam.py::TestDL4GAMAlps` failed. (See above for error)

Process finished with exit code 4

```

requirements/datasets.txt
dcodrut: I added the (current) minimum versions

torchgeo/datasets/dl4gam_alps.py
dcodrut: That's true but is there a way to make only one of them a requirement?  Or should I just mention it in the note that h5netcdf is an alternative?

requirements/datasets.txt
dcodrut: Looks like the `mininum` test is failing now, need an earlier version of xarray. Could you please explain briefly how this works or otherwise help me with it? Should I create an environment and then install only what's in requirements/min-reqs.old and then look for an xarray version that passes my dataset's tests?

torchgeo/datasets/dl4gam_alps.py
adamjstewart: We have to choose one or the other to be listed as a requirement in `pyproject.toml`/`requirements.txt` but we can specify in the dataset docs that either one is fine.

pyproject.toml
adamjstewart: Hmm, try it in CI and see how low you can go. The "latest" tests use newer versions of Python and NumPy, but the "minimum" tests use the oldest supported versions of all deps. So you're much more likely to find something compatible in the older versions. You can also try `pip install -r requirements/min-reqs.old` if you want to test this locally instead of in CI.

requirements/datasets.txt
adamjstewart: Yes, you can do that locally, or you can try different versions in CI (a bit slower). I usually do some kind of binary search to find the minimum working version. It's kinda painful, but we don't usually have to do it too often, so I haven't found a better solution.

adamjstewart: I'm also happy to help debug this in CI for you if you get stuck.

dcodrut: Found a much older version. Should be fine now :)

pyproject.toml
```diff
@@ -97,6 +99,8 @@ datasets = [
     "scikit-image>=0.19",
     # scipy 1.7.2+ required for Python 3.10 wheels
     "scipy>=1.7.2",
+    # xarray 0.12.3+ required for xarray.open_dataset
```
adamjstewart, line 102: What was added to 0.12.3 that makes it work but not 0.12.2?

docs/api/datasets.rst
```diff
@@ -321,6 +321,10 @@ GID-15
 
 .. autoclass:: GID15
 
+DL4GAM Alps
+^^^^^^^^^^^
```
adamjstewart, line 325:
```suggestion
DL4GAM
^^^^^^
```

docs/api/datasets/non_geo_datasets.csv
```diff
@@ -22,6 +22,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
...
```
adamjstewart, line 25:
```suggestion
`DL4GAM`_,S,"Sentinel-2","CC-BY-4.0","2,251 or 11,440","2","256x256","10","MSI"
```

torchgeo/datasets/dl4gam.py
```diff
@@ -0,0 +1,449 @@
...
+        This dataset requires the following additional libraries to be installed:
+
+        * `xarray <https://pypi.org/project/xarray/>`_
+        * `netcdf4 <https://pypi.org/project/netCDF4/>`_
+           or `h5netcdf <https://pypi.org/project/h5netcdf/>`_
+
+    .. versionadded:: 0.7
+    """
+
+    r_url = (
```
adamjstewart, line 95: What does `r` mean? Can we just call this `url`? Also, you don't have to split it onto multiple lines, URLs are ignored by ruff.

adamjstewart (commented):
Only minor comments remaining

pyproject.toml
dcodrut: The problem with 0.12.2 is caused by an import (`pandas.Panel`, removed starting from 0.25.0). Note that I kept the other minimum versions from min-reqs the same, including `pandas==1.3.3`. 

```
tests/datasets/test_dl4gam.py:16: in <module>
    pytest.importorskip('xarray', minversion='0.12.3')
../../../anaconda3/envs/torchgeolow2/lib/python3.10/site-packages/xarray/__init__.py:11: in <module>
    from .core.combine import combine_by_coords, combine_nested, auto_combine
../../../anaconda3/envs/torchgeolow2/lib/python3.10/site-packages/xarray/core/combine.py:8: in <module>
    from .dataarray import DataArray
../../../anaconda3/envs/torchgeolow2/lib/python3.10/site-packages/xarray/core/dataarray.py:12: in <module>
    from . import (
../../../anaconda3/envs/torchgeolow2/lib/python3.10/site-packages/xarray/core/computation.py:17: in <module>
    from .merge import expand_and_merge_variables
../../../anaconda3/envs/torchgeolow2/lib/python3.10/site-packages/xarray/core/merge.py:18: in <module>
    PANDAS_TYPES = (pd.Series, pd.DataFrame, pd.Panel)
../../../anaconda3/envs/torchgeolow2/lib/python3.10/site-packages/pandas/__init__.py:244: in __getattr__
    raise AttributeError(f"module 'pandas' has no attribute '{name}'")
E   AttributeError: module 'pandas' has no attribute 'Panel'
```

torchgeo/datasets/dl4gam.py
dcodrut: just bad naming; ok, didn't know that, thx

pyproject.toml
adamjstewart:
```suggestion
    # xarray 0.12.3+ required for pandas 1.3.3 support
```

docs/api/datasets.rst
```diff
@@ -326,6 +326,10 @@ GID-15
 
 .. autoclass:: GID15
 
+DL4GAM
```
adamjstewart, line 329: Can we sort this alphabetically?

docs/api/datasets/non_geo_datasets.csv
```diff
@@ -22,6 +22,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `Forest Damage`_,OD,Drone imagery,"CDLA-Permissive-1.0","1,543",4,"1,500x1,500",,RGB
 `GeoNRW`_,S,Aerial,"CC-BY-4.0","7,783",11,"1,000x1,000",1,"RGB, DEM"
 `GID-15`_,S,Gaofen-2,-,150,15,"6,800x7,200",3,RGB
+`DL4GAM`_,S,"Sentinel-2","CC-BY-4.0","2,251 or 11,440","2","256x256","10","MSI"
```
adamjstewart, line 25: Same here

tests/datasets/test_dl4gam.py
```diff
@@ -0,0 +1,125 @@
...
+            [1, 3, 5],
+            ['small', 'small', 'large'],
+            [DL4GAMAlps.rgb_bands, DL4GAMAlps.rgb_nir_swir_bands, DL4GAMAlps.all_bands],
+            [None, ['dem'], DL4GAMAlps.valid_extra_features],
+        )
+    )
+    def dataset(
+        self, monkeypatch: MonkeyPatch, tmp_path: Path, request: SubRequest
+    ) -> DL4GAMAlps:
+        r_url = Path('tests', 'data', 'dl4gam_alps')
```
adamjstewart, line 33: Let's use `url` here too

torchgeo/datasets/dl4gam.py
```diff
@@ -0,0 +1,446 @@
...
+
+    * https://doi.org/10.22541/essoar.173557607.70204641/v1
+
+    .. note::
+
+        This dataset requires the following additional libraries to be installed:
+
+        * `xarray <https://pypi.org/project/xarray/>`_
+        * `netcdf4 <https://pypi.org/project/netCDF4/>`_
+           or `h5netcdf <https://pypi.org/project/h5netcdf/>`_
```
adamjstewart, line 90:
```suggestion
          or `h5netcdf <https://pypi.org/project/h5netcdf/>`_
```
Sphinx is really sensitive to indentation: https://torchgeo--2508.org.readthedocs.build/en/2508/api/datasets.html#dl4gam

adamjstewart (approved):
Thanks for the contribution!

dcodrut:
Thank you too for helping me with it!

Issue #1088, opened 2023-02-05
Tests requiring internet access


adamjstewart:
### Description

Several of our tests require internet access. This is problematic for a number of reasons:

1. The GitHub Actions runners or the upload servers may have internet issues, leading to intermittent failures
2. Long download times slow down testing (the failing tests I discovered are all in our top 10 slowest tests)

We should apply the same magic we used in previous PRs to monkeypatch any downloads.

### Steps to reproduce

1. Turn off WiFi or unplug the ethernet cable
2. Run `pytest`

```
FAILED tests/models/test_farseg.py::TestFarSeg::test_valid_backbone[resnet18-True] - urllib.error.URLError: <urlopen error [Errno 8] nodename nor servname provided, or not known>
FAILED tests/models/test_farseg.py::TestFarSeg::test_valid_backbone[resnet50-True] - urllib.error.URLError: <urlopen error [Errno 8] nodename nor servname provided, or not known>
FAILED tests/trainers/test_detection.py::TestObjectDetectionTask::test_trainer[True-faster-rcnn-nasa_marine_debris-NASAMarineDebrisDataModule] - urllib.error.URLError: <urlopen error [Errno 8] nodename nor ser...
FAILED tests/trainers/test_detection.py::TestObjectDetectionTask::test_trainer[True-fcos-nasa_marine_debris-NASAMarineDebrisDataModule] - urllib.error.URLError: <urlopen error [Errno 8] nodename nor servname p...
FAILED tests/trainers/test_detection.py::TestObjectDetectionTask::test_trainer[True-retinanet-nasa_marine_debris-NASAMarineDebrisDataModule] - urllib.error.URLError: <urlopen error [Errno 8] nodename nor servn...
FAILED tests/trainers/test_detection.py::TestObjectDetectionTask::test_no_rgb[True] - urllib.error.URLError: <urlopen error [Errno 8] nodename nor servname provided, or not known>
FAILED tests/trainers/test_detection.py::TestObjectDetectionTask::test_predict[True] - urllib.error.URLError: <urlopen error [Errno 8] nodename nor servname provided, or not known>
```

### Version

0.5.0.dev0 (ce4e7ef2ed3ecff5731c2902d4cd3e8f5fab753d)

adamjstewart:
Bonus points if anyone can figure out how to jail our runners so that we can actually test this in CI and prevent it from slipping in again. We seem to fix bugs like this with every release.

[...]

PR #2513, opened 2025-01-13
Trainers: add Instance Segmentation Task

ariannasole23:


tests/trainers/test_instancesegmentation.py
adamjstewart, line 1: Every file needs copyright info at the top, see https://torchgeo.readthedocs.io/en/latest/user/contributing.html#licensing

torchgeo/trainers/instancesegmentation.py
adamjstewart, line 1: Let's rename this `instance_segmentation.py`, same for the test file

tests/trainers/test_instancesegmentation.py
adamjstewart: This file should look almost identical to the other `tests/trainers/*.py` files, you can copy-n-paste the template from there

torchgeo/trainers/instancesegmentation.py
```diff
@@ -0,0 +1,174 @@
+from typing import Any                                           
+import torch.nn as nn                                            
+import torch                                                   
+from torch import Tensor                                        
+from torchmetrics.detection.mean_ap import MeanAveragePrecision  
+from torchvision.models.detection import maskrcnn_resnet50_fpn   
+from ultralytics import YOLO
```
adamjstewart, line 7: We would need to add a dependency on this which we don't currently have. I'm hesitant to do that given that they just had their PyPI account hacked and pushed out bitcoin miners for several releases. But they are quite well known, so it may be difficult not to support them. Let me ask around and see if others have opinions.

torchgeo/trainers/instancesegmentation.py
```diff
@@ -0,0 +1,174 @@
...
+
+class InstanceSegmentationTask(BaseTask):
+    """
+    Task class for training and evaluating instance segmentation models.
+
+    This class supports Mask R-CNN and YOLO models and handles the following:
+    - Model configuration
+    - Loss computation
+    - Metric computation (e.g., Mean Average Precision)
+    - Training, validation, testing, and prediction steps
```
adamjstewart, line 18: Can you add:
```rst
.. versionadded:: 0.7
```

torchgeo/trainers/instancesegmentation.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def __init__(
+        self,
+        model: str = 'mask_rcnn',           # Model type, e.g., 'mask_rcnn' or 'yolo'
+        backbone: str = 'resnet50',         # Backbone type for Mask R-CNN (ignored for YOLO)
+        weights: str | bool | None = None,  # Pretrained weights or custom checkpoint path
+        num_classes: int = 2,               # Number of classes, including background
+        lr: float = 1e-3,                   # Learning rate for the optimizer
+        patience: int = 10,                 # Patience for the learning rate scheduler
+        freeze_backbone: bool = False,      # Whether to freeze backbone layers (useful for transfer learning)
```
adamjstewart, line 29: These descriptions should go inside of the docstring, see the other trainers for examples

torchgeo/trainers/instancesegmentation.py
```diff
@@ -0,0 +1,174 @@
...
+        super().__init__()              # Initialize the BaseTask class (inherits common functionality)
+        self.save_hyperparameters()     # Save input arguments for later use (e.g., in checkpoints or logs)
+        self.model = None               # Placeholder for the model (to be initialized later)
+        self.validation_outputs = []    # List to store outputs during validation (used for debugging or analysis)
+        self.test_outputs = []          # List to store outputs during testing
+        self.configure_models()         # Call method to set up the model
+        self.configure_metrics()        # Call method to set up metrics
+
+    def configure_models(self) -> None:
+        """
```
adamjstewart, line 46: You should copy the docstrings from our other trainers so the documentation is uniform

adamjstewart (changes requested):
Looks good, just need to make the code and tests match our existing trainers and run ruff: https://torchgeo.readthedocs.io/en/latest/user/contributing.html#linters

adamjstewart:
To solve the import issue, you also need to add 2 lines to `torchgeo/trainers/__init__.py`, just copy the style used by other classes.

torchgeo/trainers/instance_segmentation.py
```diff
@@ -0,0 +1,196 @@
...
+"""Trainers for instance segmentation."""
+
+from typing import Any                                            
+import torch.nn as nn                                            
+import torch                                                     
+from torch import Tensor                                         
+from torchmetrics.detection.mean_ap import MeanAveragePrecision  
+from torchmetrics import MetricCollection
+from torchvision.models.detection import maskrcnn_resnet50_fpn
+from base import BaseTask
```
adamjstewart, line 13:
```suggestion
from .base import BaseTask  

```

adamjstewart: That error is gone from the tests, and I don't have that error locally when I try your code. Make sure you are in the root directory of the TorchGeo repo, otherwise `from torchgeo import ...` will import the installed version of torchgeo instead of your local changes. You can double check this by running:
```python
import torchgeo
print(torchgeo.__file__)
```
It should print the path to your current directory.

ariannasole23: I can't get it to run. In which directory should I be?

adamjstewart: You should be in the root directory of the git repository.

test_trainer_instancesegmentation.py
adamjstewart, line 1: This file can be deleted

tests/trainers/test_instancesegmentation.py
adamjstewart, line 1: Let's rename to `test_instance_segmentation.py` to match the other filename

tests/trainers/test_instancesegmentation.py
```diff
@@ -0,0 +1,257 @@
...
+
+def plot_missing_bands(*args: Any, **kwargs: Any) -> None:
+    raise RGBBandsMissingError()
+
+
+class TestSemanticSegmentationTask:
+    @pytest.mark.parametrize(
+        'name',
+        [
+            'agrifieldnet',
```
adamjstewart, line 52: This is a list of files whose configuration is present in `tests/conf`. This current list is from the semantic segmentation tests. We need to create a new list (and new `tests/conf/*.yaml` files) for instance segmentation datasets. We can start with something like VHR-10 and work from there.

tests/trainers/test_instancesegmentation.py
```diff
@@ -0,0 +1,257 @@
...
+        match name:
+            case 'chabud' | 'cabuar':
+                pytest.importorskip('h5py', minversion='3.6')
+            case 'ftw':
+                pytest.importorskip('pyarrow')
+            case 'landcoverai':
+                sha256 = (
+                    'ecec8e871faf1bbd8ca525ca95ddc1c1f5213f40afb94599884bd85f990ebd6b'
+                )
+                monkeypatch.setattr(LandCoverAI, 'sha256', sha256)
```
adamjstewart, line 98: Can be removed if you aren't using any of these datasets.

tests/trainers/test_instancesegmentation.py
```diff
@@ -0,0 +1,257 @@
...
+
+        config = os.path.join('tests', 'conf', name + '.yaml')
+
+        monkeypatch.setattr(smp, 'Unet', create_model)
+        monkeypatch.setattr(smp, 'DeepLabV3Plus', create_model)
```
adamjstewart, line 103: We aren't using these models (or even smp) for this task, this can be removed

tests/trainers/test_instancesegmentation.py
```diff
@@ -0,0 +1,257 @@
...
+            InstanceSegmentationTask(model='invalid_model')
+
+    def test_invalid_loss(self) -> None:
+        match = "Loss type 'invalid_loss' is not valid."
+        with pytest.raises(ValueError, match=match):
+            InstanceSegmentationTask(loss='invalid_loss')
+
+    def test_no_plot_method(self, monkeypatch: MonkeyPatch, fast_dev_run: bool) -> None:
+        monkeypatch.setattr(SEN12MSDataModule, 'plot', plot)
+        datamodule = SEN12MSDataModule(
```
adamjstewart, line 196: This datamodule isn't compatible with instance segmentation

tests/trainers/test_instancesegmentation.py
```diff
@@ -0,0 +1,257 @@
...
+        )
+        assert all([param.requires_grad for param in model.model.decoder.parameters()])
+        assert all(
+            [
+                param.requires_grad
+                for param in model.model.segmentation_head.parameters()
+            ]
+        )
+
+    # @pytest.mark.parametrize('model_name', ['unet', 'deeplabv3+'])
```
adamjstewart, line 245: Can remove commented-out code

torchgeo/trainers/instance_segmentation.py
```diff
@@ -0,0 +1,251 @@
...
+                representation of a weight enum, True for ImageNet weights, False or
+                None for random weights, or the path to a saved model state dict.
+            in_channels: Number of input channels to model.
+            num_classes: Number of prediction classes (including the background).
+            lr: Learning rate for optimizer.
+            patience: Patience for learning rate scheduler.
+            freeze_backbone: Freeze the backbone network to fine-tune the
+                decoder and segmentation head.
+
+        .. versionadded:: 0.7
```
adamjstewart, line 47: Can you move this versionadded to the class docstring?

torchgeo/trainers/instance_segmentation.py
```diff
@@ -0,0 +1,251 @@
...
+        """
+        self.weights = weights         
+        super().__init__()              
+        self.save_hyperparameters()     
+        self.model = None               
+        self.validation_outputs = []    
+        self.test_outputs = []          
+        self.configure_models()         
+        self.configure_metrics()
```
adamjstewart, line 56: None of these are needed, they are added by the base class

torchgeo/trainers/instance_segmentation.py
```diff
@@ -0,0 +1,251 @@
...
+
+        Raises:
+            ValueError: If *model* is invalid.
+        """
+        model = self.hparams['model'].lower()      
+        num_classes = self.hparams['num_classes']  
+
+        if model == 'mask_rcnn':
+            # Load the Mask R-CNN model with a ResNet50 backbone
+            self.model = maskrcnn_resnet50_fpn(weights=MaskRCNN_ResNet50_FPN_Weights.DEFAULT)
```
adamjstewart, line 69: Weights will be chosen by the user, we can't assume ImageNet weights

torchgeo/trainers/instance_segmentation.py
```diff
@@ -0,0 +1,251 @@
...
+            for param in self.model.backbone.parameters():
+                param.requires_grad = False  
+
+
+    def configure_metrics(self) -> None:
+        """Initialize the performance metrics.
+
+        - Uses Mean Average Precision (mAP) for masks (IOU-based metric).
+        """
+        self.metrics = MetricCollection([MeanAveragePrecision(iou_type="segm")])
```
adamjstewart, line 98: No need to use MetricCollection if you only have a single metric

torchgeo/trainers/instance_segmentation.py
```diff
@@ -0,0 +1,251 @@
...
+        Returns:
+            The total loss for the batch.
+        """
+        images, targets = batch['image'], batch['target']     
+        loss_dict = self.model(images, targets)               
+        loss = sum(loss for loss in loss_dict.values())  
+
+        print('\nTRAINING LOSS\n')
+        print(loss_dict, '\n\n')
+        print(loss)
```
adamjstewart, line 119: Can remove print statements, PyTorch Lightning has builtin loggers.

adamjstewart (changes requested):
Can you run ruff on the code to make it more uniform? This will make it easier to review.

adamjstewart:
Note that #1978 was just merged, which changes some of the VHR-10 dict keys and values to be more conformant with Kornia. We should probably rebase/merge this PR to make sure it works with these new changes.

adamjstewart:
TODO: figure out how to handle multispectral inputs, will need to change input layer for 3+ bands.

torchgeo/trainers/instance_segmentation.py
```diff
@@ -0,0 +1,236 @@
...
+# Licensed under the MIT License.
+
+"""Trainers for instance segmentation."""
+
+from typing import Any
+
+import matplotlib.pyplot as plt
+import torch
+from matplotlib.figure import Figure
+from timm.models import adapt_input_conv  # type: ignore[attr-defined]
```
adamjstewart, line 11: Fixed in https://github.com/huggingface/pytorch-image-models/pull/2442

torchgeo/trainers/instance_segmentation.py
```diff
@@ -0,0 +1,236 @@
...
+                    weights_backbone=weights_backbone,
+                )
+            else:
+                msg = f"Invalid backbone type '{backbone}'. Supported backbone: 'resnet50'"
+                raise ValueError(msg)
+        else:
+            msg = f"Invalid model type '{model}'. Supported model: 'mask_rcnn'"
+            raise ValueError(msg)
+
+        self.model.backbone.body.conv1.weight = adapt_input_conv(  # type: ignore[no-untyped-call]
```
adamjstewart, line 96: Fixed in https://github.com/huggingface/pytorch-image-models/pull/2441

tests/trainers/test_instance_segmentation.py
```diff
@@ -0,0 +1,165 @@
...
+        with pytest.raises(ValueError, match=match):
+            InstanceSegmentationTask(model='invalid_model')
+
+    def test_invalid_backbone(self) -> None:
+        match = 'Invalid backbone type'
+        with pytest.raises(ValueError, match=match):
+            InstanceSegmentationTask(backbone='invalid_backbone')
+
+    def test_weights(self) -> None:
+        InstanceSegmentationTask(weights=True, num_classes=91)
```
adamjstewart, line 115: Yet another test that downloads data on the fly (#1088), but let's fix that another day when we figure out how to support custom weights.

adamjstewart:
Hmm, looks like FTW isn't supported because it only has masks, not bboxes.

adamjstewart (dismissed):
Tests are finally passing! @ariannasole23 maybe you can check VHR-10 again to make sure I didn't break anything and it still works. Would like a couple other people to review this too, but otherwise this is ready to merge once you accept the CLA.

torchgeo/trainers/instance_segmentation.py
```diff
@@ -0,0 +1,254 @@
...
+          Recall is the number of true positives divived by the number of true positives
+          + false negatives. Uses 'macro' averaging. Higher values are better.
+
+        .. note::
+           * 'Micro' averaging suits overall performance evaluation but may not
+             reflect minority class accuracy.
+           * 'Macro' averaging gives equal weight to each class, and is useful for
+             balanced performance assessment across imbalanced classes.
+        """
+        metrics = MetricCollection([MeanAveragePrecision(iou_type=('bbox', 'segm'))])  # type: ignore[arg-type]
```
adamjstewart, line 120: Fixed in https://github.com/Lightning-AI/torchmetrics/pull/2966

ariannasole23:

adamjstewart (dismissed):
Will give others a few days to test before merging.

ashnair1:
The current InstanceSegmentationTask implementation isn't really flexible. I tried training a model with it recently and found the following:
  - Training a model with `weights=True` i.e. with a pretrained backbone crashes as the FPN Weights trained on COCO require number of classes to be 91 whereas the VHR10 has 11 classes (incl background). Currently, the only way to train this model would be to have `weights=None`. See below for sample training config
  - Using the Resnet50_FPN_Weights means that users can only use the resnet50 backbone. The `ObjectDetectionTask` allows users to choose their preferred backbone i.e. resnet18, resnet34, resnet50 etc. 

 Train config used:
```yaml
trainer:
  min_epochs: 15
  max_epochs: 100

model:
  class_path: InstanceSegmentationTask
  init_args:
    model: 'mask-rcnn'
    backbone: 'resnet50'  # Only resnet50 can be used
    weights: None  # Using pretrained weights (weights:True) will not work in this implementation
    num_classes: 11
    lr: 3.0e-5 
```
I'd suggest following the implementation of how Faster-RCNN is defined in `ObjectDetectionTask` while defining Mask-RCNN in `InstanceSegmentationTask`.

adamjstewart:
* `weights=True, num_channels!=91` has been "fixed" and tested in the latest commit. I say "fixed" because I just skip the decoder weights when `num_channels!=91`. A smarter solution would be to drop the last layer of the weights, but I was lazy.
* Yep, this trainer currently lacks a lot of options, and only supports a single model and backbone. I agree that the implementation in `ObjectDetectionTask` is more flexible, but @ariannasole23 is busy with final exams and I'm busy preparing the release. I would suggest merging this as is and you can open a follow-up PR to add more backbones if you want.

adamjstewart (approved):

ashnair1 (approved):

adamjstewart:
FYI, we ended up adding an optional dependency on ultralytics in #2876. By making it optional, there are fewer licensing concerns, it's up to the user whether or not they are allowed to use YOLO. So we could actually update the trainer to support YOLO too if anyone wants to submit a PR.

PR #2520, opened 2025-01-18
Add BRIGHT dataset

nilsleh:
This PR adds the [BRIGHT](https://github.com/ChenHongruixuan/BRIGHT) dataset.

Dataset Features:

    * Pre-disaster optical images from MAXAR, NAIP, NOAA Digital Coast Raster Datasets, and the National Plan for Aerial Orthophotography Spain
    * Post-disaster SAR images from Capella Space and Umbra
    * high image resolution of 0.3-1m

Dataset Format:

    * Images are in GeoTIFF format with pixel dimensions of 1024x1024
    * Pre-disaster are three channel images
    * Post-disaster SAR images are single channel but repeated to have 3 channels

bright_example

@ChenHongruixuan @olidietrich Thank you for the nice work and making the dataset public. I was wondering whether you could include the split .txt files found [here](https://github.com/ChenHongruixuan/BRIGHT/tree/master/dfc25_benchmark/dataset/splitname) inside the zip file on Huggingface such that everything is in one place? And of course if you have any other comments about the PR, feel free to let us know below.


torchgeo/datasets/bright.py
```diff
@@ -0,0 +1,338 @@
...
+from matplotlib import colors
+from matplotlib.figure import Figure
+from torch import Tensor
+
+from .errors import DatasetNotFoundError
+from .geo import NonGeoDataset
+from .utils import Path, check_integrity, download_url, extract_archive
+
+
+class BRIGHTDFC2025(NonGeoDataset):
```
adamjstewart, line 26: Wondering if we should call it BRIGHT instead of BRIGHT DFC2025.

docs/api/datasets/non_geo_datasets.csv
```diff
@@ -3,6 +3,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `Benin Cashew Plantations`_,S,Airbus Pléiades,"CC-BY-4.0",70,6,"1,122x1,186",10,MSI
 `BigEarthNet`_,C,Sentinel-1/2,"CDLA-Permissive-1.0","590,326",19--43,120x120,10,"SAR, MSI"
 `BioMassters`_,R,Sentinel-1/2 and Lidar,"CC-BY-4.0",,,256x256, 10, "SAR, MSI"
+`BRIGHT DFC2025`_,CD,"MAXAR, NAIP, Capella, Umbra","CC-BY-SA-4.0",3239,4,"0.1-1","RGB,SAR"
```
adamjstewart, line 6: Bit confused about the license. https://huggingface.co/datasets/Kullervo/BRIGHT says CC-BY-SA-4.0, https://zenodo.org/records/14619798 says CC-BY-4.0 AND CC-BY-NC-4.0, and https://www.grss-ieee.org/technical-committees/image-analysis-and-data-fusion/?tab=data-fusion-contest says CC-BY-2.0 AND CC-BY-4.0 AND CC-BY-NC-4.0.

torchgeo/datasets/bright.py
```diff
@@ -0,0 +1,338 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""BRIGHT dataset."""
+
+import os
+import textwrap
```
adamjstewart, line 7: First time learning about this, interesting!

torchgeo/datasets/bright.py
```diff
@@ -0,0 +1,338 @@
...
+
+from .errors import DatasetNotFoundError
+from .geo import NonGeoDataset
+from .utils import Path, check_integrity, download_url, extract_archive
+
+
+class BRIGHTDFC2025(NonGeoDataset):
+    """BRIGHT DFC2025 dataset.
+
+    The `BRIGHT <https://github.com/ChenHongruixuan/BRIGHT>`__ dataset consists of bi-temporal high-resolution multimodal images for
```
adamjstewart, line 29: Can we wrap this to 88 chars?

torchgeo/datasets/bright.py
```diff
@@ -0,0 +1,338 @@
...
+        Returns:
+            data and target at that index
+        """
+        idx_paths = self.sample_paths[index]
+
+        pre_image = self._load_image(idx_paths['pre_image']).float()
+        post_image = self._load_image(idx_paths['post_image']).float()
+        # https://github.com/ChenHongruixuan/BRIGHT/blob/11b1ffafa4d30d2df2081189b56864b0de4e3ed7/dfc25_benchmark/dataset/make_data_loader.py#L101
+        # post image is stacked to also have 3 channels
+        post_image = repeat(post_image, 'c h w -> (repeat c) h w', repeat=3)
```
adamjstewart, line 137: Hmm, does this make sense? Is this only because they want to use RGB models pre-trained on ImageNet?

torchgeo/datasets/bright.py
```diff
@@ -0,0 +1,338 @@
...
+
+        sample = {'pre_image': pre_image, 'post_image': post_image}
```
adamjstewart, line 139: Not sure if we've finalized a standard for CD datasets yet. I think we're currently leaning towards T x C x H x W.

torchgeo/datasets/bright.py
```diff
@@ -0,0 +1,338 @@
...
+        if self.transforms is not None:
+            sample = self.transforms(sample)
+
+        return sample
+
+    def _get_paths(self) -> list[dict[str, str]]:
+        """Get paths to the dataset files based on specified splits.
+
+        Returns:
+            a list of dictionaries containing paths to the pre,post and target images
```
adamjstewart, line 154:
```suggestion
            a list of dictionaries containing paths to the pre, post, and target images
```

ChenHongruixuan:
> This PR adds the [BRIGHT](https://github.com/ChenHongruixuan/BRIGHT) dataset.
> 
> Dataset Features:
> 
> ```
> * Pre-disaster optical images from MAXAR, NAIP, NOAA Digital Coast Raster Datasets, and the National Plan for Aerial Orthophotography Spain
> * Post-disaster SAR images from Capella Space and Umbra
> * high image resolution of 0.3-1m
> ```
> 
> Dataset Format:
> 
> ```
> * Images are in GeoTIFF format with pixel dimensions of 1024x1024
> * Pre-disaster are three channel images
> * Post-disaster SAR images are single channel but repeated to have 3 channels
> ```
> 
> bright_example
> 
> @ChenHongruixuan @olidietrich Thank you for the nice work and making the dataset public. I was wondering whether you could include the split .txt files found [here](https://github.com/ChenHongruixuan/BRIGHT/tree/master/dfc25_benchmark/dataset/splitname) inside the zip file on Huggingface such that everything is in one place? And of course if you have any other comments about the PR, feel free to let us know below.

Hi @nilsleh ,

Thank you very much for your contribution and efforts in integrating BRIGHT into torchgeo! We are a little hesitant to include the txt file you mentioned into this for torchgeo, as that is only for the DFC25 division, and note that the current BRIGHT is not final version.

Best,

torchgeo/datasets/bright.py
nilsleh: Since this did not seem to be the final version, rather the Challenge version, I included it in the name. However, the paper and also comment below noted, that this is not the final version. Since, I would expect or hope that the dataset format stays the same for the final version, we could also add a version parameter.

docs/api/datasets/non_geo_datasets.csv
nilsleh: @ChenHongruixuan do you have input on this?

nilsleh:
> Thank you very much for your contribution and efforts in integrating BRIGHT into torchgeo! We are a little hesitant to include the txt file you mentioned into this for torchgeo, as that is only for the DFC25 division. That's why we provide these txt files in github instead of original dataset zip file, especially considering that the current BRIGHT is not final version.

Thanks for getting back @ChenHongruixuan , no problem, we can also host them on our torchgeo Hugginface, just to make downloading more easier. Also very much looking forward to the final version of the dataset. Feel free to let me know when that is openly available, such that we can integrate support for that as well here.

torchgeo/datasets/bright.py
adamjstewart: Alright, then let's assume we will have more BRIGHT datasets in the future.

docs/api/datasets.rst
```diff
@@ -217,6 +217,11 @@ BioMassters
 
 .. autoclass:: BioMassters
 
+BRIGHT DFC2025
+^^^^^^^^^^^^^^
```
adamjstewart, line 221:
```suggestion
BRIGHT
^^^^^^
```
In case we have more in the future.

torchgeo/datasets/bright.py
nilsleh: wait so do you mean keep the dataset name like it is and instead just put it under a "Bright" umbrella of possible datasets in the docs?

ChenHongruixuan:
> > Thank you very much for your contribution and efforts in integrating BRIGHT into torchgeo! We are a little hesitant to include the txt file you mentioned into this for torchgeo, as that is only for the DFC25 division. That's why we provide these txt files in github instead of original dataset zip file, especially considering that the current BRIGHT is not final version.
> 
> Thanks for getting back @ChenHongruixuan , no problem, we can also host them on our torchgeo Hugginface, just to make downloading more easier. Also very much looking forward to the final version of the dataset. Feel free to let me know when that is openly available, such that we can integrate support for that as well here.

@nilsleh , thank you so much for your support!!

torchgeo/datasets/bright.py
adamjstewart: Yep. So this dataset will be the DFC2025 version, and we may add more versions in the future. Same reason you used `bright.py` we should also use BRIGHT in the docs.

docs/api/datasets/non_geo_datasets.csv
ChenHongruixuan: hi @nilsleh , the images of BRIGHT follow CC-BY-4.0 AND CC-BY-NC-4.0. But in huggingface, we can only choose one license. But we made a detailed description in the readme. The CC-BY-2.0 is the license of some images in another dataset of that contest, called OpenEarthMap-SAR.

torchgeo/datasets/bright.py
nilsleh: https://github.com/microsoft/torchgeo/blob/8a3b60df4e82877e46ea3cd7fb97636149f0f980/torchgeo/datasets/levircd.py#L70 Levir Change detection does separate dict entries, I think 'pre_image' and 'post_image' is more clear than 'image1' and 'image2'

torchgeo/datasets/bright.py
adamjstewart: It depends. If we want to use one of our S1 pretrained models, not duplicating may make more sense. I would personally leave it up to the user to add a transform if they want this.

torchgeo/datasets/bright.py
adamjstewart: We're going to reformat Levir soon. Note that Kornia AugmentationSequential only supports "image*`, not "*image".

docs/api/datasets/non_geo_datasets.csv
```diff
@@ -3,6 +3,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `Benin Cashew Plantations`_,S,Airbus Pléiades,"CC-BY-4.0",70,6,"1,122x1,186",10,MSI
 `BigEarthNet`_,C,Sentinel-1/2,"CDLA-Permissive-1.0","590,326",19--43,120x120,10,"SAR, MSI"
 `BioMassters`_,R,Sentinel-1/2 and Lidar,"CC-BY-4.0",,,256x256, 10, "SAR, MSI"
+`BRIGHT`_,CD,"MAXAR, NAIP, Capella, Umbra","CC-BY-NC-4.0",3239,4,"0.1-1","RGB,SAR"
```
adamjstewart, line 6:
```suggestion
`BRIGHT`_,CD,"MAXAR, NAIP, Capella, Umbra","CC-BY-4.0 AND CC-BY-NC-4.0",3239,4,"0.1-1","RGB,SAR"
```
Technically these are probably equivalent, but someone may want to use only the images or only the masks.

nilsleh: one is Non-commerrcial and the other is right? That's quiet a big difference. And the HF says that the vast majority of the data is Non-commercial since it's from Maxar, so I thought safer to put NC

adamjstewart: `CC-BY-4.0 AND CC-BY-NC-4.0` implies you need to satisfy both licenses, see https://spdx.dev/learn/handling-license-info/

nilsleh: cool, learned something again, thanks!

torchgeo/datasets/bright.py
nilsleh: My thinking was that this is how it is implemented in the challenge code, so I thought it would make sense to follow that. I think a constant nummer of channels across the two images makes the dataset easier to use. For anything more specific, I would expect that a user has to write their own collate function/augmentations anyways. Also pretrained RGB models are tough to beat, even if trained on ImageNet https://arxiv.org/abs/2305.13456. And I don't think it would prevent you from using a pretrained S1 model, you would have to write a custom trainer anways to handle the inputs I believe.

nilsleh: Also note that the current pretrained S1 models https://torchgeo.readthedocs.io/en/stable/api/models.html#sentinel-1 expect 2 input channels, so they would also not work without a channel repeat for this dataset.

adamjstewart: Alright, those are convincing arguments. We can consider dropping this for the non-DFC2025 version of the dataset.

adamjstewart (approved):

adamjstewart:
Seeing these warnings in CI:
```
tests/datasets/test_bright.py::TestBRIGHTDFC2025::test_getitem[train]
tests/datasets/test_bright.py::TestBRIGHTDFC2025::test_getitem[val]
tests/datasets/test_bright.py::TestBRIGHTDFC2025::test_getitem[test]
tests/datasets/test_bright.py::TestBRIGHTDFC2025::test_plot[train]
tests/datasets/test_bright.py::TestBRIGHTDFC2025::test_plot[val]
tests/datasets/test_bright.py::TestBRIGHTDFC2025::test_plot[test]
  /opt/hostedtoolcache/Python/3.12.8/x64/lib/python3.12/site-packages/rasterio/__init__.py:356: NotGeoreferencedWarning: Dataset has no geotransform, gcps, or rpcs. The identity matrix will be returned.
    dataset = DatasetReader(path, driver=driver, sharing=sharing, **kwargs)
```
If the images are not supposed to be georeferenced, we should use PIL instead of rasterio.

PR #2523, opened 2025-01-20
Docs: ignore fewer undocumented classes

adamjstewart:
Classes are either no longer used or now have docs.


adamjstewart:
Nvm, all the lightning ones are included in an alias. The torch one could still be removed, but isn't worth a PR.

PR #2528, opened 2025-01-23
Add AI4ArcticSeaIce dataset.

nilsleh:
This PR adds the [AI4ArcticSeaIce](https://data.dtu.dk/collections/AI4Arctic_Sea_Ice_Challenge_Dataset/6244065) dataset. Rehosted to HF for faster download times (couple minutes vs ~8hours). In the [original repo](https://github.com/astokholm/AI4ArcticSeaIceChallenge/blob/4d5e3bc85e681f6c56821d96f2ebfcf4ed58b495/loaders.py) there is some additional useful information.

```markdown
The Sea Ice Challenge Dataset contains Sentinel-1 SAR imagery, passive microwave radiometer observations
    from AMSR2, and numerical weather prediction data from the ECMWF Reanalysis v5 (ERA5) dataset - all
    gridded to match the Sentinel-1 SAR scenes geometrically. As label data, the dataset contains ice charts
    manually produced by the ice analysts at the Greenland Ice Service and the Canadian Ice Service.

Dataset features:

    * Dual-polarization SAR (HH, HV) imagery for each patch.
    * Sea Ice Concentration (SIC): the percentage ratio of sea ice to open water for an area,
        discretized into 11 10% bins ranging from 0% to 100%.
    * Stage Of Development (SOD): type of sea ice, as proxy for ice thickness and
        ease of traversing with 6 classes
    * Floe size (FLOE): Classifying or segmenting distinct ice floes based on size, shape,
          or other geometric properties.

Dataset format:

    * each sample scene is stored in a separate .nc file
    * pixel dimension of varying sizes up to ~5000pxx5000px
    * 80m resolution
```

TODOS:

- [ ] check plotting again
- [ ] think about resizing since individual tiles can be really large

example_sea_ice

Thanks @astokholm for the creation and open-sourcing dataset. This dataset has some complexities due to all the different data modalities, so if you have any comments/corrections, it would be much appreciated.


khdlr:
This is the v2 dataset, right? (v1 is still around: https://data.dtu.dk/collections/AI4Arctic_Sea_Ice_Challenge_Dataset/6244065/1). Might be helpful to document this in the dataset class.

nilsleh:
> This is the v2 dataset, right? (v1 is still around: https://data.dtu.dk/collections/AI4Arctic_Sea_Ice_Challenge_Dataset/6244065/1). Might be helpful to document this in the dataset class.

Good catch, it is actually Version 3, I corrected the link above.

PR #2531, opened 2025-01-24
Add BigEarthNet Version2

nilsleh:
Superseeds #2371 after talking to Ando. 

After taking a look, the new version comes with a `metadata.parquet` file, which makes data handling quiet a bit more straightforward. With a `version=2` argument, I felt like there would be many nested if statements and therefore, cleaner to do it this way. If there was a similar `metadata.parquet` file for V1, then this could be made more condensed.

example_benv2


torchgeo/datasets/bigearthnet.py
```diff
@@ -495,7 +504,7 @@ def _download(self, url: str, filename: Path, md5: str) -> None:
             filename: output filename to write downloaded file
             md5: md5 of downloaded file
         """
-        if not os.path.exists(filename):
+        if not os.path.exists(os.path.join(self.root, filename)):
```
nilsleh, line 507: Saw this, and actually I think this is required, because it should check whether the file exists in root already right?

torchgeo/datasets/bigearthnet.py
```diff
@@ -573,3 +582,325 @@ def plot(
...
+    def _load_target(self, index: int) -> Tensor:
+        """Load the target mask for a single image.
+
+        Args:
+            index: index to return
+
+        Returns:
+            the target label
+        """
+        indices = self.metadata_df.iloc[index]['labels']
```
nilsleh, line 792: Also need to check the labels. In the V1 class it only seems possible to select the 19 label versions, because with selecting 43 they also get mapped to 19 if I understand correctly.

torchgeo/datasets/bigearthnet.py
adamjstewart: Good catch. Actually, this line isn't necessary because we already check if the zipfile exists and extract it before we reach the download. Can you open a separate PR to fix this so we can backport it to 0.6.3?

nilsleh:
@ando-shah in case you wanna have a look, since you have experience with the dataset already and find anything.

torchgeo/datasets/bigearthnet.py
```diff
@@ -573,3 +585,441 @@ def plot(
         if suptitle is not None:
             plt.suptitle(suptitle)
         return fig
+
+
+class BigEarthNetV2(NonGeoDataset):
```
adamjstewart, line 590: Could also subclass BigEarthNet if you want less code duplication. Or make a shared base class. Depends on how similar they are.

torchgeo/datasets/bigearthnet.py
```diff
@@ -573,3 +585,441 @@ def plot(
...
+        else:
+            paths = sorted(paths)
+
+        images = []
+        for path in paths:
+            with rasterio.open(path) as dataset:
+                array = dataset.read(
+                    indexes=1,
+                    out_shape=self.image_size,
+                    out_dtype='int32',
```
adamjstewart, line 820: Should we load as float32 since we're later converting to float32 anyway? Or just load in the original dtype?

adamjstewart (changes requested):
Once this is done I would love to squash this on top of #2371 so that @ando-shah gets some credit.

requirements/datasets.txt
```diff
@@ -6,3 +6,4 @@ pandas[parquet]==2.2.3
 pycocotools==2.0.8
 scikit-image==0.25.0
 scipy==1.15.1
+zstandard==0.23.0
```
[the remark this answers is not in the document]
nilsleh: mhm okay, this is a lot of work though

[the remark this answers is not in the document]
adamjstewart: Is it a lot of work? You already have the data downloaded and extracted. You just need to compress, upload, and change some URLs. Seems like less work than adding this to `pyproject.toml` and finding the minimum supported version.

We only have a single dataset that needs this dep. We already removed our zipfile-deflate64 and rarfile deps for the same reasons, fewer deps to maintain and easier compatibility with `extract_archive`.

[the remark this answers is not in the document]
nilsleh: I mean have to rewrite the whole data.py creation and file schema and downloading/extraction logic, which is always the most work when adding a dataset.

torchgeo/datasets/bigearthnet.py
```diff
@@ -572,3 +582,444 @@ def plot(
...
+    def _extract(self) -> None:
+        """Extract the tarball parts.
+
+        For each modality (s1, s2, maps), its parts are concatenated together and then extracted.
+        """
+        chunk_size = 2**15  # same as used in torchvision and ssl4eo
+        for key, meta in self.metadata_locs.items():
+            if key == 'metadata':
+                continue
+            assert not os.path.isdir(os.listdir(self.root)[2])
```
nilsleh, line 929: @adamjstewart any idea what could cause this error? Basically, when you check whether the tar ball file in the root is a directory it says no, but in the next line, when you do `os.path.join()` with the root, it all of the sudden is a directory... I don't see what I messed up.

adamjstewart: Fixed in latest commit.

requirements/datasets.txt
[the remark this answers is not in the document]
adamjstewart: zstandard is no longer used now, right? We can remove all references to it.

torchgeo/datasets/bigearthnet.py
adamjstewart: It looks like the raw data is uint16? int32 doesn't make much sense, no reason to cast dtype multiple times.

nilsleh: I don't think pytorch has uint16 datatype, so I am guessing that is why it is being read as int32 so that the tensor can be built and then cast to float32

torchgeo/datasets/bigearthnet.py
```diff
@@ -572,3 +576,430 @@ def plot(
...
+        523: 18,  # Sea and ocean
+    }
+
+    valid_splits = ('train', 'val', 'test')
+
+    def __init__(
+        self,
+        root: Path = 'data',
+        split: str = 'train',
+        bands: str = 'all',
```
adamjstewart, line 688: People have complained that some datasets have (B1, B2, etc.) and others have s1, s2, etc. I know you're just matching the existing dataset, but should we change both to the former?

torchgeo/datasets/bigearthnet.py
```diff
@@ -572,3 +576,430 @@ def plot(
...
+        exists = []
+        for key, metadata in self.metadata_locs.items():
+            if key != 'metadata':
+                exists.append(
+                    os.path.exists(os.path.join(self.root, self.dir_file_names[key]))
+                )
+            else:
+                exists.append(
+                    os.path.exists(os.path.join(self.root, self.dir_file_names[key]))
+                )
```
adamjstewart, line 864: Maybe I'm blind, but what is the difference here?

torchgeo/datasets/bigearthnet.py
```diff
@@ -572,3 +576,430 @@ def plot(
...
+                    exists.append(True)
+                else:
+                    exists.append(False)
+            else:
+                for fname in metadata['files']:
+                    fpath = os.path.join(self.root, fname)
+                    if os.path.exists(fpath):
+                        exists.append(True)
+                    else:
+                        exists.append(False)
```
adamjstewart, line 883:
```suggestion
                    exists.append(os.path.exists(fpath))
```
Same above

adamjstewart (approved):
Minor comments, can probably merge if you're in a rush.

adamjstewart (approved):
I'm happy if you're happy.

PR #2535, opened 2025-01-27
SSL4EO-L: add additional metadata

adamjstewart:


adamjstewart:
mypy error is a bit pedantic here. I don't really want to ignore it, but I also don't know an easy way to allow it.

calebrob6:
would `.get(band)` work?

torchgeo/datasets/landsat.py
```diff
@@ -148,6 +152,13 @@ class Landsat1(Landsat):
     default_bands = ('B4', 'B5', 'B6', 'B7')
     rgb_bands = ('B6', 'B5', 'B4')
 
+    wavelengths: ClassVar[dict[str, float]] = {
```
nilsleh, line 155: In the S12 PR, you also added the alternative names, would that make sense here as well?

adamjstewart: As far as I know there are no datasets that use the alternative names, so no need yet.

nilsleh (approved):

PR #2540, opened 2025-01-28
Fix file-specific test coverage

adamjstewart:
@calebrob6 does this fix the issue you reported? It works for me, but I'm not 100% sure it will always work for everyone. If you can figure out a way to break this I'll reach out to the pytest-cov developers because I really want this to work, especially as we increase the number of datasets.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -43,7 +43,7 @@ jobs:
         run: pip list
       - name: Run pytest checks
         run: |
-          pytest --cov --cov-report=xml
+          pytest --cov=torchgeo --cov-report=xml
           python3 -m torchgeo --help
       - name: Report coverage
         uses: codecov/codecov-action@v5.3.1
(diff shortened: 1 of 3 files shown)


calebrob6:
Nope, same error. Were you able to reproduce?

Similarly, there are a bunch of errors if I simply do `pytest --cov=torchgeo.datasets.landcoverai`.

calebrob6:
Tracking issue here https://github.com/microsoft/torchgeo/issues/2541

adamjstewart:
I'm not able to reproduce, it sometimes works for me and sometimes doesn't. I think we should reach out to the pytest-cov folks to get more explanation of what is and isn't supposed to work.

calebrob6:
You are saying that sometimes `pytest --cov=torchgeo.datasets.landcoverai tests/datasets/test_landcoverai.py` gives the same error that I got but sometimes it works as expected?

adamjstewart:
It's probably dataset specific. For example, advance works, but landcoverai fails.

calebrob6:
Advance doesn't work for me....
image

PR #2554, opened 2025-02-01
Update seg and class metrics

robmarkcole:
Replaces https://github.com/microsoft/torchgeo/pull/2130


robmarkcole:
Not sure why suddenly 15 files are changed when I only edited 4 on my branch

PR #2555, opened 2025-02-03
RasterDataset: assert valid bands

adamjstewart:
Curious how many datasets are missing a definition of `all_bands`.


adamjstewart:
Ah, I don't know if this will work. Some datasets like Landsat explicitly don't define `all_bands` because we want to support `B1`, `SR_B1`, etc.

PR #2557, opened 2025-02-03
EuroSAT: use percentile normalization during plotting

adamjstewart:
Related to #1263 

The current plotting method works great for our dataset, but fails as soon as normalization is added (e.g., via our data module). By using percentile normalization, both the dataset and data module should work.

P.S. We should do this for all datasets if anyone is ever bored and looking for an easy contribution. I only did EuroSAT because it's used throughout our tutorials.

### DIFF
--- a/torchgeo/datasets/eurosat.py
+++ b/torchgeo/datasets/eurosat.py
@@ -15,7 +15,14 @@ from torch import Tensor
 
 from .errors import DatasetNotFoundError, RGBBandsMissingError
 from .geo import NonGeoClassificationDataset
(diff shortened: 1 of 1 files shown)


calebrob6:
I don't think percentile normalization is a good way to get around whatever effects that the augmentations/normalization step adds. In some cases (e.g. I'm augmenting color or brightness) I'd like to be able to see differences between samples clearly. Having a reverse augmentation step is a better way to handle that case.

`image = np.clip(image / 3000, 0, 1)` produces very reasonable looking output that you can compare between images, why change this?

adamjstewart:
Superseded by #2560

PR #2559, opened 2025-02-03
Drop support for Python 3.10

adamjstewart:
See [SPEC 0](https://scientific-python.org/specs/spec-0000/) for our typical support schedule.


adamjstewart:
NEP 29 is deprecated, SPEC 0 has a 3 yr drop schedule.

calebrob6:
Nice, I see you also updated the other min-versions considering what were dropped in 2024q4

calebrob6 (approved):

adamjstewart:
Other min-versions are actually coming from Python 3.11+ support, not from SPEC 0. But numpy and friends also follow SPEC 0, so we get that for free.

fmigneault:
> See [SPEC 0](https://scientific-python.org/specs/spec-0000/) for our typical support schedule.

@adamjstewart 
Is this indicated somewhere in the README or elsewhere? The https://scientific-python.org/specs/spec-0000/ page does not seem to list `torchgeo` either. If this is the expected procedure, can it be better advertised?

This update came as a surprise in one of our project by testing upcoming 0.8 models. We are lucky to have found out about it by randomly testing stuff. In order for us to plan early upcoming changes impacting our pipelines, this kind of changes should be better exposed. Its hard to plan these migrations without further notice as a user not actively developing the package, while "minor releases" are mixed in with many other features and breaking changes.

adamjstewart:
Thanks for the feedback @fmigneault!

Our supported Python versions are informally documented in #656 and formally documented in `pyproject.toml`, which is what pip actually uses. I don't think we are important enough to be listed on the SPEC 0 webpage 🙂 

We don't _perfectly_ follow SPEC 0, as PyTorch usually takes a few months to catch up and release wheels for new Python releases. So our actual support schedule is more like "the 3 most recent Python versions that PyTorch supports".

We try to advertise these kinds of changes at the front of the release notes. Of course, there are no release notes for 0.8 yet; if you are using the main branch, some things are expected to break. Note that the next release is likely to be 1.0 and will include a very large number of backwards-incompatible changes, especially for GeoDataset and GeoSampler. We will likely make one or two release candidates for 1.0 so people can test them first.

Did you have any problems migrating to Python 3.11+? Or were you just surprised by the change? Want to know if we're moving too fast or it's just a documentation problem.

fmigneault:
So far, we where trying to use the UNet integration, which needs to explore the `torchgeo==0.8.0.dev0` pending changes. When doing so, the Python >=3.11 requirement clashed with our own package's Python >=3.10. Unless the UNet or other related tools actually did need Python 3.11 features justifying their update, that did come out as a unrelated problem. That imposes an unnecessary restrictions on versions we provide, notably given that the UNet is only exploration/demo script work for the moment, and it shouldn't impact anything else in the code base. Our basic exploration turned into a more complex dependency maintenance issue unexpectedly. 

That being said, I totally understand that `torchgeo` moves fasts and integrates a lot of features, so it is not that clean cut and easy to maintain backward support while updating requirements that need to evolve at some point. I think all the versions are correctly documented and easily findable in `pyproject.toml` for __*released versions*__. The problem is if we don't happen to check what is in `main` in the meantime until 0.8 (or 1.0) is released, supporting newer models or datasets will require more engineering work than "simply updating `torchgeo`", since it comes with all its downstream requirements that can conflict with ours. 

What seems to be lacking is the visibility/plan that the https://scientific-python.org/specs/spec-0000/#support-window or https://devguide.python.org/versions/#status-of-python-versions charts provide, with respect to `torchgeo` specifically. These allow us to better plan ahead which packages must be ready/updated for a given date/release so everything aligns, rather than going by trial/error at the moment next versions are leased. Therefore, the updates are not "moving too fast" per se, but it is just hard to guess when they will actually happen or what is impacted by them ahead of time (notably python version, or other core dependencies like numpy, torch, etc.).

If the release cycle is fixed at certain dates (with some approximate leeway for pytorch/python alignment), documenting those (directly in SPEC 0 for easy compare, or separately on https://torchgeo.readthedocs.io/) would help us mitigate these situations.

Thanks for looking into our concerns! Greatly appreciated.

adamjstewart:
> When doing so, the Python >=3.11 requirement clashed with our own package's Python >=3.10.

Can you clarify this? Did you mean <=3.10? I don't see a conflict here. 3.13 satisfies both conditions.

I would definitely like to more clearly document our typical release cycle (minor release every 6 months, patch release every 1+ month as needed). Maybe 1.0 will be a good opportunity to clarify our stability.

fmigneault:
When running installers with `>=3.10`, if the actually installed Python that performs the `pip install` or similar is 3.10, the resulting `>=3.11` defined by `torchgeo==0.8.0.dev0` generates a conflict. The installer will then indicate that the active project with `>=3.10` (which "should" be valid) contradicts the downstream requirement of `torchgeo` dependency.

adamjstewart:
> We try to advertise these kinds of changes at the front of the release notes. Of course, there are no release notes for 0.8 yet

Just realized this was actually in 0.7, and it was mentioned in the 0.7 release notes.

PR #2560, opened 2025-02-03
Trainers: denormalize images before plotting

### LINKED ISSUE #1263: Dataset plotting: normalization

adamjstewart:
Closes #1263?
Closes #2557

This should (hopefully) resolve the issue where our normalized images cannot be plotted because they are in a different range than the original dataset.

### DIFF
--- a/torchgeo/trainers/classification.py
+++ b/torchgeo/trainers/classification.py
@@ -6,6 +6,7 @@
 import os
 from typing import Any
 
+import kornia.augmentation as K
 import matplotlib.pyplot as plt
 import timm
 import torch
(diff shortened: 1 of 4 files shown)


isaaccorley (approved):

isaaccorley:
Lgtm

PR #2564, opened 2025-02-05
MODIS dataset implementation for TorchGeo

nickconst1999:
This pull request adds a new remote MODIS dataset implementation to TorchGeo. This new dataset class enables users to access and process MODIS data on–the–fly directly from the Microsoft Planetary Computer via the STAC API, without downloading the data locally.
The dataset queries the Microsoft Planetary Computer’s STAC API using a provided item ID and retrieves asset URLs for the desired MODIS product. This eliminates the need for pre-downloaded local files.
A configurable **PRODUCTS** dictionary defines several MODIS products (e.g., Snow Cover, Burned Area, Land Surface Temperature, Vegetation Indices, etc.), including their associated collection IDs, preferred asset keys, and bands. The dataset loads and combines multiple bands into a single xarray DataArray, ensuring consistency in data shape.
Furthermore, the implementation is designed as a subclass of TorchGeo’s RasterDataset, allows seamless integration into the TorchGeo ecosystem. This makes it easier for users to incorporate MODIS data into geospatial ML pipelines.

### DIFF
--- a/Torchgeo_MODIS.py
+++ b/Torchgeo_MODIS.py
@@ -1,0 +1,352 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import torch
+from torchgeo.datasets import RasterDataset
+import rioxarray
+import xarray as xr
+import matplotlib.pyplot as plt
+import pystac_client
+import planetary_computer
+from rioxarray.exceptions import NoDataInBounds
+from typing import Optional
+
+
+PRODUCTS = {
+    "10A1": {
+        "collection": "modis-10A1-061",
+        "preferred_asset_keys": [
+            "NDSI_Snow_Cover",
+            "Snow_Albedo_Daily_Tile",
+            "NDSI_Snow_Cover_Basic_QA",
+            "NDSI_Snow_Cover_Algorithm_Flags_QA",
+        ],
+        "name": "Snow Cover Daily",
+        "bands": [
+            "NDSI_Snow_Cover",
+            "Snow_Albedo_Daily_Tile",
+            "NDSI_Snow_Cover_Basic_QA",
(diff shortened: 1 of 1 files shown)


adamjstewart (changes requested):
1. File should be moved to `torchgeo/datasets/modis.py`, see https://torchgeo.readthedocs.io/en/latest/user/contributing.html#datasets
2. It doesn't look like this dataset is compatible with our other GeoDatasets, it needs to support automatic reprojection
3. It doesn't look like this dataset is compatible with our GeoSamplers, it needs to support bounding box queries
4. You should not need to override `__init__`, `__len__`, or `__getitem__`, the defaults from `RasterDataset` should be fine, see https://torchgeo.readthedocs.io/en/latest/tutorials/custom_raster_dataset.html
5. We likely can't justify adding dependencies on xarray, rioxarray, pystac_client, and planetary_computer
6. Need to add documentation
7. Need to add unit tests

adamjstewart:
@nickconst1999 is this still a work-in-progress? We're planning a 0.7.0 release in the next week or so.

adamjstewart:
Sure, happy to review again when you're done. I'll keep this PR in mind but won't hold up the release for it. It can always go into the next minor release if you miss this deadline.

adamjstewart:
Note that TorchGeo now has an XarrayDataset base class that could be useful here. Next step is to add HDF4 support to xarray.

adamjstewart:
Same question, is this still a work in progress or should we close?

nickconst1999:
Hello professor! Forgive me for making you wait. It can be closed 

Yahoo Mail: Căutare, organizare, reușită 
 
  Pe joi, nov. 13, 2025 la 11:44, Adam J. ***@***.***> a scris:   adamjstewart left a comment (torchgeo/torchgeo#2564)
Same question, is this still a work in progress or should we close?

PR #2566, opened 2025-02-07
Fix typos

giswqs:
This PR fixes several typos detected by [codespell](https://github.com/codespell-project/codespell).


isaaccorley (approved):

adamjstewart (approved):
Thanks again! Unfortunately ruff doesn't yet have any spell checkers, so we probably won't integrate this into CI, but it's probably a good idea to run every few months.

PR #2572, opened 2025-02-10
Add DIOR dataset

nilsleh:
This PR adds the [DIOR dataset](https://arxiv.org/abs/1909.00133). License found [here](https://gcheng-nwpu.github.io/). Dataset rehosted on Huggingface based on this [google drive link](https://drive.google.com/drive/folders/1UdlgHk49iu6WpcJ5467iT-UqNPpx__CC).

Dataset features:

    * 20 classes
    * 192,472 manually annotated bounding box instances

Dataset format:

    * Images are three channel .jpg files.
    * Annotations are in xml format

dior_0


docs/api/datasets/non_geo_datasets.csv
```diff
@@ -13,6 +13,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `Kenya Crop Type`_,S,Sentinel-2,"CC-BY-SA-4.0","4,688",7,"3,035x2,016",10,MSI
 `DeepGlobe Land Cover`_,S,DigitalGlobe +Vivid,-,803,7,"2,448x2,448",0.5,RGB
 `DFC2022`_,S,Aerial,"CC-BY-4.0","3,981",15,"2,000x2,000",0.5,RGB
+`DIOR`_,OD,Aerial,"CC-BY-SA","23,463",20,"800x800",0.5,RGB
```
adamjstewart, line 16: Wish we knew _which_ CC-BY-SA, without a version number it isn't a valid [SPDX identifier](https://spdx.org/licenses/).

torchgeo/datasets/dior.py
```diff
@@ -0,0 +1,392 @@
...
+    16. Tennis Court
+    17. Train Station
+    18. Vehicle
+    19. Windmill
+
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://arxiv.org/abs/1909.00133
+
```
adamjstewart, line 106: Should we mention that pyarrow is required? I'm noticing this dependency in a lot of your PRs. Is this file coming from the dataset authors or from you? I would like to avoid additional dependencies unless absolutely necessary.

torchgeo/datasets/dior.py
```diff
@@ -0,0 +1,392 @@
...
+        """Load the target mask for a single image.
+
+        Args:
+            path: path to the annotation file
+
+        Returns:
+            the target bounding boxes and labels
+        """
+        parsed = parse_pascal_voc(path)
+        boxes = torch.tensor(parsed['bboxes'], dtype=torch.float32)
```
adamjstewart, line 258: Are the boxes pixel-based? Does float make sense here?

torchgeo/datasets/dior.py
```diff
@@ -0,0 +1,392 @@
...
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://arxiv.org/abs/1909.00133
+
+
+    .. versionadded:: 0.7
+    """
+
+    url = 'https://huggingface.co/datasets/torchgeo/dior/resolve/main/{}'
```
adamjstewart, line 111:
```suggestion
    url = 'https://hf.co/datasets/torchgeo/dior/resolve/ec7be9567d2e08eb3d3401c15a52ee2145d0ef01/{}'
```
Need a stable download URL

adamjstewart (changes requested):

docs/api/datasets/non_geo_datasets.csv
adamjstewart: @gcheng-nwpu do you know?

P.S. We are adding TorchGeo data loaders for your excellent DIOR and SODA-A datasets. Hopefully this will make it even easier for people to use your datasets and cite your papers!

adamjstewart: @jbwang1997 may also know

torchgeo/datasets/dior.py
nilsleh: This file is coming from me. Part of https://github.com/microsoft/torchgeo/issues/2448?email_source=slack

torchgeo/datasets/dior.py
nilsleh: Yeah they are float values

torchgeo/datasets/dior.py
adamjstewart: I haven't yet managed to reproduce that issue, so I'm not sure if we should force people to install extra dependencies just to avoid it.

nilsleh: Alright, Ill change it to csv, parquet is more performant so today's standard, but that will remove the dependency. Also forcing sounds like a strong word, I don't think that's preventing anyone from using this dataset, and a lot of the new datasets that are generally well done do come with this, like MMEarth, BigEarthNetV2, etc.

adamjstewart: As it becomes more standard, we may start converting or supporting parquet more directly. I have similar feelings about WebDataset. I am really trying to minimize how many dependencies need to be installed to use TorchGeo, but we also want the datasets to be fast enough.

isaaccorley: Parquet and GeoParquet has significantly better compression for large vector datasets so I think including it as a dependency is ideal.

adamjstewart: Yes, but WebDataset is also better for I/O on HPC systems, mmrotate is better for object detection, etc. We need to decide a core set of dependencies we can't live without (required), then add optional dependencies sparingly. If someone wants to subclass and extend these datasets to be faster, they can (EarthNets is doing just that).

Obviously, this isn't a fair comparison. We avoid dependencies on certain libraries due to licensing, security, or difficulty of installation. My concern is less which dependencies we add and more how many dependencies we add. My priorities are roughly:

1. Ease of installation: rules out things like mmrotate
2. Licensing complexity: rules out things like YOLO
3. Performance: in favor of things like WebDataset, parquet
4. Security: rules out things like YOLO

Basically, performance only matters to me when the difference is big enough that someone would no longer want to use our implementation. If that's not the case, let's keep it simple.

torchgeo/datasets/dior.py
```diff
@@ -0,0 +1,390 @@
...
+        bboxes.append(bbox)
+        labels.append(label)
+
+    return dict(filename=filename, bboxes=bboxes, labels=labels)
+
+
+class DIOR(NonGeoDataset):
+    """DIOR dataset.
+
+    `DIOR <https://arxiv.org/abs/1909.00133>`_ dataset contains horizontal bounding box
```
adamjstewart, line 63:
```suggestion
    `DIOR <https://arxiv.org/abs/1909.00133>`__ dataset contains horizontal bounding box
```

torchgeo/datasets/dior.py
```diff
@@ -0,0 +1,390 @@
...
+    annotations of Google Earth Aerial RGB imagery. The test split does not contain bounding
+    box annotations and labels
```
adamjstewart, line 65:
```suggestion
    box annotations and labels.
```

torchgeo/datasets/dior.py
```diff
@@ -0,0 +1,390 @@
...
+        'tenniscourt',
+        'trainstation',
+        'vehicle',
+        'windmill',
+    )
+
+    def __init__(
+        self,
+        root: Path = 'data',
+        split: str = 'train',
```
adamjstewart, line 160:
```suggestion
        split: Literal['train', 'val', 'test'] = 'train',
```
Trying to push use of `typing.Literal` in more cases

torchgeo/datasets/dior.py
```diff
@@ -0,0 +1,390 @@
...
+            root: root directory where dataset can be found
+            split: split of the dataset to use, one of 'train', 'val', 'test'
+            transforms: a function/transform that takes input sample and its target as
+                entry and returns a transformed version
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+
+        Raises:
+            DatasetNotFoundError: If dataset is not found and *download* is False.
+            AssertionError: If *split* argumnet is invalid.
```
adamjstewart, line 177:
```suggestion
            AssertionError: If *split* argument is invalid.
```

torchgeo/datasets/dior.py
```diff
@@ -0,0 +1,390 @@
...
+            path: path to the image
+
+        Returns:
+            the image
+        """
+        with Image.open(path) as img:
+            array: np.typing.NDArray[np.int_] = np.array(img.convert('RGB'))
+            tensor: Tensor = torch.from_numpy(array)
+            # Convert from HxWxC to CxHxW
+            tensor = tensor.permute((2, 0, 1))
```
adamjstewart, line 245: Could use einops.rearrange here to make this easier to read

torchgeo/datasets/dior.py
```diff
@@ -0,0 +1,390 @@
...
+        if all(exists):
+            return
+
+        if not self.download:
+            raise DatasetNotFoundError(self)
+
+        self._download()
+
+    def _download(self) -> None:
+        """Downlaod the dataset and extract it."""
```
adamjstewart, line 310:
```suggestion
        """Download the dataset and extract it."""
```

adamjstewart (commented):
Just minor documentation things remaining

adamjstewart (approved):

PR #2583, opened 2025-02-17
Add EverWatch dataset

nilsleh:
This PR adds the [EverWatch](https://zenodo.org/records/11165946) dataset for bird species object detection.

Dataset features:

    * 5128 training images with 50491 annotations
    * 197 test images with 4113 annotations
    * seven different bird species

Dataset format:

    * images are three-channel pngs
    * annotations are csv file

ever_watch


docs/api/datasets/non_geo_datasets.csv
```diff
@@ -16,6 +16,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `Digital Typhoon`_,"C, R",Himawari,"CC-BY-4.0","189,364",8,512,5000,Infrared
 `ETCI2021 Flood Detection`_,S,Sentinel-1,-,"66,810",2,256x256,5--20,SAR
 `EuroSAT`_,C,Sentinel-2,"MIT","27,000",10,64x64,10,MSI
+`EverWatch`_,OD,Aerial,"CC0 1.0 Universal","5,325",7,"1,500x1500p",0.01,RGB
```
adamjstewart, line 19:
```suggestion
`EverWatch`_,OD,Aerial,CC0-1.0,"5,325",7,"1,500x1500p",0.01,RGB
```
Names should match [SPDX identifiers](https://spdx.org/licenses/). Wish we had an easy way to test this, but there are also a lot of custom licenses.

Anyways, confirmed the license from https://zenodo.org/records/11165946

torchgeo/datasets/everwatch.py
```diff
@@ -0,0 +1,291 @@
...
+        self.checksum = checksum
+        self.download = download
+
+        self._verify()
+
+        self.annot_df = pd.read_csv(
+            os.path.join(self.root, self.dir, f'{self.split}.csv')
+        )
+
+        # remove all entries where xmin == xmax or ymin == ymax
```
adamjstewart, line 116: How common is this? If you visualize them, are they false positives or just very small birds?

torchgeo/datasets/everwatch.py
```diff
@@ -0,0 +1,291 @@
...
+
+        Args:
+            sample_df: df subset with annotations for specific image
+
+        Returns:
+            bounding boxes and labels
+        """
+        boxes = torch.Tensor(
+            sample_df[['xmin', 'ymin', 'xmax', 'ymax']].values.tolist()
+        ).float()
```
adamjstewart, line 187: If I understand this line of code correctly, you convert from a pd.Series to a np.ndarray to a list to a torch.Tensor. I wonder if we can make this shorter, maybe skip the list?

adamjstewart (commented):
Very interesting datasets, excited to have it!

torchgeo/datasets/everwatch.py
nilsleh: It's 1000 annotations: before the filter there are a total of 51491 annotations and afterwards 50491. The bounding boxes won't make sense since they'd just be a horizontal or vertical line.

adamjstewart (approved):

PR #2601, opened 2025-02-21
GeoDatasets should use separate values for x and y resolution

calebrob6:
This PR refactors GeoDatasets to use separate values to store the x and y resolutions. This is pretty important because not all raster pixels are square.

Some notes:
- rasterio returns res in (width, height) format -- https://rasterio.readthedocs.io/en/stable/api/rasterio.io.html#rasterio.io.DatasetReader.res
- We expect `size` in (height, width) format for the samplers

Closes https://github.com/microsoft/torchgeo/issues/2594


torchgeo/datasets/geo.py
```diff
@@ -416,7 +416,7 @@ def __init__(
         self,
         paths: Path | Iterable[Path] = 'data',
         crs: CRS | None = None,
-        res: float | None = None,
+        res: tuple[float, float] | None = None,
```
adamjstewart, line 419: If possible, I would like all of these datasets to support both `float` and `tuple[float, float]` similar to how our Samplers work. We can use the same helper function to convert the former to the latter. This is also nice for backwards compatibility (which we have to start caring about soon).

adamjstewart: Also need to update docstrings to document the order of `xres` and `yres`.

calebrob6: Which docstrings would you like this in?

tests/datasets/test_geo.py
```diff
@@ -375,20 +376,31 @@ def test_no_all_bands(self) -> None:
...
+    def test_single_res(self) -> None:
+        root = os.path.join('tests', 'data', 'raster', 'res_2-2_epsg_32631')
+        ds = RasterDataset(root, res=10.0)
+        assert ds.res == (10.0, 10.0)
+
+    def test_change_res(self) -> None:
+        root = os.path.join('tests', 'data', 'raster', 'res_2-2_epsg_32631')
+        ds = RasterDataset(root, res=10.0)
+        assert ds.res == (10.0, 10.0)
+        ds.res = 20.0  # type: ignore[assignment]
```
calebrob6, line 388: As far as I can tell, mypy doesn't like it when your getter returns a single type but your setter accepts a union of types, maybe discussed here https://github.com/python/mypy/issues/3004

calebrob6:
The main ideas here are:
- GeoDataset's `_res` should always be a `tuple[float, float]`
- The setter for `res` handles `float | tuple[float, float]`

torchgeo/datasets/geo.py
adamjstewart: Every single docstring for every single GeoDataset with a `res` parameter.

torchgeo/datasets/globbiomass.py
```diff
@@ -139,7 +139,7 @@ def __init__(
         self,
         paths: Path | Iterable[Path] = 'data',
         crs: CRS | None = None,
-        res: float | None = None,
+        res: tuple[float, float] | None = None,
```
adamjstewart, line 142: All of these other datasets also need to be `float | tuple[float, float]` and need updated docstrings too.

adamjstewart (changes requested):
Idea is solid, just need to uniformly update the type hints and docstrings so that all datasets and samplers can handle `float | tuple[float, float]` and the order is documented. Should just be able to copy-n-paste things across all files.

torchgeo/datasets/sentinel.py
```diff
@@ -359,7 +359,7 @@ def __init__(
         """
         bands = bands or self.all_bands
         self.filename_glob = self.filename_glob.format(bands[0])
-        self.filename_regex = self.filename_regex.format(res)
+        self.filename_regex = self.filename_regex.format(int(res[0]))
```
calebrob6, line 362: Here res should really be _only_ a float

tests/datasets/test_kakuma_aerial.py
adamjstewart, line 1: Looks like this file accidentally slipped in

torchgeo/datasets/openbuildings.py
```diff
@@ -231,9 +232,10 @@ def __init__(
            *root* was renamed to *paths*.
         """
         self.paths = paths
+        if isinstance(res, float):
+            res = (res, res)
```
adamjstewart, line 236: We also have a `torchgeo.samplers.utils._to_tuple` we could be using more. Maybe we should move that to `torchgeo.utils` or something to avoid circular imports. No action required here, just a thought about future code reuse.

torchgeo/datasets/sentinel.py
```diff
@@ -359,8 +361,10 @@ def __init__(
         """
         bands = bands or self.all_bands
         self.filename_glob = self.filename_glob.format(bands[0])
-        self.filename_regex = self.filename_regex.format(res)
-
+        if isinstance(res, float):
+            self.filename_regex = self.filename_regex.format(int(res))
+        else:
+            self.filename_regex = self.filename_regex.format(int(res[0]))
```
adamjstewart, line 367: Might be cleaner to convert res, then we don't have to duplicate the whole regex thing twice.

adamjstewart (commented):
Looks good now (other than the random file that slipped in). Glad this wasn't as much work as we expected. I'm sure other minor bugs will crop up as people start to use this feature more: (xres, yres) vs. (yres, xres) or something like that. But let's get it merged first, it's too difficult to test every single corner of TorchGeo with non-square pixels.

torchgeo/datasets/sentinel.py
calebrob6: One concern I have here is that if someone passes in a different res parameter to Sentinel2 then nothing will work because it won't find the data. That's a different problem though :)

calebrob6:
CI is failing because of https://github.com/pytorch/pytorch/issues/150492

image

adamjstewart (approved):

requirements/min-reqs.old
```diff
@@ -12,7 +12,7 @@ numpy==1.23.2
 pandas==1.5.0
 pillow==9.2.0
 pyproj==3.4.0
-rasterio==1.3.3
+rasterio==1.3.11
```
adamjstewart, line 15: Somehow this slipped under my radar. Was there a reason we needed to change the minimum support rasterio version? We forgot to change this in `pyproject.toml`, which is what pip actually uses.

adamjstewart: Fixed in #2969

PR #2602, opened 2025-02-21
ObjectDetectionTask: support for non-RGB imagery (SAR, MSI, HSI)

### LINKED ISSUE #1156: ObjectDetectionTask support for MSI
### Summary

The ObjectDetectionTask has an in_channels parameter but it isn't actually used for anything. At the moment, it seems that the trainer only supports RGB imagery. We should fix this.

### Rationale

We specialize in MSI, how can we not support MSI.

### Implementation

We're currently using torchvision backbones, which makes things more challenging. How hard would it be to switch to timm backbones?

### Alternatives

Alternatively, we'll have to override the first channel of the torchvision backbone.

### Additional information

_No response_

adamjstewart:
We don't yet have a non-RGB object detection dataset to test with, but I'm pretty sure this works. Needed to bump some deps and throw away some tests to get it to work.

Closes #1156

### DIFF
--- a/torchgeo/trainers/detection.py
+++ b/torchgeo/trainers/detection.py
@@ -10,7 +10,7 @@ import matplotlib.pyplot as plt
 import torch
 import torchvision.models.detection
 from matplotlib.figure import Figure
-from timm.models import adapt_input_conv  # type: ignore[attr-defined]
+from timm.models import adapt_input_conv
 from torch import Tensor
 from torch.nn.parameter import Parameter
 from torchmetrics import MetricCollection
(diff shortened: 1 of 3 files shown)


torchgeo/trainers/detection.py
```diff
@@ -204,6 +207,9 @@ def configure_models(self) -> None:
         else:
             raise ValueError(f"Model type '{model}' is not valid.")
 
+        weight = adapt_input_conv(in_channels, self.model.backbone.body.conv1.weight)
+        self.model.backbone.body.conv1.weight = Parameter(weight)
```
isaaccorley, line 211: We should also modify the `in_channels` attribute of `model.backbone.body.conv1` because I just did this manually and even if you modify the weight it will stay `in_channels=3`. This won't show up in tests but could a silent bug later. 

```suggestion
        self.model.backbone.body.conv1.weight = Parameter(weight)
        self.model.backbone.body.conv1.in_channels = in_channels
```

isaaccorley (changes requested):
Minor suggestion but otherwise lgtm

torchgeo/trainers/detection.py
adamjstewart: Just made this same change for instance segmentation too, thanks for catching this

isaaccorley (approved):

PR #2603, opened 2025-02-23
VHR-10 plot fix

ashnair1:
Previously if split was not 'negative', unnormalized image was shown. Had accidently introduced this bug in #1978 

Before:

vhr10-plot-bug

After:

vhr10-plot-fix


adamjstewart (approved):
Good catch!

PR #2604, opened 2025-02-23
Copernicus-Bench: Level-1 datasets

wangyi111:
This PR adds a new Copernicus-Bench dataset our lab has been working on. So far, this PR only includes Level-1 datasets (cloud masking); other datasets will be added in follow-up PRs. The purpose of this PR is to decide on a standard format for benchmark suites that can also be followed by other datasets, like GEO-Bench.

cloud_s2
cloud_s3_binary
cloud_s3_multi

### References

* https://github.com/zhu-xlab/Copernicus-FM/tree/main/Copernicus-Bench/src/datasets
* https://huggingface.co/datasets/wangyi111/Copernicus-Bench
* https://arxiv.org/abs/2503.11849


wangyi111:
@adamjstewart

adamjstewart:
My first thoughts on seeing these:

* Each dataset itself is a lot of code, so it makes sense to have each one in a separate file
* We may want some kind of base class containing shared code (e.g., download, extract, len)

So as to avoid `torchgeo/datasets/` becoming too cluttered, I would propose a `torchgeo/datasets/senbench/` directory. This could look like:
```
torchgeo/datasets/
    __init__.py: from senbench import *
    senbench/
        __init__.py: from * import *, wrapper dataset
        base.py: shared base class if necessary
        cloud_s3.py: class SenBenchCloudS3
        lulc_s3.py: class SenBenchLC100SegS3
        ...
```
Here, we wouldn't actually import `import *`, we would import each individual class from each file.

This will be our first big meta dataset/benchmarking suite. Since it's the first, we have more flexibility, but also more decisions to make. Let's mull this over for a bit and I'll consult the other TorchGeo devs.

adamjstewart:
Could also organize subfiles by level or sensor. Need to look at the full table again to think about how to organize.

wangyi111:
> My first thoughts on seeing these:
> 
> * Each dataset itself is a lot of code, so it makes sense to have each one in a separate file
> * We may want some kind of base class containing shared code (e.g., download, extract, len)
> 
> So as to avoid `torchgeo/datasets/` becoming too cluttered, I would propose a `torchgeo/datasets/senbench/` directory. This could look like:
> 
> ```
> torchgeo/datasets/
>     __init__.py: from senbench import *
>     senbench/
>         __init__.py: from * import *, wrapper dataset
>         base.py: shared base class if necessary
>         cloud_s3.py: class SenBenchCloudS3
>         lulc_s3.py: class SenBenchLC100SegS3
>         ...
> ```
> 
> Here, we wouldn't actually import `import *`, we would import each individual class from each file.
> 
> This will be our first big meta dataset/benchmarking suite. Since it's the first, we have more flexibility, but also more decisions to make. Let's mull this over for a bit and I'll consult the other TorchGeo devs.

this sounds a good idea!

torchgeo/datasets/copernicus/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+            index: Index to return.
+
+        Returns:
+            Data and labels at that index.
+        """
+        return self.dataset[index]
+
+    def __getattr__(self, name: str) -> Any:
+        """Wrapper around dataset object."""
+        return getattr(self.dataset, name)
```
adamjstewart, line 61: Trying to upstream this idea: https://github.com/pytorch/pytorch/pull/149272

torchgeo/datasets/copernicus/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+from torch import Tensor
+
+from ..geo import NonGeoDataset
+from .base import CopernicusBenchBase
+from .cloud_s2 import CopernicusBenchCloudS2
+
+DATASET_REGISTRY = {'cloud_s2': CopernicusBenchCloudS2}
+
+
+class CopernicusBench(NonGeoDataset):
```
adamjstewart, line 17: It's actually not clear to me if a dataset wrapper is necessary. We could replace this whole thing with a function. Or just drop it altogether. What purpose does it serve?

torchgeo/datasets/copernicus/base.py
```diff
@@ -0,0 +1,249 @@
...
+    def _extract(self) -> None:
+        """Extract the dataset."""
+        extract_archive(os.path.join(self.root, self.zipfile))
+
+    def _download(self) -> None:
+        """Download the dataset."""
+        md5 = self.md5 if self.checksum else None
+        download_and_extract_archive(self.url, self.root, md5=md5)
+
+    def plot(
```
adamjstewart, line 184: I would actually like to move this idea to NonGeoDataset and GeoDataset in the future. 3/4 of our custom plotting methods could be removed by assuming that `all_bands`, `rgb_bands`, and `cmap` are defined as class attributes, using percentile normalization, and assuming standard sample key names. Somewhat relies on #985 though.

torchgeo/datasets/copernicus/cloud_s3.py
```diff
@@ -0,0 +1,173 @@
...
+
+    .. list-table:: Binary Classification
+       :header-rows: 1
+
+       * - Code
+         - Class
+         - Description
+       * - 255
+         - Invalid
+         - Invalid pixels, should be ignored during training.
```
adamjstewart, line 62: This may or may not work with our `task='binary'` added in #2219, need to figure out how to `ignore_index` with `nn.BCEWithLogitsLoss`.

torchgeo/datasets/copernicus/cloud_s2.py
```diff
@@ -0,0 +1,107 @@
...
+        'B07',
+        'B08',
+        'B8A',
+        'B09',
+        'B10',
+        'B11',
+        'B12',
+    )
+    rgb_bands = ('B04', 'B03', 'B02')
+    cmap = ListedColormap(['white', 'yellow', 'green', 'red'])
```
adamjstewart, line 74: Chose the colormap used by the CloudSEN12+ dataset authors in Fig. 7: https://doi.org/10.1016/j.dib.2024.110852. Could change it if we want to.

torchgeo/datasets/copernicus/cloud_s3.py
```diff
@@ -0,0 +1,178 @@
...
+    rgb_bands = ('Oa08_radiance', 'Oa06_radiance', 'Oa04_radiance')
+    cmap = ListedColormap(['red', 'gray', 'white', 'lightgray', 'black', 'snow'])
+    classes: tuple[str, ...] = (
+        'Invalid',
+        'Clear',
+        'Cloud-Sure',
+        'Cloud-Ambiguous',
+        'Cloud Shadow',
+        'Snow/Ice',
+    )
```
adamjstewart, line 114: @wangyi111 can you double check that this order is correct? The example plot I attached to your PR doesn't look right.

torchgeo/datasets/copernicus/cloud_s2.py
adamjstewart, line 1: Alternatively, we could have datasets grouped, like `level1.py` or `sentinel1.py`.

torchgeo/datasets/copernicus/__init__.py
wangyi111: As a function should be good also. just thought it might be nice to have something that wraps all datasets in the bench together.

torchgeo/datasets/copernicus/cloud_s3.py
wangyi111: yes it's correct.
- 08: 665nm (r)
- 06: 560nm (g)
- 04: 490nm (b)
maybe the cloud (bright pixel) makes the color not very natural. 

also S3 OLCI is radiance instead of reflectance, and the band values should multiply a scale factor as in https://developers.google.com/earth-engine/datasets/catalog/COPERNICUS_S3_OLCI#bands (different for each band).

adamjstewart: I mean the class order, not the bands order.

torchgeo/datasets/copernicus/__init__.py
adamjstewart: Let's see what @nilsleh thinks, I'm curious how GEO-Bench organizes their datasets. I wish PyTorch Lightning had some kind of grouped LightningDataModule concept.

P.S. We should probably add datamodules for Copernicus-Bench someday to include the transforms you are using. Doesn't necessarily have to be in this first batch of PRs, could be after the release if we run out of time.

torchgeo/datasets/copernicus/cloud_s3.py
wangyi111: ah, i will have to check my raw files on that.

wangyi111: This is my original code to generate the mask.
```
cloud_mask_multi = np.ones(cloud[0].shape, dtype='uint8') # clean
cloud_mask_multi[invalid_flag == 1] = 0
cloud_mask_multi[cloud_ambiguous_flag == 1] = 2
cloud_mask_multi[cloud_sure_flag == 1] = 3
cloud_mask_multi[cloud_shadow_flag == 1] = 4
cloud_mask_multi[snow_ice_flag == 1] = 5
```

Seems 'Cloud-Ambiguous' and 'Cloud-Sure' should be swapped. Others should be good.

torchgeo/datasets/copernicus/base.py
```diff
@@ -0,0 +1,259 @@
...
+from ..errors import DatasetNotFoundError, RGBBandsMissingError
+from ..utils import (
+    Path,
+    download_and_extract_archive,
+    extract_archive,
+    percentile_normalization,
+)
+
+
+class CopernicusBenchBase(NonGeoDataset, ABC):
```
calebrob6, line 29: This abstract structure seems very general and not specific to CopernicusBench datasets. Are we planning to implement other CopernicusBench tasks?

torchgeo/datasets/copernicus/cloud_s3.py
```diff
@@ -0,0 +1,178 @@
...
+            index: Index to return.
+
+        Returns:
+            A target sample.
+        """
+        sample: dict[str, Tensor] = {}
+        file = self.files[index]
+        mode = f'cloud_{self.mode}'
+        with rio.open(os.path.join(self.root, self.directory, mode, file)) as f:
+            sample['mask'] = torch.tensor(f.read(1).astype(np.int64))
```
calebrob6, line 176:
```suggestion
            sample['mask'] = torch.from_numpy(f.read(1)).long()
```

calebrob6 (approved):

torchgeo/datasets/copernicus/base.py
adamjstewart: Yes, see Table 3 in https://arxiv.org/abs/2503.11849 for the full list. There are classification, regression, and change detection datasets still coming for LU/LC, flood mapping, and air pollution mapping.

torchgeo/datasets/copernicus/cloud_s3.py
adamjstewart: This is actually what I used originally but it doesn't work for older PyTorch (would have to check now that we bumped the minimum PyTorch version). See https://github.com/microsoft/torchgeo/pull/2604/commits/399d0ae004a083fe89acd2d48ea849d7257d5dff for the change and https://github.com/microsoft/torchgeo/actions/runs/13882120290/job/38842005656 for the failing tests. Basically has to do with PyTorch support for uint16.

torchgeo/datasets/copernicus/__init__.py
adamjstewart: Apparently GEO-Bench also uses a wrapper class.

adamjstewart (approved):

PR #2607, opened 2025-02-23
Copernicus-Bench: Add Level-3 Sentinel-5P

wangyi111:
Adds the following Level-3 datasets from [Copernicus-Bench](https://arxiv.org/abs/2503.11849):

- [x] AQ-NO2-S5P
- [x] AQ-O3-S5P

aq_no2_s5p
aq_o3_s5p


torchgeo/datasets/copernicus/aq_no2_s5p.py
```diff
@@ -0,0 +1,108 @@
...
+from typing import Literal
+
+import torch
+from torch import Tensor
+
+from ..utils import Path, stack_samples
+from .base import CopernicusBenchBase
+
+
+class CopernicusBenchAQNO2S5P(CopernicusBenchBase):
```
adamjstewart, line 17: Could make a shared CopernicusBenchAQS5P base class, but then we need to document it as well.

wangyi111: This could be useful when we have more variables and they can be aligned, could be in further updates of Copercnius-Bench.

adamjstewart (approved):

PR #2614, opened 2025-02-28
Trainers: support nn.Module as input

adamjstewart:
Want to use a model that does not yet exist in TorchGeo? Now you can pass it directly to the trainer class.

- [x] Code
- [ ] Testing

### DIFF
--- a/tests/trainers/test_byol.py
+++ b/tests/trainers/test_byol.py
@@ -76,6 +76,10 @@ class TestBYOLTask:
 
         main(['fit', *args])
 
+    def test_custom_model(self) -> None:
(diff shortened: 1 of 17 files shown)


adamjstewart:
Unit tests have revealed that this idea doesn't actually work. We can't store `model` in `self.model` because we haven't yet called `super().__init__()`, but that calls `configure_models` which needs access to `self.models`. And we can't store `model` in `self.hparams` because it doesn't support `nn.Module`. Need to think of a clever way to break this dependency loop...

adamjstewart:
Never got this working, giving up for now.

PR #2630, opened 2025-03-05
Remove default std normalization value in base class

### LINKED ISSUE #1841: Datamodule augmentation defaults
### Summary

Currently the datamodules divide by 255 by default. This can confuse users who would expect no augmentations by default. We should make this WARNING level clear or not divide by default.

### Rationale

_No response_

### Implementation

_No response_

### Alternatives

_No response_

### Additional information

_No response_

nilsleh:
While not technically a bug, I think this is quiet inconvenient, when building custom datamodules inheriting from torchgeo base classes. For my use case, I had trouble understanding why my data was all of the sudden squashed to a tiny range when I had properly normalized it beforehand.

I think the default value here should be 1, i.e nothing happens, and users have to otherwise overwrite it. At the moment it is a base class for RGB data, which is oddly specific, and I think it would be better to keep it more general, especially because it might not be clear to all users, that the normalization "secretly" happens in the on_after_batch_transfer hook.

Closes #1841

### DIFF
--- a/torchgeo/datamodules/geo.py
+++ b/torchgeo/datamodules/geo.py
@@ -30,7 +30,7 @@ class BaseDataModule(LightningDataModule):
     """
 
     mean = torch.tensor(0)
-    std = torch.tensor(255)
+    std = torch.tensor(1)
 
     def __init__(
         self,


adamjstewart (changes requested):
I agree with this change (principle of least surprise) but we should set `std` for all the datamodules that don't currently specify it. I'm fine with just using `255` in all of them to maintain backwards compatibility but it would be good to at least check `data.py` and see what dtype the files use.

calebrob6:
+1 I've hit this issue too

isaaccorley:
> I agree with this change (principle of least surprise) but we should set `std` for all the datamodules that don't currently specify it. I'm fine with just using `255` in all of them to maintain backwards compatibility but it would be good to at least check `data.py` and see what dtype the files use.

@adamjstewart Are you okay if we merge and I fix this in a separate PR? I've been meaning to go back through and validate all the dataset stats anyway.

adamjstewart:
Let's do your PR first instead

robmarkcole:
@isaaccorley there is no link to the PR you mention above, can you add?

adamjstewart:
Anyone want to revive this PR?

calebrob6 (approved):

calebrob6:
I think this should be merged, this is a separate issue than existing datamodules `std` values as it hits me when creating _new_ datamodules

adamjstewart (changes requested):
This cannot be merged as is, as it breaks most of our existing datamodules by changing the default from 255 to 1. We need to add an explicit value to each dataset to prevent this.

adamjstewart:
Opened #4014 to unblock this PR.

PR #2639, opened 2025-03-12
Add AutoregressionTask

keves1:
This PR adds the following:

- Autoregression Trainer (see #2382)
- LSTM Sequence-to-Sequence (Seq2Seq) model for time series forecasting
- Air Quality time series data set from the [UCI Machine Learning Repository](https://archive.ics.uci.edu/dataset/360/air+quality)

Together these additions enable training air quality forecasting models as well as other autoregression tasks as additional datasets and models are added.


keves1:
@adamjstewart See below, this test passes when I run it locally but it is failing here. I don't understand this error message, do you know how to resolve this?

```
FAILED tests/trainers/test_autoregression.py::TestAutoregressionTask::test_trainer[True-air_quality] - ValueError: Does not validate against any of the Union subtypes
Errors:
  - Unexpected keyword arguments: `num_outputs`
Given value type: <class 'jsonargparse._namespace.Namespace'>
Given value: Namespace(class_path='torchgeo.trainers.AutoregressionTask', init_args=Namespace(model='lstm_seq2seq', input_size=3, input_size_decoder=1, hidden_size=1, output_size=1, target_indices=[2], encoder_indices=[2, 12, 13], decoder_indices=[2], timesteps_ahead=1, num_layers=1, loss='mse', lr=0.001, patience=10, teacher_force_prob=None))
```

docs/api/datasets/non_geo_datasets.csv
```diff
@@ -1,5 +1,6 @@
 Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `ADVANCE`_,C,"Google Earth, Freesound","CC-BY-4.0","5,075",13,512x512,0.5,RGB
+`Air Quality`_,"R,T","UCI Machine Learning Repository","CC-BY-4.0","9,358",,,,
```
adamjstewart, line 3: Confirmed the license from https://archive.ics.uci.edu/dataset/360/air+quality

tests/datasets/test_air_quality.py
```diff
@@ -0,0 +1,37 @@
...
+
+from torchgeo.datasets import AirQuality, DatasetNotFoundError
+
+
+class TestAirQuality:
+    @pytest.fixture()
+    def dataset(
+        self, monkeypatch: MonkeyPatch, tmp_path: Path, request: SubRequest
+    ) -> AirQuality:
+        url = 'tests/data/air_quality/data.csv'
```
adamjstewart, line 19:
```suggestion
        url = os.path.join('tests', 'data', 'air_quality', 'data.csv')
```

torchgeo/datamodules/air_quality.py
```diff
@@ -0,0 +1,75 @@
...
+        train_size = int(train_split_pct * len(dataset))
+        val_size = int(self.val_split_pct * len(dataset))
+        train_indices = range(train_size)
+        val_indices = range(train_size, train_size + val_size)
+        test_indices = range(train_size + val_size, len(dataset))
+        self.train_dataset = Subset(dataset, train_indices)
+        self.val_dataset = Subset(dataset, val_indices)
+        self.test_dataset = Subset(dataset, test_indices)
+
+    def on_after_batch_transfer(
```
adamjstewart, line 63: Why is this needed? Is there a problem with the base class implementation?

torchgeo/datamodules/air_quality.py
```diff
@@ -0,0 +1,75 @@
...
+            batch_size: Size of each mini-batch.
+            val_split_pct: Percentage of the dataset to use as a validation set.
+            test_split_pct: Percentage of the dataset to use as a testing set.
+            num_workers: Number of workers for parallel data loading.
+            **kwargs: Additional keyword arguments passed to
+                :class:`~torchgeo.datasets.AirQuality`.
+        """
+        super().__init__(AirQuality, batch_size, num_workers, **kwargs)
+        self.val_split_pct = val_split_pct
+        self.test_split_pct = test_split_pct
```
adamjstewart, line 44: Should we add some kind of no-op data augmentation to override the base class?

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,116 @@
...
+    * has missing features
+
+    If you use this dataset in your research, please cite:
+
+    * https://doi.org/10.1016/J.SNB.2007.09.060
+
+    .. versionadded:: 0.7
+    """
+
+    url = 'https://archive.ics.uci.edu/static/public/360/data.csv'
```
adamjstewart, line 40: Out of curiosity, how did you find this URL? I only see the URL for the zip file. This file is much cleaner than the one from the zip file, which uses all kinds of weird formats.

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,116 @@
...
+    data_file_name = 'data.csv'
+
+    def __init__(
+        self,
+        root: Path = 'data',
+        download: bool = False,
```
adamjstewart, line 46: Can we also add transforms and checksum to match other datasets?

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,116 @@
...
+        pathname = os.path.join(self.root, self.data_file_name)
+        if os.path.exists(pathname):
+            return pd.read_csv(pathname)
+
+        # Check if the user requested to download the dataset
+        if not self.download:
+            raise DatasetNotFoundError(self)
+
+        # Download the dataset
+        return pd.read_csv(self.url)
```
adamjstewart, line 116:
```suggestion
        return pd.read_csv(self.url, na_values=-200)
```
May also be worth creating a formal datetime column like:
```python
data['datetime'] = pd.to_datetime(data['Date'] + ' ' + data['Time'])
```
We don't use this yet, but may want to once we add better time-series support.

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,116 @@
...
+            + self.num_past_steps
+            + self.num_future_steps
+        ]
+        past_steps = torch.tensor(past_steps.values, dtype=torch.float32)
+        future_steps = torch.tensor(future_steps.values, dtype=torch.float32)
+
+        mean = past_steps.mean(dim=0, keepdim=True)
+        std = past_steps.std(dim=0, keepdim=True)
+        past_steps_normalized = (past_steps - mean) / (std + 1e-12)
+        future_steps_normalized = (future_steps - mean) / (std + 1e-12)
```
adamjstewart, line 96: Can we leave normalization up to the datamodule?

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,116 @@
...
+
+import pandas as pd
+import torch
+
+from .errors import DatasetNotFoundError
+from .geo import NonGeoDataset
+from .utils import Path
+
+
+class AirQuality(NonGeoDataset):
```
adamjstewart, line 16: For our time-series work, we're mostly interested in GeoDataset, but this one probably makes more sense as a NonGeoDataset.

torchgeo/models/seq2seq.py
```diff
@@ -0,0 +1,192 @@
...
+            if not teacher_force:
+                if self.target_indices:
+                    current_input[:, :, self.target_indices] = output
+                else:
+                    current_input = output
+
+        return outputs
+
+
+class LSTMSeq2Seq(nn.Module):
```
adamjstewart, line 111: Let's rename this to `Seq2Seq` and add a new parameter:
```python
cell: Literal['rnn', 'gru', 'lstm'] = 'lstm'
```
That way we can easily support all 3:
```python
match cell:
    case 'rnn':
        self.rnn = nn.RNN(...)
    case 'gru':
        self.rnn = nn.GRU(...)
    case 'lstm':
        self.rnn = nn.LSTM(...)
```

torchgeo/models/seq2seq.py
```diff
@@ -0,0 +1,192 @@
...
+    """LSTM Sequence-to-Sequence (Seq2Seq)."""
+
+    def __init__(
+        self,
+        input_size_encoder: int,
+        input_size_decoder: int,
+        target_indices: list[int] | None = None,
+        encoder_indices: list[int] | None = None,
+        decoder_indices: list[int] | None = None,
```
adamjstewart, line 120: Can we support both list and tuple? Maybe `Sequence` or `Iterable`.

torchgeo/models/seq2seq.py
```diff
@@ -0,0 +1,192 @@
...
+                if self.target_indices:
+                    current_input[:, :, self.target_indices] = output
+                else:
+                    current_input = output
+
+        return outputs
+
+
+class LSTMSeq2Seq(nn.Module):
+    """LSTM Sequence-to-Sequence (Seq2Seq)."""
```
adamjstewart, line 112: Are there any specific foundational papers we could cite for this model? Who first proposed it? Or should we instead add the original citations for Elman RNNs, GRUs, and LSTMs?

torchgeo/trainers/autoregression.py
```diff
@@ -0,0 +1,201 @@
...
+
+from .base import BaseTask
+
+
+class AutoregressionTask(BaseTask):
+    """Autoregression."""
+
+    def __init__(
+        self,
+        model: str = 'lstm_seq2seq',
```
adamjstewart, line 23:
```suggestion
        model: Literal['seq2seq'] = 'seq2seq',
        cell: Literal['rnn', 'gru', 'lstm'] = 'lstm',
```
or something like that

torchgeo/models/seq2seq.py
```diff
@@ -0,0 +1,192 @@
...
+            input_size_decoder: The number of features in the decoder input.
+            target_indices: The indices of the target(s) in the dataset. If None, uses all features. Defaults to None.
+            encoder_indices: The indices of the encoder inputs. If None, uses all features. Defaults to None.
+            decoder_indices: The indices of the decoder inputs. If None, uses all features. Defaults to None.
+            hidden_size: The number of features in the hidden states of the encoder and decoder. Defaults to 1.
+            output_size: The number of features output by the model. Defaults to 1.
+            output_sequence_len: The number of steps to predict forward. Defaults to 1.
+            num_layers: Number of LSTM layers in the encoder and decoder. Defaults to 1.
+            teacher_force_prob: Probability of using teacher forcing. If None, does not
+                use teacher forcing. Defaults to None.
```
adamjstewart, line 140: Instead of documenting and passing all of these arguments one at a time, maybe we could just have an extra `**kwargs` argument that gets passed to the LSTM class?

torchgeo/trainers/autoregression.py
```diff
@@ -0,0 +1,201 @@
...
+        input_size: int = 1,
+        input_size_decoder: int = 1,
+        hidden_size: int = 1,
+        output_size: int = 1,
+        target_indices: list[int] | None = None,
+        encoder_indices: list[int] | None = None,
+        decoder_indices: list[int] | None = None,
+        timesteps_ahead: int = 1,
+        num_layers: int = 1,
+        loss: str = 'mse',
```
adamjstewart, line 33:
```suggestion
        loss: Literal['mse', 'mae'] = 'mse',
```

torchgeo/trainers/autoregression.py
```diff
@@ -0,0 +1,201 @@
...
+        return loss
+
+    def validation_step(self, batch: Any, batch_idx: int) -> None:
+        """Compute the validation loss and additional metrics.
+
+        Args:
+            batch: The output of your DataLoader.
+            batch_idx: Integer displaying index of this batch.
+        """
+        self._shared_step(batch, batch_idx, 'val')
```
adamjstewart, line 172: Can we add plotting here?

torchgeo/trainers/autoregression.py
```diff
@@ -0,0 +1,201 @@
...
+        if target_indices:
+            future_steps = future_steps[:, :, target_indices]
+        loss: Tensor = self.criterion(y_hat, future_steps)
+        self.log(f'{stage}_loss', loss)
+
+        # Retrieve the correct metrics based on the stage
+        metrics = getattr(self, f'{stage}_metrics', None)
+        if metrics:
+            metrics(y_hat, future_steps)
+            self.log_dict({f'{k}': v for k, v in metrics.compute().items()})
```
adamjstewart, line 148: Do we somewhere revert normalization before calculating metrics? Are we normalizing both inputs and outputs or only inputs?

tests/models/test_seq2seq.py
```diff
@@ -0,0 +1,119 @@
...
+BATCH_SIZE = [1, 2, 7]
+INPUT_SIZE_ENCODER = [1, 3]
+INPUT_SIZE_DECODER = [2, 3]
+OUTPUT_SIZE = [1, 2, 3]
+NUM_LAYERS = [1, 2, 3]
+HIDDEN_SIZE = [1, 2, 3]
+
+
+class TestLSTMSeq2Seq:
+    @torch.no_grad()
```
adamjstewart, line 18: I wonder if we should be adding `torch.no_grad` to more of our tests to speed them up...

adamjstewart (changes requested):
Looks like a great start! For datasets, I'm wondering if we also want to consider spatiotemporal datasets like the one you found and whether we should add GeoDatasets. For models, I'm wondering if we should also add a GNN since that's commonly used for non-gridded data. Some kind of model that combines both and can make spatiotemporal predictions would be cool. But none of those are required for a first pass. Let's first clean this up and get it merged.

I don't know why the tests are failing, but it's only the minimum version tests, so worst case scenario we can bump the minimum supported versions. We're already doing that when we drop Python 3.10 support anyway, so I would ignore them for now and focus on everything else.

torchgeo/datasets/air_quality.py
keves1: I dug into the code for their `ucimlrepo` package to figure out the url pattern that it uses to download the data.

adamjstewart: As long as it's stable, that's pretty cool.

torchgeo/datamodules/air_quality.py
keves1: I don't think we want to be applying a Kornia augmentation (`self.aug`) to this data since it is not an image, so this was my way of avoiding that. I think your suggestion for a no-op data augmentation would make this override unnecessary.

torchgeo/datamodules/air_quality.py
keves1: That would be great, is there something like that in Kornia? I didn't see anything when I looked through the docs. Or is there another way to do it?

torchgeo/datasets/air_quality.py
keves1: I'll add transforms. I didn't see a checksum on the page for this dataset, is that something I should generate?

torchgeo/datasets/air_quality.py
keves1: Yes, I think it makes sense as a NonGeoDataset since it doesn't have location data.

torchgeo/models/seq2seq.py
keves1: That sounds great, I'll work on that.

torchgeo/datamodules/air_quality.py
adamjstewart: No not really. We would have to write our own, or use one that won't apply like normalization.

torchgeo/models/seq2seq.py
keves1: I made this change.

torchgeo/trainers/autoregression.py
keves1: I had missed reverting normalization before calculating metrics, this is now added. Inputs and outputs are normalized (all steps in the sequence).

tests/models/test_seq2seq.py
keves1: I followed what I had seen in some of the tests for other models, seems like a good idea.

torchgeo/datamodules/air_quality.py
keves1: I've been thinking about this and since I don't think we are going to want to use Kornia augmentations for this particular dataset, it seems sufficient and simpler to just override on_after_batch_transfer so that no augmentations are applied. Are you ok with that?

adamjstewart:
This keeps coming up due to my own ignorance, but @nilsleh and others keep reminding me:

* Autoregression: when a model takes its own output as input at the next time step
* Forecasting: when a model predicts the future

Should we rename this to ForecastingTask? 

This name might still be too generic, since right now it only supports pointwise datasets. Should we make the name more specific, or add support for other possible target types like CDL masks in future PRs?

isaaccorley:
This model seems like it's performing forecasting via autoregression so I think AutoRegressiveTask is still the right naming.

keves1:
> This model seems like it's performing forecasting via autoregression so I think AutoRegressiveTask is still the right naming.

I agree, this is how I intended the model to be implemented so I think it is the right name.

torchgeo/datamodules/air_quality.py
keves1: @adamjstewart let me know what you want to do here.

adamjstewart: Does the default augmentation fail? I would assume that it gets ignored if it doesn't have any image or mask keys.

keves1: The default augmentation does fail, it results in Kornia giving `TypeError: object of type 'NoneType' has no len()` since there are no keys that are recognized and so `data_keys` is `None`.
image

adamjstewart: In that case let's just override on_after_batch_transfer. It's only one datamodule for now.

torchgeo/models/seq2seq.py
keves1: This is done.

torchgeo/datamodules/air_quality.py
```diff
@@ -0,0 +1,75 @@
...
+from .geo import NonGeoDataModule
+
+
+class AirQualityDataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the AirQuality dataset.
+
+    Uses the user provided splits to divide the dataset into
+    train/val/test sets.
+
+    .. versionadded:: 0.7
```
adamjstewart, line 21:
```suggestion
    .. versionadded:: 0.8
```

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,122 @@
...
+    Dataset Features:
+
+    * hourly averaged sensor responses and reference analyzer ground truth over one year (2004-2005)
+    * has missing features
+
+    If you use this dataset in your research, please cite:
+
+    * https://doi.org/10.1016/J.SNB.2007.09.060
+
+    .. versionadded:: 0.7
```
adamjstewart, line 38:
```suggestion
    .. versionadded:: 0.8
```

torchgeo/models/seq2seq.py
```diff
@@ -0,0 +1,245 @@
...
+        output = self.rnn(x)
+
+        if isinstance(self.rnn, nn.LSTM):
+            _, (hidden, cell) = output
+            return hidden, cell
+        elif isinstance(self.rnn, nn.GRU | nn.RNN):
+            _, hidden = output
+            return hidden, None
+        else:
+            raise TypeError(f'Unsupported RNN type: {type(self.rnn)}')
```
adamjstewart, line 63: No coverage, maybe we could just remove this runtime check

torchgeo/models/seq2seq.py
```diff
@@ -0,0 +1,245 @@
...
+
+        current_input = inputs[:, 0:1, :]
+
+        for t in range(self.output_sequence_len):
+            if isinstance(self.rnn, nn.LSTM):
+                _, (hidden, cell) = self.rnn(current_input, (hidden, cell))
+            elif isinstance(self.rnn, nn.GRU | nn.RNN):
+                _, hidden = self.rnn(current_input, hidden)
+            else:
+                raise TypeError(f'Unsupported RNN type: {type(self.rnn)}')
```
adamjstewart, line 130: Same here

torchgeo/models/seq2seq.py
```diff
@@ -0,0 +1,245 @@
...
+            last_layer_hidden = hidden[-1:]
+            output = self.fc(last_layer_hidden)
+            output = output.permute(1, 0, 2)  # put batch dimension first
```
adamjstewart, line 133: I prefer to use `einops.rearrange` to make it explicit what each dimension is.

torchgeo/trainers/autoregression.py
```diff
@@ -0,0 +1,187 @@
...
+            model: Name of the model to use, currently supports 'seq2seq'.
+                Defaults to 'seq2seq'.
+            input_size: The number of features in the input. Defaults to 1.
+            input_size_decoder: The number of features in the decoder input.
+                Defaults to 1.
+            output_size: The number of features output by the model. Defaults to 1.
+            target_indices: The indices of the target(s) in the dataset. If None, uses all features. Defaults to None.
+            loss: One of 'mse' or 'mae'. Defaults to 'mse'.
+            lr: Learning rate for optimizer. Defaults to 1e-3.
+            patience: Patience for learning rate scheduler. Defaults to 10.
```
adamjstewart, line 39: The default values and valid options will be documented in the function signature, no need to duplicate in the docstring. For example, see https://torchgeo.readthedocs.io/en/stable/api/trainers.html#torchgeo.trainers.ClassificationTask.__init__

torchgeo/trainers/autoregression.py
```diff
@@ -0,0 +1,187 @@
...
+            )
+
+    def configure_losses(self) -> None:
+        """Initialize the loss criterion.
+
+        Raises:
+            ValueError: If *loss* is invalid.
+        """
+        loss: str = self.hparams['loss']
+        if loss == 'mse':
```
adamjstewart, line 65: match-case statement, no need to check invalid options if you don't want to, hard to get coverage

torchgeo/trainers/autoregression.py
```diff
@@ -0,0 +1,187 @@
...
+    def test_step(self, batch: Any, batch_idx: int) -> None:
+        """Compute the test loss and additional metrics.
+
+        Args:
+            batch: The output of your DataLoader.
+            batch_idx: Integer displaying index of this batch.
+        """
+        self._shared_step(batch, batch_idx, 'test')
+
+    def predict_step(
```
adamjstewart, line 148: Currently no test coverage

torchgeo/datasets/air_quality.py
adamjstewart: Yes, you can compute this using `md5sum data.csv` after downloading

torchgeo/models/seq2seq.py
```diff
@@ -0,0 +1,245 @@
...
+                if self.target_indices:
+                    current_input[:, :, self.target_indices] = output
+                else:
+                    current_input = output
+
+        return outputs
+
+
+class Seq2Seq(nn.Module):
+    """Sequence-to-Sequence (Seq2Seq)."""
```
adamjstewart, line 150: Add:
```rst
.. versionadded:: 0.8
```
and add a citation for the first Seq2Seq paper, maybe https://arxiv.org/abs/1409.3215

torchgeo/trainers/autoregression.py
```diff
@@ -0,0 +1,187 @@
...
+from torchmetrics import MetricCollection
+from torchmetrics.regression import MeanAbsoluteError, MeanSquaredError
+
+from torchgeo.models import Seq2Seq
+
+from .base import BaseTask
+
+
+class AutoregressionTask(BaseTask):
+    """Autoregression."""
```
adamjstewart, line 19: Add:
```rst
.. versionadded:: 0.8
```

torchgeo/trainers/autoregression.py
```diff
@@ -0,0 +1,187 @@
...
+        self,
+        model: str = 'seq2seq',
+        input_size: int = 1,
+        input_size_decoder: int = 1,
+        output_size: int = 1,
+        target_indices: list[int] | None = None,
+        loss: str = 'mse',
+        lr: float = 1e-3,
+        patience: int = 10,
+        **kwargs: dict[str, Any],
```
adamjstewart, line 28: Seems like the minimum tests may be having trouble with kwargs? Also, I think the correct type hint is `Any`, ** is already assumed to be a dict.

adamjstewart:
RtD tests may require a rebase, pillow 11.2.0 was yanked but 11.3.0 is now used in main.

keves1:
@adamjstewart  I've been realizing there are some issues with the Seq2Seq implementation here, for supporting inference (when there are no future values) and for covariates. I've found there is a lot of complexity to properly supporting different forecasting situations when you start to include covariates that may or may not be present for the decoder (i.e. for the future). So for now I simplified it down to just supporting using the target time series as an input, with no extra variables. It does work now for inference where there are no future values. How much do we want to support here? Is this an ok starting point for this PR? I also found that there would need to be some changes to logging to properly log multiple outputs, and I saw that `RegressionTask` only does single output right now as well, so I assume leaving this as single output for now is fine.

adamjstewart:
> @adamjstewart I've been realizing there are some issues with the Seq2Seq implementation here, for supporting inference (when there are no future values) and for covariates. I've found there is a lot of complexity to properly supporting different forecasting situations when you start to include covariates that may or may not be present for the decoder (i.e. for the future). So for now I simplified it down to just supporting using the target time series as an input, with no extra variables. It does work now for inference where there are no future values. 

Trying to understand this. Are you saying for past events, both inputs and expected outputs work, but for future events, the model can only make all predictions without regard to any potential new inputs?

> How much do we want to support here? Is this an ok starting point for this PR? 

Probably fine, don't want to hold this up much longer. My main concern is backwards-compatibility. If it's easy to add new features without breaking backwards-compatibility, then we are fine.

> I also found that there would need to be some changes to logging to properly log multiple outputs, and I saw that `RegressionTask` only does single output right now as well, so I assume leaving this as single output for now is fine.

Single output doesn't seem like Seq2Seq, it seems like Seq2Int. Can't we compare predicted sequences to expected sequences and log only the averaged metrics?

adamjstewart:
@keves1 I would like to move this PR forward again. Would it be possible to split the dataset, model, and trainer into separate PRs? I think the dataset is already ready to merge, and we can hack on the model separately. The trainer could use one of the other 1D models we've already added or it could wait until after the Seq2Seq model is merged.

keves1:
@adamjstewart I created PRs for the dataset and model. I'll wait for the Seq2Seq model to be merged before doing a PR for the trainer.

adamjstewart:
@f-schi may be able to finish up the trainer once the dataset/model are merged. We now have several more pre-trained 1D time series models that could be incorporated. Before we spend too much more time on this specific PR, we need to brainstorm how to organize our trainers. 1D, 3D, and 4D time series modeling may all be autoregressive. I think we should have separate trainers for each, and possibly for classification and regression variants of each.

PR #2643, opened 2025-03-14
Update license resisc45 and vhr10

nilsleh:
According to https://gcheng-nwpu.github.io/#Datasets the license for both of these datasets is "CC-BY-NC-4.0"


isaaccorley (approved):

adamjstewart (approved):
Good find! I updated the licenses on HF too. Glad we now have legal permission to redistribute them lol.

PR #2646, opened 2025-03-14
Add Copernicus-FM

wangyi111:
Add [Copernicus-FM](https://github.com/zhu-xlab/Copernicus-FM/tree/main/Copernicus-FM), an extension of the [DOFA](https://github.com/zhu-xlab/DOFA) foundation model, able to process any spectral or non-spectral sensor modality using extended dynamic hypernetworks and flexible metadata encoding.

Key features:
- **A unified model for both spectral and non-spectral modalities** -- dynamic hypernetworks with Fourier / language encoding
- **Efficient processing of any spatial resolution** -- adaptive patch embedding kernel size
- **Flexible metadata integration** -- Fourier encoding with learnable meta tokens for geolocation, scale and time

### References

* https://github.com/zhu-xlab/Copernicus-FM/tree/main/Copernicus-FM
* https://huggingface.co/wangyi111/Copernicus-FM


torchgeo/models/copernicusfm.py
```diff
@@ -0,0 +1,724 @@
...
+    pos_embed: torch.Tensor,
+    new_size: tuple[int, int],
+    old_size: int | tuple[int, int] | None = None,
+    num_prefix_tokens: int = 1,
+    interpolation: str = 'bicubic',
+    antialias: bool = True,
+) -> torch.Tensor:
+    """Resize absolute position embeddings to a target resolution via interpolation
+
+    Borrowed from https://github.com/bwconrad/flexivit.
```
adamjstewart, line 45: Contains a lot of code from FlexiViT and Aurora, but both are MIT licensed, so this doesn't change the license of the file as a whole.

torchgeo/models/copernicusfm.py
```diff
@@ -0,0 +1,645 @@
...
+            self.spectrum_bandwidth_expansion = FourierExpansion(1, 1e9)
+        elif self.hypernet == 'variable':
+            # Variable hypernetwork: Language embedding for variable names.
+            # Load pre-trained language embeddings.
+            script_dir = os.path.dirname(os.path.abspath(__file__))
+            llm_embed_path = os.path.join(script_dir, 'var_embed_llama3.2_1B.pt')
+            if not os.path.exists(llm_embed_path):
+                url = "https://huggingface.co/wangyi111/Copernicus-FM/resolve/main/var_embed_llama3.2_1B.pt"
+                
+                download_url(url, script_dir, filename='var_embed_llama3.2_1B.pt')
```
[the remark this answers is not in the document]
adamjstewart: I'll change this to download to the default directory used to cache all PyTorch downloads: https://pytorch.org/docs/stable/hub.html#torch.hub.get_dir

torchgeo/models/copernicusfm.py
```diff
@@ -325,21 +325,33 @@ def __init__(self, wv_planes=128, kernel_size=16, embed_dim=1024, hypernet='spec
...
+                'Sentinel 5P Sulfur Dioxide'
+            ]
+            self.language_embed['dem'] = self.language_embed[
+                'Copernicus Digital Elevation Model'
+            ]
             self.language_proj = nn.Linear(2048, self.wv_planes)
         else:
-            raise ValueError("Unsupported hypernet type. Choose 'spectral' or 'language'.")
+            raise ValueError(
+                "Unsupported hypernet type. Choose 'spectral' or 'language'."
```
wangyi111: I did a typo here. it should be "spectral" or "variable".

adamjstewart: Removed anyway, we can list valid inputs as type hints.

torchgeo/models/copernicusfm.py
```diff
@@ -523,18 +575,30 @@ def forward_features(self, x, meta_info, key, wave_list, bandwidth, language_emb
         if input_mode == 'spectral':
```
wangyi111, line 589: ok i think this is indeed broken. i likely only tested "spectral". these arguments have to be organized consistently.

torchgeo/models/copernicusfm.py
```diff
@@ -0,0 +1,772 @@
...
+
+        Args:
+            x: Input mini-batch.
+            meta_info: Longitudes, latitudes, times, and areas of each patch.
+                Use NaN for unknown metadata.
+            key: Key to retrieve a language embedding.
+                Only used if *input_mode=='variable'*.
+            wave_list: Wavelengths of each spectral band (μm).
+                Only used if *input_mode=='spectral'*.
+            bandwidth: Bandwidths in nm.
```
adamjstewart, line 691: @wangyi111 can you confirm the units on these?

torchgeo/models/copernicusfm.py
```diff
@@ -0,0 +1,772 @@
...
+        language_embed: Tensor | None = None,
+        input_mode: Literal['spectral', 'variable'] = 'spectral',
+        kernel_size: int | None = None,
+    ) -> Tensor:
+        """Forward pass of the model.
+
+        Args:
+            x: Input mini-batch.
+            meta_info: Longitudes, latitudes, times, and areas of each patch.
+                Use NaN for unknown metadata.
```
adamjstewart, line 686: This is an unintuitive UI. I would rather have separate values for each which are either Tensor or None. It's also a shame that we can't mix this in a single mini-batch, if a single value is NaN that metadata is ignored.

torchgeo/models/copernicusfm.py
wangyi111: both units are nm

torchgeo/models/copernicusfm.py
```diff
@@ -0,0 +1,772 @@
...
+                Only used if *input_mode=='variable'*.
+            input_mode: One of 'spectral' or 'variable'.
+            kernel_size: If provided and differs from the initialized kernel size,
+                the generated patchembed kernel weights are resized accordingly.
+
+        Returns:
+            Output mini-batch.
+        """
+        # embed patches
+        wavelist = torch.tensor(wave_list, device=x.device).float()
```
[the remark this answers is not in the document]
wangyi111: again historical issue, they are required when I use sensitive wavelength for the variable hypernet. in the language-driven variable hypernet they are not used.

torchgeo/models/copernicusfm.py
wangyi111: can make it possible to mix in the batch in principle, but needs looping over the batch dim to assign known/unknown, probably will change a lot of codes.

torchgeo/models/copernicusfm.py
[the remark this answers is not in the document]
wangyi111: can move these two lines to inside `input_mode=='spectral'`, and remove them from the variable hypernet forward arguments

torchgeo/models/copernicusfm.py
```diff
@@ -0,0 +1,716 @@
...
+            embed_dim: Embedding dimension.
+
+        Returns:
+            Coordinate position embedding.
+        """
+        coord_embed_lon = self.coord_expansion(lons + 180, embed_dim // 2)
+        coord_embed_lat = self.coord_expansion(lats + 90, embed_dim // 2)
+        coord_embed = torch.cat([coord_embed_lon, coord_embed_lat], dim=-1)
+
+        if coord_embed.shape[-1] < embed_dim:
```
adamjstewart, line 472: This was actually surprisingly hard to test. We don't hit this situation unless `embed_dim` is odd, but `embed_dim` can't be odd if `get_area_pos_embed` or `get_time_pos_embed` are used. Not only does `embed_dim` have to be odd, but `embed_dim // 2` has to be even. So I had to test a situation where `embed_dim=5` and area/time are NaN.

adamjstewart (dismissed):
This is ready from my side, but I'll give others a couple days to review. I'm particularly concerned about whether the documentation is sufficient for people to figure out how to use the model. There are ways we could make this more user-friendly, but I don't want to diverge too much from the original source code.

calebrob6:
Just read through this and found myself wanting an example of how to use it (same thing I hit previously when trying to use Scale-MAE)

E.g. even though the args in `forward` are documented -- what do they _mean_:

image


Maybe it'd be nice to put an example in the docstring? (This also applies to other pre-trained models actually)

adamjstewart:
Agreed, we actually got a similar request for DOFA: https://github.com/zhu-xlab/DOFA/issues/14

These newer models (Copernicus-FM, Panopticon) add a lot more (optional) metadata, so are even more confusing to use. Not sure if this should be API documentation or tutorials or what. I probably don't have a ton of time to work on this personally but @wangyi111 might.

wangyi111:
> Agreed, we actually got a similar request for DOFA: [zhu-xlab/DOFA#14](https://github.com/zhu-xlab/DOFA/issues/14)
> 
> These newer models (Copernicus-FM, Panopticon) add a lot more (optional) metadata, so are even more confusing to use. Not sure if this should be API documentation or tutorials or what. I probably don't have a ton of time to work on this personally but @wangyi111 might.

Should be easy for me to add the docstring. Regarding tutorial, is there such place for demonstrating a pretrained model? I only see https://torchgeo.readthedocs.io/en/stable/tutorials/pretrained_weights.html

torchgeo/models/copernicusfm.py
wangyi111: This is indeed rarely hit. I added this when I tested Cartesian coords (x,y,z) where each axis has `embed_dm // 3`. Now that we are not using that anymore it might be fine to skip it.

adamjstewart:
Yep, that's the right location. We could either expand that tutorial to cover additional models, or add a second tutorial specifically for using FMs.

adamjstewart (approved):
I vote we merge this as is and @wangyi111 can open a separate PR to expand our tutorials on Scale-MAE, DOFA, Copernicus-FM, etc. Any objections?

wangyi111:
> I vote we merge this as is and @wangyi111 can open a separate PR to expand our tutorials on Scale-MAE, DOFA, Copernicus-FM, etc. Any objections?

oh I just added some docstring to copernicusfm class

torchgeo/models/copernicusfm.py
```diff
@@ -0,0 +1,743 @@
...
+    
+    Example:
+        **1. Spectral Mode (Using Wavelength and Bandwidth):**
+
+        >>> model = CopernicusFM()
+        >>> img = torch.randn(1, 4, 224, 224) # input image
+        >>> meta = torch.full((1, 4), float('nan')) # [lon, lat, delta_time, patch_token_area], assume unknown
+        >>> wvs = [490, 560, 665, 842] # wavelength: B,G,R,NIR (Sentinel 2)
+        >>> bws = [65, 35, 30, 115] # bandwidth: B,G,R,NIR (Sentinel 2)
+        >>> language_embed = None # N/A in this mode
```
adamjstewart, line 376: We don't really need to pass None here, it's the default

torchgeo/models/copernicusfm.py
```diff
@@ -0,0 +1,743 @@
...
+
+class CopernicusFM(nn.Module):
+    """CopernicusFM: VisionTransformer backbone.
+    
+    Example:
+        **1. Spectral Mode (Using Wavelength and Bandwidth):**
+
+        >>> model = CopernicusFM()
+        >>> img = torch.randn(1, 4, 224, 224) # input image
+        >>> meta = torch.full((1, 4), float('nan')) # [lon, lat, delta_time, patch_token_area], assume unknown
```
adamjstewart, line 373: We don't really document units (degress, seconds?, km^2, nm) for most of these inputs. I'm not even sure if I know the units for delta_time.

torchgeo/models/copernicusfm.py
wangyi111: updated

torchgeo/models/copernicusfm.py
wangyi111: updated

adamjstewart:
How would you feel about renaming a few things for consistency:

* img_feat -> image or x
* meta_info -> metadata
* wave_list, wvs -> wavelengths
* wv_planes -> wavelength_dim
* bandwidth -> bandwidths
* hypernet -> input_mode

Could also split meta_info into 4 separate variables for ease of use. Don't want to diverge too much from the original implementation, but also want to make it user friendly and intuitive.

wangyi111:
> How would you feel about renaming a few things for consistency:
> 
> * img_feat -> image or x
> * meta_info -> metadata
> * wave_list, wvs -> wavelengths
> * wv_planes -> wavelength_dim
> * bandwidth -> bandwidths
> * hypernet -> input_mode
> 
> Could also split meta_info into 4 separate variables for ease of use. Don't want to diverge too much from the original implementation, but also want to make it user friendly and intuitive.

Good for me. Only one is wv_plane, which is not only the dim of wavelength but also bandwidth and language embed. maybe something like hyper_dim? i kind of wanted to call it meta_dim but metadata also means another thing in this model..

adamjstewart:
Maybe `in_dim` or `in_features`?

wangyi111:
these can still lead to input image features, maybe `hyper_planes`?

adamjstewart:
Finished renaming. Remaining ideas to improve usabillity:

- [ ] Could remove `input_mode` and key it based on whether `wavelengths/bandwidths` or `language_embed` is provided
- [ ] Could use type hints to make it clear that `wavelengths/bandwidths` _or_ `language_embed` is required, not both nor neither
- [ ] Could split `metadata` into lat/lon/time/area, makes it easier to skip certain variables

Don't want to spend too much time on this because we still need to finish Copernicus-Bench and Copernicus-Pretrain, but once it's merged it makes it harder to change without breaking backwards compatibility.

adamjstewart (dismissed):

Issue #2640, opened 2025-03-13
Pretrained ResNet/ViT models do not work when `features_only=True`


lccol:
### Description

AssertionError is raised when trying to load any ResNet152-SATLAS pretrained model when `features_only=True` due to the presence of `fc.weights` and `fc.bias` in model checkpoints.

Probably, assertion should also be modified with something like this?
`assert set(unexpected_keys) <= {'fc.weight', 'fc.bias'}`

### Steps to reproduce

```python
from torchgeo.models import resnet

# raises AssertionError on assert not unexpected_keys
model = resnet.resnet152(
    weights=resnet.ResNet152_Weights.SENTINEL2_MI_MS_SATLAS, features_only=True
)
model = resnet.resnet152(
    weights=resnet.ResNet152_Weights.SENTINEL2_MI_RGB_SATLAS, features_only=True
)
model = resnet.resnet152(
    weights=resnet.ResNet152_Weights.SENTINEL2_SI_MS_SATLAS, features_only=True
)
model = resnet.resnet152(
    weights=resnet.ResNet152_Weights.SENTINEL2_SI_RGB_SATLAS, features_only=True
)
```

### Version

0.6.2

adamjstewart:
Interesting, we never thought about `features_only` when designing this. I think your fix is correct. Want to submit a PR? This not only affects ResNet-152 Satlas but all `timm.create_model` usage:
```
torchgeo/models/resnet.py:        assert not unexpected_keys
torchgeo/models/resnet.py:        assert not unexpected_keys
torchgeo/models/resnet.py:        assert not unexpected_keys
torchgeo/models/swin.py:        assert not unexpected_keys
torchgeo/models/swin.py:        assert not unexpected_keys
torchgeo/models/vit.py:        assert not unexpected_keys
```

adamjstewart:
@lccol we're preparing a 0.7.0 release in the next few days. Do you have time to submit a PR to fix this or do you want me to fix it?

PR #2659, opened 2025-03-20
Fix Pretrained ResNet/ViT not working when features_only=True

lccol:
Fix the AssertionError raised when pretrained ResNet/ViT models (from timm) were created with `features_only = True`.

ResNet18 and Swin models were not affected by the issue (Swin models are created with torchvision instead of timm).

Closes [#2640](https://github.com/microsoft/torchgeo/issues/2640).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,7 +8,6 @@ import matplotlib
 import pytest
 import torch
 import torchvision
-from _pytest.fixtures import SubRequest
 from pytest import MonkeyPatch
 
 
@@ -29,8 +28,3 @@ def matplotlib_backend() -> None:
 @pytest.fixture(autouse=True)
 def torch_hub(tmp_path: Path) -> None:
     torch.hub.set_dir(tmp_path)
-
-
-@pytest.fixture(params=[True, False])
-def features_only(request: SubRequest) -> bool:
-    return bool(request.param)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,6 +8,7 @@ import matplotlib
 import pytest
 import torch
 import torchvision
+from _pytest.fixtures import SubRequest
 from pytest import MonkeyPatch
 
 
@@ -28,3 +29,8 @@ def matplotlib_backend() -> None:
 @pytest.fixture(autouse=True)
 def torch_hub(tmp_path: Path) -> None:
     torch.hub.set_dir(tmp_path)
+
+
+@pytest.fixture(params=[True, False])
+def features_only(request: SubRequest) -> bool:
+    return bool(request.param)
--- a/tests/models/conftest.py
+++ b/tests/models/conftest.py
@@ -4,4 +4,6 @@ from _pytest.fixtures import SubRequest
 
 @pytest.fixture(params=[True, False])
 def features_only(request: SubRequest) -> bool:
+    if request.param:
+        pytest.importorskip('timm', minversion='1.0.3')
     return bool(request.param)
--- a/tests/models/conftest.py
+++ b/tests/models/conftest.py
@@ -1,3 +1,6 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
 import pytest
 from _pytest.fixtures import SubRequest
 
--- a/tests/models/conftest.py
+++ b/tests/models/conftest.py
@@ -1,0 +1,7 @@
+import pytest
+from _pytest.fixtures import SubRequest
+
+
+@pytest.fixture(params=[True, False])
+def features_only(request: SubRequest) -> bool:
+    return bool(request.param)
--- a/tests/models/conftest.py
+++ b/tests/models/conftest.py
@@ -5,5 +5,6 @@ from _pytest.fixtures import SubRequest
 @pytest.fixture(params=[True, False])
 def features_only(request: SubRequest) -> bool:
     if request.param:
+        # features_only arg in ViT supported only from timm 1.0.3
         pytest.importorskip('timm', minversion='1.0.3')
     return bool(request.param)
--- a/tests/models/test_resnet.py
+++ b/tests/models/test_resnet.py
@@ -25,34 +25,22 @@ class TestResNet18:
     def weights(self, request: SubRequest) -> WeightsEnum:
         return request.param
 
+    @pytest.fixture(params=[True, False])
+    def features_only(self, request: SubRequest) -> bool:
+        return bool(request.param)
+
     @pytest.fixture
     def mocked_weights(
         self,
         tmp_path: Path,
         monkeypatch: MonkeyPatch,
         weights: WeightsEnum,
-        load_state_dict_from_url: None,
-    ) -> WeightsEnum:
-        path = tmp_path / f'{weights}.pth'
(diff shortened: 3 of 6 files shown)


torchgeo/models/resnet.py
```diff
@@ -9,6 +9,7 @@
 import timm
 import torch
 from timm.models import ResNet
+from timm.models._features import FeatureListNet
```
adamjstewart, line 12: Don't fully understand why this change is needed, and also hesitant to import anything from a private module that may change.

isaaccorley: This is because when you pass features_only=True to timm.create_model it wraps the model as a feature extractor class

lccol: I changed the input for FeatureListNet (from `timm.models._features` to `timm.models`). The same could not be said for FeatureGetterNet (used in ViT), which is causing failures in some tests. I'm probably going to drop the `FeatureGetterNet` wrapper import somehow...

isaaccorley:
We will probably have to bump the min version of timm we support because looks like FeatureGetterNet and FeatureListNet aren't available in the min version. Will have to search to see when they were first added

torchgeo/models/resnet.py
isaaccorley: I think it's fine to drop it

adamjstewart: Could we just use `nn.Module` instead? It would be a shame to lose all granularity, but it would also be technically correct.

lccol: Replaced it with `nn.ModuleDict`, which is the parent class for both.

lccol:
> We will probably have to bump the min version of timm we support because looks like FeatureGetterNet and FeatureListNet aren't available in the min version. Will have to search to see when they were first added

Yes. Unfortunately, in the current min version of timm, ViT model does not support the `features_only` param:

```
RuntimeError: features_only not implemented for Vision Transformer models.
```

adamjstewart:
Should we skip the tests for older versions of timm? We don't necessarily need to bump the minimum version if it's very new.

lccol:
I checked and timm 1.0.3 was released in April/May 2024, so the introduction of `features_only` on ViT is quite recent.

isaaccorley:
You can just skip the tests using `pytest.mark.skip` and check what version of timm is installed. We have some examples of this, I'll send you a link.

adamjstewart:
Or better:
```python
import pytest

def test_features_only():
    pytest.importorskip('timm', minversion='1.0.3')
```

tests/models/test_resnet.py
```diff
@@ -42,12 +42,36 @@ def mocked_weights(
             monkeypatch.setattr(weights, 'url', str(path))
         return weights
 
+    @pytest.fixture
+    def mocked_weights_features_only(
```
adamjstewart, line 46: Can we combine these two tests so there is less duplicated code? Something like:
```python
def mocked_weights(..., request: SubRequest):
    features_only = request.param
```

tests/models/test_resnet.py
```diff
@@ -42,12 +42,36 @@ def mocked_weights(
...
+            monkeypatch.setattr(weights, 'url', str(path))
+        return weights
+
     def test_resnet(self) -> None:
         resnet18()
 
     def test_resnet_weights(self, mocked_weights: WeightsEnum) -> None:
         resnet18(weights=mocked_weights)
 
+    def test_resnet_weights_features_only(
```
adamjstewart, line 70: Same here, we can do:
```python
@pytest.mark.parametrize('features_only', [True, False]):
def test_resnet_weights(..., features_only: bool):
    ...
```

adamjstewart (approved):
Just a minor request to make the tests simpler, I can do this myself if you want.

tests/models/test_resnet.py
lccol: Ok, I'll change it by the end of the day. But since the two set of weights are different, by doing this pytest won't test `mocked_weights` with `features_only=True` with `test_resnet_weights(..., features_only=False)` and viceversa? Is there a way to invoke the test function with param `features_only=True` and `mocked_weights` created with the same value of the flag?

tests/models/test_resnet.py
```diff
@@ -25,16 +25,23 @@ class TestResNet18:
     def weights(self, request: SubRequest) -> WeightsEnum:
         return request.param
 
+    @pytest.fixture(params=[True, False])
+    def features_only(self, request: SubRequest) -> bool:
+        return bool(request.param)
```
adamjstewart, line 30: Could define this a single time at the module level in a `tests/models/conftest.py` file instead of 4 times. Is the `bool(...)` cast necessary?

torchgeo/models/vit.py
```diff
@@ -242,16 +243,24 @@ def vit_small_patch16_224(
...
+    # FeatureGetterNet (extends nn.ModuleDict) is returned when features_only=True
+    model: VisionTransformer | nn.ModuleDict = timm.create_model(
         'vit_small_patch16_224', *args, **kwargs
     )
 
+    if kwargs.get('features_only', False):
+        model = cast(nn.ModuleDict, model)
+        target_model = cast(VisionTransformer, model.model)
+    else:
+        model = cast(VisionTransformer, model)
```
adamjstewart, line 255: What's the story with the casts here? Usually these are a sign of a typing bug being ignored.

adamjstewart (commented):
Tests look much cleaner now, thanks!

tests/models/test_resnet.py
lccol: For some reason, without the cast mypy raises the following error:

```
tests/models/test_vit.py:26: error: Returning Any from function declared to return "bool"  [no-any-return]
            return request.param
```

torchgeo/models/vit.py
lccol: Same here, mypy does not recognize the type for `model.model`

```
torchgeo/models/vit.py:260: error: Item "Tensor" of "Tensor | Module" has no attribute "load_state_dict"  [union-attr]
            missing_keys, unexpected_keys = target_model.load_state_dict(
                                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

tests/conftest.py
```diff
@@ -28,3 +29,8 @@ def matplotlib_backend() -> None:
 @pytest.fixture(autouse=True)
 def torch_hub(tmp_path: Path) -> None:
     torch.hub.set_dir(tmp_path)
+
+
+@pytest.fixture(params=[True, False])
```
adamjstewart, line 34: This could probably be in `tests/models/conftest.py` since it's only relevant to those tests.

tests/models/conftest.py
adamjstewart, line 1: Can you just add the copyright:
```python
# Copyright (c) Microsoft Corporation. All rights reserved.
```

lccol: Sorry, I forgot that. Just pushed it!

adamjstewart (approved):
I think this looks good now, thanks for putting up with my tedious change requests!

Note that this PR does increase the time our unit tests take to run from 11m 34s to 12m 52s (+11%). For now that's tolerable. If we add more models in the future and this becomes too much I may only run the `features_only=True` tests on the release branch.

Thanks again for the PR!

PR #2662, opened 2025-03-21
BENv2: add citation information

adamjstewart:


nilsleh (approved):
Ups, sorry, good catch!

PR #2664, opened 2025-03-22
major revision of SSL4EO downloading script

cmalbrec:
according to microsoft/torchgeo#2345

### DIFF
--- a/experiments/ssl4eo/download_ssl4eo.py
+++ b/experiments/ssl4eo/download_ssl4eo.py
@@ -1,553 +1,780 @@
 #!/usr/bin/env python3
-
-# Copyright (c) Microsoft Corporation. All rights reserved.
-# Licensed under the MIT License.
-
-""" Sample and download Satellite images with Google Earth Engine
-
-#### run the script:
-
-### Install and authenticate Google Earth Engine
-
-### match and download pre-sampled locations
-python download_ssl4eo.py \
-    --save-path ./data \
-    --collection COPERNICUS/S2_HARMONIZED \
-    --meta-cloud-name CLOUDY_PIXEL_PERCENTAGE \
-    --cloud-pct 20 \
-    --dates 2021-12-21 2021-09-22 2021-06-21 2021-03-20 \
-    --radius 1320 \
-    --bands B1 B2 B3 B4 B5 B6 B7 B8 B8A B9 B10 B11 B12 \
-    --dtype uint16 \
-    --num-workers 8 \
-    --log-freq 100 \
-    --match-file ./data/sampled_locations.csv \
-    --indices-range 0 250000
-
-### resume from interruption
-python download_ssl4eo.py \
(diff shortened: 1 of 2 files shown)


cmalbrec:

adamjstewart (changes requested):
This is more than just a refactor, this is a complete rewrite. Is this new script compatible with the command-line args used in `experiments/ssl4eo/landsat/download_*.sh`? I noticed you also removed the rate throttle, does this not suffer from exceeding the GEE API rate limit? 

To properly review this I would probably need:

* A list of changes made from the current version
* Why these changes were made
* Performance metrics before and after this change

It may be quicker and easier to simply fix the geolocation bug, I can't imagine it requires more than a few lines changed.

cmalbrec:
Hi @adamjstewart :

Yes, with @AABNassim we took the liberty to rewrite the download script to better organize for future development compared to the previous one. https://github.com/cmalbrec/torchgeo/blob/issue-2345/experiments/ssl4eo/download_ssl4eo.sh serves as an example on how to adapt `experiments/ssl4eo/landsat/download_*.sh`.

We successfully utilized the new downloading script to swiftly curate https://huggingface.co/datasets/embed2scale/SSL4EO-S12-v1.1 . I vote to move forward with the new version and add mods like rate throttling (@wangyi111 may help here), etc.

adamjstewart:
Is your version based on the [SSL4EO-S12](https://github.com/zhu-xlab/SSL4EO-S12/blob/main/src/download_data/ssl4eo_s12_downloader.py) or [SSL4EO-L](https://github.com/microsoft/torchgeo/blob/main/experiments/ssl4eo/download_ssl4eo.py) download script? The latter underwent months of work to make it more generic to support both Sentinel and Landsat, with better overlap checking, nodata checking, and optimized downloading. I wouldn't want to see those months of work undone.

If you want to move forward with this version, I'll still need to see:

* A list of changes made from the current version
* Why these changes were made
* Performance metrics before and after this change

P.S. I've had about a dozen people ask me when SSL4EO-S12 v1.1 will be added to TorchGeo so it is easier to download.

isaaccorley:
@cmalbrec  I'm guessing this was used for downloading SSL4EO-S12v1.1? Shall we just create a new script with a bumped version?

adamjstewart:
We all met offline to discuss this. We basically decided to move forward with this PR but make sure all changes made from SSL4EO-S12 to SSL4EO-L are maintained. I want to move this script out of TorchGeo eventually but not yet.

adamjstewart:
FYI, now that we have our own https://github.com/torchgeo organization, we were finally able to move the SSL4EO experiment code to a separate dedicated repo: https://github.com/torchgeo/ssl4eo-l. It would be best to open a new PR at that location. I would still recommend simple short PRs to fix reprojection and add other features instead of one giant file replacement to make the review easier.

PR #2668, opened 2025-03-24
ReforesTree: skip images with missing mappings

adamjstewart:
Reported by @lns-lns


calebrob6:
> Find better way, maybe read final_dataset.csv instead of using glob, or redistribute on HF without these images?

I think this way is fine -- why not?

adamjstewart:
It's just inelegant. If I run out of time I may leave it as is.

adamjstewart:
Bleh, we would have to create a new fake testing file in order to make this any fancier. Let's just merge as is while we wait for a response from the dataset authors.

PR #2669, opened 2025-03-24
Pin setuptools to fix CI

adamjstewart:
The latest [setuptools 78.0.0 release](https://setuptools.pypa.io/en/stable/history.html#v78-0-0) drops support for `description-file` in `setup.cfg`. Unfortunately, SMP currently depends on pretrainedmodels, which uses this syntax and is no longer maintained.

Although this dependency has been [vendored in SMP](https://github.com/qubvel-org/segmentation_models.pytorch/pull/1039), we need to wait for a new release to officially remove this dependency. Until then, we need to find another workaround.

Things I've tried so far:

- [x] ~Adding an upper bound on setuptools in pyproject.toml~
- [x] ~Using a pip constraints file~
- [x] ~Installing the main branch of SMP~
- [ ] Disabling build isolation

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -30,7 +30,7 @@ jobs:
       - name: Install pip dependencies
         if: steps.cache.outputs.cache-hit != 'true'
         run: |
-          pip install .[datasets,tests]
+          pip install .[datasets,tests] -c requirements/max-cons.old
           pip cache purge
       - name: List pip dependencies
         run: pip list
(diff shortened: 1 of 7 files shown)


adamjstewart:
This breaking change was reverted in the [setuptools 78.0.2 release](https://setuptools.pypa.io/en/stable/history.html#v78-0-2), so we no longer need to panic. Pretrainedmodels has already been dropped as a dependency in SMP. The next release should come out long before setuptools drops this again.

PR #2676, opened 2025-03-25
Rehosted Biomassters dataset to torchgeo HF

nilsleh:
I rehosted the Biomassters dataset, because the split zip file format was causing issues for people. I also renamed the csv metafile because the `.csv.csv` filename was bugging me.

I checked that I can load a sample from the dataset based on the HF torchgeo version, but don't have time to add automatic download for it.


adamjstewart (commented):
Would like to at least add a URL where people can download this from even if we don't add automatic downloading. Otherwise we have a data loader that doesn't work for the official source.

PR #2677, opened 2025-03-25
SoftCon pretrained models

wangyi111:
Add [SoftCon](https://github.com/zhu-xlab/softcon) pretrained models for Sentinel-1 GRD (2 bands) and Sentinel-2 TOA (13 bands). SoftCon conducts multi-label soft contrastive learning with land cover semantics, as well as continual pretraining from DINO (resnet) or DINOv2 (vits).

image

There're 6 new weights in total, including 1 ResNet and 2 ViTs for both Sentinel-1 and Sentinel-2:
- ResNet50
- ViT-S/14 DINOv2
- ViT-B/14 DINOv2

DINOv2 ViTs have been implemented in timm since [v0.9.1](https://github.com/huggingface/pytorch-image-models/releases/tag/v0.9.1).


wangyi111:
This PR also includes a renaming of existing Sentinel1_ALL weights to Sentinel1_GRD weights. Might be more suitable when we later add Sentinel-1 SLC models.

hubconf.py
```diff
@@ -18,7 +18,12 @@
     scalemae_large_patch16,
     swin_v2_b,
     swin_v2_t,
+    vit_base_patch16_224,
+    vit_huge_patch14_224,
+    vit_large_patch16_224,
     vit_small_patch16_224,
+    vit_small_patch14_dinov2,
```
wangyi111, line 25: Fixing ruff error would split apart the position of dinov2 small and base. Should we do that?

adamjstewart: Yes that's fine, listen to ruff.

wangyi111:
`tests/models/test_vit.py` failed not sure for what reason..

adamjstewart:
Since it isn't a timeout and there is no error msg, I'm guessing OOM error. We may have once again reached the point where there are so many large models we need to download that we have filled up the minimal storage on the CI runners. Let me see if I can squeeze out more performance...

PR #2682, opened 2025-03-27
Refactor _RandomNCrop to RandomCrop + ExtractPatches

isaaccorley:
This PR fixes https://github.com/microsoft/torchgeo/issues/2667 and refactors `_RandomNCrop` out of segmentation and change detection datamodules in favor of `K.RandomCrop` during training and `ExtractPatches` for val/test/predicting.

### DIFF
--- a/tests/datamodules/test_levircd.py
+++ b/tests/datamodules/test_levircd.py
@@ -51,9 +51,9 @@ class TestLEVIRCDPlusDataModule:
         batch = next(iter(datamodule.train_dataloader()))
         batch = datamodule.on_after_batch_transfer(batch, 0)
         assert batch['image1'].shape[-2:] == batch['mask'].shape[-2:] == (256, 256)
-        assert batch['image1'].shape[0] == batch['mask'].shape[0] == 8
+        assert batch['image1'].shape[0] == batch['mask'].shape[0] == 1
         assert batch['image2'].shape[-2:] == batch['mask'].shape[-2:] == (256, 256)
-        assert batch['image2'].shape[0] == batch['mask'].shape[0] == 8
+        assert batch['image2'].shape[0] == batch['mask'].shape[0] == 1
         assert batch['image1'].shape[1] == 3
         assert batch['image2'].shape[1] == 3
 
@@ -64,14 +64,10 @@ class TestLEVIRCDPlusDataModule:
         batch = next(iter(datamodule.val_dataloader()))
         batch = datamodule.on_after_batch_transfer(batch, 0)
         if datamodule.val_split_pct > 0.0:
-            assert (
-                batch['image1'].shape[-2:] == batch['mask'].shape[-2:] == (1024, 1024)
-            )
-            assert batch['image1'].shape[0] == batch['mask'].shape[0] == 1
-            assert (
-                batch['image2'].shape[-2:] == batch['mask'].shape[-2:] == (1024, 1024)
-            )
-            assert batch['image2'].shape[0] == batch['mask'].shape[0] == 1
+            assert batch['image1'].shape[-2:] == batch['mask'].shape[-2:] == (256, 256)
+            assert batch['image1'].shape[0] == batch['mask'].shape[0] == 16
+            assert batch['image2'].shape[-2:] == batch['mask'].shape[-2:] == (256, 256)
+            assert batch['image2'].shape[0] == batch['mask'].shape[0] == 16
             assert batch['image1'].shape[1] == 3
             assert batch['image2'].shape[1] == 3
 
@@ -81,10 +77,10 @@ class TestLEVIRCDPlusDataModule:
             datamodule.trainer.testing = True
         batch = next(iter(datamodule.test_dataloader()))
         batch = datamodule.on_after_batch_transfer(batch, 0)
-        assert batch['image1'].shape[-2:] == batch['mask'].shape[-2:] == (1024, 1024)
-        assert batch['image1'].shape[0] == batch['mask'].shape[0] == 1
-        assert batch['image2'].shape[-2:] == batch['mask'].shape[-2:] == (1024, 1024)
-        assert batch['image2'].shape[0] == batch['mask'].shape[0] == 1
+        assert batch['image1'].shape[-2:] == batch['mask'].shape[-2:] == (256, 256)
+        assert batch['image1'].shape[0] == batch['mask'].shape[0] == 32
+        assert batch['image2'].shape[-2:] == batch['mask'].shape[-2:] == (256, 256)
+        assert batch['image2'].shape[0] == batch['mask'].shape[0] == 32
         assert batch['image1'].shape[1] == 3
         assert batch['image2'].shape[1] == 3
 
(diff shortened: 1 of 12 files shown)


tests/datamodules/test_levircd.py
```diff
@@ -51,9 +51,9 @@ def test_train_dataloader(self, datamodule: LEVIRCDPlusDataModule) -> None:
         batch = next(iter(datamodule.train_dataloader()))
         batch = datamodule.on_after_batch_transfer(batch, 0)
         assert batch['image1'].shape[-2:] == batch['mask'].shape[-2:] == (256, 256)
-        assert batch['image1'].shape[0] == batch['mask'].shape[0] == 8
+        assert batch['image1'].shape[0] == batch['mask'].shape[0] == 1
         assert batch['image2'].shape[-2:] == batch['mask'].shape[-2:] == (256, 256)
-        assert batch['image2'].shape[0] == batch['mask'].shape[0] == 8
+        assert batch['image2'].shape[0] == batch['mask'].shape[0] == 1
```
isaaccorley, line 56: When using K.RandomCrop we will only get 1 crop per image

tests/datamodules/test_levircd.py
```diff
@@ -64,14 +64,10 @@ def test_val_dataloader(self, datamodule: LEVIRCDPlusDataModule) -> None:
...
         if datamodule.val_split_pct > 0.0:
-            assert (
-                batch['image1'].shape[-2:] == batch['mask'].shape[-2:] == (1024, 1024)
-            )
-            assert batch['image1'].shape[0] == batch['mask'].shape[0] == 1
-            assert (
-                batch['image2'].shape[-2:] == batch['mask'].shape[-2:] == (1024, 1024)
-            )
-            assert batch['image2'].shape[0] == batch['mask'].shape[0] == 1
+            assert batch['image1'].shape[-2:] == batch['mask'].shape[-2:] == (256, 256)
```
isaaccorley, line 67: Previously val/test was being done on the full 1024x1024 images. `ExtractPatches` fixes this so now the val/test images are patches instead.

torchgeo/datamodules/inria.py
```diff
@@ -77,3 +85,26 @@ def setup(self, stage: str) -> None:
         if stage in ['predict']:
             # Test set masks are not public, use for prediction instead
             self.predict_dataset = InriaAerialImageLabeling(split='test', **self.kwargs)
+
+    def on_after_batch_transfer(
```
isaaccorley, line 89: Because our test data only had 1 sample for train/test this bug appeared where the batch dimension wouldn't be created. This is a fix for this but we can investigate a better solution in the future.

ashnair1 (dismissed):

adamjstewart:
I would love to get rid of `_ExtractPatches` too, but I'm guessing https://github.com/kornia/kornia/issues/1300 prevents us from doing that? It may be time to work more closely with the Kornia developers to upstream and replace our broken custom transforms.

isaaccorley:
> I would love to get rid of `_ExtractPatches` too, but I'm guessing [kornia/kornia#1300](https://github.com/kornia/kornia/issues/1300) prevents us from doing that? It may be time to work more closely with the Kornia developers to upstream and replace our broken custom transforms.

I'll bring that back from the grave after this gets merged. I want to see if they can help me with extracting patches but also being able to extract bboxes since this is important for object detection datasets that have large imagery. We don't want to lose users to SAHI.

torchgeo/datamodules/deepglobelandcover.py
```diff
@@ -40,14 +40,23 @@ def __init__(
...
             K.Normalize(mean=self.mean, std=self.std),
-            _RandomNCrop(self.patch_size, batch_size),
+            _ExtractPatches(window_size=self.patch_size),
+            data_keys=None,
+            keepdim=True,
+            same_on_batch=True,
+        )
+        self.train_aug = K.AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
+            K.RandomCrop(self.patch_size, pad_if_needed=True),
```
adamjstewart, line 59: Do we need `pad_if_needed` or is that just because of our fake test data being small?

adamjstewart (commented):
Due to the bugs for `batch_size=1` and lack of support for bboxes, I'm strongly considering using `RandomCrop` for train and `CenterCrop` for val/test (with large patch size and small batch size) instead of replacing our hacky `_RandomNCrop` with another hacky `_ExtractPatches`. Then we can find time to contribute something better upstream to Kornia. How would you feel about that?

torchgeo/datamodules/deepglobelandcover.py
isaaccorley: Yes, or if the image size is not evenly divisible by the patch size it will pad appropriately in the bottom right of the patch.

isaaccorley:
I think _ExtractPatches is fine and likely more correct, particularly for datasets like LEVIR-CD where patch size is typically 256x256 and the image size is 1024x1024 so you're validation and test metrics will actually encompass the entire test sets. Using CenterCrop this would not be the case and not be able to provide an apples to apples comparison with other baselines.

torchgeo/datamodules/inria.py
adamjstewart: Would like more clarification on this. Does this bug affect all of our datasets when `batch_size=1`? Is this a bug in `_ExtractPatches` or somewhere else? Can this be fixed in `_ExtractPatches`? Might be fewer lines of code than patching it here.

isaaccorley: No this is before any augmentations happen when batch_size=1. It only seemed to happen for these 2 datasets and not any others in my tests.

isaaccorley: If I had to guess, this seems related to the masks not having a channel dimension but kornia requiring this dimension before augmentations.

However, because loss functions like CE require there to be no channel dimension in the mask so you have to squeeze this out after augmentations.

adamjstewart: Doesn't `keepdim=True` solve this?

isaaccorley: Not if the mask didn't originally have a channel dim to start with unfortunately

adamjstewart: But we don't have a single dataset in TorchGeo that has a mask channel dimension. Why are only these 2 datasets broken?

adamjstewart: I'm not surprised it's broken, I'm just surprised only 2 datasets are broken and not all of them, and want to know why it's broken.

isaaccorley: Yeah I was surprisingly struggling to figure out why other datasets I changed like LEVIR-CD worked fine but these 2 just didn't for whatever reason.

isaaccorley: I spent way too much time debugging this, so basically this is also broken for LEVIR-CD but it's hidden because we have a transform that unsqueezes the mask channel dim in the tests [here](https://github.com/microsoft/torchgeo/blob/685680960d286333ffb726d945ee7954c17675fc/tests/datamodules/test_levircd.py#L15) which was added a few years ago by yours truly.

So I think the correct solution would be to add in the channel dimension like I've done in this PR.

adamjstewart: Alright, let's just get this in and worry about a better solution later. I really want to upstream these to Kornia from now on. We clearly aren't able to maintain our own transforms (myself included).

adamjstewart (approved):

PR #2686, opened 2025-03-28
Copernicus-Pretrain dataset

wangyi111:
This PR adds the [Copernicus-Pretrain](https://github.com/zhu-xlab/Copernicus-FM) dataset in the paper ["Towards a Unified Copernicus Foundation Model for Earth Vision"](https://arxiv.org/abs/2503.11849), which is an extension of the [SSL4EO-S12](https://github.com/zhu-xlab/SSL4EO-S12) dataset to all major Sentinel missions (S1-S5P).

This PR creates a Pytorch [IterableDataset](https://pytorch.org/docs/stable/data.html#torch.utils.data.IterableDataset) for Copernicus-Pretrain (no indexing), utilizing the [WebDataset](https://github.com/webdataset/webdataset) format and library for efficient data loading.

pretrain


wangyi111:
Not sure how to properly deal with webdataset / iterabledataset, just a very simple draft here. Also, although I have simplified the data size (# local patches, # timestamps), the generated fake test data `example-000000.tar` is still quite big (15M).

torchgeo/datasets/copernicus/pretrain.py
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Initialize a new CopernicusPretrain instance.
+
+        Args:
+            *args: Arguments passed to WebDataset base class.
+            **kwargs: Keyword arguments passed to WebDataset base class.
+        """
+        wds = lazy_import('webdataset')
```
adamjstewart, line 103: If we didn't need to mess with lazy import, we could theoretically make CopernicusPretrain a direct subclass of WebDataset. Maybe we could do something like:
```python
try:
    from webdataset import WebDataset

    class CopernicusPretrain(WebDataset):
        ...
except ModuleNotFoundError:
    pass
```
We would no longer be able to automatically handle shuffle/decode/select/map, but maybe that's okay.

tests/data/copernicus/pretrain/data.py
```diff
@@ -0,0 +1,90 @@
...
+        f'{grid_id}.s1_grd.pth': torch.rand(2, 2, 2, 264, 264, dtype=torch.float32),
+        f'{grid_id}.s2_toa.pth': torch.rand(2, 2, 13, 264, 264, dtype=torch.float32).to(
+            torch.int16
+        ),
+        f'{grid_id}.s3_olci.pth': torch.rand(2, 21, 96, 96, dtype=torch.float32),
+        f'{grid_id}.s5p_co.pth': torch.rand(2, 1, 28, 28, dtype=torch.float32),
+        f'{grid_id}.s5p_no2.pth': torch.rand(2, 1, 28, 28, dtype=torch.float32),
+        f'{grid_id}.s5p_o3.pth': torch.rand(2, 1, 28, 28, dtype=torch.float32),
+        f'{grid_id}.s5p_so2.pth': torch.rand(2, 1, 28, 28, dtype=torch.float32),
+        f'{grid_id}.dem.pth': torch.rand(1, 960, 960, dtype=torch.float32),
```
adamjstewart, line 73: TODO: check whether this is supposed to be (1, 960, 960) or (960, 960). The example 100 image version of the dataset seems to use the latter, but that could be a bug.

wangyi111: Just checked, it's the latter. The test dem should be (960,960).

torchgeo/datasets/copernicus/pretrain.py
wangyi111: never thought about this, in principle true..

calebrob6:
I'm down to merge this as its cool and seems to work.

That said, putting down some thoughts and questions for @wangyi111 / @adamjstewart (we don't need to address these now, but to keep in mind):
- @wangyi111 the docs say this is a reduced dataset ("For simplicity, the current dataset class provides a minimum example:"). Is this what you trained on in the paper though? As a user who is, e.g, trying to reproduce your work, I would like to understand how this artifact relates to the results reported in the paper.
- This has sample keys that are incompatible with our trainers by default. It'd be nice to have an accompanying transformer suite that allows something like (and I'm just sketching out here, I know this syntax doesn't work:
```
transforms = T.Compose([
   ProductPicker(["s1", "s3", "s5p_no2"]),
   TimePicker("concatenate")
])
ds = CopernicusPretrain(transforms=transforms)
```
that creates a `sample["image"]` for use as normal. Maybe this doesn't work for a lot of reasons though.
- Are WebDatasets loading from HF fast? To saturate a GPU with remote loading I usually have to co-locate my compute in the same datacenter as the data. If someone is trying this out from Colab, or their laptop, what throughput can they expect in terms of `batches of 32 256x256 samples/sec`? As a user looking to pre-train a model, the first thing I do with a dataset is check to see what sort of throughput I can get.

My first reaction seeing this is "if this is really so customized, should it be in torchgeo? similarly, would a 'full-fledged' version in a standalone repo be more beneficial to the community?" Put differently, if I'm a user that is looking to train on Copernicus-Pretrain, am I ever going to use this code or am I going to use something else anyway? If, you think "I'd use something else anyway", then we would be wasting time/effort keeping this updated in torchgeo.

calebrob6 (approved):

adamjstewart:
Merging due to the release deadline, but we can always make this more powerful/flexible in 0.7.1. We would eventually like all of our pre-training datasets to migrate to WebDataset or LitData. Let's do some benchmarking experiments and see which is faster/more convenient. I personally prefer LitData more because of who makes it.

wangyi111:
> Is this what you trained on in the paper though?

Yes. "reduced" means a sampling of the full dataset (all local patches of S1/2, all time stamps of all modalities loaded at once). Although if training long enough, the random sampling can cover all data.

> sample keys that are incompatible with our trainers by default.

Yeah this is tricky. May need a unique training pipeline compared to standard datasets. This may apply to more pretraining datasets in the future.

> Are WebDatasets loading from HF fast? 

Loading from HF is not that stable from my experience (S3 was much better). Can be pretty slow sometimes (also depends on the hyperparameters of shard shuffling etc). I would still recommend loading from disk (or enabling cache) in general, but loading some examples with streaming is acceptable. 

E.g., loading the example_100_grid subset on Colab, batch size 4, dataset length 100 (each sample has all 8 modalities with various shape [2,C,H,W]), for 10 epochs:
- loading from disk: 2.2 it/s (2min in total)
- loading from HF: 1.1 it/s (4min in total)

wangyi111:
Yes we need some proper benchmarking to test these large-scale data possibilities.

wangyi111:
> Am I ever going to use this code or am I going to use something else anyway?

I think this is an important question to think about for current and future pretraining datasets in general, when they go to multimodal, multitemporal, and more. The implementation here alone doesn't yet have much advantage over other customized codes. But if we will have a unified pipeline for all pretraining datasets, prove its efficiency, enable flexibility of sampling and even easy merging of different pretraining datasets, it would be really valuable.

PR #2687, opened 2025-03-29
Models: fix features_only loading, avoid redundant weight tests

adamjstewart:
TorchGeo now has a whopping 93 builtin pre-trained model weights. These weights have become prohibitively expensive to test, resulting in a significant slowdown and intermittent test failures. This PR is designed to resolve this issue.

### Before

At the moment, we test all model weights roughly 6 times: with `features_only` True and False, and for weight-loading, bands, and transforms. This means we are instantiating and saving large models to disk (sometimes as large as ViT-H) $$93 \times 6 = 576$$ times per OS, per Python version, and per commit. This adds up very quickly:

```console
> pytest tests/models/
...
==== 666 passed, 93 deselected in 926.76s (0:15:26) ====
```

### After

We apply the following changes to reduce the amount of testing:

- [x] We only test a single weight per enum instead of all weights (93 -> 17)
- [x] We use the builtin weights instead of mocked weights when testing bands and transforms (6 -> 2)

The result is still 100% test coverage, with all weights still being tested on the release branch, but significantly less compute time and storage requirements for the average CI run:

```console
> pytest tests/models/
...
==== 380 passed, 93 deselected in 200.63s (0:03:20) ====
```
That's an 80% decrease, and shaves off 12 min from every test run!

One other change is to create our test weights without considering `features_only` but test that those weights can be loaded with or without `features_only`. This caught several bugs in #2659 @lccol

P.S. There is likely still opportunity to further speed this up by converting the fixtures from function to class scope, but `monkeypatch` and `tmp_path` do not support this.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -43,7 +43,7 @@ jobs:
         run: pip list
       - name: Run pytest checks
         run: |
-          pytest --cov=torchgeo --cov-report=xml
+          pytest --cov=torchgeo --cov-report=xml --durations=10
           python3 -m torchgeo --help
       - name: Report coverage
         uses: codecov/codecov-action@v5.4.0
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -43,7 +43,7 @@ jobs:
         run: pip list
       - name: Run pytest checks
         run: |
-          pytest --cov=torchgeo --cov-report=xml --durations=10
+          pytest --cov=torchgeo --cov-report=xml
           python3 -m torchgeo --help
       - name: Report coverage
         uses: codecov/codecov-action@v5.4.0
--- a/torchgeo/models/vit.py
+++ b/torchgeo/models/vit.py
@@ -598,7 +598,12 @@ def vit_small_patch16_224(
         )
         assert set(missing_keys) <= {'head.weight', 'head.bias'}
         # used when features_only = True
-        assert set(unexpected_keys) <= {'norm.weight', 'norm.bias'}
+        assert set(unexpected_keys) <= {
+            'norm.weight',
+            'norm.bias',
+            'head.weight',
+            'head.bias',
+        }
 
     return model
 
@@ -641,7 +646,12 @@ def vit_base_patch16_224(
             weights.get_state_dict(progress=True), strict=False
         )
         assert set(missing_keys) <= {'head.weight', 'head.bias'}
-        assert set(unexpected_keys) <= {'norm.weight', 'norm.bias'}
+        assert set(unexpected_keys) <= {
+            'norm.weight',
+            'norm.bias',
+            'head.weight',
+            'head.bias',
+        }
 
     return model
 
@@ -684,7 +694,12 @@ def vit_large_patch16_224(
             weights.get_state_dict(progress=True), strict=False
         )
         assert set(missing_keys) <= {'head.weight', 'head.bias'}
-        assert set(unexpected_keys) <= {'norm.weight', 'norm.bias'}
+        assert set(unexpected_keys) <= {
+            'norm.weight',
+            'norm.bias',
+            'head.weight',
+            'head.bias',
+        }
 
     return model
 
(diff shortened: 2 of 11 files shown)


torchgeo/models/vit.py
```diff
@@ -727,7 +742,12 @@ def vit_huge_patch14_224(
             weights.get_state_dict(progress=True), strict=False
         )
         assert set(missing_keys) <= {'head.weight', 'head.bias'}
-        assert set(unexpected_keys) <= {'norm.weight', 'norm.bias'}
+        assert set(unexpected_keys) <= {
+            'norm.weight',
```
calebrob6, line 746: Why did these change?

.github/workflows/tests.yaml
```diff
@@ -43,7 +43,7 @@ jobs:
         run: pip list
       - name: Run pytest checks
         run: |
-          pytest --cov=torchgeo --cov-report=xml
+          pytest --cov=torchgeo --cov-report=xml --durations=10
```
calebrob6, line 46: What is the purpose of this?

isaaccorley (dismissed):

calebrob6 (approved):
Is there a way we can make the tests auto-fail if they go longer than 10 minutes? That way we know when we're getting into "there should be a better way to do this" territory.

adamjstewart:
> Is there a way we can make the tests auto-fail if they go longer than 10 minutes? That way we know when we're getting into "there should be a better way to do this" territory.

On an entire test suite level: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes
On an individual unit test level: https://pypi.org/project/pytest-timeout/

But I'd rather not do that, I think there are enough active core developers to notice when things become prohibitively slow.

torchgeo/models/vit.py
adamjstewart: Previously, we tested whether weights created with `features_only=True` could be loaded with `features_only=True` (and same for False).

Now, we test whether weights created with `features_only=False` can be loaded with `features_only=True` (or False).

I believe all of the pretrained weights we host were made with `features_only=False`, but can't be certain. If we want to be really sure, we could invert things to test True/False and False/True.

.github/workflows/tests.yaml
adamjstewart: This displays the 10 slowest tests so we know where to focus future energy:
```
============================= slowest 10 durations =============================
16.82s call     tests/trainers/test_instance_segmentation.py::TestInstanceSegmentationTask::test_trainer[True-vhr10_ins_seg]
11.07s call     tests/trainers/test_detection.py::TestObjectDetectionTask::test_trainer[True-retinanet-reforestree]
11.00s call     tests/trainers/test_detection.py::TestObjectDetectionTask::test_trainer[True-fcos-vhr10_obj_det]
10.95s call     tests/trainers/test_detection.py::TestObjectDetectionTask::test_trainer[True-retinanet-vhr10_obj_det]
10.93s call     tests/trainers/test_detection.py::TestObjectDetectionTask::test_trainer[True-fcos-reforestree]
10.92s call     tests/trainers/test_detection.py::TestObjectDetectionTask::test_trainer[True-retinanet-nasa_marine_debris]
10.81s call     tests/trainers/test_detection.py::TestObjectDetectionTask::test_trainer[True-faster-rcnn-reforestree]
10.81s call     tests/trainers/test_detection.py::TestObjectDetectionTask::test_trainer[True-fcos-nasa_marine_debris]
10.80s call     tests/trainers/test_detection.py::TestObjectDetectionTask::test_trainer[True-faster-rcnn-vhr10_obj_det]
10.64s call     tests/trainers/test_detection.py::TestObjectDetectionTask::test_trainer[True-faster-rcnn-nasa_marine_debris]
```
Could remove though, not strictly needed.

torchgeo/models/vit.py
adamjstewart: Testing True/False and False/True actually revealed more bugs, let me make this change and fix.

.github/workflows/tests.yaml
adamjstewart: Removed.

PR #2690, opened 2025-03-31
Update loss function to support all task modes for Jaccard and Focal loss

japanj:
This PR is created for updating Jaccard and Focal loss to support all task modes ('binary', 'multiclass', and 'multilabel')


adamjstewart:
If you run `ruff format` it should automatically fix the line length issue.

adamjstewart (approved):

PR #2699, opened 2025-04-04
Update FTW documentation

### LINKED ISSUE #2696: Number of Classes in FieldsOfTheWorld Dataset
### Description

I am iterating through the 3-class version of the FieldsOfTheWorldDataset and noticing that for the country of brazil and sometimes india the max label value is 3, so others return [0, 1, 2] but for brazil [1, 2, 3] are returned as mask labels. 

Do you perhaps have any insights @calebrob6 ?

### Steps to reproduce

```python

ds = FieldsOfTheWorld(root="FieldsOfTheWorld", target='3-class', countries=CC_BY_COUNTRIES)

dl = DataLoader(ds, batch_size=64, num_workers=8)

for batch in tqdm(dl):
    if batch["mask"].max() > 2:
        print("Max value in mask is greater than 2")
```

or check these two examples directly:

```python
path = "FieldsOfTheWorld/brazil/label_masks/semantic_3class/g3_0000011264-0000001536.tif"
path = "FieldsOfTheWorld/india/label_masks/semantic_3class/g1_1799.tif"

    mask = src.read(1)
    print(mask.max())
    print(np.unique(mask))

```

### Version

'0.7.0.dev0'

calebrob6:
Closes #2696

### DIFF
--- a/torchgeo/datasets/ftw.py
+++ b/torchgeo/datasets/ftw.py
@@ -43,9 +43,10 @@ class FieldsOfTheWorld(NonGeoDataset):
 
     Dataset classes:
 
-    1. background
-    2. field
-    3. field-boundary (three-class only)
+    0. background
+    1. field
+    2. field-boundary (three-class only)
+    3. unlabeled (kenya, rwanda, brazil and india have presence only labels)
 
     If you use this dataset in your research, please cite the following paper:
 


adamjstewart (commented):
Should we update `docs/api/datasets/non_geo_datasets.csv` too?

adamjstewart (approved):

PR #2706, opened 2025-04-07
Fix typo in geospatial tutorial

canmike:
Clarifies satellite orbit types and their typical use cases in time-series applications.


adamjstewart (approved):
Nice catch!

Issue #2060, opened 2024-05-15
class_weights cannot be passed via config file as a tensor is expected


robmarkcole:
### Description

Using the Lightning CLI we can train the SemanticSegmentationTask, but cannot use class_weights without an error. Solution is to accept a list if int in addition to tensor

### Steps to reproduce

In Lightning CLI Yaml:
```
model:
  class_path: SemanticSegmentationTask
  init_args:
    model: unet
    backbone: resnet50
    weights: null
    lr: 0.001
    in_channels: 6
    num_classes: 2
    class_weights:
      - 1
      - 50
  ```
Will result in 
```
      Does not validate against any of the Union subtypes
      Subtypes: (<class 'torch.Tensor'>, <class 'NoneType'>)
      Errors:
        - Not a valid subclass of Tensor
          Subclass types expect one of:
          - a class path (str)
          - a dict with class_path entry
          - a dict without class_path but with init_args entry (class path given previously)
        - Expected a <class 'NoneType'>
      Given value type: <class 'list'>
      Given value: [1, 50]
```

### Version

main

**Comment by isaaccorley** [comment]
Care to make a PR to accept a list and convert to a tensor? If not then I can take it on this weekend.

**Comment by robmarkcole** [comment]
You will get to it way before me!

**Comment by adamjstewart** [comment]
For a bit of history, I added this in #1221 and it initially only supported lists. In #1413, @ntw-au modified this to support lists, numpy arrays, and torch tensors. Then in #1541, I modified it to only accept torch tensors. I agree we need a way to support `class_weights` in a YAML file (and preferably also on the command line). If omegaconf supports this, we could also easily enable omegaconf as a parser: https://lightning.ai/docs/pytorch/stable/cli/lightning_cli_advanced_2.html#enable-variable-interpolation.

**Comment by isaaccorley** [comment]
I haven't looked at the Lightning CLI in awhile but I wonder if it supports recursive instantiation like

```
class_weights:
  class_path:
    torch.tensor
  init_args:
    data: [0.5, 0.5]
```

**Comment by robmarkcole** [comment]
@isaaccorley that results in 
```
    - Problem with given class_path 'torchgeo.trainers.SemanticSegmentationTask':
        Parser key "class_weights":
          Does not validate against any of the Union subtypes
          Subtypes: [<class 'NoneType'>, <class 'torch.Tensor'>]
          Errors:
            - Expected a <class 'NoneType'>
[...]

PR #2707, opened 2025-04-07
Update class_weights to support list[float]

robmarkcole:
Close https://github.com/microsoft/torchgeo/issues/2060

### DIFF
--- a/torchgeo/trainers/classification.py
+++ b/torchgeo/trainers/classification.py
@@ -9,6 +9,7 @@ from typing import Any, Literal
 import kornia.augmentation as K
 import matplotlib.pyplot as plt
 import timm
+import torch
 import torch.nn as nn
 from matplotlib.figure import Figure
 from segmentation_models_pytorch.losses import FocalLoss, JaccardLoss
@@ -36,7 +37,7 @@ class ClassificationTask(BaseTask):
         num_classes: int | None = None,
         num_labels: int | None = None,
         loss: Literal['ce', 'bce', 'jaccard', 'focal'] = 'ce',
-        class_weights: Tensor | None = None,
+        class_weights: Tensor | list[float] | None = None,
         lr: float = 1e-3,
         patience: int = 10,
         freeze_backbone: bool = False,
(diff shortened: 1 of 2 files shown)


robmarkcole:
Not sure why docs fail to build 
image

adamjstewart:
Could use https://docs.python.org/3/library/warnings.html#warnings.deprecated or explicitly raise https://docs.python.org/3/library/exceptions.html#DeprecationWarning for deprecation.

I'm not sure why the doc tests are failing either, but you're not the only PR with failing tests. Will try to investigate this later if it keeps happening.

adamjstewart:
Restarting the docs build fixed the other PR, hoping this is just a transient issue.

robmarkcole:
Hmm did this change break this test? 
```
FAILED tests/models/test_copernicusfm.py::TestCopernicusFMBase::test_copernicusfm_variable - AssertionError: The input tensor is not within the configured range `[0.001, 510000000.0]`.
```

adamjstewart:
Nope, this is because I used random numbers which are sometimes outside the valid range. Will submit a fix. If you rerun it should only fail 1% of the time.

adamjstewart:
Yes, just hit it. I don't think you have access, but I'm happy to rerun until CI is happy.

robmarkcole:
If you could - I dont see the button 🥹

robmarkcole:
@adamjstewart thoughts?

adamjstewart:
Let's discuss more on #2060

robmarkcole:
@adamjstewart any further thoughts?

isaaccorley:
Let's support both lists and tensors only (not arrays) and not deprecate tensor support.

I can see the case for both:
- support tensors for backwards compat
- support lists so weights can be passed in through configs with hydra

adamjstewart:
_Every_ change needs a test 🙂

isaaccorley (dismissed):
Lgtm

adamjstewart (commented):
Same for both tasks.

isaaccorley (approved):

isaaccorley:
Let's merge tomorrow if no other comments/concerns with this PR

adamjstewart (approved):

Issue #2377, opened 2024-10-29
Ensure all datasets have a plot method


adamjstewart:
### Summary

We should ensure all datasets have a plot method.

### Rationale

This gives users a consistent API with which to plot and visualize data.

### Implementation

The following datasets are missing a plot method:

- [x] EDDMapS
- [x] GBIF
- [x] iNaturalist
- [x] MMEarth
- [x] Western USA Live Fuel Moisture

Once these datasets have a plot method, we should add:
```python
@abc.abstractmethod
def plot(...):
```
to `GeoDataset` and `NonGeoDataset` to make sure no new datasets are added without a plot method.

Note that we may need to add default `plot` methods to the following subclasses to make sure they remain useful:

- [ ] RasterDataset
- [ ] VectorDataset
- [ ] NonGeoClassificationDataset
- [ ] IntersectionDataset: #1971
- [ ] UnionDataset

amrirasyidi:
Hi, I would like to work on this issue
This will be my first time contributing, can you tell me first thing to do so I can pick up the rest? (I have checked plot method in copernicus and Chesapeake dataset)

Thanks!

adamjstewart:
Hi @amrirasyidi, let's start by adding plot methods to the above list of datasets that are missing one. Want to start with EDDMapS, GBIF, and iNaturalist? They are all point datasets that are largely identical, so the plot method should be the same for all 3. I'm not exactly sure what a good plot method would look like for these datasets, but try downloading and visualizing the data. Note that we are trying to plot a sample from the dataset, not the entire dataset.

amrirasyidi:
Cool, I'll start with those 3 then.

Just want to clarify on this
> Note that we are trying to plot a sample from the dataset, not the entire dataset.

By "sample" you mean a subset of the entire dataset the user have on their storage right (user has to download the data first, then provide the path)? I.e. not the test dataset in `/test/data`

amrirasyidi:
Hi @adamjstewart 
I want to continue adding plot method to `MMEarth`, this one has three different versions of the dataset and 12 modalities.
Do you have an example of dataset that has also multiple version and/or modalities?

If not, do you have a suggestion on how to approach this?
Thanks!

adamjstewart:
SatlasPretrain is another good example of this, basically just make subplots for each modality. @nilsleh may also be able to help, he wrote the initial MMEarth dataset implementation.

amrirasyidi:
Thanks for the lead!

[...]

PR #2709, opened 2025-04-07
Plot Method Added to EDDMapS Dataset

amrirasyidi:
Hi everyone, this is my first contribution ever, I hope this is suffice

- Relevant issue: #2377 
- Added contextily as dependecy to plot the basemap
- Added plot method to eddmaps dataset. The plot will take an object returned by `__getitem__` method and will plot it accordingly based on the longitude, latitude, and obsdate value.

Happy to hear your feedback if any

### DIFF
--- a/torchgeo/datasets/eddmaps.py
+++ b/torchgeo/datasets/eddmaps.py
@@ -5,24 +5,13 @@
 
 import os
 import sys
-from datetime import datetime
 from typing import Any
 
-import matplotlib.dates as mdates
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
-from matplotlib.colors import Normalize
 from matplotlib.figure import Figure
 from rasterio.crs import CRS
-
-# Make contextily optional
-try:
-    import contextily as ctx
-
-    HAS_CONTEXTILY = True
-except ImportError:
-    HAS_CONTEXTILY = False
 
 from .errors import DatasetNotFoundError
 from .geo import GeoDataset
@@ -117,18 +106,11 @@ class EDDMapS(GeoDataset):
 
         return sample
 
-    def plot(
-        self,
-        sample: dict[str, Any],
-        show_titles: bool = True,
-        suptitle: str | None = None,
-    ) -> Figure:
+    def plot(self, sample: dict[str, Any]) -> Figure:
         """Plot a sample from the dataset.
 
         Args:
             sample: a sample return by :meth:`__getitem__`
-            show_titles: flag indicating whether to show titles above each panel
-            suptitle: optional suptitle to use for figure
 
         Returns:
             a matplotlib Figure with the rendered sample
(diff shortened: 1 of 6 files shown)


amrirasyidi:
> Can this be done without contextily? We really don't want to add an additional dependency just for a single dataset. Also, can you share a picture of the resulting plot?

Yes, it can, but since this is a geographical data, it will be less intuitive without a basemap.
I haven't really checked it, but since GBIF and iNaturalist are also point dataset, I imagine they would also be benefitted with a basemap plotted. But I can just remove contextily along with the basemap if you prefer so, just let me know.

This one is the example with basemap
image

This one without
image

isaaccorley:
Not sure if any of our other dependencies include basemaps but contextily is used as the default basemap getter in the geopandas docs so this makes sense to me.

adamjstewart:
I've only ever used cartopy before, but I know there are a ton of geographic mapping libraries.

amrirasyidi:
> I've only ever used cartopy before, but I know there are a ton of geographic mapping libraries.

I also considered cartopy, but contextily is much lighter than cartopy, and I know adding dependencies wont be easily approved, hence my choice. Upon quick test, cartopy took ~2.5s to generate similar map as above, while contextily took ~0.8s.
Cartopy generates a much better look, but for quick visual, I think contextily will do.

amrirasyidi:

amrirasyidi:
Hi @adamjstewart , need approval if everything looks good

adamjstewart:
Could also make it an optional dependency. It would be used in ~3 datasets at the moment. I generally try really hard not to add new deps because otherwise TorchGeo would balloon very quickly. I guess the question is whether or not prettier plots is worth it?

amrirasyidi:
I've moved contextily in optional dataset dependency
But now the checks are stucked at "Expected"?

adamjstewart (changes requested):
I think I want to keep this simpler. The dataset went from 105 to 242 lines and the tests went from 48 to 185 lines of code. This is all stuff we will need to maintain if someone files a bug report.

For the `plot` method, let's try to keep this simple for now. So no basemap, no fancy ticks and formatters, no custom font size or weight, etc.

For the tests, the following should be sufficient to ensure that the plot method works:
```python
    def test_plot(self, dataset: EDDMapS) -> None:                                       
        x = dataset[dataset.bounds]                                                      
        dataset.plot(x, suptitle='Test')                                                 
        plt.close()
```
4 lines of code that does 90% of 137 lines of code is better for maintainability. If anyone wants super fancy plots, they can override and customize this method.

adamjstewart:
What does the new plot look like? Just want to double check.

amrirasyidi:
Here how it looks right now.

image

I'm still visualizing the observation date as is, I think it is nice to have a sense of time regarding the data.
Thanks for the feedback btw, I have to shift my "data scientist writing report" mindset it looks like :)
Let me know if you have other feedback

torchgeo/datasets/eddmaps.py
```diff
@@ -103,3 +105,50 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
         sample = {'crs': self.crs, 'bounds': bboxes}
 
         return sample
+
+    def plot(self, sample: dict[str, Any]) -> Figure:
+        """Plot a sample from the dataset.
+
+        Args:
+            sample: a sample return by :meth:`__getitem__`
```
adamjstewart, line 113: Needs the same `show_titles` and `suptitle` args as all other datasets

torchgeo/datasets/eddmaps.py
```diff
@@ -103,3 +105,50 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
...
+        scatter = ax.scatter(
+            longitudes,
+            latitudes,
+            c=timestamps,
+            cmap='coolwarm',
+            s=30,
+            alpha=0.8,
+            edgecolors='black',
+            linewidths=0.5,
+            zorder=3,
```
adamjstewart, line 142: What is zorder? Is this needed?

torchgeo/datasets/eddmaps.py
```diff
@@ -103,3 +105,50 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
...
+        fig, ax = plt.subplots(figsize=(10, 8))
+        ax.grid(ls='--')
+
+        # Extract bounding boxes (coordinates) from the sample
+        bboxes = sample['bounds']
+
+        # Extract coordinates and timestamps
+        longitudes = [bbox[0] for bbox in bboxes]  # minx
+        latitudes = [bbox[1] for bbox in bboxes]  # miny
+        timestamps = [bbox[2] for bbox in bboxes]  # mint (timestamp)
```
adamjstewart, line 130: Does matplotlib support datetime64? Maybe we can convert these from float back to datetime so the colorbar is more legible.

adamjstewart (changes requested):
Couple minor changes left, then this is ready to go!

torchgeo/datasets/eddmaps.py
amrirasyidi: added in the new commit

torchgeo/datasets/eddmaps.py
amrirasyidi: I use it to ensure the drawing order because previously a basemap was also plotted. I forgot to remove it when removing the contextily.
Removed in the new commit

torchgeo/datasets/eddmaps.py
adamjstewart: What if we convert the dates to datetime first, then use the built-in DateFormatter?

amrirasyidi:
This is how it looks like right now
image

I also remove the customization part for the scatter, because it is modifiable anyway.

torchgeo/datasets/eddmaps.py
adamjstewart: Just waiting on this comment.

amrirasyidi: I got this warning while implementing the built-in function
```
mypy .
torchgeo/datasets/eddmaps.py:144: error: Call to untyped function "date2num" in typed context  [no-untyped-call]
                longitudes, latitudes, c=mdates.date2num(dates), edgecolors='black'
                                         ^~~~~~~~~~~~~~~~~~~~~~
torchgeo/datasets/eddmaps.py:153: error: Call to untyped function "DateFormatter" in typed context  [no-untyped-call]
            date_format = cast(Any, mdates.DateFormatter('%Y-%m-%d'))
                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Found 2 errors in 1 file (checked 420 source files)
```

amrirasyidi: Workaround is probably to import typing and cast them
Or add `type: ignore` there

what do you think?

adamjstewart: What about:
```python
from datetime import datetime

dates = map(datetime.fromtimestamp, timestamps)
```
Then it should have type hints.

amrirasyidi: I can directly access the ticks to convert them like this instead of creating a custom function.
```
# Set new formatted date labels
ticks = cbar.get_ticks()
tick_labels = [datetime.fromtimestamp(tick).strftime('%Y-%m-%d') for tick in ticks]
cbar.ax.set_yticklabels(tick_labels)
```

Because, I think the problem is in the `mdates` (`matplotlib.dates`), mypy doesnt recognize its type hints.

```
dates = list(map(datetime.fromtimestamp, timestamps))
# Plot the points with colors based on date
scatter = ax.scatter(longitudes, latitudes,
                     c=mdates.date2num(dates),
                     edgecolors='black')
```
```
> mypy .
torchgeo/datasets/eddmaps.py:143: error: Call to untyped function "date2num" in typed context  [no-untyped-call]
                                 c=mdates.date2num(dates),
                                   ^~~~~~~~~~~~~~~~~~~~~~
Found 1 error in 1 file (checked 420 source files)
```
If I use `timestamps` directly while creating `scatter`, I still have to convert the label into a readable format. This way using the built-in function `mdates.DateFormatter` will throw an error.

amrirasyidi: Hi @adamjstewart need your advice on this.
Thanks!

adamjstewart: Sorry, was on vacation the last few days. Let me investigate...

adamjstewart: Couldn't find a cleaner way to do this. Let's stick with your solution for now.

adamjstewart (approved):

adamjstewart:
Thanks for the PR! If you also want to submit PRs for GBIF and iNaturalist, they should be almost identical.

amrirasyidi:
> Thanks for the PR! If you also want to submit PRs for GBIF and iNaturalist, they should be almost identical.

Thanks for the approval.
Sure, I'll continue with GBIF and iNaturalist as well.

PR #2710, opened 2025-04-07
CopernicusFM: ensure test metadata is always within range

adamjstewart:
@robmarkcole encountered the following issue in #2707:
```
FAILED tests/models/test_copernicusfm.py::TestCopernicusFMBase::test_copernicusfm_variable - AssertionError: The input tensor is not within the configured range `[0.001, 510000000.0]`.
```
The problem is that random numbers between 0 and 1 are sometimes 0, which is outside the valid range. This PR adds 1 to ensure that it is always in range.


isaaccorley (approved):
lgtm

PR #2719, opened 2025-04-10
FTW U-Net Models

isaaccorley:
This PR adds the 4 pretrained smp U-Net models from the FTW project.

This is our first pretrained U-Net weights so some thoughts I have are `UNet` or `Unet`? smp instantiates the class like `smp.Unet`.

Also the `ftw-baselines` repo doesn't indicate what version of commercial and noncommercial licenses they use, just that it's CC BY. @calebrob6 can you confirm this for me? I'm assuming it's `CC BY 4.0` and `CC BY-NC 4.0`.

I added support for using a different number of output classes since I imagine users will either want to a) use the pretrained model for the task it was trained on or b) use the pretrained weights for some new task with a possibly different number of output classes.

Sample plots:

image

image


calebrob6:
It is a mix -- see individual licenses here https://source.coop/repositories/kerner-lab/fields-of-the-world/description

isaaccorley:
@adamjstewart can you suggest a fix here for mypy not liking that I added the unet model to the models api?

```bash
torchgeo/models/api.py:88: error: Cannot call function of unknown type 
[operator]
        model: nn.Module = _model[name](*args, **kwargs)
                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

calebrob6:
> can you suggest a fix here for mypy not liking that I added the unet model to the models api?

Works for me when I remove lines 15-40 from https://github.com/microsoft/torchgeo/blob/main/.github/workflows/style.yaml 👍

isaaccorley:
> @adamjstewart can you suggest a fix here for mypy not liking that I added the unet model to the models api?
> 
> ```shell
> torchgeo/models/api.py:88: error: Cannot call function of unknown type 
> [operator]
>         model: nn.Module = _model[name](*args, **kwargs)
>                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```

Nvm fixed this just by adding type hints to the `_model` dict

nilsleh (approved):

torchgeo/models/unet.py
```diff
@@ -0,0 +1,153 @@
...
+        state_dict = weights.get_state_dict(progress=True)
+
+        # Load full pretrained model
+        if kwargs['classes'] == weights.meta['num_classes']:
+            missing_keys, _ = model.load_state_dict(state_dict, strict=True)
+        # Random initialize segmentation head for new task
+        else:
+            del state_dict['segmentation_head.0.weight']
+            del state_dict['segmentation_head.0.bias']
+            missing_keys, _ = model.load_state_dict(state_dict, strict=False)
```
adamjstewart, line 147: Can we also test unexpected keys here?

torchgeo/models/unet.py
```diff
@@ -0,0 +1,153 @@
...
+
+    * https://arxiv.org/abs/1505.04597
+
+    .. versionadded:: 0.8
+
+    Args:
+        weights: Pre-trained model weights to use.
+        classes: Number of output classes. If not specified, the number of
+            classes will be inferred from the weights.
+        *args: Additional arguments to pass to :func:`segmentation_models_pytorch.create_model`
```
adamjstewart, line 119:
```suggestion
        *args: Additional arguments to pass to ``segmentation_models_pytorch.create_model``
```
Then we don't have to ignore the lack of docs. Alternatively, could try to add docs.

adamjstewart (commented):
Looks great, just minor comments

docs/conf.py
```diff
@@ -69,6 +69,8 @@
     ('py:class', 'torchvision.models._api.WeightsEnum'),
     ('py:class', 'torchvision.models.resnet.ResNet'),
     ('py:class', 'torchvision.models.swin_transformer.SwinTransformer'),
+    # Undocumented functions
+    ('py:func', 'segmentation_models_pytorch.create_model'),
```
adamjstewart, line 73: This can be removed now

adamjstewart (dismissed):

adamjstewart:
Can you resolve the merge conflicts?

adamjstewart (approved):

PR #2733, opened 2025-04-15
Models API Incomplete

isaaccorley:
It seems that as we've been adding models, we've forgotten to add them to the torchgeo.models.api. This PR fixes that.


adamjstewart:
😅

isaaccorley:
@adamjstewart Are there any weights for these models? `dofa_huge_patch14_224, dofa_small_patch16_224`. Or is it just to support if someone wants to create a small or huge variant?

tests/models/test_api.py
```diff
@@ -9,50 +9,92 @@
...
     swin_v2_t,
+    vit_base_patch14_dinov2,
+    vit_base_patch16_224,
+    vit_huge_patch14_224,
+    vit_large_patch16_224,
+    vit_small_patch14_dinov2,
     vit_small_patch16_224,
 )
 
+models_without_weights = [dofa_huge_patch14_224]
```
adamjstewart, line 56: Seems like this isn't used, or is just overridden?

adamjstewart (approved):
To be honest, I've never really been in love with torchvision's model API design. If we can find ways to register these models as timm backbones like TerraTorch does, much of this could become obsolete.

PR #2734, opened 2025-04-15
Related Libraries: DeepForest used to support TensorFlow

adamjstewart:
Just discovered that DeepForest 0.X was TensorFlow-based and transitioned to PyTorch in 1.X: https://github.com/weecology/DeepForest/releases/tag/1.0.0

Should we continue to document historical support, or should we only document the latest backend? For example, TorchGeo is planning on switching from R-tree to geopandas.

### DIFF
--- a/docs/user/alternatives.rst
+++ b/docs/user/alternatives.rst
@@ -26,7 +26,7 @@ Features
    :header-rows: 1
    :widths: auto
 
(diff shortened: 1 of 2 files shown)


isaaccorley (approved):

isaaccorley:
I think the asterisk makes sense

PR #2735, opened 2025-04-15
FTW Plot Typo

isaaccorley:


adamjstewart (approved):

calebrob6 (approved):

calebrob6:
I don't remember which one we use, but considering you're changing it I'm assuming you ran into a problem :)

isaaccorley:
> I don't remember which one we use, but considering you're changing it I'm assuming you ran into a problem :)

Yup all of our tests use "prediction" not "predictions" so it loses test coverage because the if statement never gets hit as True

adamjstewart:
I made codecov "required" again to prevent this situation. Hopefully it doesn't get flaky again.

isaaccorley:
I could have sworn that it passed when we merged. I just happened to see it fail on the main branch

PR #2742, opened 2025-04-18
Sat-SlideMix Augmentation

isaaccorley:
Adds Sat-SlideMix augmentation from the paper [Data Augmentation Approaches for Satellite Imagery](https://ojs.aaai.org/index.php/AAAI/article/view/35028)

I've made this into a Kornia augmentation so that it will work with both images and masks. This has some nice benefits like automatically sampling from a range of percentages to shift and random sampling which dim and direction to shift.

## Example usage:

```python
import requests
import torch
import torchvision.transforms.functional as TF
from torchgeo.transforms import SatSlideMix
from PIL import Image
from IPython.display import display


image_url = "https://project.inria.fr/aerialimagelabeling/files/2011/12/kit1.jpg"
mask_url = "https://project.inria.fr/aerialimagelabeling/files/2011/12/kit2.jpg"
image = Image.open(requests.get(image_url, stream=True).raw).convert("RGB").resize((512, 512))
mask = Image.open(requests.get(mask_url, stream=True).raw).convert("L").resize((512, 512))
image = TF.pil_to_tensor(image).unsqueeze(0)
mask = TF.pil_to_tensor(mask).unsqueeze(0)
print(image.shape, mask.shape)

aug = K.AugmentationSequential(SatSlideMix(gamma=2, beta=(0.0, 1.0), p=1.0), data_keys=None)

batch = dict(image=image.clone().repeat(2, 1, 1, 1)/255.0, mask=mask.clone().repeat(2, 1, 1, 1)/255.0)
print(batch["image"].shape, batch["mask"].shape)  # image shape: (2, 3, 512, 512)  mask shape: (2, 1, 512, 512)
batch = aug(batch)
print(batch["image"].shape, batch["mask"].shape)  # image shape: (4, 3, 512, 512)  mask shape: (4, 1, 512, 512)

for i in range(batch["image"].shape[0]):
    display(TF.to_pil_image(batch["image"][i]))
    display(TF.to_pil_image(batch["mask"][i]))
```

### Sample 1
image
image

### Sample 2
image
image

### Sample 3
image
image

### Sample 4
image
image


adamjstewart:
Can we implement this in a way that is compatible with Kornia?

isaaccorley:
> Can we implement this in a way that is compatible with Kornia?

Working on it

burakekim:
Do we refer to the paper in the docstring?

isaaccorley:
> Do we refer to the paper in the docstring? 

Not yes, still converting it to work with kornia images and masks. Will add it though.

torchgeo/transforms/augmentations.py
adamjstewart, line 1: @yichiac is working on several time-series augmentations that will end up in `torchgeo/transforms/temporal.py`. Do you think Sat-SlideMix would classify as something that could go in `torchgeo/transforms/spatial.py`? Or should we reserve that for things like reprojection? Doesn't make a difference to the users, just talking about the filename.

torchgeo/transforms/augmentations.py
```diff
@@ -0,0 +1,115 @@
...
+    .. versionadded:: 0.8
+    """
+
+    def __init__(
+        self,
+        gamma: int = 1,
+        beta: Tensor | float | tuple[float, float] | list[float] = (0.0, 1.0),
+        p: float = 0.5,
+    ) -> None:
+        """Initialize a new _ExtractPatches instance.
```
adamjstewart, line 32:
```suggestion
        """Initialize a new SatSlideMix instance.
```

adamjstewart (commented):
I'm trusting you that the implementation is correct, only reviewing for typos.

torchgeo/transforms/augmentations.py
```diff
@@ -0,0 +1,115 @@
...
+            beta: The range of percentage (0.0 to 1.0) of the image
+                dimension (height or width) to shift.
+            p: Probability to apply the augmentation on each sample
+
+        Raises:
+            AssertionError: If `gamma` is not a positive integer.
+        """
+        super().__init__(p=p)
+        assert isinstance(gamma, int) and gamma > 0, 'gamma must be a positive integer'
+        self._param_generator = K.random_generator.PlainUniformGenerator(
```
adamjstewart, line 46: Don't really understand why mypy doesn't like this line, let me know if you want me to do some digging.

isaaccorley: I think in some cases the attr for param generator is set to None in the base class. I'll take a look again today.

torchgeo/transforms/augmentations.py
adamjstewart: Could also have `intensity.py` and `geometric.py`. I don't really care too much, just bikeshedding.

isaaccorley:
cc @burakekim

adamjstewart (approved):
I assume the inverse transform works too?

isaaccorley:
> I assume the inverse transform works too?

I have not implemented this. Would need to think about this a bit more tbh. I'd prefer to leave that for another PR in the event that someone actually needs the inverse transform.

PR #2746, opened 2025-04-20
FTW: add uncompressed data

adamjstewart:
When running the segmentation tests, these files get extracted and are registered as untracked files by git.


calebrob6:
Thanks! I noticed this, too (but also I thought we didn't commit the uncompressed test data for some reason)

calebrob6 (approved):

PR #2747, opened 2025-04-21
GeoDataset: rtree -> geopandas

adamjstewart:
This PR migrates our GeoDataset indices from rtree to geopandas, and is part of ongoing work to add time series support to TorchGeo: #2382. 

### Progress

- [x] Dataset Base Classes
  - [x] GeoDataset
  - [x] RasterDataset
  - [x] VectorDataset
  - [x] IntersectionDataset
  - [x] UnionDataset
- [x] Custom Datasets
  - [x] AgriFieldNet
  - [x] Chesapeake CVPR
  - [x] EDDMapS
  - [x] EnviroAtlas
  - [x] GBIF
  - [x] GlobBiomass
  - [x] iNaturalist
  - [x] L7 Irish
  - [x] LandCover.ai
  - [x] MMFlood
  - [x] Open Buildings
  - [x] South Africa Crop Type
- [x] Splitters
  - [x] random_bbox_assignment
  - [x] random_bbox_splitting
  - [x] random_grid_cell_assignment
  - [x] roi_split
  - [x] time_series_split
- [x] Samplers
  - [x] GeoSampler
  - [x] RandomGeoSampler
  - [x] GridGeoSampler
  - [x] PreChippedGeoSampler
  - [x] BatchGeoSampler
  - [x] RandomBatchGeoSampler
- [x] Dependencies
  - [x] pyproject.toml
  - [x] requirements.txt
- [x] Documentation
  - [x] related libraries
  - [x] tutorials
- [x] Testing
  - [x] Code coverage
  - [x] Silence warnings

### Backwards-Incompatible Changes

I've tried to keep backwards-incompatible changes in this particular PR to a minimum, but there were a few that were unavoidable or minor:

* `GeoDataset` now uses pyproj CRS instead of rasterio CRS
* `GeoDataset` no longer provides a default `__init__` method
* Point datasets (EDDMapS, GBIF, iNaturalist) now return a GeoDataFrame instead of a list of bounding boxes
* Point datasets (EDDMapS, GBIF, iNaturalist) no longer handle partial timestamp information
* `BoundingBox` now takes datetime objects instead of floats for mint/maxt
* `roi_split` now includes geometries like LineString and Point with zero-area overlap

Some of these changes could be undone if necessary, but I expect to have many more backwards-incompatible changes in future PRs.

### Open Questions

1. How should we handle `VectorDataset` now that we can store arbitrary Polygons in the index?
2. Can we mix `IntervalIndex`, `PeriodIndex`, and `DatetimeIndex`?
3. Can we mix `date`, `datetime`, `pd.Timestamp`, `pd.Period`?
4. Can we mix `Point`, `Polygon`, etc.?
5. [CRS compatibility](https://pyproj4.github.io/pyproj/stable/crs_compatibility.html) between rasterio, fiona, geopandas, and pyproj?
6. Should spatial intersection with zero area count (point, edge)?
7. Should we define a standard set of column names, or support arbitrary additional columns (e.g., cloud percent)?
8. Can we unify `GeoDataset` and `NonGeoDataset` (and `TileDataset`)?

### Motivation

Switching from rtree to geopandas is no small feat. This decision was made after careful consideration of several options and alternatives. The decision to switch from rtree to geopandas was made due to geopandas support for the following features that rtree does not support (items in bold are critical for time series integration):

* Array features:
  * Integer indexing
  * Integer slicing
* List features:
  * **spatial aggregation**: with `index.dissolve`
  * **temporal aggretation**: with `index.groupby`
* Set features:
  * membership test
  * uniqueness: with `pandas.unique`
  * faster intersection/union
* Database features:
  * multitheading
  * pickling
  * redundancy
* Geospatial features:
  * reprojection: with `index.to_crs`
  * arbitrary shapely `Geometry`s: including `Point`s and non-rectangular `Polygon`s
* Geotemporal features:
  * **datetime64 objects**
* Visualization features:
  * plotting


torchgeo/datasets/geo.py
```diff
@@ -354,10 +301,10 @@ class RasterDataset(GeoDataset):
     date_format = '%Y%m%d'
 
     #: Minimum timestamp if not in filename
-    mint: float = 0
+    mint: datetime = pd.Timestamp.min
```
adamjstewart, line 304: Pandas defaults to `datetime64[ns]`, which has a much smaller range than `datetime`:
```pycon
>>> import pandas as pd
>>> pd.Timestamp.min
Timestamp('1677-09-21 00:12:43.145224193')
>>> pd.Timestamp.max
Timestamp('2262-04-11 23:47:16.854775807')
>>> datetime.min
datetime.datetime(1, 1, 1, 0, 0)
>>> datetime.max
datetime.datetime(9999, 12, 31, 23, 59, 59, 999999)
```
I can't imagine this being an issue, but something to be aware of.

adamjstewart:
I mostly have the dataset base classes working now. However, the following datasets are either custom or hacky, making the porting a bit tricky. I would appreciate help from the people who wrote these datasets:

- [ ] AgriFieldNet: @yichiac 
- [ ] Chesapeake CVPR: @calebrob6
- [ ] EDDMapS: @adamjstewart 
- [ ] EnviroAtlas: @calebrob6 
- [ ] GBIF: @adamjstewart 
- [ ] GlobBiomass: @nilsleh
- [ ] iNaturalist: @adamjstewart 
- [ ] L7 Irish: @yichiac 
- [ ] LandCover.ai: @adrianboguszewski 
- [ ] MMFlood: @lccol 
- [ ] Open Buildings: @nilsleh
- [ ] South Africa Crop Type: @GeorgeHuber

@isaaccorley maybe you can help with some of these as well.

calebrob6:
First question from me is about speed -- do we have a benchmark script to understand before/after lookup speed?

adamjstewart:
We have both https://torchgeo.readthedocs.io/en/latest/user/contributing.html#i-o-benchmarking (for TorchGeo) and https://github.com/zhu-xlab/spatiotemporal-indexing and (for rtree/geopandas/etc.).

adamjstewart:
That's the goal. There will likely be minor differences but we can correct those as we go.

torchgeo/datasets/geo.py
adamjstewart: May actually convert the types directly to pd.Timestamp, datetime-compatibility is not great...

torchgeo/datasets/geo.py
```diff
@@ -990,35 +953,33 @@ def __init__(
...
+
+        # Temporal intersection
+        datetime_1 = pd.IntervalIndex(self.index.pop('datetime_1'))
+        datetime_2 = pd.IntervalIndex(self.index.pop('datetime_2'))
+        mint = np.maximum(datetime_1.left, datetime_2.left)
+        maxt = np.minimum(datetime_1.right, datetime_2.right)
+        self.index.index = pd.IntervalIndex.from_arrays(
+            mint, maxt, closed='both', name='datetime'
+        )
+        # self.index = self.index[maxt > mint]
```
adamjstewart, line 980: Adding this line of code causes a couple tests to fail (point datasets, box edge overlap). But presumably this line of code is needed in cases where some files have spatial overlap but not temporal overlap? I don't think our tests yet exercise this, may need to add a new test.

adamjstewart:
@adrianboguszewski LandCover.ai Geo doesn't look too bad, but we need to update the `__getitem__` method to replace any R-tree specific stuff with the new syntax. If you want, you can open a PR off of my branch, push directly to my branch, or simply share what the code change would look like for the new geopandas backend. If you're busy, I can try to take a stab at it and you can double check that the changes work with the real dataset.

adrianboguszewski:
@adamjstewart yeah, little busy, so would be great if you create a PR and I test it :)

torchgeo/samplers/single.py
```diff
@@ -113,48 +108,32 @@ def __init__(
...
-
-            yield bounding_box
+            point = points.sample(random_state=self.generator)
+            minx = point.iloc[0].x - self.size[1] / 2
+            maxx = point.iloc[0].x + self.size[1] / 2
+            miny = point.iloc[0].y - self.size[0] / 2
+            maxy = point.iloc[0].y + self.size[0] / 2
+            mint = point.index[0].left
+            maxt = point.index[0].right
+            yield BoundingBox(minx, maxx, miny, maxy, mint, maxt)
```
sfalkena, line 136: Not sure if this was meant to be final, but right now you are not ensuring that the sampled bounding boxes fall within the geometries to sample from. I have two alternative examples to realize this (first one inspired by your way of sampling, and my preferred option):

```
import random
from shapely.geometry import box
import geopandas as gpd

data = {
    'geometry': [
        box(0, 0, 2, 2),
        box(2.5, 2.5, 5, 4),
        box(5.5, 5.5, 9, 7),
        box(9.5, 9.5, 14, 12),
        box(14.5, 14.5, 20, 18)
    ]
}
gdf = gpd.GeoDataFrame(data)

sample_size = 1
# Create valid areas to sample bottom left corners of boxes
debuffered = gpd.GeoDataFrame({'geometry': gdf.geometry.apply(lambda geom: box(geom.bounds[0], geom.bounds[1], geom.bounds[2] - sample_size, geom.bounds[3] - sample_size))})
# Sample points within the debuffered geometries, weighted by area
points = gpd.GeoDataFrame({'geometry': debuffered.dissolve().geometry.sample_points(10).explode()})
sampled_boxes = points.geometry.apply(lambda geom: box(geom.bounds[0], geom.bounds[1], geom.bounds[2] + sample_size, geom.bounds[3] + sample_size))
# Plot the debuffered geometries on top of the original geometries
fig, ax = plt.subplots()
gdf.plot(ax=ax, color='lightblue', edgecolor='black')
debuffered.plot(ax=ax, color='red', alpha=0.5)
sampled_boxes.plot(ax=ax, color='green', alpha=0.5)
plt.show()

```

or:
```import random
from shapely.geometry import box
import geopandas as gpd

def sample_random_boxes(gdf, n=10, sample_size=2):
    """
    Samples n random boxes of a fixed size within the geometries of a GeoDataFrame, weighted by area.

    Parameters:
        gdf (GeoDataFrame): The input GeoDataFrame containing geometries.
        n (int): The number of random boxes to sample.
        box_size (float): The size of the boxes to sample.

    Returns:
        GeoDataFrame: A GeoDataFrame containing the sampled boxes.
    """
    # Calculate the area of each geometry
    gdf['area'] = gdf.geometry.area

    # Normalize the areas to create weights
    total_area = gdf['area'].sum()
    weight = gdf['area'] / total_area

    sampled_boxes = []
    for _ in range(n):
        # Sample a geometry based on weights
        sampled_geometry = random.choices(
            gdf.geometry,
            weights=weight,
            k=1
        )[0]

        # Get the bounds of the sampled geometry
        minx, miny, maxx, maxy = sampled_geometry.bounds

        # Ensure the box fits within the bounds of the geometry
        if maxx - minx >= sample_size and maxy - miny >= sample_size:
            random_x = random.uniform(minx, maxx - sample_size)
            random_y = random.uniform(miny, maxy - sample_size)
            sampled_box = box(random_x, random_y, random_x + sample_size, random_y + sample_size)

            # Check if the box is within the geometry
            if sampled_geometry.contains(sampled_box):
                sampled_boxes.append(sampled_box)

    # Create a new GeoDataFrame with the sampled boxes
    sampled_gdf = gpd.GeoDataFrame({'geometry': sampled_boxes})
    if sampled_gdf.empty:
        raise ValueError("No valid boxes were sampled. Try adjusting the parameters.")

    return sampled_gdf

data = {
    'geometry': [
        box(0, 0, 2, 2),
        box(2.5, 2.5, 5, 4),
        box(5.5, 5.5, 9, 7),
        box(9.5, 9.5, 14, 12),
        box(14.5, 14.5, 20, 18)
    ]
}
gdf = gpd.GeoDataFrame(data)

# Example usage
sampled_boxes = sample_random_boxes(gdf, n=10, sample_size=1)
# Plot the sampled boxes on top of the original geometries
ax = gdf.plot(color='lightblue', edgecolor='black')
sampled_boxes.plot(ax=ax, color='red', alpha=0.5)
plt.show()```

**Comment by sfalkena** [comment]
@adamjstewart, @adrianboguszewski, let me take up landcover.ai dataset to get me into shape in adapting dataset classes

File: torchgeo/datasets/geo.py [hunk 4]
```diff
@@ -518,7 +465,10 @@ def __init__(
 
             self._res = res
 
-        self._crs = crs
+        # Create the dataset index
+        data = {'filepath': filepaths}
+        index = pd.IntervalIndex.from_tuples(datetimes, closed='both', name='datetime')
```
sfalkena, line 470: I was thinking, right now we are creating a datetime index even though there are datasets which don't have any time information. This in combination with the min and max date of 1677 and 2262 becomes a bit confusing.  Maybe we can just check the length of `datetimes` and if it is empty we create a normal index instead of a IntervalIndex. Then for downstream steps, it is directly clear if the dataset has datetime information.

torchgeo/samplers/single.py
adamjstewart: This was meant to be final, but I may change it anyway.

Basically, the `sample_points` approach is nice because it takes into account Polygon bounds, which may not be rectangular, but annoying because it results in sampling outside the bounds by up to half a patch size. I still think it's better than pretending everything is rectangular, but may save that approach for our future SpatialSampler PR and keep this deprecated sampler as backwards-compatible as possible.

Note that `sample_points` has support for providing custom sampling strategies, so it may be possible to write something that has a buffer zone around the point. Geopandas seems to have 10x the features of R-tree and is actually actively maintained, so adding new features is definitely possible, and something we will likely end up doing.

torchgeo/datasets/geo.py
adamjstewart: Mixing index classes becomes complicated, because all of our getitem logic needs to support both types. The idea was that anything with no datetime info is static for all time. This is obviously not true in general, but it's a good assumption. But I'll keep this idea in mind.

torchgeo/samplers/single.py
adamjstewart: Okay, I reverted most backwards-incompatible changes on this. Can you test your code on the new version?

adamjstewart:
@calebrob6 @nilsleh @sfalkena @lccol @isaaccorley I would like to have this PR completed in the next couple days. If you're able to finish converting the remaining custom datasets to geopandas by then let me know. If not, I can start working on them, as everything else is now done.

torchgeo/datasets/l7irish.py
```diff
@@ -213,19 +167,10 @@ def __init__(
         self.image = L7IrishImage(paths, crs, res, bands, transforms, cache)
         self.mask = L7IrishMask(paths, crs, res, None, transforms, cache)
 
-        super().__init__(self.image, self.mask)
+        # Mask filename does not include the date, grab it from the image filename
+        self.mask.index.index = self.image.index.index
```
yichiac, line 171: For the dataset initialization:
```
dataset = L7Irish(paths='data/l7irish', download=False)
```
This line caused an error:
```
ValueError: Length mismatch: Expected axis has 102 elements, new values have 103 elements
```

yichiac: Specifically,
`len(self.image.index.index)==103` and `len(self.mask.index.index)==102`

adamjstewart: @yichiac I'm not able to reproduce this issue. I downloaded the entire dataset fresh and both `L7IrishImage` and `L7IrishMask` have length 206. However, `L7Irish` has length 210, which is the additional bug I mentioned during our meeting. It's possible that your local directory has an extra or duplicated file lying around that is messing things up.

adamjstewart:
@adrianboguszewski LandCoverAIGeo is now ported if you want to test it.

torchgeo/datasets/l7irish.py
adamjstewart: Okay, I fixed it so that the intersection also has length 206. This should be 100% backwards compatible now.

torchgeo/datasets/geo.py
```diff
@@ -990,35 +953,38 @@ def __init__(
...
+        dataset2.res = dataset1.res
+
+        # Spatial intersection
+        index1 = dataset1.index.reset_index()
+        index2 = dataset2.index.reset_index()
+        self.index = gpd.overlay(
+            index1, index2, how='intersection', keep_geom_type=True
+        )
+
+        # Temporal intersection
```
adamjstewart, line 974: TODO: does this actually work? It doesn't seem to be taking an intersection, especially if both datasets aren't sorted by time.

adamjstewart: Update: it works, but I don't know why. My best guess is that `gpd.overlay` returns all possible intersections, then the valid check filters out all invalid cases.

adrianboguszewski:
@adamjstewart, please don't wait for me. If I find any problem, I will report it later in issues.

adriantre (commented):
Looks good imo!

torchgeo/datasets/geo.py
```diff
@@ -445,18 +392,19 @@ def __init__(
...
         if self.all_bands:
             assert set(self.bands) <= set(self.all_bands)
 
-        # Populate the dataset index
-        i = 0
+        # Gather information about the dataset
         filename_regex = re.compile(self.filename_regex, re.VERBOSE)
+        filepaths = []
+        datetimes = []
+        geometries = []
```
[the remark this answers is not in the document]
adamjstewart: That is the exact plan 🙂: https://github.com/microsoft/torchgeo/issues/2382

I want to revive your PR to store the actual footprint, @isaaccorley is looking into storing the actual geometries in VectorDataset and replacing our fiona dependency with geopandas, and eventually we will use something like [sample_points](https://geopandas.org/en/v1.0.1/docs/reference/api/geopandas.GeoSeries.sample_points.html) in our samplers to avoid sampling from nodata locations outside the bounds of the true footprint.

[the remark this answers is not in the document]
adriantre: Sounds great!

isaaccorley (approved):
I've never been so happy to see so many lines of code deleted

PR #2750, opened 2025-04-22
Ignore VSCode files

isaaccorley:
Ignore Visual Studio Code project settings files that are stored in `.vscode/`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@
 /results/
 *.aux.xml
 
+# Visual Studio Code stuff
+.vscode/
+
 # Node stuff:
 node_modules/
 


adamjstewart:
Historically I have rejected OS-specific (`.DS_Store`, `Thumb.db`) or editor-specific (`*.swp`, `*.sublime-workspace`) files from our `.gitignore` because they are not specific to this project, and generally belong in a global `~/.gitignore` file instead. https://github.com/github/gitignore also separates them in a `Global` directory. My concern is that there are several OSes and hundreds of editors, and I don't really want to put all of them in our `.gitignore`. How can we choose which to include and which to exclude? I've chosen to exclude all so far, but I'm open to opinions. VS Code and macOS are clearly popular enough to warrant exceptions.

calebrob6:
I'm 100% fine with having a bunch of editor and OS specific lines in our gitignore (clearly not `results/`, `isaac_tmp_4/`) as I don't see any sort of maintenance or bug risk here. You can put a comment at the top of the file:

```
#############################
### Hello from Adam, did you know about global gitignore (https://github.com/github/gitignore )? Please consider that before adding a line here, but also we won't bite if you open a PR, thanks!
#############################
```

calebrob6 (approved):

PR #2752, opened 2025-04-23
ObjectDetection/InstanceSegmentationTask: fix support for non-RGB images

### LINKED ISSUE #2749: Detection trainer fails with In_channels>3
### Description

Eg if using a 4 channel dataset, the error will be raised:

```python
  File "/usr/local/lib/python3.11/site-packages/torchvision/models/detection/transform.py", line 141, in forward
    image = self.normalize(image)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/torchvision/models/detection/transform.py", line 169, in normalize
    return (image - mean[:, None, None]) / std[:, None, None]
            ~~~~~~^~~~~~~~~~~~~~~~~~~~~
RuntimeError: The size of tensor a (4) must match the size of tensor b (3) at non-singleton dimension 0
```
This is because [GeneralizedRCNNTransform](https://github.com/pytorch/vision/blob/17dc80e12355b815495774126a76b10d8c47a534/torchvision/models/detection/transform.py#L86) only supports 3 channels. We want to use Kornia for norm instead or disable this behaviour from torchvision

### Steps to reproduce

I'm using a proprietary 4 channel dataset

```
model:
  class_path: ObjectDetectionTask
  init_args:
    model: faster-rcnn
    backbone: resnet18
    weights: True
    lr: 5e-4
    in_channels: 4
```

### Version

torchgeo==0.7.0

isaaccorley:
One of the wonderful surprises of torchvision's detector models is that a `GeneralizedRCNNTransform` gets added under the hood which defaults to ImageNet RGB mean/std normalize + dynamic resizing in the range of (800, 1333).

This PR fixes this by loading pretrained weights but overriding this transform to simply subtract 0 and divide by 1 which is a no-op and changes the dynamic resize to allow for a min/max input shape in the range of (1, 4096).

#### Alternatives considered:
I attempted to simply replace `model.transform` with `nn.Identity()` but this doesn't work because the detection models pass multiple args to the transform which will throw an error.

Fixes #2749

### DIFF
--- a/tests/conf/vhr10_ins_seg.yaml
+++ b/tests/conf/vhr10_ins_seg.yaml
@@ -9,6 +9,6 @@ data:
   init_args:
     batch_size: 1
     num_workers: 0
-    patch_size: 4
+    patch_size: 256
   dict_kwargs:
     root: 'tests/data/vhr10'
--- a/torchgeo/trainers/detection.py
+++ b/torchgeo/trainers/detection.py
@@ -151,8 +151,8 @@ class ObjectDetectionTask(BaseTask):
                 num_classes,
                 rpn_anchor_generator=anchor_generator,
                 box_roi_pool=roi_pooler,
-                min_size=1,
-                max_size=4096,
+                min_size=800,
+                max_size=800,
                 image_mean=[0],
                 image_std=[1],
             )
@@ -176,8 +176,8 @@ class ObjectDetectionTask(BaseTask):
                 model_backbone,
                 num_classes,
                 anchor_generator=anchor_generator,
-                min_size=1,
-                max_size=4096,
+                min_size=800,
+                max_size=800,
                 image_mean=[0],
                 image_std=[1],
             )
@@ -214,8 +214,8 @@ class ObjectDetectionTask(BaseTask):
                 num_classes,
                 anchor_generator=anchor_generator,
                 head=head,
-                min_size=1,
-                max_size=4096,
+                min_size=800,
+                max_size=800,
                 image_mean=[0],
                 image_std=[1],
             )
--- a/torchgeo/trainers/detection.py
+++ b/torchgeo/trainers/detection.py
@@ -75,6 +75,9 @@ class ObjectDetectionTask(BaseTask):
     ) -> None:
         """Initialize a new ObjectDetectionTask instance.
 
+        Note that we disable the internal normalize+resize transform of the detection models.
+        Please ensure your images are appropriately resized before passing them to the model.
+
         Args:
             model: Name of the `torchvision
                 <https://pytorch.org/vision/stable/models.html#object-detection>`__
--- a/torchgeo/trainers/detection.py
+++ b/torchgeo/trainers/detection.py
@@ -24,6 +24,7 @@ from torchvision.ops import MultiScaleRoIAlign, feature_pyramid_network, misc
 
 from ..datasets import RGBBandsMissingError, unbind_samples
 from .base import BaseTask
+from .utils import GeneralizedRCNNTransformNoOp
 
 BACKBONE_LAT_DIM_MAP = {
     'resnet18': 512,
@@ -151,11 +152,8 @@ class ObjectDetectionTask(BaseTask):
                 num_classes,
                 rpn_anchor_generator=anchor_generator,
                 box_roi_pool=roi_pooler,
-                min_size=800,
-                max_size=800,
-                image_mean=[0],
-                image_std=[1],
             )
+            self.model.transform = GeneralizedRCNNTransformNoOp()
         elif model == 'fcos':
             kwargs['extra_blocks'] = feature_pyramid_network.LastLevelP6P7(256, 256)
             kwargs['norm_layer'] = (
@@ -173,14 +171,9 @@ class ObjectDetectionTask(BaseTask):
                     param.requires_grad = False
 
             self.model = torchvision.models.detection.FCOS(
-                model_backbone,
-                num_classes,
-                anchor_generator=anchor_generator,
-                min_size=800,
-                max_size=800,
-                image_mean=[0],
-                image_std=[1],
+                model_backbone, num_classes, anchor_generator=anchor_generator
             )
+            self.model.transform = GeneralizedRCNNTransformNoOp()
         elif model == 'retinanet':
             kwargs['extra_blocks'] = feature_pyramid_network.LastLevelP6P7(
                 latent_dim, 256
@@ -214,11 +207,8 @@ class ObjectDetectionTask(BaseTask):
                 num_classes,
                 anchor_generator=anchor_generator,
                 head=head,
-                min_size=800,
-                max_size=800,
-                image_mean=[0],
-                image_std=[1],
             )
+            self.model.transform = GeneralizedRCNNTransformNoOp()
         else:
             raise ValueError(f"Model type '{model}' is not valid.")
 
--- a/torchgeo/trainers/detection.py
+++ b/torchgeo/trainers/detection.py
@@ -151,6 +151,10 @@ class ObjectDetectionTask(BaseTask):
                 num_classes,
                 rpn_anchor_generator=anchor_generator,
                 box_roi_pool=roi_pooler,
+                min_size=1,
+                max_size=4096,
+                image_mean=[0],
+                image_std=[1],
             )
         elif model == 'fcos':
             kwargs['extra_blocks'] = feature_pyramid_network.LastLevelP6P7(256, 256)
@@ -169,7 +173,13 @@ class ObjectDetectionTask(BaseTask):
                     param.requires_grad = False
 
             self.model = torchvision.models.detection.FCOS(
-                model_backbone, num_classes, anchor_generator=anchor_generator
+                model_backbone,
+                num_classes,
+                anchor_generator=anchor_generator,
+                min_size=1,
+                max_size=4096,
+                image_mean=[0],
+                image_std=[1],
             )
         elif model == 'retinanet':
             kwargs['extra_blocks'] = feature_pyramid_network.LastLevelP6P7(
@@ -204,6 +214,10 @@ class ObjectDetectionTask(BaseTask):
                 num_classes,
                 anchor_generator=anchor_generator,
                 head=head,
+                min_size=1,
+                max_size=4096,
+                image_mean=[0],
+                image_std=[1],
             )
         else:
             raise ValueError(f"Model type '{model}' is not valid.")
--- a/torchgeo/trainers/instance_segmentation.py
+++ b/torchgeo/trainers/instance_segmentation.py
@@ -22,6 +22,7 @@ from torchvision.models.detection import (
 
 from ..datasets import RGBBandsMissingError, unbind_samples
 from .base import BaseTask
+from .utils import GeneralizedRCNNTransformNoOp
 
 
 class InstanceSegmentationTask(BaseTask):
@@ -46,6 +47,9 @@ class InstanceSegmentationTask(BaseTask):
         freeze_backbone: bool = False,
     ) -> None:
         """Initialize a new InstanceSegmentationTask instance.
+
+        Note that we disable the internal normalize+resize transform of the MaskRCNN model.
+        Please ensure your images are appropriately resized before passing them to the model.
 
         Args:
             model: Name of the model to use.
@@ -87,11 +91,10 @@ class InstanceSegmentationTask(BaseTask):
                     weights=weights,
                     num_classes=num_classes,
                     weights_backbone=weights_backbone,
-                    min_size=800,
-                    max_size=800,
                     image_mean=[0],
                     image_std=[1],
                 )
+                self.model.transform = GeneralizedRCNNTransformNoOp()
             else:
                 msg = f"Invalid backbone type '{backbone}'. Supported backbone: 'resnet50'"
                 raise ValueError(msg)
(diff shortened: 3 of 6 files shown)


adamjstewart:
Can you also check instance segmentation?

isaaccorley:
> OK, whilst the error is resolved, the loss for OD models I train is always zero. Is there a validated results I can reproduce? Note this might just be my datasets which I recently updated for the new format

This by default was resizing all imagery to a min of 800. What transforms are you using to preprocess your imagery?

isaaccorley:
@adamjstewart fixed the instance segmentation. It had the same issue.

robmarkcole:
@isaaccorley can you elaborate on `This by default was resizing all imagery to a min of 800`?
Example transforms:

```
        self.train_aug = Ka.AugmentationSequential(    
                Ka.Normalize(mean=self.mean, std=self.std),
                Ka.Resize(self.chip_size),
                Ka.RandomHorizontalFlip(p=0.5),
                Ka.RandomVerticalFlip(p=0.5),
                Ka.RandomRotation(degrees=(90.0, 90.0), p=0.25),
                data_keys=None,
                keepdim=True,
        )

        test_transforms: List[Ka.AugmentationBase2D] = [
            Ka.Normalize(mean=self.mean, std=self.std),
            Ka.Resize(self.chip_size),
        ]
```

where typically chip_size = 224

isaaccorley:
> @isaaccorley can you elaborate on `This by default was resizing all imagery to a min of 800`?
> 
> Example transforms:
> 
> 
> 
> ```
> 
>         self.train_aug = Ka.AugmentationSequential(    
> 
>                 Ka.Normalize(mean=self.mean, std=self.std),
> 
>                 Ka.Resize(self.chip_size),
> 
>                 Ka.RandomHorizontalFlip(p=0.5),
> 
>                 Ka.RandomVerticalFlip(p=0.5),
> 
>                 Ka.RandomRotation(degrees=(90.0, 90.0), p=0.25),
> 
>                 data_keys=None,
> 
>                 keepdim=True,
> 
>         )
> 
> 
> 
>         test_transforms: List[Ka.AugmentationBase2D] = [
> 
>             Ka.Normalize(mean=self.mean, std=self.std),
> 
>             Ka.Resize(self.chip_size),
> 
>         ]
> 
> ```
> 
> 
> 
> where typically chip_size = 224

Torchvision Faster-RCNN and MaskRCNN has a `GeneralizedRCNNTransform` transform module inside the model itself that will perform normalize + resizing to a minimum image size of 800. So any image you pass in <800 will be resized to 800. See the code [here](https://github.com/pytorch/vision/blob/5f03dc524bdb7529bb4f2e84d2d8c237233fc62a/torchvision/models/detection/faster_rcnn.py#L172).

One trick that works well for object detection in remote sensing is to simply resize your small patches to be larger. This may be why you're getting poor performance.

robmarkcole:
@isaaccorley good to know! Perhaps we should document this?

robmarkcole:
I've ruled out issues with my dataset and the remaining differences I see beween my legacy implementation and this implementation are details such as the anchor sizes I've utilised. I suggest we merge this approach and then as a follow up (and pending a suitable test dataset) work on further optimisations in another PR

adamjstewart (dismissed):
LGTM, thanks for the fix!

adamjstewart:
Except for the tests...

robmarkcole:
As discussed in slack with @isaaccorley better results are achieved with the torchvision defaults for norm with 
```
            min_size=800,
            max_size=1333,
```

Noting https://github.com/allenai/rslearn/issues/171 where they choose identical values, which makes sense as we usually deal with square images in RS

robmarkcole:
Some benchmarking based on the defaults using VHR10 for 25 epochs:

```python
task = ObjectDetectionTask(
    model="faster-rcnn",
    backbone="resnet18",
    weights=True,
    in_channels=3,
    num_classes=11,
    trainable_layers=3,
    lr=1e-3,
    patience=10,
    freeze_backbone=False,
)

Equal sizes:
```
min_size=800,
max_size=800

test_map_50        │    0.64
```

File: torchgeo/trainers/detection.py [hunk 1]
```diff
@@ -151,6 +151,10 @@ def configure_models(self) -> None:
                 num_classes,
                 rpn_anchor_generator=anchor_generator,
                 box_roi_pool=roi_pooler,
+                min_size=800,
+                max_size=800,
```
adamjstewart, line 155: Is this a hard constraint? It would be nice to be as flexible as possible here.

robmarkcole: I actually think it might be better to pass this in as an arg, image_size or similar. However as mentioned it would be preferable if we could avoid this processing altogether.

isaaccorley: I'll make it an input arg

robmarkcole: Do all the datamodules support image_size as an input arg? I have it as one for mine and always perform a resize - definitely want control over this

adamjstewart: Not consistently, no, but we could. We definitely could for all the object detection/instance segmentation ones, there aren't that many.

isaaccorley: Let me try updating the datamodules with controllable sizes and see if I can get the models to converge. Ideally we could set this range to be a no-op which is what I originally set it to (1,4096) for

isaaccorley:
The behavior of `GeneralizedRCNNTransform` is very odd. You would think that if I have a min/max size range of (32, 4096) that if my image is 512x512 that it would be within the valid range and not be resized but it actually gets resized to the min size...

```python
from torchvision.models.detection.transform import GeneralizedRCNNTransform

images = [torch.randn(3, 512, 512)]

transform = GeneralizedRCNNTransform(min_size=32, max_size=4096, image_mean=[0], image_std=[1])
print(transform(images)[0].tensors.shape)  # torch.Size([1, 3, 32, 32])

transform = GeneralizedRCNNTransform(min_size=4096, max_size=4096, image_mean=[0], image_std=[1])
print(transform(images)[0].tensors.shape)  # torch.Size([1, 3, 4096, 4096])
```

isaaccorley:
HAH found that they have a debug arg of some sort that lets you skip the resize [here](https://github.com/pytorch/vision/blob/11ed64e827a99fd7895674fd71f30261a31c7c64/torchvision/models/detection/transform.py#L117)

```python
from torchvision.models.detection.transform import GeneralizedRCNNTransform

images = [torch.randn(3, 512, 512)]
transform = GeneralizedRCNNTransform(min_size=1, max_size=1, image_mean=[0], image_std=[1], _skip_resize=True)
print(transform(images)[0].tensors.shape)  # torch.Size([1, 3, 512, 512])

**Comment by adamjstewart** [comment]
Ugh.

> I attempted to simply replace `model.transform` with `nn.Identity()` but this doesn't work because the detection models pass multiple args to the transform which will throw an error.

Can we replace it with:
```python
def identity(input: Tensor, *args: Any, **kwargs: Any) -> Tensor:
    return input
```
or even some kind of lambda?

isaaccorley:
> Ugh.
> 
> > I attempted to simply replace `model.transform` with `nn.Identity()` but this doesn't work because the detection models pass multiple args to the transform which will throw an error.
> 
> Can we replace it with:
> 
> ```python
> def identity(input: Tensor, *args: Any, **kwargs: Any) -> Tensor:
>     return input
> ```
> 
> or even some kind of lambda?

Yes we could, however my above method makes it a no-op without us having to maintain any additional LOC

isaaccorley:
Oof nevermind, it doesn't work if the model is in eval mode...

isaaccorley:
> Ugh.
> 
> > I attempted to simply replace `model.transform` with `nn.Identity()` but this doesn't work because the detection models pass multiple args to the transform which will throw an error.
> 
> Can we replace it with:
> 
> ```python
> def identity(input: Tensor, *args: Any, **kwargs: Any) -> Tensor:
>     return input
> ```
> 
> or even some kind of lambda?

I'm able to do this however it doesn't just return the input, it returns an `ImageList`. See [here](https://github.com/pytorch/vision/blob/11ed64e827a99fd7895674fd71f30261a31c7c64/torchvision/models/detection/transform.py#L119)

isaaccorley:
My solution adds LOC but it works when tested for both training and inference.

```python
class GeneralizedRCNNTransformNoOp(GeneralizedRCNNTransform):
    """GeneralizedRCNNTransform without the normalize and resize ops.

    .. versionadded:: 0.8
    """

    def __init__(self) -> None:
        """Initialize a new GeneralizedRCNNTransformNoOp instance."""
        super().__init__(min_size=0, max_size=0, image_mean=[0], image_std=[1], _skip_resize=True)

    def resize(
        self,
        image: Tensor,
        target: dict[str, Tensor] | None = None,
    ) -> tuple[Tensor, dict[str, Tensor] | None]:
        """Skip resizing and return the image and target."""
        return image, target
```

Then testable like this

```python
from torchvision.models.detection import maskrcnn_resnet50_fpn
import torch

images = [torch.randn(3, 512, 512)]
model = maskrcnn_resnet50_fpn(pretrained=True)
model.transform = GeneralizedRCNNTransformNoOp()

model.eval()
with torch.inference_mode():
    preds = model(images)
preds
"""
[{'boxes': tensor([], size=(0, 4)),
  'labels': tensor([], dtype=torch.int64),
  'scores': tensor([]),
  'masks': tensor([], size=(0, 1, 512, 512))}]
"""

boxes = torch.tensor([
    [50, 50, 150, 150],
    [200, 200, 300, 300],
    [100, 400, 200, 500],
    [400, 100, 500, 200],
    [250, 250, 350, 350]
], dtype=torch.float32)

model.train()
preds = model(images, targets=[{"boxes": boxes, "labels": torch.randint(0, 91, (10,)), "masks": torch.randn(10, 512, 512)}])
preds
"""
{'loss_classifier': tensor(0.2235, grad_fn=<NllLossBackward0>),
 'loss_box_reg': tensor(0.0202, grad_fn=<DivBackward0>),
 'loss_mask': tensor(0.4244, grad_fn=<BinaryCrossEntropyWithLogitsBackward0>),
 'loss_objectness': tensor(1.1351, grad_fn=<BinaryCrossEntropyWithLogitsBackward0>),
 'loss_rpn_box_reg': tensor(0.0442, grad_fn=<DivBackward0>)}
"""
```

tests/conf/vhr10_ins_seg.yaml
```diff
@@ -9,6 +9,6 @@ data:
   init_args:
     batch_size: 1
     num_workers: 0
-    patch_size: 4
+    patch_size: 256
```
isaaccorley, line 12: This had to get fixed because the GeneralizedRCNNTransform was resizing behind the scenes. Now that that is no longer happening because it's a no-op we need to increase the patch size so the test images are resized.

robmarkcole:
@isaaccorley results look good with this approach

torchgeo/trainers/instance_segmentation.py
```diff
@@ -47,6 +48,9 @@ def __init__(
     ) -> None:
         """Initialize a new InstanceSegmentationTask instance.
 
+        Note that we disable the internal normalize+resize transform of the MaskRCNN model.
+        Please ensure your images are appropriately resized before passing them to the model.
```
adamjstewart, line 52: Do we need this comment for ObjectDetectionTask too?

torchgeo/trainers/instance_segmentation.py
```diff
@@ -87,7 +91,10 @@ def configure_models(self) -> None:
                     weights=weights,
                     num_classes=num_classes,
                     weights_backbone=weights_backbone,
+                    image_mean=[0],
+                    image_std=[1],
```
adamjstewart, line 95: Do we need these? We didn't use them in ObjectDetectionTask

torchgeo/trainers/utils.py
```diff
@@ -11,6 +11,24 @@
 import torch.nn as nn
 from torch import Tensor
 from torch.nn.modules import Conv2d, Module
+from torchvision.models.detection.transform import GeneralizedRCNNTransform
+
+
+class GeneralizedRCNNTransformNoOp(GeneralizedRCNNTransform):  # type: ignore[misc]
+    """GeneralizedRCNNTransform without the normalize and resize ops.
+
+    .. versionadded:: 0.7.1
```
adamjstewart, line 20:
```suggestion
    .. versionadded:: 0.8
```
It won't appear in the docs anyway, so let's pretend it is added in 0.8 even though it will be there in 0.7.1.

torchgeo/trainers/utils.py
```diff
@@ -11,6 +11,24 @@
 import torch.nn as nn
 from torch import Tensor
 from torch.nn.modules import Conv2d, Module
+from torchvision.models.detection.transform import GeneralizedRCNNTransform
+
+
+class GeneralizedRCNNTransformNoOp(GeneralizedRCNNTransform):  # type: ignore[misc]
```
adamjstewart, line 17: What is the type ignore here? Should I add type hints to torchvision?

isaaccorley: this is the error I get

`torchgeo/trainers/utils.py:17: error: Class cannot subclass
"GeneralizedRCNNTransform" (has type "Any")  [misc]`

isaaccorley: This doesn't make much sense because the code itself inherits from torch.nn.Module [here](https://github.com/pytorch/vision/blob/966da7e46f65d6d49df3e31214470a4fe5cc8e66/torchvision/models/detection/transform.py#L86)

isaaccorley: Ahh okay yes likely torchvision doesn't have type hints for this

adamjstewart: Strange, I see type hints for basically every class method and attribute, not sure where it's getting `Any` from. Oh well.

adamjstewart (approved):

PR #2754, opened 2025-04-23
Intersection/UnionDataset: simplify crs/res getter/setter

adamjstewart:
I don't see any reason this wouldn't work, but let's see what the tests uncover.

### DIFF
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -1067,7 +1067,7 @@ class IntersectionDataset(GeoDataset):
         Returns:
             The :term:`coordinate reference system (CRS)`.
         """
-        return self._crs
+        return self.datasets[0].crs
 
     @crs.setter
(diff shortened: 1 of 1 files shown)


calebrob6:
What if something depends on `dataset._res` being correct? (I see the tests pass, but I don't see why we should change it just because)

adamjstewart:
`dataset._res` is internal, and no one should rely on it. We shouldn't change it *just because*, we should change it because -10 lines of code we need to maintain.

calebrob6:
Sorry, you are right, it definitely isn't _just because_! Also, how do you even find these?

calebrob6 (dismissed):
Happy whenever mypy is (but honestly, I'm usually happy way before mypy is)

adamjstewart:
Really wish we could fix this mypy bug. Hopefully ruff's red knot won't have this same issue.

calebrob6 (approved):

PR #2756, opened 2025-04-24
Changing Link to Substation Dataset

### LINKED ISSUE #2738: Change of Link to Substation Dataset
Hi, @adamjstewart I want to change the link to the substation dataset to Hugging Face for the next release. But I am getting the following error while doing so. This is because the dataset is big and I have to divide it into parts, can I implement my custom function to handle this and a external library, or is there an alternative:

File /ext3/miniforge3/lib/python3.12/zipfile/__init__.py:257, in _EndRecData64(fpin, offset, endrec)
    254     return endrec
    256 if diskno != 0 or disks > 1:
--> 257     raise BadZipFile("zipfiles that span multiple disks are not supported")
    259 # Assume no 'zip64 extensible data'
    260 fpin.seek(offset - sizeEndCentDir64Locator - sizeEndCentDir64, 2)

BadZipFile: zipfiles that span multiple disks are not supported
File /ext3/miniforge3/lib/python3.12/zipfile/__init__.py:257, in _EndRecData64(fpin, offset, endrec)
    254     return endrec
    256 if diskno != 0 or disks > 1:
--> 257     raise BadZipFile("zipfiles that span multiple disks are not supported")
    259 # Assume no 'zip64 extensible data'
    260 fpin.seek(offset - sizeEndCentDir64Locator - sizeEndCentDir64, 2)

BadZipFile: zipfiles that span multiple disks are not supported

rijuld:
Closes #2738


adamjstewart:
Is this still a WIP? We're planning a 0.9 release this week and would love to include this.

adamjstewart:
@rijuld 0.10 is now coming soon, do you want to revive this PR or just close it?

rijuld:
> @rijuld 0.10 is now coming soon, do you want to revive this PR or just close it?

Yes, I want to revive this PR

rijuld:
Hi @adamjstewart, you suggested using SSL4EO-L as inspiration but SSL4EO-L's parts are just byte slices from split -b, so joining them gives the tarball back. The substation images were made with zip -s though, which is a spanned zip that stores offsets relative to each part, so merging leaves most of the 26,523 entries pointing at the wrong place. That's where my BadZipFile came from.

I've added a small _extract_images that reads the three parts as one stream and inflates each entry without touching the central directory. I can't repack the HF repo because I don't own it.

torchgeo/datasets/substation.py
```diff
@@ -46,18 +49,31 @@ class Substation(NonGeoDataset):
...
+
+    Access:
+
+    * Dataset is hosted on Hugging Face: https://huggingface.co/datasets/neurograce/SubstationDataset
+
+    .. versionchanged:: 0.10
+       The dataset is now downloaded from Hugging Face instead of Google Cloud Storage.
+       The images are distributed as a multi-part zip archive, and the ``md5_images``,
+       ``md5_masks``, ``url_for_images``, and ``url_for_masks`` attributes have been
+       replaced by *url* and *checksums*.
```
adamjstewart: I think we can remove this part, it isn't user-facing and users won't notice the difference.

adamjstewart (commented):
Extraction logic is more complicated than I would like, but I also don't know of an easier way to do this. HF is nice for reproducibility, is that the only reason we wanted to change the link? Will let @nilsleh review and see if there is a simpler way to do this.

PR #2759, opened 2025-04-26
Plot Method Added to MMEarth Dataset

amrirasyidi:
- Relevant issue: https://github.com/microsoft/torchgeo/issues/2377
- Added plot method to MMEarth dataset.

Using the dataset from test, this is how the plot looks like
image

### DIFF
--- a/torchgeo/datasets/mmearth.py
+++ b/torchgeo/datasets/mmearth.py
@@ -628,7 +628,7 @@ class MMEarth(NonGeoDataset):
         show_titles: bool = True,
         suptitle: str | None = None,
     ) -> Figure:
-        """Plot a sample from the dataset.
+        """Plot a sample from the dataset as shown in fig. 2 from https://arxiv.org/pdf/2405.02771.
 
         Args:
             sample: A sample returned by :meth:`__getitem__`.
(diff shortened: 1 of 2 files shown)


adamjstewart:
Just need to see what the plot would look like with the real data, hoping @nilsleh has some data lying around. If not, we can do a partial download (start curl then kill it), then extract a few samples and plot them.

amrirasyidi:
I just downloaded the 100k version, and this is how the plot looks like
```
from torchgeo.datasets import MMEarth
data = MMEarth('path/to/data', subset='MMEarth100k')
f = data.plot(data[8000])
```
image

Most of the HH and HV bands are `nan`, that is why they are blank.

amrirasyidi:
# Color Map
I'll use these cmaps
- ESA WorldCover: table 2 from this document https://worldcover2020.esa.int/data/docs/WorldCover_PUM_V1.1.pdf
- Dynamic World: https://developers.google.com/earth-engine/tutorials/community/introduction-to-dynamic-world-pt-1#visualize_the_classified_image

# To be Drawn
image

Based on Fig. 2 in https://arxiv.org/pdf/2405.02771 these are the data that will be drawn in the plot
- Sentinel-2 RGB
- Sentinel-1 <- **what about this one?** is it composite, or just a single band? (ascending, descending? vv, vh?)
- ASTER Elevation
- ESA WorldCover
- Dynamic World
- Canopy Height

Any comment regarding this?

adamjstewart:
Looks good to me! For Sentinel-1, I'm not a SAR expert, but I would just pick a random one if they don't say which they plotted in the paper.

nilsleh:
For SAR you can try plotting the false color image like it's done [here](https://github.com/microsoft/torchgeo/blob/060f2ba68685f9fa67361ef95fcaf857bb677d59/torchgeo/datasets/mmflood.py#L335) for example.

adamjstewart:
I've had mixed results with this, but always fun to try.

amrirasyidi:
This is the plot so far, the cmap for esa and dynamic world have been adjusted.

# Sentinel 1 as composite
I'm just replicating the compositing method from [here](https://github.com/microsoft/torchgeo/blob/060f2ba68685f9fa67361ef95fcaf857bb677d59/torchgeo/datasets/mmflood.py#L335)
image

# Sentinel 1 vv as is
image

adamjstewart:
Hmm, looks like compositing didn't work so well. The formula is completely different for power and decibel scales, could be related to the issue. I would personally vote for the second one since it's closer to what the authors used for plotting.

P.S. I just noticed https://github.com/vishalned/MMEarth-data/blob/main/data_exp/view_h5.py, may give you some hints on how the authors plotted things.

amrirasyidi:
Okay

I'll merge using the 2nd option for now, as it is indeed closer to what the authors used for plotting.

amrirasyidi:
Hi @adamjstewart , need your approval for this merge request if everything looks good, thanks

torchgeo/datasets/mmearth.py
```diff
@@ -618,3 +621,126 @@ def __len__(self) -> int:
...
+                    norm_img = percentile_normalization(band_val)
+                    images.append(norm_img)
+
+                    modalities_name = key.split('_', 1)[1]
+                    band_name = avail_bands_dict[modalities_name][0]
+                    titles.append(
+                        (modalities_name.replace('_', ' ').title()) + ' ' + band_name
+                    )
+        fig, ax = plt.subplots(1, 6, figsize=(12, 4), squeeze=False)
+        axes = ax.flatten()
```
adamjstewart, line 724: Why use both `squeeze=False` and `ax.flatten`?

adamjstewart (commented):
One minor change and then I think this is ready to merge

adamjstewart (approved):

PR #2760, opened 2025-04-28
Add PatternNet datamodule

burakekim:
Adds PatternNet datamodule


tests/datamodules/test_patternnet.py
adamjstewart, line 1: We're actually trying to get rid of all of these files. Instead, the way to test a datamodule is with the compatible trainer. You can add a `tests/conf/patternnet.yaml` file with the configuration needed to test and add one line to `tests/trainers/test_classification.py` to make sure the datamodule is compatible with the respective trainer.

torchgeo/datamodules/patternnet.py
```diff
@@ -0,0 +1,77 @@
...
+
+        Args:
+            stage: Either 'fit', 'validate', 'test', or 'predict'.
+        """
+        if stage in ('fit', 'validate', 'test'):
+            full = PatternNet(**self.kwargs)
+            total_len = len(full)
+            test_len = round(total_len * self.test_split_pct)
+            val_len = round(total_len * self.val_split_pct)
+            train_len = total_len - val_len - test_len
```
adamjstewart, line 65: No reason to do the math here, `random_split` can handle both ints and percentages.

torchgeo/datamodules/patternnet.py
```diff
@@ -0,0 +1,77 @@
...
+            train, val, test = random_split(
+                full, [train_len, val_len, test_len], generator=generator
+            )
+
+            if stage == 'fit':
+                self.train_dataset = train
+            if stage in ('fit', 'validate'):
+                self.val_dataset = val
+            if stage == 'test':
+                self.test_dataset = test
```
adamjstewart, line 77: Since we have to create the `full` class regardless, let's just always set all three attributes, no if-statements needed.

torchgeo/datamodules/patternnet.py
```diff
@@ -0,0 +1,77 @@
...
+
+    mean = torch.tensor([91.48, 91.78, 81.23])
+    std = torch.tensor([49.74, 47.18, 45.43])
+
+    def __init__(
+        self,
+        batch_size: int = 64,
+        num_workers: int = 0,
+        val_split_pct: float = 0.2,
+        test_split_pct: float = 0.1,
```
adamjstewart, line 30: I usually use the same percent for both val or test. 0.2 or 0.1 both sound fine to me. I guess whichever is more common in other datasets. May also depend on how big the dataset is.

torchgeo/datamodules/patternnet.py
```diff
@@ -0,0 +1,77 @@
...
+        **kwargs: Any,
+    ) -> None:
+        """Initialize a new PatternNetDataModule instance.
+
+        Args:
+            batch_size: Size of each mini-batch.
+            num_workers: Number of workers for parallel data loading.
+            val_split_pct: Fraction of dataset to use for validation.
+            test_split_pct: Fraction of dataset to use for testing.
+            **kwargs: Additional keyword arguments passed to PatternNet.
```
adamjstewart, line 40: Can we link to the `PatternNet` class here?

torchgeo/datamodules/patternnet.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        super().__init__(PatternNet, batch_size, num_workers, **kwargs)
+
+        self.val_split_pct = val_split_pct
+        self.test_split_pct = test_split_pct
+
+        self.aug = K.AugmentationSequential(
```
adamjstewart, line 47: Should we add things like horizontal/vertical flip for `train_aug`? I've been meaning to add those by default to all datamodules, they almost always help.

adamjstewart (changes requested):

torchgeo/datamodules/patternnet.py
burakekim: This is a rather small dataset -- and also performance-wise saturated. Since there seems to be no class imbalance, I will then use 0.1 and 0.1 for val and test

tests/datamodules/test_patternnet.py
burakekim: Loved it! I can lead this migration actually

adamjstewart: Great, but I want you to focus on datamodules for our change detection datasets first. Some of the datamodule tests we're trying to remove are actually waiting on a change detection trainer.

torchgeo/datamodules/patternnet.py
```diff
@@ -0,0 +1,74 @@
...
+            K.Normalize(mean=self.mean, std=self.std),
+            K.RandomHorizontalFlip(p=0.5),
+            K.RandomVerticalFlip(p=0.5),
+            K.Resize(size=256),
+            data_keys=None,
+            keepdim=True,
+        )
```
adamjstewart, line 56:
```suggestion
        self.aug = K.AugmentationSequential(
            K.Normalize(mean=self.mean, std=self.std),
            K.Resize(size=256),
            data_keys=None,
            keepdim=True,
        )
        self.train_aug = K.AugmentationSequential(
            K.Normalize(mean=self.mean, std=self.std),
            K.RandomHorizontalFlip(p=0.5),
            K.RandomVerticalFlip(p=0.5),
            K.Resize(size=256),
            data_keys=None,
            keepdim=True,
        )
```
We don't want data augmentation for val/test, only preprocessing.

burakekim: Ah, my bad!

adamjstewart (approved):

PR #2769, opened 2025-05-02
Plot Method Added to Western USA Live Fuel Moisture Dataset

amrirasyidi:
- Relevant issue: https://github.com/microsoft/torchgeo/issues/2377
- Added plot method to Western USA Live Fuel Moisture dataset.

This is how the plot looks like
# Default plot
```python
from torchgeo.datasets import WesternUSALiveFuelMoisture
data = WesternUSALiveFuelMoisture('path/to/data')
sample = data[0]
f = data.plot(
    sample,
    show_titles=True, suptitle='test suptitle'
)
```
image

# Single variable defined
```python
f = data.plot(
    sample,
    variables_to_plot=['slope'],
    show_titles=True, suptitle='test suptitle'
)
```
image

# Multiple variables defined
```python
f = data.plot(
    sample,
    variables_to_plot=[
        'slope',
        'green',
        'swir'
    ],
    show_titles=True, suptitle='test suptitle'
)
```
image

### DIFF
--- a/torchgeo/datasets/western_usa_live_fuel_moisture.py
+++ b/torchgeo/datasets/western_usa_live_fuel_moisture.py
@@ -359,16 +359,16 @@ class WesternUSALiveFuelMoisture(NonGeoDataset):
         lat = input_data[-1]
         lfmc_value = sample['label'].item()
 
-        axs[-1].text(
+        fig.text(
             x=0.5,
-            y=-0.6,
+            y=-0.05,
             s=f'Live Fuel Moisture Content\nat {lon:.4f}, {lat:.4f}: {lfmc_value:.2f}%',
             ha='center',
-            transform=axs[-1].transAxes,
+            transform=fig.transFigure,
         )
 
         if suptitle is not None:
-            fig.suptitle(t=suptitle, y=1.5, fontsize=12, transform=axs[0].transAxes)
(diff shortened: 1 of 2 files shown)


adamjstewart:
Hmm, maybe there are too many variables to always plot all of them. Unless we can somehow group plots together. What's the difference between red, vv_red, and vh_red? Maybe we can only plot RGB instead of all possible color combinations? This one might require some creativity to actually be a useful plot.

torchgeo/datasets/western_usa_live_fuel_moisture.py
```diff
@@ -297,3 +299,116 @@ def _download(self) -> None:
...
+            y=-0.7,
+            s=f'Live Fuel Moisture Content\nat {lon:.4f}, {lat:.4f}: {lfmc_value:.2f}%',
+            ha='center',
+            transform=axs[-1].transAxes,
+        )
+
+        if suptitle is not None:
+            fig.suptitle(t=suptitle, y=1.6, transform=axs[0].transAxes)
+
+        # fig.tight_layout()
```
adamjstewart:
```suggestion
```

adamjstewart (approved):
I don't think this plot is perfect, but I do think it's better than no plot. Let's merge and refine in future PRs if anyone cares.

Issue #2505, opened 2025-01-08
VectorDataset: return bounding boxes and instance segmentation masks


adamjstewart:
### Summary

Currently, `VectorDataset.__getitem__` returns only a raster `mask` designed for semantic segmentation. We should extend `VectorDataset` to support:

### raster output
  - [x] semantic segmentation mask
  - [ ] instance segmentation mask
### vector output
  - [ ] object detection bounding boxes
  - [ ] keypoint detection
  - [ ] object polygons?

### Rationale

Currently `VectorDataset` can only be used for semantic segmentation, but there are many other applications that store data in vector shapefiles.

### Implementation

We should decide whether we should always return all of these options, or whether the dataset will have a new parameter that decides which output will be returned.

### Alternatives

Could also make a new dataset class.

BeritCheema:
I would like to work on this issue

adamjstewart:
Please do!

amrirasyidi:
Where should I look first to start working on this issue?

adamjstewart:
`VectorDataset.__getitem__`

mayrajeo:
I'm working with this also, and thinking of something like the following:

Use this helper function to convert geocoordinates to pixel coordinates:

```python
def convert_poly_coords(
    geom:shapely.geometry.shape, affine_obj:affine.Affine, inverse:bool=False
) -> shapely.geometry.shape:
    """Convert pixel coordinates to geocoordinates and vice versa, 
    based on `affine_obj`

    Args:
        geom: shapely.geometry.shape to convert
        affine_obj: affine.Affine object to use for geoconversion
        inverse: If true, convert geocoordinates to pixel coordinates

    Returns:
        shapely.geometry.shape: input shape converted to pixel coordinates
    """
    if inverse:
        affine_obj = ~affine_obj

    xformed_shape = shapely.affinity.affine_transform(geom, 
        [
            affine_obj.a,
            affine_obj.b,
            affine_obj.d,
            affine_obj.e,
            affine_obj.xoff,
            affine_obj.yoff,
        ],
    )
    return xformed_shape
```

And then modify `VectorDataset.__getitem__` with:

```python
        if shapes:
            masks = rasterio.features.rasterize(
                shapes, out_shape=(round(height), round(width)), transform=transform
            )
            # convert shapes to pixel coordinates
            px_shapes = [convert_poly_coords(shapely.geometry.shape(s[0]), transform, inverse=True) 
                         for s in shapes]

[...]

PR #2785, opened 2025-05-18
Feature/rasterized vectordataset support #2505

sherif-med:
## Summary

This PR introduces the `RasterizedVectorDataset`, a subclass of `VectorDataset`, to enable loading vector data as rasterized masks, addressing issue #2505 . It includes:

- ✅ Implementation of `RasterizedVectorDataset`
- ✅ Introduction of `RasterizationStrategy` and `DefaultRasterizationStrategy`
- ✅ Support for clipping geometries to the query window
- ✅ Updated documentation and tutorials
- ✅ Unit tests for rasterized datasets and clipping behavior

---

## Motivation

Many geospatial tasks (e.g., semantic segmentation) require vector labels to be rasterized. This new dataset class simplifies workflows by seamlessly integrating rasterized vector data into TorchGeo’s dataset ecosystem.

---

## Docs Updated

- `docs/tutorials/custom_raster_dataset.ipynb`: Added explanations for `RasterizedVectorDataset`
- `docs/user/alternatives.rst`: Added `RasterizedVectorDataset` to comparison overview
- API docstrings for all new classes

---

## Tests

- Added tests in `tests/datasets/test_geo.py` for:
  - `__getitem__`
  - Querying with and without geometry clipping
  - Edge cases (e.g., empty window, invalid query)
  - Multilabel support
  - Proper rasterization output

---

## Affected Files (Highlights)

- `torchgeo/datasets/geo.py`: Implementation of `RasterizedVectorDataset`, `RasterizationStrategy`
- `torchgeo/datasets/cbf.py`, `eurocrops.py`: Switched to use `RasterizedVectorDataset`
- `torchgeo/datasets/__init__.py`: Added new classes to exports
- Unit tests and docs as noted above

---

## Example Usage

```python
from torchgeo.datasets import RasterizedVectorDataset

dataset = RasterizedVectorDataset(
    root="data/vector",
    res=0.1,
    transforms=my_transforms,
    label_name="class_id",
    clip_geometries=True,
)
sample = dataset[dataset.bounds]
```

---

## Future Enhancements

- Add a new rasterizer that converts polygons into mask-of-polygons with contours
- Integrate visualization tools for debugging geometry mask alignment

### DIFF
--- a/docs/api/datasets.rst
+++ b/docs/api/datasets.rst
@@ -638,9 +638,15 @@ VectorDataset
 .. autoclass:: VectorDataset
 
 RasterizedVectorDataset
-^^^^^^^^^^^^^
+^^^^^^^^^^^^^^^^^^^^^^^
 
 .. autoclass:: RasterizedVectorDataset
+
+Vector datasets Rasterizers
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. autoclass:: RasterizationStrategy
+.. autoclass:: DefaultRasterizationStrategy
 
 NonGeoDataset
 ^^^^^^^^^^^^^
--- a/docs/api/datasets.rst
+++ b/docs/api/datasets.rst
@@ -637,6 +637,11 @@ VectorDataset
 
 .. autoclass:: VectorDataset
 
+RasterizedVectorDataset
+^^^^^^^^^^^^^
+
+.. autoclass:: RasterizedVectorDataset
+
 NonGeoDataset
 ^^^^^^^^^^^^^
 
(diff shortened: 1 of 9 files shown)


adamjstewart:
Why do we need `RasterizedVectorDataset` to solve #2505? `VectorDataset` already supports this feature.

sherif-med:
As mentioned in #2505 some tasks require geometries instead of masks as input. Also the name VectorDataset is counter-intuitive as its elements are images/masks.
VectorDataset should (in my opinion) return spatial features (including geometry and attributes)

adamjstewart:
The names are based on supported file types, not returned keys.

The idea of #2505 is to modify `VectorDataset` such that a single class can be used for semantic/instance segmentation and object/keypoint detection. Otherwise, we would need 4 copies of `EuroCrops` (for example).

sherif-med:
Great, do you think a switcher method within VectorDataset to alter the output of __getitem__ dynamically is a good solution?
In the case of object/keypoint detection, the returned items are not images/masks anymore. Should I take that into account? as different methods like "plot" expect a 'sample dict containing masks tensor'

adamjstewart:
I think we should return all of the above. Then the plot method can also plot all potential outputs. The return value is a dictionary, so we can add as many keys as we want.

adamjstewart:
Is this superseded by #2819?

adamjstewart:
Seems to be superseded, closing.

PR #2796, opened 2025-05-22
Improve numeric stability of QR losses

tayden:
These loss functions sometimes return `nan` values when r or q are 0.

This adds a small epsilon before the log transform to prevent undefined values.

### DIFF
--- a/torchgeo/losses/qr.py
+++ b/torchgeo/losses/qr.py
@@ -31,7 +31,7 @@ class QRLoss(Module):
         q_bar = q.mean(dim=(0, 2, 3))
         qbar_log_S = (q_bar * torch.log(q_bar)).sum()
 
-        q_log_p = torch.einsum('bcxy,bcxy->bxy', q, torch.log(target)).mean()
+        q_log_p = torch.einsum('bcxy,bcxy->bxy', q, torch.log(target + 1e-12)).mean()
 
         loss = qbar_log_S - q_log_p
         return loss
@@ -62,6 +62,6 @@ class RQLoss(Module):
         z = q / q.norm(p=1, dim=(0, 2, 3), keepdim=True).clamp_min(1e-12).expand_as(q)
         r = F.normalize(z * target, p=1, dim=1)
 
-        loss = torch.einsum('bcxy,bcxy->bxy', r, torch.log(r) - torch.log(q)).mean()
+        loss = torch.einsum('bcxy,bcxy->bxy', r, torch.log(r + 1e-12) - torch.log(q + 1e-12)).mean()
 
         return loss


tayden:
It may make sense to make this epsilon value configurable via a new initialization parameter with some sensible default value. Let me know if that is desired and I can make that change.

isaaccorley:
I think it's fine to hardcode it. @adamjstewart any objections?

adamjstewart:
I actually like the idea of an eps parameter. Check the PyTorch docs and see what eps they usually use.

tayden:

tests/losses/test_qr.py
```diff
@@ -18,3 +20,23 @@ def test_loss_on_prior_reversed_kl_simple(self) -> None:
         log_probs = torch.log(probs)
         targets = torch.rand(2, 4, 10, 10)
         RQLoss()(log_probs, targets)
+
+    def test_invalid_eps_value(self) -> None:
+        """Test that the QR loss raises a ValueError when eps is less than 0."""
+        with pytest.raises(ValueError):
```
adamjstewart, line 26:
```suggestion
        with pytest.raises(ValueError, match='Invalid epsilon value'):
```

tests/losses/test_qr.py
```diff
@@ -18,3 +20,23 @@ def test_loss_on_prior_reversed_kl_simple(self) -> None:
...
+            QRLoss(eps=-1.0)
+
+        with pytest.raises(ValueError):
```
adamjstewart, line 29:
```suggestion
        with pytest.raises(ValueError, match='Invalid epsilon value'):
```

torchgeo/losses/qr.py
```diff
@@ -17,6 +17,22 @@ class QRLoss(Module):
...
 
+    def __init__(self, eps: float = 1e-8) -> None:
+        """Initialize a new QRLoss instance.
+
+        Args:
+            eps: small constant for numerical stability to prevent log(0) when computing the loss.
+                Must be greater than or equal to 0.
+
+        Raises:
+            ValueError: If eps is less than 0.
```
adamjstewart, line 28: Can you add:
```rst
.. versionadded:: 0.8
   The *eps* parameter.
```

torchgeo/losses/qr.py
```diff
@@ -46,6 +62,22 @@ class RQLoss(Module):
...
 
+    def __init__(self, eps: float = 1e-8) -> None:
+        """Initialize a new RQLoss instance.
+
+        Args:
+            eps: small constant for numerical stability to prevent division by zero
+                and log(0) when computing the loss. Must be greater than or equal to 0.
+
+        Raises:
+            ValueError: If eps is less than 0.
```
adamjstewart, line 73: Can you add:
```rst
.. versionadded:: 0.8
   The *eps* parameter.
```

adamjstewart:
@estherrolf can you also review?

isaaccorley:
Merged! Thanks for the PR @tayden!

PR #2797, opened 2025-05-24
Add SolarPlantsBrazil dataset for photovoltaic panel detection (semantic segmentation)

FederCO23:
This PR adds the **SolarPlantsBrazil** dataset to TorchGeo.

SolarPlantsBrazil is a geospatial dataset curated for **binary semantic segmentation**, focused on detecting **photovoltaic (PV) power stations** in satellite imagery. It includes **272 manually annotated 256×256 GeoTIFF image patches** from various regions of Brazil. Each image patch comes with:
- A **4-band image** (RGB + NIR, float32)
- A **binary mask** indicating solar panel regions

### Changes introduced:
- Added `SolarPlantsBrazil` dataset class under `torchgeo.datasets`
- Added `SolarPlantsBrazilDataModule` datamodule class under `torchgeo.datamodules`
- Integrated automatic download from Hugging Face: [FederCO23/solar-plants-brazil](https://huggingface.co/datasets/FederCO23/solar-plants-brazil)
- Parsed RGB + NIR float32 GeoTIFFs and binary mask labels
- Implemented a `plot()` method using RGB bands for sample visualization
- Registered dataset in `__init__.py` and `non_geo_datasets.csv`
- Added full unit test coverage under `tests/datasets/test_solar_plants_brazil.py`

### Example output from `plot()`:

Sample_from_Solar_Plants_Brazil

### References:
- [ Hugging Face Dataset](https://huggingface.co/datasets/FederCO23/solar-plants-brazil)
- [ Project Repository](https://github.com/FederCO23/UCSD_MLBootcamp_Capstone)


FederCO23:

FederCO23:

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,189 @@
...
+    * Dataset is hosted on Hugging Face: https://huggingface.co/datasets/FederCO23/solar-plants-brazil
+    * Code and preprocessing steps available at: https://github.com/FederCO23/UCSD_MLBootcamp_Capstone
+
+    If you use this dataset, please cite or reference the project repository.
+
+    .. versionadded:: 0.6
+    """
+
+    url = 'https://huggingface.co/datasets/FederCO23/solar-plants-brazil'
+    bands = ('Red', 'Green', 'Blue', 'NIR')
```
robmarkcole, line 50: I wonder if we could start using enums - [e.g. like terratorch ](https://github.com/IBM/terratorch/blob/main/terratorch/datasets/utils.py
)

robmarkcole:
@FederCO23 care to add a datamodule too?

torchgeo/datasets/solar_plants_brazil.py
adamjstewart: I've thought about it, but there's a lot more info I would like to add (min/max/mean wavelength, bandwidth, etc.). I think a formal class would be better. But that's for another day, we would have to update all existing datasets first.

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,189 @@
...
+    root/train/input/img(123).tif
+    root/train/labels/target(123).tif
+
+    Access:
+    * Dataset is hosted on Hugging Face: https://huggingface.co/datasets/FederCO23/solar-plants-brazil
+    * Code and preprocessing steps available at: https://github.com/FederCO23/UCSD_MLBootcamp_Capstone
+
+    If you use this dataset, please cite or reference the project repository.
+
+    .. versionadded:: 0.6
```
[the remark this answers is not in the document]
FederCO23: done 👍

tests/datasets/test_solar_plants_brazil.py
```diff
@@ -0,0 +1,138 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""Unit tests for the SolarPlantsBrazil dataset."""
+
+import os
+import tempfile
+import warnings
+from unittest.mock import patch
```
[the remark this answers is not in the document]
FederCO23: Replaced unittest.mock.patch with pytest.MonkeyPatch

tests/datasets/test_solar_plants_brazil.py
```diff
@@ -0,0 +1,138 @@
...
+            driver='GTiff',
+            height=256,
+            width=256,
+            count=1,
+            dtype='uint8',
+        ) as dst:
+            dst.write(mask, 1)
+
+
+def test_solar_plants_brazil_getitem_and_len() -> None:
```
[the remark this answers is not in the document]
FederCO23: Refactored tests into a class (TestSolarPlantsBrazil) with individual method.

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,189 @@
...
+
+import glob
+import os
+from collections.abc import Callable
+
+import matplotlib.pyplot as plt
+import numpy as np
+import rasterio
+import torch
+from huggingface_hub import snapshot_download
```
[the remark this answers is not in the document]
FederCO23: Ok. Removed huggingface_hub; now using `download_and_extract_archive`

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,189 @@
...
+
+    If you use this dataset, please cite or reference the project repository.
+
+    .. versionadded:: 0.6
+    """
+
+    url = 'https://huggingface.co/datasets/FederCO23/solar-plants-brazil'
+    bands = ('Red', 'Green', 'Blue', 'NIR')
+
+    citation = """\
```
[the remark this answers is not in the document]
FederCO23: Ok. Removed citation block from the dataset class

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,189 @@
...
+
+        self.image_paths = sorted(
+            glob.glob(os.path.join(self.dataset_path, 'input', 'img(*).tif'))
+        )
+        self.mask_paths = sorted(
+            glob.glob(os.path.join(self.dataset_path, 'labels', 'target(*).tif'))
+        )
+
+        if len(self.image_paths) == 0:
+            raise FileNotFoundError(
```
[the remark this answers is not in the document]
FederCO23: Ok. Replaced FileNotFoundError with DatasetNotFoundError, and documented it in the docstring.

robmarkcole:
Test failures same as https://github.com/microsoft/torchgeo/pull/2707#issuecomment-2908777476

robmarkcole:
@FederCO23 hit the `Update branch` button as the failing test is fixed on main

FederCO23:
> @FederCO23 hit the `Update branch` button as the failing test is fixed on main

@robmarkcole, I’ve updated the branch as requested and all checks are passing.
I also addressed all the requested changes from @adamjstewart and left a comment summarizing them above.
Let me know if anything else is needed on my end. Thanks again!

docs/api/datasets/non_geo_datasets.csv
```diff
@@ -55,6 +55,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `SkyScript`_,IC,"NAIP, orthophotos, Planet SkySat, Sentinel-2, Landsat 8--9",MIT,5.2M,-,100--1000,0.1--30,RGB
 `So2Sat`_,C,Sentinel-1/2,"CC-BY-4.0","400,673",17,32x32,10,"SAR, MSI"
 `SODA`_,OD,Aerial,"CC-BY-NC-4.0","2513",9,"~2700x~4800","RGB"
+`SolarPlantsBrazil`_,S,Aerial,"CC-BY-NC-4.0","272",2,256x256,10,"RGB + NIR"
```
adamjstewart: The license on https://huggingface.co/datasets/FederCO23/solar-plants-brazil says MIT?

docs/api/datasets.rst
```diff
@@ -492,6 +492,10 @@ So2Sat
 
 .. autoclass:: So2Sat
 
+SolarPlantsBrazil
```
adamjstewart: Should the name be "Solar Plants Brazil" with spaces?

tests/datamodules/test_solar_plants_brazil.py
adamjstewart, line 1: Let's get rid of this file and properly test the datamodule with its trainer. So you would add a file in `tests/conf/` and one line in `tests/trainers/test_segmentation.py`.

tests/datasets/test_solar_plants_brazil.py
```diff
@@ -0,0 +1,106 @@
...
+import torch
+from matplotlib import pyplot as plt
+
+from torchgeo.datasets import DatasetNotFoundError, SolarPlantsBrazil
+
+
+class TestSolarPlantsBrazil:
+    @pytest.fixture
+    def dataset_root(self) -> str:
+        # Point to the test data folder you generated and committed
```
adamjstewart: Who is "you"? Also, this path should use `os.path.join` for Windows support

tests/datasets/test_solar_plants_brazil.py
```diff
@@ -0,0 +1,106 @@
...
+        return 'tests/data/solar_plants_brazil'
+
+    @pytest.fixture(params=['train'])
```
adamjstewart: If there's only one value being tested, no need to parametrize

tests/datasets/test_solar_plants_brazil.py
```diff
@@ -0,0 +1,106 @@
...
+            return sample
+
+        dataset = SolarPlantsBrazil(
+            root=dataset_root, split='train', transforms=dummy_transform
+        )
+        sample = dataset[0]
+        assert torch.all(sample['image'] > 0)
+
+
+def test_download_called(monkeypatch: pytest.MonkeyPatch, tmp_path: Path) -> None:
```
adamjstewart: Can't we test this like we do with other datasets?

tests/datasets/test_solar_plants_brazil.py
```diff
@@ -0,0 +1,106 @@
...
+        'torchgeo.datasets.solar_plants_brazil.download_and_extract_archive',
+        fake_download_and_extract_archive,
+    )
+
+    _ = SolarPlantsBrazil(root=tmp_path, split='train', download=True)
+
+    assert called['triggered']
+
+
+def test_missing_dataset_triggers_error(tmp_path: Path) -> None:
```
adamjstewart: These should be added to the class

torchgeo/datamodules/solar_plants_brazil.py
```diff
@@ -0,0 +1,104 @@
...
+
+import kornia.augmentation as K
+import torch
+from torch import Tensor
+
+from ..datasets import SolarPlantsBrazil
+from ..samplers.utils import _to_tuple
+from .geo import NonGeoDataModule
+
+# Per-channel statistics (mean and std) computed over the entire dataset.
```
adamjstewart: Shouldn't they be computed only on the train split?

torchgeo/datamodules/solar_plants_brazil.py
```diff
@@ -0,0 +1,104 @@
...
+        )
+
+        self.aug = K.AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
+            K.CenterCrop(size=self.patch_size),
+            data_keys=None,
+            keepdim=True,
+        )
+
+    def setup(self, stage: str) -> None:
```
adamjstewart: This can be removed, it's identical to the base class

torchgeo/datamodules/solar_plants_brazil.py
```diff
@@ -0,0 +1,104 @@
...
+
+        Args:
+            batch: A batch of data that needs to be altered or augmented.
+            dataloader_idx: The index of the dataloader to which the batch belongs.
+
+        Returns:
+            A batch of data.
+
+        """
+        # This solves a special case where if batch_size=1 the mask won't be stacked correctly
```
adamjstewart: This seems like a bug in the dataset, can we fix this?

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+class SolarPlantsBrazil(NonGeoDataset):
+    """Solar Plants Brazil dataset (semantic segmentation for photovoltaic detection).
+
+    The `Solar Plants Brazil <https://huggingface.co/datasets/FederCO23/solar-plants-brazil>`__
+    dataset provides satellite imagery and pixel-level annotations for detecting photovoltaic
+    solar power stations.
+
+    Dataset features:
```
adamjstewart, line 30: You need an empty line before and after bulleted lists, otherwise Sphinx gets confused and your lists are not formatted correctly: https://torchgeo--2797.org.readthedocs.build/en/2797/api/datasets.html#solarplantsbrazil

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,176 @@
...
+    root/train/labels/target(123).tif
+
+    Access:
+    * Dataset is hosted on Hugging Face: https://huggingface.co/datasets/FederCO23/solar-plants-brazil
+    * Code and preprocessing steps available at: https://github.com/FederCO23/UCSD_MLBootcamp_Capstone
+
+    .. versionadded:: 0.8
+
+    Raises:
+        DatasetNotFoundError: If the dataset is not found and download=False.
```
adamjstewart: Let's move this to `__init__`

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,176 @@
...
+    """
+
+    url = 'https://huggingface.co/datasets/FederCO23/solar-plants-brazil/resolve/main/solarplantsbrazil.zip'
+    bands = ('Red', 'Green', 'Blue', 'NIR')
+
+    def __init__(
+        self,
+        root: Path = 'data',
+        split: str = 'train',
```
adamjstewart:
```suggestion
        split: Literal['train', 'val', 'test'] = 'train',
```
Not all str are valid

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,176 @@
...
+        transforms: Callable[[dict[str, Tensor]], dict[str, Tensor]] | None = None,
+        download: bool = False,
+    ) -> None:
+        """Initialize a SolarPlantsBrazil dataset split.
+
+        Args:
+            root: Root directory where dataset is stored.
+            split: Dataset split to use, one of "train", "val", or "test".
+            transforms: Optional transforms to apply.
+            download: If True, download the dataset if it doesn't exist.
```
adamjstewart: Can we also add checksum support?

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,176 @@
...
+        self._download()
+
+    def _download(self) -> None:
+        """Download the dataset archive from Hugging Face and extract it."""
+        download_and_extract_archive(
+            url=self.url, download_root=self.root, filename='solarplantsbrazil.zip'
+        )
+
+    def __getitem__(self, index: int) -> dict[str, Tensor]:
+        """Retrieve an image-mask pair by index."""
```
adamjstewart: Every function docstring should define inputs (Args) and outputs (Returns) other than self/None

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,176 @@
...
+        return torch.from_numpy(arr)
+
+    def _load_mask(self, path: str) -> Tensor:
+        """Load a binary mask as a torch tensor with shape (1, H, W)."""
+        with rasterio.open(path) as src:
+            arr = src.read(1).astype(np.uint8)
+        bin_mask = (arr > 0).astype(np.uint8)
+        return torch.from_numpy(bin_mask).unsqueeze(0).long()
+
+    def plot(
```
adamjstewart, line 175: Can you add a sample plot image to your PR description?

docs/api/datasets/non_geo_datasets.csv
FederCO23: Updated. The dataset license on Hugging Face is now set to CC-BY-NC 4.0 to reflect the intended non-commercial use

docs/api/datasets.rst
FederCO23: Renamed to "Solar Plants Brazil" in datasets.rst as suggested

tests/datamodules/test_solar_plants_brazil.py
FederCO23: Removed `test_solar_plants_brazil.py`, added `tests/conf/solar_plants_brazil.yaml`, updated `test_segmentation.py`, and refreshed the dummy dataset with 256×256 instead of 32×32 images. Trainer + DataModule test runs successfully.

tests/datasets/test_solar_plants_brazil.py
FederCO23: Updated the return value to use `os.path.join` for cross-platform support. Also removed the comment, thanks for flagging it!

tests/datasets/test_solar_plants_brazil.py
FederCO23: Removed the unnecessary `@pytest.fixture(params=['train'])` since we’re only testing a single split

tests/datasets/test_solar_plants_brazil.py
FederCO23: Updated the `test_download_called` test to follow the same structure used in other dataset tests

tests/datasets/test_solar_plants_brazil.py
FederCO23: Moved the `test_missing_dataset_triggers_error` and `test_empty_split_folder_triggers_error` methods into the `TestSolarPlantsBrazil` class for consistency

torchgeo/datamodules/solar_plants_brazil.py
FederCO23: The per-channel statistics are now computed only from the training split avoiding data leakage, good catch!

torchgeo/datamodules/solar_plants_brazil.py
FederCO23: Removed the setup method since it’s redundant with the superclass. Cleaned up as suggested

torchgeo/datamodules/solar_plants_brazil.py
FederCO23: The dataset now returns mask with shape [H, W] instead of [1, H, W], aligning with the expected format for the loss function.
Unit tests and trainer test pass successfully

torchgeo/datasets/solar_plants_brazil.py
FederCO23: Adjustments done. The docstring now renders correctly in the HTML

torchgeo/datasets/solar_plants_brazil.py
FederCO23: Ok. It’s done

torchgeo/datasets/solar_plants_brazil.py
FederCO23: Replaced `split: str = 'train'` with `split: Literal['train', 'val', 'test'] = 'train'` to restrict valid values, as suggested. Thanks for pointing that out

torchgeo/datasets/solar_plants_brazil.py
FederCO23: Updated the `__init__` docstring to follow the standard format with Args: and Returns: sections. Also reviewed all other methods to ensure parameters are explicitly documented

torchgeo/datasets/solar_plants_brazil.py
FederCO23: Added a sample output image from the `plot()` method to the PR description

torchgeo/datasets/solar_plants_brazil.py
FederCO23: Ok. Added optional checksum argument to enable MD5 validation of the downloaded .zip file

FederCO23:
@adamjstewart, thanks again for the detailed feedback!
All 16 requested changes have been addressed. I replied to each one directly above.
Let me know if there’s anything else needed on my end. Happy to contribute!

@robmarkcole, just tagging you here to keep you in the loop

adamjstewart:
Accidentally deleted my notification, but remind me to review this again this week.

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,230 @@
...
+    Access:
+
+    * Dataset is hosted on Hugging Face: https://huggingface.co/datasets/FederCO23/solar-plants-brazil
+    * Code and preprocessing steps available at: https://github.com/FederCO23/UCSD_MLBootcamp_Capstone
+
+    .. versionadded:: 0.8
+
+    """
+
+    url = 'https://huggingface.co/datasets/FederCO23/solar-plants-brazil/resolve/main/solarplantsbrazil.zip'
```
adamjstewart:
```suggestion
    url = 'https://huggingface.co/datasets/FederCO23/solar-plants-brazil/resolve/1dc13a453ef6acabf08a1781c523fd1db3d9bcc5/solarplantsbrazil.zip'
```
Use a stable commit hash so that the MD5 doesn't change even if the dataset is later updated.

docs/api/datasets/non_geo_datasets.csv
```diff
@@ -55,6 +55,7 @@ Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
 `SkyScript`_,IC,"NAIP, orthophotos, Planet SkySat, Sentinel-2, Landsat 8--9",MIT,5.2M,-,100--1000,0.1--30,RGB
 `So2Sat`_,C,Sentinel-1/2,"CC-BY-4.0","400,673",17,32x32,10,"SAR, MSI"
 `SODA`_,OD,Aerial,"CC-BY-NC-4.0","2513",9,"~2700x~4800","RGB"
+`Solar Plants Brazil`_,S,Aerial,"CC-BY-NC-4.0","272",2,256x256,10,"RGB + NIR"
```
adamjstewart, line 58: License now matches on https://huggingface.co/datasets/FederCO23/solar-plants-brazil

tests/datasets/test_solar_plants_brazil.py
```diff
@@ -0,0 +1,100 @@
...
+        assert isinstance(sample['mask'], torch.Tensor)
+        assert sample['image'].shape == (4, 256, 256)
+        assert sample['mask'].shape == (256, 256)
+
+    def test_len(self, dataset: SolarPlantsBrazil) -> None:
+        assert len(dataset) == 1
+
+    def test_plot(self, dataset: SolarPlantsBrazil) -> None:
+        sample = dataset[0].copy()
+        sample['prediction'] = sample['mask'].clone()
```
adamjstewart: I don't think a copy or clone is actually needed is it?

tests/datasets/test_solar_plants_brazil.py
```diff
@@ -0,0 +1,100 @@
...
+            sample['image'] += 1
+            return sample
+
+        dataset = SolarPlantsBrazil(
+            root=dataset_root, split='train', transforms=dummy_transform
+        )
+        sample = dataset[0]
+        assert torch.all(sample['image'] > 0)
+
+    def test_download_called(
```
adamjstewart: I don't think we test this for any other datasets, it's better to test that it's possible to "download" files from the `tests/data` directory to the test directory. See all of the other dataset tests.

tests/datasets/test_solar_plants_brazil.py
```diff
@@ -0,0 +1,100 @@
...
+        monkeypatch.setattr(
+            'torchgeo.datasets.solar_plants_brazil.download_and_extract_archive',
+            fake_download_and_extract_archive,
+        )
+
+        _ = SolarPlantsBrazil(root=tmp_path, split='train', download=True)
+
+        assert called['triggered']
+
+    def test_missing_dataset_triggers_error(self, tmp_path: Path) -> None:
```
adamjstewart: Isn't this identical to `test_missing_dataset_raises`?

torchgeo/datamodules/solar_plants_brazil.py
```diff
@@ -0,0 +1,89 @@
...
+        num_workers: int = 0,
+        **kwargs: Any,
+    ) -> None:
+        """Initialize the SolarPlantsBrazilDataModule.
+
+        Args:
+            batch_size: Number of samples per batch.
+            patch_size: Spatial dimensions (H, W) to crop from images.
+            num_workers: Number of subprocesses used to load the data.
+            **kwargs: Additional arguments passed to the dataset.
```
adamjstewart: Passed to which dataset? We should link to the docs for that dataset like we do in all other datamodules

torchgeo/datamodules/solar_plants_brazil.py
```diff
@@ -0,0 +1,89 @@
...
+
+        Returns:
+            None
```
adamjstewart: This can be skipped if there is no Return value

torchgeo/datamodules/solar_plants_brazil.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+
+        self.aug = K.AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
+            K.CenterCrop(size=self.patch_size),
+            data_keys=None,
+            keepdim=True,
+        )
+
+    def on_after_batch_transfer(
```
adamjstewart: This isn't doing anything and can be deleted

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,230 @@
...
+    bands = ('Red', 'Green', 'Blue', 'NIR')
+    md5 = 'DFA0D3EFDEF4143A33B0E7BA834EAAFA'
```
adamjstewart:
```suggestion
    md5 = 'dfa0d3efdef4143a33b0e7ba834eaafa'
```
Upper case doesn't work:
```pycon
> python3
Python 3.13.2 (main, Apr  8 2025, 15:27:33) [Clang 17.0.0 (clang-1700.0.13.3)] on darwin
>>> from torchgeo.datasets import SolarPlantsBrazil
>>> ds = SolarPlantsBrazil('data/solarplantsbrazil', download=True, checksum=True)
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 134M/134M [00:06<00:00, 22.0MB/s]
  File "<python-input-1>", line 1, in <module>
    ds = SolarPlantsBrazil('data/solarplantsbrazil', download=True, checksum=True)
  File "/Users/Adam/Downloads/torchgeo/torchgeo/datasets/solar_plants_brazil.py", line 95, in __init__
    self._verify()
    ~~~~~~~~~~~~^^
  File "/Users/Adam/Downloads/torchgeo/torchgeo/datasets/solar_plants_brazil.py", line 123, in _verify
    self._download()
    ~~~~~~~~~~~~~~^^
  File "/Users/Adam/Downloads/torchgeo/torchgeo/datasets/solar_plants_brazil.py", line 131, in _download
    download_and_extract_archive(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        url=self.url,
        ^^^^^^^^^^^^^
    ...<2 lines>...
        md5=self.md5 if self.checksum else None,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/Adam/spack/var/spack/environments/default/.spack-env/view/lib/python3.13/site-packages/torchvision/datasets/utils.py", line 391, in download_and_extract_archive
    download_url(url, download_root, filename, md5)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Adam/spack/var/spack/environments/default/.spack-env/view/lib/python3.13/site-packages/torchvision/datasets/utils.py", line 140, in download_url
    raise RuntimeError("File not found or corrupted.")
RuntimeError: File not found or corrupted.
```

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,230 @@
...
+            transforms: Optional transforms to apply.
+            download: If True, download the dataset if it doesn't exist.
+            checksum: If True, verify MD5 checksum of the downloaded file.
+
+        Raises:
+            DatasetNotFoundError: If the dataset is not found and
+                ``download=False``.
+
+        Returns:
+            None
```
adamjstewart: Can skip Returns if nothing is returned

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,230 @@
...
+
+        """
+        if split not in ['train', 'val', 'test']:
+            raise ValueError(
+                f"Invalid split '{split}', expected one of: 'train', 'val', or 'test'"
+            )
```
adamjstewart, line 84: Could honestly remove this and the test since we statically check that a valid Literal is used, but up to you. This is certainly nicer for users as it gives them a more clear explanation and is checked at runtime.

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,230 @@
...
+
+        assert len(self.image_paths) == len(self.mask_paths), (
+            'Mismatch between image and mask files'
+        )
+
+    def _verify(self) -> None:
+        """Verify the dataset exists or download it.
+
+        Returns:
+            None
```
adamjstewart: Skip

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,230 @@
...
+
+        self.image_paths = sorted(
+            glob.glob(os.path.join(self.dataset_path, 'input', 'img(*).tif'))
+        )
+        self.mask_paths = sorted(
+            glob.glob(os.path.join(self.dataset_path, 'labels', 'target(*).tif'))
+        )
+
+        if len(self.image_paths) == 0:
+            raise DatasetNotFoundError(self)
```
adamjstewart: Isn't this checked by `_verify`?

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,230 @@
...
+        if not self.download:
+            raise DatasetNotFoundError(self)
+
+        self._download()
+
+    def _download(self) -> None:
+        """Download the dataset archive from Hugging Face and extract it.
+
+        Returns:
+            None
```
adamjstewart: Skip

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,230 @@
...
+            download_root=self.root,
+            filename='solarplantsbrazil.zip',
+            md5=self.md5 if self.checksum else None,
+        )
+
+    def __getitem__(self, index: int) -> dict[str, Tensor]:
+        """Retrieve an image-mask pair by index.
+
+        Args:
+            index (int): Index of the sample to retrieve.
```
adamjstewart:
```suggestion
            index: Index of the sample to retrieve.
```
Already gotten from type hints

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,230 @@
...
+
+        Returns:
+            dict: A dictionary with the following keys:
+                - 'image': A float32 tensor of shape (C, H, W)
+                - 'mask': A long tensor of shape (1, H, W), containing binary labels
```
adamjstewart: This doesn't seem to format correctly, maybe try:
```suggestion
            A dictionary with the following keys:
            
            - 'image': A float32 tensor of shape (C, H, W)
            - 'mask': A long tensor of shape (1, H, W), containing binary labels
```
or just copy-n-paste the same description used by the other 125+ builtin datasets.

adamjstewart (changes requested):
Still quite a lot of differences between this dataset and the other builtin datasets

torchgeo/datamodules/solar_plants_brazil.py
```diff
@@ -0,0 +1,89 @@
...
+MEAN = torch.tensor([927.7570, 740.1440, 492.3968, 2441.6775])
+STD = torch.tensor([544.8361, 311.5538, 252.4914, 651.2599])
+
+
+class SolarPlantsBrazilDataModule(NonGeoDataModule):
+    """LightningDataModule for SolarPlantsBrazil dataset.
+
+    This datamodule wraps the SolarPlantsBrazil dataset, which contains
+    predefined train/val/test splits. This design ensures spatial separation
+    between samples by solar plant, preventing data leakage during training.
```
adamjstewart, line 26:
```rst
.. versionadded:: 0.8
```

torchgeo/datasets/solar_plants_brazil.py
FederCO23: Ok. Updated the dataset URL to use a fixed commit hash to ensure stable MD5

tests/datasets/test_solar_plants_brazil.py
FederCO23: You're right, since prediction is not being modified, both `.clone()` and `.copy()` are unnecessary and can be removed.
For reference, I had initially followed a similar pattern used in other datasets like `cloud_cover` and `chesapeake`, which use `.clone()` in their `test_plot()` methods."

tests/datasets/test_solar_plants_brazil.py
FederCO23: Ok. Removed `test_download_called` and added `test_already_downloaded` and `test_not_downloaded` instead, following the structure used in other datasets.

tests/datasets/test_solar_plants_brazil.py
FederCO23: Removed `test_missing_dataset_triggers_error` (duplicate of `test_not_downloaded`) and renamed for consistency with other datasets

torchgeo/datamodules/solar_plants_brazil.py
FederCO23: Ok. I've updated the docstring to include a link to the `SolarPlantsBrazil` dataset documentation

torchgeo/datamodules/solar_plants_brazil.py
FederCO23: Removed the Returns section from the docstrings

torchgeo/datamodules/solar_plants_brazil.py
FederCO23: You're right, since no custom logic is added here. I've removed the `on_after_batch_transfer override`.

torchgeo/datasets/solar_plants_brazil.py
FederCO23: Yes, it’s now lowercase. I had tested it previously and it worked, most likely because I didn’t enable the checksum flag during that run...

torchgeo/datasets/solar_plants_brazil.py
FederCO23: I prefer to keep the runtime check and corresponding test to ensure clearer error messages, especially for users working in interactive environments like notebooks or without static type checking. Is it ok for you?

torchgeo/datasets/solar_plants_brazil.py
FederCO23: I’ve now removed the `if len(self.image_paths) == 0` check from `__init__()` to align with the datasets pattern.
That said, I originally added this check to catch cases where the folder exists but is empty (ex: failed unzip). In that situation, `_verify()` would silently pass, but indexing into the dataset would fail later on in a less clear way.
I transfer that check to `_verify()`

torchgeo/datasets/solar_plants_brazil.py
FederCO23: Ok, this has been addressed

torchgeo/datasets/solar_plants_brazil.py
FederCO23: Got it. I'll simplify the docstring to match the style used by the other datasets. Thanks for the tip!

torchgeo/datamodules/solar_plants_brazil.py
FederCO23: Ok, included 🙌

tests/datasets/test_solar_plants_brazil.py
adamjstewart: Yeah, we should probably simplify those tests too. Not a big deal, just something I noticed during review.

torchgeo/datasets/solar_plants_brazil.py
adamjstewart: Works for me.

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,221 @@
...
+        Args:
+            root: root directory where dataset can be found
+            split: dataset split to use, one of "train", "val", or "test"
+            transforms: a function/transform that takes an input sample
+                and returns a transformed version
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+
+        Raises:
+            DatasetNotFoundError: If dataset is not found and *download* is False
```
adamjstewart:
```suggestion
            DatasetNotFoundError: If dataset is not found and *download* is False.
            ValueError: If *split* is invalid.
```

torchgeo/datasets/solar_plants_brazil.py
```diff
@@ -0,0 +1,221 @@
...
+        self.root = root
+        self.transforms = transforms
+        self.dataset_path = os.path.join(self.root, split)
+        self.split = split
+        self.download = download
+        self.checksum = checksum
+
+        self._verify()
+
+    def _verify(self) -> None:
```
adamjstewart, line 101: Most datasets have 3 logical sections:

1. files are extracted: return
2. zipfile is downloaded but not extracted: extract zip file and return
3. nothing is downloaded: download and extract

You're currently missing the second one. Not a huge deal, but could help make the dataset easier to use.

adamjstewart (changes requested):
Code looks much better now, thanks for all of the hard work!

Only remaining thing is to iron out the download logic and testing so we reach 100% coverage. The EuroSAT zipfile download logic is pretty clean and a good example:
```python
        directory = os.path.join(self.root, self.base_dir)                               
        zipfile = os.path.join(self.root, self.filename)                                 
        if os.path.isdir(directory):                                                     
            return                                                                       
        elif os.path.isfile(zipfile):                                                    
            extract_archive(zipfile)                                                     
        elif self.download:                                                              
            download_and_extract_archive(                                                
                self.url + self.filename,                                                
                self.root,                                                               
                md5=self.md5 if self.checksum else None,                                 
            )                                                                            
        else:                                                                            
```
I actually don't like the EuroSAT tests that much. TreeSatAI has nice simple tests:
```python
    def test_download(self, dataset: TreeSatAI, tmp_path: Path, monkeypatch: MonkeyPatch) -> None:                
        monkeypatch.setattr(TreeSatAI, 'url', local_url) 
        TreeSatAI(tmp_path, download=True)                                               
                                                                                         
    def test_extract(self, dataset: TreeSatAI, tmp_path: Path) -> None:                  
        for file in glob.iglob(os.path.join(root, '*.*')):                               
            shutil.copy(file, tmp_path)                                                  
        TreeSatAI(tmp_path)                                                              
                                                                                         
    def test_not_downloaded(self, tmp_path: Path) -> None:                               
        with pytest.raises(DatasetNotFoundError, match='Dataset not found'):             
            TreeSatAI(tmp_path)    
```
Something simple like that should get you 100% test coverage.

torchgeo/datasets/solar_plants_brazil.py
FederCO23: Ok, added it.

torchgeo/datasets/solar_plants_brazil.py
FederCO23: I appreciate the clear guidance!
I’ve updated the `_verify()` method to handle the intermediate case where the zip file exists but hasn't been extracted yet, inspired by the EuroSAT logic as suggested, and also aligned the tests more closely with the TreeSatAI style.

Let me know if anything else needs attention. Thanks!

adamjstewart (approved):
Looks good now! @calebrob6 may be interested in this to help validate his global solar panel detection work.

PR #2798, opened 2025-05-24
AI4G Sentinel-1 U-Net Flood Detection Model

isaaccorley:
This PR adds the the Sentinel-1 Flood Detection pretrained smp U-Net model from the AI4G paper: ["Mapping Global Floods with 10 Years of Satellite Radar Data", Misra et al.](https://arxiv.org/abs/2411.01411)

This model is basically a change detection model that takes as input 2 stacked before and after Sentinel-1 images. Each image has 2 bands (VV & VH). The only caveat here is that the transform creates a change map from the pre and post imagery reducing the channels from 4 -> 2. The model then takes in the 2-channel change map.

cc: @amitmisra1587

### Sample

For example, before and after Sentinel-1 images from the Kenya floods in 2024

Before FCI
image

After FCI
image

VV Change Map
image

VH Change Map
image

Model Predictions (binary mask)
image

Model Predictions (binary mask burned onto after FCI image)
image


torchgeo/models/unet.py
```diff
@@ -96,6 +200,22 @@ class Unet_Weights(WeightsEnum):  # type: ignore[misc]
             'license': 'non-commercial',
         },
     )
+    SENTINEL1_AI4G_FLOOD = Weights(
+        url='https://huggingface.co/torchgeo/ai4g_flood/resolve/672bfb53b61a91114941ac9e4338ebc96dff6ec7/unet_mobilenetv2_sentinel1_ai4g_flood-d95df7aa.pth',
+        transforms=_ai4g_flood_transforms,
+        meta={
+            'dataset': 'AI4G Global Flood Extent',
+            'in_chans': 2,
+            'in_chans_transform': 4,
```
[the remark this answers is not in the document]
isaaccorley: I think the issue is that the model's input layer only takes 2 channels which could be confusing to a user. I could also see the case where the user defines the change map transform differently (maybe the average pixel change across a longer time series). I'm going to revamp this anyway in another PR because I think we should require a minimum schema of several metadata fields to be filled out for each model.

torchgeo/models/unet.py
```diff
@@ -8,19 +8,123 @@
...
+
+        if not has_batch_dim:
+            change = change.squeeze(dim=0)
+
+        return change
+
+
+_ai4g_flood_transforms = K.AugmentationSequential(
+    # Convert to decibel scale and shift to [0, 255] range
+    K.ImageSequential(Lambda(lambda x: 2 * 10 * torch.log10(x) + 135)),
```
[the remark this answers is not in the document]
isaaccorley: Added to `torchgeo/transforms/intensity.py`

torchgeo/transforms/intensity.py
```diff
@@ -0,0 +1,141 @@
...
+
+"""TorchGeo intensity transforms."""
+
+import torch
+from kornia.augmentation import IntensityAugmentationBase2D
+from torch import Tensor
+
+
+class ToDecibelScale(IntensityAugmentationBase2D):
+    """Convert input tensor to decibel scale.
```
adamjstewart: Maybe we should clarify that this is power to decibel, as opposed to amplitude to decibel.

torchgeo/transforms/intensity.py
adamjstewart, line 1: Wondering if this should be `sar.py` instead of `intensity.py`. It doesn't really matter, as we can rename it later. We will likely want to add many more SAR transforms in the future, both for intensity changes and for geometric changes.

torchgeo/transforms/intensity.py
```diff
@@ -0,0 +1,141 @@
...
+            transform: The geometric transformation tensor.
+
+        Returns:
+            The transformed input.
+        """
+        out = flags['scale'] * 10 * torch.log10(input + 1e-6) + flags['shift']
+        return out
+
+
+class Sentinel1ChangeMap(IntensityAugmentationBase2D):
```
adamjstewart: Not 100% sure if this should be public. Is it general purpose, or really specific to a single paper? Maybe we could add more description to the docstring explaining what it is for and how it works?

tests/transforms/test_intensity.py
```diff
@@ -0,0 +1,55 @@
...
+import kornia.augmentation as K
+import pytest
+import torch
+from torch import Tensor
+
+from torchgeo.transforms import Sentinel1ChangeMap, ToDecibelScale
+
+
+@pytest.fixture
+def sample() -> dict[str, Tensor]:
```
adamjstewart, line 13: Someday (not necessarily in this PR) we should probably add a `tests/transforms/conftest.py` to define all of these shared fixtures.

adamjstewart:
Curious if this has been tested with our new ChangeDetectionTask.

torchgeo/transforms/intensity.py
isaaccorley: We can keep it in intensity for now and then move to sar once we have more sar transforms.

torchgeo/transforms/intensity.py
isaaccorley: It's pretty general imo. It's just an operation to compute the difference between 2 images and then threshold any change that's above some magnitude to be considered "change" vs "no change".

isaaccorley: I'll add some more details to the docstring.

isaaccorley:
> Curious if this has been tested with our new ChangeDetectionTask.

I would if the dataset was released

adamjstewart:
We could test it with other S1 change detection datasets though.

torchgeo/transforms/intensity.py
```diff
@@ -0,0 +1,145 @@
...
+# Licensed under the MIT License.
+
+"""TorchGeo intensity transforms."""
+
+import torch
+from kornia.augmentation import IntensityAugmentationBase2D
+from torch import Tensor
+
+
+class ToDecibelScale(IntensityAugmentationBase2D):
```
adamjstewart: I would change the name to `PowerToDecibel`. Otherwise, when we add amplitude -> decibel, there will be a name conflict.

torchgeo/transforms/intensity.py
```diff
@@ -0,0 +1,145 @@
...
+            transform: The geometric transformation tensor.
+
+        Returns:
+            The transformed input.
+        """
+        out = flags['scale'] * 10 * torch.log10(input + 1e-6) + flags['shift']
+        return out
+
+
+class Sentinel1ChangeMap(IntensityAugmentationBase2D):
```
adamjstewart: If this is general purpose, maybe we can rename it to `Threshold` or something like that? Since it doesn't have to be S1 or CD.

isaaccorley: Yeah I think I can likely make it more general like take in a list of thresholds so it's not fixed to 4 bands

isaaccorley: It's not only thresholding though. It's generating a binary change mask by taking the first half of the channels and subtracting them from the second half of the channels.

adamjstewart: Maybe 2 transforms? One for change detection/comparison and one for thresholding? Idk, don't want to make this too complicated. Use your best judgement.

adamjstewart: Could also move the subtraction part back into the model and only have a post-processing transform for thresholding if that makes more sense. Really just depends on how general purpose this is. I don't want transforms that only make sense and will only be used with one model.

I assume this part of the model has no learnable weights, so it doesn't matter whether this goes in the model or in a transform?

adamjstewart:
Will review again when tests are passing.

PR #2804, opened 2025-05-29
GeoDataset: add support for slicing

adamjstewart:
This PR adds support for GeoDataset slicing, and is part of ongoing work to add time series support to TorchGeo: #2382.

### Before

```python
ds[BoundingBox(xmin, xmax, ymin, ymax, tmin, tmax)]
```

### After

```python
ds[xmin:xmax, ymin:ymax]
ds[:, :, tmin:tmax]
ds[xmin:xmax:xres, ymin:ymax:yres, tmin:tmax:tres]
```

### Progress

- [x] Dataset Base Classes
  - [x] GeoDataset
  - [x] RasterDataset
  - [x] VectorDataset
  - [x] IntersectionDataset
  - [x] UnionDataset
- [x] Custom Datasets
  - [x] AgriFieldNet
  - [x] Chesapeake CVPR
  - [x] EDDMapS
  - [x] EnviroAtlas
  - [x] GBIF
  - [x] GlobBiomass
  - [x] iNaturalist
  - [x] LandCover.ai
  - [x] Open Buildings
  - [x] South Africa Crop Type
- [x] Samplers
  - [x] GeoSampler
  - [x] RandomGeoSampler
  - [x] GridGeoSampler
  - [x] PreChippedGeoSampler
  - [x] BatchGeoSampler
  - [x] RandomBatchGeoSampler
- [x] Documentation
- [x] Testing

### Backwards-Incompatible Changes

* GeoDatasets are now indexed by GeoSlice, not BoundingBox
* GeoDataset.bounds now returns a GeoSlice, not BoundingBox
* Returned sample bounds are now a GeoSlice, not BoundingBox
* get_random_bounding_box now takes a tuple as input and returns a tuple
* tile_to_chips now takes a tuple as input

### DIFF
--- a/torchgeo/datamodules/naip.py
+++ b/torchgeo/datamodules/naip.py
@@ -82,24 +82,22 @@ class NAIPChesapeakeDataModule(GeoDataModule):
         self.chesapeake = dc | de | md | ny | pa | va | wv
         self.dataset = self.naip & self.chesapeake
 
-        roi = self.dataset.bounds
-        midx = roi.minx + (roi.maxx - roi.minx) / 2
-        midy = roi.miny + (roi.maxy - roi.miny) / 2
+        x, y, t = self.dataset.bounds
+        midx = x.start + (x.stop - x.start) / 2
+        midy = y.start + (y.stop - y.start) / 2
 
         if stage in ['fit']:
-            train_roi = BoundingBox(
-                roi.minx, midx, roi.miny, roi.maxy, roi.mint, roi.maxt
-            )
+            train_roi = BoundingBox(x.start, midx, y.start, y.stop, t.start, t.stop)
             self.train_batch_sampler = RandomBatchGeoSampler(
                 self.dataset, self.patch_size, self.batch_size, self.length, train_roi
             )
         if stage in ['fit', 'validate']:
-            val_roi = BoundingBox(midx, roi.maxx, roi.miny, midy, roi.mint, roi.maxt)
+            val_roi = BoundingBox(midx, x.stop, y.start, midy, t.start, t.stop)
             self.val_sampler = GridGeoSampler(
                 self.dataset, self.patch_size, self.patch_size, val_roi
             )
         if stage in ['test']:
-            test_roi = BoundingBox(midx, roi.maxx, midy, roi.maxy, roi.mint, roi.maxt)
+            test_roi = BoundingBox(midx, x.stop, midy, y.stop, t.start, t.stop)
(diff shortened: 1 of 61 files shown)


torchgeo/datasets/geo.py
```diff
@@ -111,18 +111,49 @@ class GeoDataset(Dataset[dict[str, Any]], abc.ABC):
...
+        xres, yres = self.res
+        tres = 1
+        out = [xmin, xmax, xres, ymin, ymax, yres, tmin, tmax, tres]
+
+        if isinstance(key, slice):
+            key = (key,)
+
+        for i in range(len(key)):
+            if key[i].start is not None:
+                out[i * 3 + 0] = key[i].start
```
isaaccorley: Why are we multiplying by 3?

isaaccorley (commented):
this approach lgtm just one small question

torchgeo/datasets/geo.py
adamjstewart: There are up to 3 slices (x, y, t), each with 3 components (start, stop, step). In order to get (0, 1, 2), (3, 4, 5), (6, 7, 8) indices we need to multiply by 3 (for 3 components).

isaaccorley: Can you add a comment in the code here for that. I think this will confuse more than just me when reading tbh. Makes sense though

adamjstewart: Added a couple comments, not sure if it's what you're looking for though.

adamjstewart: Ended up changing the syntax anyway, no more multiplication.

torchgeo/datamodules/naip.py
```diff
@@ -82,26 +82,22 @@ def setup(self, stage: str) -> None:
...
-            val_roi = BoundingBox(midx, roi.maxx, roi.miny, midy, roi.mint, roi.maxt)
+            val_roi = BoundingBox(midx, x.stop, y.start, midy, t.start, t.stop)
             self.val_sampler = GridGeoSampler(
                 self.dataset, self.patch_size, self.patch_size, val_roi
             )
         if stage in ['test']:
-            test_roi = BoundingBox(
-                roi.minx, roi.maxx, midy, roi.maxy, roi.mint, roi.maxt
-            )
+            test_roi = BoundingBox(midx, x.stop, midy, y.stop, t.start, t.stop)
```
adamjstewart, line 100: Same as #2834

PR #2806, opened 2025-05-29
EarthLoc Model

isaaccorley:
This PR adds the EarthLoc model from the CVPRW 2024 paper ["EarthLoc: Astronaut Photography Localization by Indexing Earth from Space", Berton et al.](https://arxiv.org/abs/2403.06758). The model is trained for image retrieval to be able to localize where a given non-georeferenced image is on the Earth using Sentinel-2 imagery.

The model's architecture is: RGB image input -> extract 2D embeddings using a ResNet-50 model -> compress 2D embeddings to a 1D embedding using a [MixVPR](https://arxiv.org/abs/2303.02190) MLP-Mixer encoder

cc: @gmberton @alexstoken

### Training Strategy
image

### Results
The authors compare 3 of our encoders for image retrieval and find that EarthLoc significantly outperforms them.

image


alexstoken:
This looks great @isaaccorley, happy to see EarthLoc added to the torchgeo zoo.

torchgeo/models/earthloc.py
```diff
@@ -0,0 +1,264 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# https://github.com/gmberton/EarthLoc
```
adamjstewart: Available under an MIT license, which is perfect for us. Note that if any of the below code was directly copied from the other repo, we need to add "Copyright (c) 2024 Gabriele Berton" to that code.

adamjstewart (commented):
I think we're still missing `hubconf.py` changes

torchgeo/models/earthloc.py
isaaccorley: Most of the code is modified but the `FeatureMixerLayer` and `MixVPR` is mostly copy/paste. Do you mean to add the copyright to the docstrings of each class or to the top of the file?

isaaccorley: Added to each class that I adapted from the original codebase.

adamjstewart: Both are okay with me, as long as the copied parts are under the right copyright.

adamjstewart (approved):

adamjstewart:
It's possible that this test requires internet access, we should fix that: https://github.com/torchgeo/torchgeo/actions/runs/20265343915/job/58187117566?pr=3203

adamjstewart:
Tests failing again due to internet access: https://github.com/torchgeo/torchgeo/actions/runs/20481836279/job/58856580402

PR #2812, opened 2025-06-04
GeoSampler: support arbitrary Polygon for roi, add toi

### LINKED ISSUE #1190: support for non bbox roi in sampler
### Summary

Add option to pass Polygon, MultiPolygon or oriented bounding box as the roi param to the geosampler. (and perhaps support other geometry types?)
The sampler would then sample bounding boxes that have their center fall somewhere inside the polygon / multipolygon, or in the case of the oriented bounding box, it would sample from patches from the OBB and use the rotation matrix to sample the actual image data.

### Rationale

I'm using a global coastal dataset. It is therefor made up of very narrow strips, oriented in many directions spread out over the entire globe. The minimum bounding box for this dataset is basically the whole world, but the actual coast line makes up a tiny fragment of that bounding box. So the chance that the geo sampler will sample a "valid" image for me is basically 0. 

ATM I have to manually split up my dataset many small bounding boxes and create a dataloader for each one. This is labour intensive and suboptimal because there isn't one clear way to cut the coastal region up into bboxes.

Instead it would be very convenient if I could simply pass the multi polygon to the geo sampler. I don't mind if the sampled images will have some pixels outside the roi as long as the center of the image is contained inside the roi.

### Implementation

use the bounding boxes of the polygons / oriented bounding box to sample points and then perform isContained() test for the points to see if they fall within the polygons / oriented bounding box

### Alternatives

_No response_

### Additional information

_No response_

adamjstewart:
Now that we are using geopandas, we can easily support any arbitrary shapely Polygon: #2382

* GeoSampler no longer supports BoundingBox roi, use shapely Polygon instead

Closes #1190 
Loosely related to #1170


torchgeo/datamodules/naip.py
```diff
@@ -87,21 +87,17 @@ def setup(self, stage: str) -> None:
...
-            val_roi = BoundingBox(midx, roi.maxx, roi.miny, midy, roi.mint, roi.maxt)
+            val_roi = shapely.box(midx, roi.miny, roi.maxx, midy)
             self.val_sampler = GridGeoSampler(
                 self.dataset, self.patch_size, self.patch_size, val_roi
             )
         if stage in ['test']:
-            test_roi = BoundingBox(
-                roi.minx, roi.maxx, midy, roi.maxy, roi.mint, roi.maxt
-            )
+            test_roi = shapely.box(midx, midy, roi.maxx, roi.maxy)
```
adamjstewart: Same as #2834

isaaccorley (approved):

adamjstewart:
We should, but there's infinitely many corner cases and we don't yet sample within the polygon bounds, so I got lazy.

isaaccorley:
It's okay I'd rather get it out and have users try to break it. I'm fine with it as is.

isaaccorley (approved):

PR #2819, opened 2025-06-06
Add support for bounding boxes and instance masks to VectorDataset

### LINKED ISSUE #2505: VectorDataset: return bounding boxes and instance segmentation masks

mayrajeo:
Closes #2505 , a proposal to add `bbox_xyxy`, `segmentation` and `label` as the return values for `VectorDataset.__getitem__`. After this, `VectorDataset.__getitem__` returns

```python
        sample = {
            'mask': masks, # As before
            'bbox_xyxy': boxes_xyxy, # N, 4 tensor containing the bounding boxes in xmin, ymin, xmax, ymax
            'segmentation': segmentations, # A list of N tensors containing the COCO format segmentations
            'crs': self.crs,
            'label': labels, # N tensor containing the int labels for each object
            'bounds': query,
        }
```

Major downside: as the object detection related things are returned by default, `VectorDataset` can't be used with most of the common `collate_fn`s, such as `stack_samples` if the batch size is larger than 1. 

Example gist: https://gist.github.com/mayrajeo/1de0497dd82d2c9a2b6381ef482face8

### DIFF
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -689,6 +689,9 @@ class VectorDataset(GeoDataset):
 
         .. versionchanged:: 0.5
            *root* was renamed to *paths*.
+
+        .. versionadded:: 0.8
+           The *task* and *layer* parameters
         """
         self.paths = paths
         self.transforms = transforms
@@ -847,18 +850,20 @@ class VectorDataset(GeoDataset):
                         [p.bounds[0], p.bounds[1], p.bounds[2], p.bounds[3]]
                         for p in px_shapes
                     ]
-                    # HxW mask for each instance
-                    masks = np.array(
-                        [
-                            rasterio.features.rasterize(
-                                [s],
-                                out_shape=(round(height), round(width)),
-                                transform=transform,
-                            )
-                            for s in shapes
-                        ]
+
+                    masks = rasterio.features.rasterize(
+                        [(s[0], i + 1) for i, s in enumerate(shapes)],
+                        out_shape=(round(height), round(width)),
+                        transform=transform,
                     )
 
+                    obj_ids = np.unique(masks)
+
+                    # first id is the background, so remove it
+                    obj_ids = obj_ids[1:]
+
+                    # convert (H, W) mask a set of binary masks
+                    masks = (masks == obj_ids[:, None, None]).astype(np.uint8)
         else:
             # If no features are found in this key, return an empty mask
             # with the default fill value and dtype used by rasterize
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -674,7 +674,7 @@ class VectorDataset(GeoDataset):
                 entry and returns a transformed version
             label_name: name of the dataset property that has the label to be
                 rasterized into the mask
-            task: what is the tas the dataset is used for. Supported output types
+            task: computer vision task the dataset is used for. Supported output types
                `object_detection`, `semantic_segmentation`, `instance_segmentation`
             layer: if the input is a multilayer vector dataset, such as a geopackage,
                 specify which layer to use. Can be int to specify the index of the layer,
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -14,7 +14,6 @@ from collections.abc import Callable, Iterable, Sequence
 from datetime import datetime
 from typing import Any, ClassVar, Literal
 
-import affine
 import fiona
 import fiona.transform
 import geopandas as gpd
@@ -40,6 +39,7 @@ from .utils import (
     Path,
     array_to_tensor,
     concat_samples,
+    convert_poly_coords,
     disambiguate_timestamp,
     merge_samples,
     path_is_vsi,
@@ -622,36 +622,6 @@ class RasterDataset(GeoDataset):
             return src
 
 
-def convert_poly_coords(
-    geom: shapely.geometry.shape, affine_obj: affine.Affine, inverse: bool = False
-) -> shapely.geometry.shape:
-    """Convert geocoordinates to pixel coordinates and vice versa, based on `affine_obj`.
-
-    Args:
-        geom: shapely.geometry.shape to convert
-        affine_obj: affine.Affine object to use for geoconversion
-        inverse: If true, convert geocoordinates to pixel coordinates
-
-    Returns:
-        shapely.geometry.shape: input shape converted to pixel coordinates
-    """
-    if inverse:
-        affine_obj = ~affine_obj
-
-    xformed_shape = shapely.affinity.affine_transform(
-        geom,
-        [
-            affine_obj.a,
-            affine_obj.b,
-            affine_obj.d,
-            affine_obj.e,
-            affine_obj.xoff,
-            affine_obj.yoff,
-        ],
-    )
-    return xformed_shape
-
-
 class VectorDataset(GeoDataset):
     """Abstract base class for :class:`GeoDataset` stored as vector files."""
 
@@ -691,7 +661,7 @@ class VectorDataset(GeoDataset):
         task: Literal[
             'object_detection', 'semantic_segmentation', 'instance_segmentation'
         ] = 'semantic_segmentation',
-        gpkg_layer: str | int | None = None,
+        layer: str | int | None = None,
     ) -> None:
         """Initialize a new VectorDataset instance.
 
@@ -706,9 +676,9 @@ class VectorDataset(GeoDataset):
                 rasterized into the mask
             task: what is the tas the dataset is used for. Supported output types
                `object_detection`, `semantic_segmentation`, `instance_segmentation`
-            gpkg_layer: if the input is a multilayer geopackage, specify which layer
-                to use. Can be int to specify the index of the layer, str to select
-                the layer with that name or None which opens the first layer
+            layer: if the input is a multilayer vector dataset, such as a geopackage,
+                specify which layer to use. Can be int to specify the index of the layer,
+                str to select the layer with that name or None which opens the first layer
 
         Raises:
             DatasetNotFoundError: If dataset is not found.
@@ -732,7 +702,7 @@ class VectorDataset(GeoDataset):
         if task not in allowed_tasks:
             raise ValueError(f'Invalid task: {task!r}. Must be one of {allowed_tasks}')
         self.task = task
-        self.gpkg_layer = gpkg_layer
+        self.layer = layer
         # Gather information about the dataset
         filename_regex = re.compile(self.filename_regex, re.VERBOSE)
         filepaths = []
@@ -742,7 +712,7 @@ class VectorDataset(GeoDataset):
             match = re.match(filename_regex, os.path.basename(filepath))
             if match is not None:
                 try:
-                    with fiona.open(filepath, layer=gpkg_layer) as src:
+                    with fiona.open(filepath, layer=layer) as src:
                         if crs is None:
                             crs = CRS.from_wkt(src.crs_wkt)
 
@@ -804,7 +774,7 @@ class VectorDataset(GeoDataset):
 
         shapes = []
         for filepath in index.filepath:
-            with fiona.open(filepath, layer=self.gpkg_layer) as src:
+            with fiona.open(filepath, layer=self.layer) as src:
                 # We need to know the bounding box of the query in the source CRS
                 (minx, maxx), (miny, maxy) = fiona.transform.transform(
                     self.crs.to_wkt(), src.crs, [x.start, x.stop], [y.start, y.stop]
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -876,7 +876,7 @@ class VectorDataset(GeoDataset):
             labels = np.empty((0,), dtype=np.int32)
 
         # Use array_to_tensor since rasterize may return uint16/uint32 arrays.
-        sample: dict[str, Tensor | CRS | GeoSlice] = {'crs': self.crs, 'bounds': query}
+        sample: dict[str, Any] = {'crs': self.crs, 'bounds': query}
 
         match self.task:
             case 'semantic_segmentation':
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -12,7 +12,7 @@ import re
 import warnings
 from collections.abc import Callable, Iterable, Sequence
 from datetime import datetime
-from typing import Any, ClassVar
+from typing import Any, ClassVar, Literal
 
 import affine
 import fiona
@@ -688,6 +688,9 @@ class VectorDataset(GeoDataset):
         res: float | tuple[float, float] = (0.0001, 0.0001),
         transforms: Callable[[dict[str, Any]], dict[str, Any]] | None = None,
         label_name: str | None = None,
+        task: Literal[
+            'object_detection', 'semantic_segmentation', 'instance_segmentation'
+        ] = 'semantic_segmentation',
         gpkg_layer: str | int | None = None,
     ) -> None:
         """Initialize a new VectorDataset instance.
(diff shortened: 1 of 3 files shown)


mayrajeo:

isaaccorley:
@adamjstewart Thoughts on adding an "output_type" arg to VectorDataset? My only concern is computing mask/instance mask/boxes for each polygon can add unwanted overhead in the case that I just want boxes, or I just want the mask.

adamjstewart:
I'm not completely opposed. In fact, this is one of my implementation questions in #2505.

I guess I would be curious to know exactly how much overhead we are talking about. As long as the overhead is noticeable (10% slower) and the change is minimal (a few lines of code, plus a new parameter in all subclasses) then let's do it. Only downside of a parameter is that the return type becomes even more dynamic, and it's difficult to statically type check.

Note that Mask R-CNN (our only instance segmentation model) requires both boxes and masks.

isaaccorley:
Yeah I was thinking output types would be [semantic, instance, boxes].

My concern is that instance masks have the ability to take up a ton of memory. 


- Semantic returns a mask of shape (num_classes, h, w)

- Boxes aren't memory intensive (num_objects, 4)

- for Instance we are turning boxes, and instance masks of shape (num_objects, h, w)

If num_objects is large (which is common for small object detection) this can take up quite a bit of memory in my experience, so I imagine we shouldn't return all output types and only the one the user wants.

adamjstewart:
I wasn't thinking about the difference between semantic and instance masks, that's a good point. Yeah, we should probably have a knob to control this then.

mayrajeo:
Should the `sample` keys be consistent so that `masks` is an empty list or a tensor if semantic segmentation masks are not desired, or depend on the flags specified? Maybe the latter, because it could be possible to add both oriented bounding boxes either as instance masks (`[x1, y1, x2, y2, x3, y3, x4, y4]`) or detectron2-style (`[x_center, y_center, w, h, rotation]`, -180 <= rotation < 180) annotation format, or keypoints if needed.

Something like 

```python

class VectorDataset(GeoDataset):
    ...
    def __init    def __init__(
        self,
        crs: CRS | None = None,
        res: float | tuple[float, float] = (0.0001, 0.0001),
        masks: bool = True,
        boxes: bool = False,
        instances: bool = False,
        label_name: str | None = None,
        gpkg_layer: str | int | None = None,
    ) -> None:
    ...
    self.masks = masks
    self.boxes = boxes
    self.instances = instances
    ...

    def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
         # Create things depending on self.masks, self.boxes and self.instances
         ...
        sample = {
            'crs': self.crs,
            'bounds': query,
        }
        if self.masks:
            sample['mask'] = masks
       # etc...
```
maybe?

isaaccorley:
I think it should depend on the flags specified although I don't think we should have a flag, it should be a list of output types that a user can specify.

For OBB I prefer the polygon approach like (x1, y1, ...,x4, y4) instead of using rotation angles just because there's several definitions of the rotation angle and keeping it in polygon format makes easy conversion to<->from shapely/geopandas.

adamjstewart:
The sample keys aren't up to us, they need to follow the format expected by Kornia. Note that Kornia does not yet support OBB, so we need to add this to Kornia first.

mayrajeo:
For the output types, would `output_type: List[str] = ['masks']` with possible keys being `['masks', 'boxes', 'instances']` be good? Though `masks` and `instances` are mutually exclusive, as kornia seems to expect `masks` key for both semantic and instance masks correct? Another option might be to specify `task`, which is one of `segmentation`, `object_detection`, `instance_segmentation` or something like that, since there isn't many cases where only instance masks are needed right?.

> Note that Kornia does not yet support OBB, so we need to add this to Kornia first.

They can be thought as polygons, so they need to be converted into binary masks. Getting them from polygons with shapely is easy (`oriented_envelope`) but they can be omitted for now.

adamjstewart:
See https://github.com/kornia/kornia/blob/v0.8.1/kornia/constants.py#L142 for a list of valid keys. Our TorchGeo object detection trainer is setup to look for bbox_xyxy and label, and our instance segmentation trainer is setup to look for bbox_xyxy, label, and mask. So we should use these keys in the return dict. I'm leaning towards:
```python
task: Literal['object_detection', 'semantic_segmentation', 'instance_segmentation'] = 'semantic_segmentation'`
```
for the `__init__` parameter and default, although those names are quite long.

isaaccorley:
> See [kornia/kornia@`v0.8.1`/kornia/constants.py#L142](https://github.com/kornia/kornia/blob/v0.8.1/kornia/constants.py?rgh-link-date=2025-06-15T10%3A44%3A47Z#L142) for a list of valid keys. Our TorchGeo object detection trainer is setup to look for bbox_xyxy and label, and our instance segmentation trainer is setup to look for bbox_xyxy, label, and mask. So we should use these keys in the return dict. I'm leaning towards:
> 
> ```python
> task: Literal['object_detection', 'semantic_segmentation', 'instance_segmentation'] = 'semantic_segmentation'`
> ```
> 
> for the `__init__` parameter and default, although those names are quite long.

They are long but it's likely better to be explicit here tbh. In the future we can add support for multiple output types but let's keep it simple for now. A user can always make separate datasets with different output types and join them.

mayrajeo:
Added tasks, examples here: https://gist.github.com/mayrajeo/596c43c0a0c815e6fc09c1f8e20136cd

torchgeo/datasets/geo.py
```diff
@@ -587,6 +588,36 @@ def _load_warp_file(self, filepath: Path) -> DatasetReader:
             return src
 
 
+def convert_poly_coords(
```
isaaccorley: Let's move this to `torchgeo/datasets/utils.py`

torchgeo/datasets/geo.py
```diff
@@ -635,9 +670,15 @@ def __init__(
                 entry and returns a transformed version
             label_name: name of the dataset property that has the label to be
                 rasterized into the mask
+            task: what is the tas the dataset is used for. Supported output types
+               `object_detection`, `semantic_segmentation`, `instance_segmentation`
+            gpkg_layer: if the input is a multilayer geopackage, specify which layer
```
isaaccorley: I believe this can work with files other than just gpkg so let's just call this arg "layer"

torchgeo/datasets/geo.py
```diff
@@ -743,19 +793,104 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
...
+                    # Get labels
+                    labels = [s[1] for s in shapes]
+
+                    # xmin, ymin, xmax, ymax format
+                    boxes_xyxy = [
+                        [p.bounds[0], p.bounds[1], p.bounds[2], p.bounds[3]]
+                        for p in px_shapes
+                    ]
+                    # HxW mask for each instance
+                    masks = np.array(
```
isaaccorley: I'm okay with leaving this as is for now but it may be faster to not loop over the shapes and simply create the "values" for each mask to be the range(1, len(shapes)+1) so that it creates a (H, W) mask where each instance has a unique int id.

Then we can convert this to a (N, H, W) mask using the following torch code.

```
obj_ids = torch.unique(mask)

# first id is the background, so remove it
obj_ids = obj_ids[1:]
num_objs = len(obj_ids)

# convert (H, W) mask a set of binary masks
masks = (mask == obj_ids[:, None, None]).to(dtype=torch.uint8)
```

isaaccorley:
Great work! Just a few minor comments but we are very close to being done here 🚀

robmarkcole:
Good to see this. Probably covered elsewhere, but be sure to test images with no annotated boxes/masks - Kornia throws errors if empty `masks` are returned. In my own implementation I have an arg to control the % of empty images returned.

Also I can anticipate training instance seg model with an object detection annotated dataset - I did this with airbus ships and the model still learned to contour the ships mask quite well. Be nice to have this option, perhaps off by default.

Another thing I learned, when random cropping an image, it is useful to use the `segmentation` to determine the object extent as the box wont always capture the extent of the object in the crop. So an accurate `segmentation` mask is very desired even for object detection

torchgeo/datasets/geo.py
mayrajeo: That might work, but how to handle different classes with this approach?

isaaccorley: The classes are already defined in the "labels" variable that is created before the mask

mayrajeo: Ah, true. I'll check how this would work and how fast.

adamjstewart:
This looks fantastic now! I think we've settled on a good API. Just need to polish and add tests.

torchgeo/datasets/geo.py
```diff
@@ -669,9 +674,15 @@ def __init__(
...
                 rasterized into the mask
+            task: what is the tas the dataset is used for. Supported output types
+               `object_detection`, `semantic_segmentation`, `instance_segmentation`
+            layer: if the input is a multilayer vector dataset, such as a geopackage,
+                specify which layer to use. Can be int to specify the index of the layer,
+                str to select the layer with that name or None which opens the first layer
 
         Raises:
             DatasetNotFoundError: If dataset is not found.
+            ValueError: If task is not one of allowed tasks
```
adamjstewart, line 686:
```rst
.. versionadded:: 0.8
   The *task* and *layer* parameters.
```

torchgeo/datasets/utils.py
```diff
@@ -722,3 +724,33 @@ def which(name: Path) -> Executable:
...
+) -> shapely.geometry.shape:
+    """Convert geocoordinates to pixel coordinates and vice versa, based on `affine_obj`.
+
+    Args:
+        geom: shapely.geometry.shape to convert
+        affine_obj: affine.Affine object to use for geoconversion
+        inverse: If true, convert geocoordinates to pixel coordinates
+
+    Returns:
+        shapely.geometry.shape: input shape converted to pixel coordinates
```
adamjstewart: No need to include types in docstrings, Sphinx automatically extracts those from the type hints.

Also add:
```rst
.. versionadded:: 0.8
```

torchgeo/datasets/geo.py
```diff
@@ -669,9 +674,15 @@ def __init__(
                 entry and returns a transformed version
             label_name: name of the dataset property that has the label to be
                 rasterized into the mask
+            task: what is the tas the dataset is used for. Supported output types
```
```suggestion
            task: computer vision task the dataset is used for. Supported output types
```

mayrajeo:
Added some tests for checking that `VectorDataSet` works as its supposed to. It's still open how much we need to test e.g. Kornia and empty masks. Existing datasets should work as before, as default task is `semantic_segmentation` which works exactly as before. 

As @robmarkcole suggested, it is possible to add a parameter for determining how small fragments to keep after sampling, such as minimum pixel area or minimum pct of original polygon or bounding box. Question is, which one? Also, using `task='instance_segmentation'` should work with object detection workflows, as it returns also the `bbox_xyxy`. Separate task for `object_detection` is pretty much only useful for not generating the masks when they are not needed. With the current implementation, if the input shapefile has only boxes and `task='instance_segmentation'`, then the output masks are rasterized bounding boxes.

robmarkcole:
Of those 2 options, min pixel count could be better, otherwise a 4 pixel annotation could end up cropped to a single pixel

mayrajeo:
Another possible challenge might be MultiPolygons. In cases of multiple non-overlapping polygons, they can be separated with something like

```python
def separate_multipolygons(shapes_list:list) -> list:
    """Separate multipolygons in `shapes_list` to individual polygons 
    and duplicate their labels. 

    Args:
        shapes_list: list of tuples, where each tuple is (shape, label)

    Returns:
        A list of (shape, label) tuples, with multipolygons removed
    
    .. versionadded:: 0.8
    """
    out_shapes = []

    for s in shapes_list:
        feat, label = s
        if feat['geometry']['type'] == 'MultiPolygon':
            for coords in feat['geometry']['coordinates']:
                new_feat = deepcopy(feat)
                new_feat['geometry'] = {'type': 'Polygon', 'coordinates': coords}
                out_shapes.append((new_feat, label))
        else:
            out_shapes.append(s)
    return out_shapes
```

and 

```python
...
case 'object_detection': # or case 'instance_segmentation'
    # Explode possible multipolygons to individual targets
    shapes = separate_multipolygons(shapes)

    # Get boxes for object detection or instance segmentation
    px_shapes = [
        convert_poly_coords(
            shapely.geometry.shape(s[0]), transform, inverse=True
        )
        for s in shapes
    ]
...
```
But polygons with holes are a completely different case.

adamjstewart:
My preference would be to keep things simple for now and add complexity only when we encounter a dataset that needs it. This also makes testing easier.

Just need to resolve remaining mypy errors and test coverage. Let us know if you need help with this.

mayrajeo:
I'm on vacation and away from computer for the next month or so, so if someone can resolve the mypy errors it would be great.

isaaccorley:
Some changes I made:
- updated tests and coverage
- forced labels to be `int32` and bbox_xyxy to be `float32`. I can already see a user getting a phantom error because they set the dataset dtype to uint8 and their boxes are always getting clipped to [0,255] so this fixes that

torchgeo/datasets/geo.py
```diff
@@ -692,7 +715,7 @@ def __init__(
             match = re.match(filename_regex, os.path.basename(filepath))
             if match is not None:
                 try:
-                    with fiona.open(filepath) as src:
+                    with fiona.open(filepath, layer=layer) as src:
```
adamjstewart, line 718: @isaaccorley you were talking about replacing fiona with geopandas in a future PR. Does geopandas have a similar feature? This affects the parameter name choice.

torchgeo/datasets/geo.py
```diff
@@ -776,19 +799,97 @@ def __getitem__(self, query: GeoSlice) -> dict[str, Any]:
...
         else:
             # If no features are found in this key, return an empty mask
             # with the default fill value and dtype used by rasterize
             masks = np.zeros((round(height), round(width)), dtype=np.uint8)
+            boxes_xyxy = np.empty((0, 4), dtype=np.float32)
+            labels = np.empty((0,), dtype=np.int32)
 
         # Use array_to_tensor since rasterize may return uint16/uint32 arrays.
-        masks = array_to_tensor(masks)
+        sample: dict[str, Tensor | CRS | GeoSlice] = {'crs': self.crs, 'bounds': query}
```
```suggestion
        sample: dict[str, Any] = {'crs': self.crs, 'bounds': query}
```
At some point I plan to use TypedDict but not now

torchgeo/datasets/utils.py
```diff
@@ -17,9 +17,11 @@
 from datetime import datetime, timedelta
 from typing import Any, TypeAlias, cast, overload
 
+import affine
```
adamjstewart: This is going to sound stupid, but can we import `rasterio.affine.Affine` instead? Rasterio vendors affine, so we don't officially need to add an additional dependency on affine.

adamjstewart (changes requested):
Looks good, just a couple minor comments.

torchgeo/datasets/utils.py
isaaccorley: I agree with this actually, I'll change it

torchgeo/datasets/geo.py
isaaccorley: Yes it's the same parameter name

`geopandas.read_file('file.gpkg', layer='layer_name')
`

adamjstewart:
Closing/reopening to try to bump the CLA bot...

PR #2823, opened 2025-06-07
Aurora Atmospheric/Weather Forecasting Models

isaaccorley:
Adding the Aurora Atmospheric/Weather Forecasting models from https://github.com/microsoft/aurora!

See their [docs](https://microsoft.github.io/aurora/intro.html) and [paper](https://arxiv.org/abs/2405.13063) for more info!

cc: @wesselb @crisbodnar @megstanley

### DIFF
--- a/docs/api/weights/climate.csv
+++ b/docs/api/weights/climate.csv
@@ -1,8 +1,0 @@
-Weight,Sensor,Source,Citation,License
-Aurora_Weights.HRES_T0_PRETRAINED_AURORA,HRES-T0,`link <https://github.com/microsoft/aurora>`__,`link <https://arxiv.org/abs/2405.13063>`__,"MIT"
-Aurora_Weights.HRES_T0_PRETRAINED_12HR_AURORA,HRES-T0,`link <https://github.com/microsoft/aurora>`__,`link <https://arxiv.org/abs/2405.13063>`__,"MIT"
-Aurora_Weights.HRES_T0_PRETRAINED_SMALL_AURORA,HRES-T0,`link <https://github.com/microsoft/aurora>`__,`link <https://arxiv.org/abs/2405.13063>`__,"MIT"
-Aurora_Weights.HRES_T0_AURORA,HRES-T0,`link <https://github.com/microsoft/aurora>`__,`link <https://arxiv.org/abs/2405.13063>`__,"MIT"
-Aurora_Weights.HRES_T0_HIGH_RES_AURORA,HRES-T0,`link <https://github.com/microsoft/aurora>`__,`link <https://arxiv.org/abs/2405.13063>`__,"MIT"
-Aurora_Weights.HRES_CAMS_AIR_POLLUTION_AURORA,HRES-CAMS,`link <https://github.com/microsoft/aurora>`__,`link <https://arxiv.org/abs/2405.13063>`__,"MIT"
-Aurora_Weights.HRES_WAM0_WAVE_AURORA,HRES-WAM0,`link <https://github.com/microsoft/aurora>`__,`link <https://arxiv.org/abs/2405.13063>`__,"MIT"
--- a/docs/api/weights/climate.csv
+++ b/docs/api/weights/climate.csv
@@ -1,0 +1,8 @@
+Weight,Sensor,Source,Citation,License
+Aurora_Weights.HRES_T0_PRETRAINED_AURORA,HRES-T0,`link <https://github.com/microsoft/aurora>`__,`link <https://arxiv.org/abs/2405.13063>`__,"MIT"
+Aurora_Weights.HRES_T0_PRETRAINED_12HR_AURORA,HRES-T0,`link <https://github.com/microsoft/aurora>`__,`link <https://arxiv.org/abs/2405.13063>`__,"MIT"
+Aurora_Weights.HRES_T0_PRETRAINED_SMALL_AURORA,HRES-T0,`link <https://github.com/microsoft/aurora>`__,`link <https://arxiv.org/abs/2405.13063>`__,"MIT"
+Aurora_Weights.HRES_T0_AURORA,HRES-T0,`link <https://github.com/microsoft/aurora>`__,`link <https://arxiv.org/abs/2405.13063>`__,"MIT"
+Aurora_Weights.HRES_T0_HIGH_RES_AURORA,HRES-T0,`link <https://github.com/microsoft/aurora>`__,`link <https://arxiv.org/abs/2405.13063>`__,"MIT"
+Aurora_Weights.HRES_CAMS_AIR_POLLUTION_AURORA,HRES-CAMS,`link <https://github.com/microsoft/aurora>`__,`link <https://arxiv.org/abs/2405.13063>`__,"MIT"
+Aurora_Weights.HRES_WAM0_WAVE_AURORA,HRES-WAM0,`link <https://github.com/microsoft/aurora>`__,`link <https://arxiv.org/abs/2405.13063>`__,"MIT"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -119,7 +119,7 @@ docs = [
     "sphinx>=4,<6",
 ]
 models = [
-    # microsoft-aurora 1.7+ required timm 1.0+ support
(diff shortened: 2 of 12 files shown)


torchgeo/models/aurora.py
```diff
@@ -0,0 +1,116 @@
...
+        An Aurora model.
+    """
+    lazy_import('aurora')
+    import aurora
+
+    if weights is None:
+        model = aurora.Aurora(*args, **kwargs)
+    else:
+        model = getattr(aurora, weights.meta['model'])(*args, **kwargs)
+        model.load_checkpoint()
```
isaaccorley: This still needs some work because the aurora package automatically downloads the checkpoint from huggingface however it downloads from main and we want to download from a commit hash.

torchgeo/models/aurora.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+
+    HRES_WAM0_WAVE_AURORA = Weights(
+        url='https://huggingface.co/microsoft/aurora/resolve/74598e8c65d53a96077c08bb91acdfa5525340c9/aurora-0.25-wave.ckpt',
+        transforms=_aurora_transforms,
+        meta=_aurora_meta | {'model': 'AuroraWave', 'resolution': 0.25},
+    )
+
+
+def aurora_swin_unet(
```
isaaccorley, line 226: Not sure how to name this because the package is named `aurora` so I can't also name the model `aurora`. Any suggestions on this @adamjstewart?

torchgeo/models/aurora.py
```diff
@@ -0,0 +1,116 @@
...
+
+_aurora_meta = {
+    'dataset': 'Aurora',
+    'model': None,
+    'resolution': None,
+    'architecture': '3D Swin Transformer U-Net',
+    'encoder': '3D Perceiver',
+    'publication': 'https://arxiv.org/abs/2409.16252',
+    'repo': 'https://github.com/microsoft/aurora',
+    'license': 'MIT',
```
isaaccorley, line 30: @wesselb can you comment on the license for the weights? I just pulled this from the microsoft/aurora repo license but not sure if the weights have a different license.

tests/models/test_aurora.py
```diff
@@ -0,0 +1,57 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
isaaccorley, line 1: Don't review this yet because this entire script needs to be revamped to load dummy test data as xarray.

pyproject.toml
```diff
@@ -106,6 +106,9 @@ datasets = [
     # xarray 0.12.3+ required for pandas 1.3.3 support
     "xarray>=0.12.3",
 ]
+models = [
```
isaaccorley: @adamjstewart I added this new extras group but not sure if I'm missing something else in the repo that I need to update as well.

adamjstewart:
Last I knew, Aurora only supports medium-range **weather forecasting** (2 hrs–2 wks), not **climate modeling** (decadal scale). Would be good to confirm with the authors. We need to be _very_ careful with naming here, as there is a huge distinction between data-driven stateful models and physics-driven statistical distribution models.

isaaccorley:
> Last I knew, Aurora only supports medium-range **weather forecasting** (2 hrs–2 wks), not **climate modeling** (decadal scale). Would be good to confirm with the authors. We need to be _very_ careful with naming here, as there is a huge distinction between data-driven stateful models and physics-driven statistical distribution models.

You're correct, the paper specifically says "forecasting of weather and atmospheric processes". I'll make sure to not use the term "climate modeling".

torchgeo/models/aurora.py
wesselb: As of recently, both the code and the weights are under the MIT licence, so this looks good!

wesselb:
@adamjstewart, @isaaccorley, you're correct that Aurora only does medium-term forecasting (up to 10 days in 6 hour steps). Note, however, that it doesn't just do weather forecasting: we have models for ocean waves and air pollution too.

adamjstewart:
@cnagda can you review?

docs/api/weights/climate.csv
adamjstewart, line 1: Filename: atmospheric.csv?

pyproject.toml
```diff
@@ -106,6 +106,9 @@ datasets = [
...
+    "microsoft-aurora"
```
adamjstewart: Min version

pyproject.toml
```diff
@@ -133,7 +136,7 @@ tests = [
     "pytest-cov>=4",
 ]
 all = [
-    "torchgeo[datasets,docs,style,tests]",
+    "torchgeo[datasets,models,docs,style,tests]",
```
adamjstewart: Alphabetical

docs/api/weights/climate.csv
adamjstewart: Add column for nowcasting vs. medium-range weather forecasting vs. S2S vs. decadal-scale climate. Could use a key under the table and keys like N/MWF/S2S/C.

pyproject.toml
adamjstewart:
```suggestion
    "microsoft-aurora>=1.6"
```

adamjstewart: Also need to add to `requirements/min-reqs.old` and `requirements/models.txt`

isaaccorley:
> Looks good to me. For `load_checkpoint` , can we override to use commit hash / have our own version of it in this file?

Planning to upstream this fix directly to the aurora repo https://github.com/microsoft/aurora/issues/114

docs/api/weights/climate.csv
isaaccorley: I'm actually not 100% sure but I think these are all going to be nowcasting right since they are only forecasting less than a day out in hours? but you can do it autoregressively so technically you can also do MWF

isaaccorley: Also what do you think the column should be named? `Task` maybe?

requirements/min-reqs.old
```diff
@@ -34,6 +34,9 @@ scipy==1.9.2
 webdataset==0.2.4
 xarray==0.12.3
 
+# models
+microsoft-aurora>=1.6
```
adamjstewart:
```suggestion
microsoft-aurora==1.6.0
```

requirements/models.txt
```diff
@@ -0,0 +1 @@
+microsoft-aurora>=1.6
```
adamjstewart:
```suggestion
microsoft-aurora==1.7.0
```

docs/api/weights/climate.csv
adamjstewart: Yes, they're designed for up to 10 days at 6 hr increments, so definitely MWF. Task seems fine.

adamjstewart:
Can you rebase this? Should be much simpler now that [models] was added in a different PR.

pyproject.toml
```diff
@@ -119,6 +119,8 @@ docs = [
     "sphinx>=4,<6",
 ]
 models = [
+    # microsoft-aurora 1.7+ required timm 1.0+ support
```
```suggestion
    # microsoft-aurora 1.7+ required for timm 1.0+ support
```

torchgeo/models/aurora.py
```diff
@@ -0,0 +1,116 @@
...
+    Args:
+        weights: Pre-trained model weights to use.
+        *args: Additional arguments to pass to ``aurora.Aurora``
+        **kwargs: Additional keyword arguments to pass to ``aurora.Aurora``
+
+    Returns:
+        An Aurora model.
+    """
+    lazy_import('aurora')
+    import aurora
```
adamjstewart:
```suggestion
    aurora = lazy_import('aurora')
```

adamjstewart:
Failing minimum tests due to https://github.com/huggingface/safetensors/issues/641, for now maybe we can pin to older safetensors.

isaaccorley:
Can we support older aurora? The aurora 1.6.0 version pins to a specific version of timm<1.0 which conflicts with our min version of timm being >1.0

adamjstewart:
I believe aurora 1.6 includes your PR

adamjstewart:
@isaaccorley I'll let you take over again. Now that the tests are actually running, there are a bunch of new errors.

PR #2825, opened 2025-06-08
Fix empty validation set for BigEarthNetV2 (#2824)

### LINKED ISSUE #2824: BigEarthNetV2: Empty Validation Set
### Description

The problem is:

In the metadata.parquet file for the refined BigEarthNet (=BigEarthNetV2) the split column only holds the values "train", "validation" and "test". The correct key to pass to the c-tor would accordingly be split="validation". However, the c-tor asserts split to be in ["train", "val", "test"] so one cannot use the validation  set.

For additional information on the content of the metadata.parquet refere to: https://bigearth.net/ (Dataset Description Document, page 9) 

### Steps to reproduce

Run the following code:

```
from torchgeo.datasets.bigearthnet import BigEarthNetV2

val_set = BigEarthNetV2(root=DATASET_DIR, bands="s2", split="val")

print(len(val_set))
´´´

This will print: 0

### Version

0.7

Night-dev1307:
In the metadata.parquet the split col only holds values of "train", "validation" and "test". Using split="val" resulted in an empty dataset. Therefore, update val -> validation.

Closes #2824

### DIFF
--- a/requirements/required.txt
+++ b/requirements/required.txt
@@ -6,7 +6,7 @@ einops==0.8.1
 fiona==1.10.1
 geopandas==1.0.1
 kornia==0.8.1
-lightly==1.5.20
+lightly==1.5.21
 lightning[pytorch-extra]==2.5.0.post0
 matplotlib==3.10.1
 numpy==2.2.4


MariusGehlhaus:
> @Night-dev1307 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

isaaccorley:
Thanks for the PR @Night-dev1307!

I don't think we want to change the split from val to validation because all of our datasets expect splits to be train/val/test.

However we could add a line to change the filter on `self.metadata_df` to be `"validation" if split == "val"`. Does this make sense?

MariusGehlhaus:
> I don't think we want to change the split from val to validation because all of our datasets expect splits to be train/val/test.

Thanks for your quick response!

Makes totally sense to keep the split values standardized among datasets.

I decided to add an internal mapping (see updated commit).

adamjstewart:
> all of our datasets expect splits to be train/val/test.

Many do, I wouldn't say "all". I'm personally fine with both solutions.

tests/data/bigearthnet/v2/metadata.parquet
adamjstewart, line 1: Can you update `tests/data/bigearthnet/v2/data.py` so that this can be recreated?

adamjstewart: Or is this change no longer necessary?

MariusGehlhaus: Thanks for catching that. I've added it just now.

isaaccorley:
> > all of our datasets expect splits to be train/val/test.
> 
> 
> 
> Many do, I wouldn't say "all". I'm personally fine with both solutions.

To clarify, my point is that datasets that have a val set will use the "val" split name, not "validation" so we should stay consistent with this. Not that literally every dataset has train/val/test splits.

adamjstewart:
I understand your point, but there are datasets like So2Sat/MapInWild that use train/validation/test. We aren't exactly consistent, although I agree that I would like to be.

isaaccorley:
Ahh interesting, I forgot about these. I think we shouldn't do this since it's confusing to the user. Shall I make a PR to fix?

adamjstewart:
Sure. Maybe we can support both so as not to break backwards compatibility?

nilsleh (approved):

PR #2826, opened 2025-06-10
[WIP] Adding a sinusoidal temporal embedding

yichiac:
Adding a generic cyclic/periodic embedding where the user inputs the period length and gets back sin/cos. #2382


torchgeo/transforms/temporal.py
```diff
@@ -68,3 +71,29 @@ def compute_transformation(
         """
         out = self.identity_matrix(input)
         return out
+
+
+class TemporalEmbedding(nn.Module):
```
adamjstewart: Check what other libraries call this, maybe:

Prefix:

* Periodic
* Cyclic
* Sinusoidal

Suffix:

* Embedding
* Encoder

torchgeo/transforms/temporal.py
```diff
@@ -68,3 +71,29 @@ def compute_transformation(
...
+    """Generic sinusoidal embedding for periodic temporal features."""
+
+    def __init__(self, period: int) -> None:
```
adamjstewart:
```suggestion
    def __init__(self, period: Timedelta) -> None:
```

torchgeo/transforms/temporal.py
```diff
@@ -68,3 +71,29 @@ def compute_transformation(
...
+    def forward(self, t: Tensor) -> Tensor:
+        """Compute sinusoidal embeddings for periodic time values.
+
+        Args:
+            t: Tensor of shape (B,) or (B, 1), representing time values.
+
+        Returns:
+            Tensor of shape (B, 2), sin and cos embeddings.
+        """
+        t = t.view(-1, 1).float()
```
adamjstewart: Don't need to rearrange?

Convert t from float to pd.Timestamp before division, then convert back to float after division?

adamjstewart (changes requested):

torchgeo/transforms/temporal.py
```diff
@@ -68,3 +71,29 @@ def compute_transformation(
...
+            period: The period of the sinusoidal function.
+        """
+        super().__init__()
+        self.period = period
+
+    def forward(self, t: Tensor) -> Tensor:
+        """Compute sinusoidal embeddings for periodic time values.
+
+        Args:
+            t: Tensor of shape (B,) or (B, 1), representing time values.
```
adamjstewart: I wonder if this transform would only be useful if it took in a sample as input and then used the `time` key (or a user-defined key). Currently this transform requires custom handling.

yichiac: The transform now takes a sample as input with the `time` key. I can write tests for the transform once if it looks good.

torchgeo/transforms/temporal.py
```diff
@@ -68,3 +72,42 @@ def compute_transformation(
...
+
+
+class CyclicalEncoder(nn.Module):
+    """Generic sinusoidal embedding for periodic temporal features."""
+
+    def __init__(
+        self,
+        period: pd.Timedelta,
+        time_key: str = 'time',
+        out_key: str = 'time_embedding',
```
adamjstewart: Let's split the output into two keys. May not even need an `out_key` parameter. Could just use `f"sin_{time_key}"` and `f"cos_{time_key}"`.

tests/transforms/test_temporal.py
```diff
@@ -41,3 +41,22 @@ def test_rearrange_integration_in_augmentation_sequential() -> None:
     out_batch = train_transforms(batch)
     assert 'image' in out_batch
     assert out_batch['image'].shape == (b, t * c, h, w)
+
+
+def test_temporal_embedding_shape_and_values() -> None:
```
adamjstewart, line 46: Let's also test the values themselves. They should be bounded between [-1, 1]. It may be useful to check `math.isclose(output, expected)` for a real number to make sure we do the `pd.Timestamp <-> float` conversion correctly, as I noticed that it isn't exactly intuitive and can be bug prone.

adamjstewart (changes requested):

torchgeo/transforms/temporal.py
```diff
@@ -68,3 +72,42 @@ def compute_transformation(
         """
         out = self.identity_matrix(input)
         return out
+
+
+class CyclicalEncoder(nn.Module):
+    """Generic sinusoidal embedding for periodic temporal features."""
```
adamjstewart:
```rst
.. versionadded:: 0.8
```

adamjstewart:
Let's make this even more generic. The same kind of encoder should work for lat/lon, not just time.

calebrob6:
Is this still a WIP?

yichiac:
> Is this still a WIP?

No. Happy to push forward if we have more concrete plan for this PR.

PR #2828, opened 2025-06-12
New SMP Decoders

isaaccorley:
This PR adds the newly supported SMP decoders (UPerNet, Segformer, DPT) for all dense output tasks (semantic seg, pixelwise regression, change detection).

In research papers I've been seeing a trend towards using UPerNet and DPT decoders to evaluate geospatial foundation models (particularly transformer-based ones) so we should enable users to do this within our trainers.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,8 +69,8 @@ dependencies = [
     "segmentation-models-pytorch>=0.5",
     # shapely 2+ required for shapely.Geometry
     "shapely>=2",
-    # timm 0.8+ required for timm.models.adapt_input_conv, 0.9.2 required by SMP
-    "timm>=0.9.2",
+    # timm 1.0.3+ required for forward_intermediates support for ViT
+    "timm>=1.0.3",
     # torch 2+ required for Python 3.11 wheels
     "torch>=2",
     # torchmetrics 1.2+ required for average argument in mAP metric
--- a/tests/conf/inria_dpt.yaml
+++ b/tests/conf/inria_dpt.yaml
@@ -1,0 +1,14 @@
+model:
+  class_path: PixelwiseRegressionTask
+  init_args:
+    model: 'dpt'
+    backbone: 'vit_base_patch16_224'
+    in_channels: 3
+    loss: 'mae'
+data:
+  class_path: InriaAerialImageLabelingDataModule
+  init_args:
+    batch_size: 1
+    patch_size: 2
+  dict_kwargs:
+    root: 'tests/data/inria'
--- a/tests/trainers/test_change.py
+++ b/tests/trainers/test_change.py
@@ -179,7 +179,7 @@ class TestChangeDetectionTask:
         ],
         backbone: str,
     ) -> None:
-        if not (backbone == 'tu-vit_base_patch16_224' and model_name == 'dpt'):
+        if model_name == 'dpt' and not backbone.startswith('tu-vit'):
             pytest.skip('dpt model only supports vit backbones')
 
         model = ChangeDetectionTask(
--- a/tests/trainers/test_change.py
+++ b/tests/trainers/test_change.py
@@ -167,8 +167,21 @@ class TestChangeDetectionTask:
         'backbone', ['resnet18', 'mobilenet_v2', 'efficientnet-b0']
     )
     def test_freeze_backbone(
-        self, model_name: Literal['unet', 'fcsiamdiff', 'fcsiamconc'], backbone: str
+        self,
+        model_name: Literal[
+            'unet',
+            'deeplabv3+',
+            'segformer',
+            'upernet',
+            'dpt',
+            'fcsiamdiff',
+            'fcsiamconc',
+        ],
+        backbone: str,
     ) -> None:
+        if not (backbone == 'tu-vit_base_patch16_224' and model_name == 'dpt'):
+            pytest.skip('dpt model only supports vit backbones')
+
         model = ChangeDetectionTask(
             model=model_name, backbone=backbone, freeze_backbone=True
         )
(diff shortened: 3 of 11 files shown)


torchgeo/trainers/regression.py
adamjstewart, line 1: Can we also add `Literal[...]` to `__init__` here so the possible inputs are more clear?

tests/trainers/test_regression.py
```diff
@@ -320,11 +333,21 @@ def test_weight_str_download(self, weights: WeightsEnum) -> None:
...
+        'backbone',
+        ['resnet18', 'mobilenet_v2', 'efficientnet-b0', 'tu-vit_base_patch16_224'],
     )
     def test_freeze_backbone(self, model_name: str, backbone: str) -> None:
+        if backbone == 'tu-vit_base_patch16_224':
+            if model_name != 'dpt':
+                return
+        else:
+            if model_name == 'dpt':
+                return
```
adamjstewart: Wondering if there isn't a smarter way to do this in a single line...

torchgeo/trainers/regression.py
```diff
@@ -281,33 +281,58 @@ def configure_models(self) -> None:
         """Initialize the model."""
         weights = self.weights
 
-        if self.hparams['model'] == 'unet':
+        model = self.hparams['model']
+        backbone = self.hparams['backbone']
+        in_channels = self.hparams['in_channels']
+
+        if model == 'unet':
```
adamjstewart: Can we switch to `match`?

adamjstewart (changes requested):
Needs more test coverage

torchgeo/trainers/regression.py
isaaccorley: So I tried this but the way the code is written -- `model` is also used for plain `RegressionTask` which can take in any possible timm backbone. So unless we rewrite and break some things we can't use `Literal` here.

pyproject.toml
```diff
@@ -69,8 +69,8 @@ dependencies = [
     "segmentation-models-pytorch>=0.5",
     # shapely 2+ required for shapely.Geometry
     "shapely>=2",
-    # timm 0.8+ required for timm.models.adapt_input_conv, 0.9.2 required by SMP
-    "timm>=0.9.2",
+    # timm 1.0.3+ required for forward_intermediates support for ViT
```
adamjstewart, line 72: What is `forward_intermediates`? I don't actually see where we're using it. Does one of the SMP decoders require this?

torchgeo/trainers/change.py
```diff
@@ -113,6 +124,40 @@ def configure_models(self) -> None:
...
                 )
+            case 'deeplabv3+':
+                self.model = smp.DeepLabV3Plus(
+                    encoder_name=backbone,
+                    encoder_weights='imagenet' if weights is True else None,
+                    in_channels=in_channels * 2,  # images are concatenated
+                    classes=1,
+                )
+            case 'fcn':
+                self.model = FCN(
```
adamjstewart, line 135: I think FCN is the only thing we're still missing test coverage of?

adamjstewart (commented):
Just need 1 more line of test coverage and this is ready to go

pyproject.toml
isaaccorley: Yes forward_intermediates is needed for the ViT encoders to extra intermediate feature maps to feed to decoders like DPT

isaaccorley: It wasn't available until timm 1.0.3. It was one of the major features of timm v1. https://github.com/huggingface/pytorch-image-models/releases/tag/v1.0.3

torchgeo/trainers/change.py
isaaccorley: just added a test for this

tests/conf/inria_dpt.yaml
```diff
@@ -0,0 +1,14 @@
+model:
+  class_path: PixelwiseRegressionTask
```
adamjstewart, line 2: We should really add a datamodule for a pixelwise regression dataset, it's kinda silly that we're using INRIA to test this...

(but this is unrelated to this PR, we can add this to the ever expanding TODO list for future PRs)

tests/trainers/test_change.py
```diff
@@ -167,7 +167,11 @@ def test_weight_str_download(self, weights: WeightsEnum) -> None:
         'backbone', ['resnet18', 'mobilenet_v2', 'efficientnet-b0']
```
adamjstewart, line 167: Does this need to change too, or do none of these support freezing the backbone anyway?

tests/trainers/test_change.py
```diff
@@ -199,6 +211,13 @@ def test_freeze_decoder(
             ]
         )
 
+    def test_vit_backbone(self) -> None:
+        ChangeDetectionTask(model='dpt', backbone='tu-vit_base_patch16_224')
+
+    def test_fcn_model(self) -> None:
+        """FCN has no backbone/decoder. Need separate test for full test coverage."""
+        ChangeDetectionTask(model='fcn')
```
adamjstewart, line 219: None of these tests actually run the model through a forward pass, right? Would like to change that. I'm adding a bunch of new CD datamodules, we can have each one use a different model. Can either do this now like we do with inria (e.g., oscd_unet, oscd_deeplab) or do it after I'm done adding more datamodules.

tests/trainers/test_regression.py
```diff
@@ -223,14 +223,27 @@ class TestPixelwiseRegressionTask:
...
     def test_trainer(
         self, monkeypatch: MonkeyPatch, name: str, fast_dev_run: bool
     ) -> None:
         config = os.path.join('tests', 'conf', name + '.yaml')
 
         monkeypatch.setattr(smp, 'Unet', self.create_model)
         monkeypatch.setattr(smp, 'DeepLabV3Plus', self.create_model)
+        monkeypatch.setattr(smp, 'UPerNet', self.create_model)
+        monkeypatch.setattr(smp, 'Segformer', self.create_model)
+        monkeypatch.setattr(smp, 'DPT', self.create_model)
```
adamjstewart, line 246: Honestly wondering if we should do less monkeypatching. This is designed to make the model simpler and therefore faster, but it then doesn't test a real model. We still need to monkeypatch things that download from the internet, but I'm okay with slow integration tests if they better test real models.

torchgeo/trainers/change.py
```diff
@@ -36,7 +36,16 @@ class ChangeDetectionTask(BaseTask):
 
     def __init__(
         self,
-        model: Literal['unet', 'fcsiamdiff', 'fcsiamconc'] = 'unet',
+        model: Literal[
```
adamjstewart, line 39: Should we sort these alphabetically (same in the match-case statement)?

tests/trainers/test_change.py
isaaccorley: Yeah this just tests that the model loads. We can change this though in another PR

tests/trainers/test_regression.py
isaaccorley: Yeah imo we shouldn't monkeypatch this and we should actually test the real model loads and does a forward pass.

PR #2835, opened 2025-06-16
roi_split: support arbitrary Polygon for roi

adamjstewart:
Now that we are using geopandas, we can easily support any arbitrary shapely Polygon: #2382

* roi_sampler no longer supports BoundingBox roi, use shapely Polygon instead

Loosely related to #1170 and #1190

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -121,6 +121,7 @@ intersphinx_mapping = {
     'python': ('https://docs.python.org/3', None),
     'rasterio': ('https://rasterio.readthedocs.io/en/stable/', None),
     'segmentation_models_pytorch': ('https://smp.readthedocs.io/en/stable/', None),
+    'shapely': ('https://shapely.readthedocs.io/en/stable/', None),
     'sklearn': ('https://scikit-learn.org/stable/', None),
     'timm': ('https://huggingface.co/docs/timm/main/en/', None),
     'torch': ('https://docs.pytorch.org/docs/stable/', None),
--- a/tests/datasets/test_splits.py
+++ b/tests/datasets/test_splits.py
@@ -207,9 +207,9 @@ def test_roi_split() -> None:
     train_ds, val_ds, test_ds = roi_split(
         ds,
         rois=[
-            BoundingBox(0, 2, 0, 1, MINT, MAXT),
-            BoundingBox(2, 3.5, 0, 1, MINT, MAXT),
-            BoundingBox(3.5, 4, 0, 1, MINT, MAXT),
+            shapely.box(0, 0, 2, 1),
+            shapely.box(2, 0, 3.5, 1),
+            shapely.box(3.5, 0, 4, 1),
         ],
     )
 
@@ -233,13 +233,7 @@ def test_roi_split() -> None:
 
     # Test invalid input rois
     with pytest.raises(ValueError, match="ROIs in input rois can't overlap."):
-        roi_split(
-            ds,
-            rois=[
-                BoundingBox(0, 2, 0, 1, MINT, MAXT),
-                BoundingBox(1, 3, 0, 1, MINT, MAXT),
-            ],
-        )
+        roi_split(ds, rois=[shapely.box(0, 0, 2, 1), shapely.box(1, 0, 3, 1)])
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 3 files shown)


Moritz-Langer:
There are some issues with the current implementation and potential fixes.

## Issue 1 — ROI polygon not used for chip filtering

After `GeoSampler.__init__` clips the index to the ROI, `GridGeoSampler.__iter__` calls `.bounds` on the clipped geometry, discarding the polygon shape entirely. Every chip inside the bounding box rectangle is yielded — even chips that do not intersect the actual ROI. With a triangular ROI (e.g. a ROI Polygon that splits a tile diagonally), roughly half the generated chips fall completely outside the ROI.

```python
# Current behaviour — only bbox is used, polygon shape is ignored
bounds = self.index.geometry.iloc[i].bounds   # loses polygon shape
xmin, ymin, xmax, ymax = bounds
rows, cols = tile_to_chips(bounds, self.size, self.stride)

for i in range(rows):
    for j in range(cols):
        ...
        yield slice(xmin, xmax), slice(ymin, ymax), slice(tmin, tmax)
        # no check against self.roi
```

**Proposed fix** — intersect each chip box against `self.roi` before yielding:

```python
chip_polygon = shapely.box(xmin, ymin, xmax, ymax)
if not self.roi.intersects(chip_polygon):
    continue
yield slice(xmin, xmax), slice(ymin, ymax), slice(tmin, tmax)
```
See this minimal working test as a fix on the roi:  https://github.com/torchgeo/torchgeo/compare/main...Moritz-Langer:torchgeo:feature/roi_fixed

## Issue 2 — `__len__` overcounts when ROI filtering is applied

However this produces wrong sampler lengths as this is not computed before runtime.
This can be precomputed like follows:

```python
self.length = 0
for i in range(len(self.index)):
    bounds = self.index.geometry.iloc[i].bounds
    xmin_b, ymin_b, xmax_b, ymax_b = bounds
    ...
    for ri in range(rows):
        for ci in range(cols):
            chip_polygon = shapely.box(...)
            if self.roi.intersects(chip_polygon):
                self.length += 1
```
but this is also not ideal since this takes time in all cases.
Please provide feedback for design choices.

adamjstewart:
Yep, I'm working on changing this now. The plan is to use geopandas.sample_points which respects the polygon.

Moritz-Langer:
I have commited a rework that should also fix length generation for the GridGeoSampler. However I could not understand the geopandas sample_points approach. I would guess this might come in handy for the RandomGeoSampler but I could not conceptualize a solution for the GridGeoSampler.

### Performance Comparison: `GridGeoSampler_original` vs. `GridGeoSampler_proposed_solution`

#### Original Implementation
```python
%%timeit
sampler_grid = GridGeoSampler_original(
    image_raster,
    size=img_size,
    stride=img_size,
    roi=roi if roi else None
)
# 1.71 ms ± 83 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
len(sampler_grid)  # 129,548
```

#### Proposed Solution
```python
%%timeit
sampler_grid_new = GridGeoSampler_proposed_solution(
    image_raster,
    size=img_size,
    stride=img_size,
    roi=roi if roi else None
)
# 889 ms ± 58.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
len(sampler_grid_new)  # 85,519
```

#### Additional Context
- `len(sampler_grid_no_roi)`: 152,292

#### Observations
- The proposed solution is slower as it computes the chips at init (889 ms vs. 1.71 ms).
- The proposed solution also returns fewer samples (85,519 vs. 129,548) in my example use case which reduces compute time by days.

If you approve the change I could submit a PR.

adamjstewart:
Any other time I would say "yes please" but I'm currently working on rewriting all of the samplers from scratch so I would hold off. I'll ask you to review the new implementation for me to make sure it looks okay.

Moritz-Langer:
Okay will do, thank you. I am looking forward to seeing your solution.

adamjstewart:
Okay, see my solution in #3552. I use `geopandas.sample_points` in RandomSpatialSampler and check for overlap in GridSpatialSampler. The latter does not yet have a `__len__` method, but this could easily be added by moving the `__iter__` loop into `__init__` and memoizing the valid bounding boxes.

PR #2836, opened 2025-06-16
Fix ResNet18_Weights.LANDSAT_TM_TOA weights from 5 to 7 channels in docs/api/weights/landsat.csv

kiransiitg:
In docs/api/weights/landsat.csv, updated ResNet18_Weights.LANDSAT_TM_TOA from 5 channels to 7 channels.


adamjstewart:
It should still be Landsat 4--5, the only thing you want to change is the number of channels.

PR #2837, opened 2025-06-16
IntersectionDataset: support spatial-only intersection

### LINKED ISSUE #2571: Ignoring Temporal Overlap in Spatial Intersection for Multiple Datasets
### Issue

**Apologies if this has been covered somewhere, but I haven't been able to find it.**

## Use Case

- **Dataset 1**: `['image']` — multiple overlapping images with independent timestamps (e.g., Sentinel-2 from Jan-2023 to Mar-2023).  (A,B,C,D)

- **Dataset 2**: `['mask']` — multiple overlapping images with independent timestamps (e.g., SAR from Apr-2023 to May-2023). (1,2)

Image

For the sake of this example, **no temporally overlapping data** exists. Each dataset has many overlapping regions, and not every image overlaps with every other image within the same dataset.

## Problem Description

1. If both datasets are imported with timestamps, there is **no spatiotemporal overlap** (due to distinct time ranges).  
2. If timestamps in Dataset 2 are **zeroed out**, each pair of potential spatial overlaps is recognized as expected.  
3. However, the **"hit" or query generated in the R-tree index** cannot distinguish between overlapping images in Dataset 2, because the timestamp in the ROI refers only to Dataset 1.

## Goal

- **Obtain all possible combinations** of Dataset 1 and Dataset 2 where a **spatial overlap** exists.  
in the example image this would be: [A1,A2,B1,B2,C1,C2,D1]

## Questions

1. Is there a **recommended process** to accomplish a purely spatial intersection while preserving original timestamps?  
2. It almost seems we need **two ROIs** with matching `x, y` but different `t` values to identify one image from each dataset. Has anyone tackled a similar scenario?

---

Any insights or references are greatly appreciated!

### Fix

_No response_

adamjstewart:
Part of #2382 

Needed for applications like MultiMiner where we have a collection of satellite/drone imagery inputs at different times covering the same year. @MUYang99

Closes #2571

### DIFF
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -821,6 +821,20 @@ class TestIntersectionDataset:
         assert shapely.box(1, 1, 3, 3) in ds.index.geometry
         assert shapely.box(2, 2, 3, 3) in ds.index.geometry
 
+        bounds1 = [(0, 2, 0, 2, pd.Timestamp(2025, 4, 1), pd.Timestamp(2025, 4, 2))]
+        bounds2 = [(1, 3, 1, 3, pd.Timestamp(2025, 4, 3), pd.Timestamp(2025, 4, 4))]
+        ds1 = CustomGeoDataset(bounds1)
+        ds2 = CustomGeoDataset(bounds2)
+        ds = IntersectionDataset(ds1, ds2, spatial_only=True)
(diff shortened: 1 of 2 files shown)


isaaccorley (approved):
lgtm

PR #2840, opened 2025-06-16
Download Eurosat split file if only image folder exists

### LINKED ISSUE #2753: Eurosat No such file or directory: 'data/eurosat-train.txt'
### Description

With `download=True` this should be downloaded..?

### Steps to reproduce

```
from torchgeo.datasets import EuroSAT

>>> ds = EuroSAT(root="data", download=True)
  File "/teamspace/studios/this_studio/torchgeo/torchgeo/datasets/eurosat.py", line 144, in __init__
    with open(os.path.join(self.root, self.split_filenames[split])) as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'data/eurosat-train.txt'
```

### Version

0.8.0.dev0

ashnair1:
Fixes #2753

### DIFF
--- a/tests/datasets/test_eurosat.py
+++ b/tests/datasets/test_eurosat.py
@@ -73,6 +73,38 @@ class TestEuroSAT:
         with pytest.raises(DatasetNotFoundError, match='Dataset not found'):
             EuroSAT(tmp_path)
 
+    def test_missing_split_file(self, monkeypatch: MonkeyPatch, tmp_path: Path) -> None:
+        """Test case where image folder exists but split file is missing."""
+        url = os.path.join('tests', 'data', 'eurosat') + os.sep
+        monkeypatch.setattr(EuroSAT, 'url', url)
+        dataset = EuroSAT(tmp_path, split='train', download=True)
+        os.remove(os.path.join(tmp_path, dataset.split_filenames['train']))
+        dataset = EuroSAT(tmp_path, split='train', download=True)
+        assert os.path.exists(os.path.join(tmp_path, dataset.split_filenames['train']))
+
+    def test_missing_split_file_no_download(
+        self, monkeypatch: MonkeyPatch, tmp_path: Path
+    ) -> None:
+        """Test case where split file is missing and download=False."""
+        url = os.path.join('tests', 'data', 'eurosat') + os.sep
+        monkeypatch.setattr(EuroSAT, 'url', url)
+        dataset = EuroSAT(tmp_path, split='train', download=True)
+        os.remove(os.path.join(tmp_path, dataset.split_filenames['train']))
+        with pytest.raises(FileNotFoundError):
+            EuroSAT(tmp_path, split='train', download=False)
+
+    def test_different_datasets_same_root(
+        self, monkeypatch: MonkeyPatch, tmp_path: Path
+    ) -> None:
+        """Test downloading different dataset variants to same root directory."""
+        url = os.path.join('tests', 'data', 'eurosat') + os.sep
+        monkeypatch.setattr(EuroSAT, 'url', url)
+        monkeypatch.setattr(EuroSAT100, 'url', url)
+        dataset1 = EuroSAT(tmp_path, split='train', download=True)
+        dataset2 = EuroSAT100(tmp_path, split='train', download=True)
+        assert os.path.exists(os.path.join(tmp_path, dataset1.split_filenames['train']))
+        assert os.path.exists(os.path.join(tmp_path, dataset2.split_filenames['train']))
+
     def test_plot(self, dataset: EuroSAT) -> None:
         x = dataset[0].copy()
         dataset.plot(x, suptitle='Test')
--- a/tests/datasets/test_eurosat.py
+++ b/tests/datasets/test_eurosat.py
@@ -73,37 +73,12 @@ class TestEuroSAT:
         with pytest.raises(DatasetNotFoundError, match='Dataset not found'):
             EuroSAT(tmp_path)
 
-    def test_missing_split_file(self, monkeypatch: MonkeyPatch, tmp_path: Path) -> None:
-        """Test case where image folder exists but split file is missing."""
-        url = os.path.join('tests', 'data', 'eurosat') + os.sep
-        monkeypatch.setattr(EuroSAT, 'url', url)
-        dataset = EuroSAT(tmp_path, split='train', download=True)
-        os.remove(os.path.join(tmp_path, dataset.split_filenames['train']))
-        dataset = EuroSAT(tmp_path, split='train', download=True)
-        assert os.path.exists(os.path.join(tmp_path, dataset.split_filenames['train']))
-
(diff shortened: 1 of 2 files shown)


torchgeo/datasets/eurosat.py
```diff
@@ -188,8 +189,18 @@ def _verify(self) -> None:
...
+            # Download the split file if it does not exist
+            if self.download:
+                download_url(
+                    self.url + self.split_filenames[self.split],
+                    self.root,
+                    md5=self.split_md5s[self.split] if self.checksum else None,
+                )
+                return
 
         # Check if zip file already exists (if so then extract)
```
adamjstewart: What about the case where the zip file exists but the split file does not?

I would instead organize this like:
```python
# Split file
if not os.path.exists(split_file):
    if self.download:
        download_url(...)
    else:

# Image files
if os.path.exists(filepath):
    return
elif os.path.exists(zipfile):
    extract_archive(...)
elif self.download:
    download_and_extract_archive(...)
else:
```
This may require changing or even removing `_download/_extract/_check_integrity`.

adamjstewart (changes requested):

adamjstewart:
Note that this isn't perfect, and may not work if you use the same directory for EuroSAT, EuroSATSpatial, and EuroSAT100. We would download and extract the images only once, so depending on the order, you may end up with only 100 images. The safest solution is to use a unique directory for each, although we can't enforce/check that without searching for every single file in the CSV.

torchgeo/datasets/eurosat.py
```diff
@@ -188,22 +189,30 @@ def _verify(self) -> None:
...
+            self._download()
+            self._extract()
+        else:
             raise DatasetNotFoundError(self)
 
-        # Download and extract the dataset
-        self._download()
-        self._extract()
-
     def _download(self) -> None:
```
adamjstewart: Should we remove the split downloads from `_download` now that it's handled manually? If `_download` and `_extract` are only 1 line long, we might even want to delete them and call the functions directly.

ashnair1: Yeah it's redundant now. Will be removing it shortly

ashnair1: @adamjstewart  Actually they currently do different things. In `_verify()` we're only checking if the split file specified by user is present or not. In `_download()` we're downloading all the split files.

We could just merge the `_download` and `_verify` functions so it would look something like this

```python
    def _verify(self) -> None:
        """Verify the integrity of the dataset."""
        # Check if the files already exist
        filepath = os.path.join(self.root, self.base_dir)
        split_file = os.path.join(self.root, self.split_filenames[self.split])

        # Check split file
        if not os.path.exists(split_file):
            if not self.download:

            for split in self.splits:
                download_url(
                    self.url + self.split_filenames[split],
                    self.root,
                    md5=self.split_md5s[split] if self.checksum else None,
                )
        # Check image files
        if os.path.exists(filepath):
            return
        elif self._check_integrity():
            self._extract()
        elif self.download:
            download_url(
            self.url + self.filename, self.root, md5=self.md5 if self.checksum else None
        )
            self._extract()
        else:
```

adamjstewart: I personally don't think we need to download all split files, just the user-requested split. But yes, I agree with merging all methods into one.

tests/datasets/test_eurosat.py
```diff
@@ -60,19 +60,26 @@ def test_add(self, dataset: EuroSAT) -> None:
...
         shutil.rmtree(dataset.root)
         shutil.copy(dataset.url + dataset.filename, tmp_path)
-        type(dataset)(tmp_path)
+        type(dataset)(tmp_path, split=dataset.split)
 
     def test_not_downloaded(self, tmp_path: Path) -> None:
         with pytest.raises(DatasetNotFoundError, match='Dataset not found'):
             EuroSAT(tmp_path)
 
+    def test_missing_images_and_zip_no_download(self, tmp_path: Path) -> None:
```
adamjstewart, line 76: Should we also test the opposite? If the image directory exists but the split file doesn't? That's the issue this PR is fixing.

adamjstewart (approved):
This looks good to me now, merge when you think it's ready

tests/datasets/test_eurosat.py
```diff
@@ -60,19 +60,58 @@ def test_add(self, dataset: EuroSAT) -> None:
...
+        """Test that split file is downloaded if missing but image folder is present."""
+        image_dir = (
+            tmp_path
+            / 'ds'
+            / 'images'
+            / 'remote_sensing'
+            / 'otherDatasets'
+            / 'sentinel_2'
+            / 'tif'
+        )
```
adamjstewart:
```suggestion
        image_dir = tmp_path / EuroSAT.base_dir
```

tests/datasets/test_eurosat.py
```diff
@@ -60,19 +60,58 @@ def test_add(self, dataset: EuroSAT) -> None:
...
+        class_dir = image_dir / 'AnnualCrop'
+        class_dir.mkdir(parents=True, exist_ok=True)
+        (class_dir / 'dummy.tif').touch()
+
+        split_file = tmp_path / 'eurosat-train.txt'
+        if split_file.exists():
+            split_file.unlink()
```
adamjstewart: It should never exist, right?

tests/datasets/test_eurosat.py
```diff
@@ -60,19 +60,58 @@ def test_add(self, dataset: EuroSAT) -> None:
...
+
+        called = {}
+
+        def fake_download_url(url: str, root: str, md5: str) -> None:
+            called['url'] = url
+            split_file.write_text('dummy.tif\n')
+
+        monkeypatch.setattr('torchgeo.datasets.eurosat.download_url', fake_download_url)
```
adamjstewart: I don't love this, can't we monkeypatch the URL to the fake data folder and download (copy) the real file? We already monkeypatch download_url to copy a local file.

ashnair1: Not sure I follow, the monkeypatch download_url would copy everything in the tests/data/eurosat folder (archive, image folder and split files). We're looking to have the directory have just the image folder (ds) before downloading.

adamjstewart: I'm suggesting to remove this line, monkeypatch `url`, and instead use `EuroSAT(download=True)`. Then the dataset will be doing the downloading (copying) for us.

tests/datasets/test_eurosat.py
ashnair1: It shouldn't

tests/datasets/test_eurosat.py
ashnair1: Added test for this case

adamjstewart (approved):

PR #2847, opened 2025-06-20
BoundingBox: deprecate usage

adamjstewart:
The recent time series work in #2382 has made BoundingBox obsolete. Since it is no longer used in TorchGeo, we should deprecate it and remove it in a couple of release cycles.


isaaccorley (approved):
lgtm

PR #2849, opened 2025-06-21
Add SSL4Eco SeCo-Eco Weights

isaaccorley:
Adds the RN50 weights pretrained on the new SSL4Eco dataset using the SeCo-Eco method modified from the original SeCo method. This is from the CVPR Earthvision paper ["SSL4Eco: A Global Seasonal Dataset for Geospatial Foundation Models in Ecology", Plekhanova et al. (2025)](https://arxiv.org/abs/2504.18256).

Weights were extracted from the checkpoint and tweaked to be loadable with timm and torchvision resnet50 and rehosted to huggingface [here](https://huggingface.co/torchgeo/seco-eco).

cc: @PlekhanovaElena @drprojects @jdollinger-bit

### DIFF
--- a/docs/api/weights/sentinel2.csv
+++ b/docs/api/weights/sentinel2.csv
@@ -6,7 +6,8 @@ ResNet50_Weights.SENTINEL2_ALL_DECUR,13,`link <https://github.com/zhu-xlab/DeCUR
 ResNet50_Weights.SENTINEL2_ALL_DINO,13,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,"CC-BY-4.0",90.7,99.1,63.6,
 ResNet50_Weights.SENTINEL2_ALL_MOCO,13,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,"CC-BY-4.0",91.8,99.1,60.9,
 ResNet50_Weights.SENTINEL2_ALL_SOFTCON, 13,`link <https://github.com/zhu-xlab/softcon>`__,`link <https://arxiv.org/abs/2405.20462>`__,"CC-BY-4.0"
-ResNet50_Weights.SENTINEL2_ALL_SECO_ECO, 9,`link <https://github.com/PlekhanovaElena/ssl4eco>`__,`link <https://arxiv.org/abs/2504.18256>`__,"MIT"
+ResNet50_Weights.SENTINEL2_ALL_SECO_ECO, 12,`link <https://github.com/PlekhanovaElena/ssl4eco>`__,`link <https://arxiv.org/abs/2504.18256>`__,"MIT"
+ResNet50_Weights.SENTINEL2_ALL_NDVI_SECO_ECO, 9,`link <https://github.com/PlekhanovaElena/ssl4eco>`__,`link <https://arxiv.org/abs/2504.18256>`__,"MIT"
 ResNet50_Weights.SENTINEL2_MI_MS_SATLAS,9,`link <https://github.com/allenai/satlaspretrain_models>`__,`link <https://arxiv.org/abs/2211.15660>`__,ODC-BY,,,,
 ResNet50_Weights.SENTINEL2_MI_RGB_SATLAS,3,`link <https://github.com/allenai/satlaspretrain_models>`__,`link <https://arxiv.org/abs/2211.15660>`__,ODC-BY,,,,
 ResNet50_Weights.SENTINEL2_SI_MS_SATLAS,9,`link <https://github.com/allenai/satlaspretrain_models>`__,`link <https://arxiv.org/abs/2211.15660>`__,ODC-BY,,,,
--- a/docs/api/weights/sentinel2.csv
+++ b/docs/api/weights/sentinel2.csv
@@ -6,6 +6,7 @@ ResNet50_Weights.SENTINEL2_ALL_DECUR,13,`link <https://github.com/zhu-xlab/DeCUR
 ResNet50_Weights.SENTINEL2_ALL_DINO,13,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,"CC-BY-4.0",90.7,99.1,63.6,
 ResNet50_Weights.SENTINEL2_ALL_MOCO,13,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,"CC-BY-4.0",91.8,99.1,60.9,
 ResNet50_Weights.SENTINEL2_ALL_SOFTCON, 13,`link <https://github.com/zhu-xlab/softcon>`__,`link <https://arxiv.org/abs/2405.20462>`__,"CC-BY-4.0"
+ResNet50_Weights.SENTINEL2_ALL_SECO_ECO, 9,`link <https://github.com/PlekhanovaElena/ssl4eco>`__,`link <https://arxiv.org/abs/2504.18256>`__,"MIT"
 ResNet50_Weights.SENTINEL2_MI_MS_SATLAS,9,`link <https://github.com/allenai/satlaspretrain_models>`__,`link <https://arxiv.org/abs/2211.15660>`__,ODC-BY,,,,
 ResNet50_Weights.SENTINEL2_MI_RGB_SATLAS,3,`link <https://github.com/allenai/satlaspretrain_models>`__,`link <https://arxiv.org/abs/2211.15660>`__,ODC-BY,,,,
 ResNet50_Weights.SENTINEL2_SI_MS_SATLAS,9,`link <https://github.com/allenai/satlaspretrain_models>`__,`link <https://arxiv.org/abs/2211.15660>`__,ODC-BY,,,,
--- a/torchgeo/models/resnet.py
+++ b/torchgeo/models/resnet.py
@@ -155,6 +155,18 @@ _seco_transforms = K.AugmentationSequential(
     data_keys=None,
 )
 
+
+# Normalization only available for RGB dataset, defined here:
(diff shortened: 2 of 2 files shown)


torchgeo/models/resnet.py
```diff
@@ -156,6 +156,18 @@
...
 )
 
+
+# Normalization only available for RGB dataset, defined here:
+# https://github.com/PlekhanovaElena/ssl4eco/blob/7445e048035f7ae31c0eb45e1ed8426c9989fe56/pretraining/pretrain_seco_3heads.py#L140
+# https://github.com/PlekhanovaElena/ssl4eco/blob/7445e048035f7ae31c0eb45e1ed8426c9989fe56/downstream_tasks/test_modules/secoeco_test_module.py#L28
+_seco_eco_bands = ['B2', 'B3', 'B4', 'B5', 'B6', 'B7', 'B8', 'B8A', 'NDVI']
+_seco_eco_transforms = K.AugmentationSequential(
+    K.Resize((224, 224)),
+    K.Normalize(mean=torch.tensor(0.0), std=torch.tensor(10000.0)),
```
isaaccorley, line 164: @PlekhanovaElena can you confirm these are the recommended transforms used during pretraining if someone wants to process a raw sentinel-2 uint16 patch?

adamjstewart: Definitely don't want to divide by 10K for NDVI, may need a vector of length 9

docs/api/weights/sentinel2.csv
```diff
@@ -6,6 +6,7 @@ ResNet50_Weights.SENTINEL2_ALL_DECUR,13,`link <https://github.com/zhu-xlab/DeCUR
 ResNet50_Weights.SENTINEL2_ALL_DINO,13,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,"CC-BY-4.0",90.7,99.1,63.6,
 ResNet50_Weights.SENTINEL2_ALL_MOCO,13,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,"CC-BY-4.0",91.8,99.1,60.9,
 ResNet50_Weights.SENTINEL2_ALL_SOFTCON, 13,`link <https://github.com/zhu-xlab/softcon>`__,`link <https://arxiv.org/abs/2405.20462>`__,"CC-BY-4.0"
+ResNet50_Weights.SENTINEL2_ALL_SECO_ECO, 9,`link <https://github.com/PlekhanovaElena/ssl4eco>`__,`link <https://arxiv.org/abs/2504.18256>`__,"MIT"
```
adamjstewart: @PlekhanovaElena is the 12-band version without NDVI available? That may be more general and useful to us.

Also, I don't see a license on https://huggingface.co/eplekh/secoeco. What license are the weights released under?

torchgeo/models/resnet.py
isaaccorley: In their code when they create NDVI in their dataset they actually multiply it by 10k so it can be normalized later by 10k

isaaccorley: See line 73 here https://github.com/PlekhanovaElena/ssl4eco/blob/main/data_download/download_ssl4eco.py

docs/api/weights/sentinel2.csv
PlekhanovaElena: > @PlekhanovaElena is the 12-band version without NDVI available? That may be more general and useful to us.

I added the weights of [12-band version](https://huggingface.co/eplekh/secoeco/resolve/main/ablation_B12_weights.ckpt) of the model.

> Also, I don't see a license on https://huggingface.co/eplekh/secoeco. What license are the weights released under?

I added MIT license on Hugging Face.
Thanks!

isaaccorley: Thank you! I will add these weights as well!

drprojects:
Thank you @isaaccorley @adamjstewart @plekhanovaelena for the integration :pray::pray::pray:

I have a small concern regarding the transforms. We have a conditional behavior depending on the size of the input image:

```python
class SetPatchSizeToPretraingSize(torch.nn.Module):
    def __init__(self, patch_size, interpolation='BICUBIC'):
        super().__init__()
        self.patch_size = patch_size
        self.interpolation = interpolation

    def forward(self, img):
        h, w, _ = img.shape
        if h < self.patch_size or w < self.patch_size:
            transform = cvtransforms.Resize(
                self.patch_size,
                interpolation=self.interpolation)
        else:
            transform = cvtransforms.CenterCrop(self.patch_size)
        return transform(img)
```

The proposed kornia-only implementation in TorchGeo would indeed reproduce our behavior for images smaller than `patch_size=224`, but this would not work for larger input images:

```python
_seco_eco_transforms = K.AugmentationSequential(
    K.Resize((224, 224)),
    ...
```

I can create a new PR to fix this but would need your opinion first @isaaccorley @adamjstewart. To the best of my knowledge, kornia does not allow conditional mechanisms. How essential is it that `_seco_eco_transforms` be a kornia augmentation? I can suggest two options:
- `_seco_eco_transforms` is a callable but not necessarily a kornia augmentation object
- `_seco_eco_transforms` is a custom `kornia.augmentation.GeometricAugmentationBase2D` with conditional behavior. Follow-up question would then be: where would you prefer such new augmentation class to be defined in the codebase? Is it fine with you to have it in `torchgeo/models/resent.py` just above `_seco_eco_transforms`?

adamjstewart:
We could start with CenterCrop without padding to get it to <= 224, then use Resize to get it to == 224. Then we wouldn't need any conditional logic.

drprojects:
> We could start with CenterCrop without padding to get it to <= 224, then use Resize to get it to == 224. Then we wouldn't need any conditional logic.

I though so too, but CenterCrop does not seem to like input images smaller than the required output crop size:

```python
import torch

K.CenterCrop((20, 20))(torch.rand(3, 10, 10))

>>> Traceback (most recent call last):
  File "/home/damien/miniconda3/envs/ssl4eco/lib/python3.10/site-packages/IPython/core/interactiveshell.py", line 3577, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-24-0987adfc757d>", line 1, in <module>
    K.CenterCrop((20, 20), keepdim=False)(torch.rand(3, 10, 10))
  File "/home/damien/miniconda3/envs/ssl4eco/lib/python3.10/site-packages/kornia/core/module.py", line 311, in __call__
    _output_image = decorated_forward(*inputs, **kwargs)
  File "/home/damien/miniconda3/envs/ssl4eco/lib/python3.10/site-packages/kornia/core/module.py", line 81, in wrapper
    tensor_outputs = func(*args, **kwargs)
  File "/home/damien/miniconda3/envs/ssl4eco/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1739, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/home/damien/miniconda3/envs/ssl4eco/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1750, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/damien/miniconda3/envs/ssl4eco/lib/python3.10/site-packages/kornia/augmentation/base.py", line 251, in forward
    params = self.forward_parameters(batch_shape)
  File "/home/damien/miniconda3/envs/ssl4eco/lib/python3.10/site-packages/kornia/augmentation/base.py", line 218, in forward_parameters
    _params = self.generate_parameters(torch.Size((int(to_apply.sum().item()), *batch_shape[1:])))
  File "/home/damien/miniconda3/envs/ssl4eco/lib/python3.10/site-packages/kornia/augmentation/_2d/geometric/center_crop.py", line 109, in generate_parameters
    return rg.center_crop_generator(batch_shape[0], batch_shape[-2], batch_shape[-1], self.size, self.device)
  File "/home/damien/miniconda3/envs/ssl4eco/lib/python3.10/site-packages/kornia/augmentation/random_generator/_2d/crop.py", line 302, in center_crop_generator
    raise AssertionError(f"Crop size must be smaller than input size. Got ({height}, {width}) and {size}.")
AssertionError: Crop size must be smaller than input size. Got (10, 10) and (20, 20).
```

Would you maybe have a workaround in mind ?

isaaccorley:
You would need to use K.PadTo((224,224))

drprojects:
> You would need to use K.PadTo((224,224))

Which, to the best of my knowledge, would in turn not behave as required for images larger than 224: `K.PadTo` will crop the top-left corner of an image if too large. But in our case we want to CenterCrop images larger than 224.

```python
K.PadTo((5, 5))(torch.arange(100).view(10, 10).float())
```

isaaccorley:
@drprojects Was the pretraining size set to 224x224? Or are these conditional transforms only used for downstream fine-tuning?

Doing a [search](https://github.com/search?q=repo%3APlekhanovaElena%2Fssl4eco%20SetPatchSizeToPretraingSize&type=code) for `SetPatchSizeToPretrainingSize` in the SSL4Eco repo, this transform seems to only be used for downstream tasks. The transforms we define for the weights are those used during pretraining so I think it's fine to keep them as is since it's more of a guide of what the size was during pretraining. Users are free to use whatever transforms they want for downstream tasks.

If we are dead set on using this transform though, I think it would be preferable to upstream this to Kornia directly as it's not geospatial specific and seems like torchvision doesn't have something similar to this either so it would be beneficial for the non-geospatial community as well.

drprojects:
The pretraining was done on 224x224 images. More precisely, it was performed on [random crops resized to 224x224](https://github.com/PlekhanovaElena/ssl4eco/blob/408084c2fb930e923321e86be2367db64e198c8a/pretraining/pretrain_seco_3heads.py#L301-L304). Hence, the model should in theory be robust to a certain level of stretching (which aligns with using `K.Resize((224, 224))` for images smaller than 224x224), but the behavior on larger images downsampled to 224 cannot be guaranteed. 

If you think this is fine and that TorchGeo users will adjust their transforms for downstream tasks, then I am also OK with leaving things as is!

PR #2853, opened 2025-06-22
PyTorch Tutorial: download all split files

### LINKED ISSUE #2895: Dataset download error in getting_started.ipynb
### Description

In the introduction to PyTorch tutorial (https://github.com/microsoft/torchgeo/blob/main/docs/tutorials/pytorch.ipynb) , there is an issue with downloading the EuroSAT100 required for the tutorial. The issue is that only the `'eurosat-100-train.txt'` is downloaded from the below code, and not the `'eurosat-100-val.txt'` and the `'eurosat-100-test.txt'` files.  
```
root = os.path.join(tempfile.gettempdir(), 'eurosat100')
dataset = EuroSAT100(root, download=True)

for i in torch.randint(len(dataset), (10,)):
    sample = dataset[i]
    dataset.plot(sample)
```

This causes an issue in the next block of code
```
train_dataset = EuroSAT100(root, split='train')
val_dataset = EuroSAT100(root, split='val')
test_dataset = EuroSAT100(root, split='test')
```
Error:
```DatasetNotFoundError: Dataset not found in `root='/tmp/eurosat100'` and `download=False`, either specify a different `root` or use `download=True` to automatically download the dataset.```
 
Temporary fix:
```
train_dataset = EuroSAT100(root, split='train')
val_dataset = EuroSAT100(root, split='val', download=True)
test_dataset = EuroSAT100(root, split='test', download=True)
```

### Steps to reproduce

1. Download EuroSAT100 data
```python 
root = os.path.join(tempfile.gettempdir(), 'eurosat100')
dataset = EuroSAT100(root, download=True)
```

2.  Attempt to create split variables
```python
train_dataset = EuroSAT100(root, split='train')
val_dataset = EuroSAT100(root, split='val')
test_dataset = EuroSAT100(root, split='test')
```

### Version

0.7.1

adamjstewart:
Bug introduced in 0.7.1

TODO: figure out why CI didn't catch this

Fixes #2895

### DIFF
--- a/docs/tutorials/pytorch.ipynb
+++ b/docs/tutorials/pytorch.ipynb
@@ -173,9 +173,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "train_dataset = EuroSAT100(root, split='train')\n",
-    "val_dataset = EuroSAT100(root, split='val')\n",
-    "test_dataset = EuroSAT100(root, split='test')"
+    "train_dataset = EuroSAT100(root, split='train', download=True)\n",
+    "val_dataset = EuroSAT100(root, split='val', download=True)\n",
+    "test_dataset = EuroSAT100(root, split='test', download=True)"
    ]
   },
   {


adamjstewart:
Actually, a more philosophical question: should all datasets download all splits at once? This is what our data module base classes assume, but it isn't always true. Not sure what we should do.

adamjstewart:
#2891 can be used to better test this.

There are still some philosophical questions of how datamodules should behave, but let's not let that hold up this bug fix.

isaaccorley (approved):

PR #2856, opened 2025-06-24
Lacuna African Field Boundaries Dataset #2846

isaaccorley:
Remade this PR because the other one was messy. This PR adds the African Field Boundaries dataset from the paper ["A region-wide, multi-year set of crop field boundary labels for Africa", Estes et al. (2024)](https://arxiv.org/abs/2412.18483). This dataset is similar to FTW with 3-classes, however FTW uses Sentinel-2 imagery while this dataset has ~3m res Planet RGBN imagery. This dataset is straightforward -- contains a directory of images and a directory of masks with the same filename prefix. The only caveat is that it needs to be downloaded from AWS S3 through the terminal so no auto-download. Possibly rehosting to HuggingFace would be ideal (@ldemaz)

TODO:
- datamodule
- tests

Some samples (percentile normalized to the [1, 99] percentile range):
image
image
image
image

cc: @ldemaz @cholmes


adamjstewart:
We actually do support auto-downloads from AWS and Azure. See what I did with the SpaceNet and SatlasPretrain datasets. However, depending on the dataset size and license, I would also be fine with rehosting on HF.

isaaccorley:
> We actually do support auto-downloads from AWS and Azure. See what I did with the SpaceNet and SatlasPretrain datasets. However, depending on the dataset size and license, I would also be fine with rehosting on HF.

The plan is to rehost to HF under @ldemaz org. He has given me access.

docs/api/datasets.rst
```diff
@@ -370,6 +370,11 @@ Inria Aerial Image Labeling
 
 .. autoclass:: InriaAerialImageLabeling
 
+LacunaAfricanFieldBoundaries
```
adamjstewart: Let's add spaces to the name

torchgeo/datasets/lacuna.py
```diff
@@ -0,0 +1,249 @@
...
+        Args:
+            root: root directory where dataset can be found
+            bands: load all RGBN bands or RGB only. One of 'rgb' or 'all'.
+            transforms: a function/transform that takes input sample and its target as
+                entry and returns a transformed version
+            download: if True, download dataset and store it in the root directory
+            checksum: if True, check the MD5 of the downloaded files (may be slow)
+
+        Raises:
+            DatasetNotFoundError: If dataset is not found.
```
adamjstewart, line 87: Also raises AssertionError if # images != # masks

torchgeo/datasets/lacuna.py
```diff
@@ -0,0 +1,249 @@
...
+        bands: Literal['rgb', 'all'] = 'all',
+        transforms: Callable[[dict[str, Tensor]], dict[str, Tensor]] | None = None,
+        download: bool = False,
+        checksum: bool = False,
+    ) -> None:
+        """Initialize a new LacunaAfricanFieldBoundaries dataset instance.
+
+        Args:
+            root: root directory where dataset can be found
+            bands: load all RGBN bands or RGB only. One of 'rgb' or 'all'.
```
adamjstewart: If we're going to support this, it may be more consistent to pass in a list of bands like our other RasterDatasets

torchgeo/datasets/lacuna.py
```diff
@@ -0,0 +1,249 @@
...
+            show_titles: flag indicating whether to show titles above each panel
+            suptitle: optional string to use as a suptitle
+
+        Returns:
+            a matplotlib Figure with the rendered sample
+
+        Raises:
+            RGBBandsMissingError: If *bands* does not include all RGB bands.
+        """
+        image = sample['image'][:3].permute(1, 2, 0).float().numpy()
```
adamjstewart: Seems like order somewhat depends on whether bands==all or rgb?

PR #2862, opened 2025-06-28
BRIGHT: add data module, change detection compatibility

adamjstewart:
Related to #2382

- [x] BRIGHT: add new data module
- [x] BRIGHT: convert dataset to be compatible with new ChangeDetectionTask


tests/data/bright/data.py
```diff
@@ -95,16 +95,16 @@ def populate_data(ids: list[str], dir_name: str, with_target: bool = True) -> No
...
             ROOT, DATA_DIR, dir_name, 'post-event', f'{sid}_post_disaster.tif'
         )
-        write_tif(post_path, channels=1)
+        write_tif(post_path, channels=1, classes=256)
         if with_target:
             target_path = os.path.join(
                 ROOT, DATA_DIR, dir_name, 'target', f'{sid}_building_damage.tif'
             )
-            write_tif(target_path, channels=1)
+            write_tif(target_path, channels=1, classes=2)
```
adamjstewart: TODO: check if this is correct, and if all of these are really uint8

torchgeo/datamodules/bright.py
```diff
@@ -0,0 +1,74 @@
...
+from ..datasets import BRIGHTDFC2025
+from .geo import NonGeoDataModule
+
+
+class BRIGHTDFC2025DataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the BRIGHT dataset.
+
+    .. versionadded:: 0.8
+    """
+
```
adamjstewart, line 19: TODO: check mean/std normalization range

torchgeo/datamodules/bright.py
```diff
@@ -0,0 +1,74 @@
...
+        """Apply batch augmentations to the batch after it is transferred to the device.
+
+        Args:
+            batch: A batch of data that needs to be altered or augmented.
+            dataloader_idx: The index of the dataloader to which the batch belongs.
+
+        Returns:
+            A batch of data.
+        """
+        # This solves a special case where if batch_size=1 the mask won't be stacked correctly
```
adamjstewart: TODO: I hate that we need to do this, is this a bug in Kornia?

tests/data/bright/data.py
adamjstewart: All uint8, but should be 4 classes, this is multiclass change detection

torchgeo/datamodules/bright.py
adamjstewart: Good enough

torchgeo/trainers/change.py
```diff
@@ -263,7 +263,7 @@ def _shared_step(self, batch: Any, batch_idx: int, stage: str) -> Tensor:
         x = batch['image']
         y = batch['mask']
 
-        if model == 'unet':
+        if not model.startswith('fcsiam'):
```
adamjstewart, line 266: This bug was introduced in @isaaccorley's #2828

torchgeo/trainers/change.py
```diff
@@ -281,15 +281,8 @@ def _shared_step(self, batch: Any, batch_idx: int, stage: str) -> Tensor:
         # Retrieve the correct metrics based on the stage
         metrics = getattr(self, f'{stage}_metrics', None)
         if metrics:
-            # Transform predictions for metrics calculation
-            match self.hparams['task']:
-                case 'binary' | 'multilabel':
-                    y_hat = (y_hat.sigmoid() >= 0.5).long()
-                case 'multiclass':
-                    y_hat = y_hat.argmax(dim=1)
```
adamjstewart, line 289: This bug was introduced (by me) in @hkristen's #2890

torchgeo/trainers/change.py
```diff
@@ -281,15 +281,8 @@ def _shared_step(self, batch: Any, batch_idx: int, stage: str) -> Tensor:
...
-
             metrics(y_hat, y)
-            self.log_dict({f'{k}': v for k, v in metrics.compute().items()})
+            self.log_dict(metrics, batch_size=x.shape[0])
```
adamjstewart, line 285: This has been there since @keves1's #2422, I don't think we need this much complexity.

torchgeo/trainers/change.py
```diff
@@ -311,7 +304,7 @@ def _shared_step(self, batch: Any, batch_idx: int, stage: str) -> Tensor:
                     case 'binary' | 'multilabel':
                         batch['prediction'] = (y_hat.sigmoid() >= 0.5).long()
                     case 'multiclass':
-                        batch['prediction'] = y_hat.argmax(dim=1)
+                        batch['prediction'] = y_hat.argmax(dim=1, keepdim=True)
```
adamjstewart, line 307: Could also change all multiclass CD datasets to assume prediction has no channel dimension. But then prediction and mask will not have consistent dimensions.

PR #2867, opened 2025-07-02
Add L-TAE model

MUYang99:
Add L-TAE model for satellite image time series processing: https://arxiv.org/pdf/2007.00586, https://github.com/VSainteuf/lightweight-temporal-attention-pytorch


MUYang99:

docs/api/models.rst
```diff
@@ -119,6 +119,12 @@ Vision Transformer
 .. autoclass:: ViTSmall14_DINOv2_Weights
 .. autoclass:: ViTBase14_DINOv2_Weights
 
+
```
isaaccorley: Can you move this to its place alphabetically?

tests/models/test_ltae.py
```diff
@@ -0,0 +1,51 @@
+"""
```
isaaccorley: This needs the Microsoft header that is in other files

torchgeo/models/__init__.py
```diff
@@ -109,4 +110,5 @@
     'vit_large_patch16_224',
     'vit_small_patch14_dinov2',
     'vit_small_patch16_224',
+    'LTAE',
```
isaaccorley: Can you move this to the correct place alphabetically?

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,213 @@
+"""
+Lightweight Temporal Attention Encoder module
```
isaaccorley: This needs the Microsoft header. The reference to the paper can be in the class docstring

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,213 @@
...
+            T (int): Period to use for the positional encoding
+            len_max_seq (int, optional): Maximum sequence length, used to pre-compute the positional encoding table
+            positions (list, optional): List of temporal positions to use instead of position in the sequence
+            d_model (int, optional): If specified, the input tensors will first processed by a fully connected layer
+                to project them into a feature space of dimension d_model
+            return_att (bool): If true, the module returns the attention masks along with the embeddings (default False)
+
+        """
+
+        super(LTAE, self).__init__()
```
isaaccorley: Change to `super()`

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,213 @@
...
+import torch.nn as nn
+import numpy as np
+import copy
+
+
+class LTAE(nn.Module):
+    def __init__(self, in_channels=128, n_head=16, d_k=8, n_neurons=[256,128], dropout=0.2, d_model=256,
+                 T=1000, len_max_seq=24, positions=None, return_att=False):
+        """
+        Sequence-to-embedding encoder.
```
isaaccorley: Needs a new line after this line before the args

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,213 @@
...
+            d_model (int, optional): If specified, the input tensors will first processed by a fully connected layer
+                to project them into a feature space of dimension d_model
+            return_att (bool): If true, the module returns the attention masks along with the embeddings (default False)
+
+        """
+
+        super(LTAE, self).__init__()
+        self.in_channels = in_channels
+        self.positions = positions
+        self.n_neurons = copy.deepcopy(n_neurons)
```
isaaccorley: Probably don't need the deep copy here. If the concern is that the list could be modified then we should make it a tuple

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,213 @@
...
+        self.inlayernorm = nn.LayerNorm(self.in_channels)
+
+        self.outlayernorm = nn.LayerNorm(n_neurons[-1])
+
+        self.attention_heads = MultiHeadAttention(
+            n_head=n_head, d_k=d_k, d_in=self.d_model)
+
+        assert (self.n_neurons[0] == self.d_model)
+
+        activation = nn.ReLU()
```
isaaccorley: Change to `nn.ReLU(inplace=True)`

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,213 @@
...
+
+        layers = []
+        for i in range(len(self.n_neurons) - 1):
+            layers.extend([nn.Linear(self.n_neurons[i], self.n_neurons[i + 1]),
+                           nn.BatchNorm1d(self.n_neurons[i + 1]),
+                           activation])
+
+        self.mlp = nn.Sequential(*layers)
+
+        self.dropout = nn.Dropout(dropout)
```
isaaccorley: Change to `nn.Dropout(dropout, inplace=True)`

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def get_posi_angle_vec(position):
+        return [cal_angle(position, hid_j) for hid_j in range(d_hid)]
+
+    sinusoid_table = np.array([get_posi_angle_vec(pos_i) for pos_i in positions])
+
+    sinusoid_table = np.sin(sinusoid_table)  # dim 2i
+    sinusoid_table[:, clip:] = torch.zeros(sinusoid_table[:, clip:].shape)
+
+    if torch.cuda.is_available():
```
isaaccorley: This should be an arg passed to the function instead of checking if cuda is available -- for example this wouldn't work on Mac mps devices

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,213 @@
...
+        enc_output = x + self.position_enc(src_pos)
+
+        enc_output, attn = self.attention_heads(enc_output, enc_output, enc_output)
+
+        enc_output = enc_output.permute(1, 0, 2).contiguous().view(sz_b, -1)  # Concatenate heads
+
+        enc_output = self.outlayernorm(self.dropout(self.mlp(enc_output)))
+
+        if self.return_att:
+            return enc_output, attn
```
isaaccorley: We can likely drop optionally returning the attention, this would produce a graph break. If a user wants the attention maps they could simply override the forward method

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,213 @@
+"""
+Lightweight Temporal Attention Encoder module
+
+Credits:
+The module is heavily inspired by the works of Vaswani et al. on self-attention and their pytorch implementation of
+the Transformer served as code base for the present script.
+
+paper: https://arxiv.org/abs/1706.03762
+code: github.com/jadore801120/attention-is-all-you-need-pytorch
```
adamjstewart: Did you mean to reference https://github.com/VSainteuf/lightweight-temporal-attention-pytorch? Also need to add "Copyright (c) 2020 VSainteuf (Vivien Sainte Fare Garnot)" since we are copying code from another library.

isaaccorley:
@MUYang99 any thoughts on just scrapping the custom implementations of MHA, SDPA, and positional embeddings with the implementations from the torch libraries?

I think that torch has more optimized versions of [MHA](https://docs.pytorch.org/docs/stable/generated/torch.nn.MultiheadAttention.html) and [SDPA](https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html). I think the LTAE code is just very old so the implementation didn't have these classes available in torch yet.

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,319 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# Copyright (c) 2020 VSainteuf (Vivien Sainte Fare Garnot)
+
+"""Lightweight Temporal Attention Encoder (LTAE) model."""
```
adamjstewart:
```suggestion
"""Lightweight Temporal Attention Encoder (L-TAE) model."""
```

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,319 @@
...
+class LTAE(nn.Module):
+    """Lightweight Temporal Attention Encoder (L-TAE).
+
+    This model implements a lightweight temporal attention encoder that processes
+    time series data using a multi-head attention mechanism. It is designed to
+    efficiently encode temporal sequences into fixed-length embeddings.
+
+    The model architecture is based on the paper:
+    "Lightweight Temporal Self-Attention for Classifying Satellite Images Time Series: https://arxiv.org/pdf/2007.00586"
+
```
adamjstewart:
```suggestion
    If you use this model in your research, please cite the following paper:
    
    * https://arxiv.org/abs/2007.00586
    
    .. versionadded:: 0.8
```

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,319 @@
...
+        enc_output = x + self.position_enc(src_pos)
+        enc_output, attn = self.attention_heads(enc_output, enc_output, enc_output)
+        enc_output = (
+            enc_output.permute(1, 0, 2).contiguous().view(sz_b, -1)
+        )  # Concatenate heads
+
+        enc_output = self.outlayernorm(self.dropout(self.mlp(enc_output)))
+
+        return enc_output  # type: ignore[no-any-return]
```
adamjstewart: We should try to avoid ignoring mypy errors whenever possible. In this case, maybe we can use:
```python
enc_output: Tensor = ...
```

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,312 @@
...
+    sinusoid_table_np[:, 1::2] = np.cos(sinusoid_table_np[:, 1::2])  # dim 2i+1
+
+    return torch.FloatTensor(sinusoid_table_np).to(device)
+
+
+if __name__ == '__main__':
+    model = LTAE(in_channels=128)
+    x = torch.randn(4, 24, 128)
+    output = model(x)
+    print(output.shape)
```
adamjstewart: This debugging stuff can be removed, only scripts need to check if name == main.

docs/api/models.rst
```diff
@@ -119,6 +124,7 @@ Vision Transformer
...
```
isaaccorley:
```suggestion
```

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,169 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# Copyright (c) 2020 VSainteuf (Vivien Sainte Fare Garnot)
+
+"""Lightweight Temporal Attention Encoder (L-TAE) model."""
+
+from __future__ import annotations
```
isaaccorley:
```suggestion
```

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,169 @@
...
+
```
isaaccorley:
```suggestion
from typing import Sequence

```

adamjstewart:
Can you resolve the merge conflicts with `docs/api/models.rst`? This will then allow the tests to run.

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,169 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# Copyright (c) 2020 VSainteuf (Vivien Sainte Fare Garnot)
```
isaaccorley, line 4: @adamjstewart since we've modified the code pretty heavily I'm not sure if this copyright should be at the top or just within the class docstring? Also the original codebase is MIT licensed, not sure if we need to mention that somewhere

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,169 @@
...
+        self,
+        in_channels: int = 128,
+        n_head: int = 16,
+        d_k: int = 8,
+        n_neurons: tuple[int, ...] = (256, 128),
+        dropout: float = 0.2,
+        d_model: int | None = 256,
+        T: int = 1000,
+        len_max_seq: int = 24,
+        positions: list[int] | None = None,
```
isaaccorley:
```suggestion
        positions: Sequence[int] | None = None,
```

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,169 @@
...
+        """Forward pass of the model.
+
+        Args:
+            x: Input tensor of shape (batch_size, seq_len, in_channels)
+
+        Returns:
+            torch.Tensor: Output tensor of shape (batch_size, n_neurons[-1])
+
+        Raises:
+            AssertionError: If input tensor dimensions don't match expected shape
```
isaaccorley:
```suggestion
```

isaaccorley: No need to assert here, pytorch's default assertion is self-explanatory

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        sz_b, seq_len, d = x.shape
+        assert d == self.in_channels, (
+            f'Input channels {d} does not match expected channels {self.in_channels}'
+        )
```
isaaccorley:
```suggestion
```

isaaccorley: No need to assert here, pytorch's default assertion is self-explanatory and we don't do this for any other models

tests/models/test_ltae.py
```diff
@@ -0,0 +1,50 @@
...
+        """Test forward pass."""
+        batch_size = 4
+        seq_len = 24
+        in_channels = 128
+
+        model = LTAE(in_channels=in_channels)
+        x = torch.randn(batch_size, seq_len, in_channels)
+        output = model(x)
+
+        assert output.shape[0] == batch_size
```
isaaccorley:
```suggestion
        assert output.shape[0] == (batch_size, model.n_neurons[-1])
```

torchgeo/models/ltae.py
adamjstewart: I'm fine with putting it in the class docstring, but we might need to put it in multiple places. The entire file is already MIT, so I don't think we need to mention that twice.

isaaccorley: Ok then disregard this then and we can leave it as is

tests/models/test_ltae.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def test_invalid_input(self) -> None:
+        """Test invalid input shape."""
+        batch_size = 4
+        seq_len = 24
+        in_channels = 128
+        wrong_channels = 64  # Different from model's in_channels
+
+        with pytest.raises(AssertionError):
```
isaaccorley:
```suggestion
        with pytest.raises(RuntimeError):
```

isaaccorley (commented):
Just need to format and fix mypy complaints and this is good to go!

isaaccorley (dismissed):
This lgtm @adamjstewart any final nits?

docs/api/models.rst
```diff
@@ -60,6 +60,11 @@ FC Siamese Networks
 .. autoclass:: FCSiamConc
 .. autoclass:: FCSiamDiff
 
+LTAE
+^^^^
```
adamjstewart:
```suggestion
L-TAE
^^^^^
```

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,164 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# Copyright (c) 2020 VSainteuf (Vivien Sainte Fare Garnot)
+
+"""Lightweight Temporal Attention Encoder (L-TAE) model."""
+
+from __future__ import annotations
```
adamjstewart: Is this used anywhere?

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,164 @@
...
+    .. versionadded:: 0.8
+
+    """
+
+    def __init__(
+        self,
+        in_channels: int = 128,
+        n_head: int = 16,
+        d_k: int = 8,
+        n_neurons: tuple[int, ...] = (256, 128),
```
adamjstewart:
```suggestion
        n_neurons: Sequence[int] = (256, 128),
```

torchgeo/models/ltae.py
```diff
@@ -0,0 +1,164 @@
...
+        """Forward pass.
+
+        Args:
+            x: Input tensor of shape (batch_size, seq_len, d_model)
+
+        Returns:
+            torch.Tensor: Output tensor with positional encoding added
+        """
+        # Get positional encoding up to the sequence length
+        pe = self.pe[:, : x.size(1)]  # type: ignore[index]
```
adamjstewart, line 162: Why do we need to ignore?

MUYang99: The type: ignore[index] is necessary because mypy can't automatically determine that self.pe is a torch.Tensor just from the self.register_buffer('pe', pe) call. It doesn't know that self.pe can be indexed.

torchgeo/models/ltae.py
MUYang99: To pass the ruff check

adamjstewart: I removed it and ruff still passes

PR #2868, opened 2025-07-02
EarthVision2025_Data_Challenge_Winning_Solution 

xboboji:
It includes instructions and code for implementing EarthVision2025_Data_Challenge_Winning_Solution


adamjstewart (changes requested):
First of all, congratulations on your victory!

At the moment, we are actually trying to deprecate the `experiments` directory in the TorchGeo repo. These were experiments used for official TorchGeo papers like https://arxiv.org/abs/2111.08872. We're planning to move those to their own repos in the future.

I also see a lot of code copied from timm and other repos. Unfortunately, this can't be added to TorchGeo as it would result in the entire repo being distributed under multiple licenses (MIT and Apache-2.0). You would also need to contribute all of your own code under an MIT license with a Microsoft copyright.

I do see a lot of useful stuff in here though. Perhaps you could start with separate PRs for each model or dataset you would like to add? Models would go in `torchgeo/models/` for example. We would also need unit tests for all new code.

See https://torchgeo.readthedocs.io/en/stable/user/contributing.html for more details.

adamjstewart:
@xboboji is this PR still a work in progress? Trying to close out old PRs.

adamjstewart:
No response from contributor, closing as abandoned. Feel free to open new PRs for specific components you think could be useful for others (models, tutorials) as long as its under an MIT license.

PR #2876, opened 2025-07-08
Delineate Anything YOLO Model

isaaccorley:
This PR adds the [Delineate Anything](https://lavreniuk.github.io/Delineate-Anything/) YOLO11x and YOLO11n pretrained models. This adds a `models` extras group to the dependencies to add the `ultralytics` package. The models are available as AGPL-3.0 licensed.

image


adamjstewart:
@robmarkcole wanted to check with you and see how you feel about this. Ultralytics would not be a required dependency, so this change shouldn't impact you.

robmarkcole:
@adamjstewart I think using optional imports is a good way to go

adamjstewart:
@isaaccorley might need to find a way to update this error message:
```
ultralytics is not installed and is required to use this dataset. Either run:

$ pip install ultralytics

to install just this dependency, or:

$ pip install torchgeo[datasets]

to install all optional dataset dependencies.
```

isaaccorley:
> @isaaccorley might need to find a way to update this error message:
> 
> ```
> 
> ultralytics is not installed and is required to use this dataset. Either run:
> 
> 
> 
> $ pip install ultralytics
> 
> 
> 
> to install just this dependency, or:
> 
> 
> 
> $ pip install torchgeo[datasets]
> 
> 
> 
> to install all optional dataset dependencies.
> 
> ```

Yeah I saw this. I might save this for another PR because I'm going to add a group arg so that it prints out the correct extras to install like torchgeo[models] vs torchgeo[datasets] then I'll have to go back and fix everywhere that it's used.

edit: @adamjstewart I went ahead and just changed the error message. PTAL

.github/workflows/tests.yaml
```diff
@@ -32,12 +32,12 @@ jobs:
...
+          key: ${{ runner.os }}-${{ runner.arch }}-Python-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('requirements/required.txt', 'requirements/datasets.txt', 'requirements/models.txt', 'requirements/tests.txt') }}
         if: ${{ runner.os != 'macOS' }}
       - name: Install pip dependencies
         if: steps.cache.outputs.cache-hit != 'true'
         run: |
-          pip install -r requirements/required.txt -r requirements/datasets.txt -r requirements/tests.txt
+          pip install -r requirements/required.txt -r requirements/datasets.txt -r requirements/models.txt -r requirements/tests.txt
           pip cache purge
       - name: List pip dependencies
         run: pip list
```
adamjstewart, line 43: Can we change the `datasets` test name to be `optional` or something like that? We are trying to ensure that all tests pass even if optional deps are missing. The name `datasets` was never very clear about this. Maybe `optional` isn't clear either, it makes it sound like it isn't required to pass.

hubconf.py
```diff
@@ -51,6 +52,13 @@
...
     'vit_small_patch16_224',
+    'yolo',
 )
 
-dependencies = ['timm', 'torchvision', 'segmentation_models_pytorch', 'kornia']
+dependencies = [
+    'timm',
+    'torchvision',
+    'segmentation_models_pytorch',
+    'kornia',
```
adamjstewart: Can we sort kornia first alphabetically while we're at it?

pyproject.toml
```diff
@@ -118,6 +118,10 @@ docs = [
     # https://github.com/pytorch/pytorch_sphinx_theme/issues/175
     "sphinx>=4,<6",
 ]
+models = [
+    # ultralytics 8.3.0+ for loading with weights_only=False fixed
+    "ultralytics>=8.3.0",
```
```suggestion
    # ultralytics 8.3+ fixes loading with weights_only=False
    "ultralytics>=8.3",
```

requirements/min-reqs.old
```diff
@@ -26,14 +26,17 @@ typing-extensions==4.5.0
 h5py==3.8.0
 laspy==2.0.0
 netCDF4==1.6.1
-opencv-python==4.5.5.64
+opencv-python==4.6.0.66  # ultralytics 8.3.0 depends on opencv-python>=4.6.0
```
adamjstewart: Move this comment to `pyproject.toml` and also bump the minimum there.

requirements/models.txt
```diff
@@ -0,0 +1 @@
+ultralytics==8.3.162
```
adamjstewart, line 1: ultralytics appears to have almost daily releases. For the sake of both of our sanity, let's do what we did for setuptools in `.github/dependabot.yml`

tests/models/test_api.py
```diff
@@ -102,11 +105,24 @@
...
+
+    ULTRALYTICS_INSTALLED = True
+except ImportError:
+    ULTRALYTICS_INSTALLED = False
+
+
 @pytest.mark.parametrize('builder', builders)
 def test_get_model(builder: Callable[..., nn.Module]) -> None:
+    if builder == yolo and not ULTRALYTICS_INSTALLED:
+        pytest.skip('Ultralytics is not installed, skipping YOLO model tests')
```
adamjstewart:
```suggestion
@pytest.mark.parametrize('builder', builders)
def test_get_model(builder: Callable[..., nn.Module]) -> None:
    builder == yolo and pytest.importorskip('ultralytics', minversion='8.3')
```

tests/models/test_api.py
```diff
@@ -80,6 +82,7 @@
     vit_large_patch16_224,
     vit_small_patch14_dinov2,
     vit_small_patch16_224,
+    yolo,
```
adamjstewart, line 85: Alternatively, something like:
```suggestion
    pytest.param(yolo, marks=pytest.importorskip('ultralytics', minversion='8.3')),
```

tests/models/test_yolo.py
```diff
@@ -0,0 +1,55 @@
...
+
+import pytest
+import torch
+from _pytest.fixtures import SubRequest
+from pytest import MonkeyPatch
+from torchvision.models._api import WeightsEnum
+
+from torchgeo.models import YOLO_Weights, yolo
+
+pytest.importorskip('ultralytics', minversion='8.3.0')
```
adamjstewart:
```suggestion
pytest.importorskip('ultralytics', minversion='8.3')
```

torchgeo/datasets/utils.py
```diff
@@ -699,9 +699,9 @@ def lazy_import(name: str) -> Any:
```
adamjstewart, line 699: "dataset" also appears in the first sentence, maybe "feature" or "class"?

torchgeo/models/yolo.py
```diff
@@ -0,0 +1,94 @@
...
+            'bands': ['R', 'G', 'B'],
+            'publication': 'https://arxiv.org/abs/2409.16252',
+            'repo': 'https://github.com/Lavreniuk/Delineate-Anything',
+            'resolution': None,
+            'license': 'AGPL-3.0',
+        },
+    )
+
+
+def yolo(weights: YOLO_Weights | None = None, *args: Any, **kwargs: Any) -> nn.Module:
```
adamjstewart, line 71: We have separate functions for things like `resnet18` and `resnet50`, wondering if we want/need separate functions for different versions of yolo too.

torchgeo/models/yolo.py
```diff
@@ -0,0 +1,94 @@
...
+    """YOLO model.
+
+    .. versionadded:: 0.8
+
+    Args:
+        weights: Pre-trained model weights to use.
+        *args: Additional arguments to pass to ``ultralytics.YOLO``
+        **kwargs: Additional keyword arguments to pass to ``ultralytics.YOLO``
```
adamjstewart: Can we link to the actual YOLO docs? This would make it easier to see which args/kwargs are supported

adamjstewart (changes requested):
Apologies for how nitpicky this review is, it's not every day that we add a new extra so I want to make sure we get it right.

Can we also update `docs/user/installation.rst` to advertise the new extra? Can just replace `[datasets]` with `[datasets,models]`, doesn't have to be anything fancy.

.github/workflows/tests.yaml
isaaccorley: I think optional makes the most sense just because that's how it's always defined in pyproject.toml files

torchgeo/models/yolo.py
isaaccorley: We don't actually, the `ultralytics.YOLO` function can take in any of the supported YOLO models. I think we can just split things up by the weights but have the same `yolo` function for loading

adamjstewart: But so can the `torchvision.models.ResNet` class. It's just annoying to keep track of all the right parameters to use to get a specific model, so there are convenience functions for certain setups. For example, it's currently unclear to me how I would get YOLOvX for all X.

adamjstewart (approved):
Only unsure about whether or not we should have version-specific convenience functions.

hubconf.py
```diff
@@ -51,6 +52,13 @@
...
+    'yolo',
 )
 
-dependencies = ['timm', 'torchvision', 'segmentation_models_pytorch', 'kornia']
+dependencies = [
+    'kornia',
+    'segmentation_models_pytorch',
+    'timm',
+    'torchvision',
+    'ultralytics',
```
adamjstewart, line 63: This is actually kind of unfortunate, I wonder if it's possible to make this optional...

PR #2879, opened 2025-07-09
LEVIR-CD: change detection compatibility

adamjstewart:
Related to #2382


torchgeo/datamodules/levircd.py
```diff
@@ -45,21 +45,27 @@ def __init__(
         self.patch_size = _to_tuple(patch_size)
 
         self.train_aug = K.AugmentationSequential(
-            K.Normalize(mean=self.mean, std=self.std),
-            K.RandomCrop(self.patch_size, pad_if_needed=True),
+            K.VideoSequential(
+                K.Normalize(mean=self.mean, std=self.std),
+                K.RandomCrop(self.patch_size, pad_if_needed=True),
+            ),
```
adamjstewart, line 51: This is interesting. Before train_aug, our x and y are:

* image: BxTxCxHxW (2x2x3x32x32)
* mask: BxHxW (2x32x32)

but after our train_aug, they become:

* image: BxTxCxHxW (2x2x3x32x32)
* mask: 1xHxW (1x32x32)

This appears to be due to RandomCrop, and removing RandomCrop fixes this. Perhaps VideoSequential expects both image and mask to have a time axis?

isaaccorley: I think you're missing a channel dimension for the mask (or maybe this is unique to VideoSequential)

When I make the mask (2,1,32,32) I get the correct output shape

```python
import torch

augs = K.AugmentationSequential(
   K.VideoSequential(
      K.Normalize(torch.zeros(3,), torch.ones(3,)),
      K.RandomCrop((32, 32), pad_if_needed=True)),
   data_keys=None, keepdim=True
)
batch = dict(image=torch.randn(2,2,3,32,32), mask=torch.randint(0,2,(2,1,32,32)))
output = augs(batch)
for k, v in output.items():
   print(k, v.shape)
# image torch.Size([2, 2, 3, 32, 32])
# mask torch.Size([2, 1, 32, 32])
```

adamjstewart: Yes, but why should we need a channel dimension for the mask? Should we add one to all CD datasets? Or should we do this dynamically in the trainer?

isaaccorley: I briefly remember that kornia devs solved this for augmentation sequential where they auto add the channel dimensions in if it doesn't exist for the mask. This might be a feature needed for VideoSequential though.

Seems like the short term workaround is to just add a channel dimension in the trainer methods and then squeeze it back out before computing the loss

adamjstewart: I thought about it and it doesn't work if it happens in the trainer, because that's after the transform runs.

isaaccorley: Oh yeah sorry, I mean it would need to be in the `datamodule.on_after_batch_transfer` method

adamjstewart: I'm actually thinking about adding a channel dim to all CD masks at the dataset level. It wouldn't be consistent with our other semantic segmentation datasets, but it's kinda silly to add and remove a channel dim in both the datamodule and the trainer.

isaaccorley: I'm personally fine with this and i actually like to be explicit with having a channel dimension.

I think the only reason we don't have a channel dim is for 1) matplotlib plotting expects no channel dim and 2) CE loss expects no channel dim.

However it would probably make more sense to always have a channel dim and just squeeze right before plotting and the loss computation.

isaaccorley (approved):
LGTM

PR #2891, opened 2025-07-19
Tutorials: test working downloads

adamjstewart:
Debugging #2853

### DIFF
--- a/.github/workflows/tutorials.yaml
+++ b/.github/workflows/tutorials.yaml
@@ -35,7 +35,7 @@ jobs:
       - name: List pip dependencies
         run: pip list
       - name: Run notebook checks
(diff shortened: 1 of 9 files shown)


adamjstewart:
Ready to go once #2853 is merged.

calebrob6:
What is the point here?

adamjstewart:
The point is that without this change, CI didn't fail even though downloads were broken. Now that #2853 is merged, CI is once again passing. Without this PR, we can't accurately tell whether or not each tutorial works regardless of the order in which they are run. Each tutorial was sharing resources downloaded by previous tutorials.

calebrob6:
Ahhh nice

calebrob6 (approved):

PR #2892, opened 2025-07-21
Restructure model docs

### LINKED ISSUE #2885: Add available weights for ResNet50 in the docs
### Issue

In the tutorial [about earthquake prediction](https://torchgeo.readthedocs.io/en/stable/tutorials/earthquake_detection.html) the `ResNet50_Weights.SENTINEL1_ALL_MOCO` weights are used. I've tried searching in the docs for a list of the other available weights, but I couldn't find them. With a quick inspection, I can see:

```python
from torchgeo.models import ResNet50_Weights

dir(ResNet50_Weights)

['FMOW_RGB_GASSL',
 'LANDSAT_ETM_SR_MOCO',
 'LANDSAT_ETM_SR_SIMCLR',
 'LANDSAT_ETM_TOA_MOCO',
 'LANDSAT_ETM_TOA_SIMCLR',
 'LANDSAT_OLI_SR_MOCO',
 'LANDSAT_OLI_SR_SIMCLR',
 'LANDSAT_OLI_TIRS_TOA_MOCO',
 'LANDSAT_OLI_TIRS_TOA_SIMCLR',
 'LANDSAT_TM_TOA_MOCO',
 'LANDSAT_TM_TOA_SIMCLR',
 'SENTINEL1_ALL_DECUR',
 'SENTINEL1_ALL_MOCO',
 'SENTINEL1_GRD_SOFTCON',
 'SENTINEL2_ALL_DECUR',
 'SENTINEL2_ALL_DINO',
 'SENTINEL2_ALL_MOCO',
 'SENTINEL2_ALL_SOFTCON',
 'SENTINEL2_MI_MS_SATLAS',
 'SENTINEL2_MI_RGB_SATLAS',
 'SENTINEL2_RGB_MOCO',
 'SENTINEL2_RGB_SECO',
 'SENTINEL2_SI_MS_SATLAS',
 'SENTINEL2_SI_RGB_SATLAS',
...
]
```

### Fix

Add list of available weights for ResNet50 in the docs

gcaria:
Closes #2885


gcaria:

docs/conf.py
```diff
@@ -17,16 +17,16 @@
...
+sys.path.insert(0, os.path.abspath(".."))
 
 import torchgeo
 
 # -- Project information -----------------------------------------------------
 
-project = 'torchgeo'
-copyright = '2021, Microsoft Corporation'
+project = "torchgeo"
+copyright = "2025, Microsoft Corporation"
```
adamjstewart: Would rather just remove the year, but we can do that in a different PR.

adamjstewart:
My goal is to get something a little closer to https://docs.pytorch.org/vision/main/models.html. So we can optionally include some discussion of how to choose/use models, links to all model API docs, and also tables with all pre-trained weights, all on the main page. Currently we only have links to model API docs.

gcaria:
@adamjstewart I've restructured the models docs page. Still working on a minimal set of models to get the structure right before adding all others. I'm puzzled by why I don’t see the “+” expand icon next to “Models” in the right sidebar, looking into it.

By the way, it seems not all pretrained weights are present in the csv files (e.g. I can't see `ResNet50_Weights.SENTINEL1_ALL_MOCO`)

gcaria:
@microsoft-github-policy-service rerun

gcaria:

adamjstewart:
So far I like it. I wonder if there's a way we can automatically generate the linked pages (e.g., ResNet, ViT) instead of having separate files for them. But we don't add these very often, so it isn't hard to make that part of the process.

Let me ping more people for review before we finalize the design. If this looks good, we can do the same thing for datasets, data modules, etc.

isaaccorley:
This is great and I think this is a significant improvement over the original docs.

calebrob6:
Definitely much better

docs/api/models.rst
```diff
@@ -202,12 +92,3 @@ Sentinel-2
    :file: weights/sentinel2.csv
 
 
-Atmospheric
```
adamjstewart, line 204: Did these accidentally get deleted by a rebase?

docs/api/models/aurora/index.rst
adamjstewart, line 1: I would like to reduce the nesting level so users don't have to click so many times. So the top level page (api/models.html) should list all model families, and clicking a link like Aurora should switch to a page documenting all Aurora classes/functions without having to click a second time. Even better would be a single `docs/api/models/aurora.rst` containing something like:
```rst
.. automodule:: torchgeo.models.aurora
   :members:
   :member-order: bysource
   :show-inheritance:
```
so that all public members are documented. We only want to document things in `torchgeo/models/__init__.py`, not all classes/functions. This may require passing a list to `:members:`.

docs/api/models.rst
```diff
@@ -1,157 +1,47 @@
...
-.. autoclass:: ViTSmall14_DINOv2_Weights
-.. autoclass:: ViTBase14_DINOv2_Weights
-
-YOLO
-^^^^
-
-.. autofunction:: yolo
-.. autoclass:: YOLO_Weights
-
-Utility Functions
```
adamjstewart, line 138: It looks like these utility functions got lost during the refactor?

docs/api/models/aurora/index.rst
gcaria: Done @adamjstewart

docs/api/models/aurora.rst
```diff
@@ -0,0 +1,9 @@
+Aurora
+======
+
+The Aurora family of models for geospatial computer vision tasks.
```
adamjstewart: Let's remove these sentences, the module itself has its own top-level docstring that is added.

docs/api/models/aurora.rst
```diff
@@ -0,0 +1,9 @@
...
+
+.. automodule:: torchgeo.models.aurora
```
adamjstewart: This is documenting things as `torchgeo.models.aurora.Aurora_Weights`. The preferred import method is `torchgeo.models.Aurora_Weights`. Does removing this:

```suggestion
.. automodule:: torchgeo.models
```
solve the problem, or does it complain that this module is documented multiple times?

docs/api/models/dofa.rst
```diff
@@ -0,0 +1,9 @@
+DOFA
+====
+
+The DOFA (Dense Optical Flow Attention) family of models for geospatial computer vision tasks.
```
adamjstewart: This isn't even close to what DOFA stands for lol

docs/api/models/yolo.rst
adamjstewart, line 1: Just a heads up, #3069 will remove all of our YOLO models for licensing reasons. So depending on the order in which these PRs get merged, you may need to resolve merge conflicts.

docs/api/models.rst
```diff
@@ -1,162 +1,54 @@
...
-.. autoclass:: ViTBase14_DINOv2_Weights
-
-YOLO
-^^^^
-
-.. autofunction:: yolo
-.. autoclass:: YOLO_Weights
+This section provides an overview of all models available in ``torchgeo.models``.
+
+Models
```
adamjstewart: Maybe we should refer to this as Model Architectures?

adamjstewart: Also, can we add some kind of description?

torchgeo/models/changestar.py
```diff
@@ -112,7 +112,7 @@ def __init__(
                 semantic segmentation model without semantic segmentation head.
             seg_classifier: semantic segmentation head, typically a convolutional layer
                 followed by an upsampling layer.
-            changemixin: :class:`torchgeo.models.ChangeMixin` module
+            changemixin: :class:`torchgeo.models.changestar.ChangeMixin` module
```
adamjstewart: This can be undone once the first comment is resolved

torchgeo/models/scale_mae.py
```diff
@@ -35,7 +35,7 @@ def get_2d_sincos_pos_embed_with_resolution(
         cls_token: Increase positional embedding size by 1 for class token.
 
     Returns:
-        pos_embed: Spatial resolution aware positional embeddings (Ph * Pw, D).
+        torch.Tensor: Spatial resolution aware positional embeddings (Ph * Pw, D).
```
adamjstewart: These should actually be like:

```suggestion
        Spatial resolution aware positional embeddings (Ph * Pw, D).
```
The type hint is documented in the function signature and doesn't need to be duplicated.

adamjstewart (changes requested):

PR #2893, opened 2025-07-22
Transforms Refactor

isaaccorley:
This PR has 2 goals:

1) Refactor all model inference transforms that operate on images in `torchgeo.models` to use `torchvision.transforms.v2` instead of `kornia.augmentations` primarily because kornia augmentations can't be `torch.exported`. Some benefits of exporting are described here https://github.com/microsoft/torchgeo/issues/2732.

2) An issue I had the pleasure of encountering is that the defaults for `K.Resize` and `T.Resize` are not equivalent. Mostly because Kornia is behind the times. To make them equivalent we should change all `K.Resize` or `K.RandomResizedCrop` throughout the codebase to use `K.Resize(..., align_corners=False, antialias=True)` so that they are interoperable with `T.Resize`.


tests/models/test_dofa.py
```diff
@@ -107,6 +107,14 @@ def test_transforms(self, weights: WeightsEnum) -> None:
         }
         weights.transforms(sample)
 
+    def test_export_transforms(self, weights: WeightsEnum) -> None:
+        """Test that the transforms have no graph breaks."""
+        torch = pytest.importorskip('torch', minversion='2.6.0')
```
isaaccorley, line 112: Skipping for the min tests because torchvision transforms weren't fully exportable until torch 2.6.0 but I don't want to bump the min to be 2.6 just to test exportability in recent versions.

tests/models/test_dofa.py
```diff
@@ -107,6 +107,14 @@ def test_transforms(self, weights: WeightsEnum) -> None:
...
+        torch._dynamo.reset()
```
isaaccorley: Resets need to happen because torch exporting caches results and the cache will OOM for many consecutive of the same transform being exported in tests.

torchgeo/models/dofa.py
```diff
@@ -387,12 +387,7 @@ def forward(self, x: Tensor, wavelengths: list[float]) -> Tensor:
 
 # https://github.com/zhu-xlab/DOFA/blob/master/normalize_dataset.py
 # Normalization is sensor-dependent and is therefore left out
-_dofa_transforms = K.AugmentationSequential(K.CenterCrop((224, 224)), data_keys=None)
-
-# https://github.com/pytorch/vision/pull/6883
-# https://github.com/pytorch/vision/pull/7107
-# Can be removed once torchvision>=0.15 is required
-Weights.__deepcopy__ = lambda *args, **kwargs: args[0]
+_dofa_transforms = nn.Sequential(T.CenterCrop((224, 224)))
```
isaaccorley, line 390: wrapping in `nn.Sequential` instead of `T.Compose` allows the transforms to be exportable

torchgeo/transforms/transforms.py
```diff
@@ -6,7 +6,6 @@
 from typing import Any
 
 import kornia.augmentation as K
```
rbavery: Does the _Extract_Patches transform in this file need a corresponding torchvision replacement?

torchgeo/models/earthloc.py
```diff
@@ -21,14 +20,10 @@
...
-_earthloc_transforms = K.AugmentationSequential(
-    K.Normalize(mean=torch.tensor(0.0), std=torch.tensor(255.0)),
-    K.Normalize(
-        mean=torch.tensor([0.485, 0.456, 0.406]),
-        std=torch.tensor([0.229, 0.224, 0.225]),
-    ),
-    K.Resize((320, 320)),
-    data_keys=None,
+_earthloc_transforms = nn.Sequential(
+    T.Normalize(mean=[0.0], std=[255.0], inplace=True),
```
isaaccorley, line 24: inplace here reduces memory from tensor copies

torchgeo/transforms/transforms.py
isaaccorley: It's only used for some datamodules not necessarily for inference. We can leave this for a separate PR

adamjstewart (commented):
My main concern is consistency. I would like to use a single transforms library everywhere. I don't really want to use Kornia for some parts and torchvision for others. We should decide which we want to use going forward. We've already invested quite some effort and documentation to Kornia support. Do we want to abandon them? Are there any limitations of torchvision we can't live with?

isaaccorley:
> My main concern is consistency. I would like to use a single transforms library everywhere. I don't really want to use Kornia for some parts and torchvision for others. We should decide which we want to use going forward. We've already invested quite some effort and documentation to Kornia support. Do we want to abandon them? Are there any limitations of torchvision we can't live with?

This is fine and I can refactor the augmentations to also use torchvision. On the other hand, kornia is unable to support exporting just due to the design of the library while torchvision does so I'm inclined to lean towards torchvision. I'm not aware of any other limitation that Kornia solves other than possibly more niche transforms.

adamjstewart:
Is `torchvision.transforms.v2` stable now?

isaaccorley:
> Is `torchvision.transforms.v2` stable now?

Yes and I've been using it on the job for their object detection / instance segmentation support for a couple years now and they haven't changed anything. 

I lean towards using it because all of their transforms are plain nn.Modules without graph breaks. Object detection transforms are quite good and would also integrate into our Object Detection / Instance Segmentation trainer fine.

isaaccorley:
I believe the only reason for the v2 naming is to keep compatibility with v1 which supported transforms directly on PIL images.

robmarkcole:
I will just repeat what I said on slack, which is that I support a transition to torchvision. I appreciate torch dropped support for features in the past but this feels too critical for that treatment.

adamjstewart:
This is a big change that requires input from more people than just me. Whether we choose Kornia or torchvision, we will likely need to devote quite a lot of time and upstream bugfixes. I want to make sure that this kind of change happens on the order of once every 5 years, which means we should really do more feature comparison and benchmarking like we did for rtree -> geopandas. I propose discussing this during a future technical steering committee (TSC) meeting.

isaaccorley:
> This is a big change that requires input from more people than just me. Whether we choose Kornia or torchvision, we will likely need to devote quite a lot of time and upstream bugfixes. I want to make sure that this kind of change happens on the order of once every 5 years, which means we should really do more feature comparison and benchmarking like we did for rtree -> geopandas. I propose discussing this during a future technical steering committee (TSC) meeting.

Regarding converting from kornia to torchvision, an alternative solution I figured out is to use Kornia's Lambda and partials with the Kornia functional API. I personally think this is a good tradeoff if you prefer not to pivot from kornia, however the code will look a bit ugly which is fine by me as long as it works.

```python
import torch
import torch.nn as nn
import kornia
from kornia.contrib import Lambda

x = torch.randn(1, 3, 256, 256)
transforms = nn.Sequential(
   Lambda(partial(kornia.geometry.transform.resize, size=(512, 512), interpolation="bilinear", align_corners=False, antialias=True))
)
program = torch.export.export(transforms, args=(x,))
```

isaaccorley:
Looks like `kornia.geometry.center_crop` has many graph breaks as well and would need to be completely redesigned unfortunately. So likely would need to use torchvision here in this case.

adamjstewart:
> I propose discussing this during a future technical steering committee (TSC) meeting.

Added this to the TSC agenda

tests/models/test_dofa.py
adamjstewart: Is there a non-private method for this?

tests/models/test_dofa.py
adamjstewart, line 1: We are currently testing that transforms can be exported for all models. But we aren't testing the actual model can be exported. Should we? I'm guessing most of our fancy multimodal FMs like DOFA can't be exported. This could be an interesting future project.

tests/models/test_dofa.py
```diff
@@ -107,6 +107,14 @@ def test_transforms(self, weights: WeightsEnum) -> None:
...
+        c = 4
+        inputs = (torch.randn(1, c, 224, 224, dtype=torch.float),)
```
adamjstewart, line 115: Out of curiosity, what purpose do the example `args` serve to `torch.export.export`?

adamjstewart (changes requested):
Before I dig any deeper to check for implementation differences between torchvision and kornia transforms, we have one major blocker.

Kornia transforms support dict inputs with the following keys: https://github.com/kornia/kornia/blob/v0.8.1/kornia/constants.py#L142

Torchvision does not seem to support this, at least with our current names or usage:
```python
import torch
import torch.nn as nn
import torchvision.transforms.v2 as T

sample = {
    'image': torch.rand(2, 3, 64, 64),
    'mask': torch.rand(2, 64, 64),
}

transform = nn.Sequential(T.CenterCrop((224, 224)))

sample = transform(sample)

print(sample['image'].shape)
print(sample['mask'].shape)
```
This shows that the image is cropped, but not the mask. How would we change this to augment the entire sample?

tests/models/test_dofa.py
isaaccorley: When you export you need to give an example input to do a forward pass to validate all the torch operations that occur on a forward pass.

tests/models/test_dofa.py
isaaccorley: That is going to be an effort for me in a follow-up PR. I wanted to start initially with the transforms first.

isaaccorley: Torch does support exporting with dict inputs so this does work but will need slightly more complex export tests!

isaaccorley:
> Before I dig any deeper to check for implementation differences between torchvision and kornia transforms, we have one major blocker.
> 
> Kornia transforms support dict inputs with the following keys: https://github.com/kornia/kornia/blob/v0.8.1/kornia/constants.py#L142
> 
> Torchvision does not seem to support this, at least with our current names or usage:
> 
> ```python
> import torch
> import torch.nn as nn
> import torchvision.transforms.v2 as T
> 
> sample = {
>     'image': torch.rand(2, 3, 64, 64),
>     'mask': torch.rand(2, 64, 64),
> }
> 
> transform = nn.Sequential(T.CenterCrop((224, 224)))
> 
> sample = transform(sample)
> 
> print(sample['image'].shape)
> print(sample['mask'].shape)
> ```
> 
> This shows that the image is cropped, but not the mask. How would we change this to augment the entire sample?

A couple things:
1) torchvision transforms v2 does support dict inputs, see their docs https://docs.pytorch.org/vision/0.15/auto_examples/plot_transforms_v2.html
2) the transforms in `torchgeo.models` currently only serve as inference-time transforms, not training augmentations so there's not really a need to transform both image and target. This is also how timm provides transforms (for inference).

tests/models/test_dofa.py
adamjstewart: Does this validation happen at export time or at import time or what?

adamjstewart:
1. Okay, so the difference is just the dict keys, we may have to rename things. If that is the case, I would really like to make a final decision if we want to convert everything to torchvision or keep some things in kornia. We would be renaming all keys, which would be a very backwards-incompatible change.
2. I would love to be able to use these not only in `predict_step` but also in `test_step`. One can imagine using @calebrob6's torchgeo-bench idea with these model weights and scraping the transforms from here. But you may be able to convince me that this is useless...

tests/models/test_dofa.py
isaaccorley: It's only at export time. Exporting basically creates a graph of torch operations. Because torch by default is eager you need to do a forward pass during export to create this graph also since dimensions can now bs dynamic.

isaaccorley:
1. Torchvision expands past this by using TVTensors which casts the tensor as a specific type of data rather than just depending on dict keys at which point it's deriving the type of data by the data itself rather than the keys. Imo assuming dtypes by dict keys makes too many assumptions anyway since it's not guaranteeing that it explicitly works with the data you are passing around https://docs.pytorch.org/vision/main/tv_tensors.html
2. My 2c is that the `torchgeo.models` transforms should be decoupled from the datamodules/ modules entirely because we are providing users the minimal set of operations to get any user data into the correct format for input. Making the transforms also transform a target variable is too dataset specific. In other words, test_step is dataset target dependent while predict_step is only dataset input dependent.

adamjstewart:
Interesting. TVTensors may also make some of the type hint stuff easier. I wonder if we can keep the same dict structure but use TVTensors such that our transforms work with both torchvision and kornia. If not, we may still have to make some backwards-incompatible changes here.

isaaccorley:
I actually do think this might be possible. TVTensors are just tensors with some extra methods and attrs for helping with certain data type operations. I'll see if Cursor can one-shot it.

adamjstewart:
> My 2c is that the `torchgeo.models` transforms should be decoupled from the datamodules/ modules entirely because we are providing users the minimal set of operations to get any user data into the correct format for input. Making the transforms also transform a target variable is too dataset specific. In other words, test_step is dataset target dependent while predict_step is only dataset input dependent.

Perhaps we should also decouple transforms from models 😉

tests/models/test_dofa.py
isaaccorley: yep changed to `torch.compiler.reset()`

adamjstewart: I saw that too, couldn't tell if it was the same or not. As long as we don't run OOM in CI, we definitely won't locally.

isaaccorley: Yeah this has never happened for me when exporting one-off models, but exporting the DOFAHuge models consume most of the runner's memory.

PR #2898, opened 2025-07-24
Add ConvLSTM model

MUYang99:
Add ConvLSTM model for satellite image time series processing: https://arxiv.org/pdf/1506.04214

### DIFF
--- a/torchgeo/models/convlstm.py
+++ b/torchgeo/models/convlstm.py
@@ -1,0 +1,251 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""Convolutional Long Short-Term Memory (ConvLSTM) model."""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+from typing import TypeVar
+
+import torch
+import torch.nn as nn
+
+T = TypeVar("T")
+
+
+class ConvLSTMCell(nn.Module):
+    """A single ConvLSTM cell module."""
+
+    def __init__(
+        self,
+        input_dim: int,
+        hidden_dim: int,
+        kernel_size: tuple[int, int],
+        bias: bool = True,
+    ) -> None:
+        """Initializes a ConvLSTMCell.
+
+        Args:
+            input_dim: Number of channels of input tensor.
+            hidden_dim: Number of channels of hidden state.
+            kernel_size: Size of the convolutional kernel.
+            bias: Whether or not to add the bias.
+        """
+        super().__init__()
+
+        self.input_dim = input_dim
+        self.hidden_dim = hidden_dim
+        self.kernel_size = kernel_size
+        self.padding = kernel_size[0] // 2, kernel_size[1] // 2
+        self.bias = bias
+
+        self.conv = nn.Conv2d(
+            in_channels=self.input_dim + self.hidden_dim,
+            out_channels=4 * self.hidden_dim,
+            kernel_size=self.kernel_size,
+            padding=self.padding,
+            bias=self.bias,
+        )
+
+    def forward(
+        self, input_tensor: torch.Tensor, cur_state: tuple[torch.Tensor, torch.Tensor]
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        """Forward pass of the ConvLSTMCell.
+
+        Args:
+            input_tensor: Tensor of shape (b, c, h, w).
+            cur_state: Tuple containing the current hidden and cell states.
+
+        Returns:
+            A tuple containing the next hidden and cell states.
+        """
+        h_cur, c_cur = cur_state
+        combined = torch.cat([input_tensor, h_cur], dim=1)
+        combined_conv = self.conv(combined)
+        cc_i, cc_f, cc_o, cc_g = torch.split(
+            combined_conv, self.hidden_dim, dim=1
+        )
+
+        i = torch.sigmoid(cc_i)
+        f = torch.sigmoid(cc_f)
+        o = torch.sigmoid(cc_o)
+        g = torch.tanh(cc_g)
+
+        c_next = f * c_cur + i * g
+        h_next = o * torch.tanh(c_next)
+
+        return h_next, c_next
+
+    def init_hidden(
+        self, batch_size: int, image_size: tuple[int, int]
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        """Initializes the hidden state.
+
+        Args:
+            batch_size: The batch size.
+            image_size: The height and width of the image.
+
+        Returns:
+            A tuple of tensors for the initial hidden and cell states.
+        """
+        height, width = image_size
+        device = self.conv.weight.device
+        return (
+            torch.zeros(batch_size, self.hidden_dim, height, width, device=device),
+            torch.zeros(batch_size, self.hidden_dim, height, width, device=device),
+        )
+
+
+class ConvLSTM(nn.Module):
+    """Convolutional LSTM model.
+
+    This model is a sequence-processing model that uses convolutional operations
+    within the LSTM cells. It is particularly useful for spatio-temporal data.
+    """
+
+    def __init__(
+        self,
+        input_dim: int,
+        hidden_dim: int | Sequence[int],
+        kernel_size: tuple[int, int] | Sequence[tuple[int, int]],
+        num_layers: int,
+        batch_first: bool = False,
+        bias: bool = True,
+        return_all_layers: bool = False,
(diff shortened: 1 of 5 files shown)


torchgeo/models/convlstm.py
```diff
@@ -0,0 +1,230 @@
...
+        """Forward pass of the ConvLSTM.
+
+        Args:
+            input_tensor: A 5-D Tensor of shape (t, b, c, h, w) or (b, t, c, h, w).
+            hidden_state: An optional initial hidden state.
+
+        Returns:
+            A tuple of two lists:
+            1. layer_output_list: List of Tensors of shape (b, t, c, h, w).
+            2. last_state_list: List of tuples of (h, c) for the last time step.
```
adamjstewart: Sphinx does not like this, maybe add a blank line before the start of the numbered list, play around with indentation, or just simplify the formatting a bit.

torchgeo/models/convlstm.py
```diff
@@ -0,0 +1,230 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
```
adamjstewart, line 5: This appears to be taken from https://github.com/ndrplz/ConvLSTM_pytorch? Need to add:
```python
# Copyright (c) 2017 Andrea Palazzi
```

torchgeo/models/convlstm.py
```diff
@@ -0,0 +1,230 @@
...
+"""Convolutional Long Short-Term Memory (ConvLSTM) model."""
+
+from __future__ import annotations
```
adamjstewart: I don't think this is needed

torchgeo/models/convlstm.py
```diff
@@ -0,0 +1,230 @@
...
+            torch.zeros(batch_size, self.hidden_dim, height, width, device=device),
+            torch.zeros(batch_size, self.hidden_dim, height, width, device=device),
+        )
+
+
+class ConvLSTM(nn.Module):
+    """Convolutional LSTM model.
+
+    This model is a sequence-processing model that uses convolutional operations
+    within the LSTM cells. It is particularly useful for spatio-temporal data.
```
adamjstewart, line 99: Add:
```rst

* https://arxiv.org/abs/1506.04214
```

torchgeo/models/convlstm.py
```diff
@@ -0,0 +1,236 @@
...
+        super().__init__()
+
+        # Normalize hidden_dim to a list of ints
+        if isinstance(hidden_dim, int):
+            self.hidden_dim: list[int] = [hidden_dim] * num_layers
+        else:
+            self.hidden_dim = list(hidden_dim)
+
+        # Normalize kernel_size to a list of tuples
+        if isinstance(kernel_size, tuple):
```
adamjstewart: `Sequence` includes both tuple and list. What happens when you have a tuple of tuples as input? Should we change the input type back to `tuple[int, int] | list[tuple[int, int]]`? Or check `if isinstance(kernel_size[0], int):`?

torchgeo/models/convlstm.py
```diff
@@ -0,0 +1,236 @@
...
+    If you use this model in your research, please cite the following paper:
+
+    * https://arxiv.org/abs/1506.04214
+    """
+
+    def __init__(
+        self,
+        input_dim: int,
+        hidden_dim: int | Sequence[int],
+        kernel_size: tuple[int, int] | Sequence[tuple[int, int]],
```
adamjstewart: https://github.com/ndrplz/ConvLSTM_pytorch/blob/master/convlstm.py only supports int. We only support tuple or seq[tuple]. All of these have merit.

I would like to support int, as I think that will be the most common use case. We can also support seq[int], tuple, seq[tuple] if you want, up to you. It's quite rare to see non-square kernels, but there are satellite image sources with non-square pixels.

torchgeo/models/convlstm.py
```diff
@@ -0,0 +1,236 @@
...
+
+        Args:
+            input_dim: Number of channels in the input.
+            hidden_dim: Number of hidden channels. Can be a single int (for all
+                layers) or a sequence of ints (one for each layer).
+            kernel_size: Size of the convolutional kernel. Can be a single tuple
+                (for all layers) or a sequence of tuples (one for each layer).
+            num_layers: Number of LSTM layers stacked on each other.
+            batch_first: If ``True``, then the input and output tensors are
+                provided as (b, t, c, h, w).
```
adamjstewart, line 131: TorchGeo is standardizing on B x T x C x H x W. At the very least, this should default to True, although I realize this makes us diverge from the reference implementation.

torchgeo/models/convlstm.py
```diff
@@ -0,0 +1,236 @@
...
+        if not self.return_all_layers:
+            layer_output_list = layer_output_list[-1:]
+            last_state_list = last_state_list[-1:]
+
+        return layer_output_list, last_state_list
+
+    def _init_hidden(
+        self, batch_size: int, image_size: tuple[int, int]
+    ) -> list[tuple[torch.Tensor, torch.Tensor]]:
+        """Initializes the hidden states for all layers."""
```
adamjstewart: Should we document the Args and Returns? I know it won't show up in the docs, but it might help for people viewing the source code.

torchgeo/models/convlstm.py
```diff
@@ -0,0 +1,236 @@
...
+        hidden_state: list[tuple[torch.Tensor, torch.Tensor]] | None = None,
+    ) -> tuple[list[torch.Tensor], list[tuple[torch.Tensor, torch.Tensor]]]:
+        """Forward pass of the ConvLSTM.
+
+        Args:
+            input_tensor: A 5-D Tensor of shape (t, b, c, h, w) or (b, t, c, h, w).
+            hidden_state: An optional initial hidden state.
+
+        Returns:
+            tuple:
```
adamjstewart: No need to document the return type here, it's already pulled from the function signature type hints

adamjstewart (commented):
Only very minor comments remaining

torchgeo/models/convlstm.py
adamjstewart: Looks good now: https://torchgeo--2898.org.readthedocs.build/en/2898/api/models.html#torchgeo.models.ConvLSTM.forward

torchgeo/models/convlstm.py
```diff
@@ -0,0 +1,236 @@
...
+
+class ConvLSTM(nn.Module):
+    """Convolutional LSTM model.
+
+    This model is a sequence-processing model that uses convolutional operations
+    within the LSTM cells. It is particularly useful for spatio-temporal data.
+
+    If you use this model in your research, please cite the following paper:
+
+    * https://arxiv.org/abs/1506.04214
```
adamjstewart, line 103: Also add:
```rst
.. versionadded:: 0.8
```

torchgeo/models/convlstm.py
```diff
@@ -0,0 +1,245 @@
...
+            self.hidden_dim: list[int] = [hidden_dim] * num_layers
+        else:
+            self.hidden_dim = list(hidden_dim)
+
+        # Normalize kernel_size to a list of tuples
+        if isinstance(kernel_size, int):
+            self.kernel_size: list[tuple[int, int]] = [
+                (kernel_size, kernel_size)
+            ] * num_layers
+        elif isinstance(kernel_size, tuple):
```
adamjstewart: I think this still suffers from the same bug, a tuple of tuples will turn into a list of tuples of tuples. You either need:
```suggestion
        elif isinstance(kernel_size[0], int):
```
or to change the type to `list[tuple[int, int]]` (Sequence includes both tuple and list). However, the type hints also allow `[int, tuple[int, int]]` (a mixed list of ints and tuples), so this isn't reliable either. So either change the types or make this more powerful.

torchgeo/models/convlstm.py
```diff
@@ -0,0 +1,255 @@
...
+        if isinstance(hidden_dim, int):
+            self.hidden_dim: list[int] = [hidden_dim] * num_layers
+        else:
+            self.hidden_dim = list(hidden_dim)
+
+        # Normalize kernel_size to a list of tuples
+        self.kernel_size: list[tuple[int, int]] = []
+
+        # Convert to list first to handle both tuples and lists
+        if isinstance(kernel_size, int | tuple):
```
adamjstewart, line 145: Still broken, an input like `((1, 2), (3, 4))` will be transformed into `[((1, 2), (3, 4)), ((1, 2), (3, 4))]`. Let me push a commit to try to fix this.

PR #2901, opened 2025-07-29
aws.bat and azcopy.bat python command

sfalkena:
By default, the aws.bat and azcopy.bat in the test folder only run the `python3` command. On my Windows system with conda environment, the `python3` command is not available by default. I have added a simple line which tries both the `python3` and `python` command.

### DIFF
--- a/tests/datasets/aws.bat
+++ b/tests/datasets/aws.bat
@@ -3,7 +3,4 @@ REM Licensed under the MIT License.
 
 @ECHO OFF
 
-python3 tests\datasets\aws.py %*
-IF %ERRORLEVEL% NEQ 0 (
-    python tests\datasets\aws.py %*
-)
+python tests\datasets\aws.py %*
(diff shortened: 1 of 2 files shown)


adamjstewart:
Interesting. On UNIX, the `python` command is not guaranteed to exist, and scripts are advised to use `python3` ([PEP 394](https://peps.python.org/pep-0394/)). Are there any Windows systems where `python` does not exist? Should we just always use `python`? [PEP 397](https://peps.python.org/pep-0397/) isn't completely clear, but I'm leaning towards just removing the 3 on Windows and always calling `python`.

sfalkena:
Indeed PEP394 mentions that UNIX does not require the `python` command. Why would you want to hardcode it per OS, and not try `python3` first (implicitly for UNIX) and then try `python` if that fails? In that case it is OS agnostic.

adamjstewart:
`aws` (symlinked to `aws.py`) is only used on UNIX, `aws.bat` (calls `aws.py`) is only used on Windows. It isn't possible to have a single platform-agnostic wrapper around a shell command as far as I know.

adamjstewart:
Want to update this PR to always use `python` on Windows instead of `python3`?

adamjstewart (approved):

PR #2903, opened 2025-07-29
Store valid data footprint into RasterDataset.geometries

### LINKED ISSUE #1330: How to avoid nodata-only patches

adriantre:
Started over since #1881 was so outdated :)

This PR includes the basic functionality for storing valid footprint (meaning excluding no-data regions) in the RasterDataset index, instead of just the bounds of the raster.

Closes #1330


torchgeo/datasets/geo.py
```diff
@@ -359,6 +360,12 @@ class RasterDataset(GeoDataset):
     #: Color map for the dataset, used for plotting
     cmap: ClassVar[dict[int, tuple[int, int, int, int]]] = {}
 
+    #: Whether to store the valid data footprint or the raster bounds
+    use_valid_footprint: bool = True
```
adamjstewart: We should discuss whether this should be a class attribute, an `__init__` parameter, or always True. Do you have any metrics on how slow this is for, let's say, 100 Sentinel-2 tiles?

torchgeo/datasets/geo.py
```diff
@@ -359,6 +360,12 @@ class RasterDataset(GeoDataset):
...
+
+    #: nodata value
+    nodata_value: int | float = 0
```
adamjstewart: int is a subset of float, no need to include both

torchgeo/datasets/utils.py
```diff
@@ -758,3 +760,52 @@ def convert_poly_coords(
...
+    Args:
+        masks: raster mask, like rasterio.io.DatasetReader.read_masks()
+        transform: affine transform for the raster
+        raster_width: width of the raster
+        raster_resolution_x: resolution of the raster in the x direction (meters)
+
+    Returns:
+        Polygon representing the valid data footprint of the raster
+    """
+    # Read valid/nodata-mask. Could choose only the first bands mask for peed-up.
```
adamjstewart:
```suggestion
    # Read valid/nodata-mask. Could choose only the first bands mask for speed-up.
```

torchgeo/datasets/utils.py
```diff
@@ -758,3 +760,52 @@ def convert_poly_coords(
...
+    # Currently, supports spatial shifts between the bands, and include all of this
+    # as one combined mask.
```
adamjstewart: I think we discussed this on Slack. Typically these shifts are on the order of a few pixels at most. I'm personally okay with ignoring this and only looking at a single band just to make things faster.

torchgeo/datasets/utils.py
```diff
@@ -758,3 +760,52 @@ def convert_poly_coords(
...
+    # to one multipolygon later
+    geoms = [g for g, v in shapes(sieved_mask, transform=transform) if v > 0]
+    # Create a Shapely Polygon object(s)
+    vector_footprint = MultiPolygon(
+        [Polygon(feature['coordinates'][0]) for feature in geoms]
+    )
+    # The resulting polygon(s) is very staggered/pixelated,
+    # which could result in thousands of corners.
+    # The valid data footprints are usually very simple (3-5 corners)
+    # Simplifying each side to length 1/5 of the raster size (assumes crs is meters)
```
adamjstewart: Lots of simplifying assumptions built into this. I'm okay with this for the sake of complexity, but we'll have to spend some time playing devil's advocate and testing this.

adamjstewart (commented):
Pretty cool. Don't fully understand all of the details, but it's definitely something we want now that we are using geopandas.

Note that someone still needs to submit a PR to ensure that the GeoSamplers actually use the Polygon bounds instead of the surrounding bounding box.

Also, @isaaccorley was planning on extending this to VectorDataset so that we read directly using geopandas instead of fiona. We're also planning on caching these indices so that we don't have to recompute the bounds every time.

torchgeo/datasets/geo.py
adriantre: I set it to always run in my latest commit. I'll try to benchmark this :)
I made it skip this step if all pixels are valid though, based on the mask.

torchgeo/datasets/geo.py
adriantre: True, I always forget.

I made this default to None, and relying on the raster to be "fully defined", see [this description](https://rasterio.readthedocs.io/en/latest/topics/masks.html#dataset-masks).

The user can then set this parameter if they know the actual nodata-value used in the raster, such that point 3. is reached in `src.dataset_mask()`.

torchgeo/datasets/utils.py
adriantre: Yeah, as I looked at this with fresh eyes I came across `src.dataset_mask()` that does this OR-combination across bands for us. _Could_ also be faster than my implementation.

torchgeo/datasets/utils.py
adriantre: Please do!

adriantre:
> Pretty cool. Don't fully understand all of the details, but it's definitely something we want now that we are using geopandas.

I was so happy when geopandas was introduced :)

> Note that someone still needs to submit a PR to ensure that the GeoSamplers actually use the Polygon bounds instead of the surrounding bounding box.

I'll take a look now that geopandas should make this much more straightforward than last time I did.

> Also, @isaaccorley was planning on extending this to VectorDataset so that we read directly using geopandas instead of fiona. We're also planning on caching these indices so that we don't have to recompute the bounds every time.

I see that VectorDataset stores the bounds of the data to `.geometries`. I understand that the geometry might be complex and big, but for sampling postives and negatives it would be great to have the (simplified) actual data 😅 before `__get_item__` that is.

torchgeo/datasets/geo.py
adriantre: So my findings are that it is extremely slow when relying on nodata-value. I tested this on Sentinel-2 by specifying `nodata_value=0`. This is because it have to read all pixels to find the mask. `src.dataset_mask()` will first look for a mask-file (external or internal) which should be much quicker though. Some optimisation is needed regardless. In general, we should probably extract the vector before reprojecting, then reproject the vector, as reprojecting will add no-data in the corners.

**main branch:**
Benchmark for 10 paths:
Time: 0.029s, Memory: 12.83MB

**My branch:**
Benchmark for 10 paths:
Time: 43.982s, Memory: 1336.36MB

For Sentinel-2, I would get the vector mask directly like this, though...
```python
path = "data/alberta_s2/s2/S2B_MSIL1C_20240711T182919_N0510_R027_T11UQS_20240711T232727.SAFE/MTD_MSIL1C.xml"
    print(src.tags()["FOOTPRINT"])
```

Anyhow, I benchmarked it by simulating many files from one Sentinel-2 image, I think it is a pretty neat way to save disk space.
<details markdown="1"><summary>Code</summary>

```python
import os
import time
from pathlib import Path
from tempfile import TemporaryDirectory

import psutil

# ---- Utility methods ----
def hardlink_dir(src: Path, dst: Path):
    """Recursively hard-link all files from src to dst."""
    for root, dirs, files in os.walk(src):
        rel_path = os.path.relpath(root, src)
        dst_root = dst / rel_path
        dst_root.mkdir(parents=True, exist_ok=True)

        for f in files:
            src_file = Path(root) / f
            dst_file = dst_root / f
            os.link(src_file, dst_file)  # hard link


def duplicate_sample_dirs(src_dir: str, output_dir: str, num_copies: int):
    """Create N sample_* directories in output_dir, each with a hardlinked copy of src_dir."""
    src_dir = Path(src_dir).resolve()
    output_dir = Path(output_dir).resolve()
    output_dir.mkdir(parents=True, exist_ok=True)

    for i in range(num_copies):
        sample_dir = output_dir / f"sample_{i}"
        hardlink_dir(src_dir, sample_dir)


def track_time_and_memory(func, *args, **kwargs):
    process = psutil.Process(os.getpid())
    mem_before = process.memory_info().rss
    start = time.time()

    func(*args, **kwargs)

    elapsed = time.time() - start
    mem_after = process.memory_info().rss
    mem_used = (mem_after - mem_before) / (1024 * 1024)  # MB
    return elapsed, mem_used


# ---- Run benchmark ----


class Sentinel2WithNodataSpecified(Sentinel2):
    nodata_value = 0


def my_method(root):
    _ = Sentinel2WithNodataSpecified(root, bands=Sentinel2.rgb_bands)


def benchmark():
    src_input_dir = "torchgeo/data/alberta_s2/s2/S2B_MSIL1C_20240711T182919_N0510_R027_T11UQS_20240711T232727.SAFE"
    num_paths = 10
    with TemporaryDirectory() as root_dir:
        duplicate_sample_dirs(src_input_dir, root_dir, num_paths)

        time_a, mem_a = track_time_and_memory(my_method, root_dir)

        print(f"\nBenchmark for {num_paths} paths:")
        print(f"Time: {time_a:.3f}s, Memory: {mem_a:.2f}MB")


if __name__ == "__main__":
    benchmark()
```

</details>

adamjstewart: Oof, that is quite slow...

adamjstewart: What's the speed for `src.dataset_mask()`? How often does such a mask exist?

adriantre: With the same benchmark, but opening the files "th eintended way" that utilise Sentinel2 driver of GDAL we get the masks even faster than torchgeo main branch. There are mask-files in the .SAFE products, which I assume are used automatically through the driver when `src.dataset_mask()` is called. So it should be much faster when a mask is present.

Many satellite platforms have such drivers, btw (like [Radarsat CM](https://gdal.org/en/stable/drivers/raster/rcm.html)). 

```python
def rasterio_s2_driver(root):
    pattern = os.path.join(root, "**", "MTD_MSIL1C.xml")
    paths = [os.path.abspath(path) for path in glob.glob(pattern, recursive=True)]
    for path in paths:
            _ = get_valid_footprint_from_datasource(src)


def benchmark(func):
    src_input_dir = "torchgeo/data/alberta_s2/s2/S2B_MSIL1C_20240711T182919_N0510_R027_T11UQS_20240711T232727.SAFE"
    num_paths = 100
    with TemporaryDirectory() as root_dir:
        duplicate_sample_dirs(src_input_dir, root_dir, num_paths)

        time_a, mem_a = track_time_and_memory(func, root_dir)

        print(f"\nBenchmark for {num_paths} paths:")
        print(f"Time: {time_a:.3f}s, Memory: {mem_a:.2f}MB")


if __name__ == "__main__":
    benchmark(rasterio_s2_driver)
```

Bare rasterio with S2-driver:
Benchmark for 100 paths:
Time: 0.074s, Memory: 4.20MB

Torchgeo main branch:
Benchmark for 100 paths:
Time: 0.109s, Memory: 16.56MB

adamjstewart: Hmm, this feels pretty hard-coded and unique for every dataset. I was hoping for a solution that works for most RasterDatasets out of the box...

adriantre: This was a low-effort test to see the speed of getting footprint when mask-files are present. So I assume we will experience similar speeds for datasets where such mask-files exist.

For those that does not we could [write these mask-files](https://rasterio.readthedocs.io/en/latest/api/rasterio.io.html#rasterio.io.BufferedDatasetWriter.write_mask) the first time the file is used so that the next read is fast. This would add a new file alongside the rasters. What do you think about that approach?

adamjstewart: I like the idea of caching these mask files (when possible, we won't always have access). But if the first time is still 1000x slower, that's pretty hard to adopt.

adamjstewart:
I'm currently leaning towards supporting this on a case-by-case basis only when the appropriate metadata is already available. It's prohibitively slow to calculate it on-the-fly. So we could start with checking for the Sentinel-2 metadata you found and add more satellites as we notice this.

torchgeo/datasets/geo.py
adriantre: Could the first step be to add this static method to RasterDataset

```python
@staticmethod
def get_footprint_from_datasource(src: DatasetReader | WarpedVRT) -> MultiPolygon | Polygon:
    return shapely.box(*src.bounds)
```

to let users override it, for example Sentinel2 direct footprint reading.

```python
class Sentinel2WithOverriddenGetFootprint(Sentinel2):
    @staticmethod
    def get_footprint_from_datasource(dataset) -> MultiPolygon | Polygon:
        if isinstance(dataset, WarpedVRT):
            filepath = dataset.src_dataset.name
        else:
            filepath = dataset.name
        metadata_path = filepath.split("GRANULE")[0] + "MTD_MSIL1C.xml"
        with rasterio.open(metadata_path) as metadata_src:
            true_footprint = shapely.wkt.loads(metadata_src.tags()["FOOTPRINT"])
        transformer = pyproj.Transformer.from_crs(
            pyproj.CRS("EPSG:4326"), dataset.crs, always_xy=True
        ).transform
        return transform(transformer, true_footprint)
```

or they can choose to override using the might-be-slow-the-first-time `utils.extract_valid_footprint_polygon` or their own implementation?

adriantre: Posted this before I read your latest comment below. Is this suggestion in line with what you were thinking?

adamjstewart: Yes, this is inline with what I'm thinking. Note that `@staticmethod` probably won't work as we need the ability to reproject based on `self.crs`.

adriantre: Yeah, for now the above implementation relies on the WarpedVRT to be in the correct CRS :)

I will push a suggestion soon, should I remove or include `utils.extract_valid_footprint_polygon`? (without usage)

adamjstewart: Good question. Maybe keep it in `utils.py` if you already have tests for it.

torchgeo/datasets/geo.py
```diff
@@ -359,6 +360,11 @@ class RasterDataset(GeoDataset):
     #: Color map for the dataset, used for plotting
     cmap: ClassVar[dict[int, tuple[int, int, int, int]]] = {}
 
+    #: Nodata value used in the raster for pixels outside the satellite acquisition
+    #: Optionally set this if the raster lack nodata-mask and nodata value
+    #: and you want need it to override `RasterDataset._footprint_from_datasource()`
+    nodata_value: float | None = None
```
adriantre: I left this in here. _if_ someone wants to use this in combination with overriding `_footprint_from_datasource` to call `datasets.utils.get_valid_footprint_from_datasource`, this is passed through for them.

adamjstewart:
0.8 will almost certainly be 1.0, but let's use `versionadded:: 0.8` for now just in case.

adriantre:
I'm done with this part I think, and will start a new PR for samplers to use it. If I understand them correctly I will just have to make [tile_to_chips](https://github.com/torchgeo/torchgeo/blob/b246b646ed701ef034a9cf9c2f62754c77e38b51/torchgeo/samplers/single.py#L155) respect the real geometries instead of the total bounds?

tests/datasets/test_geo.py
```diff
@@ -7,6 +7,7 @@
 from collections.abc import Iterable, Sequence
 from pathlib import Path
 from typing import Any
+from unittest.mock import MagicMock, patch
```
adamjstewart: I would rather use pytest MonkeyPatch

tests/datasets/test_geo.py
```diff
@@ -483,6 +496,26 @@ def test_change_res(self) -> None:
         assert ds.res == (10.0, 10.0)
         ds.res = 20.0
 
+    def test_nodata_value_is_passed_on(self) -> None:
+        root = os.path.join('tests', 'data', 'raster', 'res_2-2_epsg_32631')
+
+        RasterDatasetNodata = RasterDataset
+        RasterDatasetNodata.nodata_value = 0.0
```
adamjstewart: This should be monkeypatched, they're both pointers to the same variable.

tests/datasets/test_geo.py
```diff
@@ -483,6 +496,26 @@ def test_change_res(self) -> None:
         assert ds.res == (10.0, 10.0)
         ds.res = 20.0
 
+    def test_nodata_value_is_passed_on(self) -> None:
```
adamjstewart: I don't fully understand this test, the mocking is too deep. What are you actually trying to test here?

tests/datasets/test_sentinel.py
```diff
@@ -149,3 +155,36 @@ def test_invalid_query(self, dataset: Sentinel2) -> None:
...
+
+    @pytest.mark.parametrize('dataset_spec', [DatasetReader, WarpedVRT])
+    def test_footprint_from_datasource_metadata_file(
+        self, dataset: Sentinel2, dataset_spec: type[DatasetReader] | type[WarpedVRT]
+    ) -> None:
+        # WKT in EPSG:4326
+        metadata_footprint_wkt = dataset.index.geometry.to_crs(4326).values[0].wkt
+        footprint_wgs84 = shapely.wkt.loads(metadata_footprint_wkt)
+
+        # Fake dataset
```
adamjstewart: Can we test with real data instead? Just update the `tests/data/sentinel2/data.py` file to save the metadata you need.

tests/datasets/test_sentinel.py
```diff
@@ -149,3 +155,36 @@ def test_invalid_query(self, dataset: Sentinel2) -> None:
 
     def test_float_res(self, dataset: Sentinel2) -> None:
         Sentinel2(dataset.paths, res=10.0, bands=dataset.bands)
+
+    @pytest.mark.parametrize('dataset_spec', [DatasetReader, WarpedVRT])
```
adamjstewart: Can't we test data that does and doesn't need reprojection instead of these dataset_spec things? It isn't clear when a real user would hit this, since the code being tested isn't something a real user would use.

tests/datasets/test_utils.py
```diff
@@ -537,3 +542,101 @@ def test_azcopy(tmp_path: Path, azcopy: Executable) -> None:
 def test_which() -> None:
     with pytest.raises(DependencyNotFoundError, match='foo is not installed'):
         which('foo')
+
+
+def create_test_raster_file(
```
adamjstewart: Code used during testing usually belongs as a fixture, either in here or in `conftests.py`

torchgeo/datasets/geo.py
```diff
@@ -451,10 +457,14 @@ def __init__(
                         if crs is None:
                             crs = src.crs
 
-                        with WarpedVRT(src, crs=crs) as vrt:
-                            geometries.append(shapely.box(*vrt.bounds))
+                        vrt_options = {'crs': crs}
+                        if self.nodata_value is not None:
+                            vrt_options['nodata'] = self.nodata_value
+                        with WarpedVRT(src, **vrt_options) as vrt:
```
adamjstewart: Just a heads up, we're possibly getting rid of WarpedVRT in #2945

torchgeo/datasets/sentinel.py
```diff
@@ -415,3 +424,47 @@ def plot(
...
+            filepath = dataset.src_dataset.name
+        else:
+            filepath = dataset.name
+        metadata_path = filepath.split('GRANULE')[0] + 'MTD_MSIL1C.xml'
+        if not os.path.exists(metadata_path):
+            warnings.warn(
+                f'Metadata file not found in Sentinel-2 path: {metadata_path}. '
+                'This is required to extract true footprint for this dataset.'
+                'Defaulting to using raster bounds.'
+            )
```
adamjstewart: I'm worried that this will be printed hundreds of times for most users. Maybe this should not have a warning message?

torchgeo/datasets/sentinel.py
```diff
@@ -415,3 +424,47 @@ def plot(
...
+            return super()._footprint_from_datasource(dataset)
+        with rasterio.open(metadata_path) as metadata_src:
+            true_footprint = shapely.wkt.loads(metadata_src.tags()['FOOTPRINT'])
+        transformer = pyproj.Transformer.from_crs(
+            pyproj.CRS('EPSG:4326'), dataset.crs, always_xy=True
+        ).transform
```
adamjstewart: Why 4326? Is this metadata file always guaranteed to be 4326?

torchgeo/datasets/sentinel.py
```diff
@@ -415,3 +424,47 @@ def plot(
...
+                or a :class:`rasterio.vrt.WarpedVRT`, from which the footprint will
+                be derived.
+
+        Returns:
+            A :class:`shapely.geometry.Polygon` or
+            :class:`shapely.geometry.MultiPolygon` representing the footprint in
+            the dataset's CRS.
+
+        Raises:
+            KeyError: If the existing metadata file does not contain a ``FOOTPRINT`` tag.
```
adamjstewart: Should we use a try-except instead?

torchgeo/datasets/sentinel.py
```diff
@@ -415,3 +424,47 @@ def plot(
...
+
+        .. versionadded:: 0.8
+        """
+        if hasattr(dataset, 'src_dataset'):
+            # When dataset is a WarpedVRT
+            filepath = dataset.src_dataset.name
+        else:
+            filepath = dataset.name
```
adamjstewart: Instead of passing the DatasetReader, why not pass the filename as an argument? Does this work for merged files or only individual files?

torchgeo/datasets/utils.py
```diff
@@ -758,3 +761,131 @@ def convert_poly_coords(
...
+    geoms = [g for g, v in shapes(sieved_mask, transform=transform) if v > 0]
+
+    # Create a Shapely Polygon object(s)
+    vector_footprint = MultiPolygon(
+        [Polygon(feature['coordinates'][0]) for feature in geoms]
+    )
+    # The resulting polygon is very staggered/pixelated,
+    # which could result in thousands of corners.
+    # The valid data footprints are usually very simple (3-5 corners)
+    # Simplifying each side to length 1/5 of the raster size (assumes crs is meters)
```
adamjstewart: > (assumes crs is meters)

This is not a good assumption

torchgeo/datasets/utils.py
```diff
@@ -758,3 +761,131 @@ def convert_poly_coords(
...
+    return vector_footprint.simplify(max_distance_of_polygon_length)
+
+
+def get_valid_footprint_from_datasource(
+    src: DatasetReader | WarpedVRT,
+) -> MultiPolygon | Polygon:
+    """Compute the valid data footprint of a raster dataset.
+
+    NB! If your dataset rely on nodata-value to create the masks, this might
+    add a lot of overhead. Consider writing nodata masks to file.
```
adamjstewart:
```suggestion
    .. note::
    
       If your dataset rely on nodata-value to create the masks, this might
       add a lot of overhead. Consider writing nodata masks to file.
```

torchgeo/datasets/utils.py
```diff
@@ -758,3 +761,131 @@ def convert_poly_coords(
...
+
+    This function analyzes the raster's mask band to determine the spatial extent
+    of valid (non-NoData) pixels, returning the result as a `Polygon` or `MultiPolygon`
+    in the dataset's coordinate reference system.
+
+    Workflow:
```
adamjstewart: I don't think this will be formatted correctly, you need an empty line before and after lists

adamjstewart (changes requested):
Code mostly looks okay, just a bit complicated. However, the tests are far too complicated. I want to test with real data and greatly simplify the test mocking.

adamjstewart:
> will start a new PR for samplers to use it. If I understand them correctly I will just have to make tile_to_chips respect the real geometries instead of the total bounds?

It's a bit more complicated than that. `tile_to_chips` just calculates the approximate number of chips one could get out of a tile. This should probably be replaced by some simple area math (compute area of Polygon using shapely, divide by area of patch). 

We also need to update `get_random_bounding_box` to use the real bounds. I'm thinking something like [sample_points](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoSeries.sample_points.html). However, sampling the centroid will result in patches slightly outside the bounds of the Polygon. Ideally we would use some kind of custom `method` that weights things by distance from the edge.

However, @sfalkena is working on a complete rewrite of the samplers, so the 3 of us should coordinate the order in which these PRs happen to avoid too many merge conflicts. We're also thinking about precomputing all sample points at the beginning of each epoch so that users can visualize or weight these locations.

tests/datasets/test_sentinel.py
adriantre: I did it like this just for test coverage.

For the current state of this PR, only the WarpedVRT is relevant/called. ([Here in RasterDataset](https://github.com/torchgeo/torchgeo/blob/bd09c6fb485ea73bafa4ffb36dbc78684a699c77/torchgeo/datasets/geo.py#L466)). But I'm thinking it might be better/faster to extract the footprint before reprojecting the raster to the common crs, and reproject the resulting vector instead. Should be less nodata regions this way.

Thats why I kept both options in the test and in `Sentinel2._footprint_from_datasource`. I can simplify to either one once we settle on where `_footprint_from_datasource` is called.

tests/datasets/test_sentinel.py
adriantre: Thanks, I will give it a go

torchgeo/datasets/geo.py
adriantre: Cool! I'm not too familiar with xarray, but it looks like creating data masks is possible, maybe also it supports writing the masks for caching purposes.

I guess this is what replace WarpedVRT: `src = src.rio.reproject(crs)`

torchgeo/datasets/sentinel.py
adriantre: Yes, it always is 💪

torchgeo/datasets/sentinel.py
adriantre: For this specific override, we could pass the filename (and possibly the original CRS to avoid opening the same dataset twice).
But the parent method `RasterDataset._footprint_from_datasource` needs the `dataset.bounds` instead. Could pass filename to it too, but that feels like unnecessary duplication of `rasterio.open`.

This override requires individual file(s) and that the Sentinel2 image follows the `.SAFE` format/file structure, specifically that it has the `MTD_MSIL1C.xml` file in the folder `GRANULE`, which is part of the file name.

If someone, for some reason, has Sentinel-2 rasters in a different format, they have lost this information, unless they have stored it in their rasters somehow (e.g., a nodata mask).

torchgeo/datasets/utils.py
adriantre: Yeah, I noted it down since I thought it might not be. If the CRS (of the Raster!) is in degrees, the resolution is a tiny number, and this tolerance will be huge, resulting in a very simplified polygon. How often is that the case for rasters?

adriantre:
> It's a bit more complicated than that. `tile_to_chips` just calculates the approximate number of chips one could get out of a tile.

Yeah, I later saw that it did my assumption about what it did was wrong.

> I'm thinking something like [sample_points](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoSeries.sample_points.html).

I had the same idea, for storing vector data in the index. Vectors might get huge, so sample points as a proxy for the actual was what I thought too. Random sampler is a bit tricky since there's no pre-defined grid where we can precompute the intersection of labels and raster footprint. We'll figure it out.

> However, @sfalkena is working on a complete rewrite of the samplers, so the 3 of us should coordinate the order in which these PRs happen to avoid too many merge conflicts. We're also thinking about precomputing all sample points at the beginning of each epoch so that users can visualize or weight these locations.

Love it. I'm keen on joining the discussion.

tests/datasets/test_geo.py
adriantre: I wrote this test to ensure that nodata_value is received where it is (currently) used, inside the WarpedVRT. My idea was that this would catch if someone changes the code in the future and forgets to handle nodata_value, tricking the user into thinking it is used when it is not.

For test-coverage the code could be ran in any test, but it would not actually check the logic, only catch runtime errors.

Do you have other ideas of how to test this?

tests/datasets/test_sentinel.py
adriantre: #2991

I put it in a separate PR to not flood this one. Will update the tests when-if it is merged.

tests/datasets/test_geo.py
adriantre: The other option ofc is to check that the resulting footprint is correct, meaning the nodata was passed and used correctly. https://github.com/torchgeo/torchgeo/pull/2991 will make that much easier, and I can get rid of the patching.

adamjstewart:
Would like to get back to this since I'm working on samplers right now. USGS EarthExplorer Landsat data also has an MTL file containing the true footprint of the image. So it isn't just Sentinel that could benefit.

What's the easiest way to revive this? I'm happy to add MTL test data for Landsat or push changes to your Sentinel metadata PR. For this PR itself, I want to see if we can somehow make this simpler for users, similar in magnitude to your PR that refactored reprojection into a separate helper function.

torchgeo/datasets/geo.py
adamjstewart: This could probably be a separate PR. In `XarrayDataset`, we hard-code to `nodata=0`, but it should probably be configurable in both. I remember some warnings when nodata was not set in the file but nodata pixels exist.

torchgeo/datasets/utils.py
```diff
@@ -758,3 +761,133 @@ def convert_poly_coords(
         ],
     )
     return xformed_shape
+
+
+def clean_binary_mask(
```
adamjstewart: I think these changes (calculating a footprint when no metadata exists) should be moved to a separate PR. Given how slow they are, we're unlikely to use them most of the time anyway. Don't want these changes to hold up this PR.

adriantre: Moved it to here https://github.com/torchgeo/torchgeo/pull/3663
- [X] Will clean up this PR soon :)

torchgeo/datasets/geo.py
adriantre: Yes, this is the case for Sentinel-2 😎

torchgeo/datasets/sentinel.py
adriantre: Will now fallback to using bounds, as when metadata file does not exist.

tests/datasets/test_sentinel.py
adriantre, line 1: Might remove the mock when #2991 is merged. 
(Might need to test for missing metadata file, though, for coverage)

torchgeo/datasets/sentinel.py
adriantre: To hedge on my understanding of your question on merged files:

It runs in init, thus not at a point where reading from overlapping tiles / files.

It does not need to care about multiple band-files, as init specifically handles one bandfile only
`self.filename_regex = self.filename_regex.format(self.resolutions[bands[0]])`.

torchgeo/datasets/utils.py
adriantre: Fixed in #3663

torchgeo/datasets/geo.py
adriantre: Enabling the user to set nodata ties well with the ability to override `_footprint_from_datasource`, imo. Specifically, it enables the usage of rasterio.read_masks in the override.

I had a look at XarrayDataset. The hard-coding of `nodata=0` seems to be related to filling in the blanks between the merged files, not about how to respect the source datasets' nodata value. 

Unrelated, but what you _could/should_ do in that scenario is to specify `nodata=src.rio.nodata or 0` to keep the missing pieces in line with the src datasets' definition of nodata.

.pre-commit-config.yaml
```diff
@@ -20,6 +20,7 @@ repos:
         language: python
         pass_filenames: false
         additional_dependencies:
+          - ty>=0.0.33
```
adriantre: (I think this was the reason for my pre-commit complaining earlier)

adriantre: Couldn't find a way to make it respect the version defined in pyproject.toml though..

adamjstewart (commented):
Will re-review after the Sentinel-2 test data PR is merged, trying to get a second set of eyes on that one.

.pre-commit-config.yaml
adamjstewart: This should be in a separate PR, has nothing to do with RasterDataset.geometries

torchgeo/datasets/geo.py
adamjstewart: Yep, feel free to refactor our nodata handling in another PR.

adamjstewart:
Can we add a new section to the docs right after https://docs.torchgeo.org/en/stable/tutorials/custom_raster_dataset.html#Plotting that shows how to do this? Keep it simple, assuming the metadata always exists. Might require downloading additional metadata. This will help me when I write the same thing for Landsat.

.pre-commit-config.yaml
adriantre: Agreed

adriantre:
I love how much simpler the tests in `test_sentinel` became now that the metadatafile exist 🙌

> Can we add a new section to the docs ... that shows how to do this?

I would love learn how to contribute to the docs! What exactly do you want it to show? For sentinel2 my thinking is that the users do not need to do anything at all for the dataset to utilise the true footprint.

adamjstewart:
Those docs are for how to create your own custom RasterDataset. Sentinel is used as an example even though we already have a Sentinel dataset.

adriantre:
Right, yes I can do!

adriantre:
@adamjstewart Can we merge this and do the docs in a different PR?

tests/datasets/test_geo.py
```diff
@@ -502,6 +503,25 @@ def test_change_res(self) -> None:
...
 
+    def test_nodata_value_is_passed_on(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        wvrt_kwargs: list[dict[str, object]] = []
+        original_wvrt = WarpedVRT
+
+        def spy_wvrt(src: object, **kwargs: object) -> WarpedVRT:
+            wvrt_kwargs.append(dict(kwargs))
+            return original_wvrt(src, **kwargs)  # type: ignore[call-arg]
+
+        monkeypatch.setattr('torchgeo.datasets.geo.WarpedVRT', spy_wvrt)
```
adamjstewart: Why do we need to monkeypatch this? I thought the point of adding nodata pixels to the image was that we no longer need to mock anything?

torchgeo/datasets/geo.py
```diff
@@ -391,6 +392,9 @@ class RasterDataset(GeoDataset):
     #: Color map for the dataset, used for plotting
     cmap: ClassVar[dict[int, tuple[int, int, int, int]]] = {}
 
+    #: Nodata value for pixels outside the acquisition footprint; overrides raster nodata.
+    nodata_value: float | None = None
```
adamjstewart: I think we decided to move nodata handling into a separate PR?

torchgeo/datasets/geo.py
```diff
@@ -806,6 +813,29 @@ def _compute_affine_warp_grid(
...
 
+    def _footprint_from_datasource(
+        self, dataset: DatasetReader | WarpedVRT
+    ) -> MultiPolygon | Polygon:
+        """Compute the spatial footprint of the dataset from a file handle.
+
+        Called during indexing for each file in the dataset.
+        Override this in subclasses to compute a more precise footprint than
+        just the raster bounds (e.g. by reading a metadata file or by using
+        :func:`~torchgeo.datasets.utils.get_valid_footprint_from_datasource`).
```
adamjstewart: Remove this suggestion, we don't yet have this utility

torchgeo/datasets/geo.py
```diff
@@ -806,6 +813,29 @@ def _compute_affine_warp_grid(
...
+        Args:
+            dataset: An open raster dataset, either a :class:`rasterio.io.DatasetReader`
+                or a :class:`rasterio.vrt.WarpedVRT`.
+
+        Returns:
+            A :class:`shapely.geometry.Polygon` or
+            :class:`shapely.geometry.MultiPolygon` representing the footprint
+            in the dataset's CRS.
+
+        .. versionadded:: 0.9
```
adamjstewart: 0.10

torchgeo/datasets/sentinel.py
```diff
@@ -462,3 +471,40 @@ def plot(
...
+            dataset: An open raster dataset, either a :class:`rasterio.io.DatasetReader`
+                or a :class:`rasterio.vrt.WarpedVRT`, from which the footprint will
+                be derived.
+
+        Returns:
+            A :class:`shapely.geometry.Polygon` or
+            :class:`shapely.geometry.MultiPolygon` representing the footprint in
+            the dataset's CRS.
+
+        .. versionadded:: 0.9
```
adamjstewart: 0.10

adamjstewart (changes requested):

torchgeo/datasets/sentinel.py
adamjstewart: Can you add a comment to this effect?

torchgeo/datasets/sentinel.py
```diff
@@ -462,3 +471,40 @@ def plot(
...
+        Sentinel-2 metadata XML file (`MTD_MSIL1C.xml`) and transforms it from
+        geographic coordinates (EPSG:4326) into the CRS of the given dataset.
+
+        Args:
+            dataset: An open raster dataset, either a :class:`rasterio.io.DatasetReader`
+                or a :class:`rasterio.vrt.WarpedVRT`, from which the footprint will
+                be derived.
+
+        Returns:
+            A :class:`shapely.geometry.Polygon` or
```
adamjstewart: No need to document the return type here, it's automatically added from the function signature

torchgeo/datasets/sentinel.py
```diff
@@ -462,3 +471,40 @@ def plot(
...
+        metadata_path = filepath.split('GRANULE')[0] + 'MTD_MSIL1C.xml'
+        if not os.path.exists(metadata_path):
+            return super()._footprint_from_datasource(dataset)
+        with rasterio.open(metadata_path) as metadata_src:
+            tags = metadata_src.tags()
+        true_footprint = shapely.wkt.loads(tags['FOOTPRINT'])
+        transformer = pyproj.Transformer.from_crs(
+            pyproj.CRS('EPSG:4326'), dataset.crs, always_xy=True
+        ).transform
+        return cast(MultiPolygon | Polygon, transform(transformer, true_footprint))
```
adamjstewart: Why is a cast needed? Because rasterio lacks type hints?

adamjstewart (changes requested):

torchgeo/datasets/sentinel.py
adriantre: Cool! Love it

torchgeo/datasets/sentinel.py
adriantre: The transform method used here actually from shapely, as it is a vector transform.

I think this transform method is made to be very flexible and does not help with typing well. I see that a newer method exist in shapely. Will check if that one provides better type hints. After Norways national holiday on Sunday 🇳🇴

https://github.com/shapely/shapely/issues/1897

Quick question that I want your input on: 

My instinct was to make a GeoSeries and run .to_crs on it. Then I thought that it would not pass review as I know there is _some_ overhead in making series. But what do you think? Would the GeoSeries approach sit well in your view? I think it reads cleaner than transform objects, myself.

torchgeo/datasets/geo.py
adriantre: Truth!

torchgeo/datasets/geo.py
adriantre: We might have disagreed on that one. But I can do

tests/datasets/test_geo.py
adriantre: I'll move the nodata checks to a separate PR. The monkeypatch is to very that the overridden nodata value is passed on all the way through reprojected datasource.

What is your opinion on such tests? It is a failsafe for future code changes, where codecov might not catch it.

torchgeo/datasets/geo.py
adriantre: I agree more now than I did before :)

torchgeo/datasets/sentinel.py
adamjstewart: My personal philosophy is simplicity > speed, but it depends on how often it gets run. Don't worry too much about speed, we can always make things faster later.

adriantre: So the new transform method does carry the type. Would still have to cast

```python
true_footprint = cast(
    MultiPolygon | Polygon, shapely.wkt.loads(tags['FOOTPRINT'])
)
transformer = pyproj.Transformer.from_crs(
    pyproj.CRS('EPSG:4326'), dataset.crs, always_xy=True
)
return shapely.transform(  # ty: ignore[no-matching-overload]
    true_footprint, transformer.transform, interleaved=False
)
```

`interleaved=False` lets us avoid an ugly lambda extracting the coordinates into numpy, but there are no subs for `False` (only `True`) so we would need the ty: ignore.

I instead went with:

```python
return gpd.GeoSeries(
    [shapely.wkt.loads(tags['FOOTPRINT'])], crs='EPSG:4326'
).to_crs(dataset.crs).iloc[0]
```

torchgeo/datasets/sentinel.py
```diff
@@ -462,3 +467,36 @@ def plot(
             plt.suptitle(suptitle)
 
         return fig
+
+    def footprint_from_datasource(
+        self, dataset: DatasetReader | WarpedVRT
```
adriantre: call the arg `datasource` instead?

adriantre:
@calebrob6 could you please help me understand why the building of docs fails? ♥️

torchgeo/datasets/sentinel.py
adriantre: Done

torchgeo/datasets/geo.py
```diff
@@ -791,6 +792,28 @@ def _compute_affine_warp_grid(
...
+        just the raster bounds (e.g. by reading a metadata file).
+
+        Args:
+            datasource: An open raster dataset, either a :class:`rasterio.io.DatasetReader`
+                or a :class:`rasterio.vrt.WarpedVRT`.
+
+        Returns:
+            A :class:`shapely.geometry.Polygon` or
+            :class:`shapely.geometry.MultiPolygon` representing the footprint
+            in the dataset's CRS.
```
ashnair1: @adriantre Docs are failing due to Sphinx intersphinx warnings. The shapely intersphinx inventory exposes `shapely.Polygon` and `shapely.MultiPolygon`, but **not** `shapely.geometry.Polygon` or `shapely.geometry.MultiPolygon` (Shapely 2.x moved these to the top-level namespace). So this change should fix that.

```diff
-            A :class:`shapely.geometry.Polygon` or
-            :class:`shapely.geometry.MultiPolygon` representing the footprint
-            in the dataset's CRS.
+            A :class:`shapely.Polygon` or
+            :class:`shapely.MultiPolygon` representing the footprint
+            in the dataset's CRS.
```

ashnair1 (changes requested):
Once docs are fixed, this is good to go 👍.

torchgeo/datasets/geo.py
adriantre: Thank you for the explanation! 🙌🏻

ashnair1 (dismissed):

docs/tutorials/custom_raster_dataset.ipynb
```diff
@@ -313,6 +313,50 @@
     "As expected, we have a GeoDataset of size 2 because there are 2 scenes in our root data directory."
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Specifying true footprints\n",
+    "\n",
+    "By default, `RasterDataset` indexes each file by its **bounding box**. But rasters often does not fill its whole bounding box with valid data. The satellites' slanted orbit, or reprojecting tiles will for example result in triangular nodata borders. The **true footprint** is the polygon covering only the valid (non-nodata) pixels. This is the outline of the valid-data mask instead of the enclosing rectangle rectified to the cardinal directions.\n",
```
adamjstewart:
```suggestion
    "By default, `RasterDataset` indexes each file by its **bounding box**. But rasters often do not fill their whole bounding boxes with valid data. For example, a satellite's slanted orbit or tile reprojection will result in triangular nodata borders. The **true footprint** is the polygon covering only the valid (non-nodata) pixels. This is the outline of the valid-data mask instead of the enclosing rectangle rectified to the cardinal directions.\n",
```

docs/tutorials/custom_raster_dataset.ipynb
```diff
@@ -313,6 +313,50 @@
...
+    "\n",
+    "When using a true footprint geo samplers won't draw patches that lie entirely in nodata, so you never feed all-nodata patches to the model. That speeds up **training** and **prediction**, and keeps nodata-only samples from diluting your metrics.\n",
```
adamjstewart:
```suggestion
    "When using a true footprint, samplers avoid patches that lie entirely in nodata regions, so you never feed all-nodata patches to the model. That speeds up **training** and **prediction**, and keeps nodata-only samples from diluting your metrics.\n",
```
It's actually a little more nuanced than this. RandomPatchSampler will avoid all nodata pixels (unless combined with a temporal sampler) and GriddedPatchSampler will purposefully include partial nodata patches. But yes, we will never include fully nodata patches.

docs/tutorials/custom_raster_dataset.ipynb
```diff
@@ -313,6 +313,50 @@
...
+    "\n",
+    "You can make your `RasterDataset` use it by overriding `footprint_from_datasource`. It receives an open datasource and returns a shapely `Polygon`/`MultiPolygon` in the dataset's CRS. Where the footprint comes from depends on your data. Some products ship the acquisition footprint in a metadata file, and otherwise you can (pre)compute it from the nodata mask (`dataset.dataset_mask()`). Sentinel-2 is used here as an example:"
```
adamjstewart:
```suggestion
    "You can tell your `RasterDataset` how to calculate the true footprint by overriding `footprint_from_datasource`. It receives an open datasource and returns a shapely `Polygon`/`MultiPolygon` in the dataset's CRS. Where the footprint comes from depends on your data. Some products ship the acquisition footprint in a metadata file, and otherwise you can (pre)compute it from the nodata mask (`dataset.dataset_mask()`). Sentinel-2 is used here as an example:"
```

docs/tutorials/custom_raster_dataset.ipynb
```diff
@@ -313,6 +313,50 @@
...
+    "            gpd.GeoSeries.from_wkt([footprint], crs='EPSG:4326')\n",
+    "            .to_crs(datasource.crs)\n",
+    "            .iloc[0]\n",
+    "        )"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": "NB: Computing it from the nodata mask instead means reading the pixels, which can be slower for large rasters, and relies on your files declaring a nodata value (or carrying an internal mask) so the valid pixels can be identified. Consider precomputing these polygons and read from disk in the override."
```
adamjstewart: In https://docs.torchgeo.org/en/stable/tutorials/geospatial.html, we actually use a dedicated warning block for this. Could be useful here too. I don't think GitHub's Markdown warnings are valid though. Completely optional.

tests/datasets/test_sentinel.py
```diff
@@ -159,3 +165,31 @@ def test_invalid_index(self, dataset: Sentinel2) -> None:
...
+        self, dataset: Sentinel2, tmp_path: Path
+    ) -> None:
+        filepath = next(iter(dataset.files))
+        # Move this raster to a directory where it does not find metadata file
+        link = tmp_path / Path(filepath).name
+        link.symlink_to(Path(filepath).resolve())
+
+        with rasterio.open(link) as src:
+            result = dataset.footprint_from_datasource(src)
+            assert result.equals_exact(shapely.box(*src.bounds), tolerance=1e-9)  # type: ignore[arg-type]
```
adamjstewart: What is the ignore for here? Is this a bug we can fix?

torchgeo/datasets/geo.py
```diff
@@ -791,6 +792,28 @@ def _compute_affine_warp_grid(
...
+    ) -> MultiPolygon | Polygon:
+        """Compute the spatial footprint of the dataset from a file handle.
+
+        Called during indexing for each file in the dataset.
+        Override this in subclasses to compute a more precise footprint than
+        just the raster bounds (e.g. by reading a metadata file).
+
+        Args:
+            datasource: An open raster dataset, either a :class:`rasterio.io.DatasetReader`
+                or a :class:`rasterio.vrt.WarpedVRT`.
```
```suggestion
            datasource: An open raster dataset.
```
The rest comes from the function signature, it will appear in the docs

torchgeo/datasets/geo.py
```diff
@@ -791,6 +792,28 @@ def _compute_affine_warp_grid(
...
         return dst_transform, dst_width, dst_height, needs_warp
 
+    def footprint_from_datasource(
+        self, datasource: DatasetReader | WarpedVRT
+    ) -> MultiPolygon | Polygon:
+        """Compute the spatial footprint of the dataset from a file handle.
+
+        Called during indexing for each file in the dataset.
+        Override this in subclasses to compute a more precise footprint than
+        just the raster bounds (e.g. by reading a metadata file).
```
```suggestion
        just the raster bounds (e.g., by reading a metadata file).
```

torchgeo/datasets/sentinel.py
```diff
@@ -462,3 +467,36 @@ def plot(
...
+        """Extract the true geometric footprint from a Sentinel-2 datasource.
+
+        This method reads the original footprint geometry from the associated
+        Sentinel-2 metadata XML file (`MTD_MSIL1C.xml`) and transforms it from
+        geographic coordinates (EPSG:4326) into the CRS of the given dataset.
+
+        Args:
+            datasource: An open raster dataset, either a :class:`rasterio.io.DatasetReader`
+                or a :class:`rasterio.vrt.WarpedVRT`, from which the footprint will
+                be derived.
```
adamjstewart: Let's match the base class description

torchgeo/datasets/sentinel.py
```diff
@@ -462,3 +467,36 @@ def plot(
...
+
+        .. versionadded:: 0.10
+        """
+        if hasattr(datasource, 'src_dataset'):
+            # When dataset is a WarpedVRT
+            filepath = datasource.src_dataset.name
+        else:
+            filepath = datasource.name
+        metadata_path = filepath.split('GRANULE')[0] + 'MTD_MSIL1C.xml'
```
adamjstewart, line 494: I'm thinking about changing Sentinel2 to use the .SAFE GDAL driver instead of our current complicated hierarchy traversal (in a separate PR). Would this change your PR at all?

P.S. If anyone is interested in doing this or has time, please open a PR for me. Any PR that deletes more lines of code than it adds is a good PR.

torchgeo/datasets/sentinel.py
```diff
@@ -462,3 +467,36 @@ def plot(
...
+        if not os.path.exists(metadata_path):
+            return super().footprint_from_datasource(datasource)
```
adamjstewart: I imagine that this kind of fallback will be common. Another design pattern we could use is:
```python
try:
    footprint = self.footprint_from_datasource(vrt)
except FileNotFoundError:
    footprint = shapely.box(*vrt.bounds)
```
or
```python
footprint = self.footprint_from_datasource(vrt) or shapely.box(*vrt.bounds)
```
if we let it return None (no return). We would put this in RasterDataset so we don't need fallback in each subclass. Let me know what you think, I'll let you decide, whichever is simpler to understand is best.

torchgeo/datasets/sentinel.py
```diff
@@ -462,3 +467,36 @@ def plot(
...
+        with rasterio.open(metadata_path) as metadata_src:
+            tags = metadata_src.tags()
+        # The FOOTPRINT tag in MTD_MSIL1C.xml is always stored in EPSG:4326.
+        return (
+            gpd.GeoSeries.from_wkt([tags['FOOTPRINT']], crs='EPSG:4326')
+            .to_crs(datasource.crs)
+            .iloc[0]
+        )
```
adamjstewart: Could be split into 2 lines to avoid the long expression.

adamjstewart (commented):
I think this is 99% ready to go now after a couple questions and grammar fixes. Excited to try this out myself for Landsat!

P.S. I'm shooting for 0.10 later this week or early next week, definitely want to include this PR if you have time to finish it up. If you don't have time, I can also help.

torchgeo/datasets/geo.py
adriantre: This is not the first time I forget that 😅

tests/datasets/test_sentinel.py
adriantre: Added it because my IDE linter complained about this, but I see that ty don't. Removed 👍

torchgeo/datasets/sentinel.py
adriantre: I understand this as meaning for the args and returns specifically/only.

torchgeo/datasets/sentinel.py
adriantre: Depending on the implementation, the `tags['FOOTPRINT']` might be available on the open datasource directly. But this heavily relies on the implementation.

torchgeo/datasets/sentinel.py
adriantre: I like option 2. It comes with the benefit of not putting `shapely.box(*vrt.bounds)` in the `RasterDataset.footprint_from_datasource`, as this is not really the footprint, but a fallback. Instead of the `or` I check `not None` as we should allow `footprint_from_datasource` to return an empty polygon.

adriantre:
Thanks! I have now addressed all feedback 💪

tests/datasets/test_sentinel.py
adriantre: It also was redundant after returning None instead of calling super

docs/tutorials/custom_raster_dataset.ipynb
adriantre: Added it, and it rendered nicely in my IDE :)

torchgeo/datasets/geo.py
```diff
@@ -798,6 +802,26 @@ def _compute_affine_warp_grid(
...
+        Args:
+            datasource: An open raster dataset.
+
+        Returns:
+            The true footprint in the dataset's CRS, or ``None`` if the metadata
+            file is not found (falling back to the raster's bounding box).
+
+        .. versionadded:: 0.10
+        """
+        return None
```
adamjstewart: Technically `return` is equivalent to `return None`

torchgeo/datasets/sentinel.py
```diff
@@ -462,3 +467,35 @@ def plot(
...
+            The true footprint in the dataset's CRS, or ``None`` if the metadata
+            file is not found (falling back to the raster's bounding box).
+
+        .. versionadded:: 0.10
+        """
+        if hasattr(datasource, 'src_dataset'):
+            # When dataset is a WarpedVRT
+            filepath = datasource.src_dataset.name
+        else:
+            filepath = datasource.name
```
adamjstewart, line 493: Hmm, are we worried that this will also be a common thing and we should instead pass the filepath as input? It sounds like if we switch to the .SAFE driver we would want the filehandle instead of the filepath though.

torchgeo/datasets/sentinel.py
```diff
@@ -462,3 +467,35 @@ def plot(
...
+        metadata_path = filepath.split('GRANULE')[0] + 'MTD_MSIL1C.xml'
+        if not os.path.exists(metadata_path):
+            return None
```
adamjstewart, line 496: 1% of me is wondering if we should issue a warning, but I think this might be overkill. 99% of users will download the .SAFE format and everything will work.

adamjstewart (dismissed):
I think this looks good to me now! Couple comments for discussion, don't necessarily need to change anything.

torchgeo/datasets/sentinel.py
adriantre: I think for now we want to centralise file opening to avoid users reading pixels raw when they should have been warped etc. Since this method can also be used compute footprint from pixels.

Hard to foresee whether it will be common. Some users might have stored the footprint as a separate file, and might do this to find its relative path to the raster? 🤷🏼‍♂️

torchgeo/datasets/sentinel.py
adriantre: I agree, both with the conclusion and the percentage 😅 I initially put a warning, but if they don't have it they probably would not go the steps to get it

adamjstewart (approved):
Let's merge. If we change our minds later, we can always create a new helper function and deprecate the old one. This is a huge improvement towards avoiding nodata pixels! Time to download some Landsat data and try my hand at this personally 😄

ashnair1 (approved):

adamjstewart:
> Time to download some Landsat data and try my hand at this personally 😄

I started hacking on this and it's a nightmare. You have to parse both MTL.xml and ANG.txt and combine the info. But I can't find the docs for either spec, so don't know what I'm doing. We basically need to do https://pystac.readthedocs.io/en/stable/tutorials/creating-a-landsat-stac.html#Read-metadata-from-the-MTL-file to get the footprint (what STAC calls "geometry"), but I don't like copying things I don't understand (and can't copy them under their license anyway). This is as far as I got before I quit (this is just the "bbox" part):
```diff
diff --git a/torchgeo/datasets/landsat.py b/torchgeo/datasets/landsat.py
index 1f12d92f9..15a84e9b0 100644
--- a/torchgeo/datasets/landsat.py
+++ b/torchgeo/datasets/landsat.py
@@ -4,12 +4,18 @@
 """Landsat datasets."""
 
 import abc
+import os
+import xml.etree.ElementTree as ET
 from collections.abc import Callable, Iterable, Sequence
 from typing import ClassVar
 
 import matplotlib.pyplot as plt
+import shapely
 from matplotlib.figure import Figure
-from pyproj import CRS
+from pyproj import CRS, Transformer
+from rasterio import DatasetReader
+from rasterio.vrt import WarpedVRT
+from shapely import Polygon
 
 from .errors import RGBBandsMissingError
 from .geo import RasterDataset
@@ -102,6 +108,43 @@ class Landsat(RasterDataset, abc.ABC):
 
         super().__init__(paths, crs, res, bands, transforms, cache, time_series)
 
+    def footprint_from_datasource(
+        self, datasource: DatasetReader | WarpedVRT
+    ) -> Polygon | None:
+        """Compute the spatial footprint of the dataset from a file handle.
+
+        Args:
+            datasource: An open raster dataset.
+
+        Returns:
+            The true footprint in the dataset's CRS, or ``None`` if the metadata
+            file is not found (falling back to the raster's bounding box).
+
+        .. versionadded:: 0.10
+        """
+        dataset = getattr(datasource, 'src_dataset', datasource)
+        filepath = dataset.name.split(self.bands[0])[0] + 'MTL.xml'
+        if os.path.exists(filepath):
+            tree = ET.parse(filepath)
+            root = tree.getroot()
+            node = root.find('PROJECTION_ATTRIBUTES')
+
+            uly = float(node.find('CORNER_UL_LAT_PRODUCT').text)
+            ulx = float(node.find('CORNER_UL_LON_PRODUCT').text)
+            ury = float(node.find('CORNER_UR_LAT_PRODUCT').text)
+            urx = float(node.find('CORNER_UR_LON_PRODUCT').text)
+            lly = float(node.find('CORNER_LL_LAT_PRODUCT').text)
+            llx = float(node.find('CORNER_LL_LON_PRODUCT').text)
+            lry = float(node.find('CORNER_LR_LAT_PRODUCT').text)
+            lrx = float(node.find('CORNER_LR_LON_PRODUCT').text)
+
+            coords = [(lrx, lry), (urx, ury), (ulx, uly), (llx, lly), (lrx, lry)]
+            footprint = Polygon(coords)
+            wgs84 = CRS.from_epsg(4326)
+            transformer = Transformer.from_crs(wgs84, datasource.crs, always_xy=True)
+            footprint = shapely.ops.transform(transformer.transform, footprint)
+            return footprint
+
     def plot(
         self, sample: Sample, show_titles: bool = True, suptitle: str | None = None
     ) -> Figure:
```

adriantre:
Cool! Looks correct, did it run? You might not need to read the ANG.txt if this works. I saw in the link you sent that they, for some reason, don't use all of the coordinates like you do. They create just the envelope and that's their justification for the complex ANG.txt parsing. So I'd think that what you have is actually the acquisition corners.

adamjstewart:
Nah, this just gives the bbox. I tried plotting it and it was just a square. I think ANG.txt is still required.

adriantre:
Was this for a product at the edge of the acquisition, like it has nodata triangle? Else the footprint will be axis-aligned ofc

adamjstewart:
I plotted the entire Landsat tile and the "footprint" of that tile, but the tile was rotated and the footprint matched the bbox.

adriantre:
Right, Landsat is not axis aligned like Sentinel2.

I'm not too familiar with the different formats available, but this mentions a *stac.json file?

https://gis.stackexchange.com/a/479458

adamjstewart:
Yes, _if_ a STAC catalog already exists, then you don't need to do this. We are doing this in order to create a catalog because one doesn't exist yet.

PR #2904, opened 2025-07-29
Added CLOSP vision encoders

DarthReca:
As suggested by @isaaccorley, I have added the weights for the vision encoder of CLOSP models. They are all aligned with text, and GeoCLOPS is aligned with location, too.

I have tried the downloads, and they work without any issues.


isaaccorley:
Just need to update the list of pretrained weights `docs/api/weights/sentinel2.csv` and `docs/api/weights/sentinel1.csv`.

adamjstewart:
Fun fact: OpenRAIL is not yet recognized by SPDX or OSI, partly because it is technically not an open source license (it has restrictions on usage) and because many projects that use OpenRAIL have custom license templates which can't easily be recognized or categorized. We try to follow SPDX for the licenses column, but I would much rather list OpenRAIL than "custom". Are the models released under OpenRAIL-M? May be useful to be more specific.

docs/api/weights/sentinel2.csv
adamjstewart, line 1: Number of channels is wrong, should be 13 instead of 2

We have historically used "ALL" and "RGB" to differentiate between similar weights covering all bands or only RGB bands. We don't have to follow that pattern if we don't want to though.

torchgeo/models/resnet.py
```diff
@@ -391,6 +391,62 @@ class ResNet50_Weights(WeightsEnum):  # type: ignore[misc]
     .. versionadded:: 0.4
     """
 
+    SENTINEL2_CLOSP = Weights(
+        url='https://hf.co/DarthReca/CLOSP-Visual/resolve/main/closp-rn_s2_encoder.pth',
```
adamjstewart: We usually try to add the first 8 digits of the SHA256 to the filename as supported by https://docs.pytorch.org/docs/stable/hub.html#torch.hub.load_state_dict_from_url

Also, we need to use a more stable download URL, so replace `main` with the most recent commit hash (e.g., 68208978239cf17976de38cb20b95024ccdfa5a2).

DarthReca:
Thanks for the suggestion. I am using the default HF template for that license. 

I mainly like them for the general clauses of not causing harm. 
I can change this license for the vision encoders if it is an issue.

adamjstewart (changes requested):
We try to keep all weights alphabetically sorted, both in the CSV files and also in the .py files.

docs/api/weights/sentinel2.csv
DarthReca: I can use a uniform notation for clarity, but in any case, this model in RGB does not exist.

torchgeo/models/resnet.py
DarthReca: Let me update the files on HF and the URLs with the SHA.

adamjstewart:
License is definitely not an issue, keep it however you want 🙂

docs/api/weights/sentinel2.csv
adamjstewart: Good. I will never understand why people make RGB-only models for Sentinel-2.

DarthReca:
> We try to keep all weights alphabetically sorted, both in the CSV files and also in the .py files.

I have noticed that the "main" order in csv files is not alphabetical (e.g, Swin is after ViT). Do you want to keep it that way, or will you go for a full sort?

adamjstewart:
I would love to have a full sort, either in this PR or in a follow-up PR.

DarthReca:
What about refactoring also the transforms and bands for easier maintainability? 

I suggest a system like this:
BANDS = {
    "LANDSAT_TM_TOA": ['B1', 'B2', 'B3', 'B4', 'B5', 'B6', 'B7'],
    "LANDSAT_ETM_TOA": ['B1', 'B2', 'B3', 'B4', 'B5', 'B6_VCID_1', 'B6_VCID_2', 'B7', 'B8'],
    # ... other band lists
}

TRANSFORMS = {
    "SSL4EO_L": [Transforms, ...],
    "SSL4EO_S12_10K":  [Transforms, ...],
    "SECO":  [Transforms, ...],
    # ... other transforms
}

adamjstewart:
Sounds like a good idea, but let's do that in a separate PR.

docs/api/weights/sentinel2.csv
```diff
@@ -1,46 +1,50 @@
...
-ResNet50_Weights.SENTINEL2_MI_RGB_SATLAS,3,`link <https://github.com/allenai/satlaspretrain_models>`__,`link <https://arxiv.org/abs/2211.15660>`__,ODC-BY,,,,
-ResNet50_Weights.SENTINEL2_SI_MS_SATLAS,9,`link <https://github.com/allenai/satlaspretrain_models>`__,`link <https://arxiv.org/abs/2211.15660>`__,ODC-BY,,,,
-ResNet50_Weights.SENTINEL2_SI_RGB_SATLAS,3,`link <https://github.com/allenai/satlaspretrain_models>`__,`link <https://arxiv.org/abs/2211.15660>`__,ODC-BY,,,,
-ResNet50_Weights.SENTINEL2_RGB_MOCO, 3,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,"CC-BY-4.0",,,
-ResNet50_Weights.SENTINEL2_RGB_SECO, 3,`link <https://github.com/ServiceNow/seasonal-contrast>`__,`link <https://arxiv.org/abs/2103.16607>`__,"Apache-2.0",87.81,,,
+EarthLoc_Weights.SENTINEL2_RESNET50,3,`link <https://github.com/gmberton/EarthLoc>`__,`link <https://arxiv.org/abs/2403.06758>`__,MIT,,,,
 ResNet152_Weights.SENTINEL2_MI_MS_SATLAS,9,`link <https://github.com/allenai/satlaspretrain_models>`__,`link <https://arxiv.org/abs/2211.15660>`__,ODC-BY,,,,
```
adamjstewart: ResNet18 should come before ResNet152. We should use an alphanumeric sort, not just an alphabetic sort.

torchgeo/models/resnet.py
```diff
@@ -544,6 +544,19 @@ class ResNet50_Weights(WeightsEnum):  # type: ignore[misc]
         },
     )
 
+    SENTINEL1_GRD_CLOSP = Weights(
```
adamjstewart: SENTINEL1_ALL should come before SENTINEL1_GRD

adamjstewart (commented):
To be honest, some of the sorting changes have made it more difficult to review the PR. Maybe we could separate sorting and CLOSP into separate PRs. My goal was to at least maintain the current mostly-sorted quality and not introduce any additional inconsistencies.

DarthReca:
Thanks for your effort. I'll go back to the previous order and only add the models.

DarthReca:
Just a note on the license, I have used CreativeML-OpenRAIL-M. However, I feel there is no official abbreviation.

adamjstewart (approved):
Looks perfect now, thanks for all the hard work, and apologies for all the back-and-forth formatting changes.

adamjstewart:
P.S. You might want to clean up your main branch. Your main branch has diverged pretty far from TorchGeo's upstream main branch. This is why we recommend creating new feature branches for PRs.

DarthReca:
Thanks, I have noticed the divergence too far. I will open a new PR for the reworking of transforms and bands.

PR #2906, opened 2025-07-30
xView2: change detection compatibility

adamjstewart:
Changes xView2 and xView2DataModule to be compatible with our new ChangeDetectionTask.

Not 100% sure this is the right strategy. We could also change this to a time series semantic segmentation dataset. How common is the background class? Is background always the same before and after? There is currently no way to ignore background pixels.

@RitwikGupta 

Part of #2382


RitwikGupta:
The background can change. Let's say you have grass pre-disaster and water post-disaster.

adamjstewart:
Okay, so background (class 0) is not something to ignore, it just denotes something that is not a building. Then this approach should be fine.

RitwikGupta:
Correct. Since the scoring is only on the change experienced by buildings, the change mask produced over the background can simply be ignored.

RitwikGupta:
There's a world in which a building was built where there was previously background, but that's pretty niche

PR #2909, opened 2025-07-31
Models: Add ChangeVIT model

hkristen:
Cleanroom implementation of the [ChangeViT paper](https://arxiv.org/pdf/2406.12847). The [code shared](https://github.com/zhuduowang/ChangeViT) by the authors is CC-BY-NC-SA-4.0 **therefore we can't copy from there or even look at the code while implementing.** 

This initial version should follow the paper quite closely. It has a some tests, a few of them where written at the beginning to make sure I don't go to far off while implementing. I am sure I over-complicated things, very happy for suggestions to make it more lean!

You can find the ToDos that immediately come to my mind at the top of changevit.py or here below

### TODO:
- [ ] Train model on benchmark datasets (Levir-CD, WHU-CD) and compare with paper results
- [ ] Tolerance paramter size ok? -> timm vit + standard resnet vs. custom VIT + custom resnet
- [ ] Which weights should be used as standard? Should we add DINO-S weights as well?
- [ ] Go through code once more and make sure it aligns with paper

---- 

Now I want to run a benchmark training run with the same parameters as in the paper to ensure the implementation produces similar results on LEVIR-CD (which is the actual test if my implementation makes sense). I am using the CLI for it but get this error when I try to run it and can't figure out what the issue is:

`torchgeo fit --config tests/conf/changevit_levircd_benchmark.yaml `
```
error: Parser key "model":
  Does not validate against any of the Union subtypes
  Errors:
    - Problem with given class_path 'torchgeo.trainers.ChangeDetectionTask':
        Parser key "model":
          Cannot take a Union of no types.
  Given value type: <class 'dict'>
  Given value: {'class_path': 'ChangeDetectionTask', 'init_args': {'model': 'changevit_tiny', 'task': 'binary', 'loss': 'bce', 'lr': 0.0002, 'num_classes': 2}}
```
These seems like a dumb error - any suggestions @adamjstewart what the issue is here?


adamjstewart:
Maybe you can rebase on upstream main? There are quite a lot of unrelated changes that got sucked into this PR.

hkristen:
> Maybe you can rebase on upstream main? There are quite a lot of unrelated changes that got sucked into this PR.

Damn it, I rebased yesterday to get the latest changes from change.py -> will try to fix this

hkristen:
Things got messy, I ported my changes to a new clean branch based on upstream main.

PR #2910, opened 2025-07-31
Models: Add ChangeVIT model

hkristen:
Cleanroom implementation of the [ChangeViT paper](https://arxiv.org/pdf/2406.12847). The [code shared](https://github.com/zhuduowang/ChangeViT) by the authors is CC-BY-NC-SA-4.0 **therefore we can't copy from there or even look at the code while implementing.** 

This initial version should follow the paper quite closely. It has a some tests, a few of them where written at the beginning to make sure I don't go to far off while implementing. I am sure I over-complicated things, very happy for suggestions to make it more lean!

### TODO:
- [x] ~Tolerance paramter size ok? -> timm vit + standard resnet vs. custom VIT + custom resnet (I think something is off at the decoder)~
- [x] ~Which weights should be used as standard? Should we add DINO-S weights as well?~
- [x] Go through code once more and make sure it aligns with paper
- [x] Train model on benchmark datasets (Levir-CD, WHU-CD) and compare with paper results
- [x] Split out the datamodule/transforms stuff to #3028 
- [x]  after #3028 is merged, rebase changevit branch and merge to main

---- 

Now I want to run a benchmark training run with the same parameters as in the paper to ensure the implementation produces similar results on LEVIR-CD (which is the actual test if my implementation makes sense). I am using the CLI for it but get this error when I try to run it and can't figure out what the issue is:

`torchgeo fit --config tests/conf/changevit_levircd_benchmark.yaml `
```
error: Parser key "model":
  Does not validate against any of the Union subtypes
  Errors:
    - Problem with given class_path 'torchgeo.trainers.ChangeDetectionTask':
        Parser key "model":
          Cannot take a Union of no types.
  Given value type: <class 'dict'>
  Given value: {'class_path': 'ChangeDetectionTask', 'init_args': {'model': 'changevit_tiny', 'task': 'binary', 'loss': 'bce', 'lr': 0.0002, 'num_classes': 2}}
```
These seems like a dumb error - any suggestions @adamjstewart what the issue is here?


adamjstewart:
> These seems like a dumb error - any suggestions @adamjstewart what the issue is here?

This is actually a fun error. When you use `torchgeo`, you pick up whichever `torchgeo` command is in your `PATH`. For me, and probably for you, this is the latest stable release (0.7.*), which does not have `ChangeDetectionTask`. If you want to guarantee that you are using your local copy of TorchGeo from git clone, use `python -m torchgeo` instead.

torchgeo/datamodules/levircd.py
```diff
@@ -52,24 +55,44 @@ def __init__(
...
-            keepdim=True,
-            same_on_batch=True,
-        )
+
+        # Use CenterCrop for ChangeViT models to avoid multiple patches
+        if model.startswith('changevit'):
+            self.val_aug = K.AugmentationSequential(
+                K.VideoSequential(
+                    K.Normalize(mean=self.mean, std=self.std),
+                    K.CenterCrop(self.patch_size),
```
isaaccorley: We shouldn't do this because then your val/test metrics wouldn't be apples to apples with other models because they are running only on center cropped patches vs the entire image.

torchgeo/trainers/change.py
```diff
@@ -228,6 +237,14 @@ def configure_models(self) -> None:
                     classes=num_classes,
                     encoder_weights='imagenet' if weights is True else None,
                 )
+            case 'changevit_small':
+                self.model = changevit_small(
+                    weights=weights if isinstance(weights, WeightsEnum) else None
```
isaaccorley: Should raise an error or warning instead of just returning None

torchgeo/trainers/change.py
```diff
@@ -236,17 +253,43 @@ def configure_models(self) -> None:
...
+            else:
+                self.model.encoder.load_state_dict(state_dict)
 
         # Freeze backbone
         if self.hparams['freeze_backbone'] and model != 'fcn':
-            for param in self.model.encoder.parameters():
-                param.requires_grad = False
+            if model.startswith('changevit'):
+                # Freeze ViT backbone for ChangeViT models
+                for param in self.model.vit_backbone.parameters():
```
isaaccorley: We should rename "vit_backbone" in ChangeVit to "encoder" for consistency which would also make these lines not needed.

torchgeo/trainers/change.py
```diff
@@ -281,7 +344,15 @@ def _shared_step(self, batch: Any, batch_idx: int, stage: str) -> Tensor:
         # Retrieve the correct metrics based on the stage
         metrics = getattr(self, f'{stage}_metrics', None)
         if metrics:
-            metrics(y_hat, y)
+            # Transform predictions for metrics calculation (don't modify original y_hat)
```
isaaccorley: Why is this needed?

torchgeo/models/changevit.py
```diff
@@ -0,0 +1,645 @@
...
+
+import torch
+import torch.nn as nn
+from einops import rearrange
+from torch import Tensor
+from torch.nn.modules import Module
+from torchvision.models._api import Weights, WeightsEnum
+
+
+class BasicBlock(Module):
```
isaaccorley: Preferably we would use timm's implementation of this in https://github.com/huggingface/pytorch-image-models/blob/main/timm/models/resnet.py. This allows us to be consistent with the layers of timm resnets and also use pretrained weights from timm

torchgeo/models/changevit.py
```diff
@@ -0,0 +1,645 @@
...
+        if self.downsample is not None:
+            identity = self.downsample(x)
+
+        out += identity
+        result: Tensor = self.relu(out)
+
+        return result
+
+
+class DetailCaptureModule(Module):
```
isaaccorley, line 21: Same here imo this is a fancy name for just 

timm.create_model("resnet18", features_only=True)

Where you just take the outputs of the first 3 blocks. We shouldn't need to maintain this code because it's not specific to geospatial and this code also doesn't take advantage of pretrained weights.

torchgeo/datamodules/levircd.py
hkristen: I totally agree with you on this one. I will roll this back to how it was originally. 

I was trying to reproduce the training the authors did, to verify if my implementation produces similar results on levircd. Today I realized that they actually do this: They segment each 1024x1024 image into 16 x 256x256 images. 

-> I was thinking about writing a separate class for that in the levircd datamodule to follow this approach. That I then can use within the tests/conf/changevit_levircd_benchmark.yaml. 

What do you think?

isaaccorley: This is actually what _ExtractPatches does if you set window size = 256 and stride = 256

torchgeo/trainers/change.py
hkristen: This is not needed anymore. Was a fix I implemented while developing and forgot about it. Cleaned this and some other up now

hkristen:
Thx for all the comments @isaaccorley, I tried to factor them all in.

### Paramter size
Due to relying on timm the DetailCapure is now 73% smaller than reported in the paper (0.73M vs. 2.7M). Same with the ViT part, it is 38% smaller (7.3M vs. 11.68M). I assume we will still get similar results as the paper as the core architecture is the same, it is just more efficient? Or could this be an issue?

### Training
I feel this is now in a state where I can start training it with the new LEVIRCDBenchmarkDataModule + changevit_levircd_benchmark.yaml and compare the results with the paper (I am sure I will figure more things to change/fix out along the way).

torchgeo/models/changevit.py
```diff
@@ -0,0 +1,511 @@
...
+            import timm
+        except ImportError as e:
+            raise ImportError(
+                '`timm` is not installed and is required for this model. '
+                'Please install it with `pip install timm`.'
+            ) from e
+
+        # Create ResNet18 backbone with features_only=True to get intermediate features
+        self.backbone = timm.create_model(
+            'resnet18',
```
isaaccorley: Could we make the model an input argument so it doesn't always default to resnet18? For example could we use resnet50? I believe for the projection layers, you can get the number of output channels of each index using this 

```python
model.feature_info.channels()

# returns [64, 64, 128] for resnet18
# returns [64, 256, 512] for resnet50
```

torchgeo/models/changevit.py
```diff
@@ -0,0 +1,511 @@
...
+            pretrained=True,  # Use pretrained weights for better initialization
```
isaaccorley: Let's make this an arg as well, for example, for tests we don't want to use pretrained=True because this will download Imagenet pretrained weights even if we are loading from trained checkpoint

torchgeo/models/changevit.py
```diff
@@ -0,0 +1,511 @@
...
+            features_only=True,
+            out_indices=[0, 1, 2],  # Get features at 1/2, 1/4, 1/8 scales
+            in_chans=in_channels,  # Support 6-channel input for bitemporal data
+        )
+
+        # Add projection layers to match paper's channel dimensions (64, 128, 256)
+        # timm ResNet18 gives us (64, 64, 128) channels, we need (64, 128, 256)
+        self.proj1 = nn.Identity()  # 64 -> 64 (already correct)
```
isaaccorley: Are we sure the paper still doesn't use a 1x1 conv layer here just to learn features from the resnet extracted feature maps? `self.proj1 = nn.Conv2d(64, 64, kernel_size=1)`

torchgeo/models/changevit.py
```diff
@@ -0,0 +1,511 @@
...
+        """
+        super().__init__()
+
+        try:
+            import timm
+        except ImportError as e:
+            raise ImportError(
+                '`timm` is not installed and is required for this model. '
+                'Please install it with `pip install timm`.'
+            ) from e
```
isaaccorley: timm is installed by default with torchgeo so this can be imported at the top without checking for import errors

torchgeo/models/changevit.py
```diff
@@ -0,0 +1,511 @@
...
+                    nn.Conv2d(inner_channels, inner_channels, 3, 1, 1),
+                    nn.BatchNorm2d(inner_channels),
+                    nn.ReLU(True),
+                )
+                for _ in range(num_convs - 1)
+            ]
+        )
+
+        # Classification layer
+        layers.append(nn.Conv2d(inner_channels, 1, 3, 1, 1))
```
isaaccorley: let's call this last layer `self.head` just for clarity in case someone wants to later adapt the output layer

isaaccorley: Also can we make `num_classes` an arg? Not every change detection is binary change and so we shouldn't hardcode to out_channels=1

torchgeo/models/changevit.py
```diff
@@ -0,0 +1,511 @@
...
+        t2t1 = torch.cat([bi_feature[:, 1], bi_feature[:, 0]], dim=1)
+
+        # Process both orderings together
+        c1221 = self.convs(torch.cat([t1t2, t2t1], dim=0))
+
+        # Apply upsampling to match target size
+        if target_size is not None:
+            c1221 = F.interpolate(
+                c1221, size=target_size, mode='bilinear', align_corners=False
+            )
```
isaaccorley: The upsampling should be done outside of the decoder imo and in the ChangeViT module directly. Avoiding if statements inside nn.Modules is better practice in case someone wants to export the model this would be considered a graph break

torchgeo/models/changevit.py
```diff
@@ -0,0 +1,511 @@
...
+        enhanced_spatial = rearrange(
+            enhanced_features_tensor, 'b t (h w) d -> b t d h w', h=h_patch, w=w_patch
+        )
+
+        # Apply change detection with proper temporal features
+        # Pass original input size to ensure exact output dimensions
+        target_size = (x.shape[-2], x.shape[-1])  # (H, W) from original input
+        c12, c21 = self.decoder(
+            enhanced_spatial, target_size=target_size
+        )  # Returns tuple of [B, 1, H, W] tensors
```
isaaccorley: We can do the upsampling to target size after this call to self.decoder

torchgeo/models/changevit.py
```diff
@@ -0,0 +1,511 @@
...
+        if self.training:
+            # Training: return logits for BCE loss computation
+            return {
+                'change_prob': change_logits
+            }  # Actually logits, but keeping key name for compatibility
+        else:
+            # Inference: return probabilities and binary map
+            change_prob = torch.sigmoid(change_logits)
+            change_binary = (change_prob > 0.5).float()  # Threshold at 0.5
+            return {'change_prob': change_prob, 'change_binary': change_binary}
```
isaaccorley: We don't want this imo. we should always return logits. Sigmoid/Softmax is typically handle by the loss function and users may want to use a different threshold than 0.5.

isaaccorley: In other words just replace all this with `return change_logits`. Also update the return type of the forward pass to `Tensor`

torchgeo/trainers/change.py
```diff
@@ -263,13 +294,20 @@ def _shared_step(self, batch: Any, batch_idx: int, stage: str) -> Tensor:
         x = batch['image']
         y = batch['mask']
 
-        if not model.startswith('fcsiam'):
+        if not model.startswith('fcsiam') and not model.startswith('changevit'):
             x = rearrange(x, 'b t c h w -> b (t c) h w')
```
blaz-r: This might also be needed for other models, including BTC. Right now, I just do the rearrangement in the forward, but maybe we could think of a way to handle this better?

torchgeo/trainers/change.py
```diff
@@ -373,11 +412,22 @@ def predict_step(
         x = batch['image']
         if model == 'unet':
             x = rearrange(x, 'b t c h w -> b (t c) h w')
-        y_hat: Tensor = self(x)
+        elif not model.startswith('fcsiam') and not model.startswith('changevit'):
+            x = rearrange(x, 'b t c h w -> b (t c) h w')
```
blaz-r: Same here.

torchgeo/datamodules/levircd.py
```diff
@@ -1,150 +1,174 @@
...
-            K.VideoSequential(
-                K.Normalize(mean=self.mean, std=self.std),
-                _ExtractPatches(window_size=self.patch_size),
+
+    def _create_deterministic_val_aug(self) -> K.AugmentationSequential:
+        """Create deterministic validation/test augmentation for reproducible evaluation."""
+        return K.AugmentationSequential(
+            K.Normalize(mean=self.mean, std=self.std),
+            _ExtractPatches(
+                window_size=self.patch_size, stride=self.stride, keepdim=False
```
blaz-r: The updated levir and transforms don't quite work for me. I think that this should be set to `keepdim=True`, so that the correct reshaping triggers.

torchgeo/transforms/transforms.py
```diff
@@ -87,8 +204,25 @@ def apply_transform(
...
-        if flags['keepdim']:
-            out = rearrange(out, 'b t c h w -> (b t) c h w')
+        # Handle temporal data from VideoSequential
+        # If we extracted multiple patches and batch size suggests temporal data (even number),
+        # rearrange to group patches by spatial location rather than temporal sequence
+        # Only apply this fix when keepdim=True (for flattening compatibility)
+        if (
+            len(out.shape) == 5
+            and out.shape[1] > 1
+            and input.shape[0] % 2 == 0
```
blaz-r: I don't think this check here is a good way to test for temporal data. It then fails on masks since they have shape of 1.

torchgeo/trainers/change.py
hkristen: Sure, I just added changevit to the already existing fcsiam. Feel free to add more where it is needed.

torchgeo/transforms/transforms.py
hkristen: True. This is a leftover from a debugging session with the original datamodule. Feel free to remove it.

blaz-r:
The model is not training with this new datamodule. Did it work for you? I believe the problem might be in `_RandomNCrop` because it doesn't ensure that masks are aligned with image crops?

hkristen:
> The model is not training with this new datamodule. Did it work for you? I believe the problem might be in `_RandomNCrop` because it doesn't ensure that masks are aligned with image crops?

I haven't tried a longer training yet. Only 2-3 epochs to see if the validation works. Though I did create a small script to visually evaluate that the masks are aligned with the image crops, which worked for me. 

grafik


If you are willing to investigate please go ahead, I can only continue on this one sometime next week.

blaz-r:
The validation crops work fine, the training one doesn't, I think. If I also use the extract patches in training, then it starts working.

hkristen:
> The validation crops work fine, the training one doesn't, I think. If I also use the extract patches in training, then it starts working.

you are right, i only checked the validation patches 🤦 here is one example from training:
grafik

adamjstewart:
Can we move all the datamodule/transform stuff to a separate PR so that this PR is focused solely on ChangeViT?

hkristen:
I had another look, changed back to using Kornia and my training now works. Please verify @blaz-r 

@adamjstewart we would need to move the changes in those files to a new PR and then after is merged continue with the ChangeViT PR as some tests depend on it:
```
torchgeo/datamodules/__init__.py
tests/conf/levircdplus.yaml
torchgeo/datamodules/__init__.py
torchgeo/datamodules/levircd.py
torchgeo/transforms/transforms.py
```

calebrob6:
@hkristen this is looking great! Any updates on the training comparison?

hkristen:
Thanks @calebrob6  - I did one training run before my holiday, which went well but didn't fully re-produce the paper results. I will go through the latest changes proposed by @isaaccorley next week and then do another training run. Should be closer then I guess, will report back here.

torchgeo/models/changevit.py
hkristen: good idea!

torchgeo/models/changevit.py
hkristen: Good point! Changed all projections to use learnable 1x1 convs. Also made it work with any backbone via dynamic channel detection.

hkristen:
Status:
I re-based to main as I didn't work on this for quite some time and addressed all comments from @isaaccorley

hkristen:
Training is running well, I get quite good training metrics but the validation scores for F1 and Jaccard are abysmal when I use the hyperparamters reported in the paper. I tried a few different loss functions and learning rates over the weekend but the result is very similar. 

image
Screenshot from 2025-10-06 12-53-35
Screenshot from 2025-10-06 12-54-02

I suspect that the issue is still somewhere in the validation augmentation of the Levir-CD dataset but I am kinda stuck. Can someone else try and have a look? @isaaccorley or @calebrob6 maybe?

hkristen:
I finally found the culprit with the help of Claude Code / Sonnet 4.5.

[Commit 83491ac ](ttps://github.com/torchgeo/torchgeo/commit/83491accbd3134aff1f010054a015dfe3644a379) by @isaaccorley 
 refactored the datamodule to use standard Kornia transforms (RandomCrop + ExtractPatches) instead of the custom _RandomNCrop. This was a good cleanup, but introduced a train/val distribution mismatch:
  - Training: RandomCrop(256) samples patches at random positions each epoch
  - Validation: ExtractPatches extracts the same 16 fixed grid positions every time

The model learned position-dependent features from training and failed on the fixed grid positions it never saw. This wasn't model overfitting - training metrics were correct at  85% F1.

To test this I evaluated the same checkpoint with RandomCrop instead of ExtractPatches jumped and the F1 score jumped from 11.7% → 81.6% F1 instantly.

*This was fixed by*: Added val_patch_sampling parameter ('random' by default) to match train/val distributions while preserving the option for grid-based evaluation when needed.

I started a training with paper-exact hyperparameters + this fix → looks really good already, will report results after overnight run.

hkristen:
## Reproduce paper results

  ### Update: DINOv3 Results

  After additional experimentation with DINOv3 backbones and optimized hyperparameters, I achieved **90.88% F1** - only **1.1 percentage points** from the paper's target!

  | Model | Best Val F1 | Best Step | Gap to Paper | Notes |
  |-------|-------------|-----------|--------------|-------|
  | **Paper Target** | **91.98%** | - | - | Target |
  | **DINOv3-Large-SAT** | **90.88%** | 61,000 | **-1.10 pp** | Satellite pretraining (323M params) |
  | DINOv2-Small | 86.05% | 60,480 | -5.93 pp | Natural image pretraining (25M params) |
  | DINOv3-Small | 85.29% | ~4,700 | -6.69 pp | Natural image pretraining |

  **This means**
  - Satellite-specific pretraining (DINOv3-Large-SAT) provides ** roughly +5% gain** over dinov2-small
  - Config: `tests/conf/changevit_levircd_dinov3_large_sat_optimized.yaml`
  - how the ChangeViT achieved those 91.98% F1 is a miracle to me, but I guess i would just need to tune the dinov3 variant more for that

  ### Initial DINOv2 baseline

  The overnight training run completed successfully, confirming the validation strategy fix works as expected 🚀

  Results:
  - Val F1: 86.05% (peak at step 60,480)
  - Val Jaccard: 69.45%
  - Val Accuracy: 98.61%

  ## Split out dataset/transforms stuff
  I did split this out to #3028 - once this is merged and I get some nice reviews on this PR we can merge finally .:)

torchgeo/trainers/change.py
```diff
@@ -298,6 +358,7 @@ def _shared_step(self, batch: Any, batch_idx: int, stage: str) -> Tensor:
                 and hasattr(self.logger.experiment, 'add_figure')
             ):
                 datamodule = self.trainer.datamodule
+
```
isaaccorley:
```suggestion
```

torchgeo/trainers/change.py
```diff
@@ -377,6 +438,10 @@ def predict_step(
         x = batch['image']
         if model == 'unet':
             x = rearrange(x, 'b t c h w -> b (t c) h w')
+        elif not model.startswith('fcsiam') and not model.startswith('changevit'):
+            x = rearrange(x, 'b t c h w -> b (t c) h w')
+
+        # Forward pass
```
isaaccorley:
```suggestion
```

torchgeo/trainers/change.py
```diff
@@ -267,12 +325,14 @@ def _shared_step(self, batch: Any, batch_idx: int, stage: str) -> Tensor:
...
 
-        if not model.startswith('fcsiam'):
+        if not model.startswith('fcsiam') and not model.startswith('changevit'):
             x = rearrange(x, 'b t c h w -> b (t c) h w')
 
         if self.hparams['task'] == 'multiclass':
             y = y.squeeze(1)
+            y = y.long()
 
+        # Forward pass
```
isaaccorley:
```suggestion
```

torchgeo/trainers/change.py
```diff
@@ -232,6 +247,41 @@ def configure_models(self) -> None:
...
+                        'Expected None or a ChangeViT_Weights enum.'
+                    )
+                self.model = changevit_large(weights=weights)
+            case 'changevit_large_dinov3_sat':
+                if weights is not None and not isinstance(weights, WeightsEnum):
+                    raise ValueError(
+                        f'Invalid weights for changevit_large_dinov3_sat: {weights}. '
+                        'Expected None or a ChangeViT_Weights enum.'
+                    )
+                self.model = changevit_large_dinov3_sat(weights=weights)
```
isaaccorley:
```suggestion
            case 'changevit_small':
                self.model = changevit_small(weights=weights)
            case 'changevit_tiny':
                self.model = changevit_tiny(weights=weights)
            case 'changevit_small_dinov3':
                self.model = changevit_small_dinov3(weights=weights)
            case 'changevit_large':
                self.model = changevit_large(weights=weights)
            case 'changevit_large_dinov3_sat':
                self.model = changevit_large_dinov3_sat(weights=weights)
```

I think it makes more sense to just build this `weights` type validation into the `_create_changevit` function directly in `torchgeo/models/changevit.py`, otherwise during testing we will have to test all these conditional branches instead of just testing it once directly.

tests/models/test_changevit.py
```diff
@@ -0,0 +1,530 @@
+#!/usr/bin/env python3
```
isaaccorley:
```suggestion
# Copyright (c) TorchGeo Contributors. All rights reserved.
# Licensed under the MIT License.
```

tests/models/test_changevit.py
```diff
@@ -0,0 +1,530 @@
...
+
+"""
+ChangeViT model tests.
+
+Based on the paper "ChangeViT: Unleashing Plain Vision Transformers for Change Detection"
+https://arxiv.org/abs/2406.12847
+"""
```
isaaccorley:
```suggestion
"""ChangeViT model tests."""
```

tests/models/test_changevit.py
```diff
@@ -0,0 +1,530 @@
...
+        expected_dcm_params = 2.7e6
+        tolerance_pct = 10.0  # 10% tolerance for ResNet18 implementation variations
+
+        rel_error_pct = (
+            abs(dcm_params - expected_dcm_params) / expected_dcm_params * 100
+        )
+        assert rel_error_pct < tolerance_pct, (
+            f'Paper reports {expected_dcm_params / 1e6:.2f}M params, got {dcm_params / 1e6:.2f}M '
+            f'(relative error: {rel_error_pct:.1f}%, tolerance: {tolerance_pct}%)'
+        )
```
isaaccorley:
```suggestion
```

I don't think these tests are necessary to run in CI on every commit tbh

tests/models/test_changevit.py
```diff
@@ -0,0 +1,530 @@
...
+            'encoder',
+            'detail_capture',
+            'feature_injector',
+            'decoder',
+        ]
+
+        for comp_name in components_to_check:
+            component = getattr(model, comp_name)
+            has_grad = any(p.grad is not None for p in component.parameters())
+            assert has_grad, f'Component {comp_name} should receive gradients'
```
isaaccorley:
```suggestion
```

I don't think these tests need to run in CI on every commit

tests/models/test_changevit.py
```diff
@@ -0,0 +1,530 @@
...
+
+        batch_sizes = [1, 2, 4, 8]
+
+        for bs in batch_sizes:
+            x = torch.randn(bs, 2, 3, 256, 256)
+
+            with torch.no_grad():
+                output = model(x)
+
+            assert output.shape[0] == bs, f'Batch dimension should be {bs}'
```
isaaccorley:
```suggestion
```

These tests aren't needed

tests/models/test_changevit.py
```diff
@@ -0,0 +1,530 @@
...
+        # Should be ready for future WeightsEnum integration
+        # (This test ensures the API is compatible)
+        try:
+            # This will fail now but shows the intended API
+            from torchgeo.models.changevit import ChangeViT_Weights
+
+            _ = changevit_small(weights=ChangeViT_Weights.DEFAULT)
+        except (ImportError, AttributeError):
+            # Expected to fail until weights are available
+            pass
```
isaaccorley:
```suggestion
```

These tests shouldn't be performed here and instead be added to the torchgeo change detection trainer tests.

torchgeo/models/changevit.py
```diff
@@ -0,0 +1,572 @@
...
+        target_size = (x.shape[-2], x.shape[-1])  # (H, W) from original input
+        change_logits: Tensor = F.interpolate(
+            c12, size=target_size, mode='bilinear', align_corners=False
+        )
+
+        # Return raw logits - sigmoid/softmax handled by loss function
+        return change_logits
+
+
+class ChangeViT_Weights(WeightsEnum):  # type: ignore[misc]
```
isaaccorley: Weights Enums are for models pretrained on remote sensing / satellite imagery. these just seem to be using timm imagenet pretrained weights. I would recommend we just create a single `changevit` function that takes in a select backbones that are validated to work within the architecture. In other words, we can remove all this enum code as well as the `changevit_small`, `changevit_tiny`, etc. because it seems we are just hardcoding timm backbones into the models

isaaccorley (changes requested):
A few thoughts on this PR:

- model tests should be redone to match the style of other model tests. consider that each test added adds more time to CI so we should try to minimize the number of tests that will actually reach 100% coverage.

- trainer tests should go in the existing change detection trainer test scripts

- because we don't have any actual pretrained weights for this model, we should not need a WeightsEnum and this should be written similarly to FarSeg and ChangeStar https://github.com/torchgeo/torchgeo/blob/main/torchgeo/models/changestar.py
- because the model size is dictated by the timm backbone, it seems unecessary to have separate `changevit_XXX` functions e.g. `changevit_small`. I would recommend just maintaining a single `changevit` function.

hkristen:
Thanks for your comments @isaaccorley - I will factor those in tomorrow. 

One more thing: I did start a training run with the config `tests/conf/changevit_levircd_dinov3_large_sat_optimized.yaml` to see what effect the changes made by @adamjstewart have on the training metrics. They are quite bad now, see screenshot below (Purple = before changes from main where merged, Turqouise = current state of the ChangeViT branch)
image

I suspect that changing back to dice loss vs. dice+bce before and using only centercrop are the culprits. I get it, why those changes where made but I don't think it is ideal that paper results can't be reproduced with the current state of the model & levircd dataset. 

My question is how to proceed from here? As I guess this will partly be resolved by the outcomes of https://github.com/torchgeo/torchgeo/issues/3108.

adamjstewart:
Can you confirm if loss or transforms are the culprit? Adding another loss is easy (would prefer to do so in a separate PR though). Reverting the centercrop change is harder, and I don't immediately see why that would matter.

calebrob6:
Few things:

@hkristen -- heroic effort here re-implementing and verifying from scratch, that's awesome!!! I'm really excited to have changevit in torchgeo. Nice job!

Do you mean https://github.com/torchgeo/torchgeo/blob/7ba3c692c07d78a425adced27ff7fa841c211bc1/torchgeo/datamodules/levircd.py#L64 is messing up your metrics? I agree that only evaluating test metrics over center crops because the images are rectangular isn't the way to go.

hkristen:
> Can you confirm if loss or transforms are the culprit? Adding another loss is easy (would prefer to do so in a separate PR though). Reverting the centercrop change is harder, and I don't immediately see why that would matter.

**Loss is not the main culprit.** I ran a diagnostic today with bce+dice loss (recovered from commit 576d622):
- Post-merge with dice loss: 25.19% F1
- Post-merge with bce+dice loss: ~33% F1 
- Baseline (October): 86.92% F1

> Few things:
> 
> @hkristen -- heroic effort here re-implementing and verifying from scratch, that's awesome!!! I'm really excited to have changevit in torchgeo. Nice job!

Thanks @calebrob6 it honestly was a bit more effort than anticipated but once you start you need to finish it right. I am also excited to hopefully be done with it soon :P 

> Do you mean
> 
> https://github.com/torchgeo/torchgeo/blob/7ba3c692c07d78a425adced27ff7fa841c211bc1/torchgeo/datamodules/levircd.py#L64
> is messing up your metrics? I agree that only evaluating test metrics over center crops because the images are rectangular isn't the way to go.

I believe this ties back to what I "discovered" before in this PR: https://github.com/torchgeo/torchgeo/pull/2910#issuecomment-3381130492 and was changed to CenterCrop for val & test after this discussion https://github.com/torchgeo/torchgeo/pull/3028#discussion_r2494397967. However, the current diagnostic suggests that my original hypothesis was correct - the train/val distribution mismatch significantly impacts metrics. I am sure someone smarter than me can explain why :P 

I know that @adamjstewart wants to be done with this, latest Thursday to be ready for the 0.8 release, so I am suggest a pragmatic approach:
- We already know that ChangeViT works and produces good results on LevirCD with the same transforms as in the paper, as shown in https://github.com/torchgeo/torchgeo/commit/576d622ac28d5d9e6f28483af7e8bedbc37bfddf
- So I guess we can assume the model is ready and I can address @isaaccorley review comments without fixating on getting the metrics right.
- We have a separate discussion/PR after 0.8 is released to address the LEVIRCD validation strategy issue.


What do you guys think?

adamjstewart:
Yes, the model definitely seems good to go. Will give one last review and merge tomorrow if everything has been taken care of. Would love to better understand/figure out the val/test metrics in a separate PR.

I'm still very confused about the test metrics. We could easily switch to RandomCrop during validation, but it isn't obvious to me why that would result in better performance. The model is seeing the entire image during training (including the center). Does the center of the image somehow have a different distribution from the rest of the image (e.g., object of interest is always near the middle)?

Other idea is to set `patch_size=1024` by default so we basically always train and evaluate on the entire image.

hkristen:
Okidoki - I tried to address all review comments from @isaaccorley and started a full overnight training run. Let's see how it goes. 

Let's figure out the rest later

calebrob6:
I'm guessing the minimum tests are failing because dinov3 was very recent version of timm

adamjstewart:
Fine with bumping min timm or conditional testing based on timm version.

hkristen:
I did follow @adamjstewart 's suggestion and started a training with patch_size=1024 ->and miraculously F1 > 80% after 10 epochs! 🎉

Configuration:
- patch_size=1024 (full images)
- batch_size=1
- accumulate_grad_batches=12 (effective batch size: 12)
- precision=16-mixed (fp16 to fit large model in memory)
- Model: changevit with backbone='large_dinov3_sat'
- Loss: dice

Curious WHY this happens I checked the change pixel distribution in the 100 training & 64 validation images:

Full image:
- train = 7.75% change pixels
- val = 4.20% change pixels

Center crop 256×256:
- train = 7.87% change pixels
- val = 3.17% change pixels


 So when using CenterCrop on the validation set, there are ~24% fewer change pixels in the center region compared to the full image → this creates a train/val distribution mismatch that seems to be the reason for the poor validation/test performance.

I created heatmaps for both validation and training datasets to visualize this spatial bias - notice how the validation split (bottom) has visibly darker/fewer changes in the center compared to the edges, while the train split (top) has uniform distribution.

levircd_train_distribution
levircd_val_distribution

adamjstewart:
A lot of remaining issues that need to be addressed:

- [ ] No TorchGeo datasets or models should download anything by default (security, principle of least surprise, etc.)
- [ ] We don't usually add helper functions (changevit) unless there are pretrained weights
- [ ] We should export and use the `ChangeViT` class instead
- [ ] Currently this class requires you to input all other models, could be changed to input backbone str instead
- [ ] Tests are using timm backbones from a month ago, no reason we can't use ancient simple ViT tiny model instead
- [ ] Missing a lot of test coverage
- [ ] Missing documentation

Some of these are easy, some are hard. I would really like to get out the release today, but this model is currently quite far from a standard TorchGeo model. I'll check back in after reviewing a few other PRs to see if this is any closer. I can help with some things, but there's a lot of work remaining here.

tests/trainers/test_change.py
```diff
@@ -200,36 +219,44 @@ def test_freeze_backbone(
             [param.requires_grad is False for param in model.model.encoder.parameters()]
         )
         assert all([param.requires_grad for param in model.model.decoder.parameters()])
-        assert all(
-            [
-                param.requires_grad
-                for param in model.model.segmentation_head.parameters()
```
adamjstewart, line 190: Not all models have a `segmentation_head`. Maybe we should rename ChangeViT to clarify which part of the decoder is the `segmentation_head`?

tests/trainers/test_change.py
```diff
@@ -242,6 +269,38 @@ def test_btc_model(self) -> None:
...
+            model='changevit',
+            backbone='vit_tiny_patch16_224',
+            task='multiclass',
+            num_classes=5,
+        )
+        x = torch.randn(1, 2, 3, 256, 256)
+        y = model(x)
+        assert y.shape == (1, 5, 256, 256)
+
+    def test_changevit_predict(self) -> None:
```
adamjstewart, line 281: Technically none of these are needed, but I'll simplify the tests another day.

torchgeo/models/changevit.py
```diff
@@ -0,0 +1,339 @@
...
+       * Training: 80k steps with batch size 48
+       * Image size: 256x256 patches
+
+    .. versionadded:: 0.8
+    """
+
+    def __init__(
+        self,
+        backbone: str,
+        img_size: int = 256,
```
adamjstewart, line 244: At the moment, it isn't possible to use `ChangeViT` with `ChangeDetectionTask` without a 256 x 256 input. None of our other decoders require this. I wonder if we can make this more flexible like SMP.

torchgeo/trainers/change.py
adamjstewart, line 262: @hkristen I removed the logic that froze the feature_injector and detail_capture components but not the entire decoder. It would be nice to keep the same structure as most SMP decoders so we don't have to do anything custom here. Let me know if this is a big problem.

adamjstewart (dismissed):
LGTM with the following caveats. Would like to hear from @hkristen before merging to make sure I didn't break anything, but I think we have 100% test coverage finally.

tests/models/test_changevit.py
```diff
@@ -0,0 +1,116 @@
...
+            backbone='vit_tiny_patch16_224', img_size=64, pretrained=False
+        )
+
+        assert hasattr(model, 'encoder')
+        assert hasattr(model, 'detail_capture')
+        assert hasattr(model, 'feature_injector')
+        assert hasattr(model, 'decoder')
+
+
+class TestDetailCaptureModule:
```
isaaccorley, line 60: nit: Do we even need these tests? 256x256 is also a large image size for testing

tests/models/test_changevit.py
```diff
@@ -0,0 +1,116 @@
...
+        assert c3.shape[1] == 128
+        assert c4.shape[1] == 256
+
+        # Check spatial dimensions (1/2, 1/4, 1/8 of input)
+        assert c2.shape[-2:] == (128, 128)  # 256 / 2
+        assert c3.shape[-2:] == (64, 64)  # 256 / 4
+        assert c4.shape[-2:] == (32, 32)  # 256 / 8
+
+
+class TestFeatureInjector:
```
isaaccorley, line 80: nit: Same as above comment

tests/models/test_changevit.py
```diff
@@ -0,0 +1,116 @@
...
+            torch.randn(2, 256, 16, 16),  # 1/8 scale
+        )
+
+        enhanced_feats = injector(vit_feats, detail_feats)
+
+        # Output should have same shape as input ViT features
+        assert enhanced_feats.shape == vit_feats.shape
+
+
+class TestChangeViTDecoder:
```
isaaccorley, line 103: nit: Same as above comment

tests/trainers/test_change.py
isaaccorley: Agreed on this. All smp models have a `segmentation_head` but not all possible segmentation models we add will have this.

I actually like timm's implementation that adds a `get_classifier` method that returns whatever the custom name of the head is 

```python
def get_classifier(self):
    # Return whatever module acts as the classifier
    return self.head
```

tests/trainers/test_change.py
isaaccorley: Might be good to remove this tbh, the less forward passes we do in tests to achieve 100% coverage the better

torchgeo/models/changevit.py
isaaccorley: Why is this? If we pass `dynamic_img_size=True` to `timm.create_model` this actually allows for dynamic image sizes by resampling the pos embedding on the fly. Seems like this is an issue because the decoder is fixed because it's a ViT Decoder.

isaaccorley (dismissed):

tests/models/test_changevit.py
adamjstewart: I removed all of these tests and confirmed that `TestChangeViT` is sufficient for 100% coverage. But also, all of those extra tests take 0.25 s, so it doesn't really hurt us much either.

adamjstewart:
I'm not worried about the tests. All of the trainer tests are verbose and could be cleaned up to save time. We can do this whenever without worrying about backwards-incompatible changes.

My main concern is with the model. Does it have all the basic features we need? As long as we have support for dynamic image resizing, we should be good. Let's just make sure there's nothing user-facing we want to change.

adamjstewart (approved):
Let's merge, hopefully any improvements needed will be backwards-compatible.

hkristen:
> LGTM with the following caveats. Would like to hear from @hkristen before merging to make sure I didn't break anything, but I think we have 100% test coverage finally.

After running my usual test training over 5000 steps with the legacy dinov3 config, I can confirm that the model converges like before.

torchgeo/trainers/change.py
hkristen: Works for me, consistent behavior across all models is cleaner.

torchgeo/models/changevit.py
hkristen: I didn't notice this earlier because the DINOv3 backbones I was testing on, have strict_img_size=False by default. So they worked at any size. I've opened a PR to add dynamic_img_size=True which enables flexible input sizes for all ViT backbones (tested with standard vit_small_patch16_224 at 256, 512, and 1024): https://github.com/torchgeo/torchgeo/pull/3133

PR #2913, opened 2025-08-01
Add OpenEarthMap-SAR - DFC2025 baseline U-Net weights

isaaccorley:
Adds the DFC2025 Track 1 Baseline U-Net weights trained by @cliffbb in the [baseline repo](https://github.com/cliffbb/DFC2025-OEM-SAR-Baseline/). The model is a land cover land-use semantic segmentation model trained on the [OpenEarthMap-SAR dataset](https://zenodo.org/records/14622048) which contains high-resolution 1-channel uint8 SAR images from [Umbra Space
](https://registry.opendata.aws/umbra-open-data/) particularly the [Geo-Ellipsoid Corrected (GEC)](https://help.umbra.space/product-guide/umbra-products/umbra-product-specifications/geo-ellipsoid-corrected-gec-geotiff) geotiffs.

The model has an IoU of 0.41 on the private test set leaderboard
image

Before @calebrob6 asks, I've checked that it runs and it's decent but seems to get confused by the noisiness in the SAR imagery. Below is a minimum working example:

```python
import random
import glob

import torch
import torchvision.transforms.v2.functional as F
import rasterio
from matplotlib.colors import to_rgb, ListedColormap
from torchgeo.models import Unet_Weights, unet

h = w = 1024
images = sorted(glob.glob("/data/train/sar_images/*.tif"))
labels = sorted(glob.glob("/data/train/labels/*.tif"))

device = torch.device("cuda")
dtype = torch.float32
weights = Unet_Weights.UMBRA_GEC_OPENEARTHMAP_SAR
classes = weights.meta["classes"]
colormap = ListedColormap([to_rgb(hex) for hex in weights.meta["colormap"]])

transforms = weights.transforms
model = unet(weights)
model.eval()
transforms.eval()
model = model.to(device).to(dtype)
transforms = transforms.to(device).to(dtype)

idx = random.randint(0, len(images) - 1)
path = images[idx]
label = labels[idx]
    image = torch.from_numpy(src.read())
    image = F.center_crop(image, (h, w))
    x = image.clone().unsqueeze(dim=0).to(device).to(dtype)
    image = image.repeat(3, 1, 1)

with rasterio.open(label) as src:
    mask = torch.from_numpy(src.read())
    mask = F.center_crop(mask, (h, w))
    mask = mask.squeeze(dim=0).numpy()

x = transforms(dict(image=x))["image"]
preds = model(x).cpu().squeeze().argmax(dim=0)

fig, axs = plt.subplots(1, 3, figsize=(12, 5))
axs[0].imshow(image.permute(1, 2, 0).numpy())
axs[0].axis('off')
axs[1].imshow(mask, vmin=0, vmax=len(classes), cmap=colormap, interpolation='none')
axs[1].axis('off')
axs[2].imshow(preds, vmin=0, vmax=len(classes), cmap=colormap, interpolation='none')
axs[0].set_title('Image')
axs[1].set_title('Mask')
axs[2].set_title('Predictions')
```

image
image
image
image

### DIFF
--- a/docs/api/models.rst
+++ b/docs/api/models.rst
@@ -175,14 +175,14 @@ NAIP
    :file: weights/naip.csv
 
 
-Sentinel-1
-----------
+Synthetic Aperture Radar (SAR)
+------------------------------
 
 .. csv-table::
    :widths: 45 10 10 10 10
    :header-rows: 1
    :align: center
-   :file: weights/sentinel1.csv
+   :file: weights/sar.csv
 
 
 Sentinel-2
--- a/docs/api/weights/sar.csv
+++ b/docs/api/weights/sar.csv
@@ -2,6 +2,9 @@ Weight,Channels,Source,Citation,License
 ResNet50_Weights.SENTINEL1_GRD_DECUR, 2,`link <https://github.com/zhu-xlab/DeCUR>`__,`link <https://arxiv.org/abs/2309.05300>`__,"Apache-2.0"
 ResNet50_Weights.SENTINEL1_GRD_MOCO, 2,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,"CC-BY-4.0"
 ResNet50_Weights.SENTINEL1_GRD_SOFTCON, 2,`link <https://github.com/zhu-xlab/softcon>`__,`link <https://arxiv.org/abs/2405.20462>`__,"CC-BY-4.0"
+Swin_V2_B_Weights.SENTINEL1_MI_SATLAS,2,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,ODC-BY
+Swin_V2_B_Weights.SENTINEL1_SI_SATLAS,2,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,ODC-BY
+Unet_Weights.UMBRA_GEC_OPENEARTHMAP_SAR, 1,`link <https://github.com/cliffbb/DFC2025-OEM-SAR-Baseline/>`__,`link <https://arxiv.org/abs/2501.10891>`__,"CC-BY-4.0"
 ViTSmall16_Weights.SENTINEL1_GRD_MAE,2,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,"CC-BY-4.0"
 ViTSmall16_Weights.SENTINEL1_GRD_FGMAE,2,`link <https://github.com/zhu-xlab/FGMAE>`__,`link <https://arxiv.org/abs/2310.18653>`__,"CC-BY-4.0"
 ViTBase16_Weights.SENTINEL1_GRD_MAE,2,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,"CC-BY-4.0"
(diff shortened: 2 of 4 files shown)


calebrob6:
I see my job here is not needed

calebrob6 (approved):

docs/api/models.rst
```diff
@@ -175,14 +175,14 @@ NAIP
    :file: weights/naip.csv
 
 
-Sentinel-1
-----------
+Synthetic Aperture Radar (SAR)
+------------------------------
```
adamjstewart, line 179: I disagree with this change. There are many different kinds of SAR (C-band vs. L-band) just like there are many different kinds of optical imagery (Sentinel-2, Landsat, etc.). Just because two models have the same number of bands doesn't mean they will be compatible with any SAR imagery.

docs/api/weights/sar.csv
```diff
@@ -2,6 +2,9 @@ Weight,Channels,Source,Citation,License
 ResNet50_Weights.SENTINEL1_GRD_DECUR, 2,`link <https://github.com/zhu-xlab/DeCUR>`__,`link <https://arxiv.org/abs/2309.05300>`__,"Apache-2.0"
 ResNet50_Weights.SENTINEL1_GRD_MOCO, 2,`link <https://github.com/zhu-xlab/SSL4EO-S12>`__,`link <https://arxiv.org/abs/2211.07044>`__,"CC-BY-4.0"
 ResNet50_Weights.SENTINEL1_GRD_SOFTCON, 2,`link <https://github.com/zhu-xlab/softcon>`__,`link <https://arxiv.org/abs/2405.20462>`__,"CC-BY-4.0"
+Swin_V2_B_Weights.SENTINEL1_MI_SATLAS,2,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,ODC-BY
```
adamjstewart, line 5: Let's move sorting to a separate PR to make this easier to review.

torchgeo/models/unet.py
```diff
@@ -39,6 +39,46 @@ class Unet_Weights(WeightsEnum):  # type: ignore[misc]
...
+        transforms=K.AugmentationSequential(
+            K.Normalize(mean=torch.tensor(0), std=torch.tensor(255)), data_keys=None
+        ),
+        meta={
+            'dataset': 'OpenEarthMap-SAR',
+            'in_chans': 1,
+            'num_classes': 9,
+            'model': 'U-Net',
+            'encoder': 'efficientnet-b4',
+            'classes': (
```
adamjstewart, line 53: This is a lot more metadata than other models. Is this coming from STAC MLM?

torchgeo/models/unet.py
```diff
@@ -39,6 +39,46 @@ class Unet_Weights(WeightsEnum):  # type: ignore[misc]
...
+                '#949494',
+                '#FFFFFF',
+                '#226126',
+                '#0045FF',
+                '#4BB549',
+                '#DE1F07',
+            ),
+            'publication': 'https://arxiv.org/abs/2501.10891',
+            'repo': 'https://github.com/cliffbb/DFC2025-OEM-SAR-Baseline',
+            'bands': ['B1'],
```
adamjstewart, line 77: B1 doesn't sound right, is it VV or VH? These are very different polarizations, an incompatible choice could result in the poor performance you are seeing.

docs/api/weights/sar.csv
```diff
@@ -2,6 +2,9 @@ Weight,Channels,Source,Citation,License
...
+Swin_V2_B_Weights.SENTINEL1_SI_SATLAS,2,`link <https://github.com/allenai/satlas>`__,`link <https://arxiv.org/abs/2211.15660>`__,ODC-BY
+Unet_Weights.UMBRA_GEC_OPENEARTHMAP_SAR, 1,`link <https://github.com/cliffbb/DFC2025-OEM-SAR-Baseline/>`__,`link <https://arxiv.org/abs/2501.10891>`__,"CC-BY-4.0"
```
adamjstewart, line 7: Where can I find this license documented?

adamjstewart (changes requested):

torchgeo/models/unet.py
isaaccorley: Most of our weights are foundation models with no class specific outputs. I think it makes sense to add this here for the user to have information about the output order and categories.

torchgeo/models/unet.py
isaaccorley: It's a different level of processing that is used from Umbra where it converted to a grayscale uint8 image. This doesn't use the raw band data unfortunately.

docs/api/models.rst
isaaccorley: That's okay with me but you're not proposing a solution here. Do you prefer that I just make a separate group for "Umbra Space" in this case? Because this isn't Sentinel-1.

docs/api/weights/sar.csv
isaaccorley: The dataset is CC-BY-4.0 licensed so the model is also the same license found here https://registry.opendata.aws/umbra-open-data/

isaaccorley: @cliffbb can you clarify here what license you are releasing the model weights under?

docs/api/models.rst
adamjstewart: Yes, separate group/table for Umbra

PR #2914, opened 2025-08-02
SkyCLIP Models

isaaccorley:
This PR adds the pretrained image ViT image encoders from the paper ["SkyScript: A Large and Semantically Diverse Vision-Language Dataset for Remote Sensing", Wang et al. (AAAI 2024)](https://arxiv.org/abs/2312.12856). These encoders are pretrained using CLIP on the SkyScript dataset to align images to captions. SkyScript contains a variety of RGB images from different sources including those below:

image


docs/api/weights/agnostic.csv
```diff
@@ -7,6 +7,10 @@ DOFALarge16_Weights.DOFA_MAE,`link <https://github.com/zhu-xlab/DOFA>`__,`link <
...
 ResNet50_Weights.FMOW_RGB_GASSL,`link <https://github.com/sustainlab-group/geography-aware-ssl>`__,`link <https://arxiv.org/abs/2011.09980>`__,-,implicit,-,-,,,,,,,,,,,,
 ScaleMAE_ViTLarge16_Weights.FMOW_RGB_SCALEMAE,`link <https://github.com/bair-climate-initiative/scale-mae>`__,`link <https://arxiv.org/abs/2212.14532>`__,CC-BY-NC-4.0,explicit,-,-,,,,,,,,,,,
+ViTBase32_CLIP_Weights.SKYCLIP_RGB_TOP50PCT,`link <https://github.com/wangzhecheng/SkyScript>`__,`link <https://arxiv.org/abs/2312.12856>`__,MIT,implicit,-,-,,,,,,,,,,,
+ViTLarge14_CLIP_Weights.LAIONRS_CLIP_RGB,`link <https://github.com/wangzhecheng/SkyScript>`__,`link <https://arxiv.org/abs/2312.12856>`__,MIT,implicit,-,-,,,,,,,,,,,
+ViTLarge14_CLIP_Weights.SKYCLIP_RGB_TOP30PCT,`link <https://github.com/wangzhecheng/SkyScript>`__,`link <https://arxiv.org/abs/2312.12856>`__,MIT,implicit,-,-,,,,,,,,,,,
+ViTLarge14_CLIP_Weights.SKYCLIP_RGB_TOP50PCT,`link <https://github.com/wangzhecheng/SkyScript>`__,`link <https://arxiv.org/abs/2312.12856>`__,MIT,implicit,-,-,,,,,,,,,,,
```
adamjstewart, line 13: Code is MIT, where do you see the license for the weights?

torchgeo/models/vit.py
```diff
@@ -40,10 +40,17 @@
     data_keys=None,
 )
 
-# https://github.com/pytorch/vision/pull/6883
-# https://github.com/pytorch/vision/pull/7107
-# Can be removed once torchvision>=0.15 is required
-Weights.__deepcopy__ = lambda *args, **kwargs: args[0]
```
adamjstewart, line 46: Can we remove this from all files in a separate PR?

docs/api/weights/agnostic.csv
isaaccorley: I don't think weights license is defined anywhere. Seems like the datasets trained on are a mix of OSM data and several public sources of imagery. Would this make the weights license inherit from the dataset sources then?

isaaccorley: @wangzhecheng Can you comment here about what license you are releasing the SkyCLIP models under?

PR #2915, opened 2025-08-02
MOSAIKS alias for RCF

isaaccorley:
It seems to be confusing to users that MOSAIKS == RCF so we should make an alias for this. It shows up in the docs like this now 

image

### DIFF
--- a/docs/api/models.rst
+++ b/docs/api/models.rst
@@ -69,9 +69,6 @@ MOSAIKS
 ^^^^^^^
 
 .. autoclass:: MOSAIKS
-
-RCF Extractor
-^^^^^^^^^^^^^
 .. autoclass:: RCF
 
 ResNet
--- a/torchgeo/models/rcf.py
+++ b/torchgeo/models/rcf.py
@@ -211,6 +211,8 @@ class MOSAIKS(RCF):
     .. note::
 
        This Module is *not* trainable. It is only used as a feature extractor.
+
+    .. versionadded:: 0.8
     """
 
     def __init__(


adamjstewart:
Curious what @estherrolf thinks. I think RCF is just a random name @calebrob6 chose. It does seem that MOSAIKS is the better known name in the literature.

I would be in favor of renaming RCF to MOSAIKS and deprecating the old name like so:
```python
from typing_extensions import deprecated

@deprecated('Use torchgeo.models.MOSAIKS instead')
class RCF(MOSAIKS):
    """Deprecated alias of MOSAIKS."""
```
Then we would only show MOSAIKS in the docs.

calebrob6:
RCFs (random convolutional features) isn't something I made up -- it's on page 4 of the MOSAIKS paper. I would say RCFs is the name for the feature extraction step while MOSAIKS is the whole pipeline ("Multi-task Observation using Satellite Imagery and Kitchen Sinks").

I agree that we should have a MOSAIKS alias though as that is what has "caught on" (esp. given the references in the AlphaEarth paper).

All that said, I defer to whatever Esther wants to name it!

estherrolf:
Hi all! I remember weighing in on this a while back, let me clarify what I think is most accurate:
- Random Convolutional Features (RCF) is the technical name of the embedding method, where the random patches can be drawn from the patch distribution from the image dataset ("empirical") or randomly generated (e.g. from a Gaussian distribution). In practice, I've always seen "empirical" do better, that's why we use empirical features in the MOSAIKS paper (Nature Comms 2021, see also the github repo for the project which has been public since 2021).
- This is why it's kindof confusing: MOSAIKS as a _prediction method_ is essentially RCF ("empirical") + ridge regression. The MOSAIKS _embedding method_ == RCF ("empirical"). 
- That said, I don't think it would be wrong to call the Gaussian RCF a "MOSAIKS embedding", but it is the lesser performing variant and thus any benchmarks that use MOSAIKS with Gaussian RCF implementation should clearly state that, because it's not our standard implementation from the paper.

Now to weigh in on the question at hand: since MOSAIKS _embeddings_ are RCF ("empirical") embeddings, would it be possible for torchgeo.models.MOSAIKS not to point to the current RCF implementation with default args, but something that better reflects what we did? So, something like the following, where the defaults actually reflect MOSAIKS as it's used in the research community:

`torchgeo.models.RCF(mode='empirical', features=4096, kernel_size = 4)`? 

Of course users should still be tuning features, kernel size, bias, as hyperparameters, but 16 features is **really small** compared to what we usually use in research papers. FWIW I've found kernel size of 3 or 4 is a good starting place for NAIP and S2 imagery at least.

isaaccorley:
Thanks @estherrolf! I've made updates can you have a look @adamjstewart

adamjstewart:
If MOSAIKS is just RCF with different default parameters, I would rather just change the default parameters and deprecate the old version.

isaaccorley:
I think the concern also is that we want to make sure the MOSAIKS is the mode="empirical" version always and not the fully random one mode="gaussian", for example how it was used in the AlphaEarth paper.

adamjstewart:
I see. I think this makes sense. As long as @estherrolf is happy with this, so am I.

docs/api/models.rst
```diff
@@ -65,9 +65,13 @@ L-TAE
 
 .. autoclass:: LTAE
 
+MOSAIKS
+^^^^^^^
+
+.. autoclass:: MOSAIKS
+
 RCF Extractor
 ^^^^^^^^^^^^^
```
adamjstewart: Maybe delete this heading so that both are under the MOSAIKS section

torchgeo/models/rcf.py
```diff
@@ -199,3 +199,45 @@ def forward(self, x: Tensor) -> Tensor:
...
+class MOSAIKS(RCF):
+    """MOSAIKS RCF model with the recommended parameters defined in the paper.
+
+    If you use this model in your research, please cite the following paper:
+
+    * https://www.nature.com/articles/s41467-021-24638-z
+
+    .. note::
+
+       This Module is *not* trainable. It is only used as a feature extractor.
```
adamjstewart, line 213: Needs a `versionadded`

isaaccorley:
> Not really convinced of the value of two copies of the same class with different default parameters, but happy to better advertise MOSAIKS support.

Agree, ideally we just deprecate RCF but I think it makes sense to not have a MOSAIKS(mode="gaussian") to avoid confusion.

estherrolf:
Thanks! If we just want one class, it would be fine to have a MOSAIKS(mode="gaussian") option, but the default should be "empirical" for the MOSAIKS class. Maybe the documentation can explain (happy to help with that!)

adamjstewart:
@isaaccorley what do you think?

isaaccorley:
@adamjstewart I think it makes sense to deprecate RCF in favor of MOSAIKS and change the default parameters if this would end up being in the 1.0 release anyway.

adamjstewart:
Yes, 0.8 will almost certainly be 1.0, now is a good time to do this if we want to do this.

isaaccorley:
> Yes, 0.8 will almost certainly be 1.0, now is a good time to do this if we want to do this.

Will make another PR for this later!

PR #2921, opened 2025-08-06
PASTIS: add data module, SITS compatibility

yichiac:
Related to https://github.com/microsoft/torchgeo/issues/2382
Depends on multiclass time-series semantic segmentation support

- [x]  PASTIS: add new data module
- [x]  SemanticSegmentationTask: change to support T x C x H x W and C x H x W

### DIFF
--- a/tests/datasets/test_utils.py
+++ b/tests/datasets/test_utils.py
@@ -6,7 +6,7 @@ import pickle
 import re
 from datetime import datetime, timedelta
 from pathlib import Path
-from typing import Any, cast
+from typing import Any
 
 import numpy as np
 import pandas as pd
@@ -545,9 +545,6 @@ def test_pad_across_batches() -> None:
         {'image': torch.ones(2, 10, 5, 5), 'mask': torch.zeros(5, 5)},
         {'image': torch.ones(3, 10, 5, 5), 'mask': torch.zeros(5, 5)},
     ]
-
-    with pytest.raises(ValueError, match='padding_length must be specified'):
-        pad_across_batches(batch, padding_value=0.0, padding_length=cast(int, None))
 
     out = pad_across_batches(batch, padding_value=0.0, padding_length=3)
     assert out['image'].shape[1] == 3
--- a/torchgeo/datasets/utils.py
+++ b/torchgeo/datasets/utils.py
@@ -429,7 +429,7 @@ def pad_across_batches(
 
     Args:
         batch: list of sample dicts returned by dataset
-        padding_length: the length to pad the sequences
+        padding_length: the length to pad the sequences to
         padding_value: value for padded elements
 
     Returns:
@@ -437,12 +437,6 @@ def pad_across_batches(
 
     .. versionadded:: 0.8
     """
-    if padding_length is None:
-        raise ValueError(
-            'padding_length must be specified.'
-            'If pad to the max length in the batch, use torch.nn.utils.rnn.pad_sequence directly.'
-        )
-
     output: dict[str, Any] = {}
     images = [sample['image'] for sample in batch]
     feature_shape = images[0].shape[1:]


torchgeo/datamodules/__init__.py
```diff
@@ -34,6 +34,7 @@
 from .naip import NAIPChesapeakeDataModule
 from .nasa_marine_debris import NASAMarineDebrisDataModule
 from .oscd import OSCDDataModule
+from .pastis import PastisDataModule
```
adamjstewart: Needs to be added to `__all__`

tests/conf/pastis.yaml
```diff
@@ -0,0 +1,16 @@
...
+    backbone: 'resnet18'
+    in_channels: 24
+    task: 'binary'
+    in_channels: 12
+    loss: 'bce'
+data:
+  class_path: PastisDataModule
+  init_args:
+    batch_size: 2
+    val_split_pct: 0.5
```
adamjstewart: This is not (yet) a valid argument for the `PastisDataModule`

tests/conf/pastis.yaml
```diff
@@ -0,0 +1,16 @@
+model:
+  class_path: SemanticSegmentationTask
+  init_args:
+    model: 'unet'
+    backbone: 'resnet18'
+    in_channels: 24
```
adamjstewart: `in_channels` is duplicated

torchgeo/datamodules/pastis.py
```diff
@@ -0,0 +1,46 @@
...
+"""PASTIS datamodule."""
+
+from typing import Any
+
+import kornia.augmentation as K
+
+from ..datasets import PASTIS
+from .geo import NonGeoDataModule
+
+class PastisDataModule(NonGeoDataModule):
```
adamjstewart: Since the dataset is PASTIS, I think the data module should be PASTISDataModule

torchgeo/datamodules/pastis.py
```diff
@@ -0,0 +1,46 @@
...
+
+        Args:
+            stage: Either 'fit', 'validate', 'test', or 'predict'.
+        """
+        if stage in ['fit']:
+            self.train_dataset = PASTIS(split='train', **self.kwargs)
+        if stage in ['fit', 'validate']:
+            self.val_dataset = PASTIS(split='val', **self.kwargs)
+        if stage in ['predict']:
+            self.predict_dataset = PASTIS(split='test', **self.kwargs)
```
adamjstewart: PASTIS does not have pre-defined splits, you need to create your own

torchgeo/trainers/segmentation.py
```diff
@@ -100,6 +104,20 @@ class and used with 'ce' loss.
         self.weights = weights
         super().__init__()
 
+    def forward(self, x:Tensor) -> Tensor:
```
adamjstewart: While this is a clever solution, it only works for the currently available models. We also want to add time-series models like Satlas, Conv-LSTM, CNN-LSTM, etc. Satlas is already available and could be added to this PR, the others are coming soon.

torchgeo/trainers/segmentation.py
```diff
@@ -100,6 +104,20 @@ class and used with 'ce' loss.
...
+        """
+        Args:
+            x: Input tensor of shape (B, C, H, W) or (B, T, C, H, W).
+
+        Returns:
+            Output tensor of shape (B, T*C, H, W).
```
adamjstewart: I don't think these output dimensions are correct, isn't it B x "# classes" x H x W?

yichiac:
Thanks, @adamjstewart. I will work on:
- [x] create the train-val-test splits for pastis.

docs/api/datamodules.rst
```diff
@@ -175,6 +175,11 @@ OSCD
 
 .. autoclass:: OSCDDataModule
 
+PASTIS
+^^^^^^^^^^
```
adamjstewart:
```suggestion
^^^^^^
```

torchgeo/trainers/segmentation.py
```diff
@@ -100,6 +104,19 @@ class and used with 'ce' loss.
...
+
+        Args:
+            x: Input tensor of shape (B, C, H, W) or (B, T, C, H, W).
+
+        Returns:
+            Output tensor of shape (B, num_classes, H, W).
+        """
+        if x.ndim == 5:
+            x = rearrange(x, 'b t c h w -> b (t c) h w')
+        return cast(Tensor, self.model(x))
```
```suggestion
        x = self.model(x)
        return x
```

tests/conf/pastis.yaml
```diff
@@ -0,0 +1,15 @@
+model:
+  class_path: SemanticSegmentationTask
+  init_args:
+    model: 'unet'
+    backbone: 'resnet18'
+    task: 'binary'
+    in_channels: 12
```
adamjstewart: `in_channels: 12` is correct, but I think you need to tell the model that it's actually `in_channels * T`

tests/conf/pastis.yaml
```diff
@@ -0,0 +1,15 @@
...
+    loss: 'bce'
+data:
+  class_path: PASTISDataModule
+  init_args:
+    batch_size: 2
+    val_split_pct: 0.5
```
adamjstewart:
```suggestion
    val_split_pct: 0.33
    test_split_pct: 0.33
```
or something like that

tests/conf/pastis.yaml
```diff
@@ -0,0 +1,15 @@
+model:
+  class_path: SemanticSegmentationTask
+  init_args:
+    model: 'unet'
+    backbone: 'resnet18'
+    task: 'binary'
```
adamjstewart: I don't think PASTIS is binary, there are 18 + 1 classes, right?

tests/conf/pastis.yaml
```diff
@@ -0,0 +1,15 @@
...
+    in_channels: 12
+    loss: 'bce'
```
adamjstewart:
```suggestion
    loss: 'ce'
```

tests/conf/pastis.yaml
yichiac: Yes, it's not binary. Multi-label seems to be the right one.

adamjstewart: I don't think it's multilabel either, should be multiclass...

yichiac: You are right, it's multiclass.

tests/conf/pastis.yaml
yichiac: How do we define T in this case?

tests/conf/pastis.yaml
yichiac:
```suggestion
    task: 'multiclass'
```

tests/conf/pastis.yaml
adamjstewart: Hmm, I guess it won't actually be easy to know this in advance... Let's just use `in_channels: C * T`. So basically, expect the user to do the math.

adamjstewart: @keves1 @hkristen this will result in a slight API difference between `SemanticSegmentationTask` and `ChangeDetectionTask`, where the former will use `in_channels: C x T` and the latter will use `in_channels: C` where C is implicitly multiplied by 2 internally. Should we change `ChangeDetectionTask` to match `SemanticSegmentationTask`? Or make one a subclass of the other?

tests/conf/pastis.yaml
```diff
@@ -0,0 +1,16 @@
+model:
+  class_path: SemanticSegmentationTask
+  init_args:
+    model: 'unet'
+    backbone: 'resnet18'
+    task: 'multiclass'
+    in_channels: C * T
```
adamjstewart: This needs to be a number

yichiac:
The remaining issue is caused by Kornia's augmentation. It might be caused by the changing time steps for different modalities (S1A, S1D, S2). I am thinking adding customized `on_after_batch_transfer` to datamodule to pad or truncate to the target time steps. But this would limit the flexibility.

adamjstewart:
One of the TODOs is to add a padding utility for this purpose. But this use case seems particularly complicated since you have to pad 3 different things. Can you confirm that the 3 different band subsets have different length, not just the entire collection? Also, what order are they in? S1A_T1, S1D_T1, S2_T1, S1A_T2, ... or S1A_T1, S1A_T2, ..., S1D_T1?

yichiac:
Each band subset has a different length itself. I checked the time stamps (first element of data shapes) for all modalities. Here are the results:
```
S2
61: 626 times
46: 738 times
43: 537 times
42: 4 times
38: 562 times
37: 1 times

S1A
71: 1819 times
70: 108 times
65: 541 times

S1D
70: 1730 times
69: 738 times
```

yichiac:
The previous error was caused by varying max timestamps across train/val/test splits, which led to mismatched input channel sizes when passing data to the unet. My solution is to pad all samples to a consistent maximum timestamp, ensuring stable input shapes (B, 61, 10, 128, 128) across all splits for S2.

For S2, the global maximum timestamp is 61 based on the previous survey. It is also the second most frequent case. This makes it highly likely that each batch has at least one sample with 61 timestamps, so almost all batches already align to shape (B, 61, 10, 128, 128). Thus, the padding size is set to 61 timestamps for all batches to ensure consistent input shapes across all splits.

torchgeo/datamodules/pastis.py
```diff
@@ -0,0 +1,113 @@
...
+import kornia.augmentation as K
+import torch
+from torch import Tensor
+from torch.utils.data import random_split
+
+from ..datasets import PASTIS
+from .geo import NonGeoDataModule
+
+
+def collate_fn(
```
adamjstewart: This should be in `torchgeo/datasets/utils.py` with the other collation functions so we can easily reuse it across multiple datamodules.

torchgeo/datamodules/pastis.py
```diff
@@ -0,0 +1,113 @@
...
+    batch: list[dict[str, Tensor]], max_timestamp: int = 61
+) -> dict[str, Any]:
+    """Custom timeseries collate fn to handle variable length sequences.
+
+    Args:
+        batch: list of sample dicts return by dataset
+        max_timestamp: maximum length of the time series
```
adamjstewart: I want to pad each mini-batch to the max length of that mini-batch. This will result in each mini-batch having a different length. This means PASTIS won't be compatible with U-Net, it would only be compatible with models supporting a dynamic length sequence.

adamjstewart (changes requested):

torchgeo/datamodules/pastis.py
yichiac: I will move this function to `utils.py`.

yichiac: I moved the function to `torchgeo/datamodules/utils.py`

yichiac:
The current version uses the max length of the entire dataset, since most mini-batches are already padded to this length. If we still want dynamic lengths, we need models that support them.

torchgeo/datamodules/utils.py
```diff
@@ -43,6 +43,44 @@ def collate_fn_detection(batch: list[dict[str, Tensor]]) -> dict[str, Any]:
     return output
 
 
+def collate_fn_timeseries(
```
adamjstewart: * Move to torchgeo/_datasets_/utils.py (nearby stack_samples, etc.)
* Rename to temporal_collate, temporal_pad (check what other PyTorch padding functions are called)
* Add a `pad_length: int | None = None` parameter (if None, pad to max length of mini-batch, else pad to `pad_length`) pad_length or max_length is fine
* Have to truncate, either clearly document this or issue a warning

tests/trainers/test_segmentation.py
```diff
@@ -117,6 +118,9 @@ def test_trainer(
             '1',
         ]
 
+        if name == 'pastis':
+            args.extend(['--data.max_timestamp', '9'])
```
adamjstewart: Put this in `pastis.yaml`?

yichiac:
- [x] Move to torchgeo/datasets/utils.py (nearby stack_samples, etc.)
- [x] Rename to temporal_collate, temporal_pad (check what other PyTorch padding functions are called) -> `pad_sequence` 
- [x] Add a pad_length: int | None = None parameter (if None, pad to max length of mini-batch, else pad to pad_length) pad_length or max_length is fine
- [x] Have to truncate, either clearly document this or issue a warning

adamjstewart:
`pad_sequence` is missing a lot of coverage. Just to be sure, did you try the default `pad_sequence` from PyTorch to make sure it doesn't work for us?

yichiac:
I haven't experimented with the default `pad_sequence`. Will try it on our test data.

yichiac:
The efficiency is improved after using the default `pad_sequence` and `pad` from PyTorch. The time required to loop through all batches is reduced from `63.26` seconds to `42.46` seconds.

tests/conf/pastis.yaml
```diff
@@ -0,0 +1,18 @@
+model:
+  class_path: SemanticSegmentationTask
+  init_args:
+    model: 'unet'
+    backbone: 'resnet18'
+    task: 'multiclass'
+    in_channels: 610 # C * T (61 timesteps * 10 S2 channels)
```
adamjstewart: Shorten to max length of test data, not real data

torchgeo/datasets/utils.py
```diff
@@ -421,6 +424,54 @@ def _dict_list_to_list_dict(
     return uncollated
 
 
+def pad_across_batches(
+    batch: list[dict[str, Tensor]],
+    padding_value: float = 0.0,
+    padding_length: int | None = None,
```
adamjstewart: Thought about this more and it seems like `None` is just a wrapper around `pad_sequence`. Let's remove support for `None` and people can instead use `pad_sequence` directly if that's the behavior they want.

torchgeo/datasets/utils.py
```diff
@@ -421,6 +424,54 @@ def _dict_list_to_list_dict(
...
+) -> dict[str, Any]:
+    """Custom timeseries collate fn to handle variable length sequences.
```
adamjstewart:
```suggestion
    """Custom time-series collate fn to handle variable length sequences.
```

torchgeo/trainers/segmentation.py
```diff
@@ -296,12 +314,26 @@ def validation_step(
...
         ):
             datamodule = self.trainer.datamodule
-            aug = K.AugmentationSequential(
-                K.Denormalize(datamodule.mean, datamodule.std),
-                data_keys=None,
-                keepdim=True,
-            )
-            batch = aug(batch)
+            if batch['image'].ndim == 5:
+                B, T, C, H, W = batch['image'].shape
```
```suggestion
                _, T, C, _, _ = batch['image'].shape
```

tests/datasets/test_utils.py
```diff
@@ -537,3 +538,38 @@ def test_azcopy(tmp_path: Path, azcopy: Executable) -> None:
...
+
+
+def test_pad_across_batches() -> None:
+    batch = [
+        {'image': torch.ones(2, 10, 5, 5), 'mask': torch.zeros(5, 5)},
+        {'image': torch.ones(3, 10, 5, 5), 'mask': torch.zeros(5, 5)},
+    ]
+
+    with pytest.raises(ValueError, match='padding_length must be specified'):
+        pad_across_batches(batch, padding_value=0.0, padding_length=cast(int, None))
```
adamjstewart: None is not an int, you can remove this test

torchgeo/datasets/utils.py
```diff
@@ -421,6 +422,61 @@ def _dict_list_to_list_dict(
...
+    Returns:
+        batch dict output
+
+    .. versionadded:: 0.8
+    """
+    if padding_length is None:
+        raise ValueError(
+            'padding_length must be specified.'
+            'If pad to the max length in the batch, use torch.nn.utils.rnn.pad_sequence directly.'
+        )
```
adamjstewart: Let's remove this and rely on the type hints to guide people

torchgeo/datasets/utils.py
```diff
@@ -421,6 +422,61 @@ def _dict_list_to_list_dict(
...
 
 
+def pad_across_batches(
+    batch: list[dict[str, Tensor]], padding_length: int, padding_value: float = 0.0
+) -> dict[str, Any]:
+    """Custom time-series collate fn to handle variable length sequences.
+
+    Args:
+        batch: list of sample dicts returned by dataset
+        padding_length: the length to pad the sequences
```
adamjstewart:
```suggestion
        padding_length: the length to pad the sequences to
```
Otherwise it sounds like the amount we add, which varies

adamjstewart (changes requested):
Last remaining changes, then we merge and move on to the next dataset!

PR #2924, opened 2025-08-07
HySpecNet11k: fix bug in filepaths

wangyi111:


adamjstewart (approved):
Good catch! Unfortunately difficult to test as the path lengths are too long for Windows.

Issue #2920, opened 2025-08-06
Change Detection transforms


blaz-r:
### Description

I'm working on implementing BTC change detection model and I've faced some issues with the transforms.
- First issue is that OSCD transforms doe not work due to image size missmatch. Different images in the dataset have different image sizes so the batch can't be stacked causing `RuntimeError: stack expects each tensor to be equal size` in collate function. I think this could be solved by cropping the images using `_ExtractPatches` in `get_item`. 

- LevirCD also has issues with augmentations, since `_ExtractPatches` causes some shape issues when transforming **mask**: `AssertionError: Dimension 0 of input is expected to be 1, got 128`. I haven't really changed any other code, and believe this might have been missed since [test only cover case](https://github.com/microsoft/torchgeo/blob/95dba5c43a2828c04a6b0a316d1996c5d1c6b9c5/tests/transforms/test_transforms.py#L21-L23) where this is inside `AugmentationSequential`, not inside additional `VideoSequential` like in [case of levircd](https://github.com/microsoft/torchgeo/blob/95dba5c43a2828c04a6b0a316d1996c5d1c6b9c5/torchgeo/datamodules/levircd.py#L56-L58). 

2nd issue can be **partialy** fixed by adding a temporal dimension to masks (going from [B, 1, H, W] to [B, 1, 1, H, W]) using `unsqueeze` [here](https://github.com/microsoft/torchgeo/blob/95dba5c43a2828c04a6b0a316d1996c5d1c6b9c5/torchgeo/datasets/levircd.py#L83), but this would then potentially require squeezing in train code.

I also think that maybe both OSCD and Levir should use `_ExtractPatches` in train transforms, but right now they use random crop. It seems that original BIT implementation uses random crops for Levir, but in the paper they say they crop 1024x1024 into 256x256 patches. I've also seen this more often. 
OSCD protocol should probably also use crops of 96x96, not random crops on full image. This would maybe also resolve the 1st issue. 

Since these issues are up for debate, and I could be doing something wrong, I'm asking for best course of action here first, before fixing things and openin a PR.

### Steps to reproduce

Reproduce OSCD:

```python
oscd_root =  <path to oscd data>
oscd = OSCDDataModule(val_split_pct=0.1, batch_size=16, root=oscd_root, bands=OSCD.rgb_bands)
oscd.setup("fit")
for batch in oscd.train_dataloader():
    pass
```
[...]

Issue #2918, opened 2025-08-04
Add Be The Change (BTC) model.


blaz-r:
### Summary

I'm the author of Be The Change (BTC) model designed for bi-temporal remote sensing change detection.
@isaaccorley suggested on Linkedin that the model could be implemented in TorchGeo. 
If you find this suitable. I now have some time to implement it, but would have some questions.

### Rationale

The main point of BTC is that it's a relatively simple method that achives great results as a result of good selection of core components (like backbone pretraining, loss, correct augmentations...). I believe it can also serve as a solid starting point for future methods.

### Implementation

To add BTC to torchgeo, I belive that majority of components are already supported, but I would need to add:
 - [ ] port hugginface swin weights to torchvision (already have that working) and integrate loading into TorchGeo
 - [ ] add dice loss
 - [ ] modify / somehow change the default augmentations of CD datasets. 

What would be the best way to approach this? It could all be part of a single PR, or each its own with a final BTC PR that merges them all.

### Additional information

Paper arxiv: https://arxiv.org/abs/2507.03367
Original repo: https://github.com/blaz-r/BTC-change-detection

adamjstewart:
Yes, we would love to have your model in TorchGeo! Some comments:

* The majority of models in TorchGeo are based on timm, not torchvision. If your model is already compatible with timm's Swin ViT implementation, it should be very easy to add the model to https://github.com/microsoft/torchgeo/blob/main/torchgeo/models/swin.py.
* Is Dice part of the model architecture, or just used to train the model? We don't necessarily need to add it if it isn't part of the model architecture, and it may make more sense to contribute directly to PyTorch.
* Same here, just adding the pretrained model weights does not require all datasets to be made compatible. What kind of augmentations are you proposing? They need to remain compatible with all of our other CD models too.

isaaccorley:
- we only chose torchvision swin because the Satlas models use this implementation. Preferably we could use timm however if you already have them working with the torchvision swin then that is also fine
- smp actually has https://smp.readthedocs.io/en/latest/losses.html#diceloss implementation so likely don't need to add this unless yours is customized
- regarding the augmentations/transforms, we do allow for these to be added along with the models but detached from the datasets/datamodules so the user can choose to use them or not. I think it makes sense to add them into the same script containing the model. Note that we use kornia so preferably it would be compatible here so that we can leverage kornia's ability to augment both image/mask/bbox appropriately at the batch level.

blaz-r:
[...]

PR #2926, opened 2025-08-08
Add Be The Change (BTC) change detection model

blaz-r:
This PR adds the BTC model as discussed in #2918 . 
The BTC class was verified on OSCD dataset inside my own framework (which handles dataloading and loss). It closely matches the performance of the original implementation.
The final modifications (including dice loss to trainer) and full verification within the torchgeo framework will be performed later, once the data loading and transforms are fixed (related to #2920 and #2910 ). 

I'm opening this upfront, and would kindly ask if you could provide general feedback. I'm especially not sure about how the swin backbone is handled currently, but I think this is the easiest way of also loading pretrained feature layernorm weights.

### DIFF
--- a/docs/api/models/swin-transformer.rst
+++ b/docs/api/models/swin-transformer.rst
@@ -7,9 +7,9 @@ Swin Transformer
 .. autofunction:: swin_b
 .. autofunction:: swin_v2_t
 .. autofunction:: swin_v2_b
+.. autoclass:: SwinBackbone_Weights
+.. autoclass:: Swin_T_Weights
+.. autoclass:: Swin_S_Weights
+.. autoclass:: Swin_B_Weights
 .. autoclass:: Swin_V2_T_Weights
 .. autoclass:: Swin_V2_B_Weights
-.. autoclass:: Swin_B_Weights
-.. autoclass:: Swin_S_Weights
-.. autoclass:: Swin_T_Weights
-.. autoclass:: SwinBackbone_Weights


blaz-r:

torchgeo/models/btc.py
```diff
@@ -0,0 +1,233 @@
...
+        x = F.interpolate(x, (h, w), mode='bilinear', align_corners=False)
+        return x
+
+
+class SwinBackbone_Weights(WeightsEnum):  # type: ignore[misc]
+    """SwinBackbone weights."""
+
+    CITYSCAPES_SEMSEG_TINY = Weights(
+        url='https://huggingface.co/blaz-r/swin_tiny_cityscapes_semantic_torchvision/resolve/main/swin_tiny_cityscapes_semantic.pth',
+        transforms=lambda x: x,
```
isaaccorley: Change this to nn.Identity() here and below

torchgeo/models/btc.py
```diff
@@ -0,0 +1,233 @@
...
+        # scale to match input image
+        x = F.interpolate(x, (h, w), mode='bilinear', align_corners=False)
+        return x
+
+
+class SwinBackbone_Weights(WeightsEnum):  # type: ignore[misc]
+    """SwinBackbone weights."""
+
+    CITYSCAPES_SEMSEG_TINY = Weights(
+        url='https://huggingface.co/blaz-r/swin_tiny_cityscapes_semantic_torchvision/resolve/main/swin_tiny_cityscapes_semantic.pth',
```
isaaccorley: Can you change this from main to the commit hash so we can version control it. Here and below

torchgeo/models/btc.py
```diff
@@ -0,0 +1,233 @@
...
+            'license': 'mit',
+        },
+    )
+    CITYSCAPES_SEMSEG_SMALL = Weights(
+        url='https://huggingface.co/blaz-r/swin_small_cityscapes_semantic_torchvision/resolve/main/swin_small_cityscapes_semantic.pth',
+        transforms=lambda x: x,
+        meta={
+            'dataset': 'Cityscapes - semantic segmentation',
+            'in_chans': 3,
+            'model': 'SwinTransformer Tiny',
```
isaaccorley: Small not Tiny

torchgeo/models/btc.py
```diff
@@ -0,0 +1,233 @@
...
+            'license': 'mit',
+        },
+    )
+    CITYSCAPES_SEMSEG_BASE = Weights(
+        url='https://huggingface.co/blaz-r/swin_base_cityscapes_semantic_torchvision/resolve/main/swin_base_cityscapes_semantic.pth',
+        transforms=lambda x: x,
+        meta={
+            'dataset': 'Cityscapes - semantic segmentation',
+            'in_chans': 3,
+            'model': 'SwinTransformer Tiny',
```
isaaccorley: Base not Timy

torchgeo/models/btc.py
```diff
@@ -0,0 +1,233 @@
...
+            msg = f'Missing keys in pretrained weights: {missing}'
+            raise RuntimeError(msg)
+
+        # we select layers before reduction!
+        return_layers = ['features.1', 'features.3', 'features.5', 'features.7']
+        self.feature_extractor = create_feature_extractor(
+            model, return_nodes=return_layers
+        )
+        self.channels = self._get_feature_channels()
+        self.image_normalization = K.Normalize(
```
isaaccorley: Could we move this to the transforms in the weights instead so it's not hardcoded into the model and change this to nn.Identity() or K.Normalize(mean=[0], std=[1])

torchgeo/models/btc.py
```diff
@@ -0,0 +1,233 @@
...
+        for feat, norm in zip(features.values(), self.norms):
+            n, h, w, c = feat.shape
+            x = norm(feat)
+            x = rearrange(x, 'n h w c -> n c h w', n=n, h=h, w=w)
+            output.append(x)
+        return output
+
+    def _get_feature_channels(self) -> list[int]:
+        # dryrun
+        self.feature_extractor.eval()
```
isaaccorley, line 166: This sets the feature extractor to eval but doesn't set it back afterwards? Probably should be dynamic to check what mode it was in previously and set back to that

isaaccorley (changes requested):
Lgtm just a few minor comments

isaaccorley:
BTC needs to get added to the docs and registry in a few places. If you take a look at this PR https://github.com/torchgeo/torchgeo/pull/2719 you can see where to add them. Feel free to reach out here or in slack for tips on getting the tests to pass or if you want me to take over finishing up the tests.

blaz-r:
Great, thanks for the review. I'll check the copilot comments and the things you pointed out. I'll also do the docs and stuff + I do need to get this working with the latest changes to the base code. I'll report back when this is sorted.

blaz-r:
I've addressed the issues (I will recheck again tomorrow) and all the tests now pass. I will still have to verify the performance once the fixes in some related PRs are merged.

I have one more question regarding the registry thing: I did add the model in init.py, but all other places in the unet example do not seem applicable. I guess adding weight is an option, but while they are general they are Cityscapes weights, so my question is does it nake sense to add that to the weights.csv. If yes, which one?

torchgeo/models/btc.py
blaz-r: This now check for training state and reverts to original state after dryrun.

torchgeo/models/btc.py
blaz-r: I've moved that there now.

torchgeo/models/btc.py
blaz-r: All weights are now using commit hashs

torchgeo/models/btc.py
blaz-r: As discussed in other comment this is now the transform that was used before inside class (since this is the transform linked to the waights)

torchgeo/models/btc.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+class SwinBackbone(Module):
+    """Swin backbone for multi-resolution feature extraction."""
+
+    def __init__(self, model_size: str = 'swin_base') -> None:
+        """Initialise swin backbone for multi-resolution feature extraction.
+
+        Args:
+            model_size: "tiny" or "base" swin size.
```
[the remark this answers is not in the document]
blaz-r: Updated as suggested.

torchgeo/models/btc.py
```diff
@@ -0,0 +1,233 @@
...
+        """
+        h, w = x.shape[-2:]
+        # change trainer stacks in channel, we want stacked in batch dim for backbone
+        x = rearrange(x, 'b (t c) h w -> (b t) c h w', c=3)
+        # extract multi-resolution features
+        features = self.backbone(x)
+        # feature difference by subtraction
+        fused = self.difference(features)
+        # UperNet impl. skips first 2 feats, we don't want that so we pad with 0
+        fused = [torch.tensor(0), torch.tensor(0), *fused]
```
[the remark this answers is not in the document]
blaz-r: Updated by movin instantiation to init and then using it later here

tests/models/test_btc.py
```diff
@@ -0,0 +1,58 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
adamjstewart:
```suggestion
# Copyright (c) TorchGeo Contributors. All rights reserved.
```
Same in all other files.

tests/models/test_btc.py
```diff
@@ -0,0 +1,58 @@
...
+# Licensed under the MIT License.
+
+from unittest.mock import MagicMock, patch
```
adamjstewart: We don't use unittest, use `pytest.MonkeyPatch` instead. These tests should look almost identical to existing model tests.

torchgeo/models/btc.py
```diff
@@ -0,0 +1,240 @@
...
+class BTC(Module):
+    """Be The Change (BTC) change detection model.
+
+    If you use this model in your research, please cite the following paper:
+
+    * https://doi.org/10.1109/TGRS.2025.3585342
+    * https://arxiv.org/abs/2507.03367
+    """
+
+    def __init__(self, backbone: str, classes: int = 1) -> None:
```
adamjstewart, line 37: A more specific type hint would be:

```suggestion
    def __init__(self, backbone: Literal['swin_tiny', 'swin_small', 'swin_base'], classes: int = 1) -> None:
```
if those are the only accepted values.

torchgeo/models/btc.py
```diff
@@ -0,0 +1,240 @@
...
+        # padding tensors used for compatibility with UPerNet implementation
+        self.upernet_padding = [torch.tensor(0), torch.tensor(0)]
+
+    def forward(self, x: Tensor) -> Tensor:
+        """BTC forward call.
+
+        Extract multi-resolution features, fuse by subtraction, decode with UperNet.
+
+        Args:
+            x: input image tensor (b, t*c, h, w)
```
adamjstewart, line 75: Should also document the return value

adamjstewart:
> I have one more question regarding the registry thing: I did add the model in init.py, but all other places in the unet example do not seem applicable. I guess adding weight is an option, but while they are general they are Cityscapes weights, so my question is does it nake sense to add that to the weights.csv. If yes, which one?

Do you only have models pre-trained on Cityscapes? Any models fine-tuned on satellite imagery? I would like to add all model weights, but don't know which to add it to if it isn't satellite imagery. We could make a generic RGB or Miscellaneous or something.

blaz-r:
The weights here are only CityScapes ported from the HuggingFace format to the torchvision Swin. This code doesn't include anything pretrained on satellite data so yeah I'm not sure. I will also address the comments tomorrow.

blaz-r:
I've addressed all the comments except the backbone str type hint, since in that case, the literal clashes with the change trainer backbone type of "str".

tests/models/test_btc.py
blaz-r: This is now fixed.

tests/models/test_btc.py
blaz-r: Updated.

torchgeo/models/btc.py
blaz-r: This clashes with type "str" in change trainer for backbone argument, so I left it as str.

torchgeo/models/btc.py
blaz-r: I added this, and also for the other function that didn't have it.

torchgeo/models/btc.py
adamjstewart: I'll fix that later I guess.

tests/models/test_btc.py
```diff
@@ -0,0 +1,61 @@
...
+        class FakeSwinWeights:
+            CITYSCAPES_SEMSEG_TINY = FakeWeights()
+
+        monkeypatch.setattr('torchgeo.models.btc.SwinBackbone_Weights', FakeSwinWeights)
+
+        with pytest.raises(
+            RuntimeError,
+            match=r'Failed to load pretrained weights for backbone: unexpected keys: ',
+        ):
+            SwinBackbone(model_size='swin_tiny')
```
adamjstewart: This test isn't very useful. We should test both real and random weights. See all of the other model tests for examples of this.

tests/trainers/test_change.py
```diff
@@ -232,7 +232,11 @@ def test_fcn_model(self) -> None:
         """FCN has no backbone/decoder. Need separate test for full test coverage."""
         ChangeDetectionTask(model='fcn')
 
-    @pytest.mark.parametrize('loss_fn', ['bce', 'jaccard', 'focal'])
+    def test_btc_model(self) -> None:
+        """BTC uses only swin backbones. Need separate test for full test coverage."""
+        ChangeDetectionTask(model='btc', backbone='swin_tiny')
+
+    @pytest.mark.parametrize('loss_fn', ['bce', 'jaccard', 'focal', 'dice'])
```
adamjstewart, line 245: Dice addition looks unrelated?

torchgeo/models/btc.py
```diff
@@ -0,0 +1,247 @@
...
+from torchvision.models.swin_transformer import SwinTransformer, swin_s, swin_t
+
+
+class BTC(Module):
+    """Be The Change (BTC) change detection model.
+
+    If you use this model in your research, please cite the following paper:
+
+    * https://doi.org/10.1109/TGRS.2025.3585342
+    * https://arxiv.org/abs/2507.03367
```
adamjstewart: We normally only add one link per paper

torchgeo/models/btc.py
```diff
@@ -0,0 +1,247 @@
...
+import segmentation_models_pytorch as smp
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from einops import rearrange
+from torch import Tensor
+from torch.nn.modules import Module
+from torchvision.models import Weights, WeightsEnum
+from torchvision.models.feature_extraction import create_feature_extractor
+from torchvision.models.swin_transformer import SwinTransformer, swin_s, swin_t
```
adamjstewart: Is the timm Swin implementation compatible by any chance? We don't usually use the torchvision models, but we can.

torchgeo/models/btc.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        super().__init__()
+        self.encoder = SwinBackbone(backbone)
+        self.difference = subtraction_fusion
+        self.decoder = smp.decoders.upernet.decoder.UPerNetDecoder(
+            encoder_channels=[
+                0,
+                0,
+                *self.encoder.channels,
+            ],  # pad at the beginning since impl. cuts first two off
```
adamjstewart: Can we move the comment to a separate line so as not to stretch this line over multiple?

torchgeo/models/btc.py
```diff
@@ -0,0 +1,247 @@
...
+
+    Returns:
+        fused feature tensors list[(b c h w)].
+    """
+    out_features = []
+    for feat in x:
+        f1, f2 = rearrange(feat, '(b t) c h w -> t b c h w', t=2)
+        out_features.append(f1 - f2)
+
+    return out_features
```
adamjstewart, line 189: I don't see any helper functions for loading pre-trained weights?

torchgeo/trainers/change.py
```diff
@@ -128,6 +129,10 @@ def configure_losses(self) -> None:
                     ignore_index=ignore_index,
                     normalized=True,
                 )
+            case 'dice':
```
adamjstewart, line 132: dice changes seem unrelated, can be done in a different PR

torchgeo/models/btc.py
```diff
@@ -0,0 +1,247 @@
...
+        fused = self.upernet_padding + fused
+        # decode to change map
+        x = self.decoder(fused)
+        x = self.final_layer(x)
+        # scale to match input image
+        x = F.interpolate(x, (h, w), mode='bilinear', align_corners=False)
+        return x
+
+
+class SwinBackbone_Weights(WeightsEnum):  # type: ignore[misc]
```
adamjstewart: Could these be added to `torchgeo/models/swin.py` since they are only for the backbone?

tests/models/test_btc.py
blaz-r: The main reason for these tests is to get 100% coverage; to do this, we need to create some corrupted weights. The normal behavior is already tested by the trainer. I will, however, clarify this and add the "real" weights tests.

torchgeo/trainers/change.py
blaz-r: Okay I'll separate this into another PR

torchgeo/models/btc.py
blaz-r: I think that would be possible. I'll see about this and the above-mentioned Timm stuff. It will require some adjustments since swin.py is currently designed for torchvision's Swin V2, but I'll address that once I have the time.

tests/models/test_btc.py
adamjstewart: Empty weights are not as useful as real/random weights. All of the other model tests use real/random weights.

blaz-r:
The dice stuff is now removed as part of a separate PR. I'll address other comments after the CVPR deadline on Friday.

adamjstewart:
I was hoping to get this merged today or tomorrow so we can make a release...

torchgeo/models/btc.py
blaz-r: I've checked and the timm has different names. I also opted for the torchvision since the swin used in `torchgeo/models/swin.py` is torchvision implementation.

adamjstewart:
FYI, we're probably going to push the 0.8.0 release to next week. Good luck with your CVPR deadline!

torchgeo/models/btc.py
blaz-r: I don't understand what exactly it refers to, given the location of the code segment selected. The fusion has no weights since it's just elementwise subtraction.

torchgeo/models/btc.py
blaz-r: I've moved this now

torchgeo/models/btc.py
blaz-r: Done

torchgeo/models/btc.py
blaz-r: Changed to arxiv only

tests/models/test_btc.py
blaz-r: Since our model does not take weights enum directly I'd need to patch this somehow, but so far failed to do so. I'll try some more and see what I can do.

torchgeo/models/swin.py
```diff
@@ -240,6 +240,73 @@ class Swin_V2_B_Weights(WeightsEnum):  # type: ignore[misc]
...
+
+        Returns:
+            dict with state dict only if include_norms is False,
+            dict with 'state_dict' and 'feat_norms_state_dict' if include_norms is True.
+        """
+        full_state_dict = WeightsEnum.get_state_dict(self, *args, **kwargs)
+        if include_norms:
+            return full_state_dict
+        else:
+            return full_state_dict['state_dict']
```
isaaccorley:
```suggestion
```

Is this function used anywhere? I couldn't find it if so. If not we should remove otherwise we will need to add tests.

isaaccorley (changes requested):

torchgeo/models/swin.py
blaz-r: It is used by BTC that makes use of these saved norm layers. It should be covered by BTc test but I can add separate test as well?

isaaccorley: Oh I see it now nvm!

tests/models/test_btc.py
blaz-r: The tests now use real weights

blaz-r:
I'm not sure whetehr I should introduce separate helper functions for swin like there are currently for swin v2. These would then accept the weights enum. Do we even want that? The other option is to write a test to cover usage of SwinBackbone_Weights as regular swin weights (ie test that manually loads weights from this enum)?.

calebrob6:
Another option is to:
```
_ = SwinBackbone_Weights.CITYSCAPES_SEMSEG_TINY.get_state_dict(False)
```
somewhere and move on with life :)? 

I'd likely say, "write a test to cover usage of SwinBackbone_Weights as regular swin weights (ie test that manually loads weights from this enum)" though but will let @adamjstewart make a call here.

Anyway, thanks a ton for pushing on this @blaz-r! Just took a look at https://arxiv.org/pdf/2507.03367 and love it!

blaz-r:
Thanks for the comment :) The false is actually the default (so the weights work as regular swin weights by default). I did add the test that covers this exactly as you said. I didn't push it yet but if that is suitable I'll do that.

adamjstewart:
I would vote to add a helper function to match all of the other models. I would also like to know how important the layernorm flag is, as none of our other models have it. Ideally, the weights can be loaded without any further modification to the state dict. We can redistribute a paired down version of the weights if needed.

torchgeo/models/swin.py
```diff
@@ -247,6 +247,254 @@ class Swin_V2_B_Weights(WeightsEnum):  # type: ignore[misc]
     )
 
 
+class SwinBackbone_Weights(WeightsEnum):  # type: ignore[misc]
+    """SwinBackbone weights parent class.
+
+    These weights contain the encoder weights and optionally the backbone layernorm weights.
+    To select whether layernorm weights are returned pass `include_norms` to get_state_dict (default is false).
```
adamjstewart: versionadded

torchgeo/models/swin.py
```diff
@@ -247,6 +247,254 @@ class Swin_V2_B_Weights(WeightsEnum):  # type: ignore[misc]
...
+    """
+
+    def get_state_dict(
+        self, include_norms: bool = False, *args: Any, **kwargs: Any
+    ) -> Any:
+        """Get the state dict for this model from provided url, optionally including backbone layernorm weights.
+
+        Args:
+            include_norms (bool): Whether to also return backbone layernorm weights.
```
adamjstewart, line 266: Haven't seen this in other models. Can you explain why this is needed?

adamjstewart (changes requested):
Other things we are forgetting:

* Update `hubconf.py`
* Update `torchgeo/models/api.py`
* Update `tests/models/test_api.py`
* Update `docs/api/weights/aerial.csv` (or a different one, not sure what to call these weights)

@isaaccorley if you ever find time, we should make a checklist/tutorial for adding new models like we have for new datasets. Models are more complicated than anything else.

torchgeo/models/swin.py
blaz-r: When models are used in HuggingFace backbone they have these appended. I did also some testing and it does help with the performance a bit.

docs/api/weights/aerial.csv
```diff
@@ -1,3 +1,6 @@
 Weight,Channels,Source,Citation,License
+Swin_B_Weights.CITYSCAPES_SEMSEG,3,`link <https://github.com/facebookresearch/Mask2Former>`__,`link <https://arxiv.org/abs/2112.01527>`__,"CC-BY-NC-4.0"
```
adamjstewart, line 2: License can be found at https://github.com/facebookresearch/Mask2Former/blob/main/MODEL_ZOO.md#license

@blaz-r FYI, the license at https://hf.co/blaz-r/swin_tiny_cityscapes_semantic_torchvision is wrong.

adamjstewart (approved):

docs/api/weights/aerial.csv
blaz-r: Uh, thanks for the notice, I missed that and I'll update asap.

blaz-r:
What exactly is this referring to? The get state dict comment was marked as resolved since the function is used by BTC.

adamjstewart:
@isaaccorley had a change request, but I can't merge unless all reviews are "approved", so I dismissed his change request as stale.

adamjstewart:
Tests are currently very flaky due to internet access requirements. How hard would it be to refactor the model/weight code to not need to download anything during testing? This is how all (most) of our other models work. I also don't want to download anything by default unless users explicitly require it (security, principle of least surprise). Doesn't necessarily have to be today, but soon would be good to make CI happy.

See #1088 for some more background. Any other models you can hack to reduce the number of downloads during testing would be extremely appreciated.

blaz-r:
Okay, I'll have a look tomorrow.

PR #2929, opened 2025-08-10
Presto

isaaccorley:
This PR adds the Presto pixelwise geospatial foundation model from ["Lightweight, Pre-trained Transformers for Remote Sensing Timeseries"](https://arxiv.org/abs/2304.14065) -- MIT licensed

Mostly modified from https://github.com/nasaharvest/presto/blob/main/single_file_presto.py

cc: @gabrieltseng @ivanzvonkov @rubencart @drolnick @hannah-rae


torchgeo/models/presto.py
```diff
@@ -0,0 +1,802 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# https://github.com/nasaharvest/presto
```
adamjstewart: Need to preserve the copyright:
```python
# Copyright (c) 2024 Presto Authors
```

torchgeo/models/api.py
adamjstewart, line 1: I think you also need to update `tests/models/test_api.py` to resolve the test failure.

isaaccorley:
Oof lots of test coverage needed :(

torchgeo/models/presto.py
```diff
@@ -0,0 +1,809 @@
+# Copyright (c) 2024 Presto Authors
```
adamjstewart, line 3: Keep the TorchGeo copyright, then add the Presto copyright below that.

adamjstewart:
Also missing docs

tests/models/test_presto.py
```diff
@@ -0,0 +1,112 @@
...
+from torchgeo.models import Presto, Presto_Weights, presto
+from torchgeo.models.presto import (
+    BANDS_GROUPS_IDX,
+    NUM_DYNAMIC_WORLD_CLASSES,
+    get_sinusoid_encoding_table,
+    month_to_tensor,
+)
+
+
+def _make_inputs(
```
adamjstewart: Should this be a fixture?

torchgeo/models/presto.py
```diff
@@ -0,0 +1,814 @@
...
+    'S2_RGB': [2, 3, 4],
+    'S2_Red_Edge': [5, 6, 7],
+    'S2_NIR_10m': [8],
+    'S2_NIR_20m': [9],
+    'S2_SWIR': [10, 11],
+    'ERA5': [12, 13],
+    'SRTM': [14, 15],
+    'NDVI': [16],
+}
+NUM_DYNAMIC_WORLD_CLASSES = 9
```
adamjstewart: Both of these are hard-coded to the data that Presto happened to be pre-trained on. However, Presto itself should be compatible with any sensors if you train from scratch. Should we un-hardcode this and make it a parameter?

Same question applies to #3305, so let's finalize an answer and be consistent across both models.

torchgeo/models/presto.py
```diff
@@ -0,0 +1,814 @@
...
+            functions.
+
+    Returns:
+        A tensor of shape ``(len(positions), d_hid)`` containing the positional
+        encodings.
+    """
+    if isinstance(positions, int):
+        positions = list(range(positions))
+
+    def cal_angle(position: int, hid_idx: int) -> float:
```
adamjstewart, line 54: Ideally we would also have docstrings for these, but I know this is a crazy ask, so let's call it optional

torchgeo/models/presto.py
```diff
@@ -0,0 +1,814 @@
...
+    month_table_tensor: torch.Tensor = (
+        torch.from_numpy(month_table).to(torch.float).to(device)
+    )
+    return month_table_tensor
+
+
+def month_to_tensor(
+    month: torch.Tensor | int, batch_size: int, seq_len: int, device: torch.device
+) -> torch.Tensor:
+    """Convert month to a tensor.
```
adamjstewart: From the description alone, it is not really clear how this differs from `torch.tensor(month)`...

torchgeo/models/presto.py
```diff
@@ -0,0 +1,814 @@
...
+
+        # apply Transformer blocks
+        for blk in self.blocks:
+            x = blk(x)
+
+        # mask will be a boolean of shape [batch, total_num_tokens]
+        return self.norm(x), kept_indices, removed_indices
+
+
+class Decoder(nn.Module):
```
adamjstewart, line 420: Very happy that we are finally starting to add decoders!

torchgeo/models/presto.py
```diff
@@ -0,0 +1,814 @@
...
+        x = self.decoder_norm(x)
+
+        reconstructed_inputs, dw_output = self.reconstruct_inputs(x)
+        return reconstructed_inputs, dw_output
+
+
+class Presto(nn.Module):
+    """Pretrained Remote Sensing Transformer (Presto).
+
+    .. versionadded:: 0.8
```
adamjstewart:
```suggestion
    .. versionadded:: 0.9
```

torchgeo/models/presto.py
```diff
@@ -0,0 +1,814 @@
...
+        reconstructed_inputs, dw_output = self.decoder(
+            x, kept_indices, removed_indices, month
+        )
+        return reconstructed_inputs, dw_output
+
+
+class Presto_Weights(WeightsEnum):  # type: ignore[misc]
+    """Presto weights.
+
+    .. versionadded:: 0.8
```
adamjstewart:
```suggestion
    .. versionadded:: 0.9
```

torchgeo/models/presto.py
```diff
@@ -0,0 +1,814 @@
...
+
+
+def presto(weights: Presto_Weights | None = None, *args: Any, **kwargs: Any) -> Presto:
+    """Presto model.
+
+    If you use this model in your research, please cite the following paper:
+
+    * https://arxiv.org/abs/2304.14065
+
+    .. versionadded:: 0.8
```
adamjstewart:
```suggestion
    .. versionadded:: 0.9
```

torchgeo/models/presto.py
```diff
@@ -0,0 +1,814 @@
...
+    """Presto weights.
+
+    .. versionadded:: 0.8
+    """
+
+    PRESTO = Weights(
+        url='https://github.com/nasaharvest/presto/raw/5afde40850d73bfaed26078fc3bda621a55c311d/data/default_model.pt',
+        transforms=None,
+        meta={
+            'dataset': 'Presto LEM',
```
adamjstewart: I couldn't find any info about Presto LEM online, is this what the dataset is called internally?

isaaccorley: It's called LEM on Google cloud storage folder and in their [pretraining data diagram](https://docs.google.com/presentation/d/1rxUXdSmKtfHSusBFzG3Apx-pLR0UE9oVIOaEB997bAM/edit?usp=sharing) referenced on their readme.

adamjstewart: Out of curiosity, any idea what the acronym stands for?

isaaccorley: No idea maybe @gabrieltseng knows?

torchgeo/models/presto.py
isaaccorley: I've made it an optional arg that defaults to these

torchgeo/models/presto.py
```diff
@@ -0,0 +1,837 @@
...
+        ).to(device)
+    return month
+
+
+class Encoder(nn.Module):
+    """Encoder for the Presto model."""
+
+    def __init__(
+        self,
+        band_groups: dict[str, list[int]] | None = None,
```
adamjstewart: Only list, or does any Sequence work?

isaaccorley:
> Slow tests fail with:
> ```console
> > pytest --force-enable-socket -m slow tests/models/test_presto.py
> FAILED tests/models/test_presto.py::TestPresto::test_presto_download[Presto_Weights.PRESTO] - RuntimeError: Attempting to deserialize object on a CUDA device but torch.cuda.is_available() is False. If you are running on a CPU-only machine, please use torch.load with map_location=torch.device('cpu')...
> ```

~~Ahh I know the issue I need to use map_location="cpu"~~

Nvm I forgot to change the url to the huggingface one

PR #2937, opened 2025-08-12
Remove deepcopy workaround

ashnair1:
Since we bumped min supported torchvision version to `0.15.1` (in #2559), these patches can be removed


isaaccorley (approved):

adamjstewart (approved):
Wish we could get rid of the type ignores too.

PR #2938, opened 2025-08-12
Add CNN-LSTM Model

f-schi:
Add CNN-LSTM Model: https://arxiv.org/abs/1411.4389


docs/api/models.rst
```diff
@@ -16,6 +16,14 @@ Change Star
 .. autoclass:: ChangeStarFarSeg
 .. autoclass:: ChangeMixin
 
+CNNLSTM
+^^^^^^^
```
adamjstewart:
```suggestion
CNN-LSTM
^^^^^^^^
```

docs/api/models.rst
```diff
@@ -16,6 +16,14 @@ Change Star
...
+
+.. autoclass:: CNNLSTM
+.. autoclass:: CNN
+.. autoclass:: LSTM
+.. autoclass:: Task
```
adamjstewart: Only document the stuff exported in `__init__.py`

tests/models/test_cnnlstm.py
```diff
@@ -0,0 +1,856 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+from typing import Any
+from unittest.mock import patch
```
adamjstewart: Use pytest MonkeyPatch instead

tests/models/test_cnnlstm.py
```diff
@@ -0,0 +1,856 @@
...
+
+import pytest
+import torch
+import torch.nn as nn
+
+# Import from torchgeo models module
```
adamjstewart: This comment is kinda useless...

tests/models/test_cnnlstm.py
```diff
@@ -0,0 +1,856 @@
...
+        for param in model.parameters():
+            if param.requires_grad:
+                total_params += 1
+                if param.grad is not None:
+                    params_with_grad += 1
+
+        assert params_with_grad > 0, 'No parameter gradients found'
+
+
+class TestIntegrationAndEdgeCases:
```
adamjstewart: Let's merge the rest of these classes into one of the above. Running `pytest tests/models/test_cnnlstm.py::TestCNN` should result in 100% coverage for CNN. Any integration tests can go in `TestCNNLSTM` since it represents the total sum of all models.

tests/models/test_cnnlstm.py
```diff
@@ -0,0 +1,856 @@
...
+        cnn = CNN(model_name='resnet18', pretrained=False, output_dim=256)
+        assert cnn.output_dim == 256
+        # Test that the backbone was properly initialized
+        assert hasattr(cnn, 'backbone')
+        assert hasattr(cnn, 'projection')
+
+
+if __name__ == '__main__':
+    # Run with coverage: pytest --cov=torchgeo.models.cnn_lstm tests/models/test_cnn_lstm.py
+    pytest.main([__file__, '-v', '--tb=short'])
```
adamjstewart: Remove this, pytest handles all CLI stuff

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,378 @@
...
+    Raises:
+        ImportError: If TIMM library is not available.
+
+    Forward:
+        x (torch.Tensor): Input tensor of shape either
+            - (B, T, C, H, W) for sequences of images, or
+            - (B, C, H, W) for single images.
+
+    Returns:
+        torch.Tensor: Extracted features of shape (B, T, output_dim) or (B, output_dim).
```
adamjstewart: Remove all this, this goes inside the `__init__` and `forward` docstrings

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,378 @@
...
+        pretrained: bool = True,
+        output_dim: int | None = None,
+        freeze_backbone: bool = False,
+        in_chans: int = 3,
+    ) -> None:
+        """Initialize CNN backbone with TIMM model.
+
+        Args:
+            model_name: Name of the TIMM model architecture to use.
+            pretrained: Whether to load pretrained weights.
```
adamjstewart: We use `weights` for other models to match torchvision's newer weights API. It should take a Weights enum value as input

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,378 @@
...
+        self.projection: nn.Linear | nn.Identity
+        if feature_dim != output_dim:
+            self.projection = nn.Linear(feature_dim, output_dim)
+        else:
+            self.projection = nn.Identity()
+
+        # Optionally freeze backbone
+        if freeze_backbone:
+            for param in self.backbone.parameters():
+                param.requires_grad = False
```
adamjstewart: Wondering if this should be here or in the trainer. We do it in the trainer for all other models

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,378 @@
...
+        if x.dim() == 5:  # (B, T, C, H, W)
+            b, t, c, h, w = x.shape
+            x = x.view(b * t, c, h, w)
+            features = self.backbone(x)
+            features = self.projection(features)
+            features = features.view(b, t, self.output_dim)
+        else:  # (B, C, H, W)
+            features = self.backbone(x)
+            features = self.projection(features)
+        return features  # type: ignore[no-any-return]
```
adamjstewart: Instead of ignoring, keep using `x` as the output variable name instead of `features`. This will force mypy to recognize that the type is unchanged.

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,378 @@
...
+
+
+class LSTM(nn.Module):
```
adamjstewart, line 80: Would like to sync with what @keves1 has been doing in #2639. Basically, your `CNN` could become `Encoder` and you could reuse @keves1's `Decoder` without the code duplication.

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,378 @@
...
+            - (B, T, input_dim) for sequences.
+
+    Returns:
+        Dict[str, Optional[torch.Tensor]]:
+            For classification:
+                - 'outputs': Raw logits tensor (B, num_classes).
+                - 'probs': Softmax probabilities tensor (B, num_classes).
+            For regression:
+                - 'outputs': Regression predictions (B, output_dim).
+                - 'probs': None.
```
adamjstewart: classification/regression are the same thing, the only difference is the `num_classes`. Both should return raw logits, no probabilities.

Not sure if it makes sense to use this model for semantic segmentation or pixelwise regression. Maybe the `Decoder` should just have a `num_classes` parameter and no `Task` is needed?

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,378 @@
...
+        x (torch.Tensor): Input tensor, either
+            - (B, T, C, H, W) sequence of images, or
+            - (B, C, H, W) single image.
+        lengths (torch.Tensor, optional): Lengths of sequences for LSTM packing.
+
+    Returns:
+        Dict[str, Optional[torch.Tensor]]: Output dictionary from task head.
+    """
+
+    def __init__(self, cnn_backbone: CNN, rnn_encoder: LSTM, task_head: Task) -> None:
```
adamjstewart: We want this to be usable via YAML configuration file. So strings need to be supported. We _could_ support nn.Modules too, but if strings work, then nn.Modules probably aren't needed.

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,378 @@
...
+        """Initialize CNN-LSTM model.
+
+        Args:
+            cnn_backbone: CNN feature extractor module.
+            rnn_encoder: LSTM sequence encoder module.
+            task_head: Task head for classification or regression.
+
+        Raises:
+            ValueError: If the output dimension of CNN does not match input dimension of LSTM,
+                        or if the output dimension of LSTM does not match input dimension of task head.
```
adamjstewart: Indented too far?

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,378 @@
...
+        self.cnn_backbone = cnn_backbone
+        self.rnn_encoder = rnn_encoder
+        self.task_head = task_head
+
+        # Validate CNN -> RNN dimensions
+        if cnn_backbone.output_dim != rnn_encoder.input_dim:
+            raise ValueError(
+                f'CNN output dim ({cnn_backbone.output_dim}) must match '
+                f'RNN input dim ({rnn_encoder.input_dim})'
+            )
```
adamjstewart: If there is only a single CNNLSTM class exposed and the user passes strings and numbers to that class, we won't need to check for this.

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,378 @@
...
+
+        rnn_output_dim = rnn_encoder.output_dim
+        task_input_dim = task_head.head.in_features
+
+        if rnn_output_dim != task_input_dim:
+            raise ValueError(
+                f'RNN output dim ({rnn_output_dim}) must match '
+                f'Task input dim ({task_input_dim})'
+            )
```
adamjstewart: Same here

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,378 @@
...
+               (B, C, H, W) single image.
+            lengths: Lengths of sequences for LSTM packing (optional).
+            **kwargs: Additional keyword arguments.
+
+        Returns:
+            Output dictionary from task head containing predictions and probabilities.
+        """
+        # Handle single image input by adding time dimension
+        if x.dim() == 4:  # (B, C, H, W)
+            x = x.unsqueeze(1)  # (B, 1, C, H, W)
```
adamjstewart: Does single image input make sense? Maybe we should raise an error if a time series is not provided?

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,378 @@
...
+        cnn_features = self.cnn_backbone(x)  # (B, T, feature_dim)
+
+        # Sequence modeling with LSTM
+        rnn_output = self.rnn_encoder(cnn_features, lengths)  # (B, T, hidden_dim)
+
+        # Task-specific output
+
+        output = self.task_head(rnn_output)
+
+        return output  # type: ignore[no-any-return]
```
adamjstewart: Same here, use `x` as the variable name for all steps and you won't have to ignore the return type

adamjstewart (changes requested):

adamjstewart:
Where did the tests go? Also, can you resolve the merge conflicts?

docs/api/models.rst
```diff
@@ -16,6 +16,7 @@ Model Architectures
    models/changestar
    models/changevit
    models/convlstm
+   models/cnnlstm
```
adamjstewart, line 19: Should be above convlstm. Documentation page is missing.

tests/models/test_cnnlstm.py
adamjstewart, line 1: File is currently empty, needs tests.

torchgeo/models/__init__.py
adamjstewart, line 1: Run ruff to fix this file

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,251 @@
+# Licensed under the MIT License.
```
adamjstewart, line 1: Missing the rest of the copyright message

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,251 @@
...
+            lengths: Lengths of sequences for packing (optional).
+
+        Returns:
+            Output features from the LSTM of shape (B, T, hidden_dim * num_directions).
+        """
+        if lengths is not None:
+            # Pack for variable-length sequences
+            packed_x = nn.utils.rnn.pack_padded_sequence(
+                x, lengths.cpu(), batch_first=self.batch_first, enforce_sorted=False
+            )
```
adamjstewart, line 137: I'm wondering if this should be done in the model or in the datamodule...

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,251 @@
...
+        """Initialize CNN-LSTM model.
+
+        Args:
+            model_name: Name of the TIMM model architecture to use for CNN backbone.
+            weights: Weights enum value for loading pretrained weights.
+            cnn_output_dim: CNN output feature dimension. If None, uses native dimension.
+            in_chans: Number of input channels (e.g., 3 for RGB).
+            lstm_hidden_dim: Number of features in LSTM hidden state.
+            lstm_num_layers: Number of recurrent layers in LSTM.
+            lstm_bidirectional: If True, use bidirectional LSTM.
```
adamjstewart, line 172: Are bidirectional LSTMs actually useful for time series modeling? We don't generally have access to future states, and future states don't influence past states. It's quite different from language translation where bidirectional LSTMs make sense.

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,251 @@
...
+import torch.nn as nn
+
+
+class CNN(nn.Module):
+    """CNN feature extractor using TIMM backbones."""
+
+    def __init__(
+        self,
+        model_name: str,
+        weights: Any = None,
```
adamjstewart, line 18: Would like to avoid `Any` if we can.

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,251 @@
...
+
+    def forward(
+        self, x: torch.Tensor, lengths: torch.Tensor | None = None, **kwargs: Any
+    ) -> torch.Tensor:
+        """Forward pass through the complete CNN-LSTM model.
+
+        Args:
+            x: Input tensor of shape (B, T, C, H, W) - sequence of images.
+            lengths: Lengths of sequences for LSTM packing (optional).
+            **kwargs: Additional keyword arguments.
```
adamjstewart, line 217: Unused? We should drop if they aren't used, otherwise simple typos like `length` will go unnoticed.

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,251 @@
...
+        if x.dim() != 5:
+            raise ValueError(
+                f'Expected 5D input tensor (B, T, C, H, W), got {x.dim()}D tensor'
+            )
+
+        # CNN feature extraction
+        x = self.cnn_backbone(x)  # (B, T, feature_dim)
+
+        # Sequence modeling with LSTM
+        x = self.rnn_encoder(x, lengths)  # (B, T, hidden_dim)
```
adamjstewart, line 235: Not sure if this is an encoder or decoder...

torchgeo/models/cnnlstm.py
```diff
@@ -0,0 +1,251 @@
...
+        in_chans: int = 3,
+        lstm_hidden_dim: int = 256,
+        lstm_num_layers: int = 1,
+        lstm_bidirectional: bool = False,
+        lstm_dropout: float = 0.0,
+        num_classes: int = 1,
+        head_dropout: float = 0.0,
+        pooling: str = 'last',
+    ) -> None:
+        """Initialize CNN-LSTM model.
```
adamjstewart, line 163: Can you add citation information? Which paper should users cite if they want to use this model? Also needs:
```rst
.. versionadded:: 0.9
```

adamjstewart (changes requested):
Apologies for the delayed review!

robmarkcole:
@f-schi whats the status of this PR?

f-schi:
@robmarkcole its currently on hold from my side to finish up the more urgent time series PRs. Once these are merged (very soon) I would continue here. Ofc if someone wants to directly take over, feel free!

PR #2939, opened 2025-08-12
Add CDL 2024 data

calebrob6:
Updating the CDL dataset with the new 2024 data.


calebrob6:
There is a new 10m CDL product btw, will open a new PR to support it

adamjstewart (approved):
I never know whether to include this in the next patch release or minor release. There is no API changes, but there is a reproducibility change, so maybe next minor release to stay safe.

PR #2940, opened 2025-08-12
fix: Update model/dataset huggingface urls

isaaccorley:
Some non-commercial models/datasets have been moved out of the torchgeo huggingface account and into my personal account. This PR updates the urls. Note that huggingface redirects the original urls to the new user account but this is to explicitly update this.

@adamjstewart note that some datasets in the TorchGeo account were not actually being used in the library e.g. `million-aid` and `isaid` datasets.

### DIFF
--- a/torchgeo/models/yolo.py
+++ b/torchgeo/models/yolo.py
@@ -77,7 +77,7 @@ class YOLO_Weights(WeightsEnum):  # type: ignore[misc]
         },
     )
     CORE_DINO = Weights(
-        url='https://hf.co/torchgeo/core-dino/resolve/59427e13d114cbbf02f4745e1bea7570be3e2057/core_dino_rgb_yolo11x-80ca836f.pt',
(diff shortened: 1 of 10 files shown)


calebrob6 (approved):

torchgeo/models/yolo.py
```diff
@@ -77,7 +77,7 @@ class YOLO_Weights(WeightsEnum):  # type: ignore[misc]
         },
     )
     CORE_DINO = Weights(
-        url='https://hf.co/torchgeo/core-dino/resolve/59427e13d114cbbf02f4745e1bea7570be3e2057/core_dino_rgb_yolo11x-80ca836f.pt',
+        url='https://hf.co/isaaccorley/core-dino/resolve/59427e13d114cbbf02f4745e1bea7570be3e2057/core_dino_rgb_yolo11x-80ca836f.pt',
```
adamjstewart, line 80: Did this one get moved back?

isaaccorley: This one was cc-by-nc

PR #2945, opened 2025-08-17
RasterDataset: rasterio -> (rio)xarray

### LINKED ISSUE #1486: Xarray Dataset

adamjstewart:
This PR converts the `RasterDataset` backend from rasterio to (rio)xarray.

### Motivation

(rio)xarray can load anything that rasterio can, plus NetCDF/HDF5/Zarr and many other file formats with a corresponding xarray backend. We need NetCDF support for ERA5/CMIP6/WeatherBench, and ESA is planning to switch to Zarr in the near future.

Xarray is also extendable. While xarray no longer supports HDF4 files (needed by MODIS), we could [add a new backend](https://docs.xarray.dev/en/stable/internals/how-to-add-new-backend.html) to add support for HDF4.

> [!NOTE]
> NetCDF/HDF5/Zarr files do not always follow standard [metadata conventions](https://cfconventions.org/) and may require additional massaging before they can be used with rioxarray. This PR does not yet do that. I would like to do this in follow-up PRs when adding datasets that actually require it. This will make testing much easier, and the base class can grow organically only as needed.

### Benchmarking

Because @calebrob6 will definitely want to see this 🙂 

| version | raw (random) | raw (grid) | preprocessed (random) | preprocessed (grid) |
| --- | ---: | ---: | ---: | ---: |
| v0.7.1 | 7.398 | 3.411 | 6.167 | 3.018 |
| main | broken | broken | 7.891 | 4.642 |
| #2945 | broken | broken | 37.934 | 34.377 |

Slower than I would like, but it's a start.

Closes #1490 
Closes #1486 
Closes #509

### DIFF
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -441,7 +441,6 @@ class RasterDataset(GeoDataset):
                             res = (abs(_res[0]), abs(_res[1]))
                 except (OSError, ValueError):
                     # Skip files that xarray is unable to read
-                    raise
                     continue
                 else:
                     filepaths.append(filepath)
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -440,15 +440,15 @@ class RasterDataset(GeoDataset):
             if match is not None:
                 try:
                     with xr.open_dataset(filepath, decode_coords='all') as src:
-                        crs = crs or src.rio.crs or CRS.from_epsg(4326)
+                        crs = crs or src.rio.crs # or CRS.from_epsg(4326)
 
-                        if src.rio.crs is None:
-                            warnings.warn(
-                                f"Unable to decode coordinates of '{filepath}', "
-                                f'defaulting to {crs}. Set `crs` if this is incorrect.',
-                                UserWarning,
-                            )
-                            src = src.rio.write_crs(crs)
+                        # if src.rio.crs is None:
+                        #     warnings.warn(
+                        #         f"Unable to decode coordinates of '{filepath}', "
+                        #         f'defaulting to {crs}. Set `crs` if this is incorrect.',
+                        #         UserWarning,
+                        #     )
+                        #     src = src.rio.write_crs(crs)
 
                         if src.rio.crs != crs:
                             src = src.rio.reproject(crs)
@@ -613,8 +613,8 @@ class RasterDataset(GeoDataset):
         """
         src = xr.open_dataset(filepath, decode_coords='all')
 
-        if src.rio.crs is None:
-            src = src.rio.write_crs(self.crs)
+        # if src.rio.crs is None:
+        #     src = src.rio.write_crs(self.crs)
 
         if src.rio.crs != self.crs:
             src = src.rio.reproject(self.crs, resampling=self.resampling)
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -26,7 +26,6 @@ import torch
 import xarray as xr
 from geopandas import GeoDataFrame
 from pyproj import CRS
-from rasterio.enums import Resampling
 from torch import Tensor
 from torch.utils.data import Dataset
 from torchvision.datasets import ImageFolder
@@ -376,23 +375,6 @@ class RasterDataset(GeoDataset):
         else:
             return torch.long
 
-    @property
-    def resampling(self) -> Resampling:
-        """Resampling algorithm used when reading input files.
-
-        Defaults to bilinear for float dtypes and nearest for int dtypes.
-
-        Returns:
-            The resampling method to use.
-
-        .. versionadded:: 0.6
-        """
-        # Based on torch.is_floating_point
-        if self.dtype in [torch.float64, torch.float32, torch.float16, torch.bfloat16]:
-            return Resampling.bilinear
-        else:
-            return Resampling.nearest
-
     def __init__(
         self,
         paths: Path | Iterable[Path] = 'data',
@@ -585,7 +567,7 @@ class RasterDataset(GeoDataset):
         bounds = (x.start, y.start, x.stop, y.stop)
         res = (x.step, y.step)
         dataset = rioxarray.merge.merge_datasets(
-            datasets, bounds=bounds, res=res, method=self.resampling, crs=self.crs
+            datasets, bounds=bounds, res=res, crs=self.crs
         )
         # Use array_to_tensor since merge may return uint16/uint32 arrays.
         tensor = array_to_tensor(dataset.band_data.values)
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -466,7 +466,7 @@ class RasterDataset(GeoDataset):
             if self.bands:
                 if self.all_bands:
                     self.band_indexes = [
-                        self.all_bands.index(i) + 1 for i in self.bands
+                        self.all_bands.index(i) for i in self.bands
                     ]
                 else:
                     msg = (
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -441,6 +441,7 @@ class RasterDataset(GeoDataset):
                             res = (abs(_res[0]), abs(_res[1]))
                 except (OSError, ValueError):
                     # Skip files that xarray is unable to read
+                    raise
                     continue
                 else:
                     filepaths.append(filepath)
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -596,7 +596,7 @@ class RasterDataset(GeoDataset):
         # Use array_to_tensor since merge may return uint16/uint32 arrays.
         tensor = array_to_tensor(array)
 
-        return tensor.squeeze(0)
+        return tensor
 
     @functools.lru_cache(maxsize=128)
     def _cached_load_warp_file(self, filepath: Path) -> xr.Dataset:
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -26,6 +26,7 @@ import torch
 import xarray as xr
 from geopandas import GeoDataFrame
 from pyproj import CRS
+from rasterio.enums import Resampling
 from torch import Tensor
 from torch.utils.data import Dataset
 from torchvision.datasets import ImageFolder
@@ -375,6 +376,23 @@ class RasterDataset(GeoDataset):
         else:
             return torch.long
 
+    @property
+    def resampling(self) -> Resampling:
+        """Resampling algorithm used when reading input files.
+
+        Defaults to bilinear for float dtypes and nearest for int dtypes.
+
+        Returns:
+            The resampling method to use.
+
+        .. versionadded:: 0.6
+        """
+        # Based on torch.is_floating_point
+        if self.dtype in [torch.float64, torch.float32, torch.float16, torch.bfloat16]:
+            return Resampling.bilinear
+        else:
+            return Resampling.nearest
+
     def __init__(
         self,
         paths: Path | Iterable[Path] = 'data',
@@ -599,7 +617,7 @@ class RasterDataset(GeoDataset):
             src = src.rio.write_crs(self.crs)
 
         if src.rio.crs != self.crs:
-            src = src.rio.reproject(self.crs)
+            src = src.rio.reproject(self.crs, resampling=self.resampling)
 
         return src
 
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -23,6 +23,7 @@ import rasterio
 import rasterio.merge
 import shapely
 import torch
+import xarray as xr
 from geopandas import GeoDataFrame
 from pyproj import CRS
 from rasterio.enums import Resampling
@@ -440,23 +441,25 @@ class RasterDataset(GeoDataset):
             match = re.match(filename_regex, os.path.basename(filepath))
             if match is not None:
                 try:
-                    with rasterio.open(filepath) as src:
-                        # See if file has a color map
-                        if len(self.cmap) == 0:
-                            try:
-                                self.cmap = src.colormap(1)  # type: ignore[misc]
-                            except ValueError:
-                                pass
+                    with xr.open_dataset(filepath, decode_coords='all') as src:
+                        crs = crs or src.rio.crs or CRS.from_epsg(4326)
 
-                        if crs is None:
-                            crs = src.crs
+                        if src.rio.crs is None:
+                            warnings.warn(
+                                f"Unable to decode coordinates of '{filepath}', "
+                                f'defaulting to {crs}. Set `crs` if this is incorrect.',
+                                UserWarning,
+                            )
+                            src = src.rio.write_crs(crs)
 
-                        with WarpedVRT(src, crs=crs) as vrt:
-                            geometries.append(shapely.box(*vrt.bounds))
-                            if res is None:
-                                res = vrt.res
-                except rasterio.errors.RasterioIOError:
-                    # Skip files that rasterio is unable to read
+                        if src.rio.crs != crs:
+                            src = src.rio.reproject(crs)
+
+                        geometries.append(shapely.box(*src.rio.bounds()))
+                        if res is None:
+                            res = src.rio.resolution()
+                except (OSError, ValueError):
+                    # Skip files that xarray is unable to read
                     continue
                 else:
                     filepaths.append(filepath)
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -20,15 +20,13 @@ import geopandas as gpd
 import numpy as np
 import pandas as pd
 import rasterio
-import rasterio.merge
+import rioxarray.merge
 import shapely
 import torch
 import xarray as xr
 from geopandas import GeoDataFrame
 from pyproj import CRS
 from rasterio.enums import Resampling
-from rasterio.io import DatasetReader
-from rasterio.vrt import WarpedVRT
 from torch import Tensor
 from torch.utils.data import Dataset
 from torchvision.datasets import ImageFolder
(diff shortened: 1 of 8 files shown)


isaaccorley:
I believe rioxarray does support resampling it's just combined within the "reproject" method. https://corteva.github.io/rioxarray/stable/examples/resampling.html

torchgeo/datasets/geo.py
```diff
@@ -576,50 +576,50 @@ def _merge_files(
...
-            vrt_fhs = [self._load_warp_file(fp) for fp in filepaths]
+            datasets = [self._load_warp_file(fp) for fp in filepaths]
 
         x, y, t = self._disambiguate_slice(query)
         bounds = (x.start, y.start, x.stop, y.stop)
         res = (x.step, y.step)
-        dest, _ = rasterio.merge.merge(
-            vrt_fhs, bounds, res, indexes=band_indexes, resampling=self.resampling
+        dataset = rioxarray.merge.merge_datasets(
+            datasets, bounds=bounds, res=res, nodata=0, crs=self.crs
```
adamjstewart, line 588: Adding `nodata=0` made this code about 100x faster.

adamjstewart:
Figured out the bug breaking the raw I/O benchmarking script. It has to do with sampling outside the bounds of the files but still within the bounding box of the files. Fix is to always sample within the bounding boxes. Let's fix this in a separate PR.

adamjstewart:
cProfile on my system in case anyone wants to dig into performance: [torchgeo.txt](https://github.com/user-attachments/files/21919228/torchgeo.txt)

adamjstewart:
I'm able to shave off 5 sec by using `recalc=False` everywhere in https://github.com/corteva/rioxarray/blob/0.19.0/rioxarray/merge.py, but not sure if that's a good idea or not. Overall, there seems to be a lot of overhead in rioxarray that isn't easy to remove without cythonizing rioxarray. Let me call in backup.

torchgeo/datasets/geo.py
```diff
@@ -440,23 +439,26 @@ def __init__(
...
-                        if len(self.cmap) == 0:
-                            try:
-                                self.cmap = src.colormap(1)  # type: ignore[misc]
-                            except ValueError:
-                                pass
-
-                        if crs is None:
-                            crs = src.crs
-
-                        with WarpedVRT(src, crs=crs) as vrt:
```
snowman2, line 454: This is likely where you are experiencing a performance hit. The warpedvrt only warps the data on load. rio.reproject warps everything.

snowman2: One workaround is to clip the raster to the smaller bounding box before reprojecting.

adamjstewart: Oof. Is there any feature that can reproject only a window? Or maybe we can forward more kwargs through merge_datasets -> merge_arrays -> merge? We're happy to make contributions to rioxarray as needed.

adamjstewart: Actually, our performance hit is without reprojection so far. We haven't even measured how much performance hit we get from that.

snowman2: I am on my phone, so not doing a deep dive. With that in mind, this may be worth looking into: https://corteva.github.io/rioxarray/stable/examples/clip_geom.html#Clipping-larger-rasters

adamjstewart: Does `rioxarray.merge.merge_datasets` not support windowed reading (via `bounds`) like `rasterio.merge.merge` does? Ideally we would still keep support for creating a mosaic and loading a small raster window on the fly.

snowman2: It does. However, it will reproject the datasets it the CRS/resolution don't match. The WarpedVRT likely handled that for you. It may be worthwhile to find a way to reproject a window of data in the merge logic for performance.

adamjstewart: I guess there are (at least) two issues worth investigating:

1. Performance without reprojection is 5x slower
2. Performance with reprojection is likely orders of magnitude slower for large tiles

Let me work on fixing our I/O benchmarks for reprojection so I can get you actual numbers for 2.

isaaccorley: Pinging @ljstrnadiii for xarray optimization thoughts

torchgeo/datasets/geo.py
```diff
@@ -440,23 +439,26 @@ def __init__(
...
+                        # if src.rio.crs is None:
+                        #     warnings.warn(
+                        #         f"Unable to decode coordinates of '{filepath}', "
+                        #         f'defaulting to {crs}. Set `crs` if this is incorrect.',
+                        #         UserWarning,
+                        #     )
+                        #     src = src.rio.write_crs(crs)
+
+                        if src.rio.crs != crs:
+                            src = src.rio.reproject(crs)
```
ljstrnadiii, line 454: This is where you get a performance hit. This is not a lazy operation. And neither is the rioxarray merge function (I believe). 

Calling open with Xarray or rioxarray on a file type that implies reprojection or resampling will be lazy. For example, creating a VRT and warping the VRT or using GTI and making sure the grid match but specifying the extent will ensure the resolution, extent and CRS match. 

This allows you to open it all in a lazy fashion and merge via plane old xr.merge to avoid loading and triggering all the work. And scale the operation with dask if you wanted to.

In any case, it's also worth noting that performance here can be tuned with GDAL configs

torchgeo/datasets/geo.py
ljstrnadiii: This may be incorrect if no data is not 0. Assuming it gets propagated into the resampling in GDAL, if another value is the no data value it would be used as a valid value giving incorrect results.

torchgeo/datasets/geo.py
isaaccorley: @adamjstewart Probably need to still use VRT or start using the newer GTI https://gdal.org/en/stable/drivers/raster/gti.html

adamjstewart: This line of code is not yet being hit, the 5x slower performance is without reprojection. We haven't yet measured how much of a performance hit we get with reprojection.

ljstrnadiii: Ok so it's from rioxarray .merge triggering resampling then?

ljstrnadiii: Sounds like you already store footprints making it one step easier to use GTI. Happy to chime in more if you are curious. You can also manually construct the xmls representing VRT and WarpedVRT. Both these files open by rioxarray and if bounds are defined can use xr.concat a long band and time axis to build datasets completely lazily. Then when you retrieve a patch only the underlying blocks get requested.

ljstrnadiii: Either way might want to avoid this here since you end up opening again in load VRT with merge and that might cause looping over the entire file twice.

adamjstewart: How would I check if resampling is occurring? I didn't see any resampling in the cProfile output, but could be happening deeper in the C++ code. We shouldn't have any resampling in the I/O benchmark because all files are matching CRS/resolution and have TAP.

Does GTI work with rasterio/rioxarray or only with pure GDAL? How long has it been supported?

adamjstewart:
More profiler output, this time saved as HTML files: [profile.zip](https://github.com/user-attachments/files/21971804/profile.zip)

torchgeo/datasets/geo.py
calebrob6: GTI has been around since gdal 3.9 which was last May I believe (https://gdal.org/en/stable/drivers/raster/gti.html)

adamjstewart:
I've reproduced the slow behavior without using any merging. When running the following code:
```python
import time

import rasterio as rio
import xarray as xr
from rasterio.windows import Window

# From https://huggingface.co/datasets/torchgeo/io/blob/main/preprocessed.tar.gz
file = 'data/io/preprocessed/landsat/LC09_L2SP_023032_20230620_20230622_02_T1_SR_B1.TIF'
block = 512  # COG block size
number = 1024
window = Window(block * 8, block * 8, block, block)

with rio.open(file) as src:
    start = time.time()
    for _ in range(number):
        src.read(window=window)
    end = time.time()
    rio_time = end - start

with xr.open_dataset(file, decode_coords='all') as src:
    start = time.time()
    for _ in range(number):
        src.rio.isel_window(window=window).band_data.values
    end = time.time()
    xr_time = end - start

print(xr_time / rio_time)
```
I find that (rio)xarray is ~30x slower~ **45–50x slower** than pure rasterio. Again, I expect rioxarray to have some overhead (~10%), but 3000% is a bit more than I expected. Can others reproduce this on their own systems to double check this? I think I'm using the latest versions of rio/xr/rioxr.

If we can't figure this out, we can still use (rio)xarray to load things like NetCDF/HDF5 files, but will likely want to keep the rasterio version for all other file types. Our goal is only to load a window of data (random and gridded sampling) into PyTorch in an efficient way, we don't need any of the fancy features of xarray.

EDIT: added `.band_data.values` since the data needs to be converted to numpy before being converted to a torch Tensor.

isaaccorley:
~~You're actually not using rioxarray at all in this example. You haven't imported rioxarray and you also aren't loading it using rioxarray.open_rasterio(). I'm curious if this is the cause of the slowdown. See the docs https://corteva.github.io/rioxarray/stable/getting_started/getting_started.html#rioxarray~~

hmmm I reran the benchmark with `rioxarray.open_rasterio` instead and now it's only a whopping 28x slower 🫠

adamjstewart:
I think those are the same thing, xarray.open_dataset is preferred now.

isaaccorley:
Seems like there's some overhead because rioxarray is making a new xarray DataArray from the slice where rasterio is just parsing raw pixel values to numpy

isaaccorley:
Okay I'm not sure why but using `src.rio.isel_window(window=window)` is 30x slower but using `src.isel(x=slice(window.col_off, window.col_off + window.width), y=slice(window.row_off, window.row_off + window.height))` is only 1.7x slower


```python
import time

import rasterio as rio
import xarray as xr
from rasterio.windows import Window

# From https://huggingface.co/datasets/torchgeo/io/blob/main/preprocessed.tar.gz
file = 'data/io/preprocessed/landsat/LC09_L2SP_023032_20230620_20230622_02_T1_SR_B1.TIF'
block = 512  # COG block size
number = 1024
window = Window(block * 8, block * 8, block, block)

with rio.open(file) as src:
    start = time.time()
    for _ in range(number):
        src.read(window=window)
    end = time.time()
    rio_time = end - start

with xr.open_dataset(file, decode_coords='all') as src:
    start = time.time()
    for _ in range(number):
        src.isel(
            x=slice(window.col_off, window.col_off + window.width),
            y=slice(window.row_off, window.row_off + window.height)
        )
    end = time.time()
    xr_time = end - start

print(xr_time / rio_time)
```

snowman2:
Potentially useful reference: https://github.com/corteva/rioxarray/issues/317

snowman2:
Another performance related reference: https://github.com/corteva/rioxarray/discussions/234

snowman2:
Another performance related reference: https://github.com/xarray-contrib/xvec/issues/81

adamjstewart:
> Potentially useful reference: [corteva/rioxarray#317](https://github.com/corteva/rioxarray/issues/317)

Thanks, this was actually really useful!

From https://github.com/corteva/rioxarray/issues/317#issuecomment-831487998:

> > the slowness seems to be due to the use of the .values property. If I get rid of it, the masked read is about the same speed as the non-masked read.
>
> This is due to lazy loading the data. No data is loaded from disk until called for.

@isaaccorley this explains why your code is only 1.7x slower.

I added `.band_data.values` to my original benchmarking script for fair comparison, since we have to convert to numpy before converting to a torch Tensor. Now we are up to **45–50x slower** than pure rasterio.

I also added this to @isaaccorley's script and it's _only_ 20–25x slower. The remaining gap is likely overhead from the rio accessor and all the duck magic happening there, but I would have to profile the code to check.

Note that we can't simply use `isel` as we are indexing by geocoords, not by integer. This is especially important for datasets requiring reprojection and resampling, as not all datasets will have the same TAP.

From https://github.com/corteva/rioxarray/issues/317#issuecomment-831836350:

> When doing c4ann.values[0, :, :] the code will read all 86 bands and then do the slice/indexing. With c4ann[0, :, :] the indexing happens before the read and so we read only one band.

This gives me an idea to speed up our `band_indexes` loading. Again, doesn't improve the benchmarking times, but does improve performance when someone only wants to read a subset of bands.

isaaccorley:
> > Potentially useful reference: [corteva/rioxarray#317](https://github.com/corteva/rioxarray/issues/317)
> 
> Thanks, this was actually really useful!
> 
> From [corteva/rioxarray#317 (comment)](https://github.com/corteva/rioxarray/issues/317#issuecomment-831487998):
> 
> > > the slowness seems to be due to the use of the .values property. If I get rid of it, the masked read is about the same speed as the non-masked read.
> > 
> > 
> > This is due to lazy loading the data. No data is loaded from disk until called for.
> 
> @isaaccorley this explains why your code is only 1.7x slower.
> 
> I added `.band_data.values` to my original benchmarking script for fair comparison, since we have to convert to numpy before converting to a torch Tensor. Now we are up to **45–50x slower** than pure rasterio.
> 
> I also added this to @isaaccorley's script and it's _only_ 20–25x slower. The remaining gap is likely overhead from the rio accessor and all the duck magic happening there, but I would have to profile the code to check.
> 
> Note that we can't simply use `isel` as we are indexing by geocoords, not by integer. This is especially important for datasets requiring reprojection and resampling, as not all datasets will have the same TAP.
> 
> From [corteva/rioxarray#317 (comment)](https://github.com/corteva/rioxarray/issues/317#issuecomment-831836350):
> 
> > When doing c4ann.values[0, :, :] the code will read all 86 bands and then do the slice/indexing. With c4ann[0, :, :] the indexing happens before the read and so we read only one band.
> 
> This gives me an idea to speed up our `band_indexes` loading. Again, doesn't improve the benchmarking times, but does improve performance when someone only wants to read a subset of bands.

So if I add `.compute()` after the `.sel(...).compute()` call then I'm able to see the 20-25x slower performance. However, if I use `rioxarray.open_rasterio()` instead of `xr.open_dataset()` it's only 5-7x slower. Not sure what the difference is between the open functions.

snowman2:
> However, if I use rioxarray.open_rasterio() instead of xr.open_dataset() it's only 5-7x slower. Not sure what the difference is between the open functions.

This is probably why: https://github.com/corteva/rioxarray/issues/317

adamjstewart:
Trying a similar experiment with NetCDF files:
```python
import time

import netCDF4
import xarray as xr
from rasterio.windows import Window

# From https://www.unidata.ucar.edu/software/netcdf/examples/test_echam_spectral.nc
file = 'test_echam_spectral.nc'
number = 64
window = Window(0, 0, 192, 96)

with netCDF4.Dataset(file) as src:
    start = time.time()
    for _ in range(number):
        src['tpot'][:]
    end = time.time()
    nc_time = end - start

with xr.open_dataset(file, decode_coords='all') as src:
    start = time.time()
    for _ in range(number):
        src.rio.isel_window(window=window).tpot.values
    end = time.time()
    xr_time = end - start

print(xr_time / nc_time)
```
Here it seems that xarray is only **10x slower** than pure NetCDF4. Again, most time is spent in `isel_window`, not in I/O.

We might be able to make this performant if we abandon support for reprojection and/or merging, but I don't think that's something we're willing to do right now. I'm leaning towards separate RasterDataset (rasterio) and XarrayDataset (rioxarray) base classes. This will let us add the datasets we need and improve performance later.

snowman2:
How does performance compare with `mask_and_scale=False`?

adamjstewart:
> How does performance compare with `mask_and_scale=False`?

For the minimal windowed reading benchmarking, we go from 48x slower to 35x slower. For the full TorchGeo I/O Bench benchmarking, it doesn't seem to have much effect for some reason...

adamjstewart:
Interesting behavior I just noticed that likely relates to caching. If I change the number of times we read this same window, we get different degrees of slowdown:

| # reads | x slower |
| -------: | --------: |
| 1 | 1.24x |
| 10 | 3.66x |
| 100 | 17.71x |
| 1000 | 34.20x |
| 10000 | 38.99x |

If I add one preliminary read outside of the benchmarking timing, this mostly goes away (things range from 20–40x, not 1.24–40x). Updated script:
```python
import time

import rasterio as rio
import xarray as xr
from rasterio.windows import Window

# From https://huggingface.co/datasets/torchgeo/io/blob/main/preprocessed.tar.gz
file = 'data/io/preprocessed/landsat/LC09_L2SP_023032_20230620_20230622_02_T1_SR_B1.TIF'
block = 512
number = 100
window = Window(block * 8, block * 8, block, block)

print(' Num samples:', number)

start = time.time()
for _ in range(number):
    with rio.open(file) as src:
        src.read(window=window)
end = time.time()
rio_time = end - start

start = time.time()
for _ in range(number):
    with xr.open_dataset(file, mask_and_scale=False, decode_coords='all') as src:
        src.rio.isel_window(window=window).band_data.values
end = time.time()
xr_time = end - start

print('Initial read:', xr_time / rio_time)

with rio.open(file) as src:
    src.read(window=window)
    start = time.time()
    for _ in range(number):
        src.read(window=window)
    end = time.time()
    rio_time = end - start

with xr.open_dataset(file, mask_and_scale=False, decode_coords='all') as src:
    src.rio.isel_window(window=window).band_data.values
    start = time.time()
    for _ in range(number):
        src.rio.isel_window(window=window).band_data.values
    end = time.time()
    xr_time = end - start

print(' Cached read:', xr_time / rio_time)
```
and example output:
```
 Num samples: 100
Initial read: 1.9233418533720759
 Cached read: 39.34954325631381
```

To me this suggests:

* Caching is working for both rasterio and xarray, the second read is much faster than the first
* The initial read by xarray isn't that bad (compared to rasterio)
* The cached read by xarray is pretty bad (compared to rasterio)

Another way to interpret it is there is a static amount of overhead associated with xarray, and this matters much more when loading from cache than from the initial read (which is presumably much slower).

adamjstewart:
> I'm leaning towards separate RasterDataset (rasterio) and XarrayDataset (rioxarray) base classes. This will let us add the datasets we need and improve performance later.

While we wait to figure out the performance issues, let's move forward with #1490, which creates a separate XarrayDataset using rioxarray. It isn't fast, but it does work for NetCDF/HDF5 data.

PR #2959, opened 2025-08-28
Add tutorial notebook for generating embeddings from pretrained models

calebrob6:


isaaccorley:
This LGTM. Not sure if @adamjstewart has any other nits

isaaccorley (approved):

docs/tutorials/pretrained_embeddings.ipynb
```diff
@@ -0,0 +1,395 @@
...
+   "execution_count": null,
+   "metadata": {
+    "id": "5AIQ1B9DrMOG",
+    "outputId": "6bf360ea-8f60-45cf-c96e-0eac54818079"
+   },
+   "outputs": [],
+   "source": [
+    "# On Colab, this ensures the latest TorchGeo is available.\n",
+    "\n",
+    "%pip install torchgeo"
```
adamjstewart:
```suggestion
    "%pip install torchgeo scikit-learn tqdm"
```
tqdm is pulled in by torch, but scikit-learn isn't guaranteed to be installed, so we definitely want to add these

docs/tutorials/pretrained_embeddings.ipynb
```diff
@@ -0,0 +1,395 @@
...
+   "metadata": {
+    "id": "ia5ktOVerMOL"
+   },
+   "outputs": [],
+   "source": [
+    "# Build EuroSAT DataModule (train/val loaders) using TorchGeo Lightning utilities.\n",
+    "\n",
+    "root = os.path.join(tempfile.gettempdir(), 'eurosat100')\n",
+    "datamodule = EuroSAT100DataModule(\n",
+    "    root=root, batch_size=10, num_workers=2, download=True, bands=('B02', 'B03', 'B04')\n",
```
adamjstewart: Why RGB-only?

docs/tutorials/pretrained_embeddings.ipynb
```diff
@@ -0,0 +1,395 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Fit a k-NN classifier on DOFA train embeddings and evaluate on validation embeddings.\n",
+    "# This gives a quick, label-efficient baseline without fine-tuning.\n",
```
adamjstewart: Some of these would be better suited as markdown cells. I don't love having multiple code cells in a row without explanations.

docs/tutorials/pretrained_embeddings.ipynb
```diff
@@ -0,0 +1,395 @@
...
+    "        y_val, y_pred, digits=2, target_names=class_names, zero_division=0\n",
+    "    )\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now let's do the same thing with a ResNet18 model pretrained on Sentinel-2 RGB imagery (from the SSL4EO paper).\n",
```
adamjstewart, line 264: Out of curiosity (and without running the notebook myself), which did better?

docs/tutorials/pretrained_embeddings.ipynb
```diff
@@ -0,0 +1,395 @@
...
+    ")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "pca = PCA(n_components=2, whiten=True)\n",
```
adamjstewart, line 354: t-SNE might be cool as well, but more work.

docs/tutorials/pretrained_embeddings.ipynb
```diff
@@ -0,0 +1,395 @@
...
+    "        y = batch['label']\n",
+    "        x = x.to(accelerator)\n",
+    "        if transforms is not None:\n",
+    "            x = transforms(x)\n",
+    "\n",
+    "        with torch.inference_mode():\n",
+    "            embeddings = model.forward_features(x)\n",
+    "            embeddings = torch.mean(\n",
+    "                embeddings, dim=(-2, -1)\n",
+    "            )  # global average pooling over the spatial dims\n",
```
adamjstewart: Some of these comments can be moved from in-line to their own line to reduce formatting changes

docs/tutorials/pretrained_embeddings.ipynb
```diff
@@ -0,0 +1,395 @@
...
+    "datamodule = EuroSAT100DataModule(\n",
+    "    root=root, batch_size=10, num_workers=2, download=True, bands=('B02', 'B03', 'B04')\n",
+    ")\n",
+    "datamodule.setup('fit')\n",
+    "datamodule.setup('validate')\n",
+    "datamodule.setup('test')\n",
+    "\n",
+    "train_dl = datamodule.train_dataloader()\n",
+    "val_dl = datamodule.val_dataloader()\n",
+    "test_dl = datamodule.test_dataloader()"
```
adamjstewart: The test set isn't being used at the moment. Would it be better to only use val or test since we aren't doing fine-tuning?

adamjstewart (changes requested):
I always have nits 😈

But seriously though, this is mostly good, glad we finally have this!

docs/tutorials/pretrained_embeddings.ipynb
isaaccorley: I recall there being some debate on t-SNE just being for visualization and not true dimensionality reduction. Better to play it safe and just use PCA imo.

docs/tutorials/pretrained_embeddings.ipynb
calebrob6: Because I couldn't be bothered to copy paste all the wavelengths for the other bands. I think this is something that the torchgeo DOFA model implementation should know about

docs/tutorials/pretrained_embeddings.ipynb
calebrob6: Why does comment vs. markdown matter? (anyway, moved to markdown)

docs/tutorials/pretrained_embeddings.ipynb
calebrob6: SSL4EO resnet did much better, Zhitong points out that contrastive models are better suited to KNN which may be a reason for this

calebrob6: actually just tried it with logistic regression and ssl4eo still wins -- maybe DOFA is very sensitive

docs/tutorials/pretrained_embeddings.ipynb
calebrob6: sure, will remove test

docs/tutorials/pretrained_embeddings.ipynb
calebrob6: t-SNE would just be `import sklearn.manifold.TSNE` but this doesn't show anything cool anyway because it is a tiny test dataset

calebrob6:
I think the nits are resolved

docs/tutorials/pretrained_embeddings.ipynb
adamjstewart: DOFA has never heard of any satellites, there are thousands of satellites out there, each with custom wavelengths. We train on a few, but it's intended to work for all thousands of satellites.

docs/tutorials/pretrained_embeddings.ipynb
adamjstewart: Markdown looks more like documentation and tutorial, code comments look kind of hacky and unprofessional.

docs/tutorials/pretrained_embeddings.ipynb
adamjstewart: @xiong-zhitong any thoughts? Wondering if we didn't do normalization correctly or something...

adamjstewart:
Tests are failing for the same reason as we discussed in the last TSC meeting. `download=True` is inconsistent across all datasets. For some (including EuroSAT), it only downloads the requested (or default) split, not all splits. So to fix your PR, add `download=True` to all splits. But we should probably figure out a better long-term solution and make things more consistent.

docs/tutorials/pretrained_embeddings.ipynb
isaaccorley: Let's merge sooner than waiting for people to respond. We can iterate after imo

xiong-zhitong: I think you can just merge this version first. I am looking into this now.

xiong-zhitong: @adamjstewart Just tried this version. Here is my KNN result of DOFA before normalization (the current version):
image

After a proper normalization the result is:
image

isaaccorley: Can you post the correct normalization?

xiong-zhitong: I just tested an embarrassingly simple normalization to check whether the issue comes from normalization. 
first scaling to the range 0 to 1, then applying mean-std normalization with (0.5, 0.5):

```
x /= 10000.
x = (x - 0.5) / 0.5
```

As a result, the accuracy nearly doubled for both EuroSAT and the subset EuroSAT100.

If we perform a more accurate normalization for S2 data, the results might be even better.

xiong-zhitong: Another interesting finding is that for MAE pre-trained models, using the embeddings from the second-to-last layer for linear probing may yield much better results than using the final layer.

A possible explanation is that the final layer is more heavily influenced by the reconstruction objective, while the penultimate layer retains more general and transferable representations.

isaaccorley: Nice find! This makes sense to me. Contrastive models also tend to not use embeddings from the projector but from the encoder because the projector can be biased towards the contrastive objective.

isaaccorley: It would be nice to do a comparison of which layer to take embeddings from for best results.

calebrob6:
This is failing on `pytorch.ipynb` which I didn't touch (and it was failing earlier for the same reason)

docs/tutorials/pretrained_embeddings.ipynb
calebrob6: I know, but to actually _use_ it you have to a.) know that you need to go look up the central wavelengths of the bands you are using; b.) go look up the central wavelengths of the bands that you are using; and apparently (referencing the other comments from Zhitong) c.) do some dataset specific input normalization anyway.

docs/tutorials/pretrained_embeddings.ipynb
calebrob6: @adamjstewart / @xiong-zhitong  I think https://github.com/torchgeo/torchgeo/blob/main/torchgeo/models/dofa.py#L389 is incorrect then?

xiong-zhitong: For most GeoFMs, normalizing inputs to the [0, 1] range generally works well, though mean/std normalization might perform better in some cases.

For DOFA, I think it’d be useful to add a quick note about normalization when using it. So, @calebrob6, could we add some normalization code in the [pretrained_embeddings.ipynb](https://github.com/torchgeo/torchgeo/pull/2959/files/1e8bfdcd98df0032d1e35257f2a56c40cb4b8aa8#diff-4e1e9da4d55c88cda622b5c9d3ac027f466002dbe33155a02ec9a06719138f49)
 tutorial?

Also, @adamjstewart, I suggest updating [dofa.py](https://github.com/torchgeo/torchgeo/blob/main/torchgeo/models/dofa.py)
 to default to RGB if wavelength isn’t provided. Since we’re about to release DOFA+ (DOFA2), I’ll include that in the same file as well.

After that, I’ll find some time to add a tutorial on using DOFA for object detection, semantic segmentation, regression tasks, LoRA fine-tuning, etc, using Torchgeo. There may be a lot of work. We can discuss this further offline.

xiong-zhitong: > It would be nice to do a comparison of which layer to take embeddings from for best results.

The band importance part is also interesting: https://kth.diva-portal.org/smash/get/diva2:1950547/FULLTEXT01.pdf 

However, for some tasks/datasets, using only RGB is better than using the full bands, as it introduces less noise.

calebrob6: @xiong-zhitong the default for the EuroSAT100 datamodule is to do standardization with the EuroSAT train means and stdevs

adamjstewart:
Fixed by #2853

calebrob6:
It isn't -- it uses default datamodule normalization which is mean/stdev

adamjstewart:
Does it? You're iterating over the dataloader, not the datamodule. I don't think augmentations are applied unless you use the datamodule.

calebrob6:
I will check, but then how do you iterate over a datamodule....?

adamjstewart:
You could call `datamodule.on_after_batch_transfer()` manually. I think `Trainer` does this automatically.

isaaccorley:
Yeah the augmentations get applied in the on_after_batch_transfer method which lighting calls automatically in a fit/evaluate/test/predict loop.

If you use the dataloader directly though you would need to call this method manually.

adamjstewart:
Waiting on #3000 to get the tests to pass

PR #2962, opened 2025-09-01
Replace Fiona with Geopandas Everywhere

isaaccorley:


torchgeo/datasets/geo.py
```diff
@@ -715,18 +715,19 @@ def __init__(
...
+                            lambda row: shapely.geometry.box(
+                                row['minx'], row['miny'], row['maxx'], row['maxy']
+                            ),
+                            axis=1,
                         )
-                        geometry = shapely.box(minx, miny, maxx, maxy)
-                        geometries.append(geometry)
-                except fiona.errors.FionaValueError:
-                    # Skip files that fiona is unable to read
+                except pyogrio.errors.DataSourceError:
```
isaaccorley: @adamjstewart unfortunately geopandas doesn't have it's own error that it throws when it can't read a file, it defaults to letting `pyogrio` throw an error. Basically, would have to add `pyogrio` as a dependency just for this error...

torchgeo/datasets/geo.py
```diff
@@ -896,21 +903,6 @@ def __getitem__(self, query: GeoSlice) -> dict[str, Any]:
 
         return sample
 
-    def get_label(self, feature: 'fiona.model.Feature') -> int:
```
isaaccorley, line 1058: @adamjstewart Not sure if we need to deprecate this or not. This is not as performant as using geopandas directly as this function requires a for loop and my change using geopandas doesn't. The only gotcha here is that the Eurocrops dataset overrides this method with some custom logic.

torchgeo/datasets/geo.py
adamjstewart: `DataSourceError` is a subclass of `RuntimeError`, just catch that instead.

torchgeo/datasets/geo.py
adamjstewart: It is [currently documented](https://torchgeo.readthedocs.io/en/stable/api/datasets.html#torchgeo.datasets.VectorDataset.get_label), so it's technically a public API function. If we can find a way to deprecate it, that would be ideal. If we can't, well that's what the major version bump is for. I'll mark this as currently backwards-incompatible for now.

tests/data/README.md
```diff
@@ -37,16 +37,24 @@ dst.write_colormap(1, cmap)
...
 
-import fiona
+import geopandas as gpd
+from shapely.geometry import Polygon
 
 ROOT = "data/cbf"
 FILENAME = "Ontario.geojson"
 
-rec = {"type": "Feature", "id": "0", "properties": OrderedDict(), "geometry": {"type": "Polygon", "coordinates": [[(0, 0), (0, 1), (1, 1), (1, 0), (0, 0)]]}}
-with fiona.open(os.path.join(ROOT, FILENAME), "r") as src:
```
adamjstewart, line 46: The main reason we're using fiona in these docs is so that we can copy the `src.meta` profile from the real data and use it when generating our fake data. If we can ensure the same metadata is transferred using geopandas, then that's fine. If not, let's keep using fiona in the docs.

tests/data/enviroatlas/data.py
```diff
@@ -7,8 +7,7 @@
 import shutil
 from typing import Any
 
-import fiona
-import fiona.transform
+import geopandas as gpd
```
adamjstewart: Same comment here. We technically don't have to replace fiona in our `data.py` files if we don't want to, as this code isn't used at run-time or test-time. The only incentive to replace it with geopandas is that we may someday generate test data on-the-fly.

isaaccorley: Good point. I might leave the data.py files untouched for now since they are already working.

adamjstewart:
Need to resolve merge conflicts before tests can run.

torchgeo/datasets/geo.py
isaaccorley: I elected to keep it in as I can see why it's useful if a dataset needs some custom logic to convert the labels to class indices

calebrob6:
My only real question here is around performance (and I've generally found gpd really fast so am not too worried) -- has anyone tried benchmarking this vs. previous?

adamjstewart:
We don't yet have a built-in `IOBench` for `VectorDataset`, only for `RasterDataset`, but we could probably do a quick hacky benchmark just to compare.

torchgeo/datasets/openbuildings.py
```diff
@@ -240,60 +239,38 @@ def __init__(
...
-                source_crs.to_wkt(), crs.to_wkt(), [minx, maxx], [miny, maxy]
-            )
+        # Filter to only include desired polygon files
+        gdf['filepath'] = gdf['tile_url'].str.split('/').str[-1]
+        gdf = gdf[gdf['filepath'].isin(polygon_filenames)]
 
-            filepaths.append(filepath)
-            datetimes.append((mint, maxt))
-            geometries.append(shapely.box(minx, miny, maxx, maxy))
+        # Convert geometries to bounding boxes
```
adamjstewart, line 254: Why? The whole point of geopandas is that we no longer need to use bounding boxes.

torchgeo/datasets/openbuildings.py
```diff
@@ -311,7 +288,8 @@ def __getitem__(self, query: GeoSlice) -> dict[str, Any]:
         interval = pd.Interval(t.start, t.stop)
         index = self.index.iloc[self.index.index.overlaps(interval)]
         index = index.iloc[:: t.step]
-        index = index.cx[x.start : x.stop, y.start : y.stop]
+        bbox = shapely.box(x.start, y.start, x.stop, y.stop)
+        index = index[index.geometry.intersects(bbox)]
```
adamjstewart: These are the same thing?

torchgeo/datasets/openbuildings.py
```diff
@@ -357,52 +335,24 @@ def _filter_geometries(
         x, y, _ = self._disambiguate_slice(query)
 
         # We need to know the bounding box of the query in the source CRS
```
adamjstewart, line 334: Why? Didn't we reproject the entire dataframe?

torchgeo/datasets/openbuildings.py
```diff
@@ -240,60 +239,38 @@ def __init__(
...
+            lambda row: shapely.box(row['minx'], row['miny'], row['maxx'], row['maxy']),
+            axis=1,
+        )
+        filepaths = [os.path.join(self.paths, filepath) for filepath in gdf['filepath']]
+        datetimes = [(pd.Timestamp.min, pd.Timestamp.max)] * len(filepaths)
 
         if not len(filepaths):
             raise DatasetNotFoundError(self)
 
-        data = {'filepath': filepaths}
```
adamjstewart, line 292: Unnecessary change

torchgeo/datasets/idtrees.py
```diff
@@ -399,15 +397,21 @@ def _load_geometries(self, directory: Path) -> dict[int, dict[str, Any]]:
...
-                for feature in src:
-                    # The train set has a unique id for each geometry in the properties
-                    if self.split == 'train':
-                        features[feature['properties']['id']] = feature
-                    # The test set has no unique id so create a dummy id
-                    else:
-                        features[i] = feature
-                        i += 1
+            gdf = gpd.read_file(path)
+            for _, row in gdf.iterrows():
```
adamjstewart, line 401: Looks slow, wonder if there's a way to do this without a for-loop. Not required though.

torchgeo/datasets/geo.py
```diff
@@ -452,10 +451,15 @@ def __init__(
...
+                        if src.crs == crs:
+                            geometries.append(shapely.box(*src.bounds))
                             if res is None:
-                                res = vrt.res
+                                res = src.res
+                        else:
+                            with WarpedVRT(src, crs=crs) as vrt:
+                                geometries.append(shapely.box(*vrt.bounds))
+                                if res is None:
+                                    res = vrt.res
```
adamjstewart: Let's undo changes to RasterDataset, this is being worked on in #3014

torchgeo/datasets/geo.py
```diff
@@ -874,18 +878,16 @@ def __init__(
...
-                except fiona.errors.FionaValueError:
-                    # Skip files that fiona is unable to read
+                    src = gpd.read_file(filepath, layer=layer)
+                    if crs is None:
+                        crs = src.crs
+
+                    src.to_crs(crs, inplace=True)
+                    minx, miny, maxx, maxy = src.total_bounds
+                    geom = shapely.box(minx, miny, maxx, maxy)
+                    geometries.append(geom)
```
adamjstewart, line 883: Why? Isn't the goal to have every object be its own entry in the GeoDataFrame instead of one giant bounding box for each file?

torchgeo/datasets/geo.py
```diff
@@ -936,20 +938,25 @@ def __getitem__(self, query: GeoSlice) -> dict[str, Any]:
...
 
-                # Filter geometries to those that intersect with the bounding box
-                for feature in src.filter(bbox=(minx, miny, maxx, maxy)):
-                    # Warp geometries to requested CRS
-                    shape = fiona.transform.transform_geom(
-                        src.crs, self.crs.to_wkt(), feature['geometry']
-                    )
-                    label = self.get_label(feature)
-                    shapes.append((shape, label))
+            # We need to know the bounding box of the query in the source CRS
```
adamjstewart, line 938: Not if we reproject the entire GeoDataFrame...

torchgeo/datasets/spacenet/base.py
```diff
@@ -188,20 +188,24 @@ def _load_mask(
...
-                        raster_crs.to_string(),
-                        feature['geometry'],
-                    )
-                    for feature in src
-                    if feature['geometry']
-                ]
-        except (FionaError, FionaValueError):
-            # Empty geojson files, geometries that cannot be transformed (SN7)
+            with open(path) as f:
+                json.load(f)
```
adamjstewart, line 193: Is this unused?

adamjstewart:
Given how many for-loops we are dropping, I'm not too concerned about speed on this one.

torchgeo/datasets/geo.py
isaaccorley: The fiona code above is doing the same thing though

torchgeo/datasets/geo.py
isaaccorley: I'm not understanding your comment. The reprojection is happening after the filter

torchgeo/datasets/openbuildings.py
isaaccorley: I'm just replicating the previous fiona code into geopandas. If you'd like to make other changes feel free to do so

torchgeo/datasets/geo.py
adamjstewart: Yes, we're trying to get rid of the fiona code and make the datasets more useful, not preserve backwards-compatibility.

torchgeo/datasets/openbuildings.py
isaaccorley: This was throwing a warning for me when using cx so I changed it

torchgeo/datasets/openbuildings.py
isaaccorley: True but this is cleaner imo

torchgeo/datasets/spacenet/base.py
isaaccorley: No we actually have test files that are empty. Apparently the original Spacenet1 dataset has this issue

torchgeo/datasets/openbuildings.py
isaaccorley: We reprojected the index to self.crs but the files are in self._source_crs

torchgeo/datasets/spacenet/base.py
isaaccorley: This will throw the json decode error if the file is empty

adamjstewart: Maybe add a comment explaining this

torchgeo/datasets/geo.py
adamjstewart: I guess I'm okay with merging this so we can drop a dependency, but the long-term goal is to start taking advantage of geopandas more and support non-rectangular polygons everywhere.

torchgeo/datasets/openbuildings.py
adamjstewart: Ah, this is something that would be fixed if we loaded all files and polygons once and reprojected them once. Slow upfront cost, fast and powerful later.

torchgeo/datasets/openbuildings.py
adamjstewart: Hmm, what error? I'm pretty sure the former is syntactic sugar for the latter.

torchgeo/datasets/geo.py
isaaccorley: Ahh okay I thought the goal of this PR was just to drop the dependency but yes refactoring the logic makes sense. Do we want to merge this and follow up with another PR so that logic changes are more isolated?

adamjstewart: Yeah, let's do a follow-up PR since you already did the work to get to the intermediate step.

torchgeo/datasets/eurocrops.py
```diff
@@ -192,19 +192,19 @@ def _load_class_map(self, classes: list[str] | None) -> None:
         for idx, hcat_code in enumerate(classes):
             self.class_map[hcat_code] = idx + 1
 
-    def get_label(self, feature: 'fiona.model.Feature') -> int:
+    def get_label(self, feature: pd.Series) -> int:
         """Get label value to use for rendering a feature.
 
         Args:
-            feature: the :class:`fiona.model.Feature` from which to extract the label.
+            feature: the row from the GeoDataFrame from which to extract the label.
```
adamjstewart, line 199: Why can't this be a str?

torchgeo/datasets/geo.py
adamjstewart: My comment is that if we load all data into memory, and we reproject the dataset, then there is no need to ever go back and forth again. But right now, we are only storing bounding boxes of files, so we can't yet do that.

torchgeo/datasets/openbuildings.py
adamjstewart: Switched back and no error.

torchgeo/datasets/eurocrops.py
adamjstewart: I think I understand now.

adamjstewart:
Having trouble understanding this new warning message in CI:
```
tests/datasets/test_spacenet.py::TestSpaceNet::test_getitem[SpaceNet1-1]
tests/datasets/test_spacenet.py::TestSpaceNet::test_getitem[SpaceNet6-1]
tests/trainers/test_segmentation.py::TestSemanticSegmentationTask::test_trainer[True-spacenet1]
tests/trainers/test_segmentation.py::TestSemanticSegmentationTask::test_trainer[True-spacenet1]
tests/trainers/test_segmentation.py::TestSemanticSegmentationTask::test_trainer[True-spacenet6]
  /opt/hostedtoolcache/Python/3.13.9/x64/lib/python3.13/site-packages/pyogrio/raw.py:198: RuntimeWarning: driver GeoJSON does not support open option DRIVER
    return ogr_read(
```
Seems to be related to the default open options pyogrio passes to GDAL. Not sure if it's safe to ignore or we need to change how we do things...

PR #2982, opened 2025-09-11
Change order of stacking in FieldsOfTheWorld dataset

calebrob6:
OG dataset actually stacks win_b then win_a, and this matters for reproducing the results


calebrob6:
surprised I didn't manage to get flagged by ruff

PR #2987, opened 2025-09-14
RtD: linkage requires optional model deps

adamjstewart:
Something changed in an indirect dependency and now the RtD tests are failing with:
```
torchgeo/models/yolo.py:docstring of torchgeo.models.yolo.yolo:2: WARNING: py:class reference target not found: ultralytics.YOLO
```
I'm guessing one of our dependencies dropped a dependency on ultralytics and now it's no longer available in the environment.

This PR is a short-term fix. Long-term fixes include not linking to the ultralytics docs or removing the ultralytics dependency entirely.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -15,6 +15,7 @@ build:
 python:
   install:
     - requirements: requirements/required.txt
+    - requirements: requirements/models.txt
     - requirements: requirements/docs.txt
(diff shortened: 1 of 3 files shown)


isaaccorley (dismissed):

adamjstewart:
Didn't work, they must have removed it from the docs. Next solution...

PR #2989, opened 2025-09-15
Add OpenStreetMap dataset

### LINKED ISSUE #62: OpenStreetMap dataset
We should add a GeoDataset for OpenStreetMap: https://www.openstreetmap.org/

There are a couple of approaches that we could take:

1. Query the API every time we call `__getitem__` (slow)
2. Download the data in a single large rectangle and load this (fast, more complex)

ashnair1:
Fixes #62 

Adding OpenStreetMap dataset for TorchGeo. Supports:
 - Fetching vector data from the Overpass API 
 - Supports feature types (buildings, highways, amenities, etc.) or custom Overpass queries with automatic caching & rate limiting for API requests.
 
 Sample plots:
 
 
image


calebrob6:
Super cool!!

torchgeo/datasets/openstreetmap.py
```diff
@@ -0,0 +1,417 @@
...
+    def _get_data_filename(self) -> pathlib.Path:
+        """Get the filename for the cached data file."""
+        # Create a hash of the query parameters for filename
+        cache_key = {
+            'bbox': self.bbox,
+            'feature_type': self.feature_type,
+            'custom_query': self.custom_query,
+        }
+        cache_str = json.dumps(cache_key, sort_keys=True)
+        cache_hash = hashlib.md5(cache_str.encode()).hexdigest()[:16]
```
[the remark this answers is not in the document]
ashnair1: Not applicable. `md5` is just used to create a unique identifier for the downloaded data not to secure it.

docs/api/datasets/geo_datasets.csv
```diff
@@ -27,6 +27,7 @@ Dataset,Type,Source,License,Size (px),Resolution (m)
 `NCCM`_,Masks,Sentinel-2,"CC-BY-4.0",-,10
 `NLCD`_,Masks,Landsat,"public domain",-,30
 `Open Buildings`_,Geometries,"Maxar, CNES/Airbus","CC-BY-4.0 OR ODbL-1.0",-,-
+`OpenStreetMap`_,Geometries,OpenStreetMap,ODbL-1.0,-,-
```
adamjstewart, line 30: Confirmed the license

tests/datasets/test_openstreetmap.py
```diff
@@ -0,0 +1,1022 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+
+import json
+import os
+import pathlib
+import time
+import warnings
+from typing import Any, NoReturn
```
adamjstewart, line 9: First time learning about NoReturn, interesting

tests/datasets/test_openstreetmap.py
```diff
@@ -0,0 +1,1022 @@
...
+        assert gdf.crs == 'EPSG:4326'
+
+    @pytest.mark.parametrize(
+        'element,expected_geom_type,extra_checks',
+        [
+            # Valid elements
+            (
+                {'type': 'node', 'lat': 48.8566, 'lon': 2.3523},
+                'Point',
+                lambda g: g.x == 2.3523 and g.y == 48.8566,
```
adamjstewart: It's not usually a good idea to use `==` for floats, `math.isclose()` is more robust.

tests/datasets/test_openstreetmap.py
adamjstewart, line 1: Tests are quite a bit more complicated than our typical dataset tests. That doesn't mean they're wrong, just that I don't fully understand what you're doing at all times so can't review them very well. I realize that this is a difficult one to test because you need to mock an API response, but that only explains some of the complexity. I'm guessing the rest is for directly testing helper functions.

torchgeo/datasets/openstreetmap.py
```diff
@@ -0,0 +1,555 @@
...
+    `Overpass API <https://wiki.openstreetmap.org/wiki/Overpass_API>`__
+    to query and download OSM data for a specified geographic bounding box
+    at initialization, then allows spatial querying of the cached data.
+
+    Dataset features
+    ----------------
+
+    * Vector data (points, lines, polygons) for various geographic features
+    * Flexible querying by class configuration (buildings, highways, amenities, etc.)
+    * Data fetched once at initialization and cached locally
```
adamjstewart, line 47: I'm curious how performant this is. How slow would it be to fetch small patches during each `__getitem__` call? That might be easier for global training, but I agree your approach is probably much better if you're only interested in a single city at a time.

torchgeo/datasets/openstreetmap.py
```diff
@@ -0,0 +1,555 @@
...
+        'https://overpass-api.de/api/interpreter',
+        'https://overpass.kumi.systems/api/interpreter',
+    ]
+
+    _min_request_interval = 1.0
+    _last_request_time = 0.0
+
+    def __init__(
+        self,
+        bbox: tuple[float, float, float, float],
```
adamjstewart, line 87: The rest of TorchGeo is moving towards arbitrary `shapely.Geometry` objects. But I guess the overpass API doesn't support this, only a bbox?

torchgeo/datasets/openstreetmap.py
```diff
@@ -0,0 +1,555 @@
...
+        classes: list[dict[str, Any]],
+        paths: Path | Iterable[Path] = 'data',
```
adamjstewart: It seems like `Iterable[Path]` isn't actually supported. We can remove this if we don't want to support it. I think mypy is okay with this.

torchgeo/datasets/openstreetmap.py
```diff
@@ -0,0 +1,555 @@
...
+        res: float | tuple[float, float] = (0.0001, 0.0001),
+        transforms: Callable[[dict[str, Any]], dict[str, Any]] | None = None,
+        download: bool = False,
+    ) -> None:
+        """Initialize a new OpenStreetMap dataset instance.
+
+        Args:
+            bbox: bounding box for initial data fetch as (minx, miny, maxx, maxy) in EPSG:4326
```
adamjstewart:
```suggestion
            bbox: bounding box for initial data fetch as (xmin, ymin, xmax, ymax) in EPSG:4326
```
I find this a bit less confusing.

torchgeo/datasets/openstreetmap.py
```diff
@@ -0,0 +1,555 @@
...
+            classes: list of dicts defining feature classes. Each has `name` (str) and `selector` (list of OSM tag filters).
+                Features get labels 1-N based on class order, with first match taking priority.
+            paths: root directory where dataset will be stored
+            res: resolution of the dataset in units of EPSG:4326 (degrees). Default is 0.0001°.
+                For small AOIs, consider using a finer resolution to avoid pixelated plots.
+                A good rule of thumb: ``res = min(bbox_width, bbox_height) / 400`` for ~400 pixels.
+            transforms: a function/transform that takes input sample and returns
+                a transformed version
+            download: if True, download dataset and store it in the root directory
```
adamjstewart:
```suggestion
            download: if True, download dataset and store it in the paths directory
```

torchgeo/datasets/openstreetmap.py
```diff
@@ -0,0 +1,555 @@
...
+            self._download_data()
+
+        if not self._check_integrity():
+            raise DatasetNotFoundError(self)
+
+        data_file = self._get_data_filename()
+
+        super().__init__(
+            paths=data_file,
+            crs=CRS.from_epsg(4326),  # Always use WGS84 for OSM data
```
adamjstewart, line 131: This is interesting. The user can't change the CRS at instantiation time, only after the dataset has been created. I don't see an easy way to support both without a lot of extra code, so maybe this is fine.

torchgeo/datasets/openstreetmap.py
```diff
@@ -0,0 +1,555 @@
...
+            res=res,
+            transforms=transforms,
+            label_name='label',
+        )
+
+        # Defer empty class checking to first query (lazy initialization)
+        self._empty_classes_checked = False
+
+    def _validate_classes(self, classes: list[dict[str, Any]]) -> None:
+        """Validate classes configuration."""
```
adamjstewart: Need to document inputs

torchgeo/datasets/openstreetmap.py
```diff
@@ -0,0 +1,555 @@
...
+        'https://overpass.kumi.systems/api/interpreter',
+    ]
+
+    _min_request_interval = 1.0
+    _last_request_time = 0.0
+
+    def __init__(
+        self,
+        bbox: tuple[float, float, float, float],
+        classes: list[dict[str, Any]],
```
adamjstewart, line 88: I think we can do better here. Maybe use TypedDict?

torchgeo/datasets/openstreetmap.py
```diff
@@ -0,0 +1,555 @@
...
+            crs=CRS.from_epsg(4326),  # Always use WGS84 for OSM data
+            res=res,
+            transforms=transforms,
+            label_name='label',
+        )
+
+        # Defer empty class checking to first query (lazy initialization)
+        self._empty_classes_checked = False
+
+    def _validate_classes(self, classes: list[dict[str, Any]]) -> None:
```
adamjstewart, line 140: Once we add better type hints, it may be possible to remove this method and rely on mypy.

torchgeo/datasets/openstreetmap.py
```diff
@@ -0,0 +1,555 @@
...
+            if 'name' not in class_def or 'selector' not in class_def:
+                raise ValueError(f'Class {i} must have "name" and "selector" keys')
+            if not isinstance(class_def['selector'], list):
+                raise ValueError(f'Class {i} selector must be a list')
+            for j, selector in enumerate(class_def['selector']):
+                if not isinstance(selector, dict):
+                    raise ValueError(f'Class {i} selector {j} must be a dictionary')
+
+    def _get_data_filename(self) -> pathlib.Path:
+        """Get the filename for the cached data file."""
```
adamjstewart: Document returns.

Use of md5 for file naming is interesting. I like it.

torchgeo/datasets/openstreetmap.py
```diff
@@ -0,0 +1,555 @@
...
+        """Check if the dataset file exists."""
+        return self._get_data_filename().exists()
+
+    def _rate_limit(self) -> None:
+        """Implement rate limiting for API requests."""
+        current_time = time.time()
+        elapsed = current_time - OpenStreetMap._last_request_time
+        if elapsed < self._min_request_interval:
+            time.sleep(self._min_request_interval - elapsed)
+        OpenStreetMap._last_request_time = time.time()
```
adamjstewart, line 185: Trying to decide if this is right, or if this should be some kind of classmethod, or something else. So the idea is that even if you instantiate multiple instances and query in parallel, one of them should wait for the other?

torchgeo/datasets/openstreetmap.py
```diff
@@ -0,0 +1,555 @@
...
+
+    def _build_overpass_query(self) -> str:
```
adamjstewart, line 187: I appreciate your dedication to using builtin Python modules. But if we need to add a direct dependency on requests to make this simpler, we can. We already have an indirect dependency anyway.

torchgeo/datasets/openstreetmap.py
```diff
@@ -0,0 +1,555 @@
...
+            Overpass QL query string
+        """
+        minx, miny, maxx, maxy = self.bbox
+        overpass_bbox = f'{miny},{minx},{maxy},{maxx}'
+
+        queries = []
+        for class_def in self.classes:
+            for selector in class_def['selector']:
+                for tag, values in selector.items():
+                    if values == '*':
```
adamjstewart, line 200: Are all glob expressions supported or only `*`?

torchgeo/datasets/openstreetmap.py
```diff
@@ -0,0 +1,555 @@
...
+        Args:
+            query: [xmin:xmax:xres, ymin:ymax:yres, tmin:tmax:tres] coordinates to index.
+
+        Returns:
+            Sample of input, target, and/or metadata at that index.
+
+        Raises:
+            IndexError: If *query* is not found in the index.
+        """
+        self._ensure_empty_classes_checked()
```
adamjstewart, line 355: So this method is called on the entire dataset, not just the query slice. Should this be in `__init__` instead of `__getitem__`?

torchgeo/datasets/openstreetmap.py
```diff
@@ -0,0 +1,555 @@
...
+
+            h, w = arr.shape
+            rgb = np.zeros((h, w, 3), dtype=np.float64)
+
+            for label in np.unique(arr):
+                if label == 0:
+                    continue
+                class_idx = int(label - 1)
+                if class_idx < len(colors):
+                    hex_color = colors[class_idx % len(colors)]
```
adamjstewart: A bit confused here. We have a fixed number of colors, but a variable number of classes. If `class_idx > len(colors)`, we aren't labeling the classes?

torchgeo/datasets/openstreetmap.py
```diff
@@ -0,0 +1,555 @@
...
+                    r = int(hex_color[1:3], 16) / 255.0
+                    g = int(hex_color[3:5], 16) / 255.0
+                    b = int(hex_color[5:7], 16) / 255.0
+                    rgb[arr == label] = [r, g, b]
```
adamjstewart: Instead of converting hexadecimal to float, can we just directly use this cmap during plotting?

adamjstewart (commented):
This is an absolute beast of a dataset. Great job! I fear that my ability to review this PR is limited by the size of my brain, so I left comments where things raised warnings, but there are likely dozens of things I missed. I think having multiple reviewers who actually try to run the code would be beneficial, but I don't immediately see any major red flags.

torchgeo/datasets/openstreetmap.py
ashnair1: Have switched to `requests` as it does make some things simpler

torchgeo/datasets/openstreetmap.py
ashnair1: Yes. This bbox defines the AOI that the user is interested in and overpass QL expects bbox to be like so - https://wiki.openstreetmap.org/wiki/Overpass_API/Overpass_QL#Global_bounding_box_(bbox)

torchgeo/datasets/openstreetmap.py
ashnair1: Overpass QL doesn't actually allow glob expressions. It uses regex for pattern matching. 

```python
        for class_def in self.classes:
            for selector in class_def['selector']:
                for tag, values in selector.items():
                    if values == '*':
                        # Tag exists, any value
                        queries.append(f'wr["{tag}"]({overpass_bbox});')
```

The `*` here is just a shorthand for user to retrieve all the features (`ways` and `relation` ) with the specified tag and within the specified box.

requirements/required.txt
```diff
@@ -12,6 +12,7 @@ pandas==2.3.3
 pillow==12.0.0
 pyproj==3.7.2
 rasterio==1.4.3
+requests>=2.25.0
```
adamjstewart: You'll need to add this in a few places:

- [x] pyproject.toml: `requests>=2.25`
- [x] requirements/required.txt: `requests==2.32.5`
- [x] requirements/min-reqs.old: `requests=2.25.0`

I'm hesitant about adding the type stub dependency. These aren't always maintained by the developers and not always correct.

pyproject.toml
```diff
@@ -107,6 +107,7 @@ datasets = [
     "pycocotools>=2.0.7",
     # rioxarray 0.14.1+ required for rasterio 1.4+ compatibility
     "rioxarray>=0.14.1",
+    "requests>=2.25.0",
```
adamjstewart: Need description saying why 2.25+ is required

```suggestion
    # requests 2.25+ required for ... feature
    "requests>=2.25",
```

pyproject.toml
```diff
@@ -137,6 +138,7 @@ style = [
     "mypy>=0.900",
     # ruff 0.9+ required for 2025 style guide
     "ruff>=0.9",
+    "types-requests>=2.25.0",
```
adamjstewart: Let's save this for a future PR, I'm investigating these in #3074

torchgeo/datasets/openstreetmap.py
ashnair1: Simplified this using `ListedColorMap`. `apply_cmap` is no longer needed

torchgeo/datasets/openstreetmap.py
ashnair1: Yeah that was an error. Removed this function in favour of `ListedColorMap`

torchgeo/datasets/openstreetmap.py
ashnair1: Removed `Iterable[Path]`

torchgeo/datasets/openstreetmap.py
ashnair1: So I tried using `TypedDict` but it doesn't seem to be working and mypy throws a bunch of errors.

```diff
from typing import TypedDict

class ClassConfig(TypedDict):
    """Type definition for class configuration."""
    name: str
    selector: list[dict[str, Any]]

def __init__(
    self,
- classes: list[dict[str, Any]],
+ classes: list[ClassConfig], 
   
):
```

Mypy still infers the below line as `list[dict[str, Any]]`, not `list[ClassConfig]`
   ```python
   classes = [{'name': 'building', 'selector': [{'building': '*'}]}]
   ```
   
Error raised by `mypy`:
```
tests/datasets/test_openstreetmap.py:921: error: Argument "classes" to
"OpenStreetMap" has incompatible type
"list[dict[str, Sequence[Collection[str]]]]"; expected "list[ClassConfig]" 
[arg-type]
....
```

torchgeo/datasets/openstreetmap.py
```diff
@@ -0,0 +1,551 @@
...
+
+                    msg = (
+                        f'No features found in the specified area (bbox: {bbox_str}). '
+                    )
+
+                    msg += 'Try a different feature type, larger bounding box, or different geographic area.'
+                    raise ValueError(msg)
+
+            except ValueError:
+                raise
```
adamjstewart, line 264: This feels redundant?

ashnair1: Without this, the `ValueError` would be caught by the following line `except Exception as e:` and the loop would continue trying other endpoints.

adamjstewart:
Hmm, tests pass locally. Any idea how I broke the tests?

adamjstewart:
Oh, probably because we just changed VectorDataset. Let me investigate...

adamjstewart:
Yep, bug in VectorDataset when geojson has no CRS.

adamjstewart:
The tests are failing because it's trying to rasterize a vector with zero area. The problem appears to be that we only have a single Point in the test dataset. Questions for @ashnair1:

1. How would you rasterize a Point? Wouldn't it disappear?
2. What kind of OSM features are represented as Point? Do we want to somehow support these?
3. Are we doing something wrong in the way we represent our dataset? Like, if you give it a slice, it works, but if you just ask for the bounds of the dataset, it has zero area.

I can easily replace the Point with a Polygon or replace `semantic_segmentation` with `object_detection`, but want to make sure I'm not missing anything.

ashnair1:
> The tests are failing because it's trying to rasterize a vector with zero area. The problem appears to be that we only have a single Point in the test dataset. Questions for @ashnair1:
> 
> 
> 
> 1. How would you rasterize a Point? Wouldn't it disappear?
> 
> 2. What kind of OSM features are represented as Point? Do we want to somehow support these?
> 
> 3. Are we doing something wrong in the way we represent our dataset? Like, if you give it a slice, it works, but if you just ask for the bounds of the dataset, it has zero area.
> 
> 
> 
> I can easily replace the Point with a Polygon or replace `semantic_segmentation` with `object_detection`, but want to make sure I'm not missing anything.


Certain features such as 'amenities' (clinics, gyms etc) can be presented as points (nodes in OSM terms). 

At the moment, I am not considering points. Hence why in the OSM query I only retrieve ways and relations (not nodes)

Currently I don't foresee a case to support nodes.

adamjstewart:
@isaaccorley 1d0652fe5bc455f8dc2f4213635628e92fbaa502 is a bit worrying. When a query contains no elements, geopandas crashes with an obscure shapely warning. I removed this test because I think our GeoSamplers should never return locations with no data, but I'm not 100% sure we're safe from this failure case.

adamjstewart:
Next failing test: BTC requires internet access :(

torchgeo/datasets/openstreetmap.py
ashnair1: Opened #3117 to address this

PR #2991, opened 2025-09-18
Add metadata file to Sentinel2 test data gen

adriantre:
This will make testing in #2903 much simpler.

Temporarily storing raster bounds as footprint in the metadata file. To compute the actual valid footprint it will need the new method `get_valid_footprint_from_datasource`in the above mentioned PR.

Fixes an error in the raster definition, see comment below.

- [x] Add altered raster files and new metadata file to git (required to merge?)

Plot of 10m, 20m, 60m bands and computed footprint in WGS84
image


tests/data/sentinel2/data.py
```diff
@@ -154,18 +219,33 @@ def create_file(path: str, dtype: str, num_channels: int) -> None:
...
+    profile['height'] = raster_height
+    profile['width'] = raster_width
+    # NB! .SAFE format does not include nodata value in the raster profile...
 
     if 'float' in profile['dtype']:
-        Z = np.random.randn(SIZE, SIZE).astype(profile['dtype'])
+        Z = np.random.randn(raster_height, raster_width).astype(profile['dtype'])
     else:
         Z = np.random.randint(
-            np.iinfo(profile['dtype']).max, size=(SIZE, SIZE), dtype=profile['dtype']
```
adriantre, line 166: I think this has been wrong. The contents of the raster should match the width and height of the raster definition, right?

tests/data/sentinel2/data.py
```diff
@@ -144,7 +147,75 @@
...
+    # Calculate the actual valid footprint based on pixel values
+    # by specifying nodata value. This will be stored in the metadata file.
+    with rasterio.open(raster_path) as src:
+        source_crs = src.crs
+        valid_data_footprint = shapely.box(*src.bounds)
+        # TODO: Swap to the following when get_valid_footprint_from_datasource exists
+        # with WarpedVRT(src, nodata=0) as vrt:
+        #     valid_data_footprint = shapely.convex_hull(
+        #         get_valid_footprint_from_datasource(vrt)
+        #     )
```
adriantre: Doing it this way we don't need to hard-code the footprint and keep it up-to-date with the actual pixel values.

tests/data/sentinel2/data.py
adamjstewart: I think you're right, but this code must have worked at some point? I don't know how we can fit a large raster into a small file...

adamjstewart (commented):
Finally coming back to review this. So this doesn't change any of the files yet, we need to wait for #2903 before it's useful? I can't remember my previous advice, and I apologize if this is contradictory, but maybe we should combine both PRs. Or hard-code the valid footprint of the fake test file in `data.py`. 

I really want to move forward #2903 and your other PRs, but I feel like I don't know enough about them to adequately review. I wish we had more NASA/ESA folks to review these kinds of PRs.

tests/data/sentinel2/data.py
adriantre: Indeed...

adriantre:
Added the altered test-files now.

> So this doesn't change any of the files yet, we need to wait for #2903 before it's useful? [...]  maybe we should combine both PRs. Or hard-code the valid footprint of the fake test file in `data.py`.

This PR currently uses `bounds` as footprint. How about we merge this PR as-is so that #2903 is not bloated. And then we swap out the bounds for actual footprint in #2903. Should I go for it?

adamjstewart:
Getting back to this now. Perhaps we could just simply hard-code it so it doesn't need to change twice and we don't have a circular dependency between PRs?

P.S. USGS EarthExplorer Landsat data also comes with an MTL metadata file (.txt or .xml) with very similar metadata with the true bounding box. I'm planning to add support for this alongside your Sentinel-2 changes.

adriantre:
I think I found a good way of nor hard-coding or being dependant on a method that does not yet exist. Hopefully you agree! :)

pyproject.toml
```diff
@@ -265,6 +265,7 @@ include = ["torchgeo*"]
 deprecated = "ignore"
 unresolved-attribute = "ignore"
 unresolved-import = "ignore"
+unused-ignore-comment = "ignore"
```
adamjstewart: Can we remove this? I think it's just a warning. If you want to ensure there are no warnings, you need the same versions from uv.lock.

tests/data/sentinel2/data.py
```diff
@@ -154,18 +248,28 @@ def create_file(path: str, dtype: str, num_channels: int) -> None:
...
         Z = np.random.randint(
-            np.iinfo(profile['dtype']).max, size=(SIZE, SIZE), dtype=profile['dtype']
+            np.iinfo(profile['dtype']).max,
+            size=(raster_height, raster_width),
+            dtype=profile['dtype'],
         )
 
+    if nodata_value is not None:
+        # Simulates Sentinel-2 acquisitions not fully covering the MGRS cell.
+        _apply_nodata_triangle(Z, nodata_value)
```
adamjstewart: I actually don't care too much about adding nodata pixels to the fake image, the more important thing is that the metadata exists and can be read to get the true footprint.

tests/data/sentinel2/data.py
```diff
@@ -144,7 +147,98 @@
...
+    z: np.ndarray,
+    nodata_value: float,  # type: ignore[type-arg]
+) -> None:
+    """Fill the upper-left triangle of z with nodata_value in-place."""
+    height, width = z.shape
+    rows, cols = np.ogrid[:height, :width]
+    z[rows + cols < _nodata_cutoff(height, width)] = nodata_value
+
+
+def _footprint_excluding_nodata_triangle(t: Affine, width: int, height: int) -> Polygon:
```
adamjstewart: Personally I would use a rotated square (diamond) instead of a missing triangle, that's closer to the real thing.

tests/data/sentinel2/data.py
```diff
@@ -144,7 +147,98 @@
 }
 
 
-def create_file(path: str, dtype: str, num_channels: int) -> None:
+# Minimal xml content for the gdal Sentinel2 driver to be able to read
```
adamjstewart, line 151: Is there a specific driver in GDAL for Sentinel-2? Or do you mean the jp2 driver?

tests/data/sentinel2/data.py
```diff
@@ -144,7 +147,98 @@
...
+# footprint from tags: `datasource.tags()['FOOTPRINT']`
```
adamjstewart, line 152: Will have to see if this also works for Landsat, if not we can manually parse the XML file.

adamjstewart (commented):
Looking better, but anything we can do to make this as simple as possible would make it easier to merge.

tests/data/sentinel2/data.py
adriantre: I see. I wanted this to automatically gain test coverage without mocking, but it might not be needed if we can trigger the nodata handling without it

tests/data/sentinel2/data.py
adriantre: That's the real thing when you re-project. That's never natively found in s2 products. But these triangular nodata-patches is everywhere in sentinel2. On the edge of the acquisitions where the mgrs tiles are not fully covered

https://gis.stackexchange.com/questions/395456/why-is-the-sentinel-2-image-plot-triangular-in-google-earth-engine

tests/data/sentinel2/data.py
adriantre: It's specific, and built on jp2000 driver
https://gdal.org/en/stable/drivers/raster/sentinel2.html

adamjstewart: Interesting! We'll need to change our glob/regex to search for the XML file but this does indeed look much simpler than our current approach.

tests/data/sentinel2/data.py
adamjstewart: Whatever works, as long as it's simple we can match them.

pyproject.toml
adriantre: My pre-commit complained, but I'll see if it has stale versions.

tests/data/sentinel2/data.py
adriantre: Thanks for pushing me on this! I think I found a nice way where we don't have to match the logic between raster and vector.

pyproject.toml
adriantre: I think it was due to missing version specification of ty in pre-commit.

adamjstewart (approved):
I think I understand most of this? As long as the Sentinel driver can read it, it should be valid.

adriantre:
I'm not able to merge it, but when it is I will update #2903 👍

adamjstewart:
Yep, just waiting for one other maintainer to review before merging.

robmarkcole (approved):

PR #3010, opened 2025-10-03
PASTIS: correct copyright

adamjstewart:
@yichiac


yichiac:
Thanks for the update! It's TorchGeo Contributors now :)

Issue #3009, opened 2025-10-03
Assign spatial reference by Ground Control Points


adriantre:
### Summary

Some rasters does not have a CRS defined, and thus can not be used directly in torchgeo, without writing a new file by something like `gdal_translate -a_ullr ... -a_srs EPSG:4326`. This could also be done on-the-fly based on Ground Control Points instead.

### Rationale

Probably the recommended way is to write new files first, but I wanted to put this suggestion on here in case someone has the same need in the future.

In the below sketched solution I wrap rasterio.open to keep all details out of RasterDataset. It also removed the warp-logic out of RasterDataset. Could be extended to a contextmanager and cache `from_gcps` to save compute.

This derives an approximate affine georeferencing from the ground control points, and then performs an on-the-fly reprojection (like RasterDataset does in `__init__` and `_load_warp_file`). The user could swap this for a more advanced warping if they know what they want.

Thoughts?

### Implementation

```python
from pathlib import Path
import rasterio
from rasterio.vrt import WarpedVRT
from rasterio.transform import from_gcps
import geopandas as gpd
from shapely.geometry import Point

def open_georef_raster(filepath: Path, target_crs=None) -> rasterio.io.DatasetReader:
    """
    Open a raster and ensure it has a CRS/transform.

    If the raster only has GCPs, compute a virtual CRS + transform.
    Optionally, automatically warp to the "best" UTM CRS based on raster GCPs.

    Args:
        filepath: path to raster
        target_crs: if specified, return a WarpedVRT in this CRS.
                    If None, and raster only has GCPs, will attempt UTM.

    Returns:
        DatasetReader or WarpedVRT
    """
    src = rasterio.open(filepath)

    # Compute virtual CRS + transform if CRS is missing
    if src.crs is None:
        gcps, gcps_crs = src.gcps
        if gcps and gcps_crs is not None:
            virtual_crs = gcps_crs
            virtual_transform = from_gcps(gcps)
        else:
            virtual_crs = None
            virtual_transform = None
    else:
        virtual_crs = src.crs
        virtual_transform = src.transform

    # Optionally estimate best utm crs here

    # Apply warp if requested
    if target_crs is not None and virtual_crs != target_crs:
        vrt_options = {"crs": target_crs}
        if virtual_transform is not None:
            vrt_options["transform"] = virtual_transform
        vrt = WarpedVRT(src, **vrt_options)
        src.close()
        return vrt
    else:
[...]

PR #3014, opened 2025-10-06
Move all dataset open and reproject to

adriantre:


torchgeo/datasets/geo.py
```diff
@@ -603,20 +598,33 @@ def _cached_load_warp_file(self, filepath: Path) -> DatasetReader:
...
         """
+        if crs is None and hasattr(self, 'index'):
+            crs = self.crs
+
         src = rasterio.open(filepath)
 
+        # See if file has a color map
+        if len(self.cmap) == 0:
+            try:
+                self.cmap = src.colormap(1)  # type: ignore[misc]
```
adriantre: I moved this from `__init__` as I'm not sure if colormap is preserved in the vrt always. If it is, I think it is better to move it back. Or perform a rasterio.open right before plotting, instead of in init

adamjstewart:
Note that we also have a new `XarrayDataset` that works a bit differently, as it uses rioxarray and not all files will contain CRS info. We don't have to add this feature to both, but I'm trying to keep these APIs as similar as we can. Maybe let's iterate and get this perfected for `RasterDataset` before thinking about `XarrayDataset`.

adriantre:
I had a look at `XarrayDataset` now, it is a bit different indeed. I don't fully understand the differences between its `__init__` and `__getitiem__`, especially regarding projecting with or without `res` input and `decode_times=True`. But I assume it should be possible to do something similar there, like:

```python
# __getitem__
datasets = [self._load_warp_file(fp) for fp in filepaths]
```

But I agree, lets start with `RasterDataset`.

torchgeo/datasets/geo.py
adriantre: I believe colormap may only get lost if the dtype changes, which I guess is never the case here. Would love a second opinion though!

If you agree, I'll move it back to init, and get the colormap from the vrt

adriantre:
What do you think @adamjstewart am I on the right track here? Did I miss anything?

adamjstewart:
> I don't fully understand the differences between its `__init__` and `__getitiem__`, especially regarding projecting with or without `res` input and `decode_times=True`.

In `__init__`, we only need to grab the bounding box, so `res` is useless. In `__getitem__`, we may need to resample, hence the `res`. I'm not sure if or when `decode_times=True` is needed, I was hoping to add a few more datasets to debug this.

torchgeo/datasets/geo.py
```diff
@@ -603,20 +598,33 @@ def _cached_load_warp_file(self, filepath: Path) -> DatasetReader:
         """
         return self._load_warp_file(filepath)
 
-    def _load_warp_file(self, filepath: Path) -> DatasetReader:
+    def _load_warp_file(self, filepath: Path, crs: CRS | None = None) -> DatasetReader:
```
adamjstewart, line 665: What if we make crs always required? So whenever this is called, it's called with `crs`. That would make the function much simpler.

torchgeo/datasets/geo.py
adamjstewart: If the tests pass, I'm fine with moving colormap back. 

Honestly, I'm thinking about removing automatically reading colormap from files and requiring it to be set in the dataset. It isn't in a format that can quickly and easily be used by matplotlib anyway. But that's a job for a different PR.

adamjstewart:
I think you're on the right track. I don't see any particular issues with this approach.

calebrob6:
This makes sense to me, anything blocking this @adamjstewart ?

adamjstewart:
Just the couple comments I left behind. Could probably get this in a mergeable state pretty quickly.

torchgeo/datasets/geo.py
adamjstewart: Because we don't know the file CRS until we actually load the file the first time.

torchgeo/datasets/geo.py
```diff
@@ -615,20 +610,33 @@ def _cached_load_warp_file(self, filepath: Path) -> DatasetReader:
...
         Args:
             filepath: file to load and warp
+            crs: Optionally specify which CRS to reproject to. This is used in __init__
+                as self.index.crs is not defined at this point.
 
         Returns:
             file handle of warped VRT
         """
+        if crs is None and hasattr(self, 'index'):
+            crs = self.crs
```
adamjstewart: A bit confused here. If the user-provided `crs` is None (the default), the first time this is called, `self.index` will not yet exist. So `crs` will remain None, and `src.crs` will not be None, so we are creating a `WarpedVRT` to None?

torchgeo/datasets/geo.py
```diff
@@ -437,25 +437,17 @@ def __init__(
         filepaths = []
         datetimes = []
         geometries = []
+        vrt = None
```
adamjstewart: I'm wondering if this is the right place to put this, or if it should be further in. If something goes wrong in the loop, we will break out and try to close `vrt`. But if something goes wrong twice, `vrt` will already be closed.

adamjstewart:
A bit on the fence with this PR. I can hack it into shape and merge it, but it's surprisingly more complicated than I expected. Not sure if it's wise to merge this at the last minute without other people to review my review.

torchgeo/datasets/geo.py
adriantre: Makes sense, moved it in just before try:

torchgeo/datasets/geo.py
adriantre: Looking back at this, I am confused myself :D

I guess we should use src.crs the first time. Pushing a suggestion

torchgeo/datasets/geo.py
adriantre: Moved it back to init

adriantre:
Good to go? This will solve #3009 for me as I can now easily customize opening logic

adamjstewart (approved):
This looks correct as far as I can tell but would like to get another set of 👀 on this before merging.

Note that VectorDataset and XarrayDataset do not yet support this. That's fine with me since it isn't part of the public API anyway. We can make things public or add to other base classes later though.

torchgeo/datasets/geo.py
```diff
@@ -661,28 +662,37 @@ def _cached_load_warp_file(self, filepath: Path) -> DatasetReader:
...
         Returns:
             file handle of warped VRT
         """
         src = rasterio.open(filepath)
+
+        if crs is None:
+            try:
+                crs = self.crs
+            except AttributeError:
+                crs = src.crs
```
isaaccorley, line 682: This is a nit but I prefer not to use try/except everywhere if possible in case AttributeError is caused somewhere inside the crs implementation.

Happy to merge without this change though.

```suggestion
        if crs is None:
            crs = getattr(self, 'crs', src.crs)
```

isaaccorley (approved):

torchgeo/datasets/geo.py
adamjstewart: Agreed, just curious if this works since `self.crs` is a property function. So technically it's always defined, although calling it may result in an AttributeError. May be safer to check for `self.index.crs` directly if this doesn't work.

PR #3028, opened 2025-10-09
ExtractPatches: correct handling of time series dimension

### LINKED ISSUE #2920: Change Detection transforms

hkristen:
This PR is split out from #2910 as discussed with @adamjstewart to make reviewing of it easier. Should fix #2920 - @blaz-r plz confirm

- ~Add val_patch_sampling parameter to LEVIRCDDataModule ('random' or 'grid')~
- Fix temporal correspondence in patch extraction for bi-temporal images
- ~Unify LEVIRCDDataModule to support both LEVIR-CD and LEVIR-CD+ datasets~
- Add tests for temporal correspondence, patch extraction, and sampling strategies


blaz-r:
Hi, I will try this now and report back how it goes.

torchgeo/datamodules/levircd.py
```diff
@@ -34,117 +44,162 @@ def __init__(
...
+        # If patches were extracted (grid mode), reshape for ChangeViT compatibility
+        # In random crop mode, batch is already [B, T, C, H, W] so no reshaping needed
+        if len(batch['image'].shape) == 6:  # [B, T, P, C, H, W]
+            batch_size, temporal_frames, patches_per_frame = batch['image'].shape[:3]
+
+            # Reshape image: [B, T, P, C, H, W] -> [B*P, T, C, H, W]
+            batch['image'] = batch['image'].view(
+                batch_size * patches_per_frame,
+                temporal_frames,
+                *batch['image'].shape[3:],
```
blaz-r: Isn't this incorrect? If the shape really is [B, T, P, ...], then to get to [B*P, T...] we'd need to do permute and then view (like it's done for the mask). This might also explain the issue you faced in #2910 when using grid mode. I will have to verify this though.

For such cases, I think Einops could be cleaner and can even prevent such problems. This would be a simple transform of form "b t p c h w -> (b p) t c h w". Any opionions on this @adamjstewart  @isaaccorley

blaz-r: This was indeed the situation. It added a permute then reshape and it works fine:
```python
# Reshape image: [B, T, P, C, H, W] -> [B*P, T, C, H, W]
batch['image'] = batch['image'].permute(
        0, 2, 1, 3, 4, 5
    ).reshape(
    batch_size * patches_per_frame,
    temporal_frames,
    *batch['image'].shape[3:],
)
```
As I said in my opinion einops makes this even better, since this unclear notation becomes super clean:
```python
batch["image"] = einops.rearrange(batch["image"], 'b t p c h w -> (b p) t c h w')
```
It even reduces the need for comment 😄

blaz-r:
Other than the comment on the grid thing, the code seems to work and trains on BTC fine. I will need to train an actual run to fully verify this but I think this is great. Thank you for all the work 😃

torchgeo/datamodules/levircd.py
hkristen: Nice catch! +1 for einops

hkristen:
> Other than the comment on the grid thing, the code seems to work and trains on BTC fine. I will need to train an actual run to fully verify this but I think this is great. Thank you for all the work 😃

Thank you for the feedback - looking forward to hear from your full run results :)

blaz-r:
The run has finished. While the metrics are not exactly reproduced, I think it's not the dataloaders' fault. So I think that from a functional point of view this is fine once those reshapes are fixed.

hkristen:
@isaaccorley or @adamjstewart I think this one is ready for review.

adamjstewart:
Any reason we need to get rid of `LEVIRCDPlusDataModule`? If there is a lot of shared code, we could also make a subclass or shared base class.

torchgeo/datamodules/levircd.py
adamjstewart: +1 for einops

torchgeo/transforms/transforms.py
```diff
@@ -87,8 +89,25 @@ def apply_transform(
             padding=flags['padding'],
         )
 
-        if flags['keepdim']:
-            out = rearrange(out, 'b t c h w -> (b t) c h w')
+        # Handle temporal data from VideoSequential
+        # If we extracted multiple patches and batch size suggests temporal data (even number),
```
adamjstewart: Even batch size does not suggest temporal data...

tests/trainers/test_change.py
```diff
@@ -341,3 +341,171 @@ def test_multiclass_predict(self, fast_dev_run: bool) -> None:
...
         )
         trainer.predict(model=model, datamodule=datamodule)
+
+    def test_levircd_temporal_correspondence(self) -> None:
+        """Test LEVIR-CD temporal correspondence issues from GitHub issue #2920."""
+        pytest.importorskip(
+            'torchgeo.datamodules.levircd', reason='LEVIRCDDataModule required'
+        )
+
+        from torchgeo.datamodules.levircd import LEVIRCDDataModule
```
adamjstewart: Why would this ever not be importable? I think you can skip this and import at the top.

I appreciate the thorough testing, but I do wonder if it's too much. Perhaps we could instead unit test the transform? Or move the tests to `tests/datamodules/test_levircd.py`? I'm trying to slowly get rid of those tests though. I also eventually want to move this transform to Kornia so we don't have to maintain it anymore.

hkristen: Good point - I moved those tests to tests/datamodules/test_levircd.py as unit tests to make the organization clearer and avoid the trainer overhead.

hkristen:
> Any reason we need to get rid of `LEVIRCDPlusDataModule`? If there is a lot of shared code, we could also make a subclass or shared base class.

The two classes were almost identical - same transforms, same augmentation, just different dataset variants and val_split_pct defaults. Since those are configuration parameters, I unified them into one class rather than maintaining duplicate code. A subclass approach would work too, but seemed like overkill for just two config parameters.

adamjstewart:
> > Any reason we need to get rid of LEVIRCDPlusDataModule? If there is a lot of shared code, we could also make a subclass or shared base class.
>
> The two classes were almost identical - same transforms, same augmentation, just different dataset variants and val_split_pct defaults. Since those are configuration parameters, I unified them into one class rather than maintaining duplicate code. A subclass approach would work too, but seemed like overkill for just two config parameters.

We have two different dataset classes, I would prefer to also have two different datamodule classes. This also avoids backwards-incompatible changes.

torchgeo/transforms/transforms.py
```diff
@@ -87,8 +89,24 @@ def apply_transform(
...
 
-        if flags['keepdim']:
-            out = rearrange(out, 'b t c h w -> (b t) c h w')
+        # Handle temporal data from VideoSequential
+        # Rearrange to group patches by spatial location rather than temporal sequence
+        # Only apply this fix when keepdim=True (for flattening compatibility)
+        if (
+            len(out.shape) == 5
+            and out.shape[1] > 1
+            and input.shape[0] % 2 == 0
```
adamjstewart: Don't really understand these two lines

torchgeo/transforms/transforms.py
```diff
@@ -87,8 +89,24 @@ def apply_transform(
...
+        ):
+            # Rearrange to group patches by spatial location
+            # Current: patches from t1, then patches from t2
+            # Desired: patch_0 from [t1,t2], patch_1 from [t1,t2], etc.
+            temporal_frames = 2
+            out = rearrange(out, '(b t) n c h w -> (b n t) c h w', t=temporal_frames)
+        elif flags['keepdim']:
+            # Original behavior - flatten patches into batch dimension
+            out = rearrange(out, 'b n c h w -> (b n) c h w')
+        # If keepdim=False, keep the [B, N, C, H, W] shape as is
```
adamjstewart: This behavior is a bit confusing to me. I would expect something like:
```python
if flags['keepdim']:
    if out.shape == 5:
        out = rearrange(out, 'b t n c h w -> (b n) t c h w')
    elif out.shape == 4:
        out = rearrange(out, 'b n c h w -> (b n) c h w')
```
So if the input sequence has a time series, the output sequence should also have a time series.

tests/datamodules/test_levircd.py
adamjstewart, line 1: I know you added these for a good reason, I just don't know what it is. Are the typical trainer tests insufficient for testing these things? Can we add additional tests to `tests/transforms/test_transforms.py` to cover this? If we can get 100% test coverage and relatively decent guarantees of correct behavior without this file, my goal is to eventually remove everything in `tests/datamodules`.

torchgeo/datamodules/levircd.py
```diff
@@ -18,14 +20,21 @@
 class LEVIRCDDataModule(NonGeoDataModule):
     """LightningDataModule implementation for the LEVIR-CD dataset.
 
+    Implements modern random training + deterministic evaluation strategy
```
adamjstewart: What does "modern" mean in this context?

torchgeo/datamodules/levircd.py
```diff
@@ -18,14 +20,21 @@
...
+    with proper temporal correspondence fixes.
+
     .. versionadded:: 0.6
+    .. versionchanged:: 0.8
+        Fixed temporal correspondence issues from GitHub issue #2920.
+        Added val_patch_sampling parameter for flexible evaluation strategies.
```
adamjstewart: Only API changes (new parameters) need to be documented, not internals and bugs

torchgeo/datamodules/levircd.py
```diff
@@ -35,50 +44,90 @@ def __init__(
             patch_size: Size of each patch, either ``size`` or ``(height, width)``.
                 Should be a multiple of 32 for most segmentation architectures.
             num_workers: Number of workers for parallel data loading.
+            val_patch_sampling: Strategy for validation patch sampling:
```
adamjstewart: It's also used for test, right? From what I understand, random is bad and you should always use grid? Do we even need a parameter then?

torchgeo/datamodules/levircd.py
```diff
@@ -35,50 +44,90 @@ def __init__(
...
+                'random' (default): Use RandomCrop, matches training distribution.
+                'grid': Use deterministic grid patches via ExtractPatches.
             **kwargs: Additional keyword arguments passed to
                 :class:`~torchgeo.datasets.LEVIRCD`.
         """
+        stride = kwargs.pop('stride', None)
```
adamjstewart: I don't think `stride` is documented? Do we need it? Shouldn't we always set stride == patch size to avoid double counting pixels?

torchgeo/datamodules/levircd.py
```diff
@@ -35,50 +44,90 @@ def __init__(
...
-            K.VideoSequential(
-                K.Normalize(mean=self.mean, std=self.std),
-                _ExtractPatches(window_size=self.patch_size),
-            ),
+        self.train_aug = self._create_random_train_aug()
+        self.val_aug = self._create_deterministic_val_aug()
+        self.test_aug = self._create_deterministic_val_aug()
+        self.aug = self.val_aug
+
+    def _create_random_train_aug(self) -> K.AugmentationSequential:
```
adamjstewart: Do we really need helper functions for this? We don't do this in any other datamodule

torchgeo/datamodules/levircd.py
```diff
@@ -35,50 +44,90 @@ def __init__(
...
+                K.Normalize(mean=self.mean, std=self.std),
+                _ExtractPatches(
+                    window_size=self.patch_size, stride=self.stride, keepdim=False
+                ),
+                data_keys=None,
+                keepdim=False,
+                same_on_batch=True,
+            )
+
+    def on_after_batch_transfer(
```
adamjstewart: Shouldn't this happen in the trainer and depend on model? We shouldn't have any code specific to one model or another.

torchgeo/datamodules/levircd.py
hkristen: That was a word CC sneaked in when reviewing my code :P

torchgeo/datamodules/levircd.py
hkristen: This parameter exists because of a distribution mismatch issue during ChangeViT development in #2910. Training with random crops achieved 85% F1, but validation with fixed grid patches dropped to 11.7% F1 on the same checkpoint - the model learned position-dependent features. Switching validation to random crops improved it to 81.6%, catching the overfitting that grid  sampling missed.

So I suggest to keep the parameter and use it for validation, but always use grid for test.

torchgeo/datamodules/levircd.py
hkristen: You have a point there, I will simplify this to match the other datamodules.

torchgeo/datamodules/levircd.py
hkristen: This is a leftover from debugging sessions, I will get rid of it.

tests/datamodules/test_levircd.py
hkristen: The tests were added while debugging the temporal correspondence bug (#2920). Most were testing the ~~now-removed~~ on_after_batch_transfer method. I've deleted tests/datamodules/test_levircd.py entirely and moved test_extract_patches_temporal_ordering to tests/transforms/test_transforms.py where it belongs.

torchgeo/datamodules/levircd.py
adamjstewart: I don't understand. If you are training with random crops, you can't learn position-dependent features, regardless of how validation is performed. There is no learning going on during validation, only inference. Validation and test should have a similar procedure, otherwise validation isn't a useful gauge of test accuracy.

blaz-r: I'm fairly certain that the reason the fixed grid didn't work was due to the incorrect reshape, not the distribution mismatch (check the review comment I left above for more details).

hkristen: Fair points guys! Thanks for pointing this out - I have changed it so that validation & test always use grid sampling while training uses random crops.

torchgeo/datamodules/levircd.py
hkristen: You're right that this feels like model-specific logic. However, I'm struggling to find a clean solution. This is kinda my workaround, because:

1. LEVIR-CD returns temporal data [B, T=2, C, H, W] (5D)
2. Kornia's K.AugmentationSequential can't handle 5D data (expects 2-4D)
3. Most CD models expect [B, T, C, H, W], but patch-based models like ChangeViT need [B*P, T, C, H, W]

While having a closer look I realized that currently, this method doesn't even apply augmentations (doesn't call super()) .. so one could either:
- Move the reshape to ChangeViT trainer (#2910) and fix augmentations here
- Apply transforms in the dataset before batching (but loses batch augmentation benefits)
- Fix the base NonGeoDataModule to handle temporal data

What approach would you prefer? I'm open to suggestions on the cleanest architecture here or help with pushing this over the finish line :P

adamjstewart: AugmentationSequential requires 4D, but VideoSequential supports 5D. We already do the reshape you are describing in ChangeDetectionTask.

adamjstewart:
I was curious why our current unit tests did not catch this bug. It turns out, all of our test data has the same size as our cropped patches. As a result, we always create a new dimension of size 1, which is compatible with everything. The most recent commit ensures that all test data is 4x larger than the cropped patch size, reproducing the bug.

Issue #1912, opened 2024-02-27
Sentinel 2 dataset can't see files downloaded from Copernicus Browser - filename doesn't fit regex


soheir96:
### Description

Hi All,
I've downloaded band files for Sentinel 2A from the Copernicus Browser, and I think Copernicus might have changed their file naming style from what they used before? 
File names I had downloaded:
```
2023-06-14-00_00_2023-06-14-23_59_Sentinel-2_L2A_B02_(Raw).tiff
2023-06-14-00_00_2023-06-14-23_59_Sentinel-2_L2A_B03_(Raw).tiff
2023-06-14-00_00_2023-06-14-23_59_Sentinel-2_L2A_B04_(Raw).tiff
```
Trying to use the load dataset function described in the docs (https://torchgeo.readthedocs.io/en/stable/api/datasets.html#torchgeo.datasets.Sentinel2) results in the following error 
```
[FileNotFoundError](https://docs.python.org/3/library/exceptions.html#FileNotFoundError) – if no files are found in paths
``` 
as it can't recognise the filenames as they differ from the expected format.

### Steps to reproduce

1. Download files from Copernicus Browser 
2. Try to load files using Sentinel2 dataset:
```
dataset = Sentinel2(paths= PATH_TO_DATA)
```

### Version

0.5.1

adamjstewart:
Updating `filename_regex` is easy because we can use "or" in the regex, but updating `date_format` is much harder because only a single value is currently allowed. We would have to allow multiple possible values of `date_format` and try all of them. Do you know if there's any documentation that describes the Copernicus Browser naming scheme? It's frustrating when there are [several official download sites](https://www.copernicus.eu/en/copernicus-satellite-data-access) and they all use a different naming scheme...

FWIW, you can always subclass `Sentinel2` and write your own custom regex/format. But I would love to be able to support this out of the box if we can.

adamjstewart:
I just downloaded a scene from Copernicus Browser but it matches the filename hierarchy that we test in https://github.com/torchgeo/torchgeo/tree/main/tests/data/sentinel2. Can you check and see if this is still an issue? Perhaps you're downloading one file at a time instead of the whole scene?

adamjstewart:
Perhaps this is processed via Sentinel Hub? https://forum.dataspace.copernicus.eu/t/copernicus-browser-raw-data-of-sentinel-2-l2a/3142

PR #3043, opened 2025-10-17
Edit RasterDataset and Sentinel2 to work with Sentinel-2 L2A .SAFE -products

mayrajeo:
This PR fixes the error when trying to create `Sentinel2` dataset from L2A .SAFE products. L1C and L2A products have different directory structures: L1C has all the band files in a single folder, making it simple to get the data. L2A products, however, have several versions of each band in multiple resolutions: 

* B02, B03, B04 in 10m, 20m and 60m resolution
* B08 only on 10m resolution
* B05, B06, B07, B11, B12 in 20m and 60m resolution (Some .SAFE products include B01 also in 20m)
* B01 and B09 in 60m resolution

And these data have different filenames and are in different folders depending on the resolution.

This PR fixes it by editing the filenames and directories based on the most accurate resolution available for the band in question. It is now on `RasterDataset.__getitem__`, but having e.g. `Sentinel2.__getitem__` specified for these products might be possible. Also, `filename_glob` now always uses `B04`, because otherwise the code would try to look for `B01` in 10m resolution, which doesn't exist.

Tests are also edited to use all L2A bands.

Note: As L2A products do not contain B10, bands must still be specified when creating the dataset.


mayrajeo:
Test seem to be failing because they try to load B10 from L2A data, which does not exist.

adamjstewart:
Very much want this feature, people have been begging for this for years.

I've always been on the fence between doing this in RasterDataset vs. Sentinel2Dataset. Are there any other satellites/sensors that do this?

Don't yet worry about the tests, we can fix the test data if it isn't consistent with reality.

mayrajeo:
I guess the main reason for having Sentinel2Dataset would not having hacks required only for Sentinel-2 data in RasterDataset. I'm not sure whether there are other satellites that have this kind of file structure.

adamjstewart:
Will this fix #1912 by any chance?

mayrajeo:
Good question, data with that kind of naming convention seem to be [processed in SentinelHub](https://forum.dataspace.copernicus.eu/t/copernicus-browser-raw-data-of-sentinel-2-l2a/3142). I don't have experience with SentinelHub processing, and I don't actually know how to get those types of files from Copernicus Browser.

adamjstewart:
At the moment, this is only relevant to Sentinel2. I don't really want to make RasterDataset more complicated, but I also don't want to make you copy-n-paste all of RasterDataset into Sentinel2. Here are a couple more ideas before we commit to either option:

1. We could add support for `filename_glob` and `filename_regex` to include the full filepath, not just the filename. By default, we could use `*` for the resolution, sort, and choose the first one (lowest resolution). This only works when filepaths are sortable by resolution, but so far that's the case.
2. We could add a helper function that converts from one band filepath to another. By default, this would use the current filename_regex-based implementation. This tiny helper function could then be overridden in Sentinel2-only.

We should probably start adding fake test data in `tests/data/sentinel2` for this new directory structure. That will make it easier to test this.

mayrajeo:
Changed so that `RasterDataset` has 

```python
    def _update_filepath(self, band:str, filepath:str) -> str:
        """Update `filepath` to point to `band`
        
        Args:
            band: band to search for
            filepath: base filepath to use for searching

        Returns:
            updated filepath for `band`
        """
        filename = os.path.basename(filepath)
        directory = os.path.dirname(filepath)
        match = re.match(filename_regex, filename)
        if match:
            if 'band' in match.groupdict():
                start = match.start('band')
                end = match.end('band')
                filename = filename[:start] + band + filename[end:]
        filepath = os.path.join(directory, filename)
        return filepath
```

 so in `__getitem__` we can just do

```python
            data_list: list[Tensor] = []
            for band in self.bands:
                band_filepaths = []
                for filepath in index.filepath:
                    filepath = self._update_filepath(band, filepath)
                    band_filepaths.append(filepath)
                data_list.append(self._merge_files(band_filepaths, query))
            data = torch.cat(data_list)
```

And have the more complex logic in `Sentinel2._update_filepath`.

adamjstewart:
I like this implementation better. Do you want to update `tests/data/sentinel2/*` for this new directory structure so we can properly test?

Do all Sentinel-2 downloads follow this new directory structure? Would be good to compare a few different sites. We unfortunately need to support all of them, not just .SAFE.

mayrajeo:
`tests/data/sentinel2` already has the correct structure for SAFE, as there are one L1C (All files in one directory) and one L2A (Separate directories for each resolution) there. A few failing tests need some tweaks, either specify to only use L1C data or use both but only 12 bands. I can check whichever is easier.

SentinelHub seems to be at least one of the other sources for the data, but that should be able to be fixed with some regex magic. USGS Sentinel-2 archive has been decommissioned, and various STAC catalogs all have different specifications.

adamjstewart:
I'm not too worried about STAC (we will add a separate generic STACDataset soon) but would like to support as many ESA/NASA/USGS sources as possible.

tests/conf/sentinel2_cdl.yaml
```diff
@@ -16,3 +16,18 @@ data:
...
+        'B04',
+        'B05',
+        'B06',
+        'B07',
+        'B08',
+        'B8A',
+        'B09',
+        'B11',
+        'B12',
+      ]
```
adamjstewart: YAML has its own list syntax, something like:

```suggestion
    sentinel2_bands:
      - 'B01'
      - 'B02'
      - 'B03'
      - 'B04'
      - 'B05'
      - 'B06'
      - 'B07'
      - 'B08'
      - 'B8A'
      - 'B09'
      - 'B11'
      - 'B12'
```

adamjstewart:
@sfalkena you've been getting your hands dirty in `RasterDataset.__getitem__` lately and have strong opinions on how we should handle multi-band file datasets. Can you review this PR and see what you think?

adamjstewart:
FYI, you never need to merge main into your branch.

calebrob6:
> FYI, you never need to merge main into your branch.

As in always rebase instead?

adamjstewart:
> As in always rebase instead?

As in the PR only needs to be merged/rebased iff it has a conflict with main. It isn't an issue if it isn't up-to-date with the exact newest commit.

torchgeo/datasets/geo.py
```diff
@@ -560,6 +551,28 @@ def __getitem__(self, query: GeoSlice) -> dict[str, Any]:
...
+        """Update `filepath` to point to `band`.
+
+        Args:
+            band: band to search for.
+            filepath: base filepath to use for searching.
+
+        Returns:
+            updated filepath for `band`.
+        """
+        filename_regex = re.compile(self.filename_regex, re.VERBOSE)
```
adamjstewart: This is kinda slow, and now run for every band of every file. I think we can skip this line, Python will compile and cache it automatically.

adamjstewart: Reference: https://stackoverflow.com/questions/452104/is-it-worth-using-pythons-re-compile

torchgeo/datasets/sentinel.py
```diff
@@ -360,14 +381,54 @@ def __init__(
             *root* was renamed to *paths*
         """
         bands = bands or self.all_bands
-        self.filename_glob = self.filename_glob.format(bands[0])
-
+        self.filename_glob = self.filename_glob.format(self.rgb_bands[0])
```
adamjstewart: This won't work if the user doesn't download B04 for some reason. May be safer to grab the first band and the resolution of that band.

torchgeo/datasets/sentinel.py
```diff
@@ -360,14 +381,54 @@ def __init__(
...
+                folder_regex_str = folder_regex_str.format(
+                    match.groupdict()['resolution']
+                )
+                folder_regex = re.compile(folder_regex_str, re.VERBOSE)
+                folder_match = re.match(folder_regex, directory)
+                if folder_match:
+                    start = folder_match.start('resolution')
+                    end = folder_match.end('resolution')
+                    directory = (
+                        directory[:start] + self.resolutions[band] + directory[end:]
```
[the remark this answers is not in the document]
adamjstewart: I agree with this comment. Perhaps we can split by os.sep without relying on regexes at all?

torchgeo/datasets/sentinel.py
adamjstewart: This is also a bug in the previous dataset, if you choose a different res or CRS than the original, it could result in not finding matching files.

adamjstewart (approved):
I think this is clean and fast now. So glad this is finally fixed!

PR #3051, opened 2025-10-26
Mock downloading pre-trained weights

calebrob6:
Don't download resnet18 weights in the custom segmentation trainer notebook in case that is causing a timeout

### DIFF
--- a/docs/tutorials/custom_segmentation_trainer.ipynb
+++ b/docs/tutorials/custom_segmentation_trainer.ipynb
@@ -202,7 +202,8 @@
       "batch_size": 1,
       "fast_dev_run": true,
       "max_epochs": 1,
-      "num_workers": 0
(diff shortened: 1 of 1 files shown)


adamjstewart:
I think you need to declare the variable in a previous cell and mock it in a later cell? See https://github.com/treebeardtech/nbmake?tab=readme-ov-file#mock-out-variables-to-simplify-testing

calebrob6:
That's not what we do for `fast_dev_run`, etc.

adamjstewart:
Maybe the mocking cell and the later cell where it gets used need to be different?

adamjstewart:
Yep, that's working. So just move it up to the same location as fast_dev_run is mocked. The mocked value is only set after the cell completes.

adamjstewart (approved):

PR #3060, opened 2025-10-29
Tutorials: use modern GDAL CLI syntax

### LINKED ISSUE #2852: Tutorials: modern GDAL usage
### Issue

Newer versions of GDAL have a new CLI that is much more intuitive. We should update our tutorials to use this:

- [ ] Introduction to Geospatial Data
- [ ] Custom Raster Datasets

### Fix

See https://gdal.org/en/stable/programs/index.html for the new syntax.

adamjstewart:
Closes #2852

Note that this syntax was first introduced in GDAL 3.11 (as of May) and does not work with older GDAL. The syntax is also not really final, so those may be good reasons to hold off on this PR. But I'll let other reviewers decide.

### DIFF
--- a/docs/tutorials/custom_raster_dataset.ipynb
+++ b/docs/tutorials/custom_raster_dataset.ipynb
@@ -38,7 +38,7 @@
     "* `GeoDataset`: for uncurated raw data with geospatial metadata\n",
     "* `NonGeoDataset`: for curated benchmark datasets that lack geospatial metadata\n",
     "\n",
-    "If you're not sure which type of dataset you need, a good rule of thumb is to run `gdal info` on one of the files. If `gdal info` returns information like the bounding box, resolution, and CRS of the file, then you should probably use `GeoDataset`."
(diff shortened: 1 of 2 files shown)


isaaccorley (dismissed):

calebrob6:
I'm not sure about this. I don't see any immediate benefit of changing to the new syntax, but do see a good chance that this will make things not work the first time for some users (that have a non-recent version of gdal).

adamjstewart:
Let's hold off then. I do want to push more people to adopt the new CLI syntax, but I agree that it's a little too new and unstable. This can be saved for the next release in ~6 months.

adamjstewart:
@calebrob6 how do you feel about merging this now that GDAL 3.12 is out? I'm inclined to either merge or close just to reduce the number of open PRs I have.

calebrob6:
This is more like a merge in a few years thing for me. Alternatively you could update to show how to do it both ways.

adamjstewart:
Okay, added both syntaxes.

calebrob6 (approved):

PR #3061, opened 2025-10-29
DataModules: use drop_last flag for train data loader

### LINKED ISSUE #3039: Models can fail if used with combination of NonGeoDatamodule, AdaptiveAvgPool2d and BatchNorm
### Description

When a batch has a single element, it raises an error when passed through models with AdaptiveAvgPool2d(1) followed by BatchNorm (e.g., **UPerNet**), since there is only a single value per channel:
`Expected more than 1 value per channel when training, got input size torch.Size([1, 256, 1, 1])`

This happens if the data in NonGeoDatamodule cannot be split into batches with no single-element batches at the end (e.g. data of length 5 and bs=2) or if batchsize is 1.

This is not a strict issue with NonGeoDatamodule, but the current design has no nice way of avoiding this if such an issue occurs. 
One possibility is that we find a number that divides the data nicely, but an easier option is usually setting the `drop_last` property of dataloader to True. This second option is not easily doable due to the code design:
https://github.com/torchgeo/torchgeo/blob/9bf53cec1809cc0f6d6a56d093fdc9b41a6e9093/torchgeo/datamodules/geo.py#L425-L434

Would it make sense to add a `drop_last` argument to the NonGeo class so it can be set through kwargs and parsed like the current `dataset` and `batch_size` inside the factory?

### Steps to reproduce

```python
import segmentation_models_pytorch as smp
import torch

from torchgeo.datamodules import NonGeoDataModule
from torchgeo.datasets import NonGeoDataset


class MockData(NonGeoDataset):
    def __init__(self, split):
        super().__init__()

    def __getitem__(self, index):
        return torch.rand(3, 256, 256)

    def __len__(self):
        return 3

class MockDataModule(NonGeoDataModule):
    def __init__(self):
        super().__init__(MockData, batch_size=2)

model = smp.UPerNet()
data = MockDataModule()
data.setup("fit")

for batch in data.train_dataloader():
    model(batch)
```
raises error
`Expected more than 1 value per channel when training, got input size torch.Size([1, 256, 1, 1])`.

### Version

0.8.0.dev0

blaz-r:
This PR introduces a drop_last flag in NonGeo datamodule class. It enables drop_last, but only when the training dataloader is used. I also added tests to cover this.
Is this fine, or should I clarify in the docstring / variable name itself that it applies only to the trainloader?

This PR fixed #3039.

### DIFF
--- a/torchgeo/datamodules/geo.py
+++ b/torchgeo/datamodules/geo.py
@@ -287,7 +287,7 @@ class GeoDataModule(BaseDataModule):
             num_workers=self.num_workers,
             collate_fn=self.collate_fn,
             # drop_last is incompatible with batch sampler
-            drop_last=split == 'train' and batch_sampler is None,  
+            drop_last=split == 'train' and batch_sampler is None,
             persistent_workers=self.num_workers > 0,
         )
 


torchgeo/datamodules/geo.py
```diff
@@ -431,6 +434,7 @@ def _dataloader_factory(self, split: str) -> DataLoader[dict[str, Tensor]]:
             num_workers=self.num_workers,
             collate_fn=self.collate_fn,
             persistent_workers=self.num_workers > 0,
+            drop_last=split == 'train' and self.drop_last,
```
adamjstewart:
```suggestion
            drop_last=split == 'train',
```
If you make this change, we don't even need a parameter. I think we also need to do this in GeoDataModule.

adamjstewart:
Just to clarify, have you confirmed that this bug only occurs during `model.train()` and not `model.eval()`?

torchgeo/datamodules/geo.py
isaaccorley: Agree on this, preferably we just default to `drop_last=True` everywhere and if a user wants to override it they can.

blaz-r: I think that having this parameter could still be useful, but I agree that the difference wont be large even if its on by default.
I am not sure about the geodatamodule, aren't those designed in a way that always creates a "full size" batch? I can add that there as well though.

isaaccorley:
> Just to clarify, have you confirmed that this bug only occurs during `model.train()` and not `model.eval()`?

This bug is only due to model.train() requiring batch_size > 1. Model.eval() doesn't have this problem.

blaz-r:
I believe its not an issue in eval since this only applies to stats calculation, let me verify once I get home and get back to you in two hours or so.

torchgeo/datamodules/geo.py
isaaccorley: Let's add the parameter later if users start requesting it. The reasoning is that adding a parameter requires a minor release vs just changing the default could be considered a bug fix which we could get done in our patch release in the next day or so.

adamjstewart: > I am not sure about the geodatamodule, aren't those designed in a way that always creates a "full size" batch? I can add that there as well though.

I don't _think_ so. RandomBatchGeoSampler may be okay, but RandomGeoSampler should still suffer from the same bug.

adamjstewart: Although we only use RandomBatchGeoSampler. Hmm... You may be right.

blaz-r: I ran the tests and I get this:
`ValueError: batch_sampler option is mutually exclusive with batch_size, shuffle, sampler, and drop_last` 
So that'll be a problem.

blaz-r:
Thanks for chiming in, Isaac. To confirm 100%, I also checked with code, and it is not a problem in eval.

torchgeo/datamodules/geo.py
blaz-r: I think this could then be: 
```python
        if isinstance(sampler, BatchGeoSampler):
            batch_size = 1
            batch_sampler = sampler
            sampler = None
        else:
            batch_sampler = None

        return DataLoader(
            dataset=dataset,
            batch_size=batch_size,
            sampler=sampler,
            batch_sampler=batch_sampler,
            num_workers=self.num_workers,
            collate_fn=self.collate_fn,
            persistent_workers=self.num_workers > 0,
            drop_last=split == "train" and batch_sampler is None,
        ),
```

torchgeo/datamodules/geo.py
```diff
@@ -287,6 +287,9 @@ def _dataloader_factory(self, split: str) -> DataLoader[dict[str, Tensor]]:
             num_workers=self.num_workers,
             collate_fn=self.collate_fn,
             persistent_workers=self.num_workers > 0,
+            drop_last=(
+                split == 'train' and batch_sampler is None
+            ),  # drop_last is incompatible with batch sampler
```
adamjstewart: Can you put the comment on a separate (previous) line so it doesn't get wrapped?

adamjstewart:
```suggestion
            # drop_last is incompatible with batch sampler
            drop_last=split == 'train' and batch_sampler is None,  
```

torchgeo/datamodules/geo.py
```diff
@@ -287,6 +287,8 @@ def _dataloader_factory(self, split: str) -> DataLoader[dict[str, Tensor]]:
             num_workers=self.num_workers,
             collate_fn=self.collate_fn,
             persistent_workers=self.num_workers > 0,
+            # drop_last is incompatible with batch sampler
+            drop_last=split == 'train' and batch_sampler is None,
```
adamjstewart:
```suggestion
            # drop_last is incompatible with batch sampler
            drop_last=split == 'train' and batch_sampler is None,  
            persistent_workers=self.num_workers > 0,
```

torchgeo/datamodules/geo.py
```diff
@@ -431,6 +433,7 @@ def _dataloader_factory(self, split: str) -> DataLoader[dict[str, Tensor]]:
             num_workers=self.num_workers,
             collate_fn=self.collate_fn,
             persistent_workers=self.num_workers > 0,
+            drop_last=split == 'train',
```
adamjstewart, line 435:
```suggestion
            drop_last=split == 'train',
            persistent_workers=self.num_workers > 0,
```

adamjstewart (dismissed):

torchgeo/datamodules/geo.py
```diff
@@ -286,6 +286,8 @@ def _dataloader_factory(self, split: str) -> DataLoader[dict[str, Tensor]]:
             batch_sampler=batch_sampler,
             num_workers=self.num_workers,
             collate_fn=self.collate_fn,
+            # drop_last is incompatible with batch sampler
+            drop_last=split == 'train' and batch_sampler is None,
```
adamjstewart:
```suggestion
            drop_last=split == 'train' and batch_sampler is None,
```

adamjstewart:
Hmm, merging this may be causing the following tests to fail for 0.7 (but not for main): https://github.com/torchgeo/torchgeo/actions/runs/18918919530/job/54009562025

Any ideas?

adamjstewart:
Oh, I think I see what's going on. We just need to adjust the batch size in the tests to get at least one batch. These tests were removed in main, so that's why they didn't fail here.

blaz-r:
Hmm, Ill check this in a bit.

adamjstewart:
Fixed now, no worries

adamjstewart:
Just noticed new warnings in the [logs](https://github.com/torchgeo/torchgeo/actions/runs/18918137185/job/54006911719):
```
=============================== warnings summary ===============================
tests/datasets/test_utils.py::test_pad_across_batches
  /home/runner/work/torchgeo/torchgeo/torchgeo/datasets/utils.py:461: UserWarning: Truncated 2 sequences to length 1.
    warnings.warn(f'Truncated {truncated} sequences to length {padding_length}.')

tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-levircdplus]
tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-oscd]
tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-oscd_multiclass]
tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-oscd_multiclass_focal]
tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-oscd_multiclass_jaccard]
tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-xview2]
tests/trainers/test_iobench.py::TestClassificationTask::test_trainer[True-iobench]
  /opt/hostedtoolcache/Python/3.13.7/x64/lib/python3.13/site-packages/lightning/pytorch/utilities/data.py:106: Total length of `DataLoader` across ranks is zero. Please make sure this was your intention.

tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-levircdplus]
tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-oscd]
tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-oscd_multiclass]
tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-oscd_multiclass_focal]
tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-oscd_multiclass_jaccard]
tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-xview2]
tests/trainers/test_iobench.py::TestClassificationTask::test_trainer[True-iobench]
  /opt/hostedtoolcache/Python/3.13.7/x64/lib/python3.13/site-packages/lightning/pytorch/utilities/data.py:106: Total length of `CombinedLoader` across ranks is zero. Please make sure this was your intention.
```
Can you submit a PR to fix these? I think we just need to change `batch_size: 2` to `batch_size: 1` in `tests/conf/oscd.yaml` and friends.

blaz-r:
Okay. Are we still trying to get the release out today, or can I do this first thing in the morning (so in about 10h?)

adamjstewart:
Still trying to get the release out today, but this doesn't necessarily need to be in 0.7.2, it could be in 0.7.3 (or more likely 0.8.0). It doesn't affect users, it's just a warning in our tests.

isaaccorley:
> Just noticed new warnings in the [logs](https://github.com/torchgeo/torchgeo/actions/runs/18918137185/job/54006911719):
> 
> ```
> 
> =============================== warnings summary ===============================
> 
> tests/datasets/test_utils.py::test_pad_across_batches
> 
>   /home/runner/work/torchgeo/torchgeo/torchgeo/datasets/utils.py:461: UserWarning: Truncated 2 sequences to length 1.
> 
>     warnings.warn(f'Truncated {truncated} sequences to length {padding_length}.')
> 
> 
> 
> tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-levircdplus]
> 
> tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-oscd]
> 
> tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-oscd_multiclass]
> 
> tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-oscd_multiclass_focal]
> 
> tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-oscd_multiclass_jaccard]
> 
> tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-xview2]
> 
> tests/trainers/test_iobench.py::TestClassificationTask::test_trainer[True-iobench]
> 
>   /opt/hostedtoolcache/Python/3.13.7/x64/lib/python3.13/site-packages/lightning/pytorch/utilities/data.py:106: Total length of `DataLoader` across ranks is zero. Please make sure this was your intention.
> 
> 
> 
> tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-levircdplus]
> 
> tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-oscd]
> 
> tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-oscd_multiclass]
> 
> tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-oscd_multiclass_focal]
> 
> tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-oscd_multiclass_jaccard]
> 
> tests/trainers/test_change.py::TestChangeDetectionTask::test_trainer[True-xview2]
> 
> tests/trainers/test_iobench.py::TestClassificationTask::test_trainer[True-iobench]
> 
>   /opt/hostedtoolcache/Python/3.13.7/x64/lib/python3.13/site-packages/lightning/pytorch/utilities/data.py:106: Total length of `CombinedLoader` across ranks is zero. Please make sure this was your intention.
> 
> ```
> 
> Can you submit a PR to fix these? I think we just need to change `batch_size: 2` to `batch_size: 1` in `tests/conf/oscd.yaml` and friends.

@copilot can you make a PR to fix these warnings in the tests?

adamjstewart:
Copilot is for losers 😛

PR #3064, opened 2025-10-30
Investigate slow tests

adamjstewart:
macOS tests are 3x slower than Linux, and some trainer tests seem to be duplicated. Let's see if we can speed things up.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -43,7 +43,7 @@ jobs:
         run: pip list
       - name: Run pytest checks
         run: |
-          pytest --cov=torchgeo --cov-report=xml
+          pytest --cov=torchgeo --cov-report=xml --durations=10
           python3 -m torchgeo --help
       - name: Report coverage
         uses: codecov/codecov-action@v5.5.1
@@ -75,7 +75,7 @@ jobs:
         run: pip list
       - name: Run pytest checks
         run: |
-          pytest --cov=torchgeo --cov-report=xml
+          pytest --cov=torchgeo --cov-report=xml --durations=10
           python3 -m torchgeo --help
       - name: Report coverage
         uses: codecov/codecov-action@v5.5.1
@@ -107,7 +107,7 @@ jobs:
         run: pip list
       - name: Run pytest checks
         run: |
-          pytest --cov=torchgeo --cov-report=xml
+          pytest --cov=torchgeo --cov-report=xml --durations=10
           python3 -m torchgeo --help
       - name: Report coverage
         uses: codecov/codecov-action@v5.5.1


blaz-r:
Could it be related to the fact that those are arm instances (macos-15-arm64)?

adamjstewart:
I don't think so? Windows is 2x slower and still x86_64. It's across all python versions, so not a multiprocessing change.

adamjstewart:
> some trainer tests seem to be duplicated

Mystery semi-solved in #3066. The tests weren't duplicated, just run in a strange order.

Still trying to figure out why things are so much slower on macOS and Windows...

adamjstewart:
Oddly enough, the tests run in ~8 min on my macOS laptop. So this may prove difficult to reproduce.

blaz-r:
Could be linked to specifics of the VM image. The slowest test seems to be to be linked to weight loading, so maybe some I/O stuff is behaving in unusual way. From https://github.com/actions/runner-images/issues/11509 it seems to also be some I/O issues on specific image versions.

adamjstewart:
The macOS arm64 runner does have half the memory of the other runners: https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

adamjstewart:
Okay, to summarize, using pytest, we found:
```console
> pytest --durations=10
...
123.98s call     tests/models/test_api.py::test_get_model[aurora_swin_unet]
 95.45s call     tests/models/test_aurora.py::TestAurora::test_aurora_swin_unet
 74.26s call     tests/models/test_vit.py::TestViTHuge14::test_vit_weights[True]
 61.39s call     tests/models/test_vit.py::TestViTHuge14::test_vit_weights[False]
 49.72s call     tests/models/test_croma.py::TestCROMALarge::test_croma_weights
 43.62s setup    tests/models/test_vit.py::TestViTHuge14::test_vit_weights[False]
 43.41s setup    tests/models/test_vit.py::TestViTHuge14::test_vit_weights[True]
 42.36s call     tests/models/test_api.py::test_get_model[vit_huge_patch14_224]
 35.50s call     tests/models/test_vit.py::TestViTHuge14::test_vit
 21.86s call     tests/models/test_scale_mae.py::TestScaleMAE::test_scalemae_weights_diff_image_size
```
I also ran some additional experiments using pytest-memray:
```console
> pytest --memray --most-allocations=10
...
5.0GiB  tests/models/test_croma.py::TestCROMALarge::test_croma_weights
4.7GiB  tests/models/test_vit.py::TestViTHuge14::test_vit_weights[False]
4.7GiB  tests/models/test_vit.py::TestViTHuge14::test_vit_weights[True]
4.7GiB  tests/models/test_api.py::test_get_model[aurora_swin_unet]
4.7GiB  tests/models/test_aurora.py::TestAurora::test_aurora_swin_unet
3.1GiB  tests/models/test_dofa.py::TestDOFAHuge14::test_dofa
2.5GiB  tests/models/test_dofa.py::TestDOFALarge16::test_dofa_weights
2.5GiB  tests/models/test_api.py::test_get_model[croma_large]
2.5GiB  tests/models/test_croma.py::TestCROMALarge::test_croma
2.5GiB  tests/models/test_api.py::test_get_model[dofa_huge_patch14_224]
```
So it seems like there is a strong correlation between slow tests and memory footprint. Given that individual tests are hitting 5 GB each and there is only 7 GB of RAM on the `macos-latest` runners, I suspect we are RAM limited, not storage limited.

Next step would be to try some of the [larger runners](https://docs.github.com/en/actions/reference/runners/larger-runners) in GitHub Actions. Those require GitHub Team or GitHub Enterprise plans. I'll discuss this with the TSC.

We could also consider adding some of these tests to `slow` only on macOS. This would limit our ability to run these tests locally though.

calebrob6:
It seems as we add more model families (which we definitely want to do) this will become a larger problem. Any idea how we can decouple test length from number of models? (maybe test a smaller version of the model?)

adamjstewart:
We currently only test one weight per model, but we need to test all model sizes to get code coverage. Best idea I can think of would be to merge `resnet18`, `resnet50`, etc. into a single `resnet` method so we only have to test `resnet18` to get coverage.

blaz-r:
Are these test slow due to I/O of saving and loading weights or model execution? A possibility to somewhat address the second option is to only run the fastdev run on regular commits, but have a separate "pre-merge" pipeline that also runs the long tests that is manually triggered for PR before merge.

adamjstewart:
> Are these test slow due to I/O of saving and loading weights or model execution? 

On macOS, I think it's due to running out of RAM.

> A possibility to somewhat address the second option is to only run the fastdev run on regular commits, but have a separate "pre-merge" pipeline that also runs the long tests that is manually triggered for PR before merge.

We already do exactly this, fastdev on main and slow on release branches.

adamjstewart:
TODO: try this script: https://gist.github.com/haampie/d8ab09a1b4a7c067c784e38cd05c0dc2

> `python3 script.py fetch` and then `python3 script.py plot`, you need to update the URL and the name of the action in the file

This should help us visualize test time over time.

adamjstewart:
Also TODO: try to parallelize pytest and sphinx. May make the memory problem worse, but local testing faster.

adamjstewart:
Closing this for now, we can revisit this if/when things get worse. We may require a fundamental redesign of our models to avoid a linear increase in test time as the number of models we support increases.

PR #3070, opened 2025-11-03
update pre-commit-config.yaml

isaaccorley:


.pre-commit-config.yaml
```diff
@@ -22,18 +22,19 @@ repos:
...
           - pillow>=10.4.0
-          - pytest>=6.1.2
+          - pytest>=7.3
           - scikit-image>=0.22.0
           - timm>=1.0.16
           - torch>=2.8
           - torchmetrics>=1.6.2
           - torchvision>=0.18
-        exclude: (build|data|dist|logo|logs|output)/
+        exclude: (build|data|dist|logo|logs|output|docs/user/metrics)/
```
adamjstewart: Any reason why docs/user/metrics needs to be excluded? I think it's passing.

isaaccorley: It wasn't passing mypy for me locally. I can paste the errors.

adamjstewart: Please do. It's passing in CI. Maybe the min version of mypy in pre-commit needs to be bumped?

isaaccorley: Mainly the errors were around not having request mypy stubs. But maybe it's just a local thing

adamjstewart: Ah, that's something we set in `pyproject.toml`:
```
ignore_missing_imports = true
```
but pre-commit refuses to support `pyproject.toml` so we have to duplicate every single flag. Basically, there's a flag for this. Technically we should also add `--strict`, may require a couple overrides though.

ashnair1: So pre-commit creates a separate environment to run the checks. Hence why the additional dependencies are needed. These deps are installed in that environment and the checks are run.

So the `request` errors should be resolved by adding `types-requests` to mypy's `additional_dependencies`.

.pre-commit-config.yaml
```diff
@@ -22,18 +22,19 @@ repos:
...
           - einops>=0.6.0
-          - kornia>=0.6.9
+          - kornia>=0.7.4
           - lightning>=2.0.9
           - matplotlib>=3.9.2
-          - numpy>=1.22
+          - numpy>=1.24
           - pillow>=10.4.0
-          - pytest>=6.1.2
+          - pytest>=7.3
```
ashnair1, line 29: No need to bump versions here as they haven't added new typing support in these versions.

ashnair1 (changes requested):
I was able to get pre-commit passing by adding `types-requests` to additional_dependencies and fixing (and ignoring) some typing errors in `update_github.py` and `update_downloads.py`

.pre-commit-config.yaml
adamjstewart: We should probably add a bunch of stub packages as testing deps. I wish these were within the requests source code, but at least someone wrote a stub.

.pre-commit-config.yaml
adamjstewart: These are all bumped to our minimum supported version

adamjstewart:
I would like to keep pre-commit matching direct invocation of mypy. So we either ignore missing imports in pre-commit like we do in mypy, or we add type stubs to both.

calebrob6:
This is a minor thing that we should just merge

adamjstewart:
Not user facing so doesn't matter which release it goes into. Would like to sort out the type stubs stuff separately from updating dependencies here.

adamjstewart:
Coming back to this. The version changes are uncontroversial and could be quickly merged. However, we just dropped support for Python 3.11, so they may need to be raised again.

The doc ignores are the only thing holding up the PR. I would like to try to strive to use the same flags we use in `pyproject.toml` whenever possible. In this case, we now use an explicit `packages = ["torchgeo", "tests"]` and `exclude = ["^tests/data"]` and `exclude_gitignore = true`. If we can use those in pre-commit as well I would be happier.

Splitting into separate PRs would probably be easiest.

.pre-commit-config.yaml
isaaccorley: Rebased and added the stub packages which seemed to fix the issues

ashnair1 (approved):

isaaccorley:
> Coming back to this. The version changes are uncontroversial and could be quickly merged. However, we just dropped support for Python 3.11, so they may need to be raised again.
> 
> The doc ignores are the only thing holding up the PR. I would like to try to strive to use the same flags we use in `pyproject.toml` whenever possible. In this case, we now use an explicit `packages = ["torchgeo", "tests"]` and `exclude = ["^tests/data"]` and `exclude_gitignore = true`. If we can use those in pre-commit as well I would be happier.
> 
> Splitting into separate PRs would probably be easiest.

I solved this adding `pass_filenames: false` and explicitly setting the config file to use `pyproject.toml` which effectively runs `mypy --config-file pyproject.toml`.

Issue #3026, opened 2025-10-08
Add torchange to Related Libraries


adamjstewart:
### Issue

We should add https://github.com/Z-Zheng/pytorch-change-models to https://torchgeo.readthedocs.io/en/latest/user/alternatives.html

### Fix

https://torchgeo.readthedocs.io/en/latest/user/contributing.html#related-libraries

adamjstewart:
@Z-Zheng do you want to work on this?

Z-Zheng:
just adding docs into https://torchgeo.readthedocs.io/en/latest/user/alternatives.html?

adamjstewart:
Yep, see https://torchgeo.readthedocs.io/en/latest/user/contributing.html#related-libraries for instructions.

adamjstewart:
Also, FYI, we finally added [ChangeDetectionTask](https://torchgeo.readthedocs.io/en/latest/api/trainers.html#torchgeo.trainers.ChangeDetectionTask) to TorchGeo. It's mostly using the CD models you once added to TorchGeo, but we could definitely replace those with a dependency on `torchange` if it becomes stable and well-maintained.

Z-Zheng:
Good to know that. Let me add them later.

PR #3072, opened 2025-11-04
adding torchange to doc

Z-Zheng:
Closes https://github.com/torchgeo/torchgeo/issues/3026

### DIFF
--- a/docs/user/metrics/common.py
+++ b/docs/user/metrics/common.py
@@ -122,5 +122,5 @@ name_to_conda = {
     'samgeo': 'segment-geospatial',
     'GeoAI': 'geoai',
     'TerraTorch': 'terratorch',
-    'torchange': 'torchange'
+    'torchange': 'torchange',
 }


adamjstewart:
I think you also need to commit the changes to `common.py`, otherwise this will be removed during the next update.

docs/user/metrics/common.py
adamjstewart, line 1: Should probably also update `hardcoded_coverage` to 0 until unit tests are added.

Z-Zheng: added.

docs/user/metrics/features.csv
```diff
@@ -18,3 +18,4 @@ Library,ML Backend,I/O Backend,Spatial Backend,Transform Backend,Datasets,Weight
 `GeoTessera`_,scikit-learn,GDAL,-,geopandas,0,0,✅,❌,✅,❌,❌
 `OTBTF`_,TensorFlow,"GDAL, OTB",-,-,0,0,✅,❌,✅,❌,❌
 `GeoDeep`_,ONNX,GDAL,-,-,0,7,✅,❌,✅,❌,❌
+`torchange`_,PyTorch,"Apache Arrow, skimage",geopandas,Albumentations,8,8,❌,❌,❌,❌,✅
```
adamjstewart: * spatial backend: I don't see geopandas in the codebase. Perhaps there is no spatial backend and geographic metadata isn't used?
* weights: Where do I find your pretrained model weights?
* time series: Do you support variable length sequences of SITS data or only binary temporal pairs?

Z-Zheng: - spatial backend: Yeah, current version does not need a spatial backend. I will use geopandas if i will handle geo-metadata.
- weights: can be found https://huggingface.co/EVER-Z/Changen2-ChangeStar1x256, https://huggingface.co/EVER-Z/Changen2_RSDiT-B-2-OEM-256, and https://huggingface.co/EVER-Z/Changen2_RSDiT-B-2-xView2-256
- time series: current implementation only focuses on bitemporal

adamjstewart: Great, in that case let's use:
```suggestion
`torchange`_,PyTorch,"Apache Arrow, skimage",-,Albumentations,8,8,❌,❌,❌,❌,✅
```

docs/user/metrics/common.py
```diff
@@ -106,6 +109,7 @@
     'TerraTorch': 'terratorch',
     'GeoDeep': 'geodeep',
     'GeoTessera': 'geotessera',
+    'torchange': 'torchange'
```
adamjstewart:
```suggestion
    'torchange': 'torchange',
```

docs/user/metrics/common.py
```diff
@@ -118,4 +122,5 @@
     'samgeo': 'segment-geospatial',
     'GeoAI': 'geoai',
     'TerraTorch': 'terratorch',
+    'torchange': 'torchange'
```
adamjstewart:
```suggestion
    'torchange': 'torchange',
```

PR #3074, opened 2025-11-05
Test type stubs

adamjstewart:
@isaaccorley this may help with #3070

I'm hesitant to add these permanently as some of these aren't direct dependencies (only used in experimental scripts) and they aren't necessarily maintained by the authors of the library. However, we can at least fix known type issues in case someone has one of these installed anyway.


adamjstewart:
A lot of the pandas ones are difficult if not impossible to fix due to dynamic typing.

adamjstewart:
Going to integrate these and resolve issues one package at a time.

PR #3075, opened 2025-11-05
CI: speed up macOS tests

adamjstewart:
Trying to speed up macOS tests, see #3064


adamjstewart:
We drop to ~90% coverage of `torchgeo.models` on macOS, but retain coverage everywhere else. macOS test speeds are now comparable to Linux/Windows.

adamjstewart:
Hoping to avoid this if we can.

PR #3079, opened 2025-11-06
Add time-series support for SustainBench Crop Yield dataset

Thakor-Yashpal:


adamjstewart:
A bit confused by this PR. It deletes SustainBench Crop Yield and instead makes changes to ZueriCrops. Is that the intention?

adamjstewart:
@Thakor-Yashpal is this PR still a WIP or can we close?

PR #3084, opened 2025-11-09
CI: test free-threaded Python

adamjstewart:
This PR tests support for Python's new free-threaded mode in which the global interpreter lock is disabled.

We are currently waiting on some of our deps to release wheels for this mode:

- [x] aiohttp: has 3.14t but not 3.13t, https://github.com/aio-libs/aiohttp/issues/8796
- [x] ~fiona~: https://github.com/Toblerity/Fiona/issues/1507
- [x] kiwisolver: has 3.13t for Windows arm but not x86_64
- [x] opencv-python: https://github.com/opencv/opencv-python/issues/1029
- [ ] pandas: https://github.com/pandas-dev/pandas/issues/62637
- [x] pycocotools: has 3.13t but not 3.14/3.14t: https://github.com/ppwwyyxx/cocoapi/pull/31
- [x] pydantic-core: has 3.14t but not 3.13t
- [x] pyyaml: has 3.14t but not 3.13t, https://github.com/yaml/pyyaml/issues/870
- [x] netcdf4: https://github.com/Unidata/netcdf4-python/issues/1454
- [x] rasterio: https://github.com/rasterio/rasterio/issues/3428
- [ ] safetensors: https://github.com/huggingface/safetensors/issues/572
- [ ] scikit-image: https://github.com/scikit-image/scikit-image/issues/7464
- [ ] tornado: https://github.com/tornadoweb/tornado/issues/3501

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -365,3 +365,13 @@ torchgeo = ["py.typed"]
 
 [tool.setuptools.packages.find]
 include = ["torchgeo*"]
+
+[tool.ty.rules]
+deprecated = "ignore"
+unresolved-attribute = "ignore"
+unresolved-import = "ignore"
+unresolved-reference = "ignore"
+
+[tool.ty.src]
+include = ["torchgeo", "tests"]
+exclude = ["tests/data"]
--- a/requirements/datasets.txt
+++ b/requirements/datasets.txt
@@ -6,7 +6,7 @@ netcdf4==1.7.4
 opencv-python==4.12.0.88; python_version<'3.14'
 pandas[parquet]==2.3.3
 pycocotools==2.0.11
-rioxarray==0.19.0
+rioxarray==0.20.0
 scikit-image==0.26.0
 scipy==1.16.3
 webdataset==1.0.2
--- a/requirements/required.txt
+++ b/requirements/required.txt
@@ -11,7 +11,7 @@ numpy==2.2.4; python_version<'3.14'
 pandas==2.3.3
 pillow==12.1.0
 pyproj==3.7.2
-rasterio==1.4.4
+rasterio==1.5.0
 requests==2.32.5
 segmentation-models-pytorch==0.5.0
 shapely==2.1.2
--- a/tests/models/test_rcf.py
+++ b/tests/models/test_rcf.py
@@ -25,11 +25,11 @@ class TestRCF:
         model = RCF(in_channels=5, features=4, kernel_size=3, mode='gaussian')
         x = torch.randn(2, 5, 64, 64)
         y = model(x)
-        assert y.shape[1] == 4
+        assert y.shape == (2, 4)
 
         x = torch.randn(1, 5, 64, 64)
         y = model(x)
-        assert y.shape[0] == 4
+        assert y.shape == (1, 4)
 
     def test_untrainable(self) -> None:
         model = RCF(in_channels=5, features=4, kernel_size=3, mode='gaussian')
--- a/torchgeo/models/rcf.py
+++ b/torchgeo/models/rcf.py
@@ -171,6 +171,7 @@ class RCF(Module):
 
         return patches_normalized.reshape(orig_shape).astype('float32')
 
+    @torch.inference_mode()
     def forward(self, x: Tensor) -> Tensor:
         """Forward pass of the RCF model.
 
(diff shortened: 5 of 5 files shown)


adamjstewart:
Unfortunately there is no environment marker for free-threaded until [PEP 780](https://peps.python.org/pep-0780/) is approved. Best we can do is `python_version<'3.13'`.

adamjstewart:
Well we can now build wheels for all missing deps, but one of them isn't happy running in free-threaded mode. I think we should revisit in 6 months and try again.

rgommers:
I wouldn't bother with 3.13t anymore, it's more or less obsolete since (a) 3.14t is much more robust and no longer marked experimental, and (b) key packages like CFFI, aiohttp, etc. aren't going to release 3.13t wheels anymore.

Focusing on 3.14t only will make the list of needed deps a fair bit shorter.

adamjstewart:
Thanks for the suggestion @rgommers. It looks like the tests still fail even for 3.14t, but with no error message. Is there any official documentation on debugging these kind of free-threaded failures? I would at least like to be able to point to which of our many deps is causing the test failure.

rgommers:
There are multiple pages in https://py-free-threading.github.io about testing and debugging free-threading specific issues. After a quick glance at the CI results, I see multiple issues - but I think you're asking about the crash. I'd say running it locally and using `gdb`/`lldb` to obtain the traceback should tell you a lot here (see https://py-free-threading.github.io/debugging/#debugging-tests-that-depend-on-native-calls).

There are other issues like this one:
```
AttributeError: 'types.SimpleNamespace' object has no attribute 'objs'
```
which are probably easier to solve (or unrelated).

adamjstewart:
Didn't even notice those other error messages. Let me target those first, perhaps I'll accidentally fix the other failures.

adamjstewart:
Looks like geopandas does not yet support pandas 3: https://github.com/geopandas/geopandas/pull/3621

adamjstewart:
Quite possibly hitting https://github.com/pytorch/pytorch/issues/158071, may need to wait for 2.9.2.

rgommers:
Ah that would make sense. PyTorch 2.9.x isn't very usable for `torch.compile` on 3.14, so using a 2.10.0 nightly might be useful (or wait until 2.10.0 is out).

adamjstewart:
Well, the tests don't segfault locally with macOS python-3.14t and the same pip-installed environment. Could possibly be a memory issue, hard to tell without any error msg.

rgommers:
Two things that might be useful:
- add `-v` to the pytest invocation so you see exactly which test crashes
- add `--faulthandler` to `pytest.ini` or `-X faulthandler` to the `pytest` invocation to try and get a traceback rather than the silent exit

adamjstewart:
Looks like we have ourselves a Heisenbug.

rgommers:
> Looks like we have ourselves a Heisenbug.

Not completely surprising - that's what happens often with threading-related issues, since triggering the bug depends on timing. Hunting those down can be time-consuming. The simpler method is with `pytest-run-parallel`, which is easy to work with and tends to reproduce most issues when run a couple of times. For the harder ones, we end up needing ThreadSanitizer (see https://py-free-threading.github.io/thread_sanitizer/), and that catching (almost) everything but it's very very slow.

Since there is a known issue in PyTorch with a fix in the main branch but not the latest released version, I suspect that that's the cause. You might want to keep `-v`/`--faulthandler` and just wait for it to fail again to get some details, and slap a `skip` under free-threading on it if it's annoying.

adamjstewart:
Weird that it was reliably failing and is now reliably passing. Either a transient dep got updated or enabling faulthandler somehow prevented the bug from occurring.

I would be in favor of merging, but keeping 3.14t tests as non-required for merge until we have more guarantees of stability. We don't directly use `torch.compile`, but we do have a few `torch.export` tests that we could skip if needed.

rgommers:
Sounds like a reasonable strategy to me.

adamjstewart:
And now I'm seeing similar intermittent errors on different PRs unrelated to 3.14t. The plot thickens...

https://github.com/torchgeo/torchgeo/actions/runs/20370648552/job/58536156768?pr=3217

rgommers:
Okay so a general 3.14 issue? Also not wholly surprising, given that 3.14 support in PyTorch is explicitly labeled as a "preview".

adamjstewart:
Looks like `python3 -X faulthandler` didn't help here: https://github.com/torchgeo/torchgeo/actions/runs/20743773593/job/59556158194?pr=3084

Since it's failing on ViT Huge and appears to be killed externally, I have a feeling that we are running out of memory. We've seen this before on macOS where the CI runners have significantly less RAM than other OSes. What I'm surprised by is that the issue is happening on Linux (not macOS) and is restricted to Python 3.14. I don't see any known issues with Python 3.14 and PyTorch memory usage, so this is strange...

It looks like PyTorch 2.10 is already a WIP, so we may have to wait for that before Python 3.14 can work reliably: https://github.com/pytorch/pytorch/issues/170119

PR #3085, opened 2025-11-10
Add dice loss to change trainer

blaz-r:
This PR adds dice loss to ChangeDetection trainer. This loss is required by #2926.


adamjstewart (approved):
Any other trainers that could use dice loss?

blaz-r:
I think that segmentation could, but I'm not familiar with that task enough to be sure.

PR #3086, opened 2025-11-10
add CopernicusBenchBiomassS3 datamodule and SpatioTemporalPixelwiseRegression

### LINKED ISSUE #3272: Time series support in the regression tasks
### Summary

- Similar to that already implemented in https://github.com/torchgeo/torchgeo/pull/2921 with a test run with ConvLSTM
- This will be tested with dummy dataset representing a sample from CopernicusBenchBiomassS3

### Rationale

_No response_

### Implementation

_No response_

### Alternatives

_No response_

### Additional information

_No response_

robmarkcole:
Add datamodule for Copernicus-Bench Biomass-S3 dataset.

TO DISCUSS: are there official normalisation stats?

image

Closes #3272


adamjstewart:
> TO DISCUSS: are there official normalisation stats?

@wangyi111?

wangyi111:
No, but applied per-band scale factors as https://developers.google.com/earth-engine/datasets/catalog/COPERNICUS_S3_OLCI#bands.

robmarkcole:
@adamjstewart I will use these `scale` factors which I understand should be multiplied by each channel

robmarkcole:
Error implies inconsistent patch sizes:
```
 TRAIN
 Image typical: (21, 94, 94) (93.57%) from 3000 samples
  Image anomalies:
   - (21, 84, 94) (64 | 2.13%)
   - (21, 94, 84) (62 | 2.07%)
   - (21, 94, 95) (31 | 1.03%)
   - (21, 95, 94) (18 | 0.6%)
   - (21, 94, 85) (10 | 0.33%)
   - (21, 85, 94) (5 | 0.17%)
   - (21, 95, 95) (1 | 0.03%)
   - (21, 84, 95) (1 | 0.03%)
   - (21, 84, 84) (1 | 0.03%)
 Mask typical: (282, 282) (95.23%) from 3000 samples
  Mask anomalies:
   - (282, 252) (72 | 2.4%)
   - (252, 282) (70 | 2.33%)
   - (252, 252) (1 | 0.03%)

VAL
 Image typical: (21, 94, 94) (93.7%) from 1000 samples
  Image anomalies:
   - (21, 94, 84) (21 | 2.1%)
   - (21, 84, 94) (18 | 1.8%)
   - (21, 94, 95) (10 | 1.0%)
   - (21, 95, 94) (6 | 0.6%)
   - (21, 94, 85) (4 | 0.4%)
   - (21, 85, 94) (4 | 0.4%)
 Mask typical: (282, 282) (95.3%) from 1000 samples
  Mask anomalies:
   - (282, 252) (25 | 2.5%)
   - (252, 282) (22 | 2.2%)

TEST
 Image typical: (21, 94, 94) (94.2%) from 1000 samples
  Image anomalies:
   - (21, 94, 84) (19 | 1.9%)
   - (21, 84, 94) (14 | 1.4%)
   - (21, 94, 95) (14 | 1.4%)
   - (21, 94, 85) (5 | 0.5%)
   - (21, 95, 94) (3 | 0.3%)
   - (21, 85, 94) (2 | 0.2%)
   - (21, 84, 84) (1 | 0.1%)
 Mask typical: (282, 282) (95.9%) from 1000 samples
  Mask anomalies:
   - (282, 252) (24 | 2.4%)
   - (252, 282) (16 | 1.6%)
   - (252, 252) (1 | 0.1%)
```
@adamjstewart should we resize to the standard size?

adamjstewart:
I would recommend a couple things:

1. Change our `tests/data/copernicus/l3_biomass_s3` data to have inconsistent sizes to reproduce this issue
2. Use `Resize((94, 94))` as a dataset transform in the datamodule (not an aug)

robmarkcole:
@adamjstewart and resize the mask to 282? 
```
 Mask typical: (282, 282) (95.3%) from 1000 samples
```

adamjstewart:
Oh, I didn't realize the image and mask have different resolutions. That will be quite tricky. We should probably resize both to the same resolution so it becomes a simple pixel-wise regression task. @wangyi111 do you have a preference between making the image bigger or the mask smaller?

robmarkcole:
@adamjstewart for the test, I see there are some tifs, you want one of those resized to reproduce the error?

adamjstewart:
Correct. There should be a `data.py` used to generate those files. You can make the changes there so we can reproduce the test data.

wangyi111:
> Oh, I didn't realize the image and mask have different resolutions. That will be quite tricky. We should probably resize both to the same resolution so it becomes a simple pixel-wise regression task. @wangyi111 do you have a preference between making the image bigger or the mask smaller?

i'd like to keep:) but if it's must maybe make image bigger then

adamjstewart:
@wangyi111 I mean change it on-the-fly, not in your dataset. How do you evaluate your model if the input and output resolutions don't match? Are you training a J-net instead of a U-net?

robmarkcole:
@adamjstewart using 
```python
TARGET_SIZE = (282, 282)
...
        resize = K.Resize(size=TARGET_SIZE)
        extra_args = {
            DataKey.MASK: {'resample': Resample.NEAREST, 'align_corners': None}
        }

        if mode == 'time-series':
                K.VideoSequential(resize, normalizer),
                data_keys=None,
                keepdim=True,
                same_on_batch=True,
                extra_args=extra_args,
            )
        else:
                resize, normalizer, data_keys=None, keepdim=True, extra_args=extra_args
            )
```
I still get `RuntimeError: Trying to resize storage that is not resizable` - is this the correct resize approach?

**Update**: If I swap out `AugmentationSequential` the error is cleared:
```python

"""Copernicus-Bench Biomass-S3 datamodule."""

from typing import Any

import torch
from kornia.constants import Resample

from ...datasets import CopernicusBenchBiomassS3
from ..geo import NonGeoDataModule

# Multiplicative scale factors from
# https://developers.google.com/earth-engine/datasets/catalog/COPERNICUS_S3_OLCI#bands
SCALE = {
    'Oa01_radiance': 0.0139465,
    'Oa02_radiance': 0.0133873,
    'Oa03_radiance': 0.0121481,
    'Oa04_radiance': 0.0115198,
    'Oa05_radiance': 0.0100953,
    'Oa06_radiance': 0.0123538,
    'Oa07_radiance': 0.00879161,
    'Oa08_radiance': 0.00876539,
    'Oa09_radiance': 0.0095103,
    'Oa10_radiance': 0.00773378,
    'Oa11_radiance': 0.00675523,
    'Oa12_radiance': 0.0071996,
    'Oa13_radiance': 0.00749684,
    'Oa14_radiance': 0.0086512,
    'Oa15_radiance': 0.00526779,
    'Oa16_radiance': 0.00530267,
    'Oa17_radiance': 0.00493004,
    'Oa18_radiance': 0.00549962,
    'Oa19_radiance': 0.00502847,
    'Oa20_radiance': 0.00326378,
    'Oa21_radiance': 0.00324118,
}

TARGET_SIZE = (282, 282)


class CopernicusBenchBiomassS3DataModule(NonGeoDataModule):
    """LightningDataModule implementation for the Copernicus Biomass-S3 dataset.

    Uses the train/val/test splits provided with the benchmark.

    .. versionadded:: 0.81
    """

    def __init__(
        self, batch_size: int = 64, num_workers: int = 0, **kwargs: Any
    ) -> None:
        """Initialize a new CopernicusBenchBiomassS3DataModule instance.

        Args:
            batch_size: Size of each mini-batch.
                :class:`~torchgeo.datasets.CopernicusBenchBiomassS3`.
        """
        bands = kwargs.get('bands', SCALE.keys())

        scale_factors = torch.tensor([SCALE[b] for b in bands], dtype=torch.float32)

        self.mean = torch.zeros(len(bands), dtype=torch.float32)
        self.std = torch.reciprocal(scale_factors)

        super().__init__(CopernicusBenchBiomassS3, batch_size, num_workers, **kwargs)

        self.image_resizer = K.Resize(
            size=TARGET_SIZE,
            resample=Resample.BILINEAR.name,
            align_corners=False,
        )
        self.mask_resizer = K.Resize(
            size=TARGET_SIZE,
            resample=Resample.NEAREST.name,
            align_corners=False,
        )
        self.normalizer = K.Normalize(mean=self.mean, std=self.std)

        def _resize_and_normalize(batch: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
            batch['image'] = self._resize_image(batch['image'])
            if 'mask' in batch:
                batch['mask'] = self._resize_mask(batch['mask'])
            return batch

        self.aug = _resize_and_normalize

    def _resize_image(self, image: torch.Tensor) -> torch.Tensor:
        device = image.device
        self.image_resizer = self.image_resizer.to(device)
        self.normalizer = self.normalizer.to(device)

        if image.ndim == 4:
            resized = self.image_resizer(image)
            return self.normalizer(resized)

        if image.ndim == 5:
            batch, time, channels, height, width = image.shape
            flattened = image.view(-1, channels, height, width)
            resized = self.image_resizer(flattened)
            normalized = self.normalizer(resized)
            return normalized.view(batch, time, channels, *TARGET_SIZE)

        msg = (
            'Expected image tensor with shape (batch, channels, height, width) or '
            '(batch, time, channels, height, width).'
        )
        raise ValueError(msg)

    def _resize_mask(self, mask: torch.Tensor) -> torch.Tensor:
        device = mask.device
        original_dtype = mask.dtype
        self.mask_resizer = self.mask_resizer.to(device)

        if mask.ndim == 3:
            resized = self.mask_resizer(mask.unsqueeze(1).float())
            return resized.squeeze(1).to(original_dtype)

        if mask.ndim == 4:
            batch, time, height, width = mask.shape
            flattened = mask.view(-1, 1, height, width).float()
            resized = self.mask_resizer(flattened)
            return resized.view(batch, time, *TARGET_SIZE).to(original_dtype)

        msg = (
            'Expected mask tensor with shape (batch, height, width) or '
            '(batch, time, height, width).'
        )
        raise ValueError(msg)
```

The error might be a red herring: https://github.com/ultralytics/ultralytics/issues/5319

adamjstewart:
I would use the transforms parameter in https://torchgeo.readthedocs.io/en/stable/api/datasets.html#torchgeo.datasets.CopernicusBenchBiomassS3.__init__

You need to do the resize on each sample before the mini-batch is collated. Collation will fail due to size mismatch.

wangyi111:
> @wangyi111 I mean change it on-the-fly, not in your dataset. How do you evaluate your model if the input and output resolutions don't match? Are you training a J-net instead of a U-net?

i see. when training i was upsampling image to match mask

robmarkcole:
I've iterated though a few images and they just don't appear to match the mask:

```python
datamodule = CopernicusBenchBiomassS3DataModule(
    root="/data/",
    batch_size=4,
    num_workers=4,
)

datamodule.setup(stage="fit")
train_loader = datamodule.train_dataloader()

batch = next(iter(train_loader))
batch = datamodule.aug(batch)

assert batch['image'].shape == torch.Size([4, 21, 282, 282])

index = 3
channel = 10
sample_image = batch['image'][index, channel, :, :].numpy()
sample_mask = batch['mask'][index, :, :].numpy()

fig, ax = plt.subplots(1, 2, figsize=(10, 5))
ax[0].imshow(sample_image)
ax[0].set_title('Sample Image')
ax[1].imshow(sample_mask, cmap='gray')
ax[1].set_title('Sample Mask')
```

image

image

image

image

image

image

robmarkcole:
@wangyi111 can you share your code for the RGB plotting in the paper? Adding a utility for that, examples below:

image

image

image

image

adamjstewart:
Did you figure out the non-matching image/mask bug?

The base class has a plot method. Is this not working?

robmarkcole:
I didn't notice the base class plot - these (using the base class plot) look decent so my hacky plot approach in a notebook must have an error in it

image

image

torchgeo/datamodules/copernicus/biomass_s3.py
```diff
@@ -0,0 +1,110 @@
...
+
+TARGET_SIZE = (282, 282)
+
+
+class CopernicusBenchBiomassS3DataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the Copernicus Biomass-S3 dataset.
+
+    Uses the train/val/test splits provided with the benchmark.
+
+    .. versionadded:: 0.81
```
adamjstewart:
```suggestion
    .. versionadded:: 0.8
```

tests/data/copernicus/l3_biomass_s3/data.py
adamjstewart, line 1: We should make the image and mask completely different sizes (maybe ~8 and ~32?) to better represent the dataset.

adamjstewart:
To add test coverage, add a config file to `tests/conf/` and 1 line to `TestPixelwiseRegressionTask.test_trainer` parameterization in `tests/trainers/test_regression.py`.

robmarkcole:
@adamjstewart codecov still short

adamjstewart:
See https://app.codecov.io/gh/torchgeo/torchgeo/pull/3086 for the lines missing coverage, or install the codecov plugin to directly view them on GitHub.

tests/data/copernicus/l3_biomass_s3/data.py
robmarkcole: done

robmarkcole:
New error now I am testing `time-series`:
```
RuntimeError: stack expects each tensor to be equal size, but got [21, 8, 8] at entry 0 and [21, 8, 10] at entry 1
```

robmarkcole:
@adamjstewart all green

robmarkcole:
IDK inspecting another batch, these just don't look right. @wangyi111 WDYT? 

image

image

image

image

robmarkcole:
I have managed to train a model, so perhaps I am just unused to viewing the S3 images. However given these are large areas of forest, perhaps that is possible even if the images are poorly registered

image

image

torchgeo/datamodules/copernicus/biomass_s3.py
```diff
@@ -0,0 +1,110 @@
...
+        self.std = torch.reciprocal(scale_factors)
+
+        resize_transform = K.AugmentationSequential(
+            K.Resize(
+                size=TARGET_SIZE, resample=Resample.BILINEAR.name, align_corners=False
+            ),
+            data_keys=None,
+            keepdim=True,
+            extra_args={
+                DataKey.MASK: {'resample': Resample.NEAREST, 'align_corners': None}
```
isaaccorley: If the mask is continuous valued this should be bilinear not nearest

isaaccorley (changes requested):

torchgeo/datamodules/copernicus/biomass_s3.py
robmarkcole: still looks off after this change 

Image

robmarkcole: align_corners doesn't make any difference either

isaaccorley: Yeah I'm not sure about the image/mask not matching, I was just commenting that you should use bilinear resampling for continuous values targets

adamjstewart: Image/mask alignment should be fixed in Kornia 0.8.2: https://github.com/torchgeo/torchgeo/pull/3094

robmarkcole:
@adamjstewart upgraded to `kornia-0.8.2` and the image/mask still look unregistered. I don't think this is a code issue, perhaps the images themselves are not well registered. Suggest a follow up investigation be performed by @wangyi111 

image

image

image

image

tests/datamodules/test_copernicus.py
adamjstewart, line 1: We are trying to get rid of this directory. Let's instead combine your SpatioTemporalPixelWiseRegressionTask PR with one of your two datamodule PRs so we can immediately start testing. I would suggest BioMassters since it's more widely used and there are concerns of bugs in this dataset.

torchgeo/datamodules/copernicus/biomass_s3.py
```diff
@@ -0,0 +1,113 @@
...
+
+TARGET_SIZE = (282, 282)
+
+
+class CopernicusBenchBiomassS3DataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the Copernicus Biomass-S3 dataset.
+
+    Uses the train/val/test splits provided with the benchmark.
+
+    .. versionadded:: 0.8
```
adamjstewart:
```suggestion
    .. versionadded:: 0.10
```

torchgeo/datamodules/copernicus/biomass_s3.py
```diff
@@ -0,0 +1,113 @@
...
+        )
+        existing_transform = cast(
+            Callable[[dict[str, torch.Tensor]], dict[str, torch.Tensor]] | None,
+            kwargs.get('transforms'),
+        )
+
+        if existing_transform is not None:
+            transform = existing_transform
+
+            def composed(sample: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
```
adamjstewart: Is this needed because the original images have variable size?

torchgeo/datasets/copernicus/biomass_s3.py
```diff
@@ -114,6 +117,29 @@ def __getitem__(self, index: int) -> Sample:
             case 'time-series':
                 paths = os.path.join(self.root, self.directory, 's3_olci', pid, '*.tif')
                 samples = [self._load_image(path) for path in sorted(glob.glob(paths))]
+                if not samples:
```
adamjstewart: What's the motivation for all of these changes?

adamjstewart (changes requested):

tests/datamodules/test_copernicus.py
robmarkcole: Will use BioMassters for the SpatioTemporalPixelWiseRegressionTask PR

torchgeo/datamodules/copernicus/biomass_s3.py
robmarkcole: Yes Biomass-S3 samples are not uniformly sized

torchgeo/datasets/copernicus/biomass_s3.py
robmarkcole: This block addresses shape-mismatch edge cases in time-series mode before stack_samples(samples):

Frames for the same PID have different heights (H) across timestamps.
Frames for the same PID have different widths (W) across timestamps.
Both H and W differ across timestamps.
Mixed case where some frames already have max size and others are smaller (only smaller ones get padded).
Prevents stack_samples/tensor stacking failures caused by non-uniform spatial size.
It pads only on right/bottom (F.pad(image, (0, pad_w, 0, pad_h))) to make all frames [C, max_h, max_w], and copies only modified sample dicts to avoid mutating originals.

torchgeo/datamodules/copernicus/biomass_s3.py
```diff
@@ -0,0 +1,113 @@
...
+                resized = resize_transform(sample)
+                return transform(resized)
+
+            kwargs['transforms'] = composed
+        else:
```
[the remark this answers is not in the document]
robmarkcole: Not sure about this

torchgeo/datamodules/copernicus/biomass_s3.py
```diff
@@ -0,0 +1,113 @@
...
+        super().__init__(CopernicusBenchBiomassS3, batch_size, num_workers, **kwargs)
+
+        normalizer = K.Normalize(mean=self.mean, std=self.std)
+        if mode == 'time-series':
+            self.aug = K.AugmentationSequential(
+                K.VideoSequential(normalizer),
+                data_keys=None,
+                keepdim=True,
+                same_on_batch=True,
+            )
```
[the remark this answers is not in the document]
robmarkcole: added _collate_time_series_batch

robmarkcole:
@adamjstewart merged in the trainer and tested with the datamodule

adamjstewart:
Will wait for one of the codeowners to review and approve.

torchgeo/datamodules/copernicus/biomass_s3.py
```diff
@@ -0,0 +1,136 @@
...
+                resized = resize_transform(sample)
+                return transform(resized)
+
+            kwargs['transforms'] = composed
+        else:
+            kwargs['transforms'] = resize_transform
+
+        super().__init__(CopernicusBenchBiomassS3, batch_size, num_workers, **kwargs)
+
+        normalizer = K.Normalize(mean=self.mean, std=self.std)
```
nilsleh, line 77: For a regression target, both the image inputs and the image regression targets should be normalized right?

torchgeo/trainers/spatiotemporal_regression.py
```diff
@@ -0,0 +1,161 @@
...
+        self, batch: Any, batch_idx: int, dataloader_idx: int = 0
+    ) -> None:
+        """Compute the validation loss and additional metrics.
+
+        Args:
+            batch: The output of your DataLoader.
+            batch_idx: Integer displaying index of this batch.
+            dataloader_idx: Index of the current dataloader.
+        """
+        self._shared_step(batch, 'val')
```
nilsleh: maybe I missed this in the trainer design, but why are val and test not returning anything?

torchgeo/datamodules/copernicus/biomass_s3.py
robmarkcole: I have found that normalisation of the target can make the difference between a good and bad result with convlstm, so we probably should support it somehow. I don't see target norm being applied consistently by datasets in torchgeo:

- DigitalTyphoon does target normalization with dataset-wide stats. 
- AirQualityDataModule does z-score normalization for regression targets, but using training-split-wide stats rather than full-dataset stats.

In prediction you then need to un-normalise, so norm stats should be accessible and their use tracked. @adamjstewart what do you suggest?

torchgeo/trainers/spatiotemporal_regression.py
robmarkcole: This is because _shared_step logs the loss and updates the RegressionMixin metrics, you think a comment here would be useful?

torchgeo/datamodules/copernicus/biomass_s3.py
adamjstewart: My general preference is for normalization in the datamodule and denormalization in the trainer. There are exceptions to this (e.g., sometimes it's easier to normalize a pandas DataFrame once in the dataset instead of batch-wise normalization in the datamodule). But I don't love exceptions and we should try to be more consistent in general.

The new AirQuality dataset/datamodule and TemporalRegressionTask trainer are another recent example of normalizing both input and output values.

robmarkcole: Done: 

- Added TARGET_MEAN / TARGET_STD to biomass_s3.py (becomes a new standard for regression datasets, existence of these is queried by the trainer)
- CopernicusBenchBiomassS3DataModule now normalizes batch['mask'] when present and moves target stats to the active device.
- SpatioTemporalPixelwiseRegressionTask now computes loss in normalized space, but denormalizes targets/predictions for metrics and predict_step: spatiotemporal_regression.py

torchgeo/trainers/spatiotemporal_regression.py
```diff
@@ -0,0 +1,181 @@
...
+            x: Input tensor of shape ``(B, T, C, H, W)``.
+            **kwargs: Additional keyword arguments forwarded to the model,
+                e.g. ``lengths`` for variable-length sequences.
+
+        Returns:
+            Output tensor of shape ``(B, 1, H, W)``.
+        """
+        return self.model(x, **kwargs)
+
+    def _target_stats(self) -> tuple[Tensor, Tensor] | None:
```
robmarkcole: This is a bit ugly

nilsleh (dismissed):

tests/datamodules/test_copernicus.py
adamjstewart, line 1: Delete this file

tests/datasets/test_copernicus.py
```diff
@@ -126,6 +128,43 @@ def test_not_rgb(self, dataset: CopernicusBench) -> None:
...
+        root = os.path.join('tests', 'data', 'copernicus', 'l3_biomass_s3')
+        dataset = CopernicusBench('biomass_s3', root, mode='time-series')
+
+        sample = dataset[0]
+        assert sample['image'].shape == (2, 21, 8, 10)
+
+        sample = dataset[1]
+        assert sample['image'].shape == (2, 21, 9, 9)
+
+    def test_biomass_s3_time_series_missing_samples(self, tmp_path: Path) -> None:
```
adamjstewart: Do we need any of these new tests?

torchgeo/datasets/copernicus/biomass_s3.py
```diff
@@ -114,6 +116,31 @@ def __getitem__(self, index: int) -> Sample:
...
+                    msg = f'No Sentinel-3 samples found for {pid}.'
+                    raise FileNotFoundError(msg)
```
adamjstewart: Does this ever occur?

adamjstewart (changes requested):
Need to review more after downloading the full dataset, but here is a first glance.

tests/datasets/test_copernicus.py
robmarkcole: - test_biomass_s3_static_collate_error: not essential, removed
- test_biomass_s3_time_series_padding: required as dataset contains images that require padding
- test_biomass_s3_time_series_missing_samples: removed

torchgeo/datasets/copernicus/biomass_s3.py
robmarkcole: checked and it doesn't so removed

robmarkcole:
@adamjstewart comments addressed

tests/datasets/test_copernicus.py
adamjstewart: > test_biomass_s3_time_series_padding: required as dataset contains images that require padding

And the datamodule tests don't cover this?

robmarkcole: I deleted the datamodule tests

adamjstewart: I mean the trainer tests which hit the datamodule and dataset.

robmarkcole: I've updated the test dataset so that it will

robmarkcole:
## 25 epochs, convlstm experiment
```python3
datamodule = CopernicusBenchBiomassS3DataModule(
    root=root,
    mode='time-series',
    bands=bands,
    batch_size=4,
    num_workers=7,
    download=True,
)

task = SpatioTemporalPixelwiseRegressionTask(
    model='convlstm',
    in_channels=len(bands),
)
```

image

```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃        Test metric        ┃       DataLoader 0        ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│         test_MAE          │     70.92227935791016     │
│         test_MSE          │       8859.1953125        │
│         test_RMSE         │     94.12329864501953     │
│         test_loss         │    0.6255610585212708 
```

## Test predictions
image
image
image

adamjstewart:
Can you rebase and move from trainer to task?

robmarkcole:
@adamjstewart rebased and renamed trainer to `SpatioTemporalRegression`

adamjstewart:
This should be renamed to SpatioTemporalPixelwiseRegression to distinguish from #3591.

tests/tasks/test_regression.py
```diff
@@ -263,6 +272,56 @@ def test_trainer(
         except MisconfigurationException:
             pass
 
+    def test_biomass_s3_existing_transform_is_composed(self) -> None:
```
adamjstewart: Why are these new tests needed?

tests/tasks/test_spatiotemporal_pixelwise_regression.py
```diff
@@ -0,0 +1,100 @@
...
+import torch
+from lightning.pytorch import Trainer
+from torch import Tensor
+
+from torchgeo.datamodules import MisconfigurationException
+from torchgeo.main import main
+from torchgeo.tasks import SpatioTemporalPixelwiseRegression
+
+
+class ConstantRegressionModel(torch.nn.Module):
```
adamjstewart: Wonder if we should just test with the real model, we only added these in other tasks to speed up testing, but then we aren't ensuring things work with a real model.

tests/tasks/test_spatiotemporal_pixelwise_regression.py
```diff
@@ -0,0 +1,100 @@
...
+            main(['test', *args])
+        except MisconfigurationException:
+            pass
+        try:
+            main(['predict', *args])
+        except MisconfigurationException:
+            pass
+
+    @pytest.mark.parametrize('loss', ['mse', 'mae'])
+    def test_task(self, loss: Literal['mse', 'mae']) -> None:
```
adamjstewart: What is this test needed for? It seems like we could get 100% coverage without this by creating config yaml files for all loss functions.

tests/tasks/test_spatiotemporal_pixelwise_regression.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+        model.model = ConstantRegressionModel()
+        # Attach a datamodule with no target stats to exercise _target_stats()
+        # returning None after trainer lookup succeeds.
+        model._trainer = cast(Trainer, SimpleNamespace(datamodule=SimpleNamespace()))
+        batch = {'image': torch.randn(2, 4, 3, 16, 16)}
+
+        y_hat = model.predict_step(batch, 0)
+
+        assert torch.all(y_hat == 0)
```
adamjstewart: These could be combined into a single test for predict_step. See `tests/tasks/test_spatiotemporal_segmentation.py` for a simple solution.

torchgeo/datamodules/copernicus/biomass_s3.py
```diff
@@ -0,0 +1,178 @@
...
+    'Oa21_radiance': 0.00324118,
+}
+
+TARGET_SIZE = (282, 282)
+# Dataset-wide biomass statistics after resizing masks to TARGET_SIZE.
+TARGET_MEAN = torch.tensor(93.197777079690, dtype=torch.float32)
+TARGET_STD = torch.tensor(119.004185235754, dtype=torch.float32)
+
+
+def _collate_time_series_batch(
```
adamjstewart: Reuse `torchgeo.datasets.utils.pad_across_batches`

torchgeo/datamodules/copernicus/biomass_s3.py
```diff
@@ -0,0 +1,178 @@
...
+        if existing_transform is not None:
+            transform = existing_transform
+
+            def composed(sample: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
+                resized = resize_transform(sample)
+                return transform(resized)
+
+            kwargs['transforms'] = composed
+        else:
+            kwargs['transforms'] = resize_transform
```
adamjstewart: This is rather complex, can we do this automatically instead without checking? There's no reason we can't resize twice (just slower).

torchgeo/datamodules/copernicus/biomass_s3.py
```diff
@@ -0,0 +1,178 @@
...
+            batch: A batch of data that needs to be altered or augmented.
+            dataloader_idx: The index of the dataloader to which the batch belongs.
+
+        Returns:
+            A batch with normalized imagery and target masks.
+        """
+        batch = super().on_after_batch_transfer(batch, dataloader_idx)
+        batch['image'] = torch.nan_to_num(
+            batch['image'], nan=0.0, posinf=0.0, neginf=0.0
+        )
```
adamjstewart: Are there NaNs in this dataset?

torchgeo/datamodules/copernicus/biomass_s3.py
```diff
@@ -0,0 +1,178 @@
...
+                data_keys=None,
+                keepdim=True,
+                same_on_batch=True,
+            )
+        else:
+            self.aug = K.AugmentationSequential(
+                normalizer, data_keys=None, keepdim=True
+            )
+
+    def transfer_batch_to_device(
```
adamjstewart: Is this needed?

torchgeo/datasets/copernicus/biomass_s3.py
```diff
@@ -114,6 +115,28 @@ def __getitem__(self, index: int) -> Sample:
...
+                        h, w = image.shape[-2:]
+                        pad_h = max_h - h
+                        pad_w = max_w - w
+                        if pad_h or pad_w:
+                            # Pad only bottom/right so existing pixels keep their origin.
+                            padded_image = F.pad(image, (0, pad_w, 0, pad_h))
+                            sample_dict = sample_dict.copy()
+                            sample_dict['image'] = padded_image
+                        padded_samples.append(sample_dict)
+                    samples = padded_samples
```
adamjstewart: Just resize all images/masks to the same dimensions here, then you don't need to do anything else in datamodule.

torchgeo/tasks/spatiotemporal_pixelwise_regression.py
```diff
@@ -0,0 +1,135 @@
...
+        try:
+            datamodule = self.trainer.datamodule
+        except RuntimeError:
+            return None
+
+        mean = getattr(datamodule, 'target_mean', None)
+        std = getattr(datamodule, 'target_std', None)
+        if isinstance(mean, Tensor) and isinstance(std, Tensor):
+            return mean, std
+        return None
```
adamjstewart: Defensive coding

torchgeo/tasks/spatiotemporal_pixelwise_regression.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def _denormalize(self, x: Tensor) -> Tensor:
+        """Denormalize target values if datamodule statistics are available."""
+        if (stats := self._target_stats()) is not None:
+            mean, std = stats
+            x = x * std + mean
+        return x
```
adamjstewart: We were able to do this in 2 lines of code in `torchgeo/tasks/temporal_regression.py`

adamjstewart (changes requested):
Still lots of slop and defensive coding we can cut out

torchgeo/tasks/spatiotemporal_pixelwise_regression.py
robmarkcole: Removed the optional _target_stats and _denormalize fallback helpers. The task now relies directly on the configured datamodule contract.

tests/tasks/test_spatiotemporal_pixelwise_regression.py
robmarkcole: Removed test_task and its manual loss parameterization. Loss coverage will be handled through the YAML integration cases.

torchgeo/tasks/spatiotemporal_pixelwise_regression.py
robmarkcole: Simplified denormalization to direct target_std and target_mean operations in _shared_step and predict_step, matching TemporalRegression.

torchgeo/datamodules/copernicus/biomass_s3.py
robmarkcole: There was no evidence of NaNs in the dataset, so I removed the nan_to_num handling and its synthetic test.

tests/tasks/test_spatiotemporal_pixelwise_regression.py
robmarkcole: Removed ConstantRegressionModel and the mock-based prediction tests. The remaining YAML integration test exercises the real ConvLSTM model.

torchgeo/datamodules/copernicus/biomass_s3.py
robmarkcole: Correction after scanning the complete raw dataset: there are no NaNs, but one Sentinel-3 raster contains a declared -inf nodata pixel in all 21 bands. Handling is now targeted in CopernicusBenchBiomassS3._load_image: only -inf is replaced. The fixture reproduces the nodata pixel and verifies finite output for static and time-series modes.

robmarkcole:
Scanned the dataset, there are no NaNs, but real -inf nodata exists for a single pixel:

- File: S3B_20200111T014412_20200111T014558.tif, pixel (83, 0).

Added targetted handling for Sentinel-3 -inf nodata during dataset loading and covered with a test/test-data

tests/tasks/test_spatiotemporal_pixelwise_regression.py
robmarkcole: Replaced with single test, required for coverage

torchgeo/datamodules/copernicus/biomass_s3.py
robmarkcole: No, removed

torchgeo/datamodules/copernicus/biomass_s3.py
robmarkcole: Done

torchgeo/datasets/copernicus/biomass_s3.py
robmarkcole: done

torchgeo/datamodules/copernicus/biomass_s3.py
robmarkcole: resizing address in the dataset

tests/tasks/test_regression.py
robmarkcole: removed

tests/tasks/test_spatiotemporal_pixelwise_regression.py
```diff
@@ -0,0 +1,61 @@
...
+    def test_predict_step(self) -> None:
+        model = SpatioTemporalPixelwiseRegression(
+            in_channels=3, hidden_dim=8, num_layers=1
+        )
+        model._trainer = cast(
+            Trainer,
+            SimpleNamespace(
+                datamodule=SimpleNamespace(target_mean=2.0, target_std=3.0)
+            ),
+        )
```
adamjstewart: Not sure how I feel about whatever this is. I have my own try-except hack in `tests/tasks/test_spatiotemporal_segmentation.py` that I also don't love. Wish we could find a better long-term solution to test this. Perhaps monkeypatch to copy `test_dataset` to `predict_dataset` and use an actual datamodule/trainer?

adamjstewart (dismissed):
I think this is okay now, but I'm very tired and would love a second set of eyes on this. If another maintainer approves, I think we can merge.

tests/tasks/test_spatiotemporal_pixelwise_regression.py
robmarkcole: Replaced with real CopernicusBenchBiomassS3DataModule - its test dataset is copied to predict_dataset followed by an actual Trainer.predict call

adamjstewart (dismissed):

isaaccorley (changes requested):
This lgtm but it's missing the datamodule in the docs docs/api/datamodules.rst

isaaccorley (approved):

PR #3094, opened 2025-11-11
kornia 0.8.2+ required for working VideoSequential

adamjstewart:
Spent way too much time debugging this but the bug has already been fixed in the latest kornia release. Without this, all time series and change detection datasets that use `RandomCrop` will not have the same crop applied to all timestamps.

@hkristen


hkristen:
😅 so we both have spent way too much time debugging something that was already solved by someone else - thx for tracking this down!

PR #3095, opened 2025-11-12
Fix bug in plotting of BRIGHT

### LINKED ISSUE #3092: BRIGHT: legend bug
### Description

There is a bug in the `plot` method of the `BRIGHTDFC2025` dataset. Specifically, the legend colors and/or labels are incorrect when a subset of values are present.

### Steps to reproduce

```python
from torchgeo.datasets import BRIGHTDFC2025

ds = BRIGHTDFC2025('data/bright', download=True)
sample = ds[2]
ds.plot(sample)
plt.show()
```
You should see an image like:
Image

This bug occurs for most images with less than 4 classes, not just image 2.

I would personally be fine with a legend containing all classes even if all classes are not present in the image if that makes things easier.

### Version

0.8.0.dev0 (1bb6f8e63ea94c767e1e16da66bbbe6da56399b6)

nilsleh:
Closes  #3092 by showing all labels regardless of the present labels in the image.

### DIFF
--- a/torchgeo/datasets/bright.py
+++ b/torchgeo/datasets/bright.py
@@ -10,7 +10,6 @@ from typing import ClassVar
 
 import matplotlib.patches as mpatches
 import matplotlib.pyplot as plt
-import numpy as np
 import rasterio
 import torch
 from einops import repeat


adamjstewart:
This is still broken. If only 2 classes exist, regardless of which classes they are, they appear as "background" and "destroyed":

bright_intact

See #3097 for the correct fix.

PR #3103, opened 2025-11-12
OSCD: support for rectangular data

### LINKED ISSUE #2920: Change Detection transforms

adamjstewart:
The OSCD dataset contains rectangular images. Our data module was not set up to handle this. This PR includes the following changes:

- [x] Create fake test data with rectangular images
- [x] Create enough data to have non-singular mini-batches
- [x] Use RandomCrop/CenterCrop on each image at a time

Closes #2920 @blaz-r @hkristen

### DIFF
--- a/torchgeo/datasets/oscd.py
+++ b/torchgeo/datasets/oscd.py
@@ -155,7 +155,12 @@ class OSCD(NonGeoDataset):
         sample = {'image': image, 'mask': mask}
 
         if self.transforms is not None:
+            # FIXME: VideoSequential only works with a batch dimension
+            sample['image'] = sample['image'].unsqueeze(0)
+            sample['mask'] = sample['mask'].unsqueeze(0)
             sample = self.transforms(sample)
+            sample['image'] = sample['image'].squeeze(0)
+            sample['mask'] = sample['mask'].squeeze(0)
 
         return sample
 
--- a/tests/datasets/test_oscd.py
+++ b/tests/datasets/test_oscd.py
@@ -78,9 +78,9 @@ class TestOSCD:
 
     def test_len(self, dataset: OSCD) -> None:
         if dataset.split == 'train':
-            assert len(dataset) == 2
+            assert len(dataset) == 4
         else:
-            assert len(dataset) == 1
+            assert len(dataset) == 2
 
     def test_add(self, dataset: OSCD) -> None:
         ds = dataset + dataset
--- a/tests/trainers/test_change.py
+++ b/tests/trainers/test_change.py
@@ -5,6 +5,7 @@ import os
 from pathlib import Path
 from typing import Any, Literal
 
+import kornia.augmentation as K
 import pytest
 import segmentation_models_pytorch as smp
 import timm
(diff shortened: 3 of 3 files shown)


torchgeo/datasets/oscd.py
```diff
@@ -155,7 +155,12 @@ def __getitem__(self, index: int) -> dict[str, Tensor]:
         sample = {'image': image, 'mask': mask}
 
         if self.transforms is not None:
+            # FIXME: VideoSequential only works with a batch dimension
```
adamjstewart, line 158: Kinda hacky, but couldn't figure out a simpler solution. @hkristen has something a little safer in https://github.com/hkristen/torchgeo/commit/2599d581 but it's more lines of code. Would be curious which other people prefer.

blaz-r: I think this is fine. The solution by Harald is also okay, but it may create some computational overhead.

adamjstewart: Should be quite similar.

PR #3104, opened 2025-11-13
Update features.csv with the correct numbers of weights for TerraTorch

biancazadrozny:
I have updated the number of datasets and weights in TerraTorch.  We don’t have the concept of datasets in TerraTorch, we rely on TorchGeo for that, so I changed the count of datasets to 0 put in parentheses the same number as TorchGeo.  As for weights, we currently have integrated 1374 models into TerraTorch (1282 are TIMM models). The count can be obtained by running: len(BACKBONE_REGISTRY) in TerraTorch.

### DIFF
--- a/docs/user/metrics/features.csv
+++ b/docs/user/metrics/features.csv
@@ -1,7 +1,7 @@
 Library,ML Backend,I/O Backend,Spatial Backend,Transform Backend,Datasets,Weights,CLI,GUI,Reprojection,STAC,Time Series
 `TorchGeo`_,PyTorch,"GDAL, h5py, laspy, NetCDF4, OpenCV, pandas, pillow, scipy, xarray","geopandas, R-tree*",Kornia,127,120,✅,❌,✅,❌,🚧
 `OTB`_,"LibSVM, OpenCV, Shark",GDAL,-,-,0,0,✅,✅,✅,❌,❌
-`TerraTorch`_,PyTorch,"GDAL, h5py, pandas, xarray",-,Albumentations,0 (127 from TorchGeo),1374,✅,❌,✅,❌,🚧
+`TerraTorch`_,PyTorch,"GDAL, h5py, pandas, xarray",-,Albumentations,27,13,✅,❌,✅,❌,🚧
 `Raster Vision`_,"PyTorch, TensorFlow*","GDAL, OpenCV, pandas, pillow, scipy, xarray",STAC,Albumentations,0,6,✅,❌,✅,✅,🚧
 `DeepForest`_,"PyTorch, TensorFlow*","GDAL, OpenCV, pandas, pillow, scipy",R-tree,Albumentations,0,4,❌,❌,❌,❌,❌
 `samgeo`_,PyTorch,"GDAL, OpenCV, pandas, xarray",geopandas,numpy,0,0,❌,✅,✅,❌,❌


isaaccorley:
I don't believe we are considering timm backbone weights as models, otherwise TorchGeo itself would also support an additional 1282 models. Our models are pretrained weights that are trained on EO data.

adamjstewart:
Yep. For datasets, we were including formal data loaders in https://github.com/terrastackai/terratorch/tree/main/terratorch/datasets, which there are around 27 last I checked. Many of these have been moved out into geobench or terratorch iterate, but I would still count them until they're removed. I wouldn't count datasets in dependencies like TorchGeo.

For weights, only model weights specific to terratorch, not counting ones from timm or torchgeo.

adamjstewart:
I also updated the URL in #3083

biancazadrozny:
I see. For the datasets, 27 makes sense in this view.

So, for the weights, the current ones that are specific to TerraTorch (not counting timm or TorchGeo) are: 
terratorch_prithvi_eo_tiny
terratorch_prithvi_eo_v1_100
terratorch_prithvi_eo_v2_tiny_tl
terratorch_prithvi_eo_v2_100_tl
terratorch_prithvi_eo_v2_300
terratorch_prithvi_eo_v2_600
terratorch_prithvi_eo_v2_300_tl
terratorch_prithvi_eo_v2_600_tl
terratorch_terramind_v1_base
terratorch_terramind_v1_base_tim
terratorch_terramind_v01_base
terratorch_terramind_v01_base_tim
terratorch_terramind_v1_large
terratorch_terramind_v1_large_tim
terratorch_terramind_v1_tiny
terratorch_terramind_v1_tiny_tim
terratorch_terramind_v1_small
terratorch_terramind_v1_small_tim

I am not sure if we should count all these versions separately or only say 2 (for Prithvi and Terramind)? 

(We also have some manually integrated models like SatMAE and ScaleMAE that are not currently in the registry but we will update this for the next version of TerraTorch -- and then we can ask for another revision)

adamjstewart:
Let's count each one separately. So 18 in total.

isaaccorley:
What is the difference between `terratorch_terramind_v1_base` and `terratorch_terramind_v1_base_tim`?

biancazadrozny:
The _tim versions are versions for "Thinking in Modalities" (see https://research.ibm.com/blog/thinking-in-modalities-terramind). The weights are the same but the model code is different because it does generation of the missing modalities. So for the purpose of counting weights, they should not count...

adamjstewart:
I'm guessing v01 is version 0.1 and v1 is version 1.0? Or are v01 and v1 the same?

isaaccorley:
If I understand correctly I think this means there are 13 weights?

```
terratorch_prithvi_eo_tiny
terratorch_prithvi_eo_v1_100
terratorch_prithvi_eo_v2_tiny_tl
terratorch_prithvi_eo_v2_100_tl
terratorch_prithvi_eo_v2_300
terratorch_prithvi_eo_v2_600
terratorch_prithvi_eo_v2_300_tl
terratorch_prithvi_eo_v2_600_tl
terratorch_terramind_v1_base
terratorch_terramind_v01_base
terratorch_terramind_v1_large
terratorch_terramind_v1_tiny
terratorch_terramind_v1_small
```

docs/user/metrics/features.csv
```diff
@@ -1,7 +1,7 @@
 Library,ML Backend,I/O Backend,Spatial Backend,Transform Backend,Datasets,Weights,CLI,GUI,Reprojection,STAC,Time Series
 `TorchGeo`_,PyTorch,"GDAL, h5py, laspy, NetCDF4, OpenCV, pandas, pillow, scipy, xarray","geopandas, R-tree*",Kornia,127,120,✅,❌,✅,❌,🚧
 `OTB`_,"LibSVM, OpenCV, Shark",GDAL,-,-,0,0,✅,✅,✅,❌,❌
-`TerraTorch`_,PyTorch,"GDAL, h5py, pandas, xarray",-,Albumentations,27,1,✅,❌,✅,❌,🚧
+`TerraTorch`_,PyTorch,"GDAL, h5py, pandas, xarray",-,Albumentations,0 (127 from TorchGeo),1374,✅,❌,✅,❌,🚧
```
isaaccorley:
```suggestion
`TerraTorch`_,PyTorch,"GDAL, h5py, pandas, xarray",-,Albumentations,27,13,✅,❌,✅,❌,🚧
```
Adam: edited to have 27 datasets

adamjstewart:
I'll also update this in our book chapter: https://arxiv.org/abs/2510.02572

PR #3112, opened 2025-11-18
PRUE FTW Models

isaaccorley:
Models from the upcoming paper in submission titled `PRUE` from the FTW team. Note these weights are noncommercial since they are trained on the full FTW dataset. We will be adding some commercial usable models soon.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Clone repo
-        uses: actions/checkout@v5.0.0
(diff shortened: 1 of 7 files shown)


calebrob6:
Works great -- https://gist.github.com/calebrob6/2b0143a02d9137b3541ae01ac2da274c

calebrob6:
On a semi-random patch from Spain test set:

image

adamjstewart:
Is there a more specific license these are distributed under? For example, CC-BY-NC-4.0? Or is "non-commercial" all we know?

isaaccorley:
The FTW dataset has a mix of noncommercial licenses so it's unclear which one takes precedence.

PR #3119, opened 2025-11-20
fix: IDTReeS geo to pixel boxes conversion bug fix

### LINKED ISSUE #3115: IDTReeS: no bounding boxes returned
### Description

The merge of #2962 broke bounding box loading in IDTReeS. The result is that no bounding boxes are returned. Not only does this make the dataset useless, it also prevents us from testing a few lines of code in `plot`! (I'm joking, the former is obviously more important, I'm very happy that the drop in test coverage found this bug)

### Steps to reproduce

Run the following code before and after #2962 
```python
from torchgeo.datasets import IDTReeS

ds = IDTReeS('data/idtrees', download=True, checksum=True)
sample = ds[0]
ds.plot(sample)
plt.show()
```
Before #2962, the image contained bounding boxes. After, it doesn't.

### Version

0.8.0.dev0 (53410129798b59ec360beae952bb5cbbac9d5467)

isaaccorley:
This fixes a bug where all boxes were being filtered out because they were out of bounds due to not being converted to pixel coordinates correctly. Tested on the actual dataset locally

Fixes #3115

### DIFF
--- a/tests/datasets/test_idtrees.py
+++ b/tests/datasets/test_idtrees.py
@@ -62,6 +62,7 @@ class TestIDTReeS:
             if x['bbox_xyxy'].ndim != 1:
                 assert x['bbox_xyxy'].ndim == 2
                 assert x['bbox_xyxy'].shape[-1] == 4
+                assert x['bbox_xyxy'].shape[0] > 0
 
     def test_len(self, dataset: IDTReeS) -> None:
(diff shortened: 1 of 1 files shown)


adamjstewart:
Could you also modify the tests to assert that N number of bounding boxes are returned from the test data? Just check the dimensions of the sample value.

PR #3122, opened 2025-11-21
Change BTC and tests to support no weight downloading

blaz-r:
This PR implements a small refactor that enables BTC to operate without downloading weights, and also refactors the tests to leverage this change.


adamjstewart (approved):
Confirmed that all tests pass even without internet access, thanks!

Changed the default to False to match all other datasets/models. No datasets/models should ever download anything by default (security, principle of least surprise).

P.S. There are a couple other models that also need this treatment if you're ever bored and looking for something to fix: https://github.com/torchgeo/torchgeo/issues/1088

adamjstewart (approved):

isaaccorley (approved):

PR #3125, opened 2025-11-22
Related Libraries: update metrics

adamjstewart:
Quick update before our next release.

### DIFF
--- a/docs/user/metrics/update_downloads.py
+++ b/docs/user/metrics/update_downloads.py
@@ -4,7 +4,6 @@
 # Licensed under the MIT License.
 
 import argparse
-import re
 import time
 
 import pandas as pd


docs/user/metrics/common.py
```diff
@@ -6,20 +6,19 @@
...
-    'Raster Vision',
     'DeepForest',
+    'Raster Vision',
     'samgeo',
     'spopt',
     'GDL',
     'SITS',
     'SPy',
     'srai',
-    'geo-bench',
```
adamjstewart, line 17: geo-bench has not had a commit in over a year and is now removed from the list. Looking forward to adding geo-bench 2 to the list when it hits 100+ stars and stable releases @nilsleh

docs/user/metrics/common.py
```diff
@@ -69,7 +68,6 @@
 name_to_codecov = {
     'SITS': ('gh', 'e-sensing', 'sits'),
     'spopt': ('gh', 'pysal', 'spopt'),
-    'DeepForest': ('gh', 'weecology', 'DeepForest'),
```
adamjstewart, line 72: Codecov API seems to have problems with DeepForest because there is no default branch set. Switching to hard-coded coverage for now @bw4sz

docs/user/metrics/common.py
adamjstewart, line 88: @romeokienzler I've heard that TerraTorch's test coverage has recently increased but don't yet see any uploads to Codecov so can't check.

docs/user/metrics/update_downloads.py
```diff
@@ -116,19 +116,9 @@ def condaforge(package: str) -> int:
     Returns:
         Total number of downloads.
     """
-    # TODO: should really be using one of the following instead:
-    # https://github.com/conda-incubator/condastats
-    # https://github.com/anaconda/anaconda-package-data
-    url = f'https://anaconda.org/conda-forge/{package}'
-    pattern = r'<span>(\d+)</span> total downloads'
+    url = f'https://api.anaconda.org/repocore/channels/conda-forge/artifacts/conda/{package}'
```
adamjstewart, line 118: Through a happy accident, conda's website changed and the JavaScript is no longer rendered properly. This breaking change actually led me to discover the actual API being used to populate the webpage. So we no longer need to download the HTML for the entire webpage, we can just query the API directly!

nilsleh (approved):

isaaccorley (approved):

PR #3135, opened 2025-11-25
Preserve bounds through GeoDataModule pipeline

hkristen:
First step towards tiled inference as mentioned in https://github.com/torchgeo/torchgeo/discussions/3099#discussioncomment-15073150

@adamjstewart let me know if this is what you where thinking.

### DIFF
--- a/tests/datamodules/test_geo.py
+++ b/tests/datamodules/test_geo.py
@@ -211,6 +211,15 @@ class TestGeoDataModule:
         assert not dm.test_dataloader().drop_last
         assert not dm.predict_dataloader().drop_last
 
+    def test_bounds_preserved_in_batch(self, datamodule: CustomGeoDataModule) -> None:
+        """Test that bounds are preserved through DataLoader as tensors."""
+        datamodule.setup('predict')
+        if datamodule.trainer:
+            datamodule.trainer.predicting = True
+        batch = next(iter(datamodule.predict_dataloader()))
+        assert 'bounds' in batch
+        assert isinstance(batch['bounds'], Tensor)
+
 
 class TestNonGeoDataModule:
     @pytest.fixture
--- a/tests/datasets/test_utils.py
+++ b/tests/datasets/test_utils.py
@@ -421,6 +421,46 @@ class TestCollateFunctionsMatchingKeys:
             assert torch.allclose(samples[i]['image'], new_samples[i]['image'])
             assert samples[i]['crs'] == new_samples[i]['crs']
 
+    def test_stack_samples_with_bounds(self) -> None:
+        from datetime import datetime
+
+        import pandas as pd
+
+        t1 = pd.Timestamp(datetime(2020, 1, 1))
+        t2 = pd.Timestamp(datetime(2020, 6, 1))
+        samples = [
+            {
+                'image': torch.tensor([1, 2, 3]),
+                'bounds': (
+                    slice(-1.0, 0.0, None),
+                    slice(-1.0, 0.0, None),
+                    slice(t1, t2),
+                ),
+            },
+            {
+                'image': torch.tensor([4, 5, 6]),
+                'bounds': (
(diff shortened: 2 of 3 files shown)


adamjstewart:
This is kinda what I'm thinking, although I think there are a few different places we could do this:

1. **GeoDataset**: we could change the `bounds` key returned by all GeoDatasets. Not backwards-compatible, but something worth considering. I've been wanting to ensure that all datasets consistently return `dict[str, Tensor]` to make type hints easier. Currently `bounds` and `crs` are the most common things that aren't Tensor.
2. **GeoDataModule**: we currently pop bounds/crs keys in our GeoDataModules before transferring data to the GPU. We could also convert that data to a GPU-compatible format, or somehow save it so that it doesn't get moved to the GPU but all other dict keys are.
3. **collate_fn**: this is your current proposal. I was hoping to slowly get rid of our custom collation functions because most people don't think to use them and if we only have `dict[str, Tensor]`, then PyTorch's builtin collation functions work.

So I'm kinda leaning towards 1, although it is backwards-incompatible. What do you think?

Another idea could be that all GeoDatasets return:
```python
{
    'bounds': Tensor[xmin, xmax, ymin, ymax],
    'interval': Tensor[tmin, tmax],
}
```
Then we don't have very different units being combined. But this is less important to me.

For CRS, we can grab that from the dataset. Do we also need the rasterio transform? That can also be easily represented as a Tensor.

P.S. One other consideration that complicates all of this. At one time we were talking about trying to avoid unnecessary reprojection such that you can have UTM zones without having to reproject everything to a single CRS. This would mean that we can no longer reliably grab the CRS from the dataset. We would store this in our GeoDataFrame index, so if we keep track of which file we are loading from, we can grab that metadata in a more roundabout way. So don't worry too much right now, but keep this in mind.

hkristen:
>So I'm kinda leaning towards 1, although it is backwards-incompatible. What do you think?
I can see the point of Option 1 - I am still leaning towards Option 3 for pragmatic reasons:
1. No breaking changes
2. Unblocks working on tiled inference faster, which is a killer feature in my opinion

> For CRS, we can grab that from the dataset. Do we also need the rasterio transform? That can also be easily represented as a Tensor.

For tiled inference, I don't think we need the transform - we can reconstruct it from bounds + image dimensions when writing the output. For CRS, grabbing it from the dataset works for a dataset with only one CRS. 

> P.S. One other consideration that complicates all of this. At one time we were talking about trying to avoid unnecessary reprojection such that you can have UTM zones without having to reproject everything to a single CRS. This would mean that we can no longer reliably grab the CRS from the dataset. We would store this in our GeoDataFrame index, so if we keep track of which file we are loading from, we can grab that metadata in a more roundabout way. So don't worry too much right now, but keep this in mind.

For real-world use cases where you run inference over large areas like e.g. the whole Alps with multiple UTM zones, I suppose users would typically run separate inferences per zone (one per machine/node), each producing one output GeoTIFF in that zone's CRS. The current approach with dataset-level CRS handles this workflow well.
But I can also see a scenario where you have a smaller ROI that is split between two UTM zones and starting two inference runs is kinda cumbersome - maybe more of an edge case that we could tackle later?

adamjstewart:
I agree that your solution is faster and simpler, but it doesn't feel like the "right" way to do it. Since we're before 1.0, I want to make breaking changes now instead of later. The ability to guarantee that all datasets return `dict[str, Tensor]` means that:

1. We are compatible with PyTorch's builtin collation function, and no one ever needs to use a different custom one
2. We get rid of half of the `Any` type hints in TorchGeo, which currently disables type checking for those datasets

In terms of faster and making users angry, this is definitely something I'm willing to do myself. I like tedious PRs 🙂

hkristen:
Breaking stuff and doing it right before 1.0 make sense ... I would be happy if you do this, I am sure you are faster and more precise ;)

I can work on the follow up PRs regarding tiled inference then.

PR #3138, opened 2025-11-26
GeoDataset: remove 'crs', convert 'bounds'

adamjstewart:
Reboot of #2567, #2249
See #985, #3099 for discussion
Closes #3135
Alternative to #3148
Closes #3148

### Pros

* Compatibility with PyTorch's `default_collate` function
* Avoids most uses of dynamic typing (`typing.Any`)

### Cons

* Backwards-incompatible changes (removing CRS, changing format of bounds)

@hkristen

### DIFF
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -595,7 +595,6 @@ class TestVectorDataset:
         dataset_parquet.task = 'semantic_segmentation'
         x = dataset_parquet[dataset_parquet.bounds]
         assert isinstance(x, dict)
-        assert isinstance(x['crs'], CRS)
         assert isinstance(x['mask'], torch.Tensor)
         assert torch.equal(
             x['mask'].unique(),  # type: ignore[no-untyped-call]
@@ -605,7 +604,6 @@ class TestVectorDataset:
         dataset_parquet.task = 'object_detection'
         x = dataset_parquet[dataset_parquet.bounds]
         assert isinstance(x, dict)
-        assert isinstance(x['crs'], CRS)
         assert isinstance(x['bbox_xyxy'], torch.Tensor)
         assert isinstance(x['label'], torch.Tensor)
         assert x['bbox_xyxy'].shape[-1] == 4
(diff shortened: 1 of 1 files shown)


adamjstewart:
I'm wondering whether we should stick to bounds/crs changes in this PR and type hints in another PR. Last time I got stuck with object detection bboxes, which are `list[Tensor]` at the moment.

adamjstewart:
TODO: point datasets previously returned their results as `bounds`. Change this to `keypoints`.

isaaccorley:
Might want to revisit nested tensors for the boxes. I think it's fine to return boxes as Tensor though but the return type from the dataloader when collating will probably always have a custom collate fn for this. I'm not sure if there's a workaround for stacking tensors with variable shaped dims. https://docs.pytorch.org/docs/stable/nested.html

adamjstewart:
Actually, let's split this up into more manageable PRs:

- [x] Point datasets: use 'keypoints' instead of 'bounds': #3139
- [x] Convert 'bounds' from `tuple[slice]` to `Tensor`

then @hkristen will be unblocked.

- [x] Get rid of 'crs'
- [ ] Get rid of other non-Tensor outputs
- [ ] Update type hints to `dict[str, Tensor]` or `TypedDict`

adamjstewart:
I'm actually thinking about removing both 'crs' and 'bounds' and returning a new pre-computed 'transform'. Seems more useful for its intended purpose.

hkristen:
> I'm actually thinking about removing both 'crs' and 'bounds' and returning a new pre-computed 'transform'. Seems more useful for its intended purpose.

To clarify: 
1. A rasterio transform would be computed for each sample and returned as flattened Tensor[6]?
2. For CRS, would we still grab it from dataset.crs property, meaning we would only do single-CRS dataset inference?

Questions answered in https://github.com/torchgeo/torchgeo/pull/3140 @adamjstewart was faster :P

adamjstewart:
Correct. We can figure out multi-CRS datasets later when the time comes. EPSG can easily be returned as a tensor, anything else would require a dataset CRS dict for lookup.

adamjstewart:
Seems like @calebrob6 and @isaaccorley prefer this PR over #3148.

@hkristen you're the one working on stitching together predictions. Do you have any preference between the two?

hkristen:
+1 for #3138 
I actually built my current draft on it and it does what I need :)

PR #3139, opened 2025-11-26
Point datasets: return keypoints

adamjstewart:
'bounds' should be reserved for the bounding box of the query, and is needed for stitching together predictions.

'keypoints' is more appropriate for point-based data, following [Kornia's naming scheme](https://github.com/kornia/kornia/blob/v0.8.2/kornia/constants.py#L149).

### DIFF
--- a/torchgeo/datasets/gbif.py
+++ b/torchgeo/datasets/gbif.py
@@ -98,7 +98,7 @@ class GBIF(GeoDataset):
                 f'query: {query} not found in index with bounds: {self.bounds}'
             )
 
-        keypoints = torch.tensor(index.get_coordinates().values)
+        keypoints = torch.tensor(index.get_coordinates().values, dtype=torch.float32)
         sample = {'crs': self.crs, 'bounds': query, 'keypoints': keypoints}
 
         return sample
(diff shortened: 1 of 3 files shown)


torchgeo/datasets/gbif.py
```diff
@@ -124,28 +125,13 @@ def plot(
...
-        scatter = ax.scatter(longitudes, latitudes, c=timestamps, edgecolors='black')
-
-        # Create a formatter function
-        def format_date(x: float, pos: int | None = None) -> str:
-            # Convert timestamp to datetime
-            return datetime.fromtimestamp(x).strftime('%Y-%m-%d')
-
-        # Add a colorbar
-        cbar = fig.colorbar(scatter, ax=ax, pad=0.04)
-        cbar.set_label('Observed Timestamp', rotation=90, labelpad=-100, va='center')
```
isaaccorley, line 145: Do we want to remove the timestamps in plots?

adamjstewart: No way to get that without adding a new return value for t.

isaaccorley: Ahh okay I see what you're saying

isaaccorley (dismissed):
I'm good with this but I do think, similarly to what copilot suggested, that we should explicitly cast keypoints to floats.

adamjstewart:
They are guaranteed to be floats (https://geopandas.org/en/latest/docs/reference/api/geopandas.GeoSeries.get_coordinates.html), but I don't mind being explicit.

isaaccorley (approved):

PR #3142, opened 2025-11-27
Add LEVIRCD100 dataset and datamodule for tutorials

hkristen:
Selected 100 image pairs from LEVIRCD and resampled to 256x256 to achieve a small enough dataset for the Change Detection tutorial. 

Image pairs are stratified by change percentage, then randomly split into train (60),
val (20), and test (20):
* 30 samples with 0.5-2% change
* 40 samples with 2-5% change
* 20 samples with 5-10% change
* 10 samples with 10-25% change

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Clone repo
-        uses: actions/checkout@v5.0.1
+        uses: actions/checkout@v6.0.0
       - name: Set up python
         uses: actions/setup-python@v6.0.0
         with:
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Clone repo
-        uses: actions/checkout@v5.0.1
+        uses: actions/checkout@v6.0.0
       - name: Add label
         uses: actions/labeler@v6.0.1
         with:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Clone repo
-        uses: actions/checkout@v5.0.1
+        uses: actions/checkout@v6.0.0
       - name: Set up python
         id: setup-python
         uses: actions/setup-python@v6.0.0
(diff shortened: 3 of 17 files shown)


adamjstewart:
Is the license of LEVIR-CD known? I don't see it documented, which means technically no one is allowed to even use it, let alone modify and redistribute.

hkristen:
From the author's page:

>Usage
>The use of the images from Google Earth must respect the Google Earth terms of use. All images and annotations in LEVIR-CD can only be used for academic purposes, but are prohibited for any commercial use.

https://justchenhao.github.io/LEVIR/

Ofc I didn't think of that.

adamjstewart:
This is approximately equivalent to CC-BY-NC-4.0, although the exact license isn't described. I think we can use this description anyway. Not a deal breaker, although CC-BY-4.0 would be better. 

Only other thing worth considering is if we want binary or multiclass. But LEVIR-CD is well-known and can be automatically downloaded, unlike xBD. Is LEVIR-CD a good enough motivation? I wonder if disaster mapping or wildfire mapping is more interesting than new buildings.

hkristen:
I went for binary to get a decent model out in a short training time on small images, multiclass seems unlikely that this converges fast.

Disaster or wildfire are for sure more interesting! Are there any in Torchgeo that you think would be easy enough for a short training?

adamjstewart:
Only the ones I listed in my release notes: OSCD, xBD (multiclass, restrictive license), LEVIR-CD, LEVIR-CD+, ChaBuD, CaBuAr, BRIGHT (multiclass). I'm definitely not opposed to LEVIR-CD if it works.

hkristen:
> Only the ones I listed in my release notes: OSCD, xBD (multiclass, restrictive license), LEVIR-CD, LEVIR-CD+, ChaBuD, CaBuAr, BRIGHT (multiclass). I'm definitely not opposed to LEVIR-CD if it works.

What about:
- First part: Use LEVIRCD100 as a quick & fast starting point for the CD tutorial. 
- Second part: more interesting & challenging with a subset of the BRIGHT dataset (multimodal, multiclass) = BRIGHT100 - I need to see how we can make this work in a tutorial setting with a small enough dataset  (need to test how training works when I downsample to 256x256 from 1024x1024)

adamjstewart:
Completely fine with starting with LEVIR-CD, but would rather transition to GeoDatasets for the second half instead of BRIGHT.

adamjstewart:
Inquired about the license over email, let's see if we get a response.

hkristen:
Closed due to licensing issues.

PR #3145, opened 2025-11-28
add CloudCoverDetectionDataModule

robmarkcole:
Add datamodule for [CloudCoverDetection](https://github.com/torchgeo/torchgeo/blob/9289a5550818c662638085f4653b41d739f856f7/torchgeo/datasets/cloud_cover.py#L23C7-L23C26)

Do we want to normalise? e.g. 
```
    mean = torch.tensor(0)
    std = torch.tensor(10000)
```

It trains as-is however:
image

### DIFF
--- a/requirements/style.txt
+++ b/requirements/style.txt
@@ -1,6 +1,6 @@
 # style
 mypy==1.19.0
-pandas-stubs==2.3.2.250926
+pandas-stubs==2.3.3.251201
 ruff==0.14.8
 types-requests==2.32.4.20250913
 types-shapely==2.1.0.20250917
--- a/torchgeo/datasets/cropharvest.py
+++ b/torchgeo/datasets/cropharvest.py
@@ -126,9 +126,9 @@ class CropHarvest(NonGeoDataset):
 
         self.files = self._load_features(self.root)
         self.labels = self._load_labels(self.root)
-        self.classes = self.labels['properties.label'].unique()
-        self.classes = self.classes[self.classes != np.array(None)]
-        self.classes = np.insert(self.classes, 0, ['None', 'Other'])
+        classes = self.labels['properties.label'].unique()
+        classes = classes[classes != np.array(None)]
+        self.classes = np.insert(classes, 0, ['None', 'Other'])
 
     def __getitem__(self, index: int) -> dict[str, Tensor]:
         """Return an index within the dataset.
--- a/torchgeo/datasets/digital_typhoon.py
+++ b/torchgeo/datasets/digital_typhoon.py
@@ -176,7 +176,7 @@ class DigitalTyphoon(NonGeoDataset):
 
         # Compute the hour difference between the first and second entry
         self.aux_df['hour_diff_to_next'] = (
-            self.aux_df.groupby('id')['datetime']
+            self.aux_df.groupby('id')['datetime']  # type: ignore[attr-defined]
             .shift(-1)
             .sub(self.aux_df['datetime'])
             .abs()
@@ -184,7 +184,7 @@ class DigitalTyphoon(NonGeoDataset):
             / 3600
         )
 
-        self.aux_df['hour_diff'] = self.aux_df['hour_diff_consecutive'].combine_first(  # type: ignore[unreachable]
+        self.aux_df['hour_diff'] = self.aux_df['hour_diff_consecutive'].combine_first(
             self.aux_df['hour_diff_to_next']
         )
         self.aux_df.drop(


tests/datamodules/test_cloud_cover.py
adamjstewart, line 1: My goal is to remove everything in `tests/datamodules`. Instead, we should add a `tests/conf/cloud_cover.yaml` and test this via SemanticSegmentationTask.

torchgeo/datamodules/cloud_cover.py
```diff
@@ -0,0 +1,60 @@
...
+        Args:
+            batch_size: Size of each mini-batch.
+            num_workers: Number of workers for parallel data loading.
+            val_split_pct: Percentage of the training data to reserve for validation.
+            **kwargs: Additional keyword arguments passed to
+                :class:`~torchgeo.datasets.CloudCoverDetection`.
+        """
+        if not 0 < val_split_pct < 1:
+            msg = 'val_split_pct must be between 0 and 1.'
+            raise ValueError(msg)
```
adamjstewart: This will be tricky to get coverage of, can probably just remove

adamjstewart:
We definitely want to normalize to the 0-1 range. The default is to divide by 255, but if the images have larger values, we can divide by a larger number.

tests/conf/cloud_cover.yaml
```diff
@@ -0,0 +1,16 @@
...
+    task: 'binary'
+    loss: 'bce'
+    num_filters: 1
+data:
+  class_path: CloudCoverDetectionDataModule
+  init_args:
+    batch_size: 1
+    val_split_pct: 0.2
+  dict_kwargs:
+    root: 'tests/data/cloud_cover_detection'
```
adamjstewart: This is the same as `tests/data/ref_cloud_cover_detection_challenge_v1`? We can rename this directory if it's confusing.

robmarkcole:
@adamjstewart there is a bit of a mix of naming amongst `cloud_cover.py`, `CloudCoverDetectionDataModule` and cloud-cover-detection-challenge in  https://staging.source.coop/radiantearth/cloud-cover-detection-challenge I feel the latter is the most unique and could be the best choice to adopt?

tests/conf/cloud_cover.yaml
robmarkcole: rename to `cloud_cover_detection_challenge` ?

adamjstewart:
Perhaps @klwetstone has opinions on the name of the dataset since she created https://source.coop/radiantearth/cloud-cover-detection-challenge.

I would like to differentiate between the name of the competition and the name of the dataset. "Sentinel-2 Cloud Cover Segmentation" is probably correct, although a bit long.

tests/conf/cloud_cover.yaml
adamjstewart: Let's finalize the name of the dataset and dataset file first. For now, you can use `tests/data/ref_cloud_cover_detection_challenge_v1` to get the tests to pass without any file duplication.

robmarkcole:
Sentinel-2 Cloud Cover Segmentation could apply to other datasets, e.g. Cloudsen12. I assume `ref_cloud_cover_detection_challenge_v1` is a truncation of `Radiant Earth Foundation Sentinel-2 Cloud Cover Segmentation Dataset (Version 1)` although the use of `ref` is a bit confusing. `radian-earth-cloud-detection-v1` would make sense to me

robmarkcole:
the dataset expects `train/test` folders but the current test approach has them as `public/private` - @adamjstewart which approach do you want?

adamjstewart:
What does the real data look like? Did they change the directory hierarchy again? The test data should always match the real data.

robmarkcole:
The real data uses `test/train` - not sure if this has changed?
image

The testing data
image

adamjstewart:
Must have changed. Can you fix this in a separate PR so we can backport the fix to 0.8.1? This PR introduces a new datamodule, so it has to wait until 0.9.0.

robmarkcole:
@adamjstewart we can [see that](https://github.com/torchgeo/torchgeo/blob/109d83d4ba7cad4256fce74deea047641e6b88d1/torchgeo/datasets/cloud_cover.py#L168) although the [remote dir](https://staging.source.coop/radiantearth/cloud-cover-detection-challenge/final) is called `public` we sync to a `train` dir - I assume you want to keep this behaviour? This does mean if we do 
```
data:
  class_path: CloudCoverDetectionDataModule
  init_args:
    batch_size: 1
    val_split_pct: 0.2
  dict_kwargs:
    root: 'tests/data/ref_cloud_cover_detection_challenge_v1'
``` 
the paths are incorrect, since the data has not been syncd and renamed

adamjstewart:
Oh, I think I understand now. I'm fine with either. The `split` parameter should take in 'train' or 'test', but I don't care what the download directory is called. It might be simpler if we keep the public/private names in case someone manually downloads and wants to use it in TorchGeo.

robmarkcole:
@adamjstewart agreed - you want a seperate PR for that or do here?

adamjstewart:
Hmm, maybe separate.

robmarkcole:
Pending https://github.com/torchgeo/torchgeo/pull/3145

docs/api/datamodules.rst
```diff
@@ -82,6 +82,11 @@ ChaBuD
 
 .. autoclass:: ChaBuDDataModule
 
+Cloud Cover Detection
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
adamjstewart:
```suggestion
Cloud Cover Detection
^^^^^^^^^^^^^^^^^^^^^
```

tests/conf/cloud_cover.yaml
```diff
@@ -0,0 +1,16 @@
...
+    backbone: 'resnet18'
+    in_channels: 4
+    task: 'binary'
+    loss: 'bce'
+    num_filters: 1
+data:
+  class_path: CloudCoverDetectionDataModule
+  init_args:
+    batch_size: 1
+    val_split_pct: 0.2
```
adamjstewart: Could also use 0.5 to guarantee that val is not empty, we don't have very many fake images compared to the real dataset.

torchgeo/datamodules/cloud_cover.py
```diff
@@ -0,0 +1,58 @@
...
+
+        Args:
+            batch_size: Size of each mini-batch.
+            num_workers: Number of workers for parallel data loading.
+            val_split_pct: Percentage of the training data to reserve for validation.
+            **kwargs: Additional keyword arguments passed to
+                :class:`~torchgeo.datasets.CloudCoverDetection`.
+        """
+        self.mean = torch.tensor(0.0)
+        self.std = torch.tensor(10000.0)
```
adamjstewart: Can we put these at the class level like the base class?

robmarkcole:
@adamjstewart the masks contain raw uint16 values `([2732, …])` - might be why I struggled to train a model. Addressed now

adamjstewart:
Hmm, I just checked `public/train_labels/*.tif` and all files are in the range [0, 1]. Which files did you encounter 2732 in?

robmarkcole:
@adamjstewart thanks for checking, the issue was in the mock dataset

adamjstewart:
Still curious about train accuracy. Want to run a quick training run to see if accuracy has improved with normalization?

robmarkcole:
Batch size 8
```
    model="unet",
    backbone="resnet18",
    weights=True,
    in_channels=4,
    num_classes=2,
    # loss="focal",
    # labels=['background', 'cloud'],
    class_weights=Tensor([1, 10]),
    lr=1e-4,
)
```
Possibly overfitting after around 10 epochs
image

PR #3148, opened 2025-11-29
GeoDataset: remove 'bounds', 'crs'

adamjstewart:
My goal is to ensure that samples returned by datasets are (mostly) of the form `dict[str, Tensor]`. This PR removes 'bounds' and 'crs', which are 90% of the instances of non-Tensor outputs. We may still have a few lists for bboxes remaining.

### Pros

* Compatibility with PyTorch's `default_collate function`
* Fewer instances of dynamic type hints (`typing.Any`)

### Cons

* Backwards-incompatible changes (removing CRS, bounds)

Alternative to #3138
Closes #3138
Closes #3135

### DIFF
--- a/tests/datamodules/test_geo.py
+++ b/tests/datamodules/test_geo.py
@@ -40,7 +40,7 @@ class CustomGeoDataset(GeoDataset):
 
     def __getitem__(self, query: GeoSlice) -> dict[str, Any]:
         image = torch.arange(3 * 2 * 2, dtype=torch.float).view(3, 2, 2)
-        return {'image': image, 'crs': self.index.crs, 'bounds': query}
+        return {'image': image}
 
     def plot(self, *args: Any, **kwargs: Any) -> Figure:
         return plt.figure()
--- a/tests/datasets/test_agb_live_woody_density.py
+++ b/tests/datasets/test_agb_live_woody_density.py
@@ -8,7 +8,6 @@ import matplotlib.pyplot as plt
 import pytest
 import torch
 import torch.nn as nn
-from pyproj import CRS
 from pytest import MonkeyPatch
 
 from torchgeo.datasets import (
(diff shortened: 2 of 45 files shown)


isaaccorley (approved):
This overall lgtm

calebrob6:
Removing this information will break prediction/inference scripts

adamjstewart:
We added this in #3140. The idea is to get the transform from the sample and the CRS from the dataset.

isaaccorley:
Personally I'd prefer to keep the transform and bounds in so a user doesn't have to compute the transform manually. CRS can be grabbed from the dataset I think

adamjstewart:
We just added transform, we're definitely keeping that.

We could convert bounds to a Tensor if needed, but is bounds useful if you already have transform?

Yes, CRS can come from the dataset for now. At the moment, all datasets are reprojected to a single CRS. If that ever changes, different samples could have different CRS, and we would need to figure out how to pass that information through.

calebrob6:
Can you add a test that knows the bounds and asserts them against the bounds you compute with the transform?

isaaccorley:
The issue is that in the case where someone resizes the patch, we don't have the original height and width in pixels anymore. This breaks the transform.

You can recompute the new transform using the original bounds and the new height and width like this:

```python
from rasterio.transform import from_bounds

new_transform = from_bounds(minx, miny, maxx, maxy, new_width, new_height)
```

This doesn't work though if the image had a) non-orthogonal pixels, b) non-axis-aligned grid, or c) negative pixel sizes (flipped axes).

Imo it's better to keep both, it's not really expensive to keep the bounds, it is only 4 float values

adamjstewart:
It also doesn't work if the image footprint changed (e.g., a U-Net with a smaller output than input). But I'm fine with keeping bounds if needed.

isaaccorley:
Worst case we can also pass the original input h/w as a tensor as well just for consistency.

adamjstewart:
Does #3138 satisfy your requests if I resolve all merge conflicts?

isaaccorley:
Yup that PR lgtm

PR #3150, opened 2025-11-29
torchgeo.models.Unet: type is nn.Module

adamjstewart:
Flagged by `ty check`.

`smp.create_model` returns `nn.Module`, not `Unet`.

Not sure how I feel about this one, as `smp.create_model` technically returns `Any` (it is dynamic), although this any is always a subclass of `nn.Module`. Unclear if this is a bug in torchgeo or in SMP.

### DIFF
--- a/torchgeo/models/unet.py
+++ b/torchgeo/models/unet.py
@@ -209,7 +209,7 @@ def unet(
     else:
         kwargs['classes'] = 1 if classes is None else classes
 
-    model = smp.create_model(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


adamjstewart:
Weird that ty is able to resolve the type but mypy isn't. It's pretty straightforward: https://github.com/qubvel-org/segmentation_models.pytorch/blob/v0.5.0/segmentation_models_pytorch/__init__.py#L50

EDIT: I think there are settings to control this. By default, ty tries to infer the type even when `py.typed` isn't available whereas mypy doesn't.

PR #3152, opened 2025-11-29
convert_poly_coords: input/output type is Geometry

adamjstewart:
Flagged by `ty check`.

`shapely.geometry.shape` is a function, the correct type is `shapely.Geometry`.


adamjstewart:
Adding `types-shapely` as a dep would have caught this, will do in a separate PR.

PR #3153, opened 2025-11-29
mypy: only check torchgeo and tests directories

adamjstewart:
I'm sick of erroneous type hint warnings in `experiments` and `docs`. I propose we only officially type `torchgeo` and `tests` (and maybe our tutorials once we switch to ty, mypy doesn't support jupyter notebooks).

Related to #3070.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -134,8 +134,8 @@ models = [
     "microsoft-aurora>=1.6",
 ]
 style = [
-    # mypy 0.900+ required for pyproject.toml support
-    "mypy>=0.900",
+    # mypy 1.16+ required for exclude_gitignore configuration
+    "mypy>=1.16",
     # ruff 0.9+ required for 2025 style guide
     "ruff>=0.9",
 ]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -134,8 +134,8 @@ models = [
     "microsoft-aurora>=1.6",
 ]
 style = [
-    # mypy 0.900+ required for pyproject.toml support
-    "mypy>=0.900",
+    # mypy 1.16+ required for exclude_gitignore configuration
+    "mypy>=1.16",
```
adamjstewart, line 138: This is relatively new, I could be convinced to remove this change...

pyproject.toml
```diff
@@ -174,9 +174,11 @@ show_missing = true
 # https://mypy.readthedocs.io/en/stable/config_file.html
 [tool.mypy]
 # Import discovery
+packages = ["torchgeo", "tests"]
+exclude = ["^tests/data"]
```
adamjstewart, line 178: We were inadvertently ignoring these in the previous `exclude` because we forgot a `^`. As a result, these files are quite far from passing type checks. Let's continue to ignore them for now.

isaaccorley:
I approve this motion

isaaccorley (approved):

PR #3155, opened 2025-11-30
Fix pandas type hints

adamjstewart:
Mostly required variable renames or casts to avoid dynamic typing, but a couple real bugs discovered too!


torchgeo/datasets/cropharvest.py
```diff
@@ -228,11 +228,11 @@ def _load_label(self, idx: str, dataset: str) -> Tensor:
             (self.labels['properties.index'] == index)
             & (self.labels['properties.dataset'] == dataset)
         ]
-        row = row.to_dict(orient='records')[0]
```
adamjstewart, line 231: Dynamic typing: pd.Series -> dict

torchgeo/datasets/digital_typhoon.py
```diff
@@ -176,7 +176,7 @@ def __init__(
 
         # Compute the hour difference between the first and second entry
         self.aux_df['hour_diff_to_next'] = (
-            self.aux_df.groupby('id')['datetime']
+            self.aux_df.groupby('id')['datetime']  # type: ignore[attr-defined]
```
adamjstewart: I do not understand this one at all:
```
torchgeo/datasets/digital_typhoon.py:179: error: "Never" has no attribute "total_seconds"  [attr-defined]
                self.aux_df.groupby('id')['datetime']
                ^
```
It seems to think that `.dt` will never return, but it does. Could be a bug in the type hints, or dynamic typing.

torchgeo/datasets/everwatch.py
```diff
@@ -133,7 +134,8 @@ def __len__(self) -> int:
         Returns:
             length of the dataset
         """
-        return len(self.annot_df.index.levels[0])
+        index = cast(pd.MultiIndex, self.annot_df.index)
```
adamjstewart, line 137: Mypy cannot understand that our DataFrame went from Index to MultiIndex. I tried to find ways to avoid this dynamic typing, but I'm not even sure how the pandas type hints differentiate between these two types of DataFrames anyway.

torchgeo/datasets/gbif.py
```diff
@@ -61,7 +61,7 @@ def __init__(self, root: Path = 'data') -> None:
         # Read tab-delimited CSV file
         usecols = ['decimalLatitude', 'decimalLongitude', 'day', 'month', 'year']
         dtype = {'day': str, 'month': str, 'year': str}
-        df = pd.read_table(files[0], usecols=usecols, dtype=dtype)
+        df = pd.read_table(files[0], usecols=usecols, dtype=dtype)  # type: ignore[arg-type]
```
adamjstewart, line 64: This one makes no sense to me. `dtype` has many different possible values, including `Mapping[Hashable, type[str]]`. Our variable is `dict[str, type[str]]`, which should be a subset of this from what I can tell...

torchgeo/datasets/satlas.py
```diff
@@ -578,16 +578,16 @@ def __init__(
...
         )
         self.image_times = pd.read_json(
             os.path.join(root, 'metadata', f'{image_times}.json'), typ='series'
         )
 
         self.split.columns = ['col', 'row']
-        self.good_images.columns = ['col', 'row', 'directory']
-        self.good_images = self.good_images.groupby(['col', 'row'])
+        good_images_df.columns = ['col', 'row', 'directory']
+        self.good_images = good_images_df.groupby(['col', 'row'])
```
adamjstewart, line 590: Dynamic typing: DataFrame -> DataFrameGroupBy

torchgeo/datasets/splits.py
```diff
@@ -281,7 +281,8 @@ def roi_split(dataset: GeoDataset, rois: Sequence[Polygon]) -> list[GeoDataset]:
 
 
 def time_series_split(
-    dataset: GeoDataset, lengths: Sequence[float | pd.Timedelta | pd.Interval]
+    dataset: GeoDataset,
+    lengths: Sequence[float] | Sequence[Timedelta] | Sequence[Interval[Timestamp]],
```
adamjstewart: I think this is actually a bug. Previously, we allowed any `Sequence[a, b, c]`, which I think allows mixing a, b, and c in a single list. In fact, we only support pure lists of a, b, or c.

torchgeo/datasets/splits.py
```diff
@@ -295,11 +296,12 @@ def time_series_split(
 
     .. versionadded:: 0.5
     """
-    x, y, t = dataset.bounds
+    _, _, t = dataset.bounds
 
     totalt = t.stop - t.start
 
     if all(isinstance(x, int | float) for x in lengths):
+        lengths = cast(Sequence[float], lengths)
```
adamjstewart, line 304: Mypy cannot understand the guarantees provided by the previous line, need to give it a helping hand

torchgeo/datasets/splits.py
```diff
@@ -310,46 +312,44 @@ def time_series_split(
...
-    if all(isinstance(x, pd.Timedelta) for x in lengths):
+    if all(isinstance(x, Timedelta) for x in lengths):
         lengths = [
-            pd.Interval(t.start + offset - length, t.start + offset, closed='neither')
+            Interval(t.start + offset - length, t.start + offset, closed='neither')
             for offset, length in zip(accumulate(lengths), lengths)
         ]
 
-    lengths = cast(Sequence[pd.Interval], lengths)
+    lengths = cast(Sequence[Interval], lengths)  # type: ignore[type-arg]
```
adamjstewart: This one is really unsatisfying. `pd.Interval` is a generic, and can have many types, just like a tuple. However, our type hints are coming from stubs, not from pandas itself. As such, `pd.Interval` does not support subscript notation, even though the type stubs require its use. So we have to `from __future__ import annotations` for function signatures and ignore for runtime code.

torchgeo/datasets/splits.py
```diff
@@ -310,46 +312,44 @@ def time_series_split(
...
+        offset = Timedelta(0) if i == len(lengths) - 1 else Timedelta(1, unit='us')
 
         if start < t.start or end > t.stop:
             raise ValueError(
                 "Pairs of timestamps in lengths can't be out of dataset's time bounds."
             )
 
         for other in lengths:
-            x = other.left
-            y = other.right
```
adamjstewart, line 339: Dynamic typing: slice[Any, Any, Any] -> Timestamp

torchgeo/datasets/utils.py
```diff
@@ -317,44 +316,41 @@ def disambiguate_timestamp(
     Returns:
         (mint, maxt) tuple for indexing
     """
-    if not isinstance(date_str, str):
-        return pd.NaT, pd.NaT
```
adamjstewart, line 321: Another bug, these aren't a subclass of datetime or Timestamp. I remember them being needed for one of the point datasets, but let's see if the tests pass without this.

torchgeo/datasets/splits.py
adamjstewart: Hmm, RtD doesn't like this. May revert and open an issue.

adamjstewart: Could be related to https://github.com/sphinx-doc/sphinx/issues/13731, who knows. Can't open an issue until we reproduce this with a modern Sphinx version.

isaaccorley (approved):

PR #3158, opened 2025-12-01
Cloud Cover Detection: don't rename downloaded directories

robmarkcole:
We now keep the `public/private` folder naming from the source dataset, to be compatible with a dataset not downloaded via torchgeo. 

Linked to https://github.com/torchgeo/torchgeo/pull/3145#issuecomment-3596178617

### DIFF
--- a/tests/data/ref_cloud_cover_detection_challenge_v1/data.py
+++ b/tests/data/ref_cloud_cover_detection_challenge_v1/data.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python3
+
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+
+import os
+
+import numpy as np
+import rasterio
+from rasterio import Affine
+from rasterio.crs import CRS
+
+SIZE = 2
+DTYPE = np.uint16
+
+np.random.seed(0)
+
+splits = {'train': 'public', 'test': 'private'}
+chip_ids = ['aaaa']
+all_bands = ['B02', 'B03', 'B04', 'B08']
+profile = {
+    'driver': 'GTiff',
+    'dtype': DTYPE,
+    'width': SIZE,
+    'height': SIZE,
+    'count': 1,
+    'crs': CRS.from_epsg(32753),
+    'transform': Affine(10.0, 0.0, 777760.0, 0.0, -10.0, 6735270.0),
+}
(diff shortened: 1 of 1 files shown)


adamjstewart:
Hmm, not sure about the milestone. Technically, the API didn't change, but anyone who downloaded the dataset previously won't be able to use it anymore. Maybe 0.9.0 would be better?

robmarkcole:
It is a breaking change, so that makes sense. Does this need a comment in the code then?

adamjstewart:
Nah, I don't think so. It's only a breaking change in the sense that `download=True` will redownload instead of using an existing copy. Of course, `download=False` users may be confused.

robmarkcole:
OK this is GTG then

tests/data/ref_cloud_cover_detection_challenge_v1/data.py
adamjstewart, line 1: I ran this script and nothing changed. Is that the intention?

adamjstewart:
FWIW, I tested a real download and this works as expected. Only remaining question is about the data.py.

tests/data/ref_cloud_cover_detection_challenge_v1/data.py
robmarkcole: Nothing should change, what were you anticipating?

adamjstewart: Not sure. Should we undo the changes to this file if they don't do anything?

robmarkcole: These changes are to work with the dataset format as it is natively formatted - they are required

robmarkcole: Or are you saying it is no longer required as we use the native layout?

adamjstewart: `data.py` is used to generate our fake testing data. On main or on your branch, running this script results in the same files. So it doesn't seem like anything has changed in this file, so I'm confused why you made the changes.

robmarkcole:
@adamjstewart data.py removed

adamjstewart:
You should revert the changes, not delete the file.

adamjstewart:
Can you undo the file permission changes?

robmarkcole:
@adamjstewart done

Issue #3160, opened 2025-12-01
Vector Dataset Backends


isaaccorley:
### Summary

We currently use geopandas as our backend for reading and storing geometry data. While geopandas is great, it's known to be quite slow, compared to DuckDB and more recently SedonaDB, when performing spatial filtering operations on a single node when the number of geometries grows large.

[SedonaDB](https://sedona.apache.org/sedonadb/latest/) performed a benchmark comparison using [SpatialBench](https://sedona.apache.org/spatialbench/single-node-benchmarks/).

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/9a19481f-65e4-433f-bfa4-2d757217b961" />

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/a1bccfd7-be7c-424e-a584-482deba532df" />

We should support more than just geopandas as a backend. I propose we implement a `torchgeo.datasets.SedonaDBDataset` for loading vector dataset into SedonaDB dataframes and performing spatial filtering on them.

cc: @jiayuasu @rbavery @paleolimbot

jiayuasu:
Looking forward to it!!

adamjstewart:
> We should support more than just geopandas as a backend. I propose we implement a `torchgeo.datasets.SedonaDBDataset` for loading vector dataset into SedonaDB dataframes and performing spatial filtering on them.

To clarify, we use geopandas for many things:

* VectorDataset: reading vector files from disk
* GeoDataset: reading some custom datasets (e.g., point datasets)
* GeoDataset: storing spatiotemporal metadata
* GeoSampler: determining where to query from
* IntersectionDataset: computing spatiotemporal intersection
* UnionDataset: computing spatiotemporal union
* Splitters: computing splits based on bboxes, roi, toi, etc.

Depending on what you are proposing:

1. Add support for reading from a SedonaDB and converting it to our geopandas representation 👍 
2. Replace geopandas with SedonaDB 👎
3. Support multiple backends (R-tree, geopandas, Sedona, etc.) for all of the above 👎 
4. A third kind of base class (GeoDataset, NonGeoDataset, SedonaDBDataset) 👎 

[...]

PR #3161, opened 2025-12-01
SedonaDB Dataset

isaaccorley:
### Summary

See reasoning here #3160 

This PR implements the following:

- abstracts the spatial filtering ops in `VectorDataset` into a method `VectorDataset.filter_index` which can be overridden for new backends.
- implements `torchgeo.datasets.SedonaDBDataset`

### Benchmarking

I added a benchmarking script that simply iteratively queries the `filter_index` method of both methods and these are the results. The data I used is the Washington state buildings [here](https://github.com/microsoft/USBuildingFootprints) from the Microsoft Open Buildings dataset (converted to parquet):

**VectorDataset:**
Initialization time: 2.27 seconds
Filter time (50 slices): 105.54 seconds
Time per slice: 2.11 seconds
Total geometries found: 21,235
Geometries per second: **201.20**

**SedonaDBDataset Performance:**
Initialization time: 2.06 seconds
Filter time (50 slices): 12.89 seconds
Time per slice: 0.26 seconds
Total geometries found: 21,235
Geometries per second: **1,647.26**

**Speedup: 8.19x**

SedonaDBDataset is about 8.19x faster than VectorDataset for filtering operations, processing ~1,647 geometries/second vs ~201 geometries/second. Both found the same 21,235 geometries, confirming correctness.

### TODO

Possibly need to implement separate SedonaDBVectorDataset and SedonaDBRasterDataset that can use IntersectionDataset and UnionDataset (could be in a follow-up PR though)

cc: @jiayuasu @rbavery @paleolimbot

## Summary by CodeRabbit

* **New Features**
  * Added SedonaDB-backed vector dataset for advanced spatial data querying and filtering
  * Enhanced vector dataset with improved geometric querying capabilities
  
* **Dependencies**
  * Added optional Apache Sedona dependency for datasets module

* **Tests**
  * Added comprehensive test suite for SedonaDB dataset functionality across multiple tasks and edge cases


torchgeo/datasets/sedonadb.py
```diff
@@ -0,0 +1,173 @@
...
+        t_stop_ts = int(pd.Timestamp(t.stop).value)
+
+        self.index_sedona.to_view('index_df', overwrite=True)
+
+        filtered_index_df = self._sd.sql(
+            f"""
+            SELECT * FROM index_df
+            WHERE datetime_start <= {t_stop_ts}
+            AND datetime_end >= {t_start_ts}
+            AND ST_Intersects(geometry, ST_SetSRID(ST_GeomFromWKT('{query_wkt}'), {epsg}))
```
paleolimbot: You can use `ST_SetCrs(..., '{self.crs.to_json()}'), which will work even of there's no EPSG code.

torchgeo/datasets/sedonadb.py
```diff
@@ -0,0 +1,173 @@
...
+        .. note::
+
+        This dataset requires the following additional library to be installed:
+
+        * `apache-sedona[db] <https://pypi.org/project/apache-sedona/>`_ to load the dataset
+        """
+        sedona_db = lazy_import('sedona.db')
+        super().__init__(paths, crs, res, transforms, label_name, task, layer)
+
+        self._sd = sedona_db.connect()
```
paleolimbot: To really get the benefit of SedonaDB here, I think you may want to use `sd.read_parquet()` (if `paths` points to 1 or more Parquet files) or `sd.read_pyogrio()` (otherwise). Perhaps `index` can be a property? (in other words, to really get a benefit here you probably need to defer the materialization of the index as long as possible).

torchgeo/datasets/sedonadb.py
```diff
@@ -0,0 +1,173 @@
...
+        if filtered_index_df.count() == 0:
+            raise IndexError(
+                f'query: {query} not found in index with bounds: {self.bounds}'
+            )
+
+        filtered_index_gdf = filtered_index_df.to_pandas()
+
+        if t.step != 1:
+            filtered_index_gdf = filtered_index_gdf.iloc[:: t.step]
+            if len(filtered_index_gdf) == 0:
```
paleolimbot: Perhaps the initial creation of the index could be like:

```python
self._sd.read_pyogrio(files).to_view("index_raw", overwrite=True);
self._sd.sql("""SELECT row_number() as i, t, geometry FROM (SELECT * FROM index_raw ORDER BY t)"""
```

...which would possibly let you do something like `self._sd.sql("""SELECT * FROM filtered_index WHERE i % {t.step} = 0""").to_view("filtered_index", overwrite=True)` here.

torchgeo/datasets/sedonadb.py
```diff
@@ -0,0 +1,173 @@
...
+            else:
+                filtered_df = self._sd.sql(
+                    f"""
+                    SELECT * FROM temp_df
+                    WHERE ST_Intersects(geometry, ST_SetSRID(ST_GeomFromWKT('{query_wkt}'), {epsg}))
+                    """
+                )
+
+            if filtered_df.count() > 0:
+                filtered_gdf = filtered_df.to_pandas()
```
paleolimbot: You should probably do `to_pandas()` and check the `len()` (what you have here will evaluate the query twice).

pyproject.toml
```diff
@@ -92,6 +92,8 @@ dynamic = ["version"]
 
 [project.optional-dependencies]
 datasets = [
+    # apache-sedona[db] 1.8.0 required for sedona.db support
+    "apache-sedona[db]>=1.8.0",
```
paleolimbot, line 94: You may want just `sedonadb>=0.2.0` here from a developer standpoint (i.e., the number you actually want to control here is `0.2.0` to ensure you're getting the features you need)

paleolimbot (commented):
Cool!

I'm new to TorchGeo so take this all with a grain of salt! The high-level suggestion is that to get the most out of SedonaDB (or any spatial db-based dataset), you may want to do something like:

```python
self._sd.read_pyogrio(files).to_view("index_raw", overwrite=True)
self._sd.sql("""SELECT "{t_field}" as t, "{label_field}" as label, wkb_geometry as geometry FROM index_raw""").to_view("index", overwrite=True)
self._sd.sql("""SELECT row_number() as i, t, label, geometry FROM (SELECT * FROM index ORDER BY t)""").to_view("index", overwrite=True)
```

...or in other words, `self._sd.view("index)` is never materialized until it is requested. Of course, this is a bit of a shift in thinking for most GeoPandas users and so feel free to not do this (I can help with this if you get a basic implementation with tests that I can use to make the expectations of the implementation clear!).

torchgeo/datasets/sedonadb.py
isaaccorley: I refactored to use this. Actually uncovered a bug in TorchGeo where we used `gpd.read_file` everywhere which doesn't let us read parquet files or directories (!!!) cc @adamjstewart

isaaccorley:
@paleolimbot I appreciate the review for creating the index. I did some initial benchmarks and sedonadb ABSOLUTELY SLAPS at spatial intersection filtering (see PR description above) cc: @calebrob6 @adamjstewart

isaaccorley:
This PR is still a WIP at the moment but so far looking verrrry good. Most of the remaining work needs to be modifying the constructor to use sedonadb to create the index. 

Basically the following happens:
- glob all the files and create an index which contains the filepath, datetime, bounds (still needs to be sedonadb-ified)
- in __getitem__, query the index using the slice to find the files of interest, then read and filter the geometries within the files using the query (mostly done but probably could use some optimization)

calebrob6:
Would a "driver" flag in VectorDataset (that could be "geopandas" or "sedonadb") be possible? E.g. as a user if I want to use an existing vectordataset but want the sick gains of sedonadb, it'd be cool if I could just switch somehow.

Also, I like this benchmarking script and am curious how the old backend performs on it.

adamjstewart:
Would like to continue the discussion in #3160 once Isaac returns from paternity leave, but some minor comments on the proposed implementation:

I'm not opposed to supporting multiple backends, but note:

* This logic is not specific to VectorDataset, the same logic could be applied to all GeoDatasets
* As mentioned in #3160, this is only a minor fraction of the places we use geopandas, see #2747 for a full list of locations required to truly make TorchGeo backend-agnostic
* The benchmarking script here is cool, but note that we already have a [benchmarking script](https://torchgeo.readthedocs.io/en/stable/user/contributing.html#i-o-benchmarking) for RasterDataset. We should either port this to IOBench or replace IOBench with this. It would be nice to integrate this with our `torchgeo` script so that it works post-installation
* The benchmarking done here doesn't tell me much about SedonaDB's performance for insertion, intersection, union, caching, pickling, etc. It also doesn't tell me how SedonaDB scales compared to R-tree, shapely, geopandas, or STAC. I also don't know what features Sedona DB supports. We should consider adding SedonaDB to our literature review for TorchGeo 1.0 (currently private git repo, but I can give people access).

More generally, the reason we didn't consider database backends like PostGIS/DuckDB/SedonaDB is my fear that this would require users to know about these technologies, install non-Python deps, and manually set up their own databases. Also, no one suggested them when I asked around during our 6 month backend search. For reference, I tried many times to get the GeoTorchAI unit tests running on my laptop, but this didn't work as I didn't have SedonaDB installed. If this has changed and the setup process is now much easier, we can revisit these, as I expect them to be quite performant depending entirely on I/O speeds. However, note that not all users may have write access on the systems they run TorchGeo on, so we may not be able to switch to a file-based DB as the default.

jiayuasu:
@adamjstewart Hi Adam,

Thank you again for all your contributions to TorchGeo. Since Isaac is currently on leave, I wanted to clarify a few things here.

## SedonaDB and Apache Sedona

[SedonaDB](https://github.com/apache/sedona-db) is a new subproject under [Apache Sedona](https://github.com/apache/sedona), but it is not the same as SedonaSpark, SedonaFlink, or other distributed Sedona engines. SedonaDB is a single machine data processing tool that requires zero installation beyond a simple `pip install apache-sedona[db]`. It is designed for embedded and self contained environments. The wheel files for SedonaDB are available on PyPI at  
https://pypi.org/project/sedonadb/

## SedonaDB and PostGIS

SedonaDB is designed specifically for embedded and self contained use cases. It requires zero database setup and no data ingestion process like PostGIS. It works directly on your data files, identical to the GeoPandas user experience.

PostGIS is more of a transactional database. For analytical workloads such as filtering, joins, unions, and aggregations, SedonaDB is orders of magnitude faster. Because of how large the performance gap is, we did not even include PostGIS in our SpatialBench comparison since it would be an extremely unfair comparison.

## SedonaDB and GeoTorchAI

GeoTorchAI was a research prototype that relied on Apache Sedona through SedonaSpark rather than SedonaDB. I agree with you that SedonaSpark can be complicated to operate in certain environments.

## SedonaDB functionalities

SedonaDB was released in September 2025 and is positioned as a GeoPandas alternative. You can find the full list of supported functions here  
https://sedona.apache.org/sedonadb/latest/reference/sql/

We also conducted a comprehensive benchmark comparing SedonaDB, GeoPandas, and DuckDB using SpatialBench  
https://sedona.apache.org/spatialbench/single-node-benchmarks/

SpatialBench is designed to evaluate geospatial analytical performance across different systems. It examines performance from multiple angles including individual spatial functions such as

* filtering, intersection, and union
* complex and heavy spatial joins.
* automatic query optimization across combined operations

For inserting new rows into a GeoPandas DataFrame, I do not believe GeoPandas currently supports this, and neither do SedonaDB or DuckDB.

Hope this helps clarify things and addresses your concerns.

adamjstewart:
Thanks, this actually helps a lot. So we didn't consider SedonaDB during our search because it didn't exist at the time we did our literature review. Glad to know that it's easier to install than Apache Sedona and doesn't require any setup.

SpatialBench is actually of great interest to me. We have our own benchmarking comparing R-tree, Shapely, Geopandas, and STAC. Would be interesting to add some of those to the comparison, although they all obviously have very different features.

I still need to think about the best way to do this. I really don't want a new SedonaDBDataset, as we would have to duplicate all 30+ existing GeoDataset subclasses to actually take advantage of this. Maybe a `backend='sedonadb'` parameter to GeoDataset and friends would help. Again, there are hundreds of places we use the geopandas representation, not just filtering. We don't have to replace all locations, but if you really want speedups, that may be necessary.

I'll add this to the agenda for our monthly Technical Steering Committee meeting. Not sure if we'll get to it in January or February but I'm more open to this idea now that I understand it better.

isaaccorley:
Part of TorchGeo's API design is to use inheritance by providing base classes. For this reason I don't think it's necessary in this PR to completely redo the backend and add it to every single dataset that inherits from GeoDataset. I was hoping to scope this to just creating an experimental SedonaDBDataset as an optional dependency install that users can start to experiment with. As mentioned, the gains in spatial intersection are quite large and I don't think we should delay these kinds of speedups making it into the library.

It also helps our Wherobots developers to begin to consider optimizations that would help particularly for improving geospatial sampling for ML training workflows.

adamjstewart:
Well this won't make it into a release for quite some time regardless of if we merge today or in a few months. I want to speed up the release cycle after 1.0, but at the moment we're busy breaking GeoDataset and GeoSampler to add time series support. These features need time to test and mature before making it into a release, especially because they are backwards-incompatible.

isaaccorley:
> Well this won't make it into a release for quite some time regardless of if we merge today or in a few months. I want to speed up the release cycle after 1.0, 

We should discuss this at our monthly meeting since it seems you are prescribing a release schedule without input from the rest of the maintainers. My opinion is that we should really increase our release schedule frequency particularly for certain types of features. I imagine others would agree with this as well. Adding a new UNet model weights for example should not take 6 months to become available from PyPi. 1.0 or not isn't really relevant to this PR.

> but at the moment we're busy breaking GeoDataset and GeoSampler to add time series support. These features need time to test and mature before making it into a release, especially because they are backwards-incompatible.

These types of large breaking changes over multiple PRs should consider being moved to a `dev` or `time-series` branch that can get fully merged at some point in the future and not be pushed directly to main. These features don't really have anything to do with the SedonaDBDataset in this PR considering what's currently on main and shouldn't hold up any other features before they are merged

My overall 2c is that this features like this should get merged sooner than later instead of being stalled by reasons that aren't relevant to what's proposed in the PR.

torchgeo/datasets/sedonadb.py
```diff
@@ -0,0 +1,124 @@
...
+        for filepath in index_df.filepath:
+            if str(filepath).endswith('.parquet'):
+                source_df = sd.read_parquet(filepath)
+                src_crs = gpd.read_parquet(filepath).crs
+            else:
+                options: dict[str, str] = {}
+                if self.layer is not None:
+                    options['layer'] = str(self.layer)
+                source_df = sd.read_pyogrio(filepath, options=options)
+                src_crs = gpd.read_file(filepath, layer=self.layer).crs
```
paleolimbot: We should probably have a `.crs` accessor in SedonaDB's DataFrame or better expose our "primary" geometry column guessing behaviour. In any case, I think you can avoid reading the entire file using GeoPandas just to get the CRS.

```suggestion
                geometry_column_indices = df.schema.geometry_column_indices
                if len(geometry_column_indices) != 1:
                    raise ValueError("Source must have exactly one geometry column")
                src_crs_obj = df.schema.field(geometry_column_indices[0]).type.crs
                src_crs = pyproj.CRS(src_crs_obj)
```

torchgeo/datasets/sedonadb.py
```diff
@@ -0,0 +1,124 @@
...
+
+            src_crs = pyproj.CRS.from_user_input(src_crs or self.crs)
+            transformer = pyproj.Transformer.from_crs(self.crs, src_crs, always_xy=True)
+            (minx, miny) = transformer.transform(x.start, y.start)
+            (maxx, maxy) = transformer.transform(x.stop, y.stop)
+            query_wkt = shapely.to_wkt(shapely.box(minx, miny, maxx, maxy))
```
paleolimbot, line 92: I am not sure shapely provides an easy way to do this, but the best practice for transforming a bbox I believe is to densify the edges, transform, and then leave the transformed polygon (i.e., don't bbox the transformed values, which may have been rotated. I think shapely/GEOS' densify is parameterized on source CRS units which is tricky to make work for generic input.

torchgeo/datasets/sedonadb.py
```diff
@@ -0,0 +1,124 @@
...
+            src_crs_json = src_crs.to_json().replace("'", "''")
+            geometry_col = (
+                'wkb_geometry' if 'wkb_geometry' in source_df.columns else 'geometry'
+            )
```
paleolimbot:
```suggestion
            geometry_col = source_df.schema.field(geometry_column_indices[0]).name
```

I think when reading a .gpkg this is "geom" (and when reading a Parquet it can be anything)

torchgeo/datasets/sedonadb.py
```diff
@@ -0,0 +1,124 @@
...
+            filtered_df = sd.sql(
+                f"""
+                SELECT {label_select} {geometry_col} as geometry
+                FROM temp_df
+                WHERE ST_Intersects(
+                    ST_SetCrs({geometry_col}, '{src_crs_json}'),
+                    ST_SetCrs(ST_GeomFromWKT('{query_wkt}'), '{src_crs_json}')
+                )
+                """
+            )
```
paleolimbot, line 107: This won't work until 0.3 is release (maybe next week), but we added some improvements to this workflow:

```suggestion
            filtered_df = sd.sql(
                f"""
                SELECT {label_select} {geometry_col} as geometry
                FROM temp_df
                WHERE ST_Intersects(
                    ST_SetCrs({geometry_col}, $1),
                    ST_SetCrs($2, $1)
                )
                """, params = (src_crs, shapely.box(minx, miny, maxx, maxy))
            )
```

Unfortunately we can't parameterize column names via parameterized SQL (we'll add a programmatic `.select()` in 0.4).

torchgeo/datasets/sedonadb.py
```diff
@@ -0,0 +1,124 @@
...
+
+            filtered_gdf = filtered_df.to_pandas()
+            if len(filtered_gdf) > 0:
+                filtered_gdf = filtered_gdf.set_geometry('geometry', crs=src_crs)
```
paleolimbot, line 111: I'm surprised that the CRS needs resetting so many times (but I know that CRS equality is hard and I have gotten many such errors from GeoPandas)

paleolimbot (commented):
Cool!

Some optional suggestions on SedonaDB usage to take or leave 🙂

isaaccorley:
@coderabbitai review

PR #3165, opened 2025-12-02
Switch to pydata-sphinx-theme

### LINKED ISSUE #2006: Change documentation theme

ashnair1:
Fixes #2006 
Closes #3175

Widely used across the scientific Python ecosystem. Major libraries using this theme include NumPy, Pandas, SciPy, GeoPandas, Xarray etc. `pytorch_sphinx_theme2` inherits from this theme ([link](https://github.com/pytorch/pytorch_sphinx_theme/blob/0457e8af160b964e3055fffa4511f1fa0387e071/pytorch_sphinx_theme2/theme.conf#L3)) as well.

### DIFF
--- a/docs/_static/switcher.json
+++ b/docs/_static/switcher.json
@@ -39,5 +39,60 @@
     "name": "0.6.0",
     "version": "v0.6.0",
     "url": "https://torchgeo.readthedocs.io/en/v0.6.0/"
+  },
+  {
+    "name": "0.5.2",
+    "version": "v0.5.2",
+    "url": "https://torchgeo.readthedocs.io/en/v0.5.2/"
+  },
+  {
+    "name": "0.5.1",
+    "version": "v0.5.1",
+    "url": "https://torchgeo.readthedocs.io/en/v0.5.1/"
+  },
+  {
+    "name": "0.5.0",
+    "version": "v0.5.0",
+    "url": "https://torchgeo.readthedocs.io/en/v0.5.0/"
+  },
+  {
+    "name": "0.4.1",
+    "version": "v0.4.1",
+    "url": "https://torchgeo.readthedocs.io/en/v0.4.1/"
+  },
+  {
+    "name": "0.4.0",
+    "version": "v0.4.0",
+    "url": "https://torchgeo.readthedocs.io/en/v0.4.0/"
+  },
+  {
+    "name": "0.3.1",
+    "version": "v0.3.1",
+    "url": "https://torchgeo.readthedocs.io/en/v0.3.1/"
+  },
+  {
+    "name": "0.3.0",
+    "version": "v0.3.0",
+    "url": "https://torchgeo.readthedocs.io/en/v0.3.0/"
+  },
+  {
+    "name": "0.2.1",
+    "version": "v0.2.1",
+    "url": "https://torchgeo.readthedocs.io/en/v0.2.1/"
+  },
+  {
+    "name": "0.2.0",
+    "version": "v0.2.0",
+    "url": "https://torchgeo.readthedocs.io/en/v0.2.0/"
+  },
+  {
+    "name": "0.1.1",
+    "version": "v0.1.1",
+    "url": "https://torchgeo.readthedocs.io/en/v0.1.1/"
+  },
+  {
+    "name": "0.1.0",
+    "version": "v0.1.0",
+    "url": "https://torchgeo.readthedocs.io/en/v0.1.0/"
   }
 ]


docs/conf.py
```diff
@@ -69,26 +68,51 @@
...
     ('py:class', 'torchvision.models.resnet.ResNet'),
     ('py:class', 'torchvision.models.swin_transformer.SwinTransformer'),
+    # CRS string representations (not actual classes)
+    (
+        'py:class',
+        'World Geodetic System 1984 ensemble - Ellipsoid: WGS 84 - Prime Meridian: Greenwich',
+    ),
+    (
+        'py:class',
+        '- name: Popular Visualisation Pseudo-Mercator - method: Popular Visualisation Pseudo Mercator Datum: World Geodetic System 1984 ensemble - Ellipsoid: WGS 84 - Prime Meridian: Greenwich',
```
adamjstewart: These don't make a lot of sense, where are these coming from?

docs/conf.py
```diff
@@ -69,26 +68,51 @@
...
-    'logo_only': True,
-    'pytorch_project': 'docs',
-    'navigation_with_keys': True,
-    'analytics_id': 'UA-209075005-1',
+    'show_nav_level': 2,
+    'show_toc_level': 2,
+    'navigation_depth': 4,
+    'navbar_align': 'left',
+    'header_links_before_dropdown': 6,
+    'icon_links': [
```
adamjstewart, line 108: These are cool!

docs/conf.py
```diff
@@ -69,26 +68,51 @@
...
+        {
+            'name': 'GitHub',
+            'url': 'https://github.com/torchgeo/torchgeo',
+            'icon': 'fa-brands fa-github',
+        },
+        {
+            'name': 'Slack',
+            'url': 'https://torchgeo.slack.com',
```
adamjstewart: Could add the invite link. Unfortunately the current link is useless unless you've already joined.

.readthedocs.yaml
```diff
@@ -8,15 +8,14 @@ version: 2
 build:
   os: ubuntu-lts-latest
   tools:
-    # Sphinx 6.1 and older do not support Python 3.13+
-    python: '3.12'
+    # Sphinx 6.2+ and pydata-sphinx-theme support Python 3.13+
```
adamjstewart: Can remove this comment, we always want to use the latest Sphinx and theme when building the docs. This comment only explained why it wasn't possible to update Python.

pyproject.toml
```diff
@@ -122,12 +122,11 @@ docs = [
...
     "nbsphinx>=0.8.5",
-    # release versions missing files, must install from master
-    "pytorch-sphinx-theme",
-    # sphinx 4+ required for autodoc_typehints_description_target
-    # sphinx 6+ is incompatible with pytorch-sphinx-theme
-    # https://github.com/pytorch/pytorch_sphinx_theme/issues/175
-    "sphinx>=4,<6",
+    # pydata-sphinx-theme 0.16.0+ required for sphinx 6.1+ compatibility
+    "pydata-sphinx-theme>=0.16.0",
+    # sphinx 6.2+ required for Python 3.13 compatibility (imghdr module removed)
```
adamjstewart: Python 3.13 is not a lowest supported version. Which versions of pydata-sphinx-theme work for Python 3.11?

adamjstewart (commented):
Looks decent, fixes many of our bugs. Quite different from our previous theme, so may take some getting used to. Hasn't had a new release in almost a year, so either very stable or unmaintained.

I think there are a few other things we'll like to customize, such as default pages for different tabs and the version selection menu, but we can do those if we decide this is the style we want.

If it's not too much work, it would be nice to see a few other alternatives that are either simple or widely used by the community.

docs/conf.py
ashnair1: These were the error messages I got. 

```
<unknown>:1: WARNING: py:class reference target not found: World Geodetic System 1984 ensemble - Ellipsoid: WGS 84 - Prime Meridian: Greenwich [ref.class]
<unknown>:1: WARNING: py:class reference target not found: - name: Popular Visualisation Pseudo-Mercator - method: Popular Visualisation Pseudo Mercator Datum: World Geodetic System 1984 ensemble - Ellipsoid: WGS 84 - Prime Meridian: Greenwich [ref.class]
```

Haven't dug too deeply into why this occurs here. Could be due to some issues with Sphinx interpreting default CRS objects as classes in our dataset constructors. 
```python 
crs: CRS = CRS.from_epsg(4326),
```

adamjstewart: We can replace those with None and then if None replace.

docs/conf.py
ashnair1: Added invite link

ashnair1:
One note on the landing page (index). Currently it shows the docstring in [torchgeo/\_\_init\_\_.py](https://github.com/torchgeo/torchgeo/blob/main/torchgeo/__init__.py) (left image) but I could show the README instead (right image)

image

docs/conf.py
ashnair1: That worked. Have removed the warnings from the `nitpick_ignore` list

pyproject.toml
ashnair1: I didn't look into 3.11 since we don't test docs with minimum supported python and deps, so it wasn't a priority for me.

adamjstewart: Could set the lower bound based on features we use instead of Python version support.

ashnair1: Done

docs/conf.py
adamjstewart: Can remove todo now that we removed our only todo

docs/conf.py
```diff
@@ -50,8 +48,9 @@
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = ['_build']
 
-# Sphinx 4.0+ required for autodoc_typehints_description_traget
-needs_sphinx = '4.0'
+# Sphinx 5.3+ required to allow section titles inside autodoc class docstrings
+# https://github.com/sphinx-doc/sphinx/pull/10887
```
adamjstewart, line 51: I would also be okay with removing section headings from docstrings, I don't imagine it happens in very many places.

.readthedocs.yaml
```diff
@@ -8,15 +8,13 @@ version: 2
 build:
   os: ubuntu-lts-latest
   tools:
-    # Sphinx 6.1 and older do not support Python 3.13+
-    python: '3.12'
+    python: '3.13'
```
adamjstewart, line 11: Can test 3.14 now if you rebase

adamjstewart (commented):
I'm leaning towards this theme because it matches the rest of the PyTorch ecosystem but is still maintained. A couple remaining questions:

* Are we able to remove any of the files in `docs/_static`? These were CSS hacks designed to fix bugs in pytorch-sphinx-theme.
* I'm curious if there's a way to get rid of the version switcher in the bottom right corner and add one to the top menu bar like pytorch2-sphinx-theme. It's possible this is also a cache issue on my side.

ashnair1:
> I'm leaning towards this theme because it matches the rest of the PyTorch ecosystem but is still maintained. A couple remaining questions:
> 
> * Are we able to remove any of the files in `docs/_static`? These were CSS hacks designed to fix bugs in pytorch-sphinx-theme.
> * I'm curious if there's a way to get rid of the version switcher in the bottom right corner and add one to the top menu bar like pytorch2-sphinx-theme. It's possible this is also a cache issue on my side.

- I have added the version switcher to the top of the navbar
- I was able to remove `badge-height.css` and `table-scroll.css` without affecting the docs. `notebook-prompt.css` could be removed as well but I didn't go through with it as I think we don't want those line numbers. Below image shows how it would look. Left (current), Right (without `notebook-prompt.css`)
notebookprompt-removal

.readthedocs.yaml
ashnair1: Needs #3203 to be merged

adamjstewart: Technically that shouldn't matter, the pytest warning filters don't apply to sphinx anyway.

ashnair1: I tried switching to 3.14 earlier and the lightly warnings caused the RtD build to fail.

adamjstewart: Hmm, that might be a problem. I don't know how to tell RtD to ignore that, #3203 shouldn't help.

ashnair1: You were right. Warnings are still raised. I'll proceed with building docs on 3.13 for now

ashnair1:
Note: Currently colors are based on the default pydata-sphinx-theme i.e. blue and purple. This can be overridden with a custom css. Can be done in a later PR

adamjstewart:
Python 3.14 error is:
```
WARNING: Failed to get a method signature for torchgeo.datasets.GeoDataset.__annotate_func__: None is not a callable object
```
This looks like https://github.com/sphinx-doc/sphinx/issues/13701. So Sphinx 9+ is required for Python 3.14+.

ashnair1:
> Python 3.14 error is:
> 
> ```
> WARNING: Failed to get a method signature for torchgeo.datasets.GeoDataset.__annotate_func__: None is not a callable object
> ```
> 
> This looks like [sphinx-doc/sphinx#13701](https://github.com/sphinx-doc/sphinx/issues/13701). So Sphinx 9+ is required for Python 3.14+.

Only thing stopping us from supporting sphinx 9+ is `myst-parser` which currently doesn't support sphinx 9+ and is required to render the README in the documentation. Once executablebooks/MyST-Parser#1076 is merged that should be solved.

adamjstewart:
Drop the README then 🙂

ashnair1:
Docs now work with Python 3.14. There is an issue with tutorials test though.

adamjstewart:
> `notebook-prompt.css` could be removed as well but I didn't go through with it as I think we don't want those line numbers.

I vaguely recall nbsphinx having an option for this, but it didn't work with pytorch-sphinx-theme. Check nbsphinx settings and see if we can still remove this.

> Docs now work with Python 3.14. There is an issue with tutorials test though.

I've been seeing a bunch of issues with Python 3.14. Let's switch to 3.13 for now: #3228

ashnair1:
> I vaguely recall nbsphinx having an option for this, but it didn't work with pytorch-sphinx-theme. Check nbsphinx settings and see if we can still remove this.

Didn't see an option for this. nbsphinx's docs suggest the same to hide the input and output prompt - https://nbsphinx.readthedocs.io/en/latest/custom-css.html#For-a-Single-Notebook

adamjstewart:
You are completely right: https://github.com/spatialaudio/nbsphinx/issues/671

docs/_static/switcher.json
adamjstewart, line 1: Frustrating that this requires manual updates. But I read through https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/version-dropdown.html and this seems to be the recommended way.

docs/conf.py
```diff
@@ -76,25 +74,64 @@
...
-    'logo_only': True,
-    'pytorch_project': 'docs',
-    'navigation_with_keys': True,
-    'analytics_id': 'UA-209075005-1',
+    'show_nav_level': 2,
+    'show_toc_level': 2,
+    'navigation_depth': 4,
+    'navbar_align': 'left',
+    'header_links_before_dropdown': 6,
+    'icon_links': [
```
adamjstewart, line 108: Wonder if we want more here, we also have a YouTube channel, Hugging Face community, OSGeo page, etc. Basically every button in our README should work here. But I don't want it to get too crowded.

docs/_static/switcher.json
adamjstewart: When making a new release, do we update this file before or after the release? Either way, this should be added to https://github.com/torchgeo/torchgeo/wiki/Releasing-Instructions

ashnair1: File needs to be updated after the release. I'll update the wiki.

docs/conf.py
ashnair1: Will look into enhancements in a later PR

docs/_static/switcher.json
adamjstewart: But if I update this file after I make a release, won't the stable docs say I'm looking at 0.8.0 when I'm actually looking at 0.8.1? Or does it reference the live git repo as opposed to the frozen HTML build?

ashnair1: Yeah as seen in the line below

https://github.com/torchgeo/torchgeo/blob/a0725bd5db6445f94ba48bf06a9236ec683841f4/docs/conf.py#L81

All documentation versions (stable, older releases, etc.) fetch the switcher from the live `/en/latest/_static/switcher.json`, not from a frozen copy in their own build.

adamjstewart: Other option would be to only have 2 versions (stable and latest) in the switcher. Clever users will be able to guess older releases. We are pretty good at documenting when features were added/changed. And we really only support the latest version.

PR #3169, opened 2025-12-04
Added an example on how to override resampling in a RasterDataset subclass

Moritz-Langer:
Following the suggestion of @adamjstewart in https://github.com/torchgeo/torchgeo/pull/2015 I added the example to override the resampling.
This might illustrate the process for the other properties as well.


adamjstewart:
This is true for all dataset attributes, not just resizing. We have an example at the bottom where we override several dataset attributes. Maybe we could add it there?

Moritz-Langer:
Closing this and creating a new PR with the correct branch that follows your suggestion.

PR #3170, opened 2025-12-04
Fix geopandas type hints

adamjstewart:
Seems to have a lot more errors than other libs, but let's see what we can do.


adamjstewart:
It seems that pandas-stubs doesn't yet support type hints on a DataFrame index. We would have to cast everything to make mypy happy. Not sure why this problem only arises when using types-geopandas and not pandas-stubs. Perhaps types-geopandas could be changed to set `index` to `Any` since it isn't known.

PR #3176, opened 2025-12-06
Change Detection Tutorial

hkristen:
Initial draft of the Change Detection tutorial notebook based on the LEVIRCD100 dataset from #3142. 

Figured I create this while we wait for feedback regarding the LEVIRCD100 license and #3142 to merge. If we really shall use a different dataset, most of this can be re-used.


adamjstewart:
Tests are failing because the notebook currently isn't tested. Need to add it to `docs/tutorials/case_studies.rst`.

I think we should focus on the LEVIR-CD100 PR first.

hkristen:
Replaced by https://github.com/torchgeo/torchgeo/pull/3222

PR #3181, opened 2025-12-08
FLAIR-HUB dataset

gatienc:
The [FLAIR-HUB](https://github.com/IGNF/FLAIR-HUB) dataset is a large-scale, multimodal dataset for land cover and crop mapping from the French National Institute of Geographical and Forest Information (IGN). It builds upon and includes the FLAIR#1 and FLAIR#2 datasets, expanding them into a unified resource with very-high-resolution annotations.

This PR adds the [FLAIRHUB](https://ignf.github.io/FLAIR/FLAIR-HUB/flairhub.html) mono-temporal part of the dataset.

image

Let me know if anything has to changed. the monotemporal part of the dataset is already very solid and can be used independently. I plan to work on the time-series part of the dataset once it will be implemented in Torchgeo.

## Note 
As the dataset size is huge ( total of ~750 GB) I wasn't able to test training on the full dataset but it is possible to work on a subset of the dataset by commenting most of the domain years dict in  Torchgeo/datasets/flairhub.py#L174)

## Acknowledgments
Special thanks to the work done in #2394, which served as a starting point for this implementation.

## Todo
- [x] Implement the time-series part of the dataset
- [x] Add the historical part of the dataset


gatienc:
Sorry for the repost! I accidentally used the main branch on my fork for this PR. I needed to clean up my fork to work on other contributions.

adamjstewart:
@vbuchauer is interested in this dataset for time-series benchmarking. How hard would it be to add the multi-temporal version of the dataset? We're in the process of converting all time-series datasets to B x T x C x H x W, and FLAIR-HUB seems like a good candidate for this.

gatienc:
I definitely want to implement the time-series part of this. I heard in the discussion that the time-series implementation is ongoing, is there already a sample dataset I could base the behavior on so I can match the pattern on FlairHub?

adamjstewart:
Yes, PASTIS is a good example. As long as the dataset returns T x C x H x W data, then it should be compatible.

gatienc:
@adamjstewart The mono temporal part  of the dataset is done and have 100% coverage now.
It seems that there is an issue with the optional test but I don't understand why, could you help me with that?

I'll be working on adding the temporal part of the dataset from now.

adamjstewart:
Optional tests: we've been seeing segfaults (or something like that) with our Python 3.14 tests, this looks related. I made Python 3.14 optional, but should probably switch our optional tests back to Python 3.13 for stability. Don't worry about this for now.

gatienc:
For the tests data, I currently used some data sample directly, while it made the implementation simpler, it might be quite heavy for the repo, let me know if I should use generated data instead.

adamjstewart:
We are not allowed to use real data during testing, as it isn't compatible with TorchGeo's MIT license. You should instead create fake data, see the `tests/data` directory for examples.

gatienc:
I hadn't considered the licensing compatibility issue with the real data.

I removed the real data, refactored the test directory to use fake data instead, which is now generated in the data.py as most of the other test files.

gatienc:
The LPIS (crop) dataset is labeled with three hierarchical levels. Currently, the paper only utilizes the first level for model training, ignoring the second and third levels which offer deeper classification. What would be the best method to implement a hierarchical classification system ? Is there any standard datasets I could use as an example?

gatienc:
I’ve implemented the temporal part of the dataset. I haven't used this data in the test trainer yet, as I'm looking for a simple way to include it.

I’m also unsure about my implementation of simplest_preprocessing_flairhub and would like your opinion: should this stay in the datamodule file, or should it be defined only for the test trainer?

adamjstewart:
> The LPIS (crop) dataset is labeled with three hierarchical levels. Currently, the paper only utilizes the first level for model training, ignoring the second and third levels which offer deeper classification. What would be the best method to implement a hierarchical classification system ? Is there any standard datasets I could use as an example?

EuroCrops and maybe TreeSatAI also have hierarchical classification, although this isn't exposed via TorchGeo. Simplest way to implement would be a Dataset parameter letting you select the hierarchy level. This isn't really full hierarchical classification though. Does torchmetrics have any way to compute metrics for hierarchical predictions?

gatienc:
I couldn't find existing hierarchical prediction support in TorchMetrics, so I implemented each hierarchy level as a distinct entry in the class option. I retained crop_type (instead of crop_type_1) because Level 1 serves as the paper baseline and should be considered the default.

Additionally, I replicated the colormap from the original paper. Since the authors did not provide unique colors for the specific subclasses, I have kept the visualization as is for consistency with their results.

gatienc:
I removed duplicate TIF files from history by squashing them.

The temporal data is not used in the datamodule as I haven't implemented the paper's model.

@adamjstewart The dataset should now be ready for review

torchgeo/datasets/flairhub.py
adamjstewart, line 1: Should we name this file `flair.py` in case someone wants to add FLAIR#1 and FLAIR#2 later?

torchgeo/datasets/flairhub.py
```diff
@@ -0,0 +1,1361 @@
...
+
+from .errors import DatasetNotFoundError
+from .geo import NonGeoDataset
+from .utils import download_url, extract_archive, percentile_normalization
+
+
+class FLAIRHUB(NonGeoDataset):
+    """FLAIR-HUB: Large-scale Multimodal Dataset for Land Cover and Crop Mapping dataset.
+
+    FLAIR-HUB (<https://github.com/IGNF/FLAIR-HUB>) builds upon and includes the FLAIR#1 and FLAIR#2 datasets, expanding them into a unified, large-scale, multi-sensor land-cover resource with very-high-resolution annotations. Spanning over 2,500 km² of diverse French ecoclimatic regions and landscapes, FLAIR-HUB features 63 billion hand-annotated pixels across 19 land-cover and 23 crop type classes.
```
adamjstewart: Each line should be wrapped to ~88 characters to make it easier to view in a text editor. This is done automatically for code by ruff, but ruff doesn't know how to wrap docstrings without breaking formatting.

torchgeo/datasets/flairhub.py
```diff
@@ -0,0 +1,1361 @@
...
+    Dataset features:
+
+    ROI / Area Covered: 2,822 ROIs / 2,528 km²
+    Departments (France): 74
+    AI Patches (512x512 px): 241,100
+    Annotated Pixels: 63.2 billion
+    Sentinel-2 Acquisitions: 256,221
+    Sentinel-1 Acquisitions: 532,696
+    Total Files: ~2.5 million
+    Total Dataset Size: ~750 GB
```
adamjstewart: This will not get formatted correctly by Sphinx (line breaks are ignored). I would suggest a bulleted list here.

torchgeo/datasets/flairhub.py
```diff
@@ -0,0 +1,1361 @@
...
+
+    FLAIR-HUB (<https://github.com/IGNF/FLAIR-HUB>) builds upon and includes the FLAIR#1 and FLAIR#2 datasets, expanding them into a unified, large-scale, multi-sensor land-cover resource with very-high-resolution annotations. Spanning over 2,500 km² of diverse French ecoclimatic regions and landscapes, FLAIR-HUB features 63 billion hand-annotated pixels across 19 land-cover and 23 crop type classes.
+
+    The dataset integrates complementary sources including aerial imagery, SPOT and Sentinel satellite acquisitions, surface models, and historical aerial photographs. This offers rich spatial, spectral, and temporal diversity, supporting a broad range of research tasks, including semantic segmentation, multimodal fusion, and self-supervised learning. FLAIR-HUB is designed as a continuously growing resource, with new modalities and annotations to be released in future updates.
+
+    Dataset features:
+
+    ROI / Area Covered: 2,822 ROIs / 2,528 km²
+    Departments (France): 74
+    AI Patches (512x512 px): 241,100
```
adamjstewart: What are "AI Patches"?

torchgeo/datasets/flairhub.py
```diff
@@ -0,0 +1,1361 @@
...
+    Total Dataset Size: ~750 GB
+
+    Dataset structure:
+
+    The dataset is organized by domains (geographical areas) and years. Each domain-year
+    combination has multiple modalities available for download. The full dataset contains
+    66 unique domains (D004-D091, non-consecutive), with years ranging from 2017-2022.
+    Most domains have data for a single year, but some have multiple years available.
+
+    Available modalities (100% coverage across all domains):
```
adamjstewart: I think you need a blank line before the start of a bulleted list for Sphinx to format it correctly

torchgeo/datasets/flairhub.py
```diff
@@ -0,0 +1,1361 @@
...
+            22: background
+
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://arxiv.org/abs/2506.07080
+    """
+
+    # AERIAL_RGBI
+    aerial_rgb_bands: tuple[str, str, str] = ('B01', 'B02', 'B03')
```
adamjstewart: These should not require explicit type annotations, mypy can figure out the type of simple classes.

torchgeo/datasets/flairhub.py
```diff
@@ -0,0 +1,1361 @@
...
+        'cloud',
+    )  # snow and cloud probabilty mask
+
+    # SENTINEL1_ASC_TS (Sentinel-1 ASC time series)
+    sentinel1_asc_ts_bands: tuple[str, str] = ('VV', 'VH')
+    # SENTINEL1_DESC_TS (Sentinel-1 DESC time series)
+    sentinel1_desc_ts_bands: tuple[str, str] = ('VV', 'VH')
+
+    # HuggingFace dataset URL
+    url_base: ClassVar[str] = (
```
adamjstewart: ClassVar isn't needed because strings are immutable

torchgeo/datasets/flairhub.py
```diff
@@ -0,0 +1,1361 @@
...
+        'https://huggingface.co/datasets/IGNF/FLAIR-HUB/resolve/main/data'
```
adamjstewart:
```suggestion
        'https://hf.co/datasets/IGNF/FLAIR-HUB/resolve/e8ed7981d488508aa70bb05c37cf6585432b7d5f/data'
```
Git commit hash is required for stable downloads

torchgeo/datasets/flairhub.py
```diff
@@ -0,0 +1,1361 @@
...
+                '#80c1d7',  # other permanent crops
+                '#000000',  # mixed crops
+                '#000000',  # background
+            ]
+        ),
+    }
+
+    def __init__(
+        self,
+        root: Path = Path('data'),
```
adamjstewart: Let's default to a str to match all other existing datasets.

torchgeo/datasets/flairhub.py
```diff
@@ -0,0 +1,1361 @@
...
+        transforms: Callable[[dict[str, Tensor]], dict[str, Tensor]] | None = None,
+        download: bool = False,
+        use_aerial_rgbi: bool = False,
+        use_dem_elev: bool = False,
+        use_spot_rgbi: bool = False,
+        use_aerial_rlt_pan: bool = False,
+        use_sentinel2_ts: bool = False,
+        use_sentinel2_msk_sc: bool = False,
+        use_sentinel1_asc_ts: bool = False,
+        use_sentinel1_desc_ts: bool = False,
```
adamjstewart: Could all of these be combined into a single `bands` parameter?

torchgeo/datasets/flairhub.py
```diff
@@ -0,0 +1,1361 @@
...
+            use_sentinel1_desc_ts: If True, load Sentinel-1 DESC time series data.
+            dataset_type: Type of labels to use. Choose 'land_cover' for 19-class COSIA labels
+                or 'crop_type' for 23-class LPIS crop classification labels (baseline) or 'crop_type_2' for 31-class LPIS crop classification labels or 'crop_type_3' for 46-class LPIS crop classification labels. crop_type_2 and crop_type_3 are deeper levels of the LPIS crop classification labels.
+
+        Raises:
+            DatasetNotFoundError: If dataset is not found and *download* is False.
+            ValueError: If *dataset_type* is not 'land_cover' or 'crop_type' or 'crop_type_2' or 'crop_type_3'.
+            ValueError: If at least one data modality must be enabled (set to True) to load samples.
+
+        .. versionadded:: 0.9
```
adamjstewart: I would move this to the class-level docstring instead of `__init__`

torchgeo/datasets/flairhub.py
```diff
@@ -0,0 +1,1361 @@
...
+
+    def __getitem__(self, index: int) -> dict[str, Tensor]:
+        """Return an index within the dataset.
+
+        Args:
+            index: index to return
+
+        Returns:
+            dictionary containing tensors for each modality
+            keys are the modality names : 'mask', 'aerial_rgbi', 'spot_rgbi', 'dem_elev', 'aerial_rlt_pan', 'sentinel2_ts', 'sentinel2_msk_sc', 'sentinel1_asc_ts', 'sentinel1_desc_ts'
```
adamjstewart: Newline is removed, so without a period on the first sentence this won't make sense.

torchgeo/datasets/flairhub.py
```diff
@@ -0,0 +1,1361 @@
...
+        self.use_sentinel1_desc_ts = use_sentinel1_desc_ts
+        self.requested_modalities = self._get_requested_modalities()
+        # Verify at least one modality is selected
+        if len(self.requested_modalities) == 0:
+            raise ValueError('At least one data modality must be enabled')
+
+        self._verify()
+        self.files = self._load_files()
+
+    def __getitem__(self, index: int) -> dict[str, Tensor]:
```
adamjstewart: Return type should be `Sample`, this is something that changed recently

torchgeo/datasets/flairhub.py
```diff
@@ -0,0 +1,1361 @@
...
+                    # Check if zip exists but not extracted
+                    if modality_zip.is_file():
+                        to_extract.append((domain, year_str, modality_suffix))
+                    else:
+                        to_download.append((domain, year_str, modality_suffix))
+
+        # Extract any zips that exist but haven't been extracted
+        if to_extract:
+            print(f'Extracting {len(to_extract)} modality archives...')
+            with ThreadPoolExecutor(max_workers=4) as executor:
```
adamjstewart: Parallelism is neat, but perhaps this could be added as a feature to `download_url` and `extract_archive` later? Simple code without duplication is more important than speed for maintainability.

torchgeo/datasets/flairhub.py
```diff
@@ -0,0 +1,1361 @@
...
+            tensor = torch.from_numpy(array).long()
+        return tensor
+
+    def _load_time_series(self, path: Path, num_bands: int) -> Tensor:
+        """Load a time series from a path and process it to be in the T x C x H x W format instead of the usual (TxC) x H x W format.
+
+        Args:
+            path: path to the time series
+            num_bands: number of bands to load
+        Returns:
```
adamjstewart, line 1605: I think you need an empty line between Args and Returns

torchgeo/datasets/flairhub.py
```diff
@@ -0,0 +1,1361 @@
...
+        Args:
+            path: path to the image
+        Returns:
+            Tensor: the loaded image
+        """
+        with rasterio.open(str(path)) as f:
+            array: np.typing.NDArray[np.int32] = f.read()
+            if (
+                array.dtype == np.uint16
+            ):  # Had to add this as it seems that Pytorch on Windows doesn't support uint16 tensors
```
adamjstewart: If you move comments to their own separate lines, you end up with much simpler and fewer lines after formatting.

We also have a builtin `array_to_tensor` method that could help here. I don't think this is a Windows-specific problem.

torchgeo/datasets/flairhub.py
```diff
@@ -0,0 +1,1361 @@
...
+        plt.tight_layout()
+        return fig
+
+
+class FLAIRHUBToy(FLAIRHUB):
+    """Toy Version of the FLAIRHUB dataset. For further information refer to the FLAIRHUB dataset."""
+
+    download_link: str = 'https://storage.gra.cloud.ovh.net/v1/AUTH_366279ce616242ebb14161b7991a8461/defi-ia/flair_hub/FLAIR-HUB_TOY_DATASET.zip'
+
+    def __init__(
```
adamjstewart, line 1921: Is this `__init__` different from the super class? Wondering if we can remove this.

torchgeo/datamodules/flairhub.py
```diff
@@ -0,0 +1,287 @@
...
+
+    The exact splits used in the research paper can be found here:
+    https://huggingface.co/datasets/IGNF/FLAIR-HUB/resolve/main/data/GLOBAL_ALL_MTD.zip
+    in the file:
+    GLOBAL_ALL_MTD_SPLIT.gpkg
+
+    .. versionadded:: 0.9
+    """
+
+    official_splits_url = 'https://huggingface.co/datasets/IGNF/FLAIR-HUB/resolve/main/data/GLOBAL_ALL_MTD.zip'
```
adamjstewart: Stable git commit hash

torchgeo/datamodules/flair.py
```diff
@@ -0,0 +1,287 @@
...
+        train_val_split: float = 0.8,
+        random_state: int = 0,
+        official_splits: Literal[
+            'split_1',
+            'split_2',
+            'split_3',
+            'split_4',
+            'split_5',
+            'split_flairchallenge',
+        ]
```
adamjstewart: Wonder if this should be added to the dataset

torchgeo/datamodules/flairhub.py
```diff
@@ -0,0 +1,287 @@
...
+
+    def __init__(
+        self,
+        batch_size: int = 64,
+        num_workers: int = 0,
+        n_test_domains: int = 16,
+        train_val_split: float = 0.8,
+        random_state: int = 0,
```
adamjstewart: None of our other datamodules have this. Maybe we should just hard-code it to 0?

torchgeo/datamodules/flairhub.py
```diff
@@ -0,0 +1,287 @@
...
+        official_splits: Literal[
+            'split_1',
+            'split_2',
+            'split_3',
+            'split_4',
+            'split_5',
+            'split_flairchallenge',
+        ]
+        | None = None,
+        use_toy: bool = False,
```
adamjstewart: Would rather have 2 separate datamodules to match 2 separate datasets.

torchgeo/datamodules/flairhub.py
```diff
@@ -0,0 +1,287 @@
...
+        """
+        if not (0 < train_val_split < 1):
+            raise ValueError('train_val_split must be in the range (0, 1)')
+
+        if n_test_domains < 1:
+            raise ValueError('n_test_domains must be at least 1')
+
+        # Apply default normalization if no transforms provided
+        if 'transforms' not in kwargs:
+            kwargs['transforms'] = normalize_flairhub
```
adamjstewart: We should be using K.Normalize so that transforms are applied on the GPU

torchgeo/datamodules/flair.py
```diff
@@ -0,0 +1,287 @@
...
+
+        Args:
+            stage: Either 'fit', 'validate', 'test', or 'predict'.
+        """
+        # Create the full dataset to extract domain information
+        dataset = cast(FLAIRHUB | FLAIRHUBToy, self.dataset_class(**self.kwargs))
+
+        if self.official_splits is not None:
+            # Use official splits from the gpkg file
+            train_indices, val_indices, test_indices = self._get_official_splits(
```
adamjstewart: Downloading can't occur in `setup` as this runs in parallel on each worker. It should instead go in `prepare_data`. The default implementation implements `dataset_class(download=True)` so if you move this to the dataset that also solves this problem.

tests/trainers/test_segmentation.py
```diff
@@ -97,6 +141,20 @@ def test_trainer(
                     'ecec8e871faf1bbd8ca525ca95ddc1c1f5213f40afb94599884bd85f990ebd6b'
                 )
                 monkeypatch.setattr(LandCoverAI, 'sha256', sha256)
+            case 'flairhub_croptype' | 'flairhub_landcover':
+                # Mock domain_years to only include domains present in test data
+                test_domain_years = {'D006': ['2020'], 'D012': ['2019']}
+                monkeypatch.setattr(FLAIRHUB, 'domain_years', test_domain_years)
```
adamjstewart: Can this go in the config file?

tests/trainers/test_segmentation.py
```diff
@@ -97,6 +141,20 @@ def test_trainer(
...
+
+                # Patch normalize_flairhub to also concatenate modalities
+                # This is needed because standard segmentation models expect 'image'
+                import torchgeo.datamodules.flairhub as flairhub_module
+
+                monkeypatch.setattr(
+                    flairhub_module,
+                    'normalize_flairhub',
+                    flairhub_concat_modalities_transform,
+                )
```
adamjstewart: This implies that our current dataset/datamodule is incompatible with all existing trainers. Can we instead add a `bands` parameter to the dataset and change the dataset (or datamodule) to return a single `image` key with all modalities concatenated into one? Otherwise this datamodule isn't useful.

tests/datasets/test_flairhub.py
```diff
@@ -0,0 +1,455 @@
...
+            (False, False, False, True, False, False, False, False, 'land_cover'),
+            (False, False, False, False, True, False, False, False, 'land_cover'),
+            (False, False, False, False, False, True, False, False, 'land_cover'),
+            (False, False, False, False, False, False, True, False, 'land_cover'),
+            (False, False, False, False, False, False, False, True, 'land_cover'),
+            (True, True, True, True, True, True, True, True, 'land_cover'),
+            (True, True, True, True, True, True, True, True, 'crop_type'),
+            (True, True, True, True, True, True, True, True, 'crop_type_2'),
+            (True, True, True, True, True, True, True, True, 'crop_type_3'),
+        ]
```
adamjstewart: 1. This could probably be simplified using itertools somehow
2. We don't really need to test every single method with every single possible input, just a couple common examples to make sure most things work

tests/datasets/test_flairhub.py
```diff
@@ -0,0 +1,455 @@
...
+        )
+
+        (root / 'D006-2020_SPOT_RGBI.zip').unlink()
+
+        # Try to extract a non-existent zip
+        with pytest.raises(FileNotFoundError, match='Archive not found'):
+            dataset._extract('D006', '2020', 'SPOT_RGBI')
+
+
+class TestFLAIRHUBToy:
```
adamjstewart: Instead of duplicating all testing, we can parametrize the fixture over both `[FLAIRHUB, FLAIRHUBToy]` like we do in our EuroSAT tests.

tests/data/flair/data.py
```diff
@@ -0,0 +1,269 @@
...
+np.random.seed(0)
+
+DOMAIN_YEARS = {
+    'D006-2020': ['FF-S1-14', 'FU-S2-12'],
+    'D012-2019': ['AF-S1-27', 'AU-S1-18'],
+    'D006-195X': ['FF-S1-14', 'FU-S2-12'],
+    'D012-195X': ['AF-S1-27', 'AU-S1-18'],
+}
+
+MODALITIES = {
```
adamjstewart, line 68: Wonder if we can generate a subset of these files for testing so we don't need 150+ new files...

tests/datamodules/test_flair.py
adamjstewart, line 1: We are trying to remove this directory, as datamodule tests aren't actually useful. Instead, we should test all datamodules with their corresponding trainers. I'll add a README to make this more clear.

adamjstewart (changes requested):
Currently ~150 files and ~2.7k lines of code. If we can get this to ~10 files and ~500 lines of code, I would feel much more comfortable about merging this. Whether or not the dataset works is one thing, but whether or not it is maintainable is another.

There is a lot of complexity added by the dataset being multimodal. If we return a single 'image' with all bands stacked together, that solves most of the complexity problems. We also can't have any tests that download anything during CI, so make sure we aren't downloading test split files or things like that.

Also needs documentation, updates to `docs/api/datasets.rst`, `docs/api/datamodules.rst`, and `docs/api/datasets/non_geo_datasets.csv`.

gatienc:
> There is a lot of complexity added by the dataset being multimodal. If we return a single 'image' with all bands stacked together, that solves most of the complexity problems. 

I agree that using a single "image" with all bands stacked could simplify the API and enable it to have something closer to the usual Torchgeo API, however I'm concerned about stacking imagery that have very different spatial resolutions. Aerial imagery is 512*512px while sentinel 1 or 2 are 10*10px.
Furthermore Flair-Hub also has temporal modality with variable amounts of observation for each patch.
"image" could be a list of tensors, but I don't think it would simplify the implementation a lot.
 
I feel like giving the choice in the API of which satellite imagery to enable is necessary. The total dataset size is 708 GB with all modality enabled. However only using specific combinations of smaller imagery could be very useful for the user, for example SPOT6-7 and land cover masks is only 8.5 GB. Also Flair-Hub provides multiple models trained with specific combinations of satellite imagery.

> We also can't have any tests that download anything during CI, so make sure we aren't downloading test split files or things like that.

There is no download during CI, I monkey patched the tests when necessary to avoid any download during CI. I think that a lot of the tests are not necessary but I aimed for the 100% test coverage which neccessitates monkey patching. 

If we want to reduce the number of lines of code, should we remove the toy dataset completely? If a user would want a reduced dataset he could change the domain_years dict to only get a small number of department.

I agree that the created data necessitates a very large number of files, I had to create them for the tests. A simple solution would be to generate data dynamically during the test run in the CI.

However I think that the number of files can be greatly reduced by removing the toy dataset. Removing it already divides the number of files by two. 
We could use only one department instead of two and only one patch for the department, which would lead to reduce the number of created tif file to 10 tif files (one per satellite type). I think it might be the best option.

tests/datamodules/test_flair.py
gatienc: I see, most of the coverage is done with the datamodule trainer, however to reach 100% test coverage, I had to implement a test for the specific split that have to be downloaded, I am monkeypatching the split so there is no download in the test. 
There is other tests for testing the handling of wrong input in the datamodule, as the data modules offers to split depending on the number of department instead of a percentage, I felt that it coud help the users.

adamjstewart: If we remove the handling for wrong input, we can get rid of this file. I try to use type hints to at least statically check that all inputs are valid.

torchgeo/datasets/flairhub.py
gatienc: For the Flair-Hub docstring I was heavily inspired by Flair-Hub website:
https://ignf.github.io/FLAIR/FLAIR-HUB/flairhub.html

AI Patches here seems to correspond to the number of individual patches per imagery, in a way that if you merge all imagery, you obtain 241 100 individual patches ready to train "AI".

adamjstewart:
> I agree that using a single "image" with all bands stacked could simplify the API and enable it to have something closer to the usual Torchgeo API, however I'm concerned about stacking imagery that have very different spatial resolutions. Aerial imagery is 512512px while sentinel 1 or 2 are 1010px.
> Furthermore Flair-Hub also has temporal modality with variable amounts of observation for each patch.
"image" could be a list of tensors, but I don't think it would simplify the implementation a lot.

Understood. Let's keep the dataset as is then. However, the data module isn't useful unless we start making opinionated decisions. We could resize all images to the max resolution and pad all sequences to the max length in order to use existing trainers. We don't really want to write a new trainer from scratch just for this dataset.

> I feel like giving the choice in the API of which satellite imagery to enable is necessary. The total dataset size is 708 GB with all modality enabled. However only using specific combinations of smaller imagery could be very useful for the user, for example SPOT6-7 and land cover masks is only 8.5 GB. 

Yes, but with a single `bands` parameter that either takes a list like `['S1_VV', 'S1_VH', 'S2_B1', 'B2_B2', ...]` or a list like `['S1', 'S2']` we could get that same support with fewer parameters to keep track of. Either way, I agree with having some kind of feature to only download/load a subset of data.

> Also Flair-Hub provides multiple models trained with specific combinations of satellite imagery.

Would love to add these models in a follow-up PR if you're interested! And if the amount of work involved in this PR doesn't scare you off 😅 

> If we want to reduce the number of lines of code, should we remove the toy dataset completely? If a user would want a reduced dataset he could change the domain_years dict to only get a small number of department.
>
> However I think that the number of files can be greatly reduced by removing the toy dataset. Removing it already divides the number of files by two.

Up to you whether you think the toy dataset is useful. We have a bunch of other toy datasets and they are really useful for making tutorials. I think there are ways to reduce testing if both the full dataset and toy dataset have identical APIs. See EuroSAT vs. EuroSAT100 in `torchgeo/datasets/eurosat.py` and `tests/datasets/test_eurosat.py` for what I'm envisioning. For that, we reuse the same test files for both and use a for-loop over tests.

> I agree that the created data necessitates a very large number of files, I had to create them for the tests. A simple solution would be to generate data dynamically during the test run in the CI.

What if we generate data for only 2 modalities and 2–3 timestamps instead of all modalities? This would exercise the dataset and get full coverage without too many files.

torchgeo/datasets/flairhub.py
gatienc: Flair-Hub is a direct improvement over Flair#1 and Flair#2, already incorporating all their data with fixes and additional content. I don't know if there is a need to support Flair#1 and Flair#2 then.

torchgeo/datasets/flairhub.py
gatienc: There is a need to define the root folder for the Flair-Hub toy dataset, as its data is nested. However, we can significantly reduce the code in the `__init__` method by removing duplicate information.

torchgeo/datamodules/flair.py
gatienc: How would you recommend adding it to the dataset? In the `Sample` adding each split as a key and the split ("train","test","valid") as value?

gatienc:
I agree with the bands method will implement it. I'll implement the ['S1', 'S2',...] version otherwise it would be very difficult to manage the list when wanting to get most bands.

> Would love to add these models in a follow-up PR if you're interested! And if the amount of work involved in this PR doesn't scare you off 😅

I’m also keen to implement them, but given that each model version can have significantly different architectures and considering the density of the Flair-Hub code. I’m concerned that a PR covering all versions could become quite extensive. Testing, in particular, feels daunting, but I’ll keep you updated on progress.

> Up to you whether you think the toy dataset is useful. We have a bunch of other toy datasets and they are really useful for making tutorials. I think there are ways to reduce testing if both the full dataset and toy dataset have identical APIs. See EuroSAT vs. EuroSAT100 in torchgeo/datasets/eurosat.py and tests/datasets/test_eurosat.py for what I'm envisioning. For that, we reuse the same test files for both and use a for-loop over tests.

I agree it’s valuable. For implementation, I think merging the two tests is feasible. I may need to independently test the download process for each dataset.

tests/datamodules/test_flair.py
gatienc: How should we handle the tests for the official splits of the dataset then? (If we keep it in datamodules, otherwise : https://github.com/torchgeo/torchgeo/pull/3181#discussion_r2752004254)

gatienc:
I implemented the selection of only a subset of the necessary modality for the tests data but I feel like it might be better to keep them, in term of number of files it enables to go from 41 files with all enabled to 21 files with only a subset.

tests/trainers/test_segmentation.py
gatienc: Added a concatenate_modalities boolean to make it possible to concatenate all mono-temporal modalities into a simple "image" key

torchgeo/datasets/flairhub.py
gatienc: Not sure of how should be called each imagery in the `bands` parameter, furthermore wouldn't it be more appropriate to call this parameter `imagery` as bands actually selects which satellite imagery to add.
Is there a Torchgeo underlying reason why bands would be better?

torchgeo/datasets/flairhub.py
adamjstewart: FLAIR#1 and FLAIR#2 would still be useful for reproducibility and comparison, just not for new research projects.

torchgeo/datasets/flairhub.py
adamjstewart: Most other datasets use bands [B1, B2, B3, ...] or [VV, VH] or even [VV, VH, B1, B2, B3, ...]. This gives maximum flexibility for cases where you want to test a pre-trained model but it has its bands in a weird order or only a subset of bands are supported.

Only a few datasets have sensors S1, S2, etc. I would like to slowly phase these out and replace them with the above.

torchgeo/datamodules/flair.py
adamjstewart: It would work just like all other datasets with pre-defined splits. So the user would select `split='split_1'` and the dataset would only load that split.

tests/datamodules/test_flair.py
adamjstewart: The config file would include `split: 'split_1'` or something like that. We don't need to test every split, a single split is sufficient if they are all the same. We're testing on fake data anyway, so we just need to hit 100% coverage.

tests/trainers/test_segmentation.py
```diff
@@ -22,6 +23,47 @@
 from torchgeo.trainers import SemanticSegmentationTask
 
 
+def flairhub_concat_modalities_transform(sample: dict[str, Any]) -> dict[str, Any]:
```
adamjstewart: Doesn't seem to be used anymore, can this be deleted or moved to the datamodule?

torchgeo/datasets/flairhub.py
gatienc: made it a list

tests/trainers/test_segmentation.py
gatienc: Yes, I deleted it as concatenated modalities is now a datamodule parameter

gatienc:
Calling the  dataset `FLAIR-HUB` in the doc is giving this error when building: 
```api/datasets/non_geo_datasets.csv:1: ERROR: Duplicate target name, cannot be used as a unique reference: "flair-hub". [docutils]```
I am using `FLAIRHUB` for the moment in the doc, any ideas why it occurs?

**Comment by adamjstewart** [comment]
> Calling the dataset `FLAIR-HUB` in the doc is giving this error when building: `api/datasets/non_geo_datasets.csv:1: ERROR: Duplicate target name, cannot be used as a unique reference: "flair-hub". [docutils]` I am using `FLAIRHUB` for the moment in the doc, any ideas why it occurs?

In your dataset definition, you need to use a double underscore after the link name, otherwise Sphinx gets confused.

This PR needs a rebase, the dataset docs were completely redesigned in a couple different PRs.

**Comment by gatienc** [comment]
I’ve implemented all the requested changes, except for the band naming issue.

> "Most other datasets use bands [B1, B2, B3, ...] or [VV, VH] or even [VV, VH, B1, B2, B3, ...]. This gives maximum flexibility for cases where you want to test a pre-trained model but it has its bands in a weird order or only a subset of bands are supported.
> 
> Only a few datasets have sensors S1, S2, etc. I would like to slowly phase these out and replace them with the above."


For this dataset, since the same band numbers refer to different spectral ranges across sensors, I’m unsure how to uniquely name the bands.
One option could be: bands = [SPOT_B1, SPOT_B2, ... , S1_ASC_VV,...]
But this option doesn't show which dataset are monotemporal and which are multi-temporal so I assume this is better to avoid confusion
bands = [SPOT_B1, SPOT_B2, ... , TS_S1_ASC_VV, ...]
However, this approach feels overly verbose. Any suggestions or preferences for how to handle this?

**Comment by gatienc** [comment]
Also I suppose the version added should be 0.10.0 now?

File: docs/api/datasets/benchmark.csv [hunk 29]
```diff
@@ -24,6 +24,7 @@ Dataset,Task,Source,# Samples,# Classes,Size (px),Resolution (m),Bands,License
 :ref:`FAIR1M`,OD,Gaofen/Google Earth,"15,000",37,"1,024x1,024",0.3--0.8,RGB,CC-BY-NC-SA-3.0
 :ref:`Fields Of The World`,"S,I",Sentinel-2,"70795","2,3",256x256,10,MSI,Various
 :ref:`FireRisk`,C,NAIP Aerial,"91,872",7,"320x320",1,RGB,CC-BY-NC-4.0
+:ref:`FLAIR-HUB`,S,"Aerial, SPOT, Sentinel-1/2","241,100","19,23","512x512 and 256x256 and 64x64 and 10x10 and 10x10xT","0.2--10","RGB, NIR, MSI, SAR, DEM","OPEN LICENCE 2.0/LICENCE OUVERTE"
```
adamjstewart: Confirmed the license: https://github.com/IGNF/FLAIR-HUB#dataset-license

[SPDX](https://spdx.org/licenses/) has a specific identifier for this license we should use:

```suggestion
:ref:`FLAIR-HUB`,S,"Aerial, SPOT, Sentinel-1/2","241,100","19,23","512x512 and 256x256 and 64x64 and 10x10 and 10x10xT","0.2--10","RGB, NIR, MSI, SAR, DEM",etalab-2.0
```

docs/api/datasets/toys.csv
```diff
@@ -3,3 +3,4 @@ Dataset,Task,Source,# Samples,# Classes,Size (px),Resolution (m),Bands,License
 :ref:`LandCover.ai` 100,S,Aerial,"10,674",5,512x512,0.25--0.5,RGB,CC-BY-NC-SA-4.0
 :ref:`OSCD` 100,CD,Sentinel-2,24,2,"241--1,180",60,MSI,CC-BY-NC-SA-4.0
 :ref:`PASTIS` 100,I,Sentinel-1/2,"2,433",19,128x128xT,10,MSI,CC-BY-4.0
+:ref:`FLAIR-HUB` Toy,S,"Aerial, SPOT, Sentinel-1/2","250","19 or 23","512x512 and 256x256 and 64x64 and 10x10 and 10x10xT","0.2--10","RGB, NIR, MSI, SAR, DEM","OPEN LICENCE 2.0/LICENCE OUVERTE"
```
adamjstewart: Same here:

```suggestion
:ref:`FLAIR-HUB` Toy,S,"Aerial, SPOT, Sentinel-1/2","250","19 or 23","512x512 and 256x256 and 64x64 and 10x10 and 10x10xT","0.2--10","RGB, NIR, MSI, SAR, DEM",etalab-2.0
```

adamjstewart: Can you also sort these alphabetically?

tests/datasets/test_flair.py
```diff
@@ -0,0 +1,311 @@
...
+            product(
+                [FLAIRHUB, FLAIRHUBToy],
+                [
+                    (['AERIAL_RGBI', 'SENTINEL2_TS'], t)
+                    for t in ['land_cover', 'crop_type', 'crop_type_2', 'crop_type_3']
+                ],
+            )
+        )
+        + [(cls, (['AERIAL_RGBI'], 'land_cover')) for cls in [FLAIRHUB, FLAIRHUBToy]]
+        + [(cls, (None, 'land_cover')) for cls in [FLAIRHUB, FLAIRHUBToy]]
```
adamjstewart: This seems really complicated. Can we test a simple subset that ensures 100% coverage?

tests/datasets/test_flair.py
```diff
@@ -0,0 +1,311 @@
...
+
+        if base_class is FLAIRHUB:
+            monkeypatch.setattr(FLAIRHUB, 'download_link', str(FLAIRHUB_TEST_DATA_DIR))
+            return FLAIRHUB(
+                root=tmp_path,
+                transforms=transforms,
+                download=True,
+                bands=bands,
+                dataset_type=dataset_type,
+                domain_years=FLAIRHUB_DOMAIN_YEARS,
```
adamjstewart: If you remove this single line, then the entire if-statement disappears and you can instead use `base_class(root=...)` instead of `FLAIRHUB(root=...)`.

tests/datasets/test_flair.py
```diff
@@ -0,0 +1,311 @@
...
+            return FLAIRHUBToy(
+                root=tmp_path,
+                transforms=transforms,
+                download=True,
+                bands=bands,
+                dataset_type=dataset_type,
+            )
+
+    def test_len(self, dataset: FLAIRHUB | FLAIRHUBToy) -> None:
+        assert len(dataset) > 0
```
adamjstewart: I would rather assert a specific length, we can play around with the fake test data and test params to ensure that this number is always the same.

tests/datasets/test_flair.py
```diff
@@ -0,0 +1,311 @@
...
+
+    @pytest.mark.parametrize(
+        'dataset_type,bands,suptitle',
+        [
+            ('land_cover', None, 'All modalities'),
+            ('crop_type', ['AERIAL_RGBI'], 'LPIS 1 (crop_type)'),
+            ('crop_type_2', ['AERIAL_RGBI'], 'LPIS 2 (crop_type_2)'),
+            ('crop_type_3', ['AERIAL_RGBI'], 'LPIS 3 (crop_type_3)'),
+        ],
+    )
```
adamjstewart: We could inherit the dataset fixture instead of parametrizing this separately.

tests/datasets/test_flair.py
```diff
@@ -0,0 +1,311 @@
...
+            bands=bands,
+            dataset_type=dataset_type,
+            domain_years=FLAIRHUB_DOMAIN_YEARS_SINGLE,
+        )
+        x = dataset[0]
+        fig = dataset.plot(x, suptitle=suptitle)
+        assert fig is not None
+        plt.close()
+
+    def test_getitem_with_transforms(self, tmp_path: Path) -> None:
```
adamjstewart: We don't test this for any other datasets, could remove if it's not needed for coverage

tests/datasets/test_flair.py
```diff
@@ -0,0 +1,311 @@
...
+
+        x = no_transform_dataset[0]
+        x_transformed = transformed_dataset[0]
+        assert (
+            abs(x_transformed['AERIAL_RGBI'].max() - x['AERIAL_RGBI'].max() * 2.0)
+            < 1e-6
+        )
+
+
+class TestFLAIRHUBSpecific:
```
adamjstewart: Would be nice if FLAIRHUB and FLAIRHUBToy were similar enough that these weren't needed.

tests/tasks/test_segmentation.py
```diff
@@ -94,7 +98,6 @@ def test_trainer(
         match name:
             case 'ftw':
                 pytest.importorskip('pyarrow')
-
```
adamjstewart, line 100: Can you undo this change?

torchgeo/datamodules/flair.py
```diff
@@ -0,0 +1,325 @@
...
+    Implements domain-based train/val/test splits to ensure that samples from
+    the same domain stay together in the same split. This prevents data leakage
+    between train/validation, and test sets.
+
+    If ``official_splits`` is provided, a file containing the official splits
+    will be downloaded from HuggingFace. You can choose between ``split_1``,
+    ``split_2``, ``split_3``, ``split_4``, ``split_5``, and
+    ``split_flairchallenge``. If no ``official_splits`` is provided, the splits
+    method will be the same as the research paper but with a different random
+    distribution.
```
adamjstewart: I would rather drop the random option and stick with one of the many official splits.

torchgeo/datamodules/flair.py
```diff
@@ -0,0 +1,325 @@
...
+
+    The first split (train+val vs test) is domain-based. The second split
+    (train vs val) is random and ignores domain boundaries.
+
+    The default splits are: 16 test domains, 80% train / 20% val split.
+
+    The exact splits used in the research paper can be found at:
+    ``https://hf.co/datasets/IGNF/FLAIR-HUB/resolve/e8ed7981d488508aa70bb05c37cf6585432b7d5f/data/GLOBAL_ALL_MTD.zip``
+    in the file ``GLOBAL_ALL_MTD_SPLIT.gpkg``.
```
adamjstewart: This might be too much detail for a docstring, the dataset download should handle this automatically anyway

torchgeo/datamodules/flair.py
```diff
@@ -0,0 +1,325 @@
...
+
+    .. versionadded:: 0.10.0
```
adamjstewart:
```suggestion
    .. versionadded:: 0.10
```
New features are always minor versions, no need to include patch versions

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1384 @@
...
+)
+
+
+class FLAIRHUB(NonGeoDataset):
+    """FLAIR-HUB Dataset.
+
+    Large-scale Multimodal Dataset for Land Cover and Crop Mapping dataset.
+
+    `FLAIR-HUB <https://github.com/IGNF/FLAIR-HUB>`__ builds upon and includes the FLAIR#1 and FLAIR#2 datasets,
+    expanding them into a unified,large-scale,
```
adamjstewart:
```suggestion
    expanding them into a unified, large-scale,
```

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1384 @@
...
+
+    # SENTINEL1_ASC_TS (Sentinel-1 ASC time series)
+    sentinel1_asc_ts_bands = ('VV', 'VH')
+    # SENTINEL1_DESC_TS (Sentinel-1 DESC time series)
+    sentinel1_desc_ts_bands = ('VV', 'VH')
+
+    download_link = 'https://hf.co/datasets/IGNF/FLAIR-HUB/resolve/e8ed7981d488508aa70bb05c37cf6585432b7d5f/data'
+
+    splits = ('train', 'val', 'test')
+    valid_split_columns: ClassVar[tuple[str, ...]] = (
```
adamjstewart:
```suggestion
    valid_split_columns = (
```
Tuples are immutable, no need to mark them as such

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1384 @@
...
+        'AERIAL-RLT_PAN',
+        'SENTINEL2_TS',
+        'SENTINEL2_MSK-SC',
+        'SENTINEL1-ASC_TS',
+        'SENTINEL1-DESC_TS',
+    ]
+
+    # Note: the original dataset contains 19 classes, but the dataset paper
+    # suggests not using clear cut, ligneous & mixed as they are nearly empty
+    cosia: ClassVar[dict[str, Any]] = {
```
adamjstewart: If you split classes and cmap into separate attributes, they would be much easier to type hint without resorting to Any or TypedDict. I'm trying to get rid of most uses of Any throughout the codebase.

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1384 @@
...
+                '#000000',  # background
+            ]
+        ),
+    }
+
+    def __init__(
+        self,
+        root: Path = 'data',
+        split: str = 'train',
+        split_column: str | None = None,
```
adamjstewart:
```suggestion
        split_column: Literal[
            'split_1',
            'split_2',
            'split_3',
            'split_4',
            'split_5',
            'split_flairchallenge',
        ] = 'split_1',
```
I don't think None or arbitrary strings should be allowed, let's only support official splits.

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1384 @@
...
+                '#000000',  # mixed crops
+                '#000000',  # background
+            ]
+        ),
+    }
+
+    def __init__(
+        self,
+        root: Path = 'data',
+        split: str = 'train',
```
adamjstewart:
```suggestion
        split: Literal['train', 'val', 'test'] = 'train',
```

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1384 @@
...
+            dataset_type: Type of labels to use. Choose ``land_cover`` for
+                19-class COSIA labels or ``crop_type`` for 23-class LPIS crop
+                classification labels (baseline) or ``crop_type_2`` for 31-class
+                LPIS crop classification labels or ``crop_type_3`` for 46-class
+                LPIS crop classification labels. ``crop_type_2`` and
+                ``crop_type_3`` are deeper levels of the LPIS crop classification
+                labels.
+            domain_years: Optional dictionary mapping domain names to lists of years.
+                If provided, restricts the dataset to only these domain-year combinations.
+                Defaults to None, which uses all available domain-year combinations.
```
adamjstewart: Is this something you need or is officially recommended? Trying to find ways to make the dataset simpler, could remove this.

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1384 @@
...
+        """Return an index within the dataset.
+
+        Args:
+            index: index to return
+
+        Returns:
+            Dictionary containing tensors for each modality. Keys are the
+            modality names: ``mask``, ``AERIAL_RGBI``, ``SPOT_RGBI``,
+            ``DEM_ELEV``, ``AERIAL-RLT_PAN``, ``SENTINEL2_TS``,
+            ``SENTINEL2_MSK-SC``, ``SENTINEL1-ASC_TS``, ``SENTINEL1-DESC_TS``.
```
adamjstewart: Note that this won't be compatible with any other datasets/datamodules/trainers in TorchGeo...

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1384 @@
...
+            filename_stem = label_path.stem
+            patch_id = filename_stem.replace(f'{label_dir}_', '')
+            if allowed_patch_ids is not None and patch_id not in allowed_patch_ids:
+                continue
+            if allowed_domain_years is not None:
+                domain_year = patch_id.split('_')[0]
+                if domain_year not in allowed_domain_years:
+                    continue
+
+            file_dict: dict[str, Any] = {}
```
adamjstewart: Trying to avoid `Any` because it disables the type checker, could we use a TypedDict here?

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1384 @@
...
+            the time series as tensor
+        """
+        with rasterio.open(str(path)) as f:
+            tensor = array_to_tensor(f.read()).float()
+
+            # Reshape from (T*C) x H x W to T x C x H x W
+            c = num_bands
+            tc, h, w = tensor.shape
+            t = tc // c
+            tensor = tensor.view(t, c, h, w)
```
adamjstewart: More clear to use:
```python
tensor = einops.reshape(tensor, '(t c) h w -> t c h w', c=num_bands)
```

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1384 @@
...
+                (23 classes).
+
+        Raises:
+            DatasetNotFoundError: If dataset is not found and *download* is False.
+            ValueError: If *dataset_type* is not ``land_cover`` or ``crop_type``.
+            FileNotFoundError: Requested modality file is missing from the toy dataset.
+
+        Warning:
+            This is a TOY DATASET for testing only. Do not use for actual training or
+            evaluation. Use the full FLAIRHUB dataset for research purposes.
```
adamjstewart: Toy datasets can also be useful for few-shot learning.

adamjstewart (changes requested):
Apologies for the delayed review, I think I'm finally caught up now from vacation.

I realize that FLAIR-HUB is far from a simple dataset, but your current flair.py is more than double the length of the previously longest dataset (satlas.py). I think this dataset is pretty comparable to SatlasPretrain in terms of complexity and scope? If we could somehow cut the number of SLOCs in half it would make this much easier to maintain.

torchgeo/datamodules/flair.py
adamjstewart: This parameter can now be removed from the datamodule, it will be used via `**kwargs` and passed directly to the dataset.

torchgeo/datasets/flair.py
gatienc: Was convenient to scope the data downloaded. I guess I can remove it to make the implementation simpler

torchgeo/datasets/flair.py
gatienc: It means that I also have to add split_toy so the typecker accepts it as split_column is already declared by the main datamodule

torchgeo/datasets/flair.py
gatienc: Do you have an idea on how to provide the data in a "torchgeo" way, A full image with all the modality is not possible as the shape of every modality is very different (  variable temporal data, and very variable size). I am very interested to know your opinion on it.

adamjstewart: Hmm, I guess there is no standard way of handling multi-modal datasets if images can't be stacked into a single image. Unfortunately this means it won't be possible to use any existing trainers in TorchGeo. For Kornia, you would want to prefix each modality name with `mask` or `image` to make it clear which augmentations to apply. Perhaps `image_sentinel1_asc`?

gatienc:
I tried to reduce the number of SLOC as much as I could, I agree that dataset/flair.py is a very long file : 1204 LOC however on those 1204 LOC the dataset code actually starts only from Line 647, everything before is defining GLOBAL values ( color map, bands etc) and documenting the dataset info.

tests/data/flair/data.py
gatienc: There is now 10 files for 1 domain to be able to test the plot method with all modalities. The split file split the data in "train","valid" and "test" so there is 2 other domains used in the test trainers. 

A total of of 22 image are generated : 10 for the first domain for potting and 6 for test training.
Because of the dataset download format ( 1 zip per domain and doubled because of flair-hub toy) but In my opinion it is necessary so we don't have to add lot of path manipulation. 

Compared to last iteration, to keep the code simple I had to re-add DEM and AERIAL-RLT_PAN to the dataset for test training so we don't have to add test coverage on their normalization separately.

Let me know if I should implement it differently

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1203 @@
...
+        split: Literal['train', 'val', 'test'] = 'train',
+        split_column: Literal[
+            'split_1',
+            'split_2',
+            'split_3',
+            'split_4',
+            'split_5',
+            'split_flairchallenge',
+        ]
+        | Literal['split_toy'] = 'split_1',
```
gatienc: Here I had to add the  | Literal['split_toy'] even if it doesn't make sense for FLAIRHUB dataset itself (necessary for the FLAIRHUBToy to reuse base class). Do you have an idea on how to implement it in a better way?

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1203 @@
...
+
+        Args:
+            index: index to return
+
+        Returns:
+            Dictionary containing tensors for each modality. Keys are the
+            modality names prefixed with ``image_`` or ``mask_``: ``mask``,
+            ``image_aerial_rgbi``, ``image_spot_rgbi``, ``image_dem_elev``,
+            ``image_aerial_rlt_pan``, ``image_sentinel2``,
+            ``mask_sentinel2_snowcloud``, ``image_sentinel1_asc``,
```
gatienc, line 1436: I called the sentinel 2 snow cloud data as a mask here? Should I use a different name?

gatienc:
To do temporal embedding, the date of year are necessary, I should probably add a list of date of year for temporal data as a key in sample.

docs/api/datasets/toys.csv
adamjstewart, line 1: Usually when git says every line changed, it's due to a change in line ending. Can you double check this?

tests/datasets/test_flair.py
```diff
@@ -0,0 +1,159 @@
...
+from torch import Tensor
+
+from torchgeo.datasets import FLAIRHUB, DatasetNotFoundError, FLAIRHUBToy
+from torchgeo.datasets.flair import AvailableBands
+
+_TEST_DATA = Path('tests') / 'data' / 'flair'
+_DOMAIN_YEARS = {'D006': ['2020'], 'D012': ['2019'], 'D032': ['2019']}
+
+
+class TestFLAIRHUB:
```
adamjstewart, line 23: Nothing "wrong" with these tests, but they don't look anything like our tests for other datasets, meaning they're harder to maintain than they need to be. See the EuroSAT tests for a good example where we parametrize the dataset fixture over all classes and test basic features like getitem/len/plot, download (good and bad), and any other weird custom stuff.

tests/trainers/test_segmentation.py
```diff
@@ -96,6 +98,11 @@ def test_trainer(
             case 'ftw':
                 pytest.importorskip('pyarrow')
 
+        if name in ('flairhub_multimodal_landcover', 'flairhub_toy_croptype'):
```
adamjstewart: Would be nice if this wasn't required, but I'm okay with this if there isn't an easy way to avoid this. Maybe:

```suggestion
        if name.startswith('flairhub'):
```

torchgeo/datamodules/flair.py
```diff
@@ -0,0 +1,214 @@
...
+            ]
+            if present_keys:
+                max_resolution = max(batch[k].shape[-1] for k in present_keys)
+
+                # Resize and concatenate modalities for each sample in batch
+                concatenated_modalities = []
+                for batch_key in present_keys:
+                    tensor = batch[batch_key]  # [B, C, H, W]
+                    if tensor.shape[-1] != max_resolution:
+                        tensor = TF.interpolate(
```
adamjstewart: Kornia/Pillow also support interpolation. We're trying to reduce our dependence on torchvision as it doesn't seem to be well-maintained anymore.

torchgeo/datamodules/flair.py
```diff
@@ -0,0 +1,214 @@
...
+        num_workers: int = 0,
+        official_splits: Literal[
+            'split_1',
+            'split_2',
+            'split_3',
+            'split_4',
+            'split_5',
+            'split_flairchallenge',
+            'split_toy',
+        ] = 'split_1',
```
adamjstewart: Remove this, it can be a **kwarg via `split_column`

torchgeo/datamodules/flair.py
```diff
@@ -0,0 +1,214 @@
...
+            'SPOT_RGBI': K.Normalize(
+                mean=SPOT_RGBI_MEAN, std=SPOT_RGBI_STD, keepdim=True
+            ),
+            'DEM_ELEV': K.Normalize(mean=DEM_ELEV_MEAN, std=DEM_ELEV_STD, keepdim=True),
+            'AERIAL-RLT_PAN': K.Normalize(
+                mean=AERIAL_RLT_PAN_MEAN, std=AERIAL_RLT_PAN_STD, keepdim=True
+            ),
+        }
+
+    def setup(self, stage: str) -> None:
```
adamjstewart: Once you use `**kwargs` for `split_column`, you can delete this method and use the base class implementation

torchgeo/datamodules/flair.py
```diff
@@ -0,0 +1,214 @@
...
+        Args:
+            batch_size: Size of each mini-batch.
+            num_workers: Number of workers for parallel data loading.
+            official_splits: Official splits from the research paper.
+                Choose from 'split_1', 'split_2', 'split_3', 'split_4', 'split_5',
+                or 'split_flairchallenge'.
+            concatenate_modalities: If True, concatenate all mono-temporal modalities
+                (AERIAL_RGBI, SPOT_RGBI, DEM_ELEV, AERIAL-RLT_PAN) into an 'image' key.
+                Modalities will be resized to the maximum resolution before concatenation.
+                If False, modalities remain separate. Defaults to False.
```
adamjstewart, line 59: Neat workaround, but it kinda doubles the complexity to support both. We wouldn't need to override `on_after_batch_transfer` at all if we just always resize/concatenate. Could even do this directly in the dataset and call it a day. I'm not sure who is using FLAIR-HUB in its native resolution in a multi-modal way, it would require a custom Frankenstein model architecture. Can we get the FLAIR folks to comment on this PR and see what they think?

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1212 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+
+"""FLAIRHUB dataset."""
```
adamjstewart:
```suggestion
"""FLAIR (French Land cover from Aerospace ImageRy) datasets."""
```
We might add #1 and #2 someday, not just HUB.

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1212 @@
...
+    - Total Dataset Size: ~750 GB
+
+    Dataset structure:
+
+    The dataset is organized by domains (geographical areas) and years.
+    Each domain-year combination has multiple modalities available for
+    download.
+    The full dataset contains
+    66 unique domains (D004-D091, non-consecutive), with years ranging from 2017-2022.
+    Most domains have data for a single year, but some have multiple years available.
```
adamjstewart, line 381: Formatting is kinda weird, can we combine both sentences in one paragraph?

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1212 @@
...
+    - ``SENTINEL2_TS``: Sentinel-2 multispectral time series (12 bands, 10m) — key: ``image_sentinel2``
+    - ``SENTINEL2_MSK-SC``: Sentinel-2 scene classification mask — key: ``mask_sentinel2_snowcloud``
+    - ``AERIAL_LABEL-COSIA``: Land cover labels (19 classes)
+    - ``ALL_LABEL-LPIS``: Crop type labels (23 classes)
+
+    Automatic download:
+
+    Set ``download=True`` to automatically download requested modalities from
+    HuggingFace. Only the modalities you select will be downloaded, saving
+    bandwidth and storage.
```
adamjstewart: Not needed, this is true of every dataset in TorchGeo and torchvision, it should be self-explanatory

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1212 @@
...
+            'split_1',
+            'split_2',
+            'split_3',
+            'split_4',
+            'split_5',
+            'split_flairchallenge',
+        ]
+        | Literal['split_toy'] = 'split_1',
+        transforms: Callable[[Sample], Sample] | None = None,
+        download: bool = False,
```
adamjstewart, line 1323: We should probably also add sha256 checksums and `checksum: bool = False`. I know this will make the file longer, but it will also be more secure. Note that I have a separate PR to convert all of our existing HF downloads from MD5 to SHA256. Luckily HF tells you the SHA256 of every file so you don't need to download anything.

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1212 @@
...
+            tensor = array_to_tensor(f.read()).float()
+            # Reshape from (T*C) * H * W to T * C * H * W
+            tensor = rearrange(tensor, '(t c) h w -> t c h w', c=num_bands)
+        return tensor
+
+    def _load_image(self, path: Path) -> Tensor:
+        """Load a single image.
+
+        Args:
+            path: path to the image
```
adamjstewart, line 1618: I think you need a space between sections? Sphinx isn't complaining because it isn't rendering docs for hidden functions, but still worth fixing.

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1212 @@
...
+            title: Title for the subplot
+        """
+        # T * C * H * W -> take last timepoint, VV band (index 0)
+        vv = data[-1, 0].numpy()
+        p2, p98 = np.percentile(vv, (2, 98))
+        vv_norm = np.clip((vv - p2) / (p98 - p2 + 1e-6), 0, 1)
+        ax.imshow(vv_norm, cmap='gray')
+        ax.set_title(title)
+
+    def plot(self, sample: Sample, suptitle: str | None = None) -> Figure:
```
adamjstewart, line 1812: Feels like flair.py is 50% plotting. That's _okay_, but it might be difficult to maintain someday.

torchgeo/datasets/flair.py
```diff
@@ -0,0 +1,1212 @@
...
+
+        Args:
+            data: Sentinel-1 time series tensor (T, C, H, W)
+            ax: Matplotlib axes to plot on
+            title: Title for the subplot
+        """
+        # T * C * H * W -> take last timepoint, VV band (index 0)
+        vv = data[-1, 0].numpy()
+        p2, p98 = np.percentile(vv, (2, 98))
+        vv_norm = np.clip((vv - p2) / (p98 - p2 + 1e-6), 0, 1)
```
adamjstewart: quantile_normalization

adamjstewart (commented):
I think we're getting closer here. The dataset is still quite complex, but I'm also close to giving up trying to make it simpler. If we can quickly resolve the remaining change requests before I become swamped with work again, I think we can merge and save additional improvements for future PRs. This dataset is obviously very important, and we would love to have it in TorchGeo, even if it starts out complex and becomes simpler later.

torchgeo/datasets/flair.py
adamjstewart: Only way to avoid this would be to have a FLAIRHUBBase dataset with 99% of the code and two tiny FLAIRHUB and FLAIRHUBToy datasets that subclass this.

torchgeo/datasets/flair.py
adamjstewart: Nah, feels like a mask to me.

tests/trainers/test_segmentation.py
gatienc: I put it in the case as I guess is the most expensible method

tests/trainers/test_segmentation.py
```diff
@@ -97,6 +99,10 @@ def test_trainer(
         match name:
             case 'ftw':
                 pytest.importorskip('pyarrow')
+            case _ if name.startswith('flairhub'):
```
adamjstewart: Seems strange, I would use both names if you want to use case

adamjstewart (commented):
Can you rebase? We renamed trainers to tasks and added a lot of new ruff checks.

torchgeo/datamodules/flair.py
agarioud: Even for modotemporal modalities, we usually use separate backbones to avoid mixing information coming from distinct acquisition dates and having different spectral bandwidths before fusing the representations.

Nonetheless, you are right about the risk of a Frankenstein architecture, and if simplicity is key, since all patches cover the same geographical location, interpolation is possible. No strong opinion on this.

adamjstewart:
Moved this to 0.11 because it seems like there's still a lot of work to do, but if you happen to finish in the next couple days, we could still sneak this into 0.10.

gatienc:
I implemented the changes to the tests to be more similar to the EuroSat one, so I think for the Dataset and Test are soon to be done.
The output of the datamodule remains to be discussed @adamjstewart .

tests/tasks/test_segmentation.py
```diff
@@ -97,6 +99,14 @@ def test_trainer(
         match name:
             case 'ftw':
                 pytest.importorskip('pyarrow')
+            case 'flairhub_multimodal_landcover':
+                monkeypatch.setattr(
+                    FLAIRHUB, 'domain_years', {'D006': ['2020'], 'D012': ['2019']}
+                )
+            case 'flairhub_toy_croptype':
```
adamjstewart:
```suggestion
            case 'flairhub_multimodal_landcover' | 'flairhub_toy_croptype':
```

adamjstewart (dismissed):
I think this is about as simple as it's going to get. If one other maintainer can approve, I think we should merge and let people play with it and report bugs.

torchgeo/datamodules/flair.py
```diff
@@ -0,0 +1,171 @@
...
+
+import kornia.augmentation as K
+import torch
+from kornia.geometry.transform import resize
+
+from ..datasets import FLAIRHUB, FLAIRHUBToy
+from ..datasets.utils import Sample
+from .geo import NonGeoDataModule
+
+# From the research paper : https://arxiv.org/abs/2506.07080 p16
```
nilsleh, line 16: I know the differences are quite minor in this case but looking at the referenced table, it seemes that the values here are from the "ALL" partition, when instead normalization statistics should come from the train split only to prevent leakage of information from test set into training process. So I would recommend using the "TRAIN" partition values from the first column in that table

nilsleh (dismissed):

nilsleh:
Otherwise this also LGTM, and rely on people testing this in their training procedures.

torchgeo/datamodules/flair.py
agarioud: Indeed ALL should be discarded, but benchmark metrics from the different papers are using TRAIN+VAL statistics.

isaaccorley: Something to follow up in another PR I think if people feel strongly about it

isaaccorley (approved):

PR #3185, opened 2025-12-09
NLCD: fix download, v0 -> v1

### LINKED ISSUE #2877: NLCD - Dataset not found in `paths='data'`
### Description

With `download=True` a 42 byte tif is downloaded, but the error `DatasetNotFoundError: Dataset not found in `paths='data'` and cannot be automatically downloaded, either specify a different `paths` or manually download the dataset.` is raised

### Steps to reproduce

```python
from torchgeo.datasets import NLCD

dataset = NLCD(
    years=[2023], 
    download=True, 
    checksum=False
)
```

### Version

0.7.1

gatienc:
As discussed in #2877, this PR updates the link to download the NLCD dataset.

The dataset now provides a zip file containing the TIF file, so I used a similar method as in the Chesapeake dataset and updated the test file accordingly.

The dataset version on their website has been incremented. I haven't analyzed the changes further, but I tested the dataset by plotting it.

~2024 data has also been added.~

Closes #2877


adamjstewart:
Can we focus this PR on fixing downloads and open a separate PR to add 2024 data? The former can go into a 0.8.1 release soon but the latter should be saved for 0.9.0 since it changes a default.

gatienc:
Oh nice catch, indeed, I haven't thought about that

adamjstewart:
I think this can be made simpler. See CDL as an example.

gatienc:
Indeed the CDL implementation is simpler!

Regarding the test coverage, should we add a test case to ensure an error is raised when the checksum is invalid?

adamjstewart:
I would just copy the CDL tests.

tests/data/nlcd/Annual_NLCD_LndCov_2019_CU_C1V1.zip
adamjstewart, line 1: If any changes need to be made to the testing data, can you update the `data.py` script to generate these?

gatienc: Updated data.py to change the file version in the filename to match the new dataset and added the generation of the zipfile.

adamjstewart (approved):
Looks much cleaner now, thanks for the fix!

PR #3186, opened 2025-12-10
Custom Raster Datasets: add examples of dtype/resampling overrides

Moritz-Langer:
Following the suggestion of @adamjstewart in https://github.com/torchgeo/torchgeo/pull/2015 I added the example to override the resampling.

### DIFF
--- a/docs/tutorials/custom_raster_dataset.ipynb
+++ b/docs/tutorials/custom_raster_dataset.ipynb
@@ -276,6 +276,7 @@
     "    filename_regex = r'^.{6}_(?P<date>\\d{8}T\\d{6})_(?P<band>B0[\\d])'\n",
     "    date_format = '%Y%m%dT%H%M%S'\n",
     "    is_image = True\n",
+    "    dtype = torch.float64\n",
     "    resampling = Resampling.cubic\n",
     "    separate_files = True\n",
     "    all_bands = ('B02', 'B03', 'B04', 'B08')\n",


docs/tutorials/custom_raster_dataset.ipynb
```diff
@@ -275,6 +275,7 @@
     "    filename_regex = r'^.{6}_(?P<date>\\d{8}T\\d{6})_(?P<band>B0[\\d])'\n",
     "    date_format = '%Y%m%dT%H%M%S'\n",
     "    is_image = True\n",
+    "    resampling = \"nearest\"\n",
```
adamjstewart: Nearest isn't great for imagery. We default to bilinear, but maybe we could use cubic instead?

I would use:
```python
from rasterio.enums import Resampling
...
resampling = Resampling.cubic
```
it isn't clear if strings work.

calebrob6: @adamjstewart you suggested this change in https://github.com/torchgeo/torchgeo/pull/2015#issuecomment-3612028618 -- what's the difference?

adamjstewart: My suggestion was wrong 🙂

ashnair1:
Hi @Moritz-Langer, could you add the suggested change? We're planning on a `0.8.1` release soon and would love to include this PR.

adamjstewart:
Tutorial test failure is unrelated, either a transient issue with MPC or something requiring debugging. Will be good to fix that before the release anyway.

adamjstewart:
Seems like MPC is down today. I would suggest merging this once all required tests pass and worrying about MPC during the release process.

PR #3190, opened 2025-12-11
CropHarvest: add test split and update version

yichiac:
- [x] Update from v13 to v14
- [x] Add split for train and test


yichiac:
Two questions:
1. The dataset has `is_test` indicating whether or not the point is part of the test data [[source](https://github.com/nasaharvest/cropharvest/tree/main/data)]. Shall we use it for split?
2. Shall we use `NonGeoClassificationDataset`? The dataset currently use `NonGeoDataset`.

yichiac:
Since it's the 1-d time-series dataset, the current trainers don't support this dataset. Thus, this PR could be just for adding the datamodule.

adamjstewart:
> Two questions:
> 
> 1. The dataset has `is_test` indicating whether or not the point is part of the test data [[source](https://github.com/nasaharvest/cropharvest/tree/main/data)]. Shall we use it for split?
> 2. Shall we use `NonGeoClassificationDataset`? The dataset currently use `NonGeoDataset`.

1. Yes, a `split: Literal['train', 'test'] = 'train'` parameter would be great.
2. The former is a convenience base class that is only useful if the dataset follows the expected file hierarchy, the latter is more flexible if it doesn't.

torchgeo/datasets/cropharvest.py
```diff
@@ -46,13 +47,13 @@ class CropHarvest(NonGeoDataset):
...
 
-    * https://openreview.net/forum?id=JtjzUXPEaCu
+    * https://datasets-benchmarks-proceedings.neurips.cc/paper_files/paper/2021/hash/54229abfcfa5649e7003b83dd4755294-Abstract-round2.html
 
     This dataset requires the following additional library to be installed:
 
        * `h5py <https://pypi.org/project/h5py/>`_ to load the dataset
 
-    .. versionadded:: 0.6
+    .. versionadded:: 0.9
```
adamjstewart: This shouldn't change.

torchgeo/datamodules/cropharvest.py
```diff
@@ -0,0 +1,41 @@
...
+"""CropHarvest datamodule."""
+
+from typing import Any
+
+from torchgeo.datasets import CropHarvest
+
+from .geo import GeoDataModule
+
+
+class CropHarvestDataModule(GeoDataModule):
```
adamjstewart: If I understand correctly, the dataset returns data in `T x C` format? We don't yet have a trainer for this, so it may be best to wait until #2639 is merged before adding this datamodule.

torchgeo/trainers/classification.py
```diff
@@ -64,6 +65,9 @@ class and used with 'ce' loss.
             freeze_backbone: Freeze the backbone network to linear probe
                 the classifier head.
 
+        .. versionadded:: 0.9
+           Time-series support.
```
adamjstewart: Not sure if we yet have a dataset to test this. Might need to leave this for a future PR.

torchgeo/datamodules/cropharvest.py
yichiac: Yes, the dataset returns `TxC`. Since we don't have a trainer for this, let's wait until #2638 is merged for the datamodule.

tests/datasets/test_cropharvest.py
```diff
@@ -17,8 +17,17 @@
...
+    def dataset(
+        self, monkeypatch: MonkeyPatch, tmp_path: Path, request: pytest.FixtureRequest
+    ) -> CropHarvest:
+        split = request.param
+        monkeypatch.setitem(
+            CropHarvest.file_dict['features'], 'md5', 'ef6f4f00c0b3b50ed8380b0044928572'
+        )
+        monkeypatch.setitem(
+            CropHarvest.file_dict['labels'], 'md5', 'f990f1bdfd9dc7efe99e94a9f511efde'
+        )
```
adamjstewart, line 30: Let's remove md5 monkeypatching here, see #3367

tests/datasets/test_cropharvest.py
```diff
@@ -33,7 +42,9 @@ def dataset(self, monkeypatch: MonkeyPatch, tmp_path: Path) -> CropHarvest:
         root = tmp_path
         transforms = nn.Identity()
 
-        dataset = CropHarvest(root, transforms, download=True)
+        dataset = CropHarvest(
+            root, split=split, transforms=transforms, download=True, checksum=True
```
adamjstewart, line 46: Same here, no checksum

tests/data/cropharvest/labels.geojson
adamjstewart, line 1: Are these changes necessary? If so you should also update the data.py script that generates them.

tests/datasets/test_cropharvest.py
```diff
@@ -42,11 +53,12 @@ def test_getitem(self, dataset: CropHarvest) -> None:
         assert isinstance(x['array'], torch.Tensor)
         assert isinstance(x['label'], torch.Tensor)
         assert x['array'].shape == (12, 18)
-        y = dataset[2]
-        assert y['label'] == 1
 
     def test_len(self, dataset: CropHarvest) -> None:
-        assert len(dataset) == 5
+        if dataset.split == 'train':
```
adamjstewart, line 58: If we make the fake test data simpler, we could have these be the same length and the tests will become simpler too

tests/datasets/test_cropharvest.py
```diff
@@ -42,11 +53,12 @@ def test_getitem(self, dataset: CropHarvest) -> None:
         assert isinstance(x['array'], torch.Tensor)
         assert isinstance(x['label'], torch.Tensor)
         assert x['array'].shape == (12, 18)
-        y = dataset[2]
-        assert y['label'] == 1
```
adamjstewart, line 46: Why remove these tests?

torchgeo/datasets/cropharvest.py
```diff
@@ -168,17 +174,24 @@ def _load_features(self, root: Path) -> list[dict[str, str]]:
...
         """
+        is_test = self.split == 'test'
         files = []
         chips = glob.glob(
             os.path.join(root, self.file_dict['features']['extracted_filename'], '*.h5')
         )
         chips = sorted(os.path.basename(chip) for chip in chips)
         for chip in chips:
+            index = chip.split('_')[0]
+            dataset = chip.split('_')[1][:-3]
```
adamjstewart, line 185: Can make this faster by moving index/dataset back where they were, after the possible skip.

torchgeo/datasets/cropharvest.py
```diff
@@ -168,17 +174,24 @@ def _load_features(self, root: Path) -> list[dict[str, str]]:
...
+            row = self.labels[
+                (self.labels['properties.index'] == int(index))
+                & (self.labels['properties.dataset'] == dataset)
+            ]
+            if row.empty or row.iloc[0]['properties.is_test'] != is_test:
```
adamjstewart, line 190: This property was newly added in v14?

torchgeo/datasets/cropharvest.py
adamjstewart, line 297: This references a key that no longer exists, should either change to sha256 or keep md5

torchgeo/datasets/cropharvest.py
```diff
@@ -78,26 +79,29 @@ class CropHarvest(NonGeoDataset):
...
+        'https://zenodo.org/records/10251170/files/features.tar.gz?download=1'
+    )
+    labels_url = 'https://zenodo.org/records/10251170/files/labels.geojson?download=1'
     file_dict: ClassVar[dict[str, dict[str, str]]] = {
         'features': {
             'url': features_url,
             'filename': 'features.tar.gz',
             'extracted_filename': os.path.join('features', 'arrays'),
-            'md5': 'cad4df655c75caac805a80435e46ee3e',
+            'sha256': 'f88d688bac805e1c7ae8bd518ed6d14528b3811f46c47a8f95ec0669ce772a68',
```
adamjstewart, line 91: I almost wonder whether we should keep md5, since it's the only official checksum published on Zenodo...

torchgeo/datasets/cropharvest.py
```diff
@@ -118,14 +123,15 @@ def __init__(
...
+        self.split = split
         self.transforms = transforms
         self.checksum = checksum
         self.download = download
 
         self._verify()
 
-        self.files = self._load_features(self.root)
         self.labels = self._load_labels(self.root)
+        self.files = self._load_features(self.root)
```
adamjstewart, line 134: Unnecessary change

adamjstewart (changes requested):

robmarkcole:
@yichiac whats the plan for this PR?

PR #3201, opened 2025-12-14
Python: drop 3.11 support

adamjstewart:
Now that Python 3.14 has been added to our tests, it's time to drop support for Python 3.11 in line with [SPEC 0](https://scientific-python.org/specs/spec-0000/).

Python 3.12 is now the minimum supported version. The release notes highlight the following [new typing features](https://docs.python.org/3/whatsnew/3.12.html#new-features-related-to-type-hints) of interest to us:

* New `type` builtin to replace `typing.TypeAlias`
* `TypedDict` can be used to annotate `**kwargs` (144 of our Any uses)
* `typing.override` decorator

We can slowly integrate these in separate PRs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,8 +40,8 @@ dependencies = [
     "einops>=0.3",
     # geopandas 0.13+ required for GeoDataFrame.get_coordinates()
     "geopandas>=0.13",
-    # jsonargparse 4.25+ required for proper LightningCLI YAML parsing
-    "jsonargparse[signatures]>=4.25",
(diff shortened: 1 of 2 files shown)


adamjstewart:
Sphinx support for the `type` statement was not introduced until Sphinx 9: https://github.com/sphinx-doc/sphinx/pull/13508

Unfortunately, pytorch-sphinx-theme doesn't support Sphinx 6+. We may have to tell ruff to ignore this or wait until we change our docs theme.

PR #3203, opened 2025-12-15
CI: silence Python 3.14 warnings

adamjstewart:
Reported upstream, nothing else we can do to prepare.


isaaccorley (approved):

adamjstewart:
Your copilot review is useless, the error message still persists in CI lol.

adamjstewart:
Remaining warning happens not during pytest but during `python3 -m torchgeo`. I think we can live with that while we wait for lightly or pydantic to fix this.

PR #3210, opened 2025-12-16
BigEarthNet: only download required bands

goergen95:
The current implementation of the `BigEarthNetV2` dataset downloads files to the root directory with the md5 checksum as filename which then causes errors when the `_extract()` method is called. This PR fixes the respective call to `download_url()` ~and also adds a `BigEarthNetV2DataModule`~. Also note that the current implementation of the `BigEarthNetV2` dataset downloads both `s1` and `s2` inputs no matter what was specified in the `bands` argument. This can/should be changed in a future PR.

### DIFF
--- a/tests/datasets/test_bigearthnet.py
+++ b/tests/datasets/test_bigearthnet.py
@@ -206,6 +206,14 @@ class TestBigEarthNetV2:
             root=tmp_path, bands=dataset.bands, split=dataset.split, download=True
         )
 
+    def test_multiple_band_instances(
+        self, dataset: BigEarthNetV2, tmp_path: Path
+    ) -> None:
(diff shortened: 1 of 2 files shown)


goergen95:
This only partly resolves the issues with this particular dataset. Now `_verify()` seems to return early in case individual components were previously downloaded but not yet extracted. Hope to push a fix for this soon.

adamjstewart:
Can we separate the download bug fix and the new datamodule into separate PRs? The bug fix can be backported to 0.8.1, but the new datamodule will have to wait for 0.9.0.

nilsleh:
I checked out the PR and tested the download and verify which works! Probably just needs a rebase and fixing ruff!

adamjstewart:
@goergen95 is this PR still a WIP?

PR #3213, opened 2025-12-17
Geodatamodule predict setup

hkristen:
Adds roi and stride parameters to GeoDataModule.setup() for the predict stage, enabling GridGeoSampler-based inference over a region of interest. Also updates SemanticSegmentationTask.predict_step to return a dict with logits, bounds, and transform metadata needed for downstream spatial reconstruction.

Preparation for the tiled-inference PR

### DIFF
--- a/torchgeo/trainers/segmentation.py
+++ b/torchgeo/trainers/segmentation.py
@@ -380,7 +380,7 @@ class SemanticSegmentationTask(BaseTask):
 
     def predict_step(
         self, batch: Any, batch_idx: int, dataloader_idx: int = 0
-    ) -> dict[str, Any]:
+    ) -> dict[str, Tensor]:
         """Compute the predicted class probabilities.
 
         Args:
@@ -390,7 +390,7 @@ class SemanticSegmentationTask(BaseTask):
 
         Returns:
             Dictionary containing:
-                - logits: Model predictions (B, C, H, W) with softmax/sigmoid applied
+                - probabilities: Model predictions (B, C, H, W) with softmax/sigmoid applied
                 - bounds: Patch bounds tensor (B, 9) if available
                 - transform: Per-patch affine transform (B, 6) if available
         """
@@ -404,7 +404,7 @@ class SemanticSegmentationTask(BaseTask):
                 y_hat = y_hat.softmax(dim=1)
 
         return {
-            'logits': y_hat,
+            'probabilities': y_hat,
             'bounds': batch.get('bounds'),
             'transform': batch.get('transform'),
         }
--- a/tests/trainers/test_segmentation.py
+++ b/tests/trainers/test_segmentation.py
@@ -301,12 +301,12 @@ class TestSemanticSegmentationTask:
         batch = {'image': torch.randn(2, 3, 64, 64)}
         result = task.predict_step(batch, 0)
 
-        assert 'logits' in result
+        assert 'probabilities' in result
         assert 'bounds' in result
         assert 'transform' in result
 
-    def test_predict_step_logits_shape(self) -> None:
-        """Test that logits have correct shape."""
+    def test_predict_step_probabilities_shape(self) -> None:
+        """Test that probabilities have correct shape."""
         num_classes = 10
         task = SemanticSegmentationTask(task='multiclass', num_classes=num_classes)
         batch_size = 2
@@ -314,8 +314,8 @@ class TestSemanticSegmentationTask:
         batch = {'image': torch.randn(batch_size, 3, height, width)}
         result = task.predict_step(batch, 0)
 
-        logits = result['logits']
-        assert logits.shape == (batch_size, num_classes, height, width)
+        probabilities = result['probabilities']
+        assert probabilities.shape == (batch_size, num_classes, height, width)
 
     def test_predict_step_preserves_metadata(self) -> None:
         """Test that bounds and transform are passed through correctly."""


adamjstewart (changes requested):
Not sure if I have a strong argument for this, but I kinda feel like these would make more sense as `__init__` parameters instead of `setup` parameters. When using a `Trainer` (e.g., `torchgeo predict --config config.yaml`), `setup` gets called for us and we have no control over what arguments get passed to `setup`. We only control arguments to `__init__`.

I'm not sure if `roi` is necessary, ideally `setup` would be defined to use the right roi? This is interesting to think about, I'm not sure how useful a generic datamodule would be without the user explicitly overriding the `predict` dataset. `stride` may be okay, but I would use it for val/test/predict, not just predict.

hkristen:
You're right that setup() should follow the Lightning API since users can't control args when using Trainer.predict(). 

I moved roi and stride to __init__ as predict_roi and predict_stride - the predict_ prefix makes it explicit these only affect the predict stage, leaving room to add val_stride/test_stride later if needed. I kept predict_roi because I believe it's useful for quick subset testing during development, without needing to create a custom datamodule.

torchgeo/trainers/segmentation.py
```diff
@@ -380,7 +380,7 @@ def test_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) -> None
 
     def predict_step(
         self, batch: Any, batch_idx: int, dataloader_idx: int = 0
-    ) -> Tensor:
+    ) -> dict[str, Any]:
```
```suggestion
    ) -> dict[str, Tensor]:
```

torchgeo/trainers/segmentation.py
```diff
@@ -389,7 +389,10 @@ def predict_step(
             dataloader_idx: Index of the current dataloader.
 
         Returns:
-            Output predicted probabilities.
+            Dictionary containing:
+                - logits: Model predictions (B, C, H, W) with softmax/sigmoid applied
```
adamjstewart: "logits" are by definition raw outputs before softmax/sigmoid are applied. "probabilities" might be a better name. "predictions" would imply that argmax has been applied, which is not the case here.

hkristen:
> I'm still quite skeptical about the changes to GeoDataModule. I don't expect very many (if any) datamodules will use this since it requires the same dataset to have both train/val/test splits and a generic area you want to stitch together predictions for. More realistically, I would expect custom setup methods with custom IntersectionDataset for train/val/test and RasterDataset for predict, each with their own roi. All of our current GeoDataModule subclasses have their own custom setup method, making this less than useful.

Fair point. The use case I had in mind was using GeoDataModule directly for ad-hoc inference (the tiled inference callback builds on this), but I agree it clutters the base class.

I am happy to split this PR and see two options for the DataModule:

1. Custom setup in subclasses - users add their own predict handling with their own dataset/roi/sampler as you outlined.
2. Dedicated InferenceGeoDataModule - A simpler class specifically for prediction that doesn't have the train/val/test baggage. Could look like this: 

```python
datamodule = InferenceGeoDataModule(
      dataset_class=RasterDataset,
      patch_size=256,
      stride=192,
      roi=my_polygon,
      paths='/data/rasters/'
  )
```
-> Which one do you prefer?

adamjstewart:
> The use case I had in mind was using GeoDataModule directly for ad-hoc inference

This goes back to a fundamental design difference we have been dealing with since the start of TorchGeo. When I first designed `RasterDataset`, I never intended for anyone to use it directly. I intended it to be an ABC that everyone would subclass for their own data source. Unfortunately, it works a little too well, and many people only use `RasterDataset` and complain when it doesn't support all customization via `__init__`. I don't think we have any documentation that suggests directly using `RasterDataset`, but people do it anyway.

I still stand by requiring subclasses of `RasterDataset`/`GeoDataModule`, but I would also love to hear arguments from others against this design. At the moment, any custom dataset handling (e.g., timestamps, bands, dtype, resampling, plotting) or datamodule handling (e.g., train/val/test splits, transforms) require subclasses. Much like there are many kinds of Datasets, there are many subclasses of GeoDataset/NonGeoDataset. In my mind, this is exactly what object-oriented programming and subclasses are designed for.

I'm not worried about people from CS backgrounds (they know how to read source code and make subclasses). I'm more worried about people from non-CS backgrounds. TorchGeo should be as easy as possible for people to use.

> Which one do you prefer?

If you want to save yourself time, I would vote for sticking with the current object-oriented design and require people to make subclasses with a `predict` split that they can customize however they want. Especially for GeoDatasets, there is rarely a pre-defined train-val-test split, so we almost always need to customize this. We can add docs to demonstrate how to do this. Any movement in the direction of using ABCs for everything will likely take a lot more time and backwards-incompatible changes.

hkristen:
I think optional parameters would make it easier for non-CS users - passing predict_roi=my_polygon is way more approachable than writing a subclass with setup() override and super() calls. I've struggled with it myself when I started out with TorchGeo tbh.

I am not sure if they conflict with the subclass philosophy, no expert here - to me they are just convenience for the common predict case. Complex stuff still requires subclassing. No breaking changes, fully backwards-compatible.

Happy to add docs showing both approaches.

adamjstewart:
I just think we have too many parameters that users would almost always want to override. It's not just `predict_roi`, it's also `train/val/test/predict_dataset`, `train/val/test/predict_aug`, `train/val/test/predict_roi`, etc. I don't want to add all of those to the base class. Some type checkers will also complain if all subclasses don't support the same parameters.

My personal opinion is we should abandon this for now and focus on the stitching callback. We can always revisit this later. There have been several requests to make transforms easier to override, and this is within the same line of thought.

hkristen:
Alright, let's drop the GeoDataModule changes and keep the SemanticSegmentationTask changes - I'll split them into a separate PR ->  #3357

I second you that we should focus on the tiled inference callback.

PR #3214, opened 2025-12-17
Tiled Inference Callback

### LINKED ISSUE #30: Add stitching utilities

### LINKED ISSUE #35: Add utility to compute bounding box of resulting prediction
### Rationale

For GeoDatasets, at sampling time, we should know the CRS, bounding box, and resolution of the image that gets sampled. As the image is passed through transforms like `Resample` or `Warp`, we should be able to recompute the new CRS/bbox/res pretty easily. However, depending on the padding and stride used in convolutional and pooling layers, the bounding box and resolution may change significantly as the image is passed through the model. In order to save or stitch together our predictions, we'll need to be able to compute the new bbox/res.

### Description

We have two possible options:

1. Build a utility that takes in a `nn.Module` (the neural network) and computes the resulting bbox/res.
2. Modify PyTorch's builtin modules (at least the convolution and pooling ones) to take the bbox/res as input and modify them directly.

In the short-term, we will likely go with 1 since it involves the least work. In the long-term, we may end up going with 2 since we'll want to be able to design networks that can take advantage of this kind of geospatial information.

### LINKED ISSUE #1407: Stitching: Access patch geo-transform in callback after predict
> Now that `train.py` supports both training and inference, I think we should close this PR. I think the best path forward is to modify `predict_step` in our trainers to store and stitch together predictions for each tile. Then we don't need `PredictDataset` or another script or anything else.

_Originally posted by @adamjstewart in https://github.com/microsoft/torchgeo/issues/560#issuecomment-1522199639_

Relevant issues and PRs: https://github.com/microsoft/torchgeo/issues/30, https://github.com/microsoft/torchgeo/issues/35, https://github.com/microsoft/torchgeo/pull/411

To stitch the prediction output on [patches from the dataset](https://torchgeo.readthedocs.io/en/stable/api/samplers.html#torchgeo.samplers.GridGeoSampler) we need to transform them to a common grid/reference system to be able to stitch them.

Let's say my `predict_step`s output is relative to the resolution of the sample-patch (e.g. a seg mask or boxes/points relative to the patch resolution and size). Then the transformation to the original (full) image could be performed using [pyproj.Tranformer](https://pyproj4.github.io/pyproj/stable/api/transformer.html#pyproj.transformer.Transformer.transform) for coords, or [rasterio.window.transform](https://rasterio.readthedocs.io/en/stable/api/rasterio.windows.html#rasterio.windows.transform). These functions would also let us convert to any crs.

But then we need to reference this sample's window and source image file.

My initial thought on where to implement this was in Lightning Callback [on_predict_batch_end](https://lightning.ai/docs/pytorch/2.0.0/api/lightning.pytorch.callbacks.Callback.html#lightning.pytorch.callbacks.Callback.on_predict_batch_end). It has access to the trainer, and thus the dataset. Then, if we had a way to access the patch/window_transform and the source file, we could do this.

To my understanding, [GeoDataModule.transfer_batch_to_device](https://github.com/microsoft/torchgeo/blob/2f0458db4420f95fd7230fd4b99b4ba6c38288bd/torchgeo/datamodules/geo.py#L360) removes the reference to the patch/window due to it not being a tensor. For regular datasets, we can pass the index through. Would we need something similar here?

Thoughts? Are there any utilities for RasterDataset that I have yet to find that can help solve this?

### LINKED ISSUE #1440: Stitching feature request: Transform non-geospatial model inference back into geospatial space.
### Summary

`GridGeoSampler` allows for efficient imagery chipping in preparation for ML-based inference. My problem is converting these non-geospatial inference results back into geospatial space. E.g. an inference file in `geojson` format that contains model inference bounding boxes, segmentations , etc. 

Does this functionality already exist in `TorchGeo`? If not, how can we make this happen?

### Rationale

From my perspective, this would enable machine learning experts to more easily interact with geospatial data.

### Implementation

I have not thought deeply about implementation.

### Alternatives

Not that I am aware of. Although there do seem to be related issues and PRs, e.g. https://github.com/microsoft/torchgeo/issues/1407.

### Additional information

I would be interested in contributing to a solution. However, I'm not quite sure how to get started.

hkristen:
This is the PR based on the discussions in https://github.com/torchgeo/torchgeo/discussions/3099.

Closes #30 
Closes #35 
Closes #1407
Closes #1440

**Overview**
Adds a TiledInferenceCallback that enables large-area geospatial inference by splitting a region into overlapping tiles, running predictions on each tile, and merging results into a single Cloud-Optimized GeoTIFF with weighted blending. Without this, users must manually handle patch-to-scene reconstruction, overlap blending, CRS management, and memory-efficient output writing.

**Usage Example**
See https://gist.github.com/hkristen/ee0b3f0e2324c9d606da3a7e537a2582 for a complete CLI script demonstrating tiled inference on a large raster with ROI filtering.

**ToDos for further versions**
- [ ] BUG: When using delta=0, blending is fully disabled
- [ ] Configurable output format/compression options (currently hardcoded COG)
- [ ] Quantized uint8 probabilities instead of one-hot (better blending, same disk cost)
- [ ] Support regression tasks, not just classification
- [ ] Try-except `gdal raster overview add` with fallback to `gdaladdo` in writer
- [ ] Restore `trainer.return_predictions` to original value after predict ends
- [ ] Widen output dtype to `uint16` when `num_classes > 255` to avoid label truncation
- [ ] Investigate normalization mismatch between `GeoDataModule` and plain `DataLoader` predict

**Architecture Overview**
The callback hooks into Lightning's predict lifecycle across three phases:
```
┌─────────────────────────────────────────────────────────────────────────┐
│                         PREDICTION LIFECYCLE                            │
├─────────────────────────────────────────────────────────────────────────┤
│                                                                         │
│  ┌───────────────────────────────────────────────────────────────────┐  │
│  │  Phase 1: on_predict_start()                                      │  │
│  │                                                                   │  │
│  │  - Disables Lightning's prediction storage (memory leak fix)      │  │
│  │  - Introspects datamodule.predict_dataset for CRS, bounds, res    │  │
│  │  - Creates temp directory for patch files                         │  │
│  └───────────────────────────────────────────────────────────────────┘  │
│                              │                                          │
│                              ▼                                          │
│  ┌───────────────────────────────────────────────────────────────────┐  │
│  │  Phase 2: on_predict_batch_end()    (called per batch)            │  │
│  │                                                                   │  │
│  │  For each sample in the batch:                                    │  │
│  │    1. Extract probabilities, bounds, transform from predict_step  │  │
│  │    2. Convert softmax → argmax → one-hot (float32 → uint8)        │  │
│  │    3. Write as georeferenced GeoTIFF to temp dir                  │  │
│  │    4. Collect metadata: patch_id, file path, geo_bbox, transform  │  │
│  │                                                                   │  │
│  │  Temp patch layout:                                               │  │
│  │    .tmp_predictions/                                              │  │
│  │      patch_000000.tif   ← one-hot uint8, LZW compressed           │  │
│  │      patch_000001.tif      with affine transform + CRS            │  │
│  │      ...                                                          │  │
│  └───────────────────────────────────────────────────────────────────┘  │
│                              │                                          │
│                              ▼                                          │
│  ┌───────────────────────────────────────────────────────────────────┐  │
│  │  Phase 3: on_predict_epoch_end()                                  │  │
│  │                                                                   │  │
│  │  weighted_merge():                                                │  │
│  │    1. Reconstruct scene geometry                                  │  │
│  │       - Use dataset_bounds/res if available                       │  │
│  │       - Otherwise infer from patch transforms                     │  │
│  │                                                                   │  │
│  │    2. Build grid spatial index over all patches                   │  │
│  │       ┌─────┬─────┬─────┐                                         │  │
│  │       │ p0  │ p1  │ p2  │  Grid cells map to patch lists          │  │
│  │       ├─────┼─────┼─────┤  for O(1) lookup per chunk              │  │
│  │       │ p3  │ p4  │ p5  │                                         │  │
│  │       └─────┴─────┴─────┘                                         │  │
│  │                                                                   │  │
│  │    3. Process output in chunks (default 4096×4096 px):            │  │
│  │       For each chunk:                                             │  │
│  │         a. Query spatial index → overlapping patches              │  │
│  │         b. Read each patch from disk                              │  │
│  │         c. Apply edge-aware delta cropping:                       │  │
│  │            - Interior edges: crop δ pixels (remove artifacts)     │  │
│  │            - Scene boundary edges: crop 0 (preserve coverage)     │  │
│  │         d. Generate blend mask (cosine or linear ramp):           │  │
│  │                                                                   │  │
│  │            1.0 ┤  ╭────────────────╮                              │  │
│  │                │ ╱                  ╲   ← cosine ramp in overlap  │  │
│  │            0.0 ┤╱                    ╲                            │  │
│  │                └──┬──────────────┬──┘                             │  │
│  │                overlap        overlap                             │  │
│  │                                                                   │  │
│  │         e. Accumulate: output += patch × mask; weights += mask    │  │
│  │         f. Normalize: output /= weights                           │  │
│  │         g. Argmax → uint8 class labels                            │  │
│  │         h. Write chunk to output via GeoTIFFWriter                │  │
│  │                                                                   │  │
│  │    4. GeoTIFFWriter.finalize()                                    │  │
│  │       - Build COG overviews [2, 4, 8, 16, 32, 64]                 │  │
│  │                                                                   │  │
│  │    5. Clean up temp directory                                     │  │
│  └───────────────────────────────────────────────────────────────────┘  │
│                                                                         │
│  Output: Cloud-Optimized GeoTIFF with class labels, CRS, overviews      │
└─────────────────────────────────────────────────────────────────────────┘
```

**Files to Review**
All other changes in the diff are from main or other merged PRs. The files specific to this PR are:

torchgeo/callbacks/__init__.py
What it does: New module, exports TiledInferenceCallback
────────────────────────────────────────
torchgeo/callbacks/tiled_inference.py
What it does: Main callback: lifecycle hooks, temp file management, patch serialization, dataset introspection
────────────────────────────────────────
torchgeo/callbacks/blending.py
What it does: weighted_merge(), get_blend_mask(), scene reconstruction from patch transforms, grid spatial index, edge-aware delta cropping
────────────────────────────────────────
torchgeo/callbacks/writer.py
What it does: GeoTIFFWriter: chunked rasterio writing, COG overview building
────────────────────────────────────────
torchgeo/datamodules/geo.py
What it does: Added predict_roi + predict_stride params to GeoDataModule; predict stage in setup()
────────────────────────────────────────
torchgeo/trainers/segmentation.py
What it does: predict_step() returns dict with {probabilities, bounds, transform} instead of bare Tensor
────────────────────────────────────────
docs/api/callbacks.rst + docs/api/index.rst
What it does: Add callbacks to API docs
────────────────────────────────────────
tests/trainers/test_segmentation.py
What it does: 4 tests for the new predict_step return contract

**Key Design Decisions**
- Patches saved to disk as temp GeoTIFFs rather than held in memory. Allows arbitrarily large scenes without OOM. One-hot encoding (uint8) stored instead of
raw logits (float32) to reduce temp storage by ~4x per class.
- Chunked merging with spatial indexing: The merge processes the output in configurable chunks (default 4096px) and only reads patches that overlap each
chunk. A grid-based spatial index avoids O(n) scans per chunk.
- Edge-aware delta cropping: Interior patch edges are cropped by delta pixels to remove neural network boundary artifacts. Patches touching scene boundaries
keep their edge pixels (delta=0) to avoid black borders.
- Cosine blending (default) in overlap regions for smooth transitions. Linear also available.
- predict_step returns dict (breaking change, versionchanged:: 0.9): Needed to propagate bounds and transform metadata for georeferenced output.

**What to Focus On in a Review**
Critical -- needs careful review:
- blending.py: The core algorithm. Coordinate math (geo to pixel), edge delta logic, chunk/patch overlap intersection. Bugs here produce silent visual
artifacts.
- tiled_inference.py: The lifecycle hooks, especially bounds indexing in on_predict_batch_end and dataset introspection in on_predict_start.
- segmentation.py: Breaking change to predict_step return type.
- geo.py: The predict stage setup, especially the stage string matching ('predict' vs 'predicting' enum).

Mechanical / low-risk:
- writer.py is a straightforward rasterio wrapper
- Doc additions
- Datamodule subclass signature changes (if any remain in the diff)

Wanting feedback on:
- The TODO: Get rid of linear comment -- should linear blending be removed before merge?
- COG config defaults (overview levels [2, 4, 8, 16, 32, 64]) -- reasonable?
- Error messages referencing "PR #3138" / "PR #3140" -- should be user-facing docs references

**Testing**
Tests in tests/trainers/test_segmentation.py cover the predict_step contract (dict return, required keys, shape, metadata passthrough).

Gap: No integration test for the full pipeline (patch saving → merge → GeoTIFF output) and no unit tests for blending.py functions.

## Summary by CodeRabbit

* **New Features**
  * Tiled inference callback for large-scale semantic segmentation
  * Patch blending and weighted merge with cosine/linear methods for seamless outputs
  * Cloud-Optimized GeoTIFF writer with chunked/overview support
  * ROI and stride options for prediction datamodules
  * New public export for tiled inference callback

* **Documentation**
  * Added callbacks API docs and silenced select doc warnings

* **Tests**
  * Extensive tests covering callbacks, blending, tiled inference, writer, and datamodule predict behavior


hkristen:
I tested the tiled inference on a larger scene 46317x40717, discovered a memory overflow and fixed it. Besides that and that mypy is bugging me, stuff is working as it should. 
@adamjstewart feel free to add you thoughts on the PR

calebrob6:
This is awesome!! What can I do to help get this over the finish line?

hkristen:
@calebrob6 I put together a minimal example script for testing the TiledInferenceCallback locally:
https://gist.github.com/hkristen/ee0b3f0e2324c9d606da3a7e537a2582

Try it out on your own data & models of various sizes, try to break it, and tell me how you like the integration/entry point into torchgeo. That would help a lot :) 

Other than that maybe you can share the review load with @adamjstewart as it is again a quite big PR ;)

hkristen:
FYI I was testing this on a larger GeoTiff with 108,262 x 97,256 pixels (1.9 GB, 3 bands RGB) and saw the temp storage explode - I was storing raw float32 logits with torch.save() with about 5 MB per patch which caused my disk to run full very fast. 

So I switched to storing argmax → one-hot as uint8 GeoTIFF with LZW compression (~0.06 MB per patch) which is a bit slower than but let's you ran way larger areas with enough disk space.

tests/callbacks/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright (c) TorchGeo Contributors
+# All rights reserved.
+# Licensed under the MIT License.
```
adamjstewart: This is a bit different from our standard copyright (split onto 3 lines instead of 2)

tests/callbacks/test_writer.py
```diff
@@ -0,0 +1,113 @@
...
+# Licensed under the MIT License.
+
+"""Tests for GeoTIFF writer."""
+
+from pathlib import Path
+
+import numpy as np
+import pytest
+import rasterio
+from affine import Affine
```
adamjstewart: We try to import from `rasterio.transform.Affine` to avoid a separate dep on affine

torchgeo/callbacks/__init__.py
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) TorchGeo Contributors
+# All rights reserved.
+# Licensed under the MIT License.
+
+"""TorchGeo callbacks."""
+
+from torchgeo.callbacks.tiled_inference import TiledInferenceCallback
```
adamjstewart:
```suggestion
from .tiled_inference import TiledInferenceCallback
```
Relative imports for core code, absolute imports for tests

torchgeo/callbacks/blending.py
adamjstewart, line 1: I don't usually bother with `_hidden.py` naming. If it isn't in the documentation, it's private.

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,144 @@
...
+from pathlib import Path
+from typing import Any
+
+import numpy as np
+import rasterio
+from affine import Affine
+from rasterio.windows import Window
+
+
+class GeoTIFFWriter:
```
adamjstewart, line 17: Praying we don't get stuck in bikeshedding over the "best" file format to write to. I'm pro COGs, although if I had to choose it would be the same file format/compression/etc. used by the input data. But I don't know how easy it is to track that metadata since it can't easily be passed through the data loader and isn't exposed in any of our current datasets anyway.

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,144 @@
...
+        if self.cog_config.get('overviews'):
+            import subprocess
+
+            overview_levels = self.cog_config['overviews']
+            resampling = self.cog_config.get('overview_resampling', 'nearest')
+            compress = self.cog_config.get('compress', 'lzw')
+
+            subprocess.run(
+                [
+                    'gdaladdo',
```
adamjstewart: Controversial, but I wonder if we should try-except with the new `gdal raster overview add` interface and fallback on the old `gdaladdo` interface.

torchgeo/callbacks/tiled_inference.py
```diff
@@ -0,0 +1,225 @@
...
+
+        callback = TiledInferenceCallback(
+            output_path='predictions.tif',
+            overlap=32,
+            delta=8,
+        )
+        trainer = Trainer(callbacks=[callback])
+        trainer.predict(task, datamodule)
+
+    .. versionadded:: 0.7
```
adamjstewart:
```suggestion
    .. versionadded:: 0.9
```

adamjstewart (commented):
Tbh, I've never written a callback before so will need to read a few tutorials. Before I review in detail, let's start with the overall design. What's the motivation behind a callback as opposed to always saving predictions in `SemanticSegmentationTask.predict_step` and stitching predictions in `SemanticSegmentationTask.predict_step_end`? Are callbacks just optional additions to LightningModules such that the same code can be optionally added to all tasks?

hkristen:
> Tbh, I've never written a callback before so will need to read a few tutorials. Before I review in detail, let's start with the overall design. 

This was based on your recommendation in #3099 where you said "YES" to "override predict_step, use Lightning callbacks for stitching. Did I misunderstand this? 

> What's the motivation behind a callback as opposed to always saving predictions in `SemanticSegmentationTask.predict_step` and stitching predictions in `SemanticSegmentationTask.predict_step_end`? Are callbacks just optional additions to LightningModules such that the same code can be optionally added to all tasks?

A callback keeps the stitching logic opt-in rather than always-on, and separates I/O concerns from the task's model logic.

Yes, callbacks are optional additions - same code can be attached to any compatible task.

torchgeo/callbacks/writer.py
hkristen: Good idea!

torchgeo/callbacks/blending.py
hkristen: Fine for me :)

torchgeo/callbacks/writer.py
hkristen: I would say we keep COG for the temp files and for the inital PR also as output format. 

Giving the user the ability to specify the output format/compression/etc. is something we ould do in a follow up PR to not make this one even bigger :P

hkristen:
Tested this on an even larger COG with 20cm resolution and 108k × 97k pixel raster, with a UNET - ofc got another memory overflow :P 

The culprit was the tensor accumulation in patch_metadata, switching to Python lists fixed it completely. Now I can run inference on the huge test area, which is equivalent to ~87 Sentinel-2A tiles/granules at 10m resolution, with a stable 9-11GB RAM usage while merging🥳

robmarkcole:
Noting that rslearn is also using a callback approach https://github.com/allenai/rslearn?tab=readme-ov-file#apply-the-model

torchgeo/callbacks/tiled_inference.py
```diff
@@ -0,0 +1,251 @@
...
+from pathlib import Path
+from typing import Any
+
+import rasterio
+import torch
+from lightning.pytorch.callbacks import Callback
+from rasterio.transform import Affine
+
+
+def _gdal_available() -> bool:
```
adamjstewart: We don't currently have a dependency on osgeo, and probably don't want to add one until they start publishing wheels. Is it possible to do all of this with rasterio.merge?

hkristen: Sure, I was not aware that this can be done with rasterio's dst.build_overviews() - much cleaner. The merging is already done with rasterio btw.

hkristen:
A issue for another day mentioned in the [TorchGeo Slack channel:](https://torchgeo.slack.com/archives/C06NQM78TAQ/p1772115486737719)

`GeoDataModule` silently applies `Normalize(mean=0, std=255)` to all inputs during predict. When `predict_aug` is not explicitly set (which is the default), it falls back to dividing every pixel value by 255. @Wyclife Agumba Oluoch’s model was trained on raw Sentinel-2 values (~100–5000), so after this hidden normalization the model sees completely different inputs and produces garbage.
That's why single-patch inference works (plain DataLoader, no normalization) but tiled inference via GeoDataModule doesn't.

I need to think a bit more how to fix this behaviour in an elegant way. Probably tackle this in a follow up PR once this is merged.

isaaccorley:
> A issue for another day mentioned in the [TorchGeo Slack channel:](https://torchgeo.slack.com/archives/C06NQM78TAQ/p1772115486737719)
> 
> 
> 
> `GeoDataModule` silently applies `Normalize(mean=0, std=255)` to all inputs during predict. When `predict_aug` is not explicitly set (which is the default), it falls back to dividing every pixel value by 255. @Wyclife Agumba Oluoch’s model was trained on raw Sentinel-2 values (~100–5000), so after this hidden normalization the model sees completely different inputs and produces garbage.
> 
> That's why single-patch inference works (plain DataLoader, no normalization) but tiled inference via GeoDataModule doesn't.
> 
> 
> 
> I need to think a bit more how to fix this behaviour in an elegant way. Probably tackle this in a follow up PR once this is merged. 

@adamjstewart Might need to revive the discussion on changing this to default dividing by 1 instead

torchgeo/datamodules/sentinel2_south_america_soybean.py
```diff
@@ -77,11 +78,15 @@ def __init__(
...
+    def setup(
+        self, stage: str, roi: Polygon | None = None, stride: int | None = None
+    ) -> None:
         """Set up datasets and samplers.
 
         Args:
             stage: Either 'fit', 'validate', 'test', or 'predict'.
+            roi: Optional region of interest for predict stage.
+            stride: Optional stride for GridGeoSampler (predict stage only).
         """
```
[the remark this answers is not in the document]
hkristen: Outside diff range, won't fix here

torchgeo/datasets/gbif.py
```diff
@@ -61,7 +61,7 @@ def __init__(self, root: Path = 'data') -> None:
         usecols = ['decimalLatitude', 'decimalLongitude', 'day', 'month', 'year']
         dtype = {'day': str, 'month': str, 'year': str}
         # https://github.com/pandas-dev/pandas-stubs/pull/1696
-        df = pd.read_table(files[0], usecols=usecols, dtype=dtype)  # type: ignore[invalid-argument-type]
+        df = pd.read_table(files[0], usecols=usecols, dtype=dtype)
```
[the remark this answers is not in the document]
hkristen: Outside diff range, won't fix here

torchgeo/datasets/splits.py
```diff
@@ -81,7 +83,7 @@ def random_bbox_assignment(
     new_datasets = []
     for offset, length in zip(itertools.accumulate(lengths), lengths):
         ds = deepcopy(dataset)
-        ds.index = dataset.index.iloc[indices[offset - length : offset]]  # type: ignore[invalid-argument-type]
+        ds.index = dataset.index.iloc[indices[offset - length : offset]]
```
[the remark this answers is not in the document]
hkristen: Outside diff range, won't fix here

torchgeo/datasets/splits.py
```diff
@@ -166,7 +168,7 @@ def random_bbox_splitting(
                             new_geometry, geometry_remaining = geom1, geom2
                         else:
                             geometry_remaining, new_geometry = geom1, geom2
-            new_datasets[j].index.iloc[i, i_geom] = new_geometry  # type: ignore[invalid-assignment]
+            new_datasets[j].index.iloc[i, i_geom] = new_geometry
```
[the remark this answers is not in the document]
hkristen: Outside diff range, won't fix here

Wycology:
Hi all,

I am happy to report that I have successfully used the `tiled_inference.py` callback [file](https://gist.github.com/hkristen/ee0b3f0e2324c9d606da3a7e537a2582) to generate tiled-inference on three Sentinel-2 raster files with 12 channels each at 10 m spatial resolution:

- Training (Kericho County, Kenya): 3252 × 2249 ≈ 7.3M pixels → ~73,000 ha (~730 km²)
- Validation (Nandi County, Kenya): 1671 × 2784 ≈ 4.7M pixels → ~47,000 ha (~470 km²)
- Test (W. Rwanda region): 448 × 419 ≈ 188K pixels → ~1,880 ha (~18.8 km²)

**Task:**
I am segmenting tea plantations throughout Africa at 10 m resolution (Tea in Africa is primarily grown in Kenya, Uganda, Rwanda, Burundi, Ethiopia, Malawi, Tanzania, South Africa, Zimbabwe, Mozambique, and Cameroon). Some parts of Madagascar and the Mauritius Islands, too. However, for this task, I rely on training and validation areas in Kenya and a small test site in western Rwanda. I manually digitized mask polygons during 2024-2025 period (took me months :) but worth it) using Google Hybrid basemap in QGIS. Leveraging on the fact that the tea canopy remains generally green throughout the year, and once planted, can stay productive for over 100 years, the annual median cloud-free Sentinel-2 was a great candidate for the segmentation task. The only time when the tea canopy is not green is about 2 months in a year when it is recovering from prunning, which happens every ~4 years, but is smoothed out in the case of the annual median image.

**Notes:**

1. Image preprocessing, like normalization, should be done before model training and inferencing, preferably to have pixel values range between 0 and 1. This is usually achieved in Sentinel-2 images by dividing the digital numbers by 10000 or multiplying by 0.0001, which is what I did via [GEE](https://code.earthengine.google.com/a05e27f567274a751e9e755847ea32f1). The scale varies for other bands.
2. Image transforms like adding indices should be avoided as they increase channels, as opposed to the original image, unless one tweaks the `tiled_inference.py` code locally to append similar indices during inference. In my case, I did not.
3. One can train patches at patch size 256 by 256, for instance, but do the inference at a similar or different patch size, say 64 by 64, without any problem, which is what I did.
4. I noticed the `tiled_inference ` callback ran way faster (not measured here) than manual inferencing and stitching.
5. FYI, tea is normally grown by large-scale tea growers (>50 acres), such as by Unilever, George Williamson, Brown, James Finlay (the obviously visible plantation areas in the images attached), medium-scale tea growers (10 - 50 acres), and small-scale tea growers (< 10 acres). Accordingly, I ensured my masks (training polygons) captured all three scales.

Here is the plot of my original images (rgb) and their corresponding inference outputs for train (Kericho County, Kenya), validation (Nandi County, Kenya), and test (western Rwanda) sites.
comparison_figure_small

Once the `tiled_inference ` callback is merged to the main `TorchGeo` package, I can share a full tutorial on this, if necessary.

These were my inference parameters:

```
!python tiled_inference.py \
        --raster path_to_the_image.tif \
        --checkpoint path_to_the_best_model.ckpt \
        --output tiled_inference.tif \
        --patch-size 64 \
        --delta 1 \
        --overlap 8
```

Happy to hear experience from others or answer any questions in my modeling and inference workflow.

Best regards,
Wyclife

docs/conf.py
```diff
@@ -69,6 +69,10 @@
     ('py:class', 'torchvision.models._api.WeightsEnum'),
     ('py:class', 'torchvision.models.resnet.ResNet'),
     ('py:class', 'torchvision.models.swin_transformer.SwinTransformer'),
+    # Internal pandas types not exposed in public docs
+    ('py:class', 'pandas._libs.interval.Interval'),
+    ('py:class', 'pandas._libs.tslibs.timedeltas.Timedelta'),
+    ('py:class', 'pandas.core.series.Series'),
```
adamjstewart: I believe these _should_ be found at the root `pd.*` location. I've had this problem before and it was due to rate limits or unavailability of the pandas docs during CI. I would try removing this, I can debug if the problem reoccurs.

tests/callbacks/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) TorchGeo Contributors
```
adamjstewart:
```suggestion
# Copyright (c) TorchGeo Contributors. All rights reserved. 
```
Same in all files.

tests/callbacks/test_blending.py
```diff
@@ -0,0 +1,688 @@
...
+    _get_edge_deltas,
+    _query_grid_index,
+    _reconstruct_scene_from_patches,
+    get_blend_mask,
+    weighted_merge,
+)
+
+
+def _save_test_patch(
+    path: Path, logits: torch.Tensor, transform: list[float], crs: Any = 'EPSG:32631'
```
adamjstewart:
```suggestion
    path: Path, logits: torch.Tensor, transform: list[float], crs: str = 'EPSG:32631'
```
My goal is to avoid Any whenever possible, it disables the type checker.

tests/datamodules/test_geo.py
adamjstewart, line 1: Let's remove all datamodule changes for now, the PR is already complex enough to review.

torchgeo/datamodules/agrifieldnet.py
adamjstewart, line 1: Let's remove all datamodule changes for now, the PR is already complex enough to review.

torchgeo/datasets/splits.py
```diff
@@ -3,6 +3,8 @@
 
 """Dataset splitting utilities."""
 
+from __future__ import annotations
```
adamjstewart: Why is this needed?

torchgeo/callbacks/tiled_inference.py
```diff
@@ -0,0 +1,226 @@
+# Copyright (c) TorchGeo Contributors
+# Licensed under the MIT License.
+
+"""Tiled inference callback for semantic segmentation."""
```
adamjstewart, line 4: I actually think the ASCII diagram from the PR description is pretty helpful. Should we consider placing this in this file with comments? Or should we somehow convert it into a tutorial introducing the callback and showing how to use it?

torchgeo/callbacks/blending.py
```diff
@@ -0,0 +1,517 @@
+# Copyright (c) TorchGeo Contributors
+# Licensed under the MIT License.
+
+"""Blending utilities for tiled inference."""
+
+from __future__ import annotations
```
adamjstewart, line 6: This should be avoided unless absolutely necessary

torchgeo/callbacks/blending.py
```diff
@@ -0,0 +1,517 @@
...
+
+from collections import defaultdict
+from pathlib import Path
+from typing import Any
```
adamjstewart: Would like to avoid Any, can we use TypedDict?

torchgeo/callbacks/blending.py
```diff
@@ -0,0 +1,517 @@
...
+) -> tuple[int, int, int, int]:
+    """Compute per-edge crop amounts based on boundary proximity.
+
+    Patches touching scene boundaries preserve their edge pixels (delta=0 on that
+    edge) to avoid black borders. Interior edges are cropped normally to remove
+    neural network edge artifacts.
+
+    A tolerance of 1.5 * pixel_size is used for boundary detection to handle
+    floating-point imprecision. This means patches within 1.5 pixels of a scene
+    boundary are treated as boundary-touching.
```
adamjstewart, line 43: 1.5 pixels seems like a very narrow boundary, I would have expected more like 10 pixels? Is this boundary cropped entirely or does it have blending applied?

torchgeo/callbacks/blending.py
```diff
@@ -0,0 +1,517 @@
...
+            if apply_left_ramp:
+                x[:overlap] = ramp
+            if apply_right_ramp:
+                x[-overlap:] = ramp[::-1]
+
+        mask = y[:, None] * x[None, :]
+    else:
+        raise ValueError(f'Unknown blend method: {method}')
+
+    mask += 1e-6
```
adamjstewart: Why? Avoid divide by zero? Do we ever divide by this?

torchgeo/callbacks/tiled_inference.py
```diff
@@ -0,0 +1,226 @@
...
+        from lightning import Trainer
+        from torchgeo.callbacks import TiledInferenceCallback
+
+        callback = TiledInferenceCallback(
+            output_path='predictions.tif', overlap=32, delta=8
+        )
+        trainer = Trainer(callbacks=[callback])
+        trainer.predict(task, datamodule)
+
+    .. versionadded:: 0.9
```
adamjstewart:
```suggestion
    .. versionadded:: 0.10
```

torchgeo/callbacks/tiled_inference.py
```diff
@@ -0,0 +1,226 @@
...
+    """
+
+    def __init__(
+        self,
+        output_path: str | Path,
+        overlap: int = 32,
+        delta: int = 8,
+        blend_method: str = 'cosine',  # TODO: Get rid of linear
```
adamjstewart:
```suggestion
        blend_method: Literal['cosine', 'linear'] = 'cosine',  # TODO: Get rid of linear
```

torchgeo/callbacks/tiled_inference.py
```diff
@@ -0,0 +1,226 @@
...
+        chunk_size: int = 4096,
+        cog_config: dict[str, Any] | None = None,
```
adamjstewart, line 130: This might be one of the few places I'm okay with Any, it's basically the same as `**kwargs`. We could even consider calling it that so that users can add as many as they want without creating a dict.

torchgeo/callbacks/tiled_inference.py
```diff
@@ -0,0 +1,226 @@
...
+        self.cog_config = {**default_cog_config, **(cog_config or {})}
+
+        self.temp_dir: Path | None = None
+        self.patch_metadata: list[dict[str, Any]] = []
+        self.num_classes: int | None = None
+        self.crs: Any = None
+        self.dataset_bounds: tuple[float, float, float, float] | None = None
+        self.dataset_res: float | None = None
+
+    def on_predict_start(self, trainer: Any, pl_module: Any) -> None:
```
adamjstewart: Remove Any if possible

adamjstewart (commented):
Finally found time to review this!

I would like to cite https://arxiv.org/abs/1805.12219 or any other relevant docs somewhere as a useful references for best practices for stitching. Anything we can do to follow standard naming conventions set by older papers would be nice.

Ignoring tests for now, and will need to review each individual utility function in more detail later. I wonder if we should combine blending.py and writer.py into utils.py to make it clear that these aren't callbacks, but helper functions for other callbacks. I do think there are other things we could add callbacks for in the future.

Ideally this same callback will work for SemanticSegmentationTask and PixelwiseRegressionTask (including the 3D versions we're planning on adding for time series support). I don't expect any other tasks to be compatible with this. Just keep both pixelwise classification and regression tasks in mind. For classification, we don't want a soft blend, we need a hard vote and uint8. For regression, we do want a soft blend and float32. Let's make sure we test both. If you want to focus on classification first and figure out regression later in a separate PR, I could be convinced. I'm not actually sure which is easier.

Reminder for me later: test this using a checkerboard pattern to make sure stitching behaves as expected in all situations.

torchgeo/callbacks/blending.py
hkristen: The 1.5 is a floating-point comparison tolerance for boundary detection, not a blending width. It just ensures patches within 1.5 pixels of a scene edge are treated as boundary-touching (to handle FP imprecision).

torchgeo/callbacks/blending.py
hkristen: You are right, this is unnecessary. Dropped it.

torchgeo/callbacks/blending.py
hkristen: Alrighty. I added PatchMetadata TypedDict. Not very pretty as the test fixtures use minimal dicts so call sites need # type: ignore[arg-type].

torchgeo/datasets/splits.py
hkristen: It is not. Removed

hkristen:
> Finally found time to review this!

Thanks Adam :) 

> I would like to cite https://arxiv.org/abs/1805.12219 or any other relevant docs somewhere as a useful references for best practices for stitching. Anything we can do to follow standard naming conventions set by older papers would be nice.

Good idea, added to my todo list.

> I wonder if we should combine blending.py and writer.py into utils.py to make it clear that these aren't callbacks, but helper functions for other callbacks.

My reasoning behind this is that I don't want a super long utils.py -> but I also don't care too much. Your call. 

> If you want to focus on classification first and figure out regression later in a separate PR, I could be convinced. I'm not actually sure which is easier.

I would prefer to stick to classification for now, as it already works. We can work on regression in a follow up PR

torchgeo/callbacks/blending.py
```diff
@@ -0,0 +1,517 @@
...
+    num_classes: int,
+    overlap: int,
+    delta: int,
+    blend_method: str = 'cosine',
+    crs: Any = None,
+    output_path: str | Path | None = None,
+    chunk_size: int = 4096,
+    cog_config: dict[str, Any] | None = None,
+    dataset_bounds: tuple[float, float, float, float] | None = None,
+    dataset_res: float | None = None,
```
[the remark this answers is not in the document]
hkristen: Wow good catch!

tests/datamodules/test_geo.py
hkristen: Done! I rebased and dropped all datamodule changes.

torchgeo/callbacks/blending.py
adamjstewart: We could use the same TypedDict at all call sites too...

adamjstewart:
> Reminder for me later: test this using a checkerboard pattern to make sure stitching behaves as expected in all situations.

I'm struggling to get even the simplest example working here. Can you figure out what I'm doing wrong?

```python
"""Test weighted_merge via checkerboard pattern."""

import einops
import numpy as np
import rasterio as rio
from rasterio.transform import from_bounds

from torchgeo.callbacks.blending import weighted_merge

PATCH_SIZE = 8
GRID_SIZE = 3
OVERLAP = 0

# For simplicity, use identity transform: origin (0, 0), res (1, -1)

patch_metadata = []
patch_id = 0
for row in range(GRID_SIZE):
    for col in range(GRID_SIZE):
        shape = (PATCH_SIZE, PATCH_SIZE)
        fill_value = (row + col) % 2  # checkerboard pattern
        Z = np.full(shape, fill_value, dtype=np.uint8)

        west = col * (PATCH_SIZE - OVERLAP)
        south = row * (PATCH_SIZE - OVERLAP)
        east = west + PATCH_SIZE
        north = south + PATCH_SIZE
        transform = from_bounds(west, south, east, north, PATCH_SIZE, PATCH_SIZE)

        geo_bbox = (west, south, east, north)

        profile = {
            'transform': transform,
            'width': PATCH_SIZE,
            'height': PATCH_SIZE,
            'count': 1,
            'dtype': np.uint8,
        }

        file = f'patch_{patch_id}.tif'
        with rio.open(file, 'w', **profile) as dst:
            dst.write(Z, 1)

        patch_metadata.append(
            {
                'patch_id': patch_id,
                'file': file,
                'geo_bbox': geo_bbox,
                'transform': transform,
            }
        )

        patch_id += 1

weighted_merge(patch_metadata, 2, OVERLAP, OVERLAP, output_path='out.tif')

with rio.open('out.tif') as src:
    img = src.read() * 255
    img = einops.rearrange(img, 'c h w -> h w c')

plt.imshow(img, cmap='gray')
plt.show()
```
The saved `out.tif` is all zeros, despite the inputs being 0 and 1.

hkristen:
The issue is that `weighted_merge` expects one-hot encoded patch files (one band per class), matching what `TiledInferenceCallback` actually writes. See [Line 248-271](https://github.com/hkristen/torchgeo/blob/tiled-inference-callback/torchgeo/callbacks/tiled_inference.py#L248-L271) where it converts argmax predictions to one-hot before writing.

Your script writes 1-band files with raw labels (0 or 1). When `weighted_merge` reads a 1-band file with `num_classes=2`, numpy broadcasts that single band to both `chunk_output[0]` and `chunk_output[1]`, making them equal — so `argmax` always returns 0.

Fix: write 2-band one-hot tifs:

```python
fill_value = (row + col) % 2
Z = np.zeros((2, PATCH_SIZE, PATCH_SIZE), dtype=np.uint8)
Z[fill_value] = 1  # one-hot: band k=1 where class k is predicted

profile = {
    'transform': transform,
    'width': PATCH_SIZE,
    'height': PATCH_SIZE,
    'count': 2,  # num_classes bands
    'dtype': np.uint8,
}
```

Should produce a checkerboard pattern.

hkristen:
Here is the corrected version:

```python

import tempfile
from pathlib import Path

import numpy as np
import rasterio as rio
from rasterio.transform import from_bounds

from torchgeo.callbacks.blending import weighted_merge

PATCH_SIZE = 8
GRID_SIZE = 3
OVERLAP = 0

patch_metadata = []
patch_id = 0

with tempfile.TemporaryDirectory() as tmpdir:
    tmpdir = Path(tmpdir)

    for row in range(GRID_SIZE):
        for col in range(GRID_SIZE):
            fill_value = (row + col) % 2  # checkerboard: 0 or 1

            # One-hot encode: num_classes bands, band k=1 where class k is predicted.
            Z = np.zeros((2, PATCH_SIZE, PATCH_SIZE), dtype=np.uint8)
            Z[fill_value] = 1

            west = col * PATCH_SIZE
            south = row * PATCH_SIZE
            east = west + PATCH_SIZE
            north = south + PATCH_SIZE
            transform = from_bounds(west, south, east, north, PATCH_SIZE, PATCH_SIZE)

            profile = {
                "driver": "GTiff",
                "transform": transform,
                "width": PATCH_SIZE,
                "height": PATCH_SIZE,
                "count": 2,  # num_classes bands
                "dtype": np.uint8,
            }

            patch_path = tmpdir / f"patch_{patch_id}.tif"
            with rio.open(patch_path, "w", **profile) as dst:
                dst.write(Z)

            patch_metadata.append(
                {
                    "patch_id": patch_id,
                    "file": patch_path,
                    "geo_bbox": (west, south, east, north),
                    "transform": list(transform),
                }
            )
            patch_id += 1

    output_path = tmpdir / "out.tif"
    weighted_merge(patch_metadata, 2, OVERLAP, OVERLAP, output_path=output_path)

    with rio.open(output_path) as src:
        img = src.read(1)

    print("Unique values:", np.unique(img))  # [0 1] ✓
    print(img)
```

**Two fixes as described above:**

1. `count: 2` patch files must have `num_classes` bands, not 1.
2. `Z = np.zeros((2, ...)); Z[fill_value] = 1` one-hot encoding, not raw labels.

---- 

My reasoning for using one-hot and not float32 logits -> one-hot uint8 is **4× smaller** than float32 logits per pixel, and for large scenes that adds up fast.


| Format | Bands | dtype | bytes/pixel (2 classes) |
|--------|-------|-------|------------------------|
| float32 logits | num_classes | float32 | 8 |
| one-hot | num_classes | uint8 | 2 |
| argmax label | 1 | uint8 | 1 |

So the design decision was for one-hot as to me it hits the sweet spot: blends well enough and is cheap enough. The only thing we lose over float32 is soft probabilities, which seems ok for a first iteration -> I am planning to add a option for the user to get the float32 probalities in a follow up PR -> @Wycology already asked for that.

adamjstewart:
Okay, I made a couple changes to my code:
```python
"""Test weighted_merge via checkerboard pattern."""

import einops
import numpy as np
import rasterio as rio
from rasterio.transform import from_bounds

from torchgeo.callbacks.blending import weighted_merge

CLASSES = 2
PATCH_SIZE = 8
GRID_SIZE = 5
OVERLAP = 3

# For simplicity, use identity transform: origin (0, 0), res (1, -1)

patch_metadata = []
patch_id = 0
for row in range(GRID_SIZE):
    for col in range(GRID_SIZE):
        shape = (CLASSES, PATCH_SIZE, PATCH_SIZE)
        fill_value = (row + col) % 2  # checkerboard pattern
        Z = np.zeros(shape, dtype=np.uint8)
        Z[fill_value] = 1

        west = col * (PATCH_SIZE - OVERLAP)
        south = row * (PATCH_SIZE - OVERLAP)
        east = west + PATCH_SIZE
        north = south + PATCH_SIZE
        transform = from_bounds(west, south, east, north, PATCH_SIZE, PATCH_SIZE)

        geo_bbox = (west, south, east, north)

        profile = {
            'transform': transform,
            'width': PATCH_SIZE,
            'height': PATCH_SIZE,
            'count': CLASSES,
            'dtype': np.uint8,
        }

        file = f'patch_{patch_id}.tif'
        with rio.open(file, 'w', **profile) as dst:
            dst.write(Z)

        patch_metadata.append(
            {
                'patch_id': patch_id,
                'file': file,
                'geo_bbox': geo_bbox,
                'transform': transform,
            }
        )

        patch_id += 1

weighted_merge(patch_metadata, 2, OVERLAP, OVERLAP, output_path='out.tif')

with rio.open('out.tif') as src:
    img = src.read() * 255
    img = einops.rearrange(img, 'c h w -> h w c')

plt.imshow(img, cmap='gray')
plt.show()
```
The code now runs, but I'm not convinced the output is correct.

For regression (not yet implemented, future PR), I would expect a linear/cosine ramp of gray values when overlap > 0

For classification (current PR), I would expect stitching to occur such that the boundary between black and white (hard classes) is halfway between overlapping tiles. But the actual result is that black slowly replaces white entirely:
checkerboard

Am I missing something or is this a bug?

hkristen:
My bad! The example I showed earlier passed `OVERLAP` for both `overlap` and `delta` when calling `weighted_merge`. Like this you crop away the `overlap=3` with the `delta=3` resulting in the odd behaviour.

For your test use `delta=1` that should produce the result you are looking for

```python
weighted_merge(patch_metadata, 2, OVERLAP, delta=1, output_path='out.tif')
```
checkerboard_delta1

Btw. while experimenting with this I also found a bug where `delta=0` disables blending entirely, will need to fix that!

adamjstewart:
Why are the inner black tiles still larger than the inner white tiles? Shouldn't they be the same size if stitching is symmetric?

hkristen:
As the test case has perfect one-hot inputs the overlap pixel is always a tie and `np.argmax` consequently picks class = 0 (black).

adriantre:
This is great! At some point I will make something similar for detection output "global" NMS across patches and rasters.

torchgeo/callbacks/tiled_inference.py
```diff
@@ -0,0 +1,318 @@
...
+        self, trainer: Trainer, pl_module: LightningModule
+    ) -> None:
+        """Merge patches and write GeoTIFF.
+
+        Args:
+            trainer: PyTorch Lightning trainer.
+            pl_module: PyTorch Lightning module.
+        """
+        from torchgeo.callbacks.blending import weighted_merge
+
```
adriantre, line 297: I believe you want to return if not on the main process (ddp)

```python
if trainer.is_global_zero:
    Return
```

https://lightning.ai/docs/pytorch/stable/common/trainer.html#is-global-zero

torchgeo/callbacks/tiled_inference.py
```diff
@@ -0,0 +1,318 @@
...
+        if hasattr(trainer, 'predict_loop'):
+            trainer.predict_loop.return_predictions = False
+
+        datamodule = trainer.datamodule
+        if hasattr(datamodule, 'predict_dataset'):
+            dataset = datamodule.predict_dataset
+            if hasattr(dataset, 'dataset'):
+                dataset = dataset.dataset
+            self.crs = getattr(dataset, 'crs', None)
```
adriantre, line 187: This is the choice of the final merged file's crs. I suggest naming the variable `self.output_crs`.
I also think it would be very clean to pull it into its own method. Foreshadowing changes related to the coming "per-sample crs" 😎

```python
def _resolve_output_crs(self, dataset: GeoDataset) -> CRS:
    """Resolve the CRS the mosaic is written in."""
    return getattr(dataset, 'crs', None)
```

torchgeo/callbacks/blending.py
```diff
@@ -0,0 +1,531 @@
...
+
+    for meta in patch_metadata[1:]:
+        t = meta['transform']
+        if abs(t[0] - x_res) > 1e-6 or abs(t[4] - y_res) > 1e-6:
+            raise ValueError(
+                f'Inconsistent resolutions: first patch has ({x_res}, {y_res}), '
+                f'but patch {meta["patch_id"]} has ({t[0]}, {t[4]})'
+            )
+
+    all_geo_xmin = [meta['geo_bbox'][0] for meta in patch_metadata]
```
adriantre, line 108: Here I suggest extracting this to a method that clarify the choice of output crs. I kept the int of what I forsee in the docstring.

```python
def _patch_bounds_in_output_crs(
    meta: PatchMetadata, output_crs: Any
) -> tuple[float, float, float, float]:
    """Patch bounds in the output CRS.

    Today every patch already shares the output CRS. When patches carry
    per-sample CRSs this is the single place that reprojects:
    ``transform_bounds(meta['crs'], output_crs, *meta['geo_bbox'])``.
    """
    return meta['geo_bbox']

def _reconstruct_scene_from_patches(
    patch_metadata: list[PatchMetadata],
    patch_size: tuple[int, int],
    delta: int = 0,
    output_crs: CRS | string | None = None,   # NEW
) -> tuple[tuple[int, int], Affine]:
    ...

    boxes = [_patch_bounds_in_output_crs(m, output_crs) for m in patch_metadata]
    all_geo_xmin = [b[0] for b in boxes]
    all_geo_xmax = [b[2] for b in boxes]
    all_geo_ymin = [b[1] for b in boxes]
    all_geo_ymax = [b[3] for b in boxes]
    ... 
    for meta in patch_metadata:
    geo_bbox = _patch_bounds_in_output_crs(meta, output_crs)
```

torchgeo/callbacks/blending.py
```diff
@@ -0,0 +1,531 @@
...
+            meta['bbox'] = (
+                patch_col_start,
+                patch_row_start,
+                patch_col_start + effective_patch_w,
+                patch_row_start + effective_patch_h,
+            )
+    else:
+        output_shape, scene_transform = _reconstruct_scene_from_patches(
+            patch_metadata, (patch_h, patch_w), delta
+        )
```
adriantre, line 411: Third and last method explicitly stating the choice of the output. 

Would be called right here like
```python
output_shape, scene_transform = _resolve_output_grid(
    patch_metadata, crs, (patch_h, patch_w), delta, dataset_bounds, dataset_res
)
```

```python
def _resolve_output_grid(
    patch_metadata: list[PatchMetadata],
    output_crs: Any,
    patch_size: tuple[int, int],
    delta: int,
    dataset_bounds: tuple[float, float, float, float] | None,
    dataset_res: tuple[float, float] | None,
) -> tuple[tuple[int, int], Affine]:
    """Resolve the output mosaic grid (shape + transform) and per-patch placement.

    Single decision point for the output grid: extent + resolution → transform, plus
    each patch's pixel window. Two things change when datasets return per-sample CRSs,
    both localized here:
    * each patch's bounds go through ``_patch_bounds_in_output_crs`` (reproject there);
    * ``dataset_bounds``/``dataset_res`` are then interpreted in ``output_crs``.
    Assigns ``meta['bbox']`` and ``meta['edge_deltas']`` per patch.
    """
    patch_h, patch_w = patch_size

    if dataset_bounds is not None and dataset_res is not None:
        minx, miny, maxx, maxy = dataset_bounds
        x_res_ds, y_res_ds = dataset_res
        output_width = round((maxx - minx) / x_res_ds)
        output_height = round((maxy - miny) / y_res_ds)
        output_shape = (output_height, output_width)
        scene_transform = Affine(x_res_ds, 0, minx, 0, -y_res_ds, maxy)
        scene_bounds = dataset_bounds

        first_transform = patch_metadata[0]['transform']
        x_res = first_transform[0]
        y_res = abs(first_transform[4])

        for meta in patch_metadata:
            geo_bbox = _patch_bounds_in_output_crs(meta, output_crs)   # ← only change
            top, bottom, left, right = _get_edge_deltas(
                geo_bbox, scene_bounds, x_res, delta
            )
            meta['edge_deltas'] = (top, bottom, left, right)

            patch_geo_xmin = geo_bbox[0] + left * x_res
            patch_geo_ymax = geo_bbox[3] - top * y_res
            effective_patch_w = patch_w - left - right
            effective_patch_h = patch_h - top - bottom

            patch_col_start = round((patch_geo_xmin - minx) / x_res)
            patch_row_start = round((maxy - patch_geo_ymax) / y_res)

            meta['bbox'] = (
                patch_col_start,
                patch_row_start,
                patch_col_start + effective_patch_w,
                patch_row_start + effective_patch_h,
            )
    else:
        output_shape, scene_transform = _reconstruct_scene_from_patches(
            patch_metadata, (patch_h, patch_w), delta, output_crs
        )

    return output_shape, scene_transform
```

adriantre (commented):
Please consider extracting the three decision points I have commented on into named methods.

In my per-patch-crs work, I'm adding per-sample native-CRS reads, where a RasterDataset can span multiple UTM zones. There, dataset.crs becomes a global index CRS (possibly EPSG:6933) used only for bookkeeping. It's no longer the CRS the pixels are actually in (necessarily). So this would make it super easy to adapt in the future.

Regardless of my feature, it would make for a clean and explicit structure.

Happy to open a small PR against your branch with these extractions if you want. Really nice work, hope it lands soon!

hkristen:
> Please consider extracting the three decision points I have commented on into named methods.
> 
> In my per-patch-crs work #3804, I'm adding per-sample native-CRS reads, where a RasterDataset can span multiple UTM zones. There, dataset.crs becomes a global index CRS (possibly EPSG:6933) used only for bookkeeping. It's no longer the CRS the pixels are actually in (necessarily). So this would make it super easy to adapt in the future.
> 
> Regardless of my feature, it would make for a clean and explicit structure.
> 
> Happy to open a small PR against your branch with these extractions if you want. Really nice work, hope it lands soon!

Thanks for the detailed suggestions!!

One heads up though -> this PR got split up, the blending code now is in #3667, to make it easier to review (same as with #3666). The callback itself will follow as a third PR once that merges.

Your two blending extractions (_patch_bounds_in_output_crs and _resolve_output_grid) would apply 100% to #3667 , so if the offer still stands, a PR against hkristen:feat/callbacks-blending would be very welcome! 
Only timing note: once #3666 merges I'll rebase that branch, so I'd merge your PR before that. 

When #3666 and #3667 are merged I will open the final callback PR #3??? XD and include the output_crs rename and the is_global_zero guard, ok?

hkristen:
Closing this PR, as it is superseeded by #3666 & #3667 and a future callback PR once those two are merged. 

Just to avoid anymore confusion ;)

PR #3221, opened 2025-12-23
Add OSCD100 dataset and datamodule for tutorials

### LINKED ISSUE #3239: OSCD download not working
### Description

I'm getting SSL cert errors trying to download OSCD. Can we rehost on HF?

This works: https://partage.imt.fr/index.php/s/gKRaWgRnLMfwMGo/download
This doesn't: https://partage.mines-telecom.fr/index.php/s/2D6n03k58ygBSpu/download
This works: https://partage.imt.fr/index.php/s/gpStKn4Mpgfnr63/download

### Steps to reproduce

Try to download.

### Version

main

hkristen:
Selected 100 image pairs from the OSCD dataset and sampled 256x256 from them, to achieve a small enough dataset for the Change Detection tutorial. Will be used instead of #3142. 

License is ~CC-BY-4.0~ CC-BY-NC-SA-4.0

Closes #3239

### DIFF
--- a/torchgeo/datamodules/oscd.py
+++ b/torchgeo/datamodules/oscd.py
@@ -95,23 +95,18 @@ class OSCDDataModule(NonGeoDataModule):
         Args:
             stage: Either 'fit', 'validate', 'test', or 'predict'.
         """
+        transforms = K.AugmentationSequential(
+            K.VideoSequential(K.RandomCrop(self.patch_size)),
+            data_keys=None,
+            keepdim=True,
+        )
         if stage in ['fit', 'validate']:
-            transforms = K.AugmentationSequential(
-                K.VideoSequential(K.RandomCrop(self.patch_size)),
-                data_keys=None,
-                keepdim=True,
-            )
             self.dataset = OSCD(split='train', transforms=transforms, **self.kwargs)
             generator = torch.Generator().manual_seed(0)
             self.train_dataset, self.val_dataset = random_split(
                 self.dataset, [1 - self.val_split_pct, self.val_split_pct], generator
             )
         if stage in ['test']:
-            transforms = K.AugmentationSequential(
-                K.VideoSequential(K.CenterCrop(self.patch_size)),
-                data_keys=None,
-                keepdim=True,
-            )
             self.test_dataset = OSCD(split='test', transforms=transforms, **self.kwargs)
 
 
@@ -124,12 +119,18 @@ class OSCD100DataModule(NonGeoDataModule):
     """
 
     def __init__(
-        self, batch_size: int = 8, num_workers: int = 0, **kwargs: Any
+        self,
+        batch_size: int = 8,
+        patch_size: tuple[int, int] | int = 64,
+        num_workers: int = 0,
+        **kwargs: Any,
     ) -> None:
         """Initialize a new OSCD100DataModule instance.
 
         Args:
             batch_size: Size of each mini-batch.
+            patch_size: Size of each patch, either ``size`` or ``(height, width)``.
+                Should be a multiple of 32 for most segmentation architectures.
             num_workers: Number of workers for parallel data loading.
             **kwargs: Additional keyword arguments passed to
                 :class:`~torchgeo.datasets.OSCD100`.
@@ -138,8 +139,38 @@ class OSCD100DataModule(NonGeoDataModule):
             OSCD100, batch_size=batch_size, num_workers=num_workers, **kwargs
         )
 
+        self.patch_size = _to_tuple(patch_size)
+
+        self.bands = kwargs.get('bands', OSCD.all_bands)
+        self.mean = torch.tensor([MEAN[b] for b in self.bands])
+        self.std = torch.tensor([STD[b] for b in self.bands])
+
         self.aug = K.AugmentationSequential(
             K.VideoSequential(K.Normalize(mean=self.mean, std=self.std)),
             data_keys=None,
             keepdim=True,
         )
+
+    def setup(self, stage: str) -> None:
+        """Set up datasets.
+
+        Args:
+            stage: Either 'fit', 'validate', 'test', or 'predict'.
+        """
+        transforms = K.AugmentationSequential(
+            K.VideoSequential(K.RandomCrop(self.patch_size)),
+            data_keys=None,
+            keepdim=True,
+        )
+        if stage in ['fit']:
+            self.train_dataset = OSCD100(
+                split='train', transforms=transforms, **self.kwargs
+            )
+        if stage in ['fit', 'validate']:
+            self.val_dataset = OSCD100(
+                split='val', transforms=transforms, **self.kwargs
+            )
+        if stage in ['test']:
+            self.test_dataset = OSCD100(
+                split='test', transforms=transforms, **self.kwargs
+            )


calebrob6:
Seems good to me!

docs/api/datamodules.rst
```diff
@@ -180,6 +180,11 @@ OSCD
 
 .. autoclass:: OSCDDataModule
 
+OSCD100
+^^^^^^^
```
adamjstewart: Let's throw this under the OSCD header, doesn't need a separate section.

docs/api/datasets.rst
```diff
@@ -427,6 +427,11 @@ OSCD
 
 .. autoclass:: OSCD
 
+OSCD100
+^^^^^^^
+
+.. autoclass:: OSCD100
```
adamjstewart, line 464: Same here

torchgeo/datasets/oscd.py
```diff
@@ -351,3 +353,236 @@ def get_masked(img: Tensor) -> 'np.typing.NDArray[np.uint8]':
...
+
+class OSCD100(NonGeoDataset):
+    """Subset of OSCD with 100 RGB image pairs at 256x256 resolution.
+
+    Intended for tutorials and demonstrations, not benchmarking.
+
+    Dataset features:
+
+    * 100 image pairs cropped from OSCD Sentinel-2 scenes
+    * RGB bands only (B04, B03, B02) at 10m resolution
```
adamjstewart: This version of the dataset is quite different from the original. I'm fine with fewer images and smaller images, but wonder if fewer bands and completely different file formats are actually a good idea. If we keep the datasets more similar, OSCD100 could become a subclass of OSCD with only a few lines of code for different download URLs. See EuroSAT100 for an example of this. I want to take advantage of MS imagery as often as possible, since that's what distinguishes TorchGeo from other libraries like torchvision.

torchgeo/datasets/oscd.py
```diff
@@ -351,3 +353,236 @@ def get_masked(img: Tensor) -> 'np.typing.NDArray[np.uint8]':
...
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://doi.org/10.1109/IGARSS.2018.8518015
+
+    .. versionadded:: 0.8
+    """
+
+    splits: ClassVar[dict[str, dict[str, str]]] = {
+        'train': {
+            'url': 'https://drive.google.com/file/d/1IbpGIBwYaw9dYvbhVXewA6A9cGBxi5Be',
```
adamjstewart: Can convert this to HF now, same for OSCD

adamjstewart (changes requested):
I'm worried that OSCD100 is too different from OSCD. If we make them more similar, this entire PR would only be a few lines per file. We wouldn't need any extra test data, and all other classes would simply become subclasses with a slightly different URL.

torchgeo/datamodules/oscd.py
```diff
@@ -113,3 +113,45 @@ def setup(self, stage: str) -> None:
...
             )
             self.test_dataset = OSCD(split='test', transforms=transforms, **self.kwargs)
+
+
+class OSCD100DataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the OSCD100 dataset.
+
+    Intended for tutorials and demonstrations, not benchmarking.
+
+    .. versionadded:: 0.8
```
adamjstewart:
```suggestion
    .. versionadded:: 0.9
```

torchgeo/datasets/oscd.py
```diff
@@ -351,3 +353,236 @@ def get_masked(img: Tensor) -> 'np.typing.NDArray[np.uint8]':
...
+    Dataset classes:
+
+    0. no change
+    1. change
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://doi.org/10.1109/IGARSS.2018.8518015
+
+    .. versionadded:: 0.8
```
adamjstewart:
```suggestion
    .. versionadded:: 0.9
```

torchgeo/datasets/oscd.py
hkristen: Fair point, but the goal was ro tiny dataset for tutorials - RGB keeps it small and simple for quick demos. Anyone doing real work would use full OSCD anyway. I could add a note in the docstring if that helps clarify?

calebrob6: I agree with @hkristen -- this dataset is for the purposes of having something to make the tutorial tests run fast, lets not overcomplicate it.

torchgeo/datamodules/oscd.py
```diff
@@ -113,3 +113,45 @@ def setup(self, stage: str) -> None:
...
+            data_keys=None,
+            keepdim=True,
+            same_on_batch=True,
+        )
+        self.test_aug = K.AugmentationSequential(
+            K.VideoSequential(K.Normalize(mean=self.mean, std=self.std)),
+            data_keys=None,
+            keepdim=True,
+            same_on_batch=True,
+        )
```
adamjstewart:
```suggestion
        self.aug = K.AugmentationSequential(
            K.VideoSequential(K.Normalize(mean=self.mean, std=self.std)),
            data_keys=None,
            keepdim=True,
        )
```
These are equivalent

torchgeo/datasets/oscd.py
```diff
@@ -54,13 +56,16 @@ class OSCD(NonGeoDataset):
 
     urls: ClassVar[dict[str, str]] = {
         'Onera Satellite Change Detection dataset - Images.zip': (
-            'https://partage.imt.fr/index.php/s/gKRaWgRnLMfwMGo/download'
+            'https://huggingface.co/datasets/hkristen/oscd/resolve/main/'
+            'Onera%20Satellite%20Change%20Detection%20dataset%20-%20Images.zip'
```
adamjstewart: 1. Can be a single line, ruff ignore long URLs
2. Need to use a git commit hash for stable downloads

To avoid duplication and save space, I would recommend the same format you used for OSCD100, with separate `url` and `splits` attributes. This is optional though since it works as is.

torchgeo/datasets/oscd.py
```diff
@@ -351,3 +356,235 @@ def get_masked(img: Tensor) -> 'np.typing.NDArray[np.uint8]':
...
+    1. change
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://doi.org/10.1109/IGARSS.2018.8518015
+
+    .. versionadded:: 0.9
+    """
+
+    url = 'https://huggingface.co/datasets/hkristen/oscd100/resolve/main'
```
adamjstewart:
```suggestion
    url = 'https://hf.co/datasets/hkristen/oscd100/resolve/318475c71c42da38ade70dade29035117be178eb'
```
Git commit hash needed for stable downloads

torchgeo/datasets/oscd.py
adamjstewart: I'm okay with RGB-only if that's what you want, but this would be a lot less code that someone needs to maintain if we simply change PNG to GeoTIFF. I don't see any other major differences between the datasets that requires us to reimplement an entire dataset from scratch.

torchgeo/datasets/oscd.py
```diff
@@ -351,3 +356,235 @@ def get_masked(img: Tensor) -> 'np.typing.NDArray[np.uint8]':
...
+
+        if download:
+            self._download()
+
+        if not self._check_integrity():
+            raise DatasetNotFoundError(self)
+
+        self.files = self._load_files()
+
+    def __getitem__(self, index: int) -> dict[str, Tensor]:
```
adamjstewart:
```suggestion
    def __getitem__(self, index: int) -> Sample:
```
After rebase, same in a few other places

hkristen:
> Entire PR could be reduced to 10 lines of code if we make OSCD100 a subclass of OSCD with the only difference being the url/md5s/all_bands. But I don't want to complicate things 🙂

The [oscd100 dataset](https://huggingface.co/datasets/hkristen/oscd100) is now ~213 MB large which contradicts the orignal requirement of very tiny datasets for tutorials, but we have less code as upon your request @adamjstewart. 

Please review and let's get this merged soon, the whole tutorial dataset situation is keeping me busy for a little too long already :P

tests/datasets/test_oscd.py
```diff
@@ -111,3 +111,67 @@ def test_failed_plot(self, dataset: OSCD) -> None:
         ):
             x = single_band_dataset[0].copy()
             single_band_dataset.plot(x, suptitle='Test')
+
+
+class TestOSCD100:
```
adamjstewart: See `tests/datasets/test_eurosat.py` where we parametrize our fixture over `[EuroSAT, EuroSAT100]` so we don't need any extra testing code.

torchgeo/datasets/oscd.py
```diff
@@ -78,7 +72,7 @@ class OSCD(NonGeoDataset):
 
     zipfile_glob = '*Onera*.zip'
     filename_glob = '*Onera*'
-    splits = ('train', 'test')
+    splits: ClassVar[tuple[str, ...]] = ('train', 'test')
```
adamjstewart: Shouldn't need ClassVar here, tuples are immutable

torchgeo/datasets/oscd.py
```diff
@@ -54,15 +54,9 @@ class OSCD(NonGeoDataset):
     """
 
     urls: ClassVar[dict[str, str]] = {
```
adamjstewart, line 56: Completely optional: _could_ separate the base `url` from each filename to make these a little simpler to override.

adamjstewart (changes requested):
Dataset/datamodule looks much cleaner now! We could still reduce to RGB-only without any extra code (just 1 line to change the default `all_bands`) if we're worried about data volume. Also not sure if we want to add a new validation split (diverging from the original dataset), or maybe we want to add that to the original dataset too. But I don't want to hold this up any longer.

Only remaining requests are to simplify the tests in the same way. Can we delete `tests/data/oscd/oscd100` and instead test OSCD100 on `tests/data/oscd`? That will remove ~175 new files.

P.S. Let me know if you get busy with your new job and want me to get this across the finish line. I think it's really close now.

hkristen:
> P.S. Let me know if you get busy with your new job and want me to get this across the finish line. I think it's really close now.

@adamjstewart if you would be so kind to push this over the finish line? 

I just tried to vibe this, with a mediocore result haha (feel free to get rid of my last commit)

adamjstewart:
@hkristen will give you time to review to make sure I didn't break anything during my refactor.

hkristen:
lgtm! 

The datamodule setup with patch_size and normalization is a nice addition...test simplification makes sense too. thanks for pushing this over the finish line :)

PR #3222, opened 2025-12-23
Tutorials/change detection

hkristen:
Replaces #3176 to use the OSCD100 dataset from #3221 instead of the LEVIRCD100 dataset (which we couldnt use due to licensing issues).


adamjstewart:
Will review after #3221 is merged.

robmarkcole:
Could make this `quiet` to be a bit cleaner 

image

adamjstewart:
@robmarkcole I'm on the fence. `pip install` is a long-running process, and some feedback (even just a progress bar) is kinda nice. We don't save this in the docs, it only appears at run-time. If you open a PR to convert our existing notebooks to just a simple progress bar instead of hundreds of lines of output, I would probably merge.

adamjstewart (changes requested):
Ton of unrelated changes for OSCD100 duplicated from prior PRs. Can you remove all of this so it's only the notebook change?

torchgeo/datasets/oscd.py
```diff
@@ -364,15 +364,15 @@ class OSCD100(OSCD):
     """
 
     urls: ClassVar[dict[str, str]] = {
-        'Onera Satellite Change Detection dataset - Images.zip': 'https://hf.co/datasets/hkristen/oscd100/resolve/cac594329ee7f1a7f7781b1fe9c3d57befa50044/Onera%20Satellite%20Change%20Detection%20dataset%20-%20Images.zip',
```
adamjstewart: Just noticed that the HF repo is missing a license, might want to add that

docs/tutorials/change_detection.ipynb
adamjstewart, line 1: Any way to make this not 20k lines? I think you need to strip metadata and cell outputs to save space, otherwise this isn't easy to store in git or see diffs between versions.

adamjstewart (commented):
Can't view the file on GitHub and ReviewNB isn't working, but I think we discussed switching from U-Net to a more CD-specific architecture.

docs/tutorials/change_detection.ipynb
```diff
@@ -0,0 +1,462 @@
...
+    "## Change Detection with TorchGeo Tutorial\n",
+    "_Written by: Harald Kristen_\n",
+    "\n",
+    "In this tutorial, we demonstrate how to train a change detection model using TorchGeo. \n",
+    "\n",
+    "**What is Change Detection**: \n",
+    "Change detection is the process of identifying differences between images of the same location captured at different times. It's a fundamental task in remote sensing with numerous real-world applications:\n",
+    "\n",
+    "- **Urban Monitoring**: Track city growth, new construction, and infrastructure development\n",
+    "- **Disaster Response**: Assess damage from floods, earthquakes, wildfires, and hurricanes\n",
```
adamjstewart:
```suggestion
    "- **Disaster Response**: Assess damage from floods, earthquakes, wildfires, and tropical cyclones\n",
```
"hurricane" is what we call them in the US, but I think "[tropical cyclone](https://en.wikipedia.org/wiki/Tropical_cyclone)" is the more general term used everywhere. Completely optional, I think people will know what you mean regardless.

docs/tutorials/change_detection.ipynb
```diff
@@ -0,0 +1,462 @@
...
+    "- **Deforestation Tracking**: Monitor illegal logging and forest loss\n",
+    "- **Agriculture**: Detect crop changes, irrigation patterns, and land use shifts\n",
+    "- **Environmental Monitoring**: Track glacial retreat, coastal erosion, and wetland changes\n",
+    "\n",
+    "In this tutorial, we'll train a binary change detection model on the [OSCD100](https://torchgeo.readthedocs.io/en/stable/api/datasets.html#oscd100) dataset to detect urban changes in bi-temporal Sentinel-2 imagery. Our model will learn to identify where new buildings have been constructed or existing ones removed.\n",
```
adamjstewart:
```suggestion
    "In this tutorial, we will train a binary change detection model on the [OSCD100](https://torchgeo.readthedocs.io/en/stable/api/datasets.html#oscd100) dataset to detect urban changes in bi-temporal Sentinel-2 imagery. Our model will learn to identify where new buildings have been constructed or existing ones removed.\n",
```
In the actual paper, we probably ~won't~ _will not_ want to use contractions. They may also ask us to avoid first-person, but we can wait and see. My goal is that the paper text matches the doc text as closely as makes sense (we wouldn't include things like Google Colab mentions in the paper).

docs/tutorials/change_detection.ipynb
```diff
@@ -0,0 +1,462 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%pip install torchgeo tensorboard gdown"
```
adamjstewart: Why is gdown needed? The dataset is on HF, not GDrive.

docs/tutorials/change_detection.ipynb
```diff
@@ -0,0 +1,462 @@
...
+    "# Copyright (c) TorchGeo Contributors. All rights reserved.\n",
+    "# Licensed under the MIT License."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "1",
+   "metadata": {},
+   "source": [
+    "## Change Detection with TorchGeo Tutorial\n",
```
adamjstewart:
```suggestion
    "# Change Detection with TorchGeo Tutorial\n",
```
Top-level title should be a higher heading level than other subsections

docs/tutorials/change_detection.ipynb
```diff
@@ -0,0 +1,462 @@
...
+   "id": "6",
+   "metadata": {},
+   "source": [
+    "## Visualize the Dataset\n",
+    "\n",
+    "Let's download the dataset and look at some examples from OSCD100.\n",
+    "\n",
+    "**About OSCD100**: This is a smaller, downsampled version of the full [OSCD](https://torchgeo.readthedocs.io/en/stable/api/datasets.html#oscd) (Onera Satellite Change Detection) dataset. It is designed for tutorials and quick experimentation. It contains 100 image pairs at 256\u00d7256 resolution with 60 training, 20 validation, and 20 test samples.\n",
+    "\n",
+    "Each sample contains all 13 Sentinel-2 spectral bands, though we only use the RGB bands for simplicity in this notebook. For research and benchmarking, use the full OSCD dataset."
```
adamjstewart, line 67: I still think it would be more interesting and realistic to use all 13 bands and show off our Sentinel-2 pre-trained model weights, but it's up to you.

docs/tutorials/change_detection.ipynb
```diff
@@ -0,0 +1,462 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "8",
+   "metadata": {},
+   "source": [
+    "## Lightning Modules\n",
+    "\n",
+    "TorchGeo uses [Lightning](https://lightning.ai/docs/pytorch/stable/) to organize the training code and setup the dataloader. The `OSCD100DataModule`:\n",
```
adamjstewart:
```suggestion
    "TorchGeo uses [PyTorch Lightning](https://lightning.ai/docs/pytorch/stable/) to organize the training code and set up the data loader. The `OSCD100DataModule`:\n",
```
They keep going back-and-forth with how they name the repo. I think this will be most familiar to others.

docs/tutorials/change_detection.ipynb
```diff
@@ -0,0 +1,462 @@
...
+    "- Training with binary cross-entropy loss\n",
+    "- Computing metrics like accuracy, F1-score, and Jaccard Index\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "14",
+   "metadata": {},
+   "source": [
+    "We use [**BTC** (Bitemporal Change Transformer)](https://torchgeo.readthedocs.io/en/stable/api/models/btc.html), a change-detection-specific architecture that uses a Swin Transformer backbone pretrained on Cityscapes followed by a UPerNet decoder. Unlike general segmentation models that simply concatenate both images into a single tensor, BTC processes them through separate encoder paths and fuses temporal features in a change-aware way.\n",
```
adamjstewart: Probably worth citing Swin Transformer, UPerNet, and Cityscapes here (via URL) and in the paper (via citation).

docs/tutorials/change_detection.ipynb
```diff
@@ -0,0 +1,462 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "26",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Use full 256x256 images \u2014 test_dataset applies RandomCrop so predictions\n",
```
adamjstewart:
```suggestion
    "# Use full 256x256 images - test_dataset applies RandomCrop so predictions\n",
```
Unicode not really necessary here

docs/tutorials/change_detection.ipynb
```diff
@@ -0,0 +1,462 @@
...
+    "# from the dataloader wouldn't align with ground truth on direct access.\n",
```
adamjstewart: Can't we use the tensorboard log for visualization? Then we can remove all of this code and still show multiple examples.

docs/tutorials/change_detection.ipynb
```diff
@@ -0,0 +1,462 @@
...
+   "source": [
+    "### Interpreting the Results\n",
+    "\n",
+    "With BTC (swin_tiny) you should expect:\n",
+    "\n",
+    "- **Accuracy**: ~82\u201395%\n",
+    "- **F1 Score**: ~0.35\u20130.55\n",
+    "- **IoU (Jaccard Index)**: ~0.24\u20130.40\n",
+    "\n",
+    "Change detection is hard due to class imbalance as most pixels show no change, especially on such a small toy dataset. We set a random seed to improve reproducibility, but results may vary slightly between runs -> Swin Transformer's attention uses CUDA kernels that have no deterministic implementation.\n",
```
adamjstewart:
```suggestion
    "Change detection is hard due to class imbalance as most pixels show no change, especially on such a small toy dataset. We set a random seed to improve reproducibility, but results may vary slightly between runs, as Swin Transformer's attention uses CUDA kernels that have no deterministic implementation.\n",
```
Is this true? Even with `deterministic=True`? These ranges are quite large.

docs/tutorials/change_detection.ipynb
```diff
@@ -0,0 +1,462 @@
...
+    "for idx in [1, 7, 13]:\n",
+    "    show_sample(idx)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "27",
+   "metadata": {},
+   "source": [
+    "### Interpreting the Results\n",
```
adamjstewart: You switched from ## to ### here, not sure if that was on purpose. Let's just make sure we're consistent with sections and subsections.

docs/tutorials/change_detection.ipynb
```diff
@@ -0,0 +1,462 @@
...
+    "- **Accuracy**: ~82\u201395%\n",
+    "- **F1 Score**: ~0.35\u20130.55\n",
+    "- **IoU (Jaccard Index)**: ~0.24\u20130.40\n",
+    "\n",
+    "Change detection is hard due to class imbalance as most pixels show no change, especially on such a small toy dataset. We set a random seed to improve reproducibility, but results may vary slightly between runs -> Swin Transformer's attention uses CUDA kernels that have no deterministic implementation.\n",
+    "\n",
+    "### To Go Further\n",
+    "\n",
+    "- Train on the full [OSCD](https://torchgeo.readthedocs.io/en/stable/api/datasets.html#oscd) dataset (24 city pairs) with full resolution\n",
+    "- Add all 13 Sentinel-2 bands (`in_channels=3` \u2192 `in_channels=13`, set `weights=False`)\n",
```
adamjstewart: We should figure out how to use our pre-trained Sentinel-2 13 band weights instead of recommending `weights=False`.

adamjstewart (commented):
Mostly looks good now, just a few minor grammar comments. I would prefer to merge this soon and keep iterating on it in future PRs or in the actual paper itself.

docs/tutorials/change_detection.ipynb
hkristen: That is a leftover from the time before I switched the dataset to HF.

docs/tutorials/change_detection.ipynb
hkristen: Sure we could use the Tensorbaord log for viz, but oscd built-in plot() method isnt particulary great, especially not for a tutorial -> it doesn't show the prediction alongside ground truth with proper contrast normalization. 

Ideally we would improve this in the OSCD datamodule ofc. but it might be insightful for user to see how to build a custom inference viz themselfs.

adamjstewart: Can we fix the OSCD plot method then? Yes, it's insightful, but that doesn't mean it belongs in a _TorchGeo_ tutorial/paper as opposed to a _matplotlib_ tutorial.

hkristen: Sure, I can open a PR for make OSCD plotting nicer.

docs/tutorials/change_detection.ipynb
hkristen: I looked into this: The S2 pretrained Swin weights (SENTINEL2_MI_MS_SATLAS) are swin_v2 only, which is architecturally incompatible with BTC's standard swin_t/s/b, and BTC also hardcodes c=3 in its forward pass. 

Enabling this is separate PR, that we can mention here if you want. But it is out of scope for this tutorial imho.

docs/tutorials/change_detection.ipynb
hkristen: For this limited dataset, using only BTC + Cityscapes weights with RGB bands actually led to the by far best metrics. Hence me showing it here. 

Regarding pretrained weights see comment below.

docs/tutorials/change_detection.ipynb
hkristen: https://github.com/torchgeo/torchgeo/pull/3571

adamjstewart (approved):
I think this could still be made simpler, but let's merge as is and try to later incorporate inference on geospatial data. Once we try to add this to the paper, it will be more clear which things are easy or hard to write in that constrained environment.

PR #3223, opened 2025-12-23
Update Sentinel-1 regex to support S1C, S1D

lnicola:
https://www.esa.int/Applications/Observing_the_Earth/Copernicus/Sentinel-1

### DIFF
--- a/torchgeo/datasets/sentinel.py
+++ b/torchgeo/datasets/sentinel.py
@@ -121,7 +121,7 @@ class Sentinel1(Sentinel):
     # ssss:       Product ID
     filename_glob = 'S1*{}.*'
     filename_regex = r"""
-        ^S1(?P<mission>[AB])
+        ^S1(?P<mission>[A-D])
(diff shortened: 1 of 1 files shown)


torchgeo/datasets/sentinel.py
```diff
@@ -121,7 +121,7 @@ class Sentinel1(Sentinel):
     # ssss:       Product ID
     filename_glob = 'S1*{}.*'
     filename_regex = r"""
-        ^S1(?P<mission>[AB])
+        ^S1(?P<mission>[ABCD])
```
adamjstewart:
```suggestion
        ^S1(?P<mission>[A-D])
```

adamjstewart:
Are there any plans for 1E? We could use A-Z if we want to be future-proof.

lnicola:
Not as far as I know, but one can can hope :).

adamjstewart (approved):

PR #3230, opened 2025-12-29
Add Tile2Vec ResNet-22 model

Pratik0405:
- Used  "timm.create_model" no copied resnet code.
- Removes classification head.
- Registers model via "api.py".
- Includes unit test for forward pass.

### DIFF
--- a/torchgeo/models/api.py
+++ b/torchgeo/models/api.py
@@ -99,7 +99,7 @@ _model: dict[str, Callable[..., nn.Module]] = {
     'vit_small_patch14_dinov2': vit_small_patch14_dinov2,
 }
 
-_model_weights: dict[str | Callable[..., nn.Module], WeightsEnum] = {
+_model_weights: dict[str | Callable[..., nn.Module], WeightsEnum] = {  # type:ignore[invalid-assignment]
     aurora_swin_unet: Aurora_Weights,
     copernicusfm_base: CopernicusFM_Base_Weights,
     croma_base: CROMABase_Weights,
@@ -202,7 +202,7 @@ def get_weight(name: str) -> WeightsEnum:
     """
     for weight_name, weight_enum in _model_weights.items():
         if isinstance(weight_name, str):
-            for sub_weight_enum in weight_enum:
+            for sub_weight_enum in weight_enum:  # type: ignore[non-iterable]
                 if name == str(sub_weight_enum):
                     return sub_weight_enum
 
--- a/tests/models/test_api.py
+++ b/tests/models/test_api.py
@@ -152,7 +152,7 @@ def test_get_model_weights(builder: Callable[..., nn.Module]) -> None:
 
 @pytest.mark.parametrize('enum', enums)
 def test_get_weight(enum: WeightsEnum) -> None:
-    for weight in enum:
+    for weight in enum:  # type: ignore[not-iterable]
         assert weight == get_weight(str(weight))
 
 
--- a/tests/models/test_aurora.py
+++ b/tests/models/test_aurora.py
@@ -8,7 +8,6 @@ import pytest
 import torch
 from _pytest.fixtures import SubRequest
 from pytest import MonkeyPatch
-from torchvision.models._api import WeightsEnum
 
 from torchgeo.models import Aurora_Weights, aurora_swin_unet
 
@@ -17,11 +16,13 @@ pytest.importorskip('aurora')
 
 class TestAurora:
     @pytest.fixture(params=[*Aurora_Weights])
-    def weights(self, request: SubRequest) -> WeightsEnum:
-        return request.param
+    def weights(self, request: SubRequest) -> Aurora_Weights:
+        return request.param  # type: ignore[no-any-return]
 
     @pytest.fixture
-    def mocked_weights(self, tmp_path: Path, monkeypatch: MonkeyPatch) -> WeightsEnum:
+    def mocked_weights(
+        self, tmp_path: Path, monkeypatch: MonkeyPatch
+    ) -> Aurora_Weights:
         import aurora
 
         weights = Aurora_Weights.HRES_T0_PRETRAINED_SMALL_AURORA
@@ -29,21 +30,21 @@ class TestAurora:
         monkeypatch.setattr(
             aurora.Aurora, 'load_checkpoint', lambda self, *args, **kwargs: None
         )
-        return weights
+        return weights  # type: ignore[no-any-return]
 
     def test_aurora_swin_unet(self) -> None:
         aurora_swin_unet()
 
-    def test_aurora_swin_unet_weights(self, mocked_weights: WeightsEnum) -> None:
+    def test_aurora_swin_unet_weights(self, mocked_weights: Aurora_Weights) -> None:
         aurora_swin_unet(weights=mocked_weights)
 
     @pytest.mark.slow
-    def test_aurora_swin_unet_download(self, weights: WeightsEnum) -> None:
+    def test_aurora_swin_unet_download(self, weights: Aurora_Weights) -> None:
         aurora_swin_unet(weights=weights)
 
     @pytest.mark.slow
     @torch.inference_mode()
-    def test_aurora_prediction(self, weights: WeightsEnum) -> None:
+    def test_aurora_prediction(self, weights: Aurora_Weights) -> None:
         from aurora import Batch, Metadata
 
         model = aurora_swin_unet(weights=weights)
--- a/tests/models/test_copernicusfm.py
+++ b/tests/models/test_copernicusfm.py
@@ -8,7 +8,6 @@ import torch
 from _pytest.fixtures import SubRequest
 from pytest import MonkeyPatch
 from torch import Tensor
-from torchvision.models._api import WeightsEnum
 
 from torchgeo.models import CopernicusFM, CopernicusFM_Base_Weights, copernicusfm_base
 from torchgeo.models.copernicusfm import (
@@ -120,19 +119,19 @@ class TestCopernicusFM:
 
 class TestCopernicusFMBase:
     @pytest.fixture(params=[*CopernicusFM_Base_Weights])
-    def weights(self, request: SubRequest) -> WeightsEnum:
-        return request.param
+    def weights(self, request: SubRequest) -> CopernicusFM_Base_Weights:
+        return request.param  # type: ignore[no-any-return]
 
     @pytest.fixture
     def mocked_weights(
         self, tmp_path: Path, monkeypatch: MonkeyPatch, load_state_dict_from_url: None
-    ) -> WeightsEnum:
+    ) -> CopernicusFM_Base_Weights:
         weights = CopernicusFM_Base_Weights.CopernicusFM_ViT
         path = tmp_path / f'{weights}.pth'
         model = copernicusfm_base()
         torch.save(model.state_dict(), path)
         monkeypatch.setattr(weights.value, 'url', str(path))
-        return weights
+        return weights  # type: ignore[no-any-return]
 
     def test_copernicusfm_spectral(self) -> None:
         model = copernicusfm_base()
@@ -157,9 +156,11 @@ class TestCopernicusFMBase:
         input_mode = 'variable'
         model(x, metadata, language_embed=language_embed, input_mode=input_mode)
 
-    def test_copernicusfm_weights(self, mocked_weights: WeightsEnum) -> None:
+    def test_copernicusfm_weights(
+        self, mocked_weights: CopernicusFM_Base_Weights
+    ) -> None:
         copernicusfm_base(weights=mocked_weights)
 
     @pytest.mark.slow
-    def test_copernicusfm_download(self, weights: WeightsEnum) -> None:
+    def test_copernicusfm_download(self, weights: CopernicusFM_Base_Weights) -> None:
         copernicusfm_base(weights=weights)
--- a/tests/models/test_croma.py
+++ b/tests/models/test_croma.py
@@ -7,7 +7,6 @@ import pytest
 import torch
 from _pytest.fixtures import SubRequest
 from pytest import MonkeyPatch
-from torchvision.models._api import WeightsEnum
 
 from torchgeo.models import (
     CROMA,
@@ -55,53 +54,53 @@ class TestCROMA:
 
 class TestCROMABase:
     @pytest.fixture(params=[*CROMABase_Weights])
-    def weights(self, request: SubRequest) -> WeightsEnum:
-        return request.param
+    def weights(self, request: SubRequest) -> CROMABase_Weights:
+        return request.param  # type: ignore[no-any-return]
 
     @pytest.fixture
     def mocked_weights(
         self, tmp_path: Path, monkeypatch: MonkeyPatch, load_state_dict_from_url: None
-    ) -> WeightsEnum:
+    ) -> CROMABase_Weights:
         weights = CROMABase_Weights.CROMA_VIT
         path = tmp_path / f'{weights}.pth'
         model = croma_base()
         save_model(model, path)
         monkeypatch.setattr(weights.value, 'url', str(path))
-        return weights
+        return weights  # type: ignore[no-any-return]
 
     def test_croma(self) -> None:
         croma_base()
 
-    def test_croma_weights(self, mocked_weights: WeightsEnum) -> None:
+    def test_croma_weights(self, mocked_weights: CROMABase_Weights) -> None:
         croma_base(weights=mocked_weights)
 
     @pytest.mark.slow
-    def test_croma_download(self, weights: WeightsEnum) -> None:
+    def test_croma_download(self, weights: CROMABase_Weights) -> None:
         croma_base(weights=weights)
 
 
 class TestCROMALarge:
     @pytest.fixture(params=[*CROMALarge_Weights])
-    def weights(self, request: SubRequest) -> WeightsEnum:
-        return request.param
+    def weights(self, request: SubRequest) -> CROMALarge_Weights:
+        return request.param  # type: ignore[no-any-return]
 
     @pytest.fixture
     def mocked_weights(
         self, tmp_path: Path, monkeypatch: MonkeyPatch, load_state_dict_from_url: None
-    ) -> WeightsEnum:
+    ) -> CROMALarge_Weights:
         weights = CROMALarge_Weights.CROMA_VIT
         path = tmp_path / f'{weights}.pth'
         model = croma_large()
         save_model(model, path)
         monkeypatch.setattr(weights.value, 'url', str(path))
-        return weights
+        return weights  # type: ignore[no-any-return]
 
     def test_croma(self) -> None:
         croma_large()
 
-    def test_croma_weights(self, mocked_weights: WeightsEnum) -> None:
+    def test_croma_weights(self, mocked_weights: CROMALarge_Weights) -> None:
         croma_large(weights=mocked_weights)
 
     @pytest.mark.slow
-    def test_croma_download(self, weights: WeightsEnum) -> None:
+    def test_croma_download(self, weights: CROMALarge_Weights) -> None:
         croma_large(weights=weights)
--- a/tests/models/test_dofa.py
+++ b/tests/models/test_dofa.py
@@ -7,7 +7,6 @@ import pytest
 import torch
 from _pytest.fixtures import SubRequest
 from pytest import MonkeyPatch
-from torchvision.models._api import WeightsEnum
 
 from torchgeo.models import (
     DOFA,
@@ -77,19 +76,19 @@ class TestDOFASmall16:
 
 class TestDOFABase16:
     @pytest.fixture(params=[*DOFABase16_Weights])
-    def weights(self, request: SubRequest) -> WeightsEnum:
-        return request.param
+    def weights(self, request: SubRequest) -> DOFABase16_Weights:
+        return request.param  # type: ignore[no-any-return]
 
     @pytest.fixture
     def mocked_weights(
         self, tmp_path: Path, monkeypatch: MonkeyPatch, load_state_dict_from_url: None
-    ) -> WeightsEnum:
+    ) -> DOFABase16_Weights:
         weights = DOFABase16_Weights.DOFA_MAE
         path = tmp_path / f'{weights}.pth'
         model = dofa_base_patch16_224()
         torch.save(model.state_dict(), path)
         monkeypatch.setattr(weights.value, 'url', str(path))
-        return weights
+        return weights  # type: ignore[no-any-return]
 
     def test_dofa(self) -> None:
         model = dofa_base_patch16_224()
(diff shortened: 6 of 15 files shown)


isaaccorley:
This PR currently just creates a timm resnet18 backbone which has nothing about it particular to Tile2Vec. 

I would recommend adding a Tile2Vec module (similar to my implementation here https://github.com/isaaccorley/torchgeo-experimental/blob/main/src/tgexp/tile2vec/model.py) to our trainers instead which allows for training any arbitrary timm backbone using the Tile2Vec sampling (sampling can come later).

We can also use PyTorch's native TripletMarginLoss https://docs.pytorch.org/docs/stable/generated/torch.nn.TripletMarginLoss.html

adamjstewart:
@Pratik0405 is a student working with me. We're planning on adding the Tile2Vec weights for benchmark comparison. They are the oldest GeoFM I can think of.

We could definitely also add a trainer someday, but I want @Pratik0405 to focus on the model weights first.

isaaccorley:
Ahh makes sense. I was just going off the current PR but so far I'm also excited to have Tile2Vec in TorchGeo.

tests/models/test_tile2vec.py
adamjstewart, line 1: Needs license/copyright text, same as all other files

torchgeo/models/api.py
adamjstewart, line 1: API tests are failing, you need to update `tests/models/test_api.py` as well.

adamjstewart (changes requested):
For licensing and linting, see https://torchgeo.readthedocs.io/en/stable/user/contributing.html.

Let's next focus on getting the weights loaded correctly.

adamjstewart:
I dug into this and it's a lot more complicated than I expected:

* timm and torchvision ResNets are hard-coded to 4 layers, with N basic blocks per layer, and 2 convs per basic block.
* Tile2Vec's ResNet is hard-coded to 5 layers, with N=2 basic blocks per layer, and 3 convs per basic block (the 3rd conv isn't actually used though). This should really be called a ResNet-32 (1 + 5x2x3 + 1) or ResNet-22 (1 + 5x2x2 + 1).
* there are a few parameter name differences (downsample -> shortcut)
* conv1 has a different kernel size (7 -> 3)

So we either:

1. remove layer5 and rename a bunch of stuff in the weights until it can load into a standard ResNet-18 with a new conv1, or
2. copy-n-paste the full ResNet implementation from Tile2Vec, rename it TileNet, and load the Tile2Vec weights normally, or
3. convince timm/torchvision models to be more flexible and accept a variable number of layers (would take too long for @Pratik0405) 

My personal preference is for 2, but I'm curious if @isaaccorley @calebrob6 have thoughts. Which would be fairest to compare Tile2Vec performance vs. other models?

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,143 @@
+# Copyright (c) TorchGeo Contributors.
```
adamjstewart:
```suggestion
# Copyright (c) TorchGeo Contributors. All rights reserved.
```
Same for all other files you create.

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,143 @@
...
+# Licensed under the MIT License.
+
```
adamjstewart, line 4: The MIT license requires attribution and copyright notice preservation:
```python
# Adapted from https://github.com/ermongroup/tile2vec. Copyright (c) 2024 Ermon Group
```

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,143 @@
...
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torch.hub import load_state_dict_from_url
+
+__all__ = ["TileNet", "make_tilenet", "tilenet"]
+
+_TILENET_NAIP_URL = (
+    "https://huggingface.co/pgangapurwala/TileNet_Weights.NAIP_ALL_TILE2VEC/resolve/main/TileNet_Weights.NAIP_ALL_TILE2VEC.pth"
```
adamjstewart:
```suggestion
    "https://hf.co/pgangapurwala/TileNet_Weights.NAIP_ALL_TILE2VEC/resolve/af12210f5c130af76579ce8ec5e7036c1551ba25/TileNet_Weights.NAIP_ALL_TILE2VEC.pth"
```
Using a git commit hash instead of "main" will result in stable downloads.

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,143 @@
...
+        x = self.layer3(x)
+        x = self.layer4(x)
+        x = self.layer5(x)
+        x = F.avg_pool2d(x, 4)
+        return x.view(x.size(0), -1)
+
+
+def make_tilenet(in_channels=4, z_dim=512):
+    """Internal factory."""
+    return TileNet(in_channels=in_channels, z_dim=z_dim)
```
adamjstewart: This doesn't look very useful, let's get rid of this.

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torch.hub import load_state_dict_from_url
+
+__all__ = ["TileNet", "make_tilenet", "tilenet"]
+
+_TILENET_NAIP_URL = (
```
adamjstewart: We should use a Weights enum just like all other models.

adamjstewart (changes requested):
Looking better now. We need to make ruff and mypy happy by reformatting some things and adding type hints. Also need to add weight enums and rewrite the tests to match all existing model tests. Look at `torchgeo/models/resnet.py` and `tests/models/test_resnet.py` for examples of how to structure things. You should be able to mostly copy-n-paste the tests and just rename a couple things.

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,224 @@
...
+
+# -----------------------------------------------------------------------------
+# Weights
+# -----------------------------------------------------------------------------
+class TileNet_Weights(WeightsEnum):  # type: ignore[misc]
+    """TileNet (Tile2Vec) weights.
+
+    NAIP-pretrained Tile2Vec encoder.
+
+    .. versionadded:: 0.7
```
adamjstewart:
```suggestion
    .. versionadded:: 0.9
```

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,224 @@
...
+    """
+
+    NAIP = Weights(
+        url=(
+            'https://hf.co/pgangapurwala/'
+            'TileNet_Weights.NAIP_ALL_TILE2VEC/resolve/'
+            'af12210f5c130af76579ce8ec5e7036c1551ba25/'
+            'TileNet_Weights.NAIP_ALL_TILE2VEC.pth'
```
adamjstewart: Can be on a single line, ruff ignores long URLs automatically

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,224 @@
...
+        out = self.bn2(self.conv2(out))
+        out += self.shortcut(x)
+        return F.relu(out)
+
+
+# -----------------------------------------------------------------------------
+# TileNet model
+# -----------------------------------------------------------------------------
+class TileNet(nn.Module):
+    """TileNet encoder (Tile2Vec NAIP exact)."""
```
adamjstewart: What does "NAIP exact" mean here? The architecture should work for any inputs, not just NAIP. It is only the weights that expect NAIP.

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,224 @@
...
+
+    name: str = 'tilenet'
+    embedding_dim: int = 512
```
adamjstewart: What are these for? I don't see them in the original implementation and they don't seem to be used anywhere?

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,224 @@
...
+        )
+        self.bn1 = nn.BatchNorm2d(64)
+
+        self.layer1 = self._make_layer(64, 2, stride=1)
+        self.layer2 = self._make_layer(128, 2, stride=2)
+        self.layer3 = self._make_layer(256, 2, stride=2)
+        self.layer4 = self._make_layer(512, 2, stride=2)
+        self.layer5 = self._make_layer(z_dim, 2, stride=2, no_relu=True)
+
+    def _make_layer(
```
adamjstewart: Can you add a docstring here too?

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,224 @@
...
+        strides = [stride] + [1] * (num_blocks - 1)
+        layers = []
+
+        for i, s in enumerate(strides):
+            layers.append(
+                BasicBlock(
+                    self.in_planes,
+                    planes,
+                    stride=s,
+                    no_relu=no_relu and i == num_blocks - 1,
```
adamjstewart: Seems different from the original implementation?

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,224 @@
...
+        return nn.Sequential(*layers)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        """Compute TileNet embeddings.
+
+        Args:
+        x: Input image tensor of shape (B, C, H, W).
+
+        Returns:
+        Embedding tensor of shape (B, embedding_dim).
```
adamjstewart: Indentation is wrong here

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,224 @@
...
+    """Create a TileNet encoder.
+
+    Args:
+      in_channels: Number of input channels.
+      z_dim: Output embedding dimension.
+
+    Returns:
+      A TileNet model instance.
+    """
+    return TileNet(in_channels=in_channels, z_dim=z_dim)
```
adamjstewart: Not needed, this 11-line function only saves 1-line of code. I would remove this and directly call TileNet.

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,224 @@
...
+# Public factory (TorchGeo API)
+# -----------------------------------------------------------------------------
+def tilenet(
+    weights: TileNet_Weights | None = None, *args: Any, **kwargs: Any
+) -> nn.Module:
+    """TileNet (Tile2Vec) encoder.
+
+    Args:
+        weights: Pre-trained TileNet weights to load.
+        *args: Positional arguments (unused, kept for API compatibility).
```
adamjstewart: Seems like they are used to me?

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,224 @@
...
+        missing_keys, unexpected_keys = model.load_state_dict(
+            weights.get_state_dict(progress=True), strict=True
+        )
+        assert missing_keys == []
+        assert unexpected_keys == []
+
+    return model
+
+
+__all__ = ['TileNet', 'TileNet_Weights', 'make_tilenet', 'tilenet']
```
adamjstewart: This is not needed. When it does exist, it's usually at the top of the file. I would delete this line.

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,224 @@
...
+# -----------------------------------------------------------------------------
+class TileNet_Weights(WeightsEnum):  # type: ignore[misc]
+    """TileNet (Tile2Vec) weights.
+
+    NAIP-pretrained Tile2Vec encoder.
+
+    .. versionadded:: 0.7
+    """
+
+    NAIP = Weights(
```
adamjstewart:
```suggestion
    NAIP_ALL_TILE2VEC = Weights(
```
Makes this a little more specific

tests/models/test_tilenet.py
```diff
@@ -0,0 +1,94 @@
...
+        c = weights.meta['in_chans']
+        sample = {'image': torch.arange(c * 50 * 50, dtype=torch.float).view(c, 50, 50)}
+        weights.transforms(sample)
+
+    @pytest.mark.slow
+    def test_tilenet_download(self, weights: WeightsEnum) -> None:
+        """Test real weight download."""
+        tilenet(weights=weights)
+
+    @pytest.mark.optional
```
adamjstewart: This mark does not exist in our pyproject.toml and thus cannot be used. It's also causing warnings during test runs. Either make this test required or remove it.

adamjstewart (changes requested):
Looking much better now, mostly minor comments remaining.

Also missing changes to:

* `docs/api/models.rst`
* `docs/api/models/tilenet.rst`
* `docs/api/weights/naip.csv`
* `hubconf.py`

@isaaccorley anything else I'm missing?

torchgeo/models/tilenet.py
adamjstewart: Also needs:
```rst
.. versionadded:: 0.9
```

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,224 @@
...
+    return TileNet(in_channels=in_channels, z_dim=z_dim)
+
+
+# -----------------------------------------------------------------------------
+# Public factory (TorchGeo API)
+# -----------------------------------------------------------------------------
+def tilenet(
+    weights: TileNet_Weights | None = None, *args: Any, **kwargs: Any
+) -> nn.Module:
+    """TileNet (Tile2Vec) encoder.
```
adamjstewart, line 161: Also needs:
```rst
.. versionadded:: 0.9
```

torchgeo/models/tilenet.py
Pratik0405: docs/api/models.rst
docs/api/models/tilenet.rst
docs/api/weights/naip.csv
hubconf.py

Dear @adamjstewart what kind of changes are required in here ?

torchgeo/models/tilenet.py
Pratik0405: yes my bad, I was a lot thinking about the proper mentioning and the citations so made a mistake it.

torchgeo/models/tilenet.py
Pratik0405: yes I will remove it.

torchgeo/models/tilenet.py
adamjstewart: * `docs/api/models.rst`: add `models/tilenet` (next file) to the toctree
* `docs/api/models/tilenet.rst`: document the model API, see neighboring files for examples
* `docs/api/weights/naip.csv`: documents the weights, see other weights for examples
* `hubconf.py`: add tilenet to the list of models supported

torchgeo/models/tilenet.py
Pratik0405: You’re right! In the updated PR, I will remove the no_relu logic so that every block matches the original architecture and pretrained weights.

tests/models/test_tilenet.py
Pratik0405: It's optional so I will remove it.

torchgeo/models/tilenet.py
Pratik0405: yup rookie mistake.

torchgeo/models/tilenet.py
Pratik0405: I’ve removed the make_tilenet helper and now instantiate TileNet directly.

torchgeo/models/tilenet.py
Pratik0405: I didn't understand this.

adamjstewart: Ah I understand now. Can we change the code so that these actually get used? See `dofa.py` for an example of this.

Pratik0405: Yes, *args are forwarded to TileNet(), so thought to add the docstrings for it

calebrob6:
I remember reproducing some of the Tile2Vec results back in the day -- do you see any result from the paper that you could try to reproduce with these to make sure its working as expected @Pratik0405 ?

docs/api/models/tilenet.rst
```diff
@@ -0,0 +1,34 @@
...
+
+Model API
+---------
+
+.. autofunction:: torchgeo.models.tilenet
+
+Internal Factory
+----------------
+
+.. autofunction:: torchgeo.models.tilenet
```
adamjstewart: This is duplicated, causing the doc build to fail. As usual, if you simply copy the existing style and change a few lines, things will build fine, no need to add most of this.

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,197 @@
...
+
+        .. versionadded:: 0.9
```
adamjstewart: Reduce indentation

docs/api/models/tilenet.rst
Pratik0405: Ah, okay

adamjstewart (dismissed):
As far as I can tell this is ready now. Would love a review from our models expert @isaaccorley. Also hoping @Pratik0405 can show some quick model performance metrics to confirm that the weights are working as expected.

Pratik0405:
[tile2vec_test.py](https://github.com/user-attachments/files/24568825/tile2vec_test.py)

tested the model, everything looks fine, if all the dependencies are installed correctly. @adamjstewart @calebrob6 

Results:
Loaded 1000 valid tiles
Example tile shape: (4, 50, 50)
Embedding shape: torch.Size([1, 8192])
Mean embedding norm: 95921.875
Std embedding norm: 13902.6826171875
Self-distance: 0.0

Pratik0405:
Tile2Vec linear probe accuracy: 0.225
Raw pixel accuracy: 0.170
Number of classes: 28

Some more results @adamjstewart @calebrob6

tests/models/test_tilenet.py
```diff
@@ -0,0 +1,87 @@
...
+from pytest import MonkeyPatch
+from torchvision.models._api import WeightsEnum
+
+from torchgeo.models import TileNet_Weights, tilenet
+
+
+class TestTileNet:
+    """Tests for TileNet (Tile2Vec) model."""
+
+    # ------------------------------------------------------------------
```
isaaccorley: Can we remove these comment headers throughout

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,197 @@
...
+        ),
+        transforms=T.Normalize(mean=[0], std=[255], inplace=True),
+        meta={
+            'dataset': 'NAIP',
+            'in_chans': 4,
+            'model': 'tilenet',
+            'ssl_method': 'tile2vec',
+            'publication': 'https://arxiv.org/abs/1805.02855',
+            'repo': 'https://github.com/ermongroup/tile2vec',
+            'bands': ['R', 'G', 'B', 'NIR'],
```
isaaccorley: Let's make this a tuple not a list so it's immutable

adamjstewart:
> I do think we should call it Tile2VecNet to be explicit and not "TileNet" which could confuse some users on what this is.

I dug into this too. In the paper, they call the model a ResNet-18, which is really misleading because it has 22 layers (32 if you enable a flag). In the source code, there are two names:

* [ResNet](https://github.com/ermongroup/tile2vec/blob/master/src/resnet.py): this is the version they actually use to load the weights, and the version we implement here
* [TileNet](https://github.com/ermongroup/tile2vec/blob/master/src/tilenet.py): this looks like an experimental effort that wasn't used in the end?

In my mind, Tile2Vec is the sampling strategy, whereas ResNet is the model architecture. But their ResNet is kinda custom, so I was wary of naming it `resnet22` or something like that. So personally, I would prefer "ResNet-22" or "TileNet" with Tile2Vec mentioned in the docstring but not in the name.

Would be cool to get opinions from the original Tile2Vec authors if they check GitHub: @nealjean @sherriewang @ermonste

torchgeo/models/tilenet.py
adamjstewart: Someday we should make a TypedDict for our model metadata to enforce this.

isaaccorley:
True but since the code lives inside tile2vec.py it feels awkward to call it resnet22

Edit: I think this can actually be rewritten using timm's ResNet class and just passing custom args to it to make it a ResNet-22. Then we could have it live in torchgeo/models/resnet.py and put the weights enum in there as well.

adamjstewart:
> Edit: I think this can actually be rewritten using timm's ResNet class and just passing custom args to it to make it a ResNet-22. Then we could have it live in torchgeo/models/resnet.py and put the weights enum in there as well.

Trust me, I went there: https://github.com/torchgeo/torchgeo/pull/3230#issuecomment-3715129098

Pratik0405:
Yes @isaaccorley me and @adamjstewart had a lot of discussion on this it is not feasible to implement this using timm

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,197 @@
...
+        x = self.layer3(x)
+        x = self.layer4(x)
+        x = self.layer5(x)
+        x = F.avg_pool2d(x, 4)
+        return x.view(x.size(0), -1)
+
+
+# -----------------------------------------------------------------------------
+# Public factory (TorchGeo API)
+# -----------------------------------------------------------------------------
```
adamjstewart: Still a lot of comment headers we could remove

Pratik0405: Ah okay, I thought that only the ones in tests were to be removed; I will remove these as well.

calebrob6:
> Tile2Vec linear probe accuracy: 0.225 Raw pixel accuracy: 0.170 Number of classes: 28
> 
> Some more results @adamjstewart @calebrob6

What is this linear probe accuracy on?

adamjstewart:
I believe this was on a crop type mapping dataset but don't know which one. I asked @Pratik0405 to test his model on https://github.com/ermongroup/tile2vec/blob/master/examples/Example%203%20-%20Tile2Vec%20features%20for%20CDL%20classification.ipynb to make sure it gets the same accuracy. We both agree the current accuracy is quite low.

torchgeo/models/tilenet.py
```diff
@@ -0,0 +1,185 @@
...
+    """TileNet (Tile2Vec) weights.
+
+    NAIP-pretrained Tile2Vec encoder.
+
+    .. versionadded:: 0.9
+    """
+
+    NAIP_ALL_TILE2VEC = Weights(
+        url=(
+            'https://hf.co/pgangapurwala/TileNet_Weights.NAIP_ALL_TILE2VEC/resolve/af12210f5c130af76579ce8ec5e7036c1551ba25/TileNet_Weights.NAIP_ALL_TILE2VEC.pth'
```
isaaccorley, line 30: the repo naming is a bit odd. let's just call the repo tile2vec and the filename should just be `model-XXXXXX.pth` where ~~the `XXXXXX` is the last 6 digits of the output of `md5sum model.pth`~~

First 8 digits of sha256sum actually

adamjstewart: Isn't it supposed to be the first 8 digits of the sha256? Could also rename if/when we onboard to our HF org.

isaaccorley: You're right it's the first 8 of sha256 my memory was wrong

adamjstewart: Can also do the rename during you .pt2 conversion

isaaccorley:
@adamjstewart I'm good with this if you want to merge. I previously also noticed their pretrained weights weren't particularly good because their dataset is NAIP only and quite small with small patches so the low acc doesn't surprise me. Someone should create a new Tile2Vec++ with some of the newest training tricks and a larger dataset **hint hint**

Pratik0405:
I tried the tile2vec example 3 - Tile2Vec features for CDL classification.ipynb
these following were the results that I got 
Mean accuracy: 0.6541
Standard deviation: 0.0332

And the implementation on the github has following results:
Mean accuracy: 0.6876
Standard deviation: 0.0299

@adamjstewart @isaaccorley @calebrob6

adamjstewart:
Seems within margin of error to me.

PR #3235, opened 2026-01-01
fix: XarrayDataset file handle leak

isaaccorley:
Close xarray dataset handles after merging in XarrayDataset._merge_files(), preventing memory leaks during dataset operations.

Using contextlib ExitStack because we can't use the typical `with xr.open...` since we are opening multiple files in a loop.


adamjstewart (approved):
Interesting, first time seeing this. I think RasterDataset will be much harder to fix though, as we purposefully cache open filehandles to speed up reading.

PR #3237, opened 2026-01-02
feat: add TileDataset and tile samplers for patch sampling

calebrob6:
Adds `TileDataset` and tile samplers for sampling patches from large raster files without geospatial overhead.

- **`TileDataset`**: Standard PyTorch `Dataset` that loads image (and optional mask) patches by `(file_index, y, x, patch_size)` tuple
- **`RandomTileSampler`**: Samples random patches weighted by tile area for uniform spatial coverage  
- **`GridTileSampler`**: Samples patches in regular grid pattern for inference with edge coverage

```python
from torchgeo.datasets import TileDataset
from torchgeo.samplers import RandomTileSampler, GridTileSampler

dataset = TileDataset(image_paths=["tile1.tif", "tile2.tif"])
sampler = RandomTileSampler(dataset, patch_size=256, length=1000)
loader = DataLoader(dataset, sampler=sampler)
```
Closes #1353


calebrob6:
@adamjstewart the last commit (https://github.com/torchgeo/torchgeo/pull/3237/changes/3bda321e0b32818c8b225240ab77b14fa2205de8) shows what ChesapeakeCVPR would look like if it extended TileDataset. Would put this in a different PR but I think it greatly simplifies that dataset (note that the DataModule is basically the same as the NonGeoDatamodule).

Some notes from before about why this is helpful. There exist plenty of lightly preprocessed datasets that span many CRSs in the form of (imagery,) or (imagery, masks) that you don't _need_ to project to a single CRS to work with. These don't work with GeoDataset (I know we talked about a no reprojection thing but that hasn't happened) and only work with NonGeoDataset if they happen to all have the same patch size. This TileDataset handles these cases in a nice way.

adamjstewart:
I'm still concerned about having a third type of base class that is not compatible with GeoDataset or NonGeoDataset, but I added this to the TSC agenda for further discussion.

adriantre:
@calebrob6 what do you think about a best-of-both-worlds solution, where this works with GeoSamplers?

The motivation is to speed up data loading by avoiding unnecessary (and potentially harmful) reprojection, not to move away from geosampling.

The data/files could still be indexed in some common crs, and `__getitem__` could convert/snap it to a rectified `x, y, patch_size, patch_size` to avoid reprojection.

I know there are a lot of considerations wrt. overlaps and crs, but starting with an opinionated first version that immensely speed up training might make it easier to shape the future of GeoSampler, GDS.index etc.

Thoughts?

calebrob6:
@adriantre if I'm understanding you correctly, this is what I had implemented for the ChesapeakeCVPR dataset and it is a bit of a mess. Say you have a raster dataset from different CRSs, if you generate bounding boxes in a single CRS, then reproject those to the CRSs of the individual rasters, then they will (potentially) be different sizes (in the ChesapeakeCVPR dataset I take center crops of a certain size in pixels and make the sampled boxes big enough). To get around this, we'd have to have a GeoDataset that doesn't reproject, and that we could index by (raster index, bbox) and at that point its kind of a TileDataset as I've implemented here anyway.

adamjstewart:
@adriantre what you're suggesting is basically #409? I think that's also a good direction worth pursuing. I think there's a lot of room for improvement here by storing the native CRS of each file and using majority voting to decide what CRS to load as. For datasets in UTM, we would only ever reproject near the boundary of two zones.

To add on to what @calebrob6 already said, one other major difference is that TileDataset would work even for datasets that lack geographic metadata. For example, OSCD has very large images we would like to randomly sample from, but it can never be a GeoDataset (unless we create some kind of fake index).

In summary, making GeoDataset faster and NonGeoDataset more powerful are both worth doing. I think of TileDataset as a proof-of-concept improvement to NonGeoDataset. It's an experimental feature that will eventually be merged into NonGeoDataset.

This also gets to my desire to merge GeoDataset and NonGeoDataset into a single base class (or two very similar API-compatible base classes). For example, both could optionally support integer and/or bbox indexing. This would make it a little easier for developers to decide whether they should use NonGeo or GeoDataset for curated benchmark datasets that also have geographic metadata. But that's even further off in the dream pipeline.

adriantre:
> This also gets to my desire to merge GeoDataset and NonGeoDataset into a single base class (or two very similar API-compatible base classes). For example, both could optionally support integer and/or bbox indexing. 

Yes! This is essentially what I thought as well. De-coupling the index from the reading, and bbox from `__getitem__`. Some middleware (override of getitem) that converts geo-query to pixel-query.

And if that middleware knows what patch size and resolution you want to read, it can create a rectified bbox in the source file crs, avoiding the issue @calebrob6 mentioned (?). So the GeoSampler would produce points to cover, rather than final boxes, that's up to getitem.

And the box can be constructed around the point without always centring on it. (Or this can be done in transforms later by first sampling bigger-than-you-need, then cropping the way you want)

adamjstewart:
> I think there's a lot of room for improvement here by storing the native CRS of each file and using majority voting to decide what CRS to load as.

Note that this will break #3214, so it isn't that simple.

adriantre:
From how I understand #3214 this idea is not breaking it. 

They store each patch output to separate temp tifs. This could be stored in the same crs as the patch was read from. Then in on_predict_epoch_end it would warp and merge them into a common crs.

Kind of the opposite way from how torchgeo does it now, if that makes sense?

adamjstewart:
We only have access to metadata stored in the sample dictionary. And only Tensors can be transferred to the GPU. Since PyTorch doesn't support string Tensors, we would have to use EPSG codes, which don't cover all possible CRSs. Unless we can figure out a clever way to store the CRS in a dict key or something.

adriantre:
Actually, the callback hooks have access to the trainer, which gives access to the datamodule and datasets and so on. So it is definitely doable imo. The output needs to be transferred to cpu before writing anyways, so I don't think we need to transfer crs to gpu.

https://lightning.ai/docs/pytorch/stable/_modules/lightning/pytorch/callbacks/prediction_writer.html#BasePredictionWriter.on_predict_batch_end

calebrob6:
@adamjstewart is this something you hate? I just worked on ChesapeakeCVPR and was reminded that this PR is a thing I still want in main torchgeo, but it doesn't seem like we agreed on a plan here.

adamjstewart:
My opinion remains the same: I'm okay with making NonGeoDataset smarter but not okay with creating a third competing base class that is incompatible with both NonGeoDataset and GeoDataset. @evgeniialappo is going to do some benchmarking and see if we can quickly get all RasterDatasets to be as fast as TileDataset. If that can't be done, we can revisit this.

calebrob6:
How is the benchmarking/optimization process going?

PR #3242, opened 2026-01-03
perf: RCF/MOSAIKS speedup

isaaccorley:
Previously MOSAIKS/RCF was performing very large conv operation twice. this PR fuses this into a single conv operation and reduce memory usage on the relu operations as well. Also added `torch.inference_mode()` decorator to the forward pass. This model isn't trainable but we should explicitly turn off gradient computation to reduce GPU memory usage as well.

```python
        x1a = F.relu(
            F.conv2d(x, self.weights, bias=self.biases, stride=1, padding=0),
            inplace=True,
        )
        x1b = F.relu(
            -F.conv2d(x, self.weights, bias=self.biases, stride=1, padding=0),
            inplace=False,
        )

        x1a = F.adaptive_avg_pool2d(x1a, (1, 1)).squeeze()
        x1b = F.adaptive_avg_pool2d(x1b, (1, 1)).squeeze()

        if len(x1a.shape) == 1:  # case where we passed a single input
            output = torch.cat((x1a, x1b), dim=0)
            return output
        else:  # case where we passed a batch of > 1 inputs
            assert len(x1a.shape) == 2
            output = torch.cat((x1a, x1b), dim=1)
            return output
```

to 

```python
        y = F.conv2d(x, self.weights, bias=self.biases, stride=1, padding=0)
        x1a = y.clamp_min(0).mean(dim=(-2, -1))
        x1b = y.clamp_max(0).neg_().mean(dim=(-2, -1))
        return torch.cat((x1a, x1b), dim=1)
```

### Benchmarks

Did some benchmarks across different batch sizes at image size 256x256 and it's roughly `2x` faster and uses `half` the memory as the original while the outputs are identical.

Note: on my Macbook Pro the original implementation OOM with batch size=128, while the new OOMs after 256.

cc: @estherrolf

### DIFF
--- a/tests/models/test_rcf.py
+++ b/tests/models/test_rcf.py
@@ -25,11 +25,11 @@ class TestRCF:
         model = RCF(in_channels=5, features=4, kernel_size=3, mode='gaussian')
         x = torch.randn(2, 5, 64, 64)
         y = model(x)
-        assert y == (2, 4)
+        assert y.shape == (2, 4)
 
         x = torch.randn(1, 5, 64, 64)
         y = model(x)
-        assert y == (1, 4)
+        assert y.shape == (1, 4)
 
     def test_untrainable(self) -> None:
         model = RCF(in_channels=5, features=4, kernel_size=3, mode='gaussian')


adamjstewart (approved):
LGTM but will wait for @estherrolf to review

estherrolf:
looks good to me, thanks for making the improvement!

adamjstewart (approved):

Issue #1746, opened 2023-11-28
RasterDataset: no support for images with positive y-resolution


tanpinquan:
### Description

Certain geotifs have their axes flipped such as [this one](http://umbra-open-data-catalog.s3.amazonaws.com/sar-data/tasks/ad%20hoc/Sebarok%20Island%2C%20Singapore/6f073ed1-efdd-499f-bfdc-c102b7daf307/2023-06-23-03-23-26_UMBRA-06/2023-06-23-03-23-26_UMBRA-06_GEC.tif), which has the following transform flipping the direction of the x-axis:
`Affine(-0.346241067933152, 0.2915401535597796, 366317.96341952495, 0.2915401535597782, 0.3462410679331555, 130406.04105973356)`.

Depending on the transform, I believe this affects how random bounding boxes bounds should be ordered . For the example geotif above with x-axis flipped, [this line](https://github.com/microsoft/torchgeo/blob/4cc0cbb3276d4aae614a3c12381ad9b1e6a53292/torchgeo/datasets/geo.py#L536) maybe should be `bounds = (query.maxx, query.miny, query.minx, query.maxy)` instead of `bounds = (query.minx, query.miny, query.maxx, query.maxy)`. I have not tested it myself.
 
RandomGeoSampler and RasterDataset do not check for this, resulting in a `Bounds and transform are inconsistent` error

### Steps to reproduce

1. Download the sample data from [here](http://umbra-open-data-catalog.s3.amazonaws.com/sar-data/tasks/ad%20hoc/Sebarok%20Island%2C%20Singapore/6f073ed1-efdd-499f-bfdc-c102b7daf307/2023-06-23-03-23-26_UMBRA-06/2023-06-23-03-23-26_UMBRA-06_GEC.tif)
2. Run the following code to reproduce
 ```python
from torch.utils.data import DataLoader
from torchgeo.datasets import RasterDataset
from torchgeo.samplers import RandomGeoSampler
    
    
class UmbraDataset(RasterDataset):

    # umbra tifs are named like: 2023-08-02-17-08-40_UMBRA-05_GEC.tif
    filename_glob = '*.tif'
    filename_regex = r"^(?P<date>\d{4}-\d{2}-\d{2}-\d{2}-\d{2}-\d{2})"
    date_format = "%Y-%m-%d-%H-%M-%S"
    is_image = True
    separate_files = False 
    

dataset = UmbraDataset('/path/to/data/')

[...]

PR #3244, opened 2026-01-04
Add Google Satellite Embedding (AlphaEarth Foundations) dataset

adamjstewart:
- [x] Basic data loader
- [x] Timestamp extraction
- [x] Upside down raster support: #1746
- [x] Documentation
- [x] Tests

@shelhamer

Depends on #3200, #3249

sc
hf

### DIFF
--- a/tests/datasets/test_gse.py
+++ b/tests/datasets/test_gse.py
@@ -40,7 +40,7 @@ class TestGoogleSatelliteEmbedding:
         assert isinstance(x, dict)
         assert isinstance(x['image'], torch.Tensor)
         # Rasterio may return nodata for upside down rasters
-        assert not torch.any(x['image'] == -128.0)
+        assert not torch.all(x['image'] == -128.0)
 
     def test_and(self, dataset: GoogleSatelliteEmbedding) -> None:
         ds = dataset & dataset


torchgeo/datasets/gse.py
```diff
@@ -0,0 +1,107 @@
...
+       * `Source Cooperative <https://source.coop/tge-labs/aef>`__: 2018--2024
+       * `Hugging Face <https://huggingface.co/datasets/Major-TOM/Core-AlphaEarth-Embeddings>`__: subset matching Major TOM
+
+    .. versionadded:: 0.9
+    """
+
+    # TODO: only possible to get year from directory path, e.g.,
+    #
+    # * GCS/SC: 2024/10N/x086q72fv2f9q1x4a-0000000000-0000000000.tiff
+    # * HF:     2024/U/1/L/7/471U_587L.tif
```
adamjstewart: @calebrob6 has complained about the complexity and inflexibility of our current `filename_regex` approach before. I'm thinking about adding a `filepath_to_timestamp` (name TBD) method that defaults to using `filename_regex` but allows users to write custom code to extract the timestamp from the file path.

torchgeo/datasets/gse.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def plot(
+        self, sample: Sample, show_titles: bool = True, suptitle: str | None = None
+    ) -> Figure:
+        """Plot a sample from the dataset.
+
+        .. warning::
+           Visualizations are generated using PCA on each image *individually*, and
+           are thus not comparable across images. The plot method is provided for
+           visualization purposes only and should not be used to draw conclusions.
```
adamjstewart, line 97: Also, `torch.pca_lowrank` is non-deterministic, requiring a fixed seed for reproducibility.

docs/api/datasets/geo_datasets.csv
```diff
@@ -7,6 +7,7 @@ Dataset,Type,Source,License,Size (px),Resolution (m)
 `Chesapeake Land Cover`_,"Imagery, Masks",NAIP,"CC0-1.0",-,1
 `GlobalBuildingMap`_,Masks,PlanetScope,CC-BY-4.0,180K,3
 `Global Mangrove Distribution`_,Masks,"Remote Sensing, In Situ Measurements","public domain",-,3
+`Google Satellite Embedding`_,Embedding,"Sentinel 1/2, Landsat 8/9",CC-BY-4.0,"8,192",10
```
adamjstewart, line 11: I'm planning to eventually make separate tables for embeddings with more details once we have more embeddings datasets.

torchgeo/datasets/gse.py
```diff
@@ -0,0 +1,128 @@
...
+        """
+        # Example file paths:
+        #
+        # * GCS/SC: 2024/10N/x086q72fv2f9q1x4a-0000000000-0000000000.tiff
+        # * HF:     2024/U/1/L/7/471U_587L.tif
+        for part in pathlib.Path(filepath).parts[::-1]:
+            try:
+                return disambiguate_timestamp(part, self.date_format)
+            except ValueError:
+                pass
```
adamjstewart, line 85: Not the prettiest code I've ever written, but it works?

torchgeo/datasets/gse.py
adamjstewart: Honestly, PCA doesn't look significantly better than just using the first 3 channels:
gse2

torchgeo/datasets/gse.py
isaaccorley: I couldn't come up with a better solution either. Maybe in a future PR we can raise specific errors so we don't catch ValueError

adamjstewart:
> Google Satellite Embedding is less known compared to AlphaEarth Embeddings I think. We should either rename or just make an alias.

Trying to follow the same branding that Google uses. A google search for "AlphaEarth Embeddings" only turns up links to "Google Satellite Embedding" or "AlphaEarth Foundations". I did add AEF to the docstring though. Hopefully anyone searching for that keyword will find it.

> since the HF version from Major TOM is not the complete dataset, I think we can drop supporting it but I'll leave that up to you

There aren't really any lines of code specific to this other than the extra test data. Might as well keep.

PR #3251, opened 2026-01-07
fix: incorrect dataset splits with random_grid_cell_assignment

### LINKED ISSUE #3191: Incorrect dataset splits with random_grid_cell_assignment when the dataset uses a non-rectangular ROI
### Description

The random_grid_cell_assignment function excludes grid cells that fall outside the ROI. This can happen when the function is used on a dataset that has been split with a custom ROI (via roi_split).

In this case, the distribution of cells to the splits is incorrect. The problem is that the split lengths are determined assuming that all (grid_size × grid_size) cells are included in the new datasets, even though some may be discarded.

The culprit line in the function seems to be:

```python
lengths = _fractions_to_lengths(fractions, len(dataset) * grid_size**2)
```

### Steps to reproduce

Run the following code:

```python
import geopandas as gpd
from shapely.geometry import Polygon, box
from torchgeo.datasets import VectorDataset, random_grid_cell_assignment, roi_split

# Create a test dataset with a square polygon
geom = box(0, 0, 2, 2) # Save a simple box (0,0)-(2,2)
gdf = gpd.GeoDataFrame({"geometry": [geom]})
gdf.to_file("test.geojson", driver="GeoJSON")

dataset = VectorDataset("test.geojson")

# Define a triangle ROI within the bounds of the polygon
coords = [(0, 0), (2, 0), (1, 2)]
roi = Polygon(coords)

# Split the dataset using the ROI
dataset = roi_split(dataset, [roi])[0]

# Run random_grid_cell_assignment with ROI
# 2 of the 16 grid cells will be outside the triangle ROI
splits = random_grid_cell_assignment(
    dataset=dataset,
    fractions=[0.5, 0.5],
    grid_size=4,
)

print("Split sizes:", [len(s) for s in splits])
```
The output is:
```Split sizes: [8, 6]```

But should be:
```Split sizes: [7, 7]```

(Total number of cells is 14 because 2 are excluded)

### Version

0.8.0

danielKlmr:
Fix incorrect dataset split proportions in the random_grid_cell_assignment function when grid cells are excluded for falling outside a non‑rectangular ROI.

Fixes #3191

### DIFF
--- a/torchgeo/datasets/splits.py
+++ b/torchgeo/datasets/splits.py
@@ -205,8 +205,6 @@ def random_grid_cell_assignment(
     if grid_size < 2:
         raise ValueError('Input grid_size must be greater than 1.')
 
-    lengths = _fractions_to_lengths(fractions, len(dataset) * grid_size**2)
-
     # Generate the grid's cells for each bbox in index
     left = []
     right = []
@@ -231,6 +229,8 @@ def random_grid_cell_assignment(
                     right.append(index.right)
                     rows.append(row)
                     geometry.append(geom)
+
+    lengths = _fractions_to_lengths(fractions, len(rows))
 
     indexes_sr = pd.IntervalIndex.from_arrays(
         left, right, closed='both', name='datetime'


adamjstewart (approved):
Looks more correct to me. I won't make you add tests because we may end up redesigning a lot of these splitters in the future. For example, if `dataset.index` contains multiple files with overlapping bounds, there is currently nothing in place to ensure that the same location doesn't end up in multiple splits.

PR #3255, opened 2026-01-07
Air Quality dataset

keves1:
This PR adds the Air Quality time series data set from the [UCI Machine Learning Repository](https://archive.ics.uci.edu/dataset/360/air+quality). This is split off from PR #2639 to be merged separately.


torchgeo/datamodules/air_quality.py
```diff
@@ -0,0 +1,75 @@
...
+from typing import Any
+
+from torch import Tensor
+from torch.utils.data import Subset
+
+from ..datasets import AirQuality
+from .geo import NonGeoDataModule
+
+
+class AirQualityDataModule(NonGeoDataModule):
```
adamjstewart, line 15: Should probably remove this data module since we can't test it until we add the trainer.

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,122 @@
...
+            + self.num_past_steps
+            + self.num_future_steps
+        ]
+        past_targets = torch.tensor(past_targets.values, dtype=torch.float32)
+        future_targets = torch.tensor(future_targets.values, dtype=torch.float32)
+
+        mean = past_targets.mean(dim=0, keepdim=True)
+        std = past_targets.std(dim=0, keepdim=True)
+        past_targets_normalized = (past_targets - mean) / (std + 1e-12)
+        future_targets_normalized = (future_targets - mean) / (std + 1e-12)
```
adamjstewart, line 97: I normally like to put normalization in the datamodule. I realize it's a little harder for pandas datasets. We also need to ensure that targets are unnormalized before metrics calculation.

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,122 @@
...
+        pathname = os.path.join(self.root, self.data_file_name)
+        if os.path.exists(pathname):
+            return pd.read_csv(pathname)
+
+        # Check if the user requested to download the dataset
+        if not self.download:
+            raise DatasetNotFoundError(self)
+
+        # Download the dataset
+        return pd.read_csv(self.url, na_values=-200)
```
adamjstewart, line 122: Missing special handling of Date/Time columns which are strings that can't be cast to floats. Something like:
```python
df['datetime'] = pd.to_datetime(df['Date'] + ' ' + df['Time'])
df.drop(columns=['Date', 'Time'], inplace=True)
```
Also need special handling to convert datetime64 to a Tensor.

adamjstewart: Also missing a plot method. I have some nice plotting examples in my example notebook. Can steal from there.

torchgeo/datamodules/air_quality.py
```diff
@@ -0,0 +1,75 @@
...
+        Args:
+            stage: Either 'fit', 'validate', 'test', or 'predict'.
+        """
+        dataset = AirQuality(**self.kwargs)
+        train_split_pct = 1 - (self.val_split_pct + self.test_split_pct)
+        train_size = int(train_split_pct * len(dataset))
+        val_size = int(self.val_split_pct * len(dataset))
+        train_indices = range(train_size)
+        val_indices = range(train_size, train_size + val_size)
+        test_indices = range(train_size + val_size, len(dataset))
```
adamjstewart, line 58: This will result in train/val/test overlap, as the last element of train includes the first few elements of val, and so on.

tests/data/air_quality/data.py
```diff
@@ -0,0 +1,33 @@
...
+    'NO2(GT)',
+    'PT08.S4(NO2)',
+    'PT08.S5(O3)',
+    'T',
+    'RH',
+    'AH',
+]
+
+nrows = 50
+data = np.random.rand(nrows, len(columns))
```
adamjstewart, line 28: Need to update example fake test data to match real data. For example, Date/Time are not floats, and we should randomly add -200s to simulate missing data. Otherwise this dataset doesn't actually work.

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def __len__(self) -> int:
+        """Return the number of data points in the dataset.
+
+        Returns:
+            length of the dataset
+        """
+        return len(self.data) - (self.num_past_steps + self.num_future_steps)
+
+    def __getitem__(self, index: int) -> dict[str, Any]:
```
adamjstewart, line 76:
```suggestion
    def __getitem__(self, index: int) -> Sample:
```

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,122 @@
...
+    * has missing features
+
+    If you use this dataset in your research, please cite:
+
+    * https://doi.org/10.1016/J.SNB.2007.09.060
+
+    .. versionadded:: 0.9
+    """
+
+    url = 'https://archive.ics.uci.edu/static/public/360/data.csv'
```
adamjstewart, line 41: Could redistribute on HF if we don't want to rely on the original source, it's CC-BY-4.0.

adamjstewart (changes requested):
I likely won't have time to finish this dataset before the 0.9 release this week. @f-schi would you be able to finish this PR since you'll be using it in your AQ tutorial? @keves1 is too busy to complete this at the moment.

adamjstewart:
Closed by #3619

PR #3262, opened 2026-01-08
Add missing copyright headers to 1 Python files

ominiverdi:
## Summary
- Add standard MIT license copyright headers to 2 files that were missing them
- Discovered during OSGeo incubation code provenance review
## Files Changed
- `docs/user/metrics/update_timeline.py`
- `tests/data/landcoverai/split.py`
## Context
During the OSGeo incubation review, a scan of all 557 Python files found these 2 files were missing the standard copyright header used throughout the codebase.

### DIFF
--- a/docs/user/metrics/update_timeline.py
+++ b/docs/user/metrics/update_timeline.py
@@ -1,4 +1,6 @@
 #!/usr/bin/env python3
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
 
 import argparse
 


adamjstewart:
Changes to `update_timeline.py` look good.

Changes to `split.py` _should_ result in the sha256 of that file changing, causing the tests to fail. The fact that it doesn't is not good. Anyway, we are removing this file in #3238 so that problem shouldn't exist for long.

adamjstewart:
`split.py` has now been removed. Can you rebase?

PR #3264, opened 2026-01-09
Dataset.__getitem__: use index consistently

adamjstewart:
`ty check` flags the following errors in ~30 places:
```
error[invalid-method-override]: Invalid override of method `__getitem__`
    --> torchgeo/datasets/geo.py:1483:9
     |
1481 |         self.index = pd.concat([dataset1.index, dataset2.index])
1482 |
1483 |     def __getitem__(self, query: GeoSlice) -> Sample:
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Definition is incompatible with `Dataset.__getitem__`
1484 |         """Retrieve input, target, and/or metadata indexed by spatiotemporal slice.
     |
    ::: /Users/Adam/spack/var/spack/environments/default/.spack-env/._view/a3mkzy47bhv6e4txcybzfff66oh5q6a5/lib/python3.13/site-packages/torch/utils/data/dataset.py:58:9
     |
  56 |     """
  57 |
  58 |     def __getitem__(self, index) -> _T_co:
     |         --------------------------------- `Dataset.__getitem__` defined here
  59 |         raise NotImplementedError("Subclasses of Dataset should implement __getitem__.")
     |
info: This violates the Liskov Substitution Principle
info: rule `invalid-method-override` is enabled by default
```
ty is _technically_ correct, subclasses should use the same parameter names as parent classes, otherwise they aren't truly subclasses. mypy is more relaxed, making an exception for `__getitem__` because no human writes code like:
```python
ds.__getitem__(index=1)
```
instead of:
```python
ds[1]
```

### Pros

* Better to fix the error than to silence ty every time?
* Consistency between PyTorch's Dataset and all TorchGeo datasets.
* Docstrings subjectively make a little more sense with this name?

### Cons

* "index" is an **insanely** overloaded term.

  We are literally indexing `self.index.index` (the catalog/dataframe's index) using our `index` (GeoSlice). We also have band indexes and spectral indexes if that isn't enough.

* _Technically_ not backwards compatible.

  The following code will no longer work:
  ```python
  ds.__getitem__(query=(slice(xmin, xmax), slice(ymin, ymax)))
  ```
  Users will unfortunately have to switch to:
  ```python
  ds[xmin:xmax, ymin:ymax]
  ```

I made the change to `torchgeo/datasets/geo.py` first to get feedback. I really don't like overloading `index` even more, but ty isn't _wrong_, just pedantic, and I want to make ty happy. If this gets a couple 👍, then I'll update the other GeoDatasets to match.

### DIFF
--- a/torchgeo/datasets/inaturalist.py
+++ b/torchgeo/datasets/inaturalist.py
@@ -79,7 +79,7 @@ class INaturalist(GeoDataset):
             Sample of input, target, and/or metadata at that index.
 
         Raises:
-            IndexError: If *index* is not found in the index.
+            IndexError: If *index* is not found in the dataset.
         """
         x, y, t = self._disambiguate_slice(index)
         interval = pd.Interval(t.start, t.stop)
--- a/torchgeo/datasets/l7irish.py
+++ b/torchgeo/datasets/l7irish.py
@@ -66,7 +66,7 @@ class L7IrishMask(RasterDataset):
             Sample of input, target, and/or metadata at that index.
 
         Raises:
-            IndexError: If *index* is not found in the index.
+            IndexError: If *index* is not found in the dataset.
         """
         sample = super().__getitem__(index)
         sample['mask'] = self.ordinal_map[sample['mask']]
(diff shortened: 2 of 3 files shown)


adamjstewart:
Could also rename `self.index` to `self.catalog` or something like that if we want things to be more backwards-incompatible but less overloaded.

isaaccorley:
This lgtm, Copilot found a couple things I missed but other than that this is good to merge

Issue #3258, opened 2026-01-08
PASTIS100 dataset/datamodule


robmarkcole:
### Summary

To determine the requirements for a PASTIS100 dataset for testing. Propose the first 100 samples from the PASTIS-R dataset

### Additional information

https://huggingface.co/datasets/satellite-image-deep-learning/PASTIS-R-100

PR #3265, opened 2026-01-10
ADD PASTIS100

robmarkcole:
Closes https://github.com/torchgeo/torchgeo/issues/3258

### DIFF
--- a/tests/datasets/test_pastis.py
+++ b/tests/datasets/test_pastis.py
@@ -41,26 +41,18 @@ class TestPASTIS:
         mode = params['mode']
         transforms = nn.Identity()
 
-        if base_class is PASTIS:
-            md5 = '135a29fb8221241dde14f31579c07f45'
-            monkeypatch.setattr(base_class, 'md5', md5)
-            url = os.path.join('tests', 'data', 'pastis', 'PASTIS-R.zip')
-            monkeypatch.setattr(base_class, 'url', url)
-            return base_class(
-                root=root,
-                folds=folds,
-                bands=bands,
-                mode=mode,
-                transforms=transforms,
-                download=True,
-                checksum=True,
-            )
-
-        src = os.path.join('tests', 'data', 'pastis', 'PASTIS-R')
-        dst = tmp_path / 'PASTIS-R'
-        shutil.copytree(src, dst)
+        md5 = '135a29fb8221241dde14f31579c07f45'
+        monkeypatch.setattr(base_class, 'md5', md5)
+        url = os.path.join('tests', 'data', 'pastis', 'PASTIS-R.zip')
+        monkeypatch.setattr(base_class, 'url', url)
         return base_class(
-            root=root, folds=folds, bands=bands, mode=mode, transforms=transforms
+            root=root,
+            folds=folds,
+            bands=bands,
+            mode=mode,
+            transforms=transforms,
+            download=True,
+            checksum=True,
         )
 
     def test_getitem_semantic(self, dataset: PASTIS) -> None:


robmarkcole:
@adamjstewart the correct url should be https://huggingface.co/datasets/torchgeo/PASTIS-R-100/resolve/7288f86b79e87da71d4789bebce05c4f6e242b8b/PASTIS-R-100.zip ? Any further thoughts on the naming?

tests/datamodules/test_pastis.py
adamjstewart, line 1: Would like to get rid of this file and instead test these with `SemanticSegmentationTask`.

tests/datasets/test_pastis.py
```diff
@@ -101,3 +101,43 @@ def test_plot(self, dataset: PASTIS) -> None:
             x['prediction_labels'] = x['label'].clone()
         dataset.plot(x)
         plt.close()
+
+
+class TestPASTISR100:
```
adamjstewart: Can we do what we did for eurosat and instead parametrize the fixture over `[PASTIS, PASTISR100]`? Then we don't need to add any extra lines of testing code.

robmarkcole: @adamjstewart  Do this or https://github.com/torchgeo/torchgeo/pull/3265#discussion_r2736163386 ?

adamjstewart: Both, these are 2 different files.

tests/datamodules/test_pastis.py
robmarkcole: Done, LMK if we need this new test data or not

tests/trainers/test_segmentation.py
```diff
@@ -92,6 +93,12 @@ def test_trainer(
         match name:
             case 'ftw':
                 pytest.importorskip('pyarrow')
+            case 'pastis100':
```
robmarkcole: @adamjstewart is there a better way?

tests/datasets/test_pastis.py
```diff
@@ -13,31 +15,52 @@
...
-        bands = request.param['bands']
-        mode = request.param['mode']
+        folds = params['folds']
+        bands = params['bands']
+        mode = params['mode']
         transforms = nn.Identity()
-        return PASTIS(
-            root, folds, bands, mode, transforms, download=True, checksum=True
+
+        if base_class is PASTIS:
```
adamjstewart: PASTIS100 doesn't need mocking to prevent download during testing?

If you remove `checksum=True`, you don't even need to mock md5.

robmarkcole: You want the 2GB dataset downloaded in testing then?

tests/trainers/test_segmentation.py
adamjstewart: No idea what you're doing here. The `root` directory in `tests/conf/pastis100.yaml` is correct, so it shouldn't need any extra handling to work.

adamjstewart: Could rename your extracted directory, or change the design of the dataset to remove directory entirely (assume `root` contains the files directly without any extra level of abstraction).

tests/datamodules/test_pastis.py
adamjstewart: We should not need any new test data.

tests/datasets/test_pastis.py
adamjstewart: No tests should have any downloads. I'm asking why PASTIS has mocking but not PASTIS100.

robmarkcole: PASTIS100 isn’t mocked because the fixture avoids download entirely: it copies the already-extracted local fixture directory tests/data/pastis/PASTIS-R to the expected on-disk name PASTIS-R-100 on line 60

tests/trainers/test_segmentation.py
```diff
@@ -114,6 +116,13 @@ def test_trainer(
             '1',
         ]
 
+        if name == 'pastis100':
```
robmarkcole: instead of 
```python
case 'pastis100':
                monkeypatch.setattr(PASTIS100, 'directory', 'PASTIS-R')
```

adamjstewart: If you rename the directory in your zipfile (or the extracted directory) so that both are PASTIS-R, you won't need this at all.

tests/datasets/test_pastis.py
adamjstewart: Can this be simplified now that they are the same "download" url? Basically change back to the previous code and do this for both datasets.

PR #3266, opened 2026-01-11
Update the CLI tutorial with more options

calebrob6:
The CLI is much more useful when it saves the models during training :) -- it also allows you to pick LR scheduler and optimizer without overriding the trainer.

### DIFF
--- a/docs/tutorials/cli.ipynb
+++ b/docs/tutorials/cli.ipynb
@@ -215,7 +215,7 @@
    "source": [
     "## Optimizer and Learning Rate Scheduler\n",
     "\n",
-    "By default, TorchGeo trainers use Adam with a ReduceLROnPlateau scheduler. However, you can customize these via the `optimizer` and `lr_scheduler` sections in your config.\n",
+    "By default, TorchGeo trainers use AdamW with a ReduceLROnPlateau scheduler. However, you can customize these via the `optimizer` and `lr_scheduler` sections in your config.\n",
     "\n",
     "In our example, we use:\n",
     "\n",
@@ -235,7 +235,7 @@
    "source": [
     "## Callbacks\n",
     "\n",
-    "Lightning callbacks allow you to hook into the training loop at various points. In our config, we use two essential callbacks:\n",
+    "Lightning callbacks allow you to hook into the training loop at various points. In this example config, we use two callbacks:\n",
     "\n",
     "### ModelCheckpoint\n",
     "\n",
@@ -249,7 +249,7 @@
     "\n",
     "### LearningRateMonitor\n",
     "\n",
-    "`LearningRateMonitor` conveniently logs the current learning rate to your logger which is useful to make sure your learning rate schedule is working as expected."
+    "`LearningRateMonitor` logs the current learning rate to your logger which is useful to make sure your learning rate schedule is working as expected."
    ]
   },
   {


docs/tutorials/cli.ipynb
```diff
@@ -172,13 +197,99 @@
...
    "metadata": {},
    "source": [
-    "This YAML file has three sections:\n",
+    "This YAML file has five main sections:\n",
+    "\n",
+    "* **trainer**: Arguments to pass to the [Trainer](https://lightning.ai/docs/pytorch/stable/common/trainer.html), including callbacks\n",
+    "* **model**: Arguments to pass to the task\n",
+    "* **data**: Arguments to pass to the data module\n",
+    "* **optimizer**: Custom optimizer configuration\n",
+    "* **lr_scheduler**: Learning rate scheduler configuration\n",
```
adamjstewart: Can we make it clear, either in the YAML file or the comment here, which sections are optional and which are required? I think only model and data are technically required? Don't want to scare away new users for the sake of helping power users.

docs/tutorials/cli.ipynb
```diff
@@ -146,21 +146,46 @@
     "root = os.path.join(tempfile.gettempdir(), 'cli')\n",
     "config = f\"\"\"\n",
     "trainer:\n",
-    "  max_epochs: 1\n",
+    "  max_epochs: 2\n",
     "  default_root_dir: '{root}'\n",
+    "  callbacks:\n",
+    "    - class_path: lightning.pytorch.callbacks.ModelCheckpoint\n",
```
adamjstewart, line 152: Out of curiosity, does `ModelCheckpoint` without the full path work? What if you add:
```python
import lightning.pytorch.callbacks
```
in `torchgeo/main.py`?

docs/tutorials/cli.ipynb
```diff
@@ -146,21 +146,46 @@
...
     "data:\n",
     "  class_path: EuroSAT100DataModule\n",
     "  init_args:\n",
     "    batch_size: 8\n",
     "  dict_kwargs:\n",
     "    root: '{root}'\n",
     "    download: true\n",
+    "\n",
+    "optimizer:\n",
+    "  class_path: torch.optim.AdamW\n",
```
adamjstewart, line 179: Same here, we could theoretically add:
```python
import torch.optim
import torch.optim.lr_scheduler
```
to `torchgeo/main.py` if we want things to be more magic.

docs/tutorials/cli.ipynb
calebrob6: I think ModelCheckpoint without the full path worked, but I want to keep the full path as I don't really like magic

docs/tutorials/cli.ipynb
calebrob6: Annotated optimizer and lr_scheduler in this list as optional, keeping the others as is because a user is going to need to change trainer anyway

docs/tutorials/cli.ipynb
```diff
@@ -172,13 +197,99 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "693bb09d",
+   "metadata": {},
+   "source": [
+    "## Optimizer and Learning Rate Scheduler\n",
+    "\n",
+    "By default, TorchGeo trainers use Adam with a ReduceLROnPlateau scheduler. However, you can customize these via the `optimizer` and `lr_scheduler` sections in your config.\n",
```
adamjstewart: We actually default to AdamW, not Adam

docs/tutorials/cli.ipynb
```diff
@@ -172,13 +197,99 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ed5072f5",
+   "metadata": {},
+   "source": [
+    "## Callbacks\n",
+    "\n",
+    "Lightning callbacks allow you to hook into the training loop at various points. In our config, we use two essential callbacks:\n",
```
adamjstewart: I would personally consider EarlyStopping to be essential but LearningRateMonitor to not be essential. In fact, I rarely use any learning rate decay with optimizers like Adam that have dynamic lr built in. Maybe worth swapping these examples?

docs/tutorials/cli.ipynb
adamjstewart, line 370: Do you want to add more links here, for example, to the full list of optimizers/lr schedulers/callbacks?

calebrob6: I have that link above "See the [Lightning Callbacks documentation](https://lightning.ai/docs/pytorch/stable/extensions/callbacks.html) for more options."

docs/tutorials/cli.ipynb
calebrob6: Changed the "essential" wording as this is just an example

adamjstewart (approved):

PR #3269, opened 2026-01-12
torchgeo.trainers: better type checking of loggers

adamjstewart:
Resolves ty errors like:
```
> ty check torchgeo/trainers/change.py
error[call-non-callable]: Object of type `object` is not callable
   --> torchgeo/trainers/change.py:336:27
    |
334 |                 fig: Figure | None = None
335 |                 try:
336 |                     fig = datamodule.plot(sample)
    |                           ^^^^^^^^^^^^^^^^^^^^^^^
337 |                 except RGBBandsMissingError:
338 |                     pass
    |
info: rule `call-non-callable` is enabled by default

error[call-non-callable]: Object of type `object` is not callable
   --> torchgeo/trainers/change.py:342:21
    |
340 |                   if fig:
341 |                       summary_writer = self.logger.experiment
342 | /                     summary_writer.add_figure(
343 | |                         f'image/{batch_idx}', fig, global_step=self.global_step
344 | |                     )
    | |_____________________^
345 |                       plt.close()
    |
info: rule `call-non-callable` is enabled by default

Found 2 diagnostics
```
Pretty sure mypy just ignores these as Any, whereas ty doesn't like not knowing the type.

### DIFF
--- a/torchgeo/trainers/change.py
+++ b/torchgeo/trainers/change.py
@@ -15,7 +15,6 @@ import torch.nn as nn
 from einops import rearrange
 from matplotlib.figure import Figure
 from torch import Tensor
-from torch.utils.tensorboard.writer import SummaryWriter
 from torchmetrics import Accuracy, F1Score, JaccardIndex, MetricCollection
 from torchvision.models._api import WeightsEnum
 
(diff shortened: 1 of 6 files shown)


adamjstewart:
How are we testing these lines of code if we don't have any loggers installed in CI?

I don't immediately see an easy way of solving this other than adding a tensorboard dep... Only alternative I can think of is to define a Protocol for `add_figure` or ignore it. Probably going to ignore it if there are no better suggestions.

adamjstewart:
@robmarkcole reminded me that we discussed removing all figure logging from our trainers and leaving it up to the user to define their own custom callbacks. This would also be a good solution.

calebrob6 (dismissed):

isaaccorley:
Seems like we need to explicitly install tensorboard now by adding it to our deps

isaaccorley (approved):

PR #3275, opened 2026-01-13
Update documentation theme 

ashnair1:
- Updating theme to match TorchGeo colors
- Added link to Youtube channel

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -13,3 +13,9 @@ html {
   border-left-color: #ee4c2c;
   color: #ee4c2c;
 }
+
+/* Hide Jupyter notebook input/output prompts */
+.nbinput .prompt,
+.nboutput .prompt {
+  display: none;
+}
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -136,7 +136,7 @@ html_theme_options = {
 html_favicon = os.path.join('..', 'logo', 'favicon.ico')
 
 html_static_path = ['_static']
-html_css_files = ['notebook-prompt.css', 'custom.css']
+html_css_files = ['custom.css']
 
 # -- Extension configuration -------------------------------------------------
 
--- a/docs/_static/notebook-prompt.css
+++ b/docs/_static/notebook-prompt.css
@@ -1,4 +1,0 @@
-.nbinput .prompt,
-.nboutput .prompt {
-  display: none;
-}


docs/_static/custom.css
adamjstewart, line 1: Absolutely loving this. As far as I know, TorchGeo's colors are the same as Microsoft's colors because Microsoft designed our logo.

docs/conf.py
```diff
@@ -116,6 +116,23 @@
...
         },
+        {
+            'name': 'YouTube',
+            'url': 'https://www.youtube.com/@TorchGeo',
+            'icon': 'fa-brands fa-youtube',
+        },
+        {
+            'name': 'Hugging Face',
+            'url': 'https://huggingface.co/torchgeo',
+            'icon': '_static/hf-logo.svg',
```
adamjstewart: Would like to avoid non-standard logos, the color is also different. We use https://simpleicons.org/ for our README badges and they have HF/OSGeo/etc. Let's stick to YouTube only for now. In the future we can try to convince pydata-sphinx-theme to vendor simpleicons or font-awesome to add HF/OSGeo.

ashnair1: So the logos were from their branding guide but I agree it doesn't fit with the overall theme. Have dropped HF and OSGeo logos from this PR.
 
Open discussion for HF to be added to FontAwesome - FortAwesome/Font-Awesome#21322. Maybe we can open a similar one for OSGeo

docs/_static/custom.css
adamjstewart: Should we name this something more unique like colors.css?

ashnair1: We could. But then if we make other updates/customizations to the theme, they'll have to live in other css files. I don't foresee a lot of modifications so I prefer to have all them in a single css file.

`custom.css` is also what pydata uses - https://github.com/pydata/pydata-sphinx-theme/blob/main/docs/_static/custom.css

adamjstewart: In that case should we merge `notebook-prompt.css`? I'm relatively new to web development, never know what the best practices are.

ashnair1: We can. I think the general rule is if we only aim to do slight tweaks to the theme, a single css is fine. 

Full disclosure: I'm not a webdev either :)

PR #3278, opened 2026-01-13
TypeAlias -> type

adamjstewart:
Now that we upgraded Sphinx and dropped Python 3.11, we should be able to use the Python 3.12+ `type` statement.


adamjstewart:
Hmm, maybe we need to explicitly document these types?

adamjstewart:
Giving up for now, type alias support in Sphinx seems to be pretty immature.

adamjstewart:
Opened a bug report: https://github.com/sphinx-doc/sphinx/issues/14313

calebrob6:
I know this was waiting on Sphinx but no response on that issue. Is this still something we want to keep open?

adamjstewart:
These typing PRs are all still blocked by Sphinx. Definitely still something I want someday, but doesn't necessarily have to stay open until then.

adamjstewart:

Issue #3279, opened 2026-01-14
PASTIS with SemanticSegmentationTask results in expected input[1, 610, 128, 128] to have 10 channels, but got 610 channels instead


robmarkcole:
### Description

As title, the error is raised if we configure `in_channels=10` (number of S2 spectral channels) but as this is multi-temporal this should be 610 (`padding_length x num_channels`). This was not expected to me - perhaps the docs need clarification on this point?

CC @yichiac 

### Steps to reproduce

```python
import lightning as L
from torchgeo.datamodules import PASTISDataModule
from torchgeo.trainers import SemanticSegmentationTask

from torch import Tensor
import torch

batch_size = 1

datamodule = PASTISDataModule(
    root="data/",
    download=False,
    batch_size=batch_size,
)

datamodule.setup("fit")

task = SemanticSegmentationTask(
    model="unet",
    backbone="resnet18",
    weights=False,
    in_channels=10,
    num_classes=len(dataset.classes),
)

trainer = L.Trainer(
    max_epochs=1,
)

trainer.fit(task, datamodule)
```
error:
```
RuntimeError: Given groups=1, weight of size [64, 10, 7, 7], expected input[1, 610, 128, 128] to have 10 channels, but got 610 channels instead
```

### Version
`0.9.0.dev0`

adamjstewart:
This is something @yichiac and I debated in #2921 but never came up with a good design. What about:
```python
in_channels: int = 3,
time_steps: int = 1,
```
Then for models like U-Net, we can use `in_channels * time_steps`. This will be backwards-compatible for those using `in_channels=610` and forwards-compatible for those wanting to use `in_channels=10, time_steps=61`.

robmarkcole:
@adamjstewart so default `time_steps: int = 1,` then do we document `time_steps=61` on the dataset itself?

adamjstewart:
I don't really know how to document that, maybe it could squeeze into the dataset table. Many datasets have a variable number of time steps, but not all models support that. So it's kinda tricky.

robmarkcole:
Yes not really sure either - for the unet we actually pad to a specific number.. @MarcCoru any suggestions?

adamjstewart:
Closed by #3671, use the dedicated SpatioTemporalSegmentation task instead.

PR #3280, opened 2026-01-14
ConvLSTM model changes only

robmarkcole:
Support for convlstm in segmentation mode:

- Added a segmentation head `(nn.Conv2d)` controlled by `num_classes` and `head_kernel_size` parameters
- Renamed forward → forward_features for the raw hidden-state output; new forward runs the head and returns `(B, num_classes, H, W)` supports optional lengths for variable-length sequences
- Removed `batch_first` parameter — only `(b, t, c, h, w)` input is supported, consistent with TorchGeo's standard

### DIFF
--- a/torchgeo/models/convlstm.py
+++ b/torchgeo/models/convlstm.py
@@ -111,7 +111,6 @@ class ConvLSTM(nn.Module):
         hidden_dim: int | list[int],
         kernel_size: int | tuple[int, int] | list[int | tuple[int, int]],
         num_layers: int,
-        batch_first: bool = True,
         bias: bool = True,
         return_all_layers: bool = False,
         num_classes: int = 1,
@@ -129,8 +128,6 @@ class ConvLSTM(nn.Module):
                 * a tuple of two integers (for rectangular kernels)
                 * a list of integers or tuples (one for each layer)
             num_layers: Number of LSTM layers stacked on each other.
-            batch_first: If ``True``, then the input and output tensors are
-                provided as (b, t, c, h, w).
             bias: If ``True``, adds a learnable bias to the output.
             return_all_layers: If ``True``, will return the list of computations
                 for all layers.
@@ -159,7 +156,6 @@ class ConvLSTM(nn.Module):
 
         self.input_dim = input_dim
         self.num_layers = num_layers
-        self.batch_first = batch_first
         self.bias = bias
         self.return_all_layers = return_all_layers
         self.num_classes = num_classes
@@ -195,15 +191,12 @@ class ConvLSTM(nn.Module):
         .. versionadded:: 0.10
 
         Args:
-            input_tensor: A 5-D Tensor of shape (t, b, c, h, w) or (b, t, c, h, w).
+            input_tensor: A 5-D Tensor of shape (b, t, c, h, w).
             hidden_state: An optional initial hidden state.
 
         Returns:
             A tuple containing layer_output_list and last_state_list.
         """
-        if not self.batch_first:
-            input_tensor = input_tensor.permute(1, 0, 2, 3, 4)
-
         b, _, _, h, w = input_tensor.size()
 
         if hidden_state is None:
@@ -245,7 +238,7 @@ class ConvLSTM(nn.Module):
         """Forward pass for segmentation with the prediction head.
 
         Args:
-            input_tensor: A 5-D Tensor of shape (t, b, c, h, w) or (b, t, c, h, w).
+            input_tensor: A 5-D Tensor of shape (b, t, c, h, w).
             lengths: Optional sequence lengths (B,) before padding/truncation.
                 Values larger than the available sequence length use the final
                 timestep.
--- a/torchgeo/models/convlstm.py
+++ b/torchgeo/models/convlstm.py
@@ -114,6 +114,8 @@ class ConvLSTM(nn.Module):
         batch_first: bool = True,
         bias: bool = True,
         return_all_layers: bool = False,
+        num_classes: int | None = None,
+        head_kernel_size: int = 1,
     ) -> None:
         """Initializes the ConvLSTM model.
 
@@ -132,6 +134,9 @@ class ConvLSTM(nn.Module):
             bias: If ``True``, adds a learnable bias to the output.
             return_all_layers: If ``True``, will return the list of computations
                 for all layers.
+            num_classes: Optional number of segmentation classes for an attached
+                prediction head.
+            head_kernel_size: Kernel size for the optional segmentation head.
         """
         super().__init__()
 
@@ -157,6 +162,7 @@ class ConvLSTM(nn.Module):
         self.batch_first = batch_first
         self.bias = bias
         self.return_all_layers = return_all_layers
+        self.num_classes = num_classes
 
         cell_list = []
         for i in range(self.num_layers):
@@ -171,13 +177,24 @@ class ConvLSTM(nn.Module):
             )
 
         self.cell_list = nn.ModuleList(cell_list)
+        self.head: nn.Conv2d | None = None
+        if self.num_classes is not None:
+            padding = head_kernel_size // 2
+            self.head = nn.Conv2d(
+                in_channels=self.hidden_dim[-1],
+                out_channels=self.num_classes,
+                kernel_size=head_kernel_size,
+                padding=padding,
+            )
 
-    def forward(
+    def forward_features(
         self,
         input_tensor: torch.Tensor,
         hidden_state: list[tuple[torch.Tensor, torch.Tensor]] | None = None,
     ) -> tuple[list[torch.Tensor], list[tuple[torch.Tensor, torch.Tensor]]]:
-        """Forward pass of the ConvLSTM.
+        """Forward pass of ConvLSTM feature extraction.
+
+        .. versionadded:: 0.10
 
         Args:
             input_tensor: A 5-D Tensor of shape (t, b, c, h, w) or (b, t, c, h, w).
@@ -220,6 +237,48 @@ class ConvLSTM(nn.Module):
             last_state_list = last_state_list[-1:]
 
         return layer_output_list, last_state_list
+
+    def forward(
+        self,
+        input_tensor: torch.Tensor,
+        lengths: torch.Tensor | None = None,
+        hidden_state: list[tuple[torch.Tensor, torch.Tensor]] | None = None,
+    ) -> torch.Tensor:
+        """Forward pass for segmentation with the prediction head.
+
+        Args:
+            input_tensor: A 5-D Tensor of shape (t, b, c, h, w) or (b, t, c, h, w).
+            lengths: Optional sequence lengths (B,) before padding/truncation.
+                Values larger than the available sequence length use the final
+                timestep.
+            hidden_state: An optional initial hidden state.
+
+        Returns:
+            Output tensor of shape (B, num_classes, H, W).
+
+        Raises:
+            ValueError: If the segmentation head is not configured.
+        """
+        if self.head is None:
+            raise ValueError(
+                "Segmentation head is not configured. Set 'num_classes' to use "
+                'ConvLSTM for segmentation.'
+            )
+
+        layer_output_list, _ = self.forward_features(
+            input_tensor, hidden_state=hidden_state
+        )
+        layer_output = layer_output_list[-1]
+
+        if lengths is None:
+            features = layer_output[:, -1]
+        else:
+            idx = lengths.to(device=layer_output.device, dtype=torch.long) - 1
+            idx = idx.clamp(min=0, max=layer_output.size(1) - 1)
+            batch_idx = torch.arange(layer_output.size(0), device=idx.device)
+            features = layer_output[batch_idx, idx]
+
+        return self.head(features)
 
     def _init_hidden(
         self, batch_size: int, image_size: tuple[int, int]
--- a/torchgeo/models/convlstm.py
+++ b/torchgeo/models/convlstm.py
@@ -177,15 +177,13 @@ class ConvLSTM(nn.Module):
             )
 
         self.cell_list = nn.ModuleList(cell_list)
-        self.head: nn.Conv2d | None = None
-        if self.num_classes is not None:
-            padding = head_kernel_size // 2
-            self.head = nn.Conv2d(
-                in_channels=self.hidden_dim[-1],
-                out_channels=self.num_classes,
-                kernel_size=head_kernel_size,
-                padding=padding,
-            )
+        padding = head_kernel_size // 2
+        self.head = nn.Conv2d(
+            in_channels=self.hidden_dim[-1],
+            out_channels=self.num_classes,
+            kernel_size=head_kernel_size,
+            padding=padding,
+        )
 
     def forward_features(
         self,
--- a/torchgeo/models/convlstm.py
+++ b/torchgeo/models/convlstm.py
@@ -114,7 +114,7 @@ class ConvLSTM(nn.Module):
         batch_first: bool = True,
         bias: bool = True,
         return_all_layers: bool = False,
-        num_classes: int | None = None,
+        num_classes: int = 1,
         head_kernel_size: int = 1,
     ) -> None:
         """Initializes the ConvLSTM model.
--- a/torchgeo/models/convlstm.py
+++ b/torchgeo/models/convlstm.py
@@ -256,15 +256,7 @@ class ConvLSTM(nn.Module):
         Returns:
             Output tensor of shape (B, num_classes, H, W).
 
-        Raises:
-            ValueError: If the segmentation head is not configured.
         """
-        if self.head is None:
-            raise ValueError(
-                "Segmentation head is not configured. Set 'num_classes' to use "
-                'ConvLSTM for segmentation.'
-            )
-
         layer_output_list, _ = self.forward_features(
             input_tensor, hidden_state=hidden_state
         )
--- a/tests/models/test_convlstm.py
+++ b/tests/models/test_convlstm.py
@@ -205,18 +205,3 @@ class TestConvLSTM:
         y_hat_last_step = model(input_tensor, lengths=torch.tensor([t, t]))
 
         torch.testing.assert_close(y_hat, y_hat_last_step)
-
-    def test_convlstm_forward_requires_head(self) -> None:
-        """Test that segmentation forward requires a configured prediction head."""
-        b = 1
-        t = 2
-        c = 3
-        h = 8
-        w = 8
-        input_tensor = torch.rand(b, t, c, h, w)
-        model = ConvLSTM(
-            input_dim=c, hidden_dim=8, kernel_size=3, num_layers=1, batch_first=True
-        )
-
-        with pytest.raises(ValueError, match='Segmentation head is not configured'):
-            model(input_tensor)
--- a/tests/models/test_convlstm.py
+++ b/tests/models/test_convlstm.py
@@ -12,8 +12,8 @@ from torchgeo.models import ConvLSTM
 class TestConvLSTM:
     """Tests for the ConvLSTM model."""
 
-    def test_convlstm_forward(self) -> None:
-        """Test the forward pass of the ConvLSTM model."""
+    def test_convlstm_forward_features(self) -> None:
+        """Test the feature forward pass of the ConvLSTM model."""
         b = 1
         t = 4
         c = 3
@@ -28,7 +28,7 @@ class TestConvLSTM:
             num_layers=1,
             batch_first=True,
         )
-        layer_output_list, last_state_list = model(input_tensor)
+        layer_output_list, last_state_list = model.forward_features(input_tensor)
 
         assert len(layer_output_list) == 1
         assert len(last_state_list) == 1
@@ -53,7 +53,7 @@ class TestConvLSTM:
             batch_first=True,
             return_all_layers=True,
         )
-        layer_output_list, _ = model(input_tensor)
+        layer_output_list, _ = model.forward_features(input_tensor)
 
         assert len(layer_output_list) == num_layers
         assert layer_output_list[0].shape == (b, t, hidden_dims[0], h, w)
@@ -75,7 +75,7 @@ class TestConvLSTM:
             num_layers=1,
             batch_first=True,
         )
-        layer_output_list, last_state_list = model(input_tensor)
+        layer_output_list, last_state_list = model.forward_features(input_tensor)
 
         assert len(layer_output_list) == 1
         assert len(last_state_list) == 1
@@ -97,7 +97,7 @@ class TestConvLSTM:
             num_layers=1,
             batch_first=True,
         )
-        layer_output_list, last_state_list = model(input_tensor)
+        layer_output_list, last_state_list = model.forward_features(input_tensor)
 
         assert len(layer_output_list) == 1
         assert len(last_state_list) == 1
@@ -119,7 +119,7 @@ class TestConvLSTM:
             num_layers=1,
             batch_first=False,
         )
-        layer_output_list, last_state_list = model(input_tensor)
+        layer_output_list, last_state_list = model.forward_features(input_tensor)
 
         assert len(layer_output_list) == 1
         assert len(last_state_list) == 1
@@ -152,9 +152,71 @@ class TestConvLSTM:
             batch_first=True,
             return_all_layers=True,
         )
-        layer_output_list, last_state_list = model(input_tensor)
+        layer_output_list, last_state_list = model.forward_features(input_tensor)
 
         assert len(layer_output_list) == 2
         assert len(last_state_list) == 2
         assert layer_output_list[0].shape == (b, t, 16, h, w)
         assert layer_output_list[1].shape == (b, t, 32, h, w)
+
+    def test_convlstm_forward(self) -> None:
+        """Test segmentation forward pass with prediction head."""
+        b = 2
+        t = 4
+        c = 3
+        h = 16
+        w = 16
+        num_classes = 5
+        input_tensor = torch.rand(b, t, c, h, w)
+
+        model = ConvLSTM(
+            input_dim=c,
+            hidden_dim=16,
+            kernel_size=3,
+            num_layers=1,
+            batch_first=True,
+            num_classes=num_classes,
+            head_kernel_size=1,
+        )
+        y_hat = model(input_tensor, lengths=torch.tensor([4, 2]))
+
+        assert y_hat.shape == (b, num_classes, h, w)
+
+    def test_convlstm_forward_uses_last_timestep_without_lengths(self) -> None:
+        """Test segmentation forward pass defaults to the final timestep."""
+        b = 2
+        t = 4
+        c = 3
+        h = 16
+        w = 16
+        num_classes = 5
+        input_tensor = torch.rand(b, t, c, h, w)
+
+        model = ConvLSTM(
+            input_dim=c,
+            hidden_dim=16,
+            kernel_size=3,
+            num_layers=1,
+            batch_first=True,
+            num_classes=num_classes,
+            head_kernel_size=1,
+        )
+        y_hat = model(input_tensor)
+        y_hat_last_step = model(input_tensor, lengths=torch.tensor([t, t]))
+
+        torch.testing.assert_close(y_hat, y_hat_last_step)
+
+    def test_convlstm_forward_requires_head(self) -> None:
+        """Test that segmentation forward requires a configured prediction head."""
+        b = 1
+        t = 2
+        c = 3
+        h = 8
+        w = 8
+        input_tensor = torch.rand(b, t, c, h, w)
+        model = ConvLSTM(
+            input_dim=c, hidden_dim=8, kernel_size=3, num_layers=1, batch_first=True
+        )
+
+        with pytest.raises(ValueError, match='Segmentation head is not configured'):
+            model(input_tensor)
--- a/tests/models/test_convlstm.py
+++ b/tests/models/test_convlstm.py
@@ -21,13 +21,7 @@ class TestConvLSTM:
         w = 64
         input_tensor = torch.rand(b, t, c, h, w)
 
-        model = ConvLSTM(
-            input_dim=c,
-            hidden_dim=16,
-            kernel_size=(3, 3),
-            num_layers=1,
-            batch_first=True,
-        )
+        model = ConvLSTM(input_dim=c, hidden_dim=16, kernel_size=(3, 3), num_layers=1)
         layer_output_list, last_state_list = model.forward_features(input_tensor)
 
         assert len(layer_output_list) == 1
@@ -50,7 +44,6 @@ class TestConvLSTM:
             hidden_dim=hidden_dims,
             kernel_size=(3, 3),
             num_layers=num_layers,
-            batch_first=True,
             return_all_layers=True,
         )
         layer_output_list, _ = model.forward_features(input_tensor)
@@ -73,7 +66,6 @@ class TestConvLSTM:
             hidden_dim=16,
             kernel_size=3,  # Pass as integer
             num_layers=1,
-            batch_first=True,
         )
         layer_output_list, last_state_list = model.forward_features(input_tensor)
 
@@ -95,29 +87,6 @@ class TestConvLSTM:
             hidden_dim=16,
             kernel_size=[(3, 3)],  # Pass as list of tuples
             num_layers=1,
-            batch_first=True,
-        )
-        layer_output_list, last_state_list = model.forward_features(input_tensor)
-
-        assert len(layer_output_list) == 1
-        assert len(last_state_list) == 1
-        assert layer_output_list[0].shape == (b, t, 16, h, w)
-
-    def test_convlstm_batch_first_false(self) -> None:
-        """Test the forward pass with batch_first=False."""
-        b = 1
-        t = 4
-        c = 3
-        h = 64
-        w = 64
-        input_tensor = torch.rand(t, b, c, h, w)  # Note the different order
-
-        model = ConvLSTM(
-            input_dim=c,
-            hidden_dim=16,
-            kernel_size=(3, 3),
-            num_layers=1,
-            batch_first=False,
         )
         layer_output_list, last_state_list = model.forward_features(input_tensor)
 
@@ -149,7 +118,6 @@ class TestConvLSTM:
             hidden_dim=[16, 32],
             kernel_size=[3, (5, 5)],  # Mix of int and tuple
             num_layers=2,
-            batch_first=True,
             return_all_layers=True,
         )
         layer_output_list, last_state_list = model.forward_features(input_tensor)
@@ -174,7 +142,6 @@ class TestConvLSTM:
             hidden_dim=16,
             kernel_size=3,
             num_layers=1,
-            batch_first=True,
             num_classes=num_classes,
             head_kernel_size=1,
         )
@@ -197,7 +164,6 @@ class TestConvLSTM:
             hidden_dim=16,
             kernel_size=3,
             num_layers=1,
-            batch_first=True,
             num_classes=num_classes,
             head_kernel_size=1,
         )


torchgeo/trainers/segmentation.py
```diff
@@ -46,20 +100,27 @@ def __init__(
         patience: int = 10,
         freeze_backbone: bool = False,
         freeze_decoder: bool = False,
+        convlstm_hidden_dim: int | list[int] = 64,
```
robmarkcole: should these be in kwargs? Personally I find kwargs confusing to use. Or maybe we choose defaults and don't expose these

torchgeo/trainers/segmentation.py
```diff
@@ -315,18 +403,12 @@ def validation_step(
         ):
             datamodule = self.trainer.datamodule
             if batch['image'].ndim == 5:
-                _, T, C, _, _ = batch['image'].shape
-                batch['image'] = rearrange(batch['image'], 'b t c h w -> b (t c) h w')
```
robmarkcole, line 261: Why do this?

torchgeo/trainers/segmentation.py
```diff
@@ -46,20 +100,27 @@ def __init__(
...
                 <https://smp.readthedocs.io/en/latest/encoders_timm.html>`__ or `smp
                 <https://smp.readthedocs.io/en/latest/encoders.html>`__ backbone to use.
             weights: Initial model weights. Either a weight enum, the string
                 representation of a weight enum, True for ImageNet weights, False or
                 None for random weights, or the path to a saved model state dict. FCN
                 model does not support pretrained weights.
-            in_channels: Number of input channels to model.
+            in_channels: Number of input channels to model. For time series inputs
+                (B, T, C, H, W), set this to ``T * C`` for all models except
+                ``model='convlstm'`` where it should be ``C``.
```
robmarkcole: Relates to https://github.com/torchgeo/torchgeo/issues/3279

torchgeo/trainers/segmentation.py
```diff
@@ -46,20 +102,27 @@ def __init__(
         patience: int = 10,
         freeze_backbone: bool = False,
         freeze_decoder: bool = False,
+        convlstm_hidden_dim: int | list[int] = 64,
+        convlstm_kernel_size: int | tuple[int, int] | list[int | tuple[int, int]] = 3,
+        convlstm_num_layers: int = 1,
+        convlstm_head_kernel_size: int = 1,
```
robmarkcole: 1 or 3 is a good choice here?

torchgeo/trainers/segmentation.py
MarcCoru: I’m also not sure why the implementation merges the **time** and **channel** dimensions here.

With **ConvLSTMs**, the point is to keep the temporal axis *explicit* and apply convolutions **per time step**. Concretely, each 2D convolution inside the model should operate on inputs shaped like **(B, C, H, W)** at a single time step `t`, rather than on a tensor where time has been flattened into the channel dimension (i.e., `C' = T * C`).

Because of that, treating time as extra channels (via `T*C` concatenation) feels a bit mismatched in this context. I’d see this kind of concatenation more as a pragmatic “hack” to inject temporal information into a conventional, non-temporal CNN, rather than something you’d do when using an explicitly temporal model like a ConvLSTM.

torchgeo/trainers/segmentation.py
MarcCoru: on the ConvLSTM configuration:

- `convlstm_num_layers = 1`: This seems reasonable. It’s a bit shallow, but that also tends to make the model easier to train and more stable. A 2–3 layer ConvLSTM *might* yield a small performance gain, but the 1-layer choice is a solid default.

- `convlstm_head_kernel_size: int = 1`: Also reasonable. A 1×1 conv here is essentially a per-pixel dense/FC layer, so it avoids mixing neighboring pixels in the final projection. That can be beneficial for segmentation masks, since it reduces the tendency to produce overly “blobby” outputs due to spatial smoothing at the last stage.

- ConvLSTM `kernel_size = 3`: Makes sense and is consistent with common CNN practice (3×3 spatial context per update).

torchgeo/trainers/segmentation.py
isaaccorley: The segmentation trainer currently assumes everything is BCHW. This was added so that time-series segmentation datasets can be used with this trainer. My 2c is that we need a `MultitemporalSegmentation` trainer that can inherit from the `Segmentation` trainer but modify to natively support time-series instead of hacking at the existing Segmentation trainer.

robmarkcole: @adamjstewart WDYT? Separate `MultitemporalSegmentation` trainer supporting `time_steps` arg as discussed in  https://github.com/torchgeo/torchgeo/issues/3279#issuecomment-3764135143 ?

adamjstewart: Need to think about this. We combined all binary/multiclass/multilabel into one trainer because it was easy, but time series might be a little harder. There are tons of 1D, 2D, 3D, and 4D setups we need to support, and that's just for inputs. Outputs might also be 0D, 1D, 2D, 3D, or 4D. Each might be classification or regression as well. I don't necessarily want 4 x 5 x 2 trainers (not all would make sense anyway) but if it makes the code simpler and more maintainable, I could definitely be convinced. Other option is to use mixins like I did in #3328 so we have less code duplication. We need to make a similar decision for samplers too...

robmarkcole:
@adamjstewart to confirm, this will now be implemented as a new VideoSemanticSegmentation trainer?

adamjstewart:
That would be my vote, yes.

torchgeo/trainers/video_segmentation.py
```diff
@@ -0,0 +1,405 @@
...
+        Returns:
+            Output tensor of shape (B, num_classes, H, W).
+        """
+        layer_output_list, _ = self.backbone(x)
+        layer_output = layer_output_list[-1]
+
+        if lengths is None:
+            features = layer_output[:, -1]
+        else:
+            idx = (lengths.to(device=layer_output.device) - 1).clamp(min=0)
```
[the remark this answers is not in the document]
robmarkcole: addressed

robmarkcole:
@adamjstewart updated but there is now significant overlap with the semantic trainer. Going to strip this trainer back a bit

robmarkcole:
100 epoch run:
```
│                 test_OverallAccuracy                 │                  0.5166717767715454                  │
│                 test_OverallF1Score                  │                  0.5166717767715454                  │
│               test_OverallJaccardIndex               │                 0.34831923246383667                  │
│                test_OverallPrecision                 │                  0.5166717767715454                  │
│                  test_OverallRecall                  │                  0.5166717767715454                  │
│                      test_loss                       │                   1.4813312292099     
```

image

```yaml
seed_everything: 0

model:
  class_path: VideoSemanticSegmentationTask
  init_args:
    model: convlstm
    in_channels: 10
    task: multiclass
    num_classes: 20
    labels:
      - background
      - meadow
      - soft_winter_wheat
      - corn
      - winter_barley
      - winter_rapeseed
      - spring_barley
      - sunflower
      - grapevine
      - beet
      - winter_triticale
      - winter_durum_wheat
      - fruits_vegetables_flowers
      - potatoes
      - leguminous_fodder
      - soybeans
      - orchard
      - mixed_cereal
      - sorghum
      - void_label
    loss: ce

data:
  class_path: PASTIS100DataModule
  init_args:
    batch_size: 3
    num_workers: 0
    val_split_pct: 0.2
    test_split_pct: 0.2
    padding_length: 9
  dict_kwargs:
    root: /Users/robin.cole/data
    download: true

trainer:
  accelerator: auto
  devices: 1
  max_epochs: 100
  log_every_n_steps: 1
  default_root_dir: lightning_logs/pastis100_video_segmentation
  callbacks:
    - class_path: lightning.pytorch.callbacks.ModelCheckpoint
      init_args:
        monitor: val_loss
        mode: min
        save_last: true
        save_top_k: 1
  logger:
    class_path: lightning.pytorch.loggers.WandbLogger
    init_args:
      project: torchgeo-pastis100
      name: pastis100-video-segmentation
      save_dir: lightning_logs
      log_model: false
```

Script:
```python
#!/usr/bin/env python3

"""Run config-driven train and test commands for PASTIS100 video segmentation."""

# Example:
# uv run python scripts/run_pastis100_video_segmentation.py

from __future__ import annotations

import argparse
import subprocess
from datetime import datetime
from pathlib import Path


def parse_args() -> argparse.Namespace:
    """Parse command-line arguments."""
    parser = argparse.ArgumentParser(
        description='Run torchgeo fit and test with a shared PASTIS100 video segmentation config.'
    )
    parser.add_argument(
        '--config',
        type=Path,
        default=Path('scripts/configs/pastis100_video_segmentation.yaml'),
        help='Path to the shared config.',
    )
    parser.add_argument(
        '--ckpt-path',
        type=str,
        default='last',
        help='Checkpoint path passed to `torchgeo test`.',
    )
    return parser.parse_args()


def resolve_base_root_dir(config_path: Path) -> Path:
    """Extract the base ``trainer.default_root_dir`` from the config file."""
    for line in config_path.read_text().splitlines():
        stripped = line.strip()
        if stripped.startswith('default_root_dir:'):
            _, value = stripped.split(':', maxsplit=1)
            return Path(value.strip())

    msg = f'Could not find trainer.default_root_dir in {config_path}'
    raise ValueError(msg)


def checkpoint_search_root(config_path: Path) -> Path:
    """Return the directory tree to search for checkpoints."""
    return resolve_base_root_dir(config_path).parent


def create_run_root_dir(config_path: Path) -> Path:
    """Create a unique run directory under the configured base root."""
    base_root_dir = resolve_base_root_dir(config_path)
    timestamp = datetime.now().strftime('%Y%m%d_%H%M%S')
    run_root_dir = base_root_dir / timestamp
    run_root_dir.mkdir(parents=True, exist_ok=False)
    return run_root_dir


def checkpoint_candidates(search_root: Path) -> list[Path]:
    """Return all candidate ``last`` checkpoints under the search root."""
    return sorted(search_root.rglob('last*.ckpt'))


def checkpoint_mtimes(paths: list[Path]) -> dict[Path, float]:
    """Map checkpoints to their modification times."""
    return {path: path.stat().st_mtime for path in paths}


def resolve_checkpoint_path(
    search_root: Path,
    ckpt_path: str,
    previous_mtimes: dict[Path, float] | None = None,
) -> str:
    """Resolve the checkpoint path to an explicit file for testing."""
    if ckpt_path != 'last':
        return ckpt_path

    candidates = checkpoint_candidates(search_root)
    if previous_mtimes is not None:
        updated = [
            path
            for path in candidates
            if path not in previous_mtimes
            or path.stat().st_mtime > previous_mtimes[path]
        ]
        if updated:
            return str(max(updated, key=lambda path: path.stat().st_mtime))

    if candidates:
        return str(max(candidates, key=lambda path: path.stat().st_mtime))

    msg = f'No last checkpoint found under {search_root}'
    raise FileNotFoundError(msg)


def main() -> None:
    """Run fit followed by test using TorchGeo CLI configs."""
    args = parse_args()

    if not args.config.is_file():
        msg = f'Config not found: {args.config}'
        raise FileNotFoundError(msg)

    run_root_dir = create_run_root_dir(args.config)
    search_root = checkpoint_search_root(args.config)
    previous_mtimes = checkpoint_mtimes(checkpoint_candidates(search_root))

    fit_cmd = [
        'uv',
        'run',
        'torchgeo',
        'fit',
        '--config',
        str(args.config),
        '--trainer.default_root_dir',
        str(run_root_dir),
    ]
    test_cmd = [
        'uv',
        'run',
        'torchgeo',
        'test',
        '--config',
        str(args.config),
        '--trainer.default_root_dir',
        str(run_root_dir),
        '--ckpt_path',
        args.ckpt_path,
    ]

    print('Run directory:', run_root_dir)
    print('Running:', ' '.join(fit_cmd))
    subprocess.run(fit_cmd, check=True)

    resolved_ckpt_path = resolve_checkpoint_path(
        search_root, args.ckpt_path, previous_mtimes
    )
    test_cmd[-1] = resolved_ckpt_path

    print('Running:', ' '.join(test_cmd))
    subprocess.run(test_cmd, check=True)


if __name__ == '__main__':
    main()
```

adamjstewart (changes requested):
Can we split the two trainers into separate PRs? We'll also want datamodules to test them with real data. See the Google Sheets doc I shared with you for suggestions.

robmarkcole:
@adamjstewart for VideoPixelwiseRegressionTask  we could use CopernicusBenchBiomassS3 which I recall was my motivation behind https://github.com/torchgeo/torchgeo/pull/3086 so I suggest we merge that first?

robmarkcole:
TODO:

1. confirm if _ConvLSTMSegmentation is required or not
2. Subclass SemanticSegmentationTask?
3. Viz some predictions (sanity check)
4. Train on PASTIS full dataset

On 1. `_ConvLSTMSegmentation` -> Yes, in practice it is necessary. The ConvLSTM produces hidden feature maps, but the segmentation task needs per-pixel class logits, so you still need a final projection from hidden channels to num_classes. That projection does not have to be exactly the current 1x1 convolution, but some segmentation head is needed unless you force the ConvLSTM hidden size to equal the number of classes, which would unnecessarily couple the feature representation to the label space.

On 2. SemanticSegmentationTask is built around 2D segmentation models from SMP plus the special flattened-time-series path, while VideoSemanticSegmentationTask is built around a ConvLSTM that consumes (B, T, C, H, W) directly and uses length to select the last valid timestep. Those differences sit in the core model construction and forward path, so subclassing SemanticSegmentationTask would mostly mean overriding the important parts and inheriting a bunch of irrelevant SMP-oriented surface area like backbone, weights, and freeze options. The current structure is already using the right shared piece, ClassificationMixin; if we want further simplification, the better move would be extracting small shared classification-step helpers, not making the video trainer inherit from the 2D segmentation trainer.

## Example small shared classification-step helpers
```python
def _compute_classification_loss(
    self, y_hat: Tensor, y: Tensor, loss_name: str
) -> Tensor:
    if loss_name == 'bce':
        y = y.float()
    return self.criterion(y_hat, y)


def _update_and_log_classification_metrics(
    self, stage: str, y_hat: Tensor, y: Tensor, batch_size: int
) -> None:
    metrics = getattr(self, f'{stage}_metrics')
    metrics(y_hat, y)
    self.log_dict(metrics, batch_size=batch_size)

def _log_classification_loss(
    self, stage: str, loss: Tensor, batch_size: int
) -> None:
    self.log(f'{stage}_loss', loss, batch_size=batch_size)


def _classification_probs(self, y_hat: Tensor) -> Tensor:
    match self.hparams['task']:
        case 'binary' | 'multilabel':
            return y_hat.sigmoid()
        case 'multiclass':
            return y_hat.softmax(dim=1)
```

@adamjstewart @yichiac thoughts on 2?

robmarkcole:
Some predictions from a 100 epoch run on pastis-100
```
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│                 test_AverageAccuracy                 │                 0.15500332415103912                  │
│                 test_AverageF1Score                  │                 0.13013575971126556                  │
│               test_AverageJaccardIndex               │                 0.08765462040901184                  │
│                test_AveragePrecision                 │                 0.12002260237932205                  │
│                  test_AverageRecall                  │                 0.15500332415103912        
```

prediction_004
prediction_005
prediction_006
prediction_007

robmarkcole:
Metrics for 50 epoch run on full PASTIS:
```
                     Test metric                      ┃                     DataLoader 0                     ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│                 test_AverageAccuracy                 │                 0.20961925387382507                  │
│                 test_AverageF1Score                  │                 0.20383641123771667                  │
│               test_AverageJaccardIndex               │                 0.14001913368701935                  │
│                test_AveragePrecision                 │                 0.32717961072921753                  │
│                  test_AverageRecall                  │                 0.20961925387382507                  │
```

prediction_000
prediction_001
prediction_002

adamjstewart:
> @adamjstewart @yichiac thoughts on 2?

Let's not worry too much about it, I'll refactor things with subclasses or mixins later.

tests/datasets/test_pastis.py
```diff
@@ -117,3 +119,27 @@ def test_plot(self, dataset: PASTIS) -> None:
             x['prediction_labels'] = x['label'].clone()
         dataset.plot(x)
         plt.close()
+
+    def test_plot_quantile_normalizes_rgb(
```
adamjstewart: What's the purpose of this test?

tests/conf/pastis.yaml
adamjstewart, line 1: I would personally be fine with replacing the existing `pastis.yaml` with this but then I think we'll have a missing line of code tested in SemanticSegmentationTask. We'll have to think about whether we want to keep time-series support in that trainer or remove it. Would be nice to deprecate it first...

tests/trainers/test_segmentation.py
```diff
@@ -332,3 +332,148 @@ def test_predict_step_preserves_metadata(self) -> None:
         assert result['transform'] is not None
         assert torch.equal(result['bounds'], bounds_tensor)
         assert torch.equal(result['transform'], transform_tensor)
+
+
+class TestSpatioTemporalSegmentationTask:
```
adamjstewart: Let's put this in a separate file just like the class it's testing.

tests/trainers/test_segmentation.py
```diff
@@ -332,3 +332,148 @@ def test_predict_step_preserves_metadata(self) -> None:
...
+        try:
+            main(['test', *args])
+        except MisconfigurationException:
+            pass
+        try:
+            main(['predict', *args])
+        except MisconfigurationException:
+            pass
+
+    @staticmethod
```
adamjstewart: This sounds like it should be a fixture instead

tests/trainers/test_segmentation.py
```diff
@@ -332,3 +332,148 @@ def test_predict_step_preserves_metadata(self) -> None:
...
+    def _create_video_model(**kwargs: Any) -> SpatioTemporalSegmentationTask:
+        model = SpatioTemporalSegmentationTask(
+            convlstm_hidden_dim=8, convlstm_num_layers=1, **kwargs
+        )
+        # Avoid Lightning warnings when calling step hooks without a Trainer.
+        model.log = lambda *args, **kwargs: None  # type: ignore[method-assign]
+        model.log_dict = lambda *args, **kwargs: None  # type: ignore[method-assign]
+        return model
+
+    def test_video_forward_defaults_to_convlstm(self) -> None:
```
adamjstewart: More tests is usually good, but someday I want to go through all of our trainers and try to minimize the tests to only get coverage. Some of these tests can be quite slow.

torchgeo/datasets/pastis.py
```diff
@@ -355,8 +362,15 @@ def plot(
...
-        # Keep the RGB bands and convert to T x H x W x C format
-        images = sample['image'][:, [2, 1, 0], :, :].numpy().transpose(0, 2, 3, 1)
+        # Keep the RGB bands and quantile-normalize the displayed frames.
+        rgb_frames = sample['image'][:, [2, 1, 0], :, :]
+        images = torch.stack(
+            [
+                quantile_normalization(rgb_frames[0]),
+                quantile_normalization(rgb_frames[1]),
+            ]
+        )
```
adamjstewart:
```suggestion
        images = quantile_normalization(rgb_frames)
```
Simpler and keeps relative brightness differences when plotting.

torchgeo/trainers/regression.py
```diff
@@ -138,7 +138,7 @@ def configure_metrics(self) -> None:
         """
         metrics = MetricCollection(
             # https://github.com/astral-sh/ty/issues/2985
-            {  # type: ignore[invalid-argument-type]
+            {
```
adamjstewart, line 141: I think I already made this change in main, could remove this change.

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,220 @@
...
+import torch
+import torch.nn as nn
+from torch import Tensor
+
+from ..models import ConvLSTM
+from .base import BaseTask
+from .mixins import ClassificationMixin
+
+
+class _ConvLSTMSegmentation(nn.Module):
```
adamjstewart: Wondering if we should just add this directly to `ConvLSTM` instead of keeping it here. Any reason not too?

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,220 @@
...
+
+        return cast(Tensor, self.head(features))
+
+
+class SpatioTemporalSegmentationTask(ClassificationMixin, BaseTask):
+    """Spatiotemporal Semantic Segmentation."""
+
+    def __init__(
+        self,
+        model: Literal['convlstm'] | str = 'convlstm',
```
adamjstewart:
```suggestion
        model: Literal['convlstm'] = 'convlstm',
```

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,220 @@
...
+        pos_weight: Tensor | None = None,
+        loss: Literal['ce', 'bce', 'jaccard', 'focal', 'dice'] = 'ce',
+        class_weights: Tensor | Sequence[float] | None = None,
+        ignore_index: int | None = None,
+        lr: float = 1e-3,
+        patience: int = 10,
+        convlstm_hidden_dim: int | list[int] = 64,
+        convlstm_kernel_size: int | tuple[int, int] | list[int | tuple[int, int]] = 3,
+        convlstm_num_layers: int = 1,
+        convlstm_head_kernel_size: int = 1,
```
adamjstewart: Eventually we will have many more models. I wonder if we shouldn't just wrap all of this into `**kwargs`.

.gitignore
```diff
@@ -148,3 +148,4 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+.DS_Store
```
adamjstewart: I don't normally add OS-specific ignores to a project `.gitignore`, I keep them in my global `.gitignore`.

adamjstewart (changes requested):

tests/datasets/test_pastis.py
robmarkcole: The purpose is to verify that plot applies quantile normalization/clipping before visualization - will remove as excessive

tests/conf/pastis.yaml
robmarkcole: IMO since we are not 1.0, ripping off the band aid is acceptable :-)

tests/trainers/test_segmentation.py
robmarkcole: Moved TestSpatioTemporalSegmentationTask into its own file

tests/trainers/test_segmentation.py
robmarkcole: Replaced staticmethod def _create_video_model(...) with a fixture factory:

tests/trainers/test_segmentation.py
robmarkcole: Removed the forward pass to speed up

torchgeo/trainers/regression.py
robmarkcole: Confirmed

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: _ConvLSTMSegmentation is removed, and its functionality is now integrated into ConvLSTM

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: Personally I loath kwargs, and have had meany headaches debugging them. here is what codex has to say:

I wouldn’t use raw **kwargs here. It would hurt this repo’s strengths:

- Worse typing (ty) and docs: explicit params in SpatioTemporalSegmentationTask.__init__ are discoverable and type-checked.
- Worse config UX: your YAML init_args (like tests/conf/pastis_convlstm.yaml) becomes opaque and easier to misconfigure.
- Harder validation: unknown keys silently pass unless you add custom checks.

Better approach when more models arrive:

- Keep shared explicit args.
- Add a single model_kwargs (or per-model config object) and validate keys per selected model in configure_models.
- Keep typed model-specific config structures rather than untyped **kwargs.

@adamjstewart WDYT?

adamjstewart: I also loathe kwargs, but only when it's used incorrectly. I'm not proposing it for all functions/parameters, only when function parameters are immediately passed to another documented function. This way, regardless of how many models we support, how many models we add, and whether or not these models add new parameters, the trainer will always be up-to-date with no required API changes.

* Worse typing (ty) and docs: agreed about typing, but we would document kwargs and make it clear that they are passed to the model, so see the model docs for details
* Worse config UX: `init_args` is just combined with `dict_kwargs`. I don't see it as more opaque, although I know LightningCLI will skip type verification, and maybe IDEs won't auto-suggest supported values
* Harder validation: although there are no static checks, there are run-time checks. Unknown or mispelled keys would result in the model `__init__` crashing, so users would have immediate feedback

I agree that kwargs sucks when it acts like a catch-all and invalid keys are simply ignored, but that wouldn't be the case here. Especially for other tasks with dozens (SMP) or hundreds (timm) of valid model architectures, it doesn't seem feasible to manually forward all possible parameters in the trainer.

torchgeo/models/convlstm.py
```diff
@@ -171,13 +177,22 @@ def __init__(
...
+            )
 
-    def forward(
+    def forward_features(
         self,
         input_tensor: torch.Tensor,
         hidden_state: list[tuple[torch.Tensor, torch.Tensor]] | None = None,
     ) -> tuple[list[torch.Tensor], list[tuple[torch.Tensor, torch.Tensor]]]:
-        """Forward pass of the ConvLSTM.
+        """Forward pass of ConvLSTM feature extraction.
```
adamjstewart, line 189: Can we get a `.. versionadded::` or `.. versionchanged::` here?

torchgeo/models/convlstm.py
```diff
@@ -132,6 +134,9 @@ def __init__(
             bias: If ``True``, adds a learnable bias to the output.
             return_all_layers: If ``True``, will return the list of computations
                 for all layers.
+            num_classes: Optional number of segmentation classes for an attached
+                prediction head.
+            head_kernel_size: Kernel size for the optional segmentation head.
```
adamjstewart, line 136: `.. versionadded::` for new params

torchgeo/datasets/pastis.py
adamjstewart: Missing this one

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,163 @@
...
+
+from torch import Tensor
+
+from ..models import ConvLSTM
+from .base import BaseTask
+from .mixins import ClassificationMixin
+
+
+class SpatioTemporalSegmentationTask(ClassificationMixin, BaseTask):
+    """Spatiotemporal Semantic Segmentation."""
```
adamjstewart:
```rst
.. versionadded:: 0.10
```

adamjstewart:
@MUYang99 can you review the changes to ConvLSTM if you have time?

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: Thanks for the explainer, moved to kwargs

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: added

torchgeo/models/convlstm.py
robmarkcole: added versionadded:: 0.10

torchgeo/models/convlstm.py
robmarkcole: added 
```
        .. versionadded:: 0.10
           The *num_classes* and *head_kernel_size* parameters.
```

tests/conf/pastis.yaml
robmarkcole: SemanticSegmentationTask no longer supports the 5D flatten hack.

- It now requires (B, C, H, W) and raises a ValueError for other ranks: segmentation.py (line 113)
- Removed the 0.8 “Time series … support” claim from the docstring:
- Removed the 5D-specific rearrange logic in validation plotting:
- Added a test that 5D input is rejected:
- Removed pastis100 from semantic-segmentation config sweep:
- Removed obsolete tests/conf/pastis100.yaml (it depended on the old flatten behavior).

robmarkcole:
@adamjstewart @piperwolters olmo added 
- currently does not implement `rslearn.models.olmoearth_pretrain.norm.OlmoEarthNormalize` which we should consider
- currently we don't pass timestamps to the model but these are supported?

Confirmed the model can be trained
image

100 epoch run on pastis-100 Nano model - [worse than convLSTM](https://github.com/torchgeo/torchgeo/pull/3280#issuecomment-4128843467) and very 'blobby' predictions
```
│                 test_AverageAccuracy                 │                 0.12502354383468628                  │
│                 test_AverageF1Score                  │                 0.10861381888389587                  │
```

The large model is slightly worse still
```
│                 test_AverageAccuracy                 │                 0.10425371676683426                  │
│                 test_AverageF1Score                  │                 0.08186521381139755                  │
```

And the tiny:
```
│                 test_AverageAccuracy                 │                 0.10522707551717758                  │
│                 test_AverageF1Score                  │                 0.08800695091485977                  │
```
Tiny predictions:
image
image
image

torchgeo/models/olmoearth_temporal_segmentation.py
```diff
@@ -0,0 +1,193 @@
...
+from typing import Any
+
+import segmentation_models_pytorch as smp
+from segmentation_models_pytorch.base.model import SegmentationModel
+from segmentation_models_pytorch.decoders.unet.decoder import UnetDecoder
+from torch import Tensor, nn
+from torch.nn import functional as F
+
+
+class OlmoEarthTemporalSegmentation(SegmentationModel):
```
robmarkcole: Perhaps `OlmoEarthV1Segmentation` instead?

robmarkcole:
50 epoch run, olmo nano on full PASTIS - similar metrics to the PASTIS100, may need more epochs..?

image

```
test_AverageAccuracy:          0.12158840149641036
test_AverageF1Score:            0.10757429897785188
```

image
image
image

@adamjstewart you think it is worth doing more experiments, or proceed as-is?

torchgeo/models/olmoearth_temporal_segmentation.py
adamjstewart, line 1: What is this? Just an OlmoEarth encoder with a U-Net decoder? Can we do this in 3-lines of code in SpatioTemporalSegmentationTask? That would save us 700+ lines of code we don't need to maintain...

torchgeo/trainers/segmentation.py
```diff
@@ -115,13 +114,19 @@ def forward(self, x: Tensor) -> Tensor:
...
         Returns:
             Output tensor of shape (B, num_classes, H, W).
+
+        Raises:
+            ValueError: If ``x`` is not 4-dimensional.
         """
-        if x.ndim == 5:
-            x = rearrange(x, 'b t c h w -> b (t c) h w')
+        if x.ndim != 4:
+            raise ValueError(
```
adamjstewart: Would be nice if we could deprecate this first instead of removing immediately since we just added this feature in the last release.

torchgeo/datasets/pastis.py
```diff
@@ -355,8 +362,10 @@ def plot(
         Returns:
             a matplotlib Figure with the rendered sample
         """
-        # Keep the RGB bands and convert to T x H x W x C format
-        images = sample['image'][:, [2, 1, 0], :, :].numpy().transpose(0, 2, 3, 1)
+        # Keep the RGB bands and quantile-normalize the displayed frames.
+        rgb_frames = sample['image'][:, [2, 1, 0], :, :]
+        images = quantile_normalization(rgb_frames)
+        images = images.numpy().transpose(0, 2, 3, 1)
```
adamjstewart: Not sure why ty didn't catch this, but this is a dynamic type change. `images` was torch.Tensor and is now np.NDArray.

```suggestion
        rgb_frames = quantile_normalization(rgb_frames)
        images = rgb_frames.numpy().transpose(0, 2, 3, 1)
```
Alternatively, keep everything as a torch.Tensor, but that's more lines of code to change and review.

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,223 @@
...
+
+class SpatioTemporalSegmentationTask(ClassificationMixin, BaseTask):
+    """Spatiotemporal Semantic Segmentation.
+
+    .. versionadded:: 0.10
+    """
+
+    def __init__(
+        self,
+        model: Literal['convlstm', 'olmoearth_v1'] | str = 'convlstm',
```
adamjstewart:
```suggestion
        model: Literal['convlstm', 'olmoearth_v1'] = 'convlstm',
```
Other strings aren't valid.

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,223 @@
...
+            lr: Learning rate for optimizer.
+            patience: Patience for learning rate scheduler.
+            **kwargs: Additional model-specific kwargs. For ``model='convlstm'``,
+                kwargs are passed to ``ConvLSTM``. For ``model='olmoearth_v1'``,
+                kwargs are passed to ``olmoearth_v1`` except
+                ``decoder_channels``, ``decoder_in_channels``, and
+                ``head_kernel_size``, which configure the segmentation decoder.
+
+        .. versionchanged:: 0.10
+           Added support for ``model='olmoearth_v1'``.
```
adamjstewart: Not needed since this didn't exist before then anyway

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,223 @@
...
+from .mixins import ClassificationMixin
+
+
+class SpatioTemporalSegmentationTask(ClassificationMixin, BaseTask):
+    """Spatiotemporal Semantic Segmentation.
+
+    .. versionadded:: 0.10
+    """
+
+    def __init__(
```
adamjstewart: How hard would it be to add support for pretrained weights like our other trainers? This would be very nice for Olmo for example.

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,223 @@
...
+        """Forward pass of the model.
+
+        Args:
+            x: Input tensor of shape (B, T, C, H, W).
+            lengths: Optional sequence lengths (B,) before padding/truncation.
+
+        Returns:
+            Output tensor of shape (B, num_classes, H, W).
+        """
+        model = cast(ConvLSTM | OlmoEarthTemporalSegmentation, self.model)
```
adamjstewart: Not sure why this is needed.

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,223 @@
...
+                backbone = olmoearth_v1(**olmoearth_kwargs)
+                self.model = OlmoEarthTemporalSegmentation(
+                    backbone=backbone,
+                    num_classes=num_classes,
+                    decoder_channels=decoder_channels,
+                    decoder_in_channels=decoder_in_channels,
+                    head_kernel_size=head_kernel_size,
+                    patch_size=patch_size,
+                )
+            case _:
```
adamjstewart: Optional, I'm always torn between having nice run-time error msgs for users and less code/testing/ty ignores for developers.

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,223 @@
...
+                raise ValueError(
+                    f"Model type '{model}' is not supported. "
+                    'Currently, SpatioTemporalSegmentationTask only supports '
+                    "'convlstm' and 'olmoearth_v1'."
+                )
+
+    def _get_convlstm_kwargs(self) -> dict[str, Any]:
```
adamjstewart: Would like to replace this whole function with a 1-liner like below where we need it:
```python
kwargs = self.hparams['kwargs']
kwargs |= {
    'hidden_dim': 64,
    'kernel_size': 3,
    ...
}
```

adamjstewart (commented):
Overall good, but would like to know if we can greatly simplify combining OlmoEarth with U-Net and reduce this PR from 1.5k lines to ~500 lines. Every line of code we add is something I have to maintain 🙂

torchgeo/trainers/segmentation.py
robmarkcole: done

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: removed

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: Removed

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: removed

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: removed the helper

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: SpatioTemporalSegmentationTask now accepts weights: OlmoEarthV1_Weights | str | None

torchgeo/models/olmoearth_temporal_segmentation.py
robmarkcole: Simplified the file so it is now OlmoEarth-only instead of trying to support arbitrary backbone output shapes and fallback code paths. Removed the generic tensor/dict/list/token handling and now always go through the actual OlmoEarth encoder contract, including the Sentinel-2 channel adaptation and conversion from `tokens_and_masks` into a spatial feature map. This file is still required because the trainer still needs a real model wrapper that turns OlmoEarth encoder outputs into segmentation logits with a decoder, handles the input/output reshaping, and encapsulates the Olmo-specific glue instead of dumping that logic into the trainer.

robmarkcole:
@adamjstewart comments addressed, stripped this down whilst adding support for pretrained weights as requested

torchgeo/datasets/pastis.py
```diff
@@ -355,8 +362,10 @@ def plot(
         Returns:
             a matplotlib Figure with the rendered sample
         """
-        # Keep the RGB bands and convert to T x H x W x C format
-        images = sample['image'][:, [2, 1, 0], :, :].numpy().transpose(0, 2, 3, 1)
+        # Keep the RGB bands and quantile-normalize the displayed frames.
+        rgb_frames = sample['image'][:, [2, 1, 0], :, :]
+        rgb_frames = quantile_normalization(rgb_frames)
```
favyen2, line 399: I'm not sure about dividing by 5000, since I usually visualize B04/B03/B02 together with a range of [0, 2500] (I usually use 2500 not 5000), but I always have issues with quantile normalization because there could be cloudy or snowy or ocean images that end up looking very strange. For PASTIS it may be okay, but [0, 2500] seems to always give consistently good images. It is different from Landsat where a fixed normalization doesn't seem to work across all Landsat images.

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,257 @@
...
+    OlmoEarthV1_Weights,
+    get_weight,
+    olmoearth_v1,
+)
+from .base import BaseTask
+from .mixins import ClassificationMixin
+
+
+class SpatioTemporalSegmentationTask(ClassificationMixin, BaseTask):
+    """Spatiotemporal Semantic Segmentation.
```
favyen2: Is the distinction between SemanticSegmentationTask and SpatioTemporalSegmentationTask that SemanticSegmentationTask can only input one image while SpatioTemporalSegmentationTask can input an image time series? Or is the output for SpatioTemporalSegmentationTask also an image time series (different labels for each timestep)? I wonder if the name could make this more clear, like SemanticSegmentationTaskWithMultitemporalInput or something along those lines. I feel like the "spatio" may be redundant since semantic segmentation generally implies per-pixel classification.

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,257 @@
...
+                    patch_size=patch_size,
+                )
+
+    def _get_olmoearth_kwargs(self) -> dict[str, Any]:
+        """Build OlmoEarth kwargs."""
+        olmoearth_kwargs: dict[str, Any] = {}
+        if 'kwargs' in self.hparams:
+            olmoearth_kwargs.update(cast(dict[str, Any], self.hparams['kwargs']))
+
+        for key in (
```
favyen2: Might make sense to only have one route to configure the model instead of being able to set things in either e.g. `hparams["kwargs"]["model_size"]` OR `hparams["model_size"]`.

torchgeo/models/olmoearth_temporal_segmentation.py
```diff
@@ -0,0 +1,206 @@
...
+                        (batch_size, height, width, timesteps, 2),
+                        device=sentinel2_l2a.device,
+                        dtype=sentinel2_l2a.dtype,
+                    ),
+                ],
+                dim=-1,
+            )
+            if not self._warned_input_padding:
+                warnings.warn(
+                    'OlmoEarth expects 12 Sentinel-2 L2A bands. Received 10 channels; '
```
favyen2: In our evaluations we pad with the closest band. So B01 is filled with B02, and B09 is filled with B8A.

https://raw.githubusercontent.com/allenai/rslearn_projects/refs/heads/master/data/olmoearth_evals/tasks/pastis_ts.yaml

It is a bit awkward to handle this PASTIS special case in this module, but I'm not sure what the convention is for handling these discrepancies in torchgeo.

torchgeo/models/olmoearth_temporal_segmentation.py
```diff
@@ -0,0 +1,206 @@
...
+    def _build_feature_pyramid(self, features: Tensor) -> list[Tensor]:
+        """Build a coarse-to-fine feature pyramid for SMP decoders."""
+        pyramid = [features]
+        for _ in range(self.pyramid_depth):
+            last = pyramid[-1]
+            if last.shape[-2:] == (1, 1):
+                pyramid.append(last)
+            else:
+                pyramid.append(
+                    F.avg_pool2d(last, kernel_size=2, stride=2, ceil_mode=True)
```
favyen2: I think it ends up having more code to have compatibility with the SMP decoder than could be done just be directly making the prediction.

If E is the encoder output channels and C is the number of classes, consider just doing `(H/patch_size W/patch_size E) -> [1x1 Conv2d] -> (H/patch_size W/patch_size C*patch_size*patch_size) -> [Reshape] -> (H, W, C)`. Just need to be careful about the reshaping to be done with the right axes (best to use einops).

torchgeo/datasets/pastis.py
robmarkcole: There are some cloudy images, but for those we can't see much anyway. I found the previous 5000 norm wasn't working well and quantile_normalization gave a better result but I only spot checked. @adamjstewart not sure if you want to go with the suggestion above?

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: The discussion around naming is ongoing, but complex.. Actually SemanticSegmentationTask also supports time series but via stacking - this is deprecated however

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: Raised this previously, @adamjstewart and comments on the current implementation here?

torchgeo/models/olmoearth_temporal_segmentation.py
robmarkcole: @adamjstewart patch this in the PASTIS dataset?

torchgeo/models/olmoearth_temporal_segmentation.py
robmarkcole: Implemented that suggestion

torchgeo/datasets/pastis.py
adamjstewart: I'm generally pro `quantile_normalization` because it's what QGIS uses. It also works regardless of normalization which is nice. @calebrob6 is generally pro stable division because of the above mentioned points. I think it doesn't matter too much as long as the visualizations usually look good.

P.S. I'm still looking for a student to lead this effort, but what I really want is a `PlottingMixin` class that can have a good default `plot()` method used by 95% of classes. Someday we'll start removing all of this code duplication and we can decide which of the two strategies we want.

torchgeo/trainers/spatiotemporal_segmentation.py
adamjstewart: @favyen2 I did a study on all existing benchmark datasets in TorchGeo: [spreadsheet](https://docs.google.com/spreadsheets/d/11rENiUywoGp93FSyEaYYIqTd12CCRNDEaWrfqbSZBlI/edit?usp=sharing). 1st tab is the input/output dimensions of all datasets, 2nd tab is a pivot table grouping and counting all combinations and trying to assign names to them.

We spent a while back-and-forth trying to come up with good names. We decided to use `Temporal` for 1D inputs, keep 2D as is instead of adding `Spatial`, use `SpatioTemporal` for 3D inputs, etc. Backup naming was going to be `Sequence` for 1D and `Video` for 3D, but this may be more confusing. Luckily there are currently no datasets with time series output. If there were, this naming situation would become much more complicated.

But if you have better ideas, nothing is finalized just yet. I'll eventually reformat the API docs to make these input/output dimensions more clear so users can more easily choose the right trainer for them.

torchgeo/trainers/spatiotemporal_segmentation.py
adamjstewart: I agree with having one route. The definition should use `**kwargs`, and users should use `model_size=...` instead of `kwargs={'model_size': ...}`. However `save_hyperparameters` handles these kwargs is how they should be accessed, I'm hoping it's `hparams['model_size']` because that's less typing for everyone.

torchgeo/models/olmoearth_temporal_segmentation.py
adamjstewart: If we do put this in the dataset, I would suggest a `bands` argument that allows you to choose a subset (or superset). This could for example be `bands=['B02', 'B02', 'B03', ..., 'B08', 'B8A', 'B8A', B10', ...]` in the `config.yaml` so that the dataset isn't ugly and the model isn't either. We already support this for many other datasets, although I'm not sure if anyone realizes they can use it to _duplicate_ bands, not just _remove_ bands.

Alternatively, I would be happy to do things how timm does where weights bands are either duplicated or removed dynamically with appropriate scaling added. This would make OlmoEarth much more flexible for variable inputs.

adamjstewart:
Summarizing discussions over Slack:

* Robin: is trying to add an SMP U-Net decoder to OlmoEarth
* Adam: doesn't love having an extra 700+ lines just to add a segmentation head to OlmoEarth
* Adam: notes that OlmoEarth may have its own decoder, at least this is mentioned in the paper
* Favyen: never tried using this MAE decoder for downstream tasks, wouldn't recommend
* Favyen: instead suggests a very simple U-Net without skip connections
* Adam: is fine with doing this in `torchgeo/trainers/spatiotemporal_segmentation.py` or in `torchgeo/models/olmoearth.py`

Basically, my goal is 5–10 lines of code, not 700+. Specific suggestions from Favyen:

* linear probing: (H/patch_size W/patch_size 768) -> [1x1 Conv2d] -> (H/patch_size W/patch_size C*patch_size*patch_size) -> [Reshape] -> (H, W, C) where C is the number of classes
* fine-tuning with patch_size=4: (H/4 W/4 768) -> [3x3 Conv2d] -> (H/4 W/4 512) -> Upsample -> (H/2 W/2 512) -> [3x3 Conv2d] -> (H/2 W/2 256) -> (H/2 W/2 256) -> [Upsample] -> (H W 256) -> [3x3 Conv2d] -> (H W 128) -> [3x3 Conv2d] -> (H W C)

I guess I would prefer to choose only one of these to make things simpler.

torchgeo/datasets/pastis.py
robmarkcole: keep quantile_normalization

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: No action

torchgeo/models/olmoearth_temporal_segmentation.py
robmarkcole: Moved the PASTIS-specific 10 to 12 channel workaround out of the model and into dataset/config selection

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: Changed to use a single configuration route. It no longer reads nested hparams['kwargs']; both the ConvLSTM and OlmoEarth paths now pull model-specific settings directly from self.hparams, e.g. self.hparams['model_size'] and self.hparams['patch_size'].

robmarkcole:
@adamjstewart @favyen2 implemented both:

- `decoder_type='linear'` keeps the patch-wise linear probe head.
- `decoder_type='simple_unet'` adds the no-skip decoder and currently requires patch_size=4 plus three decoder_channels.

If I had to choose one, I guess I would keep the simple_unet, but I think I will want to experiment with both.

torchgeo/models/olmoearth_temporal_segmentation.py
yichiac, line 1: I'm thinking whether we should separate OlmoEarth support into a separate PR. It seems the OlmoEarth decoder and its tests are making the PR too large. We can split this into two PRs. This PR could focus on the core spatiotemporal segmentation, including convlstm + pastis. The second PR could add support for OlmoEarth-specific segmentation.

robmarkcole: @adamjstewart what do you think?

adamjstewart: If you can get all OlmoEarth changes down to 5 lines of code, keep it in this PR. If not, separate. I really think it's possible to add a trivial U-Net decoder without skip connections with 5 lines.

If it's 5 lines and generic enough to work for other encoders, put it in `torchgeo/trainers/spatiotemporal_segmentation.py`. If it's > 5 lines or hard-coded to OlmoEarth, put it in `torchgeo/models/olmoearth.py`.

robmarkcole: - Moved OlmoEarth temporal segmentation implementation and hard-coded kwargs/weight/freeze handling into `olmoearth.py`
- Added `olmoearth_v1_temporal_segmentation` builder in `olmoearth.py`
- Reduced the trainer OlmoEarth branch to a 3-line delegation in `spatiotemporal_segmentation.py`

adamjstewart: Will review tomorrow after hacking on GeoCLIP. Trying to prioritize this PR and others needed for time series tutorials.

adamjstewart:
Talked to Robin during our time-series meeting, I think this could be split into several PRs:

* ConvLSTM model changes
* OlmoEarth model changes
* PASTIS band changes
* PASTIS plot changes
* SpatioTemporalSegmentationTask addition
* SemanticSegmentationTask ts deprecation
* Possibly more

The point is that some of these (e.g., ConvLSTM) are trivial and ready to merge, whereas others (e.g., OlmoEarth) are holding up the pipeline and making it harder to review.

robmarkcole:
This branch is now ConvLSTM model changes only.

Other branches/PR created:
- split/spatiotemporal-segmentation-task: SpatioTemporalSegmentationTask addition, based on testing-time-series-with-pastis [PR](https://github.com/robmarkcole/torchgeo/pull/2)
- split/olmoearth-model: OlmoEarth model/trainer hook changes, based on split/spatiotemporal-segmentation-task [PR](https://github.com/robmarkcole/torchgeo/pull/1)
- split/semantic-segmentation-ts-deprecation: SemanticSegmentationTask 5D input deprecation, based on main [PR](https://github.com/torchgeo/torchgeo/pull/3648)

## MERGED
- split/pastis-band-changes: PASTIS custom/duplicate S2 band support, based on main [PR](https://github.com/torchgeo/torchgeo/pull/3647)
- split/pastis-plot-changes: PASTIS RGB plot normalization, based on main [PR](https://github.com/torchgeo/torchgeo/pull/3646)

torchgeo/models/convlstm.py
```diff
@@ -114,6 +114,8 @@ def __init__(
         batch_first: bool = True,
         bias: bool = True,
         return_all_layers: bool = False,
+        num_classes: int | None = None,
```
adamjstewart: Why not make this required to be int and set a default? I'm not sure if the model would even be useful without a segmentation head.

robmarkcole:
@adamjstewart comments addressed, any outstanding issues?

torchgeo/models/convlstm.py
```diff
@@ -221,6 +236,40 @@ def forward(
...
+    def forward(
+        self,
+        input_tensor: torch.Tensor,
+        lengths: torch.Tensor | None = None,
+        hidden_state: list[tuple[torch.Tensor, torch.Tensor]] | None = None,
+    ) -> torch.Tensor:
+        """Forward pass for segmentation with the prediction head.
+
+        Args:
+            input_tensor: A 5-D Tensor of shape (t, b, c, h, w) or (b, t, c, h, w).
```
adamjstewart: I'm surprised that both formats work here. We only need to support the latter, as that is what TorchGeo has standardized on.

adamjstewart (dismissed):
Length looks much better to me now, only comment is on docs

torchgeo/models/convlstm.py
robmarkcole: Now only support batch first

adamjstewart (approved):

PR #3283, opened 2026-01-15
Show README on landing page

ashnair1:
With the release of myst-parser (v5.0.0) which supports sphinx 9+, we should now be able to display README on the landing page.


adamjstewart:
Seems to struggle with the TorchGeo logo, always adding a white background even though the original file has no background color. Is this a known bug, or is there anything we can do to make it more dynamic?

ashnair1:
Fixed by adding `dark-light` CSS class as mentioned here - https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/light-dark.html#images-and-content-that-work-in-both-themes

PR #3288, opened 2026-01-16
Presto Embeddings: add new dataset

adamjstewart:
- [x] Dataset
- [x] Tests
- [x] Docs

@ivanzvonkov @gabrieltseng

presto2


ivanzvonkov:
Thanks for putting this together @adamjstewart

PR #3294, opened 2026-01-18
GeoSlice: slice is now generic

adamjstewart:
In Python 3.15, `slice` becomes [generic](https://docs.python.org/3/library/typing.html#generics). This allows us to be more specific about our type hints. For example, the following is valid:
```python
ds[1:2, 3.4:5.6, datetime(2012, 1, 1):datetime(2012, 12, 31)]
```
However, the following is not valid:
```python
ds['east':'west', datetime(2012, 1, 1):datetime(2012, 12, 31), 1:2]
```
All GeoDataset slices must be of the form:

* xslice: [float:float:float]
* yslice: [float:float:float]
* tslice: [datetime:datetime:int]

All values [start:stop:step] are optional, and all components [xslice, yslice, tslice] are optional. This PR enables static type checking to ensure that slices are valid.

Does not yet seem to work in ~ty or~ sphinx, hold off on merging.


calebrob6:
I'm not sure what this actually does? Also why change the datetimes to datetime.datetime?

adamjstewart:
> I'm not sure what this actually does? 

Updated the PR description to be more clear, hopefully this helps.

> Also why change the datetimes to datetime.datetime?

Weird sphinx bug where it couldn't find the documentation for `datetime` but it could for `datetime.datetime`. No idea why.

adamjstewart:

PR #3296, opened 2026-01-18
Trainers: fix type hints for SSL methods

adamjstewart:
Last remaining problems with `torchgeo.trainers` flagged by `ty check`.

### DIFF
--- a/torchgeo/trainers/moco.py
+++ b/torchgeo/trainers/moco.py
@@ -6,7 +6,7 @@
 import os
 import warnings
 from collections.abc import Sequence
-from typing import Any
+from typing import Any, cast
 
 import kornia.augmentation as K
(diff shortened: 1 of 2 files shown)


torchgeo/trainers/moco.py
```diff
@@ -408,7 +408,8 @@ def training_step(
                 k = self.forward_momentum(x2)
             loss = self.criterion(q, k)
         if self.hparams['version'] == 3:
-            m = cosine_schedule(self.current_epoch, self.trainer.max_epochs, m, 1)
+            max_steps = self.trainer.max_epochs or 200
+            m = cosine_schedule(self.current_epoch, max_steps, m, 1)
```
adamjstewart, line 412: Pretty sure this was an actual bug caught by ty

PR #3300, opened 2026-01-18
Run tests in parallel with pytest-xdist

isaaccorley:
## Before

- Each CI job ran pytest serially

## After

- pytest uses xdist with \`--dist loadgroup\`
- Linux and Windows use all available workers, macOS uses two workers
- Memory-intensive model tests (marked \`xdist_group('memory_intensive')\`, see #4016) are scheduled onto the same worker and run in bfloat16 on macOS

### Checklist

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -37,8 +37,11 @@ jobs:
       - name: List installed dependencies
         run: uv tree
       - name: Run checks
+        env:
+          PYTEST_WORKERS: ${{ runner.os == 'macOS' && '2' || 'auto' }}
+          TORCHGEO_USE_BFLOAT16: ${{ runner.os == 'macOS' && 'true' || 'false' }}
         run: |
-          pytest --cov=torchgeo --cov-report=xml
+          OMP_NUM_THREADS=1 pytest -n "$PYTEST_WORKERS" --dist loadgroup --cov=torchgeo --cov-report=xml
           python -m torchgeo --help
       - name: Report coverage
         uses: codecov/codecov-action@v7.0.0
@@ -65,7 +68,7 @@ jobs:
         run: uv tree
       - name: Run checks
         run: |
-          pytest --cov=torchgeo --cov-report=xml
+          OMP_NUM_THREADS=1 pytest -n auto --dist loadgroup --cov=torchgeo --cov-report=xml
           python -m torchgeo --help
       - name: Report coverage
         uses: codecov/codecov-action@v7.0.0
@@ -93,7 +96,7 @@ jobs:
         run: uv tree
       - name: Run checks
         run: |
-          pytest --cov=torchgeo --cov-report=xml
+          OMP_NUM_THREADS=1 pytest -n auto --dist loadgroup --cov=torchgeo --cov-report=xml
           python -m torchgeo --help
       - name: Report coverage
         uses: codecov/codecov-action@v7.0.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -161,6 +161,8 @@ tests = [
     "pytest>=7.3.2",
     # pytest-cov 4+ required for pytest 7.2+ compatibility
     "pytest-cov>=4",
+    # pytest-xdist 3.0.2+ required for --dist=loadgroup with pytest 9
+    "pytest-xdist>=3.0.2",
     # pytest-socket 0.3.4+ required for wheels
     "pytest-socket>=0.3.4",
 ]
--- a/uv.lock
+++ b/uv.lock
@@ -855,6 +855,15 @@ wheels = [
 ]
 
 [[package]]
+name = "execnet"
+version = "2.1.2"
+source = { registry = "https://pypi.org/simple" }
+sdist = { url = "https://files.pythonhosted.org/packages/bf/89/780e11f9588d9e7128a3f87788354c7946a9cbb1401ad38a48c4db9a4f07/execnet-2.1.2.tar.gz", hash = "sha256:63d83bfdd9a23e35b9c6a3261412324f964c2ec8dcd8d3c6916ee9373e0befcd", size = 166622, upload-time = "2025-11-12T09:56:37.75Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/ab/84/02fc1827e8cdded4aa65baef11296a9bbe595c474f0d6d758af082d849fd/execnet-2.1.2-py3-none-any.whl", hash = "sha256:67fba928dd5a544b783f6056f449e5e3931a5c378b128bc18501f7ea79e296ec", size = 40708, upload-time = "2025-11-12T09:56:36.333Z" },
+]
+
+[[package]]
 name = "executing"
 version = "2.2.1"
 source = { registry = "https://pypi.org/simple" }
@@ -2996,6 +3005,19 @@ wheels = [
 ]
 
 [[package]]
+name = "pytest-xdist"
+version = "3.8.0"
+source = { registry = "https://pypi.org/simple" }
+dependencies = [
+    { name = "execnet" },
+    { name = "pytest" },
+]
+sdist = { url = "https://files.pythonhosted.org/packages/78/b4/439b179d1ff526791eb921115fca8e44e596a13efeda518b9d845a619450/pytest_xdist-3.8.0.tar.gz", hash = "sha256:7e578125ec9bc6050861aa93f2d59f1d8d085595d6551c2c90b6f4fad8d3a9f1", size = 88069, upload-time = "2025-07-01T13:30:59.346Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/ca/31/d4e37e9e550c2b92a9cbc2e4d0b7420a27224968580b5a447f420847c975/pytest_xdist-3.8.0-py3-none-any.whl", hash = "sha256:202ca578cfeb7370784a8c33d6d05bc6e13b4f25b5053c30a152269fd10f0b88", size = 46396, upload-time = "2025-07-01T13:30:56.632Z" },
+]
+
+[[package]]
 name = "python-dateutil"
 version = "2.9.0.post0"
 source = { registry = "https://pypi.org/simple" }
@@ -3790,6 +3812,7 @@ all = [
     { name = "pytest" },
     { name = "pytest-cov" },
     { name = "pytest-socket" },
+    { name = "pytest-xdist" },
     { name = "rioxarray" },
     { name = "ruff" },
     { name = "scipy" },
@@ -3842,6 +3865,7 @@ tests = [
     { name = "pytest" },
     { name = "pytest-cov" },
     { name = "pytest-socket" },
+    { name = "pytest-xdist" },
 ]
 
 [package.metadata]
@@ -3874,6 +3898,7 @@ requires-dist = [
     { name = "pytest", marker = "extra == 'tests'", specifier = ">=7.3.2" },
     { name = "pytest-cov", marker = "extra == 'tests'", specifier = ">=4" },
     { name = "pytest-socket", marker = "extra == 'tests'", specifier = ">=0.3.4" },
+    { name = "pytest-xdist", marker = "extra == 'tests'", specifier = ">=3.0.2" },
     { name = "rasterio", specifier = ">=1.4.3" },
     { name = "requests", specifier = ">=2.25" },
     { name = "rioxarray", marker = "extra == 'datasets'", specifier = ">=0.14.1" },


adamjstewart:
Reposting comments from Slack.

My concern is that we are already exceeding the memory limits of the macOS runners. I don't want to merge anything that makes CI slower due to memory thrashing.

Can definitely install pytest-xdist and add extra flags locally though.

isaaccorley:
> Reposting comments from Slack.
> 
> My concern is that we are already exceeding the memory limits of the macOS runners. I don't want to merge anything that makes CI slower due to memory thrashing.
> 
> Can definitely install pytest-xdist and add extra flags locally though.

I still think speeding up all other tests except the macosx tests is valuable imo because it reduces the number of concurrent jobs at a time even if the mac ones take awhile.

I've disabled parallelization for the macos tests in ci.

adamjstewart:
Another option that may provide the best of both worlds. Add a new pytest marker for "memory"-intensive tests (ViT-L and above?). Then use the following in CI:
```console
> pytest -m 'not memory' -n auto
> pytest -m 'memory' -n 0
```
Apparently one of the most requested features is the ability to mark certain tests as "cannot be run in parallel" but it's not that easy to add. This seems to be the best current workaround.

adamjstewart:
Tested locally on Apple M2 Pro 10-core with 32 GB RAM:

### Serial

```console
> time pytest -n 0
...
________________________________________________________
Executed in  457.03 secs    fish           external
   usr time  631.68 secs    1.46 millis  631.68 secs
   sys time   66.17 secs    4.73 millis   66.17 secs
```

### Parallel

```console
> time pytest -n auto
...
________________________________________________________
Executed in  138.92 secs    fish           external
   usr time   15.56 mins    1.36 millis   15.56 mins
   sys time    2.78 mins    4.49 millis    2.78 mins
```
So a 3.3x speedup is theoretically possible.

There are likely other places we can play around with fixture scope and smaller models to improve speeds.

isaaccorley:
I'm curious actually if we can just test the large models with half precision (torch.float16) which should reduce memory quite a bit

adamjstewart:
Or all models...

isaaccorley:
Did some profiling of `tests/models/*` with `pytest-memray` and unsurprisingly...

Notably the issue actually isn't the inference, it's the fact that we serialize and load a dummy checkpoint. GPT 5.2 seems to think that we can simply cast the mocked_weights to float16 during saving and loading which should significantly reduce memory usage.

```bash
====================================================== MEMRAY REPORT =======================================================
Allocation results for tests/models/test_croma.py::TestCROMALarge::test_croma_weights at the high watermark

         📦 Total memory allocated: 5.0GiB
         📏 Total allocations: 205
         📊 Histogram of allocation sizes: |█▄ ▁▁|
         🥇 Biggest allocating functions:
                - load_tensor:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/serialization.py:2039 -> 2.5GiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 384.0MiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 384.0MiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 288.0MiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 192.0MiB


Allocation results for tests/models/test_vit.py::TestViTHuge14::test_vit_weights[False] at the high watermark

         📦 Total memory allocated: 4.7GiB
         📏 Total allocations: 64
         📊 Histogram of allocation sizes: |▃█▁ ▁|
         🥇 Biggest allocating functions:
                - load_tensor:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/serialization.py:2039 -> 2.3GiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 800.0MiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 800.0MiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 600.0MiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 200.0MiB


Allocation results for tests/models/test_vit.py::TestViTHuge14::test_vit_weights[True] at the high watermark

         📦 Total memory allocated: 4.7GiB
         📏 Total allocations: 70
         📊 Histogram of allocation sizes: |▄█▁ ▁|
         🥇 Biggest allocating functions:
                - load_tensor:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/serialization.py:2039 -> 2.3GiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 800.0MiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 800.0MiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 600.0MiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 200.0MiB


Allocation results for tests/models/test_api.py::test_get_model[aurora_swin_unet] at the high watermark

         📦 Total memory allocated: 4.7GiB
         📏 Total allocations: 309
         📊 Histogram of allocation sizes: |█▃▂▁▁|
         🥇 Biggest allocating functions:
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 696.0MiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 696.0MiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 696.0MiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 696.0MiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 522.0MiB


Allocation results for tests/models/test_aurora.py::TestAurora::test_aurora_swin_unet at the high watermark

         📦 Total memory allocated: 4.7GiB
         📏 Total allocations: 283
         📊 Histogram of allocation sizes: |█▃▂▁ |
         🥇 Biggest allocating functions:
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 696.0MiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 696.0MiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 696.0MiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 696.0MiB
                - __init__:/Users/isaaccorley/github/torchgeo/.venv/lib/python3.13/site-packages/torch/nn/modules/linear.py:109 -> 522.0MiB
```

adamjstewart:
Would be nice to revive all of these memory-saving/xdist PRs.

.github/workflows/tests.yaml
```diff
@@ -37,8 +37,11 @@ jobs:
       - name: List installed dependencies
         run: uv tree
       - name: Run checks
+        env:
+          PYTEST_WORKERS: ${{ runner.os == 'macOS' && '2' || 'auto' }}
+          TORCHGEO_USE_BFLOAT16: ${{ runner.os == 'macOS' && 'true' || 'false' }}
         run: |
-          pytest --cov=torchgeo --cov-report=xml
+          OMP_NUM_THREADS=1 pytest -n "$PYTEST_WORKERS" --dist loadgroup --cov=torchgeo --cov-report=xml
```
adamjstewart, line 44: Can you explain why things like OMP_NUM_THREADS and loadgroup are needed?

tests/models/conftest.py
```diff
@@ -1,10 +1,32 @@
...
 from _pytest.fixtures import SubRequest
 
 
 @pytest.fixture(params=[True, False])
 def features_only(request: SubRequest) -> bool:
     return bool(request.param)
+
+
+@pytest.fixture(autouse=True)
+def set_precision(request: SubRequest) -> Iterator[None]:
```
adamjstewart: Could be a separate PR

tests/models/test_croma.py
```diff
@@ -65,6 +65,10 @@ def test_missing_modality(
         match: str,
     ) -> None:
         model = CROMA(image_size=8)
+        if sar_images is not None:
+            sar_images = sar_images.to(torch.get_default_dtype())
+        if optical_images is not None:
+            optical_images = optical_images.to(torch.get_default_dtype())
```
adamjstewart: Separate PR

tests/conftest.py
```diff
@@ -10,13 +10,36 @@
 import torchvision
 from pytest import MonkeyPatch
 
+MEMORY_INTENSIVE_TESTS = (
```
adamjstewart: Instead of hard-coding these in `tests/conftest.py`, can we add a mark on the tests themselves?

.github/workflows/tests.yaml
isaaccorley: pytest-xdist spawns N worker processes but PyTorch/numpy's OpenMP thread pool inside each worker will try to grab all available CPU cores

isaaccorley: the --dist loadgroup tests marked with the same group name are guaranteed to run on the same worker otherwise you would get memory intensive tests potentially running at the same time which we don't want

calebrob6 (approved):

.github/workflows/tests.yaml
```diff
@@ -37,8 +37,11 @@ jobs:
       - name: List installed dependencies
         run: uv tree
       - name: Run checks
+        env:
+          PYTEST_WORKERS: ${{ runner.os == 'macOS' && '2' || 'auto' }}
+          TORCHGEO_USE_BFLOAT16: ${{ runner.os == 'macOS' && 'true' || 'false' }}
```
adamjstewart, line 42: Why not always use this for all tests for all OSes? Can't it make all tests faster?

.github/workflows/tests.yaml
adamjstewart, line 44: Should probably add a comment documenting the meaning of all these env vars, it isn't obvious. Better yet, add them to the step `env` so the code doesn't become so complex.

.github/workflows/tests.yaml
```diff
@@ -37,8 +37,11 @@ jobs:
       - name: List installed dependencies
         run: uv tree
       - name: Run checks
+        env:
+          PYTEST_WORKERS: ${{ runner.os == 'macOS' && '2' || 'auto' }}
```
adamjstewart, line 41: Is 2 based on trial and error or guessing?

adamjstewart:
^ still a lot of unanswered questions on this PR

.github/workflows/tests.yaml
isaaccorley: No I noticed that it was faster to run at float32 since the other OS runners have more RAM so this was the way to reduce RAM specifically for Mac runners so it can run tests in parallel without maxing RAM

.github/workflows/tests.yaml
isaaccorley: Sort of, macOS runners have 3 CPU and 7GB of RAM. If you use all 3 it will crash because our test use so much RAM even with bfloat16. We just use 1 less CPU which worked.

.github/workflows/tests.yaml
adamjstewart: Hmm, why would float32 be faster than float16? I think I would still align them to reduce testing complexity assuming they're close in time.

isaaccorley: I don't think you should. I think you should defer to my advice because I actually was the one deep in the weeds running it. I don't currently have the time to rebenchmark it though.

isaaccorley: Float32 is generally faster than float16 on some hardware which is different for each runner. Although float16 reduces memory. It also is unique to each PyTorch op. There are many articles online about this.

PR #3302, opened 2026-01-19
check_integrity: replace torchvision utility

adamjstewart:
torchvision has a number of aging utilities we rely on. These utilities are neither public nor documented. We have tried to add many new features to these utilities, but `torchvision.datasets` is [bugfix only right now](https://github.com/pytorch/vision/issues/9036), so most of these changes have been rejected.

In particular, there are two features we would like to see:

* Support for cryptographically secure hashing algorithms: https://github.com/pytorch/vision/pull/4102
* Support for os.PathLike: https://github.com/pytorch/vision/pull/9336

This PR replaces `torchvision.datasets.utils.check_integrity` with a new API-compatible TorchGeo implementation that adds support for these features. In follow-up PRs, we can replace the remaining torchvision utilities by building on requests/fsspec to support more download sources.

### DIFF
--- a/torchgeo/datasets/utils.py
+++ b/torchgeo/datasets/utils.py
@@ -32,10 +32,7 @@ from pandas import Timedelta, Timestamp
 from rasterio import Affine
 from shapely import Geometry
 from torch import Tensor
-from torchvision.datasets.utils import (
-    download_and_extract_archive,
-    download_url,
-)
+from torchvision.datasets.utils import download_and_extract_archive, download_url
 from torchvision.utils import draw_segmentation_masks
 from typing_extensions import deprecated
 


torchgeo/datasets/utils.py
```diff
@@ -302,6 +297,34 @@ def __call__(self, *args: Any, **kwargs: Any) -> subprocess.CompletedProcess[byt
...
+    Returns:
+        True if file exists and *checksum* is None or matches, else False.
+    """
+    if not os.path.isfile(fpath):
+        return False
+
+    if checksum is None:
+        return True
+
+    match len(checksum):
```
calebrob6: this seems really brittle, why not just pass the digest as an arg?

adamjstewart: A couple options that preserve API compatibility:

### New digest arg

Function signature:
```python
def check_integrity(fpath: Path, checksum: str | None = None, digest: Literal['md5', 'sha256'] = 'md5') -> bool:
```
Calling:
```python
check_integrity(fpath, checksum, 'sha256')
```

### New sha256 arg

Function signature:
```python
def check_integrity(fpath: Path, md5: str | None = None, sha256: str | None = None) -> bool:
```
Calling:
```python
check_integrity(fpath, sha256=checksum)
```

Which looks better? The former makes it easy to support dozens of digest algorithms, but the latter looks a little nicer during usage.

adamjstewart: All other things considered equal, I'm leaning towards the second option as that best matches the torchvision API. Either way, I would like to eventually deprecate and remove md5, but that can happen another day.

isaaccorley (dismissed):

torchgeo/datasets/utils.py
calebrob6: yeah second option is good with me if you don't think we'd expand to other digests in the future

adamjstewart: Actually, I think I have a better design idea. I'm only planning on sha256 for now, but sha256 is also one hash collision away from being deprecated just like md5. I think we can make this more future proof.

adamjstewart:
Okay, I'm pretty happy with this implementation now. It's fully backwards-compatible with torchvision, but also adds support for any future hashing algorithms supported by hashlib. In follow-up PRs, I will try to convert as many datasets as possible to sha256 and deprecate md5.

isaaccorley (approved):

PR #3303, opened 2026-01-19
WeightsEnum: fix type hints

adamjstewart:
Fixes most ty complaints about our models, will fix the others in a separate PR to keep this one easy to review.


adamjstewart:
Bleh. Torchvision doesn't explicitly make type hints public, so mypy doesn't use them. Ty does, and complains when they disagree. So we have to add `type: ignore` to all of these places to make both happy. Or convince torchvision to make their type hints public: https://github.com/pytorch/vision/pull/9343

adamjstewart:
Or ignore `no-any-return`...

isaaccorley (dismissed):

isaaccorley:
I know it's ugly but I'd rather add type ignores inline than globally ignore

PR #3304, opened 2026-01-19
Resolve remaining model type hint problems

adamjstewart:
In combination with #3303, all model-related ty errors are fixed.

### DIFF
--- a/tests/models/test_api.py
+++ b/tests/models/test_api.py
@@ -148,7 +148,7 @@ def test_get_model_weights(builder: Callable[..., nn.Module]) -> None:
 
 @pytest.mark.parametrize('enum', enums)
 def test_get_weight(enum: WeightsEnum) -> None:
-    for weight in enum:
+    for weight in enum:  # type: ignore[not-iterable]
         assert weight == get_weight(str(weight))
 
 
--- a/tests/models/test_earthloc.py
+++ b/tests/models/test_earthloc.py
@@ -23,14 +23,13 @@ class TestEarthLoc:
     ) -> WeightsEnum:
         weights = EarthLoc_Weights.SENTINEL2_RESNET50
         path = tmp_path / f'{weights}.pth'
-        kwargs = {
-            'in_channels': weights.meta['in_chans'],
-            'image_size': weights.meta['image_size'],
-            'desc_dim': weights.meta['desc_dim'],
-            'backbone': weights.meta['encoder'],
-            'pretrained': False,
-        }
-        model = earthloc(**kwargs)
+        model = earthloc(
+            in_channels=weights.meta['in_chans'],
+            image_size=weights.meta['image_size'],
+            desc_dim=weights.meta['desc_dim'],
+            backbone=weights.meta['encoder'],
+            pretrained=False,
+        )
         torch.save(model.state_dict(), path)
         monkeypatch.setattr(weights.value, 'url', str(path))
         return weights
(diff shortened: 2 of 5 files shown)


tests/models/test_api.py
```diff
@@ -148,7 +148,7 @@ def test_get_model_weights(builder: Callable[..., nn.Module]) -> None:
 
 @pytest.mark.parametrize('enum', enums)
 def test_get_weight(enum: WeightsEnum) -> None:
-    for weight in enum:
+    for weight in enum:  # type: ignore[not-iterable]
```
adamjstewart, line 151: No idea why it thinks WeightsEnum or Enum aren't iterable. I'm unable to reproduce this in a simpler context, so can't report a bug to ty. Hopefully someone else will fix this someday.

tests/models/test_earthloc.py
```diff
@@ -23,14 +23,13 @@ def mocked_weights(
     ) -> WeightsEnum:
         weights = EarthLoc_Weights.SENTINEL2_RESNET50
         path = tmp_path / f'{weights}.pth'
-        kwargs = {
```
adamjstewart, line 26: ty is unable to detect the type of `weights.meta`, therefore resolves the entire `kwargs` to `dict[str, bool]`, then complains because most kwargs can't be bool. We either cast all types or drop the kwargs.

tests/models/test_swin.py
```diff
@@ -148,8 +148,8 @@ def mocked_weights(
             stochastic_depth_prob=0.5,
         )
         num_channels = weights.meta['in_chans']
-        out_channels = model.features[0][0].out_channels
-        model.features[0][0] = torch.nn.Conv2d(
+        out_channels = model.features[0][0].out_channels  # type: ignore[not-subscriptable]
+        model.features[0][0] = torch.nn.Conv2d(  # type: ignore[invalid-assignment]
```
adamjstewart, line 152: `model.features` is a Sequential of a Sequential. Ty struggles to detect this (I think it's a Sequential of a Sequential or something else?) and complains.

torchgeo/models/api.py
```diff
@@ -97,7 +97,7 @@
     'vit_small_patch14_dinov2': vit_small_patch14_dinov2,
 }
 
-_model_weights: dict[str | Callable[..., nn.Module], WeightsEnum] = {
+_model_weights: dict[str | Callable[..., nn.Module], WeightsEnum] = {  # type:ignore[invalid-assignment]
```
adamjstewart, line 100: Ty panics and can't tell that Aurora_Weights and friends are subclasses of WeightsEnum.

torchgeo/models/swin.py
```diff
@@ -461,21 +461,20 @@ def swin_b(
...
         # here we use the class directly to support non-default larger window-size
         window_size = weights.meta.get('window_size', 7)
-        model: SwinTransformer = torchvision.models.SwinTransformer(
-            patch_size=[4, 4],
-            embed_dim=128,
-            depths=[2, 2, 18, 2],
-            num_heads=[4, 8, 16, 32],
-            window_size=[window_size, window_size],
-            stochastic_depth_prob=0.5,
-            *args,
```
adamjstewart, line 471: Ty doesn't like `*args` coming after all these kwargs because it means `patch_size` and friends could be passed twice. It isn't wrong, we probably shouldn't use args at all. Using the same workaround we use in other models.

PR #3305, opened 2026-01-19
Add Galileo encoder with pretrained Sentinel-2 weights

Pratik0405:
## Summary

This PR adds an encoder-only, fixed-resolution implementation of the Galileo
vision transformer to TorchGeo, along with pretrained Sentinel-2 weights.

### Key features
- Encoder-only inference model (nano, tiny, base variants)
- Pretrained weights pinned to immutable commit hashes
- CPU-safe weight loading (`map_location="cpu"`)
- Variant/weight compatibility checks
- Unit tests for architecture wiring and registry integration

### Design notes
- Fixed input resolution (224×224) to match pretrained positional embeddings
- Helper modules marked private to keep the public API minimal
- Training heads and SSL components are intentionally excluded

### References
- Original implementation: https://github.com/nasaharvest/galileo
- Pretrained weights hosted on Hugging Face

Closes: N/A


tests/models/test_galileo.py
adamjstewart, line 1: Rewrite in the style of existing model tests.

torchgeo/models/__init__.py
```diff
@@ -49,6 +50,7 @@
     swin_v2_b,
     swin_v2_t,
 )
+from .tilenet import TileNet, TileNet_Weights, tilenet
```
adamjstewart: Can rebase to remove tilenet changes from the review

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,214 @@
...
+# Adapted from https://github.com/nasaharvest/galileo. Copyright (c) nasaharvest
+
+"""Galileo encoder models.
+
+Reference:
+Gabriel et al., Galileo: Learning Global & Local Features of Many Remote Sensing Modalities
+
+"""
+
+from __future__ import annotations
```
adamjstewart: Why was this needed?

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,214 @@
...
+
+from functools import partial
+from typing import Any, Literal
+
+import torch
+import torch.nn as nn
+from torchvision.models._api import Weights, WeightsEnum
+from torchvision.transforms import Resize
+
+GalileoVariant = Literal['nano', 'tiny', 'base']
```
adamjstewart: This is only used in one place, not sure if it's worth making a type alias.

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,214 @@
...
+        hidden_dim = int(dim * mlp_ratio)
+        self.fc1 = nn.Linear(dim, hidden_dim)
+        self.act = nn.GELU()
+        self.fc2 = nn.Linear(hidden_dim, dim)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return self.fc2(self.act(self.fc1(x)))
+
+
+class _Attention(nn.Module):
```
adamjstewart: This looks almost identical to PyTorch/timm's builtin attention layer, we might be able to replace this

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,214 @@
...
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x = x + self.attn(self.norm1(x))
+        x = x + self.mlp(self.norm2(x))
+        return x
+
+
+class GalileoEncoder(nn.Module):
+    """Encoder-only Galileo vision transformer.
+
+    Supported variants
```
adamjstewart: I think you need a newline before the start of a list for this to be formatted correctly.

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,214 @@
...
+    def __init__(self, variant: GalileoVariant = 'base') -> None:
+        """Initialize a Galileo encoder.
+
+        Args:
+            variant: Model size variant ('nano', 'tiny', or 'base').
+        """
+        super().__init__()
+
+        if variant not in _GALILEO_CONFIGS:
+            raise ValueError(f'Unknown Galileo variant: {variant}')
```
adamjstewart: This won't be reachable without ignoring the type hints, may be challenging to get coverage of

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,214 @@
...
+            [
+                _Block(dim=embed_dim, num_heads=num_heads, mlp_ratio=MLP_RATIO)
+                for _ in range(depth)
+            ]
+        )
+
+        self.norm = nn.LayerNorm(embed_dim)
+
+        # Init
+        nn.init.trunc_normal_(self.pos_embed, std=0.02)
```
adamjstewart: The reference implementation does not have this?

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,214 @@
...
+    'base': {'embed_dim': 768, 'depth': 12, 'num_heads': 12},
+}
+
+IMAGE_SIZE = 224
+PATCH_SIZE = 16
+IN_CHANNELS = 4
+MLP_RATIO = 4.0
+
+
+class _MLP(nn.Module):
```
adamjstewart: Why prefixed with underscore? Missing the second dropout layer

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,214 @@
...
+        q, k, v = qkv.permute(2, 0, 3, 1, 4)
+
+        attn = (q @ k.transpose(-2, -1)) * self.scale
+        attn = attn.softmax(dim=-1)
+
+        x = (attn @ v).transpose(1, 2).reshape(B, N, C)
+        return self.proj(x)
+
+
+class _Block(nn.Module):
```
adamjstewart: Looks completely different from the Galileo implementation

adamjstewart (changes requested):

torchgeo/models/galileo.py
Pratik0405: Okay will look into it.

torchgeo/models/galileo.py
Pratik0405: Ruff aaparently was failing for this line so added an underscore

Pratik0405: Also will check if timm has similar implementation for MLP then we can use that instead. Also will add dropout layer

torchgeo/models/galileo.py
Pratik0405: Will make sure to change in upcoming commit

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1234 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+# Adapted from https://github.com/nasaharvest/galileo. Copyright (c) nasaharvest
```
adamjstewart:
```suggestion
# Adapted from https://github.com/nasaharvest/galileo. Copyright (c) Presto Authors.
```
From https://github.com/nasaharvest/galileo/blob/main/LICENSE

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1234 @@
...
+"""Galileo encoder models.
+
+Reference:
+Gabriel et al., Galileo: Learning Global & Local Features of Many Remote Sensing Modalities
+
+"""
+import collections.abc
+import itertools
+from collections import OrderedDict
+from collections import OrderedDict as OrderedDictType
```
adamjstewart, line 14: Why import this twice?

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1234 @@
...
+        "DW_static": [STATIC_BANDS.index(b) for b in STATIC_DW_BANDS],
+        "WC_static": [STATIC_BANDS.index(b) for b in STATIC_WC_BANDS],
+    }
+)
+
+
+def get_2d_sincos_pos_embed_with_resolution(
+    embed_dim, grid_size, res, cls_token=False, device="cpu"
+):
+    """grid_size: int of the grid height and width
```
adamjstewart: Need to reformat docstrings to follow Google style.

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1234 @@
...
+            [
+                torch.zeros([n, 1, embed_dim], device=pos_embed.device),
+                pos_embed,
+            ],
+            dim=1,
+        )
+    return pos_embed
+
+
+def get_2d_sincos_pos_embed_from_grid_torch(embed_dim, grid):
```
adamjstewart: Missing docstring

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1234 @@
...
+    month_table = torch.concatenate([sin_table[:-1], cos_table[:-1]], axis=-1)
+
+    return month_table  # (M, D)
+
+
+# thanks to https://github.com/bwconrad/flexivit/ for this nice implementation
+
+# of the FlexiPatchEmbed module
+def to_2tuple(x: Any) -> tuple:
+    if isinstance(x, collections.abc.Iterable) and not isinstance(x, str):
```
adamjstewart, line 225:
```python
from collections.abc import Iterable
```

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1234 @@
...
+            x = rearrange(x, "(b t) c h w -> b h w t c", b=batch_size, t=num_timesteps)
+        else:
+            x = rearrange(x, "b c h w -> b h w c")
+        x = self.norm(x)
+
+        return x
+
+
+class Attention(nn.Module):
+    # https://github.com/huggingface/pytorch-image-models/blob/main/timm/models/vision_transformer.py
```
adamjstewart: Can we import and use this instead of copy-n-paste?

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1234 @@
...
+        base_patch_size: int = 4,
+        use_channel_embs: bool = True,
+        drop_path: float = 0.0,
+    ):
+        super().__init__()
+
+        self.space_time_groups = SPACE_TIME_BANDS_GROUPS_IDX
+        self.space_groups = SPACE_BAND_GROUPS_IDX
+        self.time_groups = TIME_BAND_GROUPS_IDX
+        self.static_groups = STATIC_BAND_GROUPS_IDX
```
adamjstewart: These shouldn't be hard-coded, we should make these init args. Otherwise this model only works for one sensor.

adamjstewart (changes requested):
This is looking much better now.

Some remaining changes:

* We might be able to reuse the Attention class from timm
* We should make sure the model doesn't only support Sentinel-2, things like # channels and dimensions shouldn't be hard-coded
* All classes/functions should have Google-style docstrings like the following:

```python
"""short 1-line description of function.

Args:
    arg1: The first argument.
    arg2: The second argument.

Returns:
    Whatever the function returns.
"""
```
Otherwise I don't actually know what this function is for and how to use it.

torchgeo/models/galileo.py
Pratik0405: the OrderedDict was throwing some import error, so was trial and erroring it worked when imported twice hence kept it the way

adamjstewart: Let me know what the error is and I can fix it

Pratik0405: image

does not recognise the OrderedDict

Pratik0405: image
cant run pytests without it

adamjstewart: Well that's because you imported it and then changed the name, but you refer to it by two different names. Why did you rename it?

Pratik0405: one is the type alias and other is the dictonary hence I had to import it twice so it does not give an import error

Pratik0405: When I just import the first one, i.e., "OrderedDict," then it throws the error on the type alias, but then when I import the other (ordered dictType) and remove first, i.e just "OrderedDict" it gives error on the ordered dictType
.

Pratik0405: I checked the original implementation; one import is from typing and the other from collections. I get it now, we should also do something similar so the pytests work and the type hints are there

adamjstewart: One is a deprecated alias for the other: https://docs.python.org/3/library/typing.html#typing.OrderedDict

You should only use `collections.OrderedDict` for both.

docs/api/models/galileo.rst
```diff
@@ -0,0 +1,8 @@
+Galileo
+=======
+
+.. currentmodule:: torchgeo.models.galileo
+.. autofunction:: galileo
+.. autoclass:: torchgeo.models.galileo.GalileoEncoder
```
adamjstewart:
```suggestion
.. autoclass:: GalileoEncoder
```

docs/api/models/galileo.rst
```diff
@@ -0,0 +1,8 @@
+Galileo
+=======
+
+.. currentmodule:: torchgeo.models.galileo
```
adamjstewart:
```suggestion
.. currentmodule:: torchgeo.models
```

docs/api/models/galileo.rst
```diff
@@ -0,0 +1,8 @@
...
+.. autofunction:: galileo
+.. autoclass:: torchgeo.models.galileo.GalileoEncoder
+.. autoclass:: GalileoWeights
+   :members:
```
adamjstewart:
```suggestion
```

docs/api/weights/sentinel2.csv
```diff
@@ -51,3 +51,6 @@ Unet_Weights.SENTINEL2_FTW_PRUE_EFNETB3,8,`link <https://github.com/fieldsofthew
 Unet_Weights.SENTINEL2_FTW_PRUE_EFNETB5,8,`link <https://github.com/fieldsoftheworld/ftw-baselines>`__,,"non-commercial",,,
 Unet_Weights.SENTINEL2_FTW_PRUE_EFNETB7,8,`link <https://github.com/fieldsoftheworld/ftw-baselines>`__,,"non-commercial",,,
 EarthLoc_Weights.SENTINEL2_RESNET50,3,`link <https://github.com/gmberton/EarthLoc>`__,`link <https://arxiv.org/abs/2403.06758>`__,"MIT",,,
+GalileoWeights.GALILEO_S2_NANO_V1,13,`link <https://github.com/nasaharvest/galileo>`__,`link <https://arxiv.org/abs/2502.09356>`__,"MIT",,,
+GalileoWeights.GALILEO_S2_TINY_V1,13,`link <https://github.com/nasaharvest/galileo>`__,`link <https://arxiv.org/abs/2502.09356>`__,"MIT",,,
+GalileoWeights.GALILEO_S2_BASE_V1,13,`link <https://github.com/nasaharvest/galileo>`__,`link <https://arxiv.org/abs/2502.09356>`__,"MIT",,,
```
adamjstewart, line 59: Galileo is not only trained on Sentinel-2, I think these belong in `agnostic.csv`

docs/api/models.rst
```diff
@@ -31,6 +31,9 @@ TorchGeo contains a number of model architectures depending on the task you are
    models/earthloc
    models/farseg
    models/fcn
+   models/fc-siamese-networks
+   models/galileo
+   models/l-tae
```
adamjstewart, line 37: fc-siamese-networks and l-tae are now duplicated, this should only have galileo

tests/models/test_galileo.py
```diff
@@ -0,0 +1,136 @@
...
+SPACE_TIME_GROUPS = OrderedDict({'g0': list(range(SPACE_TIME_BANDS))})
+
+SPACE_GROUPS = OrderedDict({'g0': list(range(SPACE_BANDS))})
+
+TIME_GROUPS = OrderedDict({'g0': list(range(TIME_BANDS))})
+
+STATIC_GROUPS = OrderedDict({'g0': list(range(STATIC_BANDS))})
+
+
+def make_inputs(
```
adamjstewart, line 31: This could be a fixture

tests/models/test_galileo.py
```diff
@@ -0,0 +1,136 @@
...
+def weights(request: SubRequest) -> WeightsEnum:
+    return request.param
+
+
+@pytest.fixture
+def mocked_weights(
+    tmp_path: Path, monkeypatch: MonkeyPatch, load_state_dict_from_url: Any
+) -> GalileoWeights:
+    """Create a tiny local checkpoint to test weight loading."""
+    w: GalileoWeights = GalileoWeights.GALILEO_S2_NANO_V1
```
adamjstewart, line 58: Can we name this `weights` instead of `w` to make it more clear?

tests/models/test_galileo.py
```diff
@@ -0,0 +1,136 @@
...
+    )
+
+    inputs = make_inputs(batch=1)
+    out = model(**inputs, patch_size=16)
+
+    assert isinstance(out[0], torch.Tensor)
+
+
+@pytest.mark.slow
+@pytest.mark.parametrize('w', list(GalileoWeights))
```
adamjstewart, line 126: You created a fixture for this but you never actually used it, why not use it here like all the other tests?

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1616 @@
...
+
+Reference:
+Gabriel et al., Galileo: Learning Global & Local Features of Many Remote Sensing Modalities
+"""
+
+import collections.abc
+import itertools
+from collections import OrderedDict
+from collections import OrderedDict as OrderedDictType
+from collections.abc import Sequence
```
adamjstewart, line 15:
```suggestion
from collections.abc import Iterable, Sequence
```

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1616 @@
...
+"""Galileo encoder models.
+
+Reference:
+Gabriel et al., Galileo: Learning Global & Local Features of Many Remote Sensing Modalities
+"""
+
+import collections.abc
+import itertools
+from collections import OrderedDict
+from collections import OrderedDict as OrderedDictType
```
adamjstewart, line 14:
```suggestion
```

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1616 @@
...
+# Licensed under the MIT License.
+# Adapted from https://github.com/nasaharvest/galileo. Copyright (c) Presto Authors.
+
+"""Galileo encoder models.
+
+Reference:
+Gabriel et al., Galileo: Learning Global & Local Features of Many Remote Sensing Modalities
+"""
+
+import collections.abc
```
adamjstewart, line 11:
```suggestion
```

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1616 @@
...
+)
+
+STATIC_BAND_GROUPS_IDX: OrderedDictType[str, list[int]] = OrderedDict(
+    {
+        'LS': [STATIC_BANDS.index(b) for b in LANDSCAN_BANDS],
+        'location': [STATIC_BANDS.index(b) for b in LOCATION_BANDS],
+        'DW_static': [STATIC_BANDS.index(b) for b in STATIC_DW_BANDS],
+        'WC_static': [STATIC_BANDS.index(b) for b in STATIC_WC_BANDS],
+    }
+)
```
adamjstewart, line 113: Any of these constants that are unused can be removed. All sensor configuration should be a parameter that the user can define themselves.

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1616 @@
...
+    device: str = 'cpu',
+) -> Tensor:
+    """Generate 2D sine-cosine positional embeddings with resolution scaling.
+
+    Args:
+        embed_dim (int): Dimension of the positional embedding.
+        grid_size (int): Height and width of the spatial grid.
+        res (Tensor): Resolutions for each sample, shape (N,).
+        cls_token (bool, optional): Whether to prepend a class token. Defaults to False.
+        device (str, optional): Device for tensors. Defaults to "cpu".
```
adamjstewart, line 130: Types can be removed, there are automatically extracted from the function signature.

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1616 @@
...
+def to_2tuple(x: Any) -> tuple[Any, Any]:
+    """Convert input into a 2-tuple.
+
+    Args:
+        x (Any): Input object.
+
+    Returns:
+        tuple: A tuple of length 2.
+    """
+    if isinstance(x, collections.abc.Iterable) and not isinstance(x, str):
```
adamjstewart, line 225:
```suggestion
    if isinstance(x, Iterable) and not isinstance(x, str):
```

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1616 @@
...
+    angles = torch.arange(0, 13) / (12 / (2 * np.pi))
+
+    sin_table = torch.sin(torch.stack([angles for _ in range(embed_dim // 2)], dim=-1))
+    cos_table = torch.cos(torch.stack([angles for _ in range(embed_dim // 2)], dim=-1))
+    month_table = torch.concatenate([sin_table[:-1], cos_table[:-1]], dim=-1)
+
+    return month_table
+
+
+def to_2tuple(x: Any) -> tuple[Any, Any]:
```
adamjstewart, line 216:
```suggestion
def to_2tuple(x: object) -> tuple[object, object]:
```
Might actually be better to use a Generic here since the type does not change from before to after.

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1616 @@
...
+    Returns:
+        tuple: A tuple of length 2.
+    """
+    if isinstance(x, collections.abc.Iterable) and not isinstance(x, str):
+        return tuple(x)
+    return tuple(itertools.repeat(x, 2))
+
+
+# thanks to https://github.com/bwconrad/flexivit/ for this nice implementation
+# of the FlexiPatchEmbed module
```
adamjstewart, line 231: If this is copied from https://github.com/bwconrad/flexivit/, it needs:
```
Copyright (c) 2023 Ben Conrad
```

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1616 @@
...
+    This subclass of nn.ModuleList adds an `_init_weights` method that can be
+    applied to child modules to initialize linear layers with Xavier uniform
+    initialization.
+    """
+
+    def _init_weights(self, m: nn.Module) -> None:
+        """Initialize weights of linear layers.
+
+        Args:
+            m (nn.Module): Module to initialize.
```
adamjstewart, line 605: Can we use `module` instead of `m` to make this parameter more clear?

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1616 @@
...
+        if isinstance(m, nn.Linear):
+            torch.nn.init.xavier_uniform_(m.weight)
+            if isinstance(m, nn.Linear) and m.bias is not None:
+                nn.init.constant_(m.bias, 0)
+
+
+class GalileoBase(nn.Module):
+    """Base class containing shared logic for Galileo encoders.
+
+    This class manages:
```
adamjstewart, line 616: You probably need a newline before the start of your list for Sphinx to format it correctly

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1616 @@
...
+        use_channel_embs (bool): Whether channel embeddings are learnable.
+        drop_path (float): Stochastic depth drop probability.
+    """
+
+    def __init__(
+        self,
+        space_time_groups: dict[str, Any],
+        space_groups: dict[str, Any],
+        time_groups: dict[str, Any],
+        static_groups: dict[str, Any],
```
adamjstewart, line 639: Try to avoid Any. What types should dictionary values actually be?

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1616 @@
...
+        self.t_channel_embed = nn.Parameter(
+            torch.zeros(len(self.time_groups), int(embedding_size * 0.25)), **args
+        )
+        self.st_channel_embed = nn.Parameter(
+            torch.zeros(len(self.static_groups), int(embedding_size * 0.25)), **args
+        )
+
+        self.apply(self._init_weights)
+
+    def _init_weights(self, m: nn.Module) -> None:
```
adamjstewart, line 711: Why do we need both this and `ModuleListWithInit`?

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1616 @@
...
+        """
+        if isinstance(m, nn.Linear):
+            torch.nn.init.xavier_uniform_(m.weight)
+            if m.bias is not None:
+                nn.init.constant_(m.bias, 0)
+
+    @classmethod
+    def collapse_and_combine_hwtc(
+        cls,
+        s_t_x: torch.Tensor,
```
adamjstewart, line 725:
```python
from torch import Tensor
```

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1616 @@
...
+
+    Attributes:
+        GALILEO_S2_NANO_V1: Pretrained 'nano' variant.
+        GALILEO_S2_TINY_V1: Pretrained 'tiny' variant.
+        GALILEO_S2_BASE_V1: Pretrained 'base' variant.
+    """
+
+    GALILEO_S2_NANO_V1 = Weights(
+        url='https://huggingface.co/nasaharvest/galileo/resolve/0bbc53293a37dea7d563566f015c5527bdaf3793/models/nano/encoder.pt',
+        transforms=partial(Resize, size=(IMAGE_SIZE, IMAGE_SIZE)),
```
adamjstewart, line 1513: I don't think you want to use a partial here, you can just instantiate the class with the right inputs.

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1616 @@
...
+        meta={
+            'variant': 'nano',
+            'in_channels': IN_CHANNELS,
+            'embed_dim': 192,
+            'dataset': 'Sentinel-2',
```
adamjstewart, line 1518: Not only trained on Sentinel-2

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1616 @@
...
+            'embed_dim': 768,
+            'dataset': 'Sentinel-2',
+            'license': 'MIT',
+        },
+    )
+
+
+def galileo(
+    *,
+    variant: str | None = None,
```
adamjstewart, line 1550:
```suggestion
    variant: Literal['nano', 'tiny', 'base'] | None = 'base',
```

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1616 @@
...
+            'variant': 'base',
+            'in_channels': IN_CHANNELS,
+            'embed_dim': 768,
+            'dataset': 'Sentinel-2',
+            'license': 'MIT',
+        },
+    )
+
+
+def galileo(
```
adamjstewart, line 1548: I wonder if we should instead have different helper functions and weights enums for different sizes...

torchgeo/models/galileo.py
```diff
@@ -0,0 +1,1616 @@
...
+            f'Unknown Galileo variant: {variant!r}. '
+            f'Available: {list(_GALILEO_CONFIGS.keys())}'
+        )
+
+    cfg = _GALILEO_CONFIGS[variant]
+
+    assert space_time_groups is not None
+    assert space_groups is not None
+    assert time_groups is not None
+    assert static_groups is not None
```
adamjstewart, line 1599: I don't understand. You allow None as input, but None isn't valid?

adamjstewart (changes requested):

PR #3306, opened 2026-01-19
docs: add GitHub Copilot instructions

isaaccorley:
I scraped all maintainer PR review comments since the start of TorchGeo and then used Opus 4.5 and Codex 5.2 to create a concise copilot instructions file.

Note that this is what the Copilot PR review bot consumes which should improve the automated reviews.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -9,16 +9,15 @@ Datasets, samplers, transforms, and pre-trained models for geospatial data.
 pip install -e ".[all]"
 
 # Lint (run from repo root)
-pre-commit run --all-files
-# or: ruff format && ruff check && mypy && prettier --write .
+ruff format && ruff check && mypy && prettier --write .
 
 # Test
 pytest --cov=torchgeo tests/                                                        # all (skip slow)
-pytest --cov=torchgeo.datasets.foo tests/datasets/test_foo.py                       # single file
-pytest --cov=torchgeo.datasets.foo tests/datasets/test_foo.py::TestFooDataset::test_getitem  # single method
+pytest --cov=torchgeo.datasets tests/datasets/test_foo.py                          # single file
+pytest --cov=torchgeo.datasets tests/datasets/test_foo.py::TestFooDataset::test_getitem  # single method
 pytest -m "" --cov=torchgeo tests/                                                  # include slow
 
-# Docs (requires Pandoc + pip install -e ".[docs]")
+# Docs (requires Pandoc + pip install ".[docs]")
 cd docs && make clean && make html
 ```
 
@@ -43,13 +42,13 @@ tests/{data/<dataset>/,datasets/,...}
 - Single quotes, no trailing commas
 - Imports: isort order (stdlib → third-party → local)
 - Local imports (`from .utils`) in `torchgeo/`; absolute (`from torchgeo`) in `tests/`
-- `# noqa: E501` only for unsplittable URLs
 
 ### Type Hints (strict mypy)
 
 - All functions require full annotations
 - Union: `X | Y` (not `Union[X, Y]`)
-- Avoid `Any`; use `Sequence` over `List`/`Tuple`
+- Avoid `Any`; use `Sequence` for parameters, `list`/`tuple` for return values
+- Prefer built-in types (`list`, `dict`, `tuple`) over `typing.List`, etc.
 - `type: ignore` only for external lib issues
 
 ### Docstrings (Google style)
@@ -71,7 +70,6 @@ def func(x: Tensor, weights: Weights | None = None) -> Tensor:
     Warns:
         UserWarning: If weights are outdated.
 
-    .. versionadded:: 0.6
     """
 ```
 
@@ -84,8 +82,9 @@ def func(x: Tensor, weights: Weights | None = None) -> Tensor:
 ### RST Conventions (docs/docstrings)
 
 - Monospace: double backticks ` ``code`` `
+- Bold: `**text**`
 - Italics: `*text*`
-- Single backticks create links (not monospace)
+- Single backticks create links only when followed by underscore; otherwise italics
 - Blank line required before bulleted lists
 
 ### Classes
@@ -102,8 +101,6 @@ class Foo(NonGeoDataset):
     def __init__(self, root: Path = 'data', download: bool = False) -> None:
 ```
 
-- Class attributes for constants (url, classes, filename_glob, etc.)
-
 ## Testing
 
 ```python
@@ -116,16 +113,12 @@ class TestFooDataset:
         x = dataset[0]
         assert x['image'].ndim == 3
         assert x['mask'].ndim == 2
-
-    @pytest.mark.slow
-    def test_download(self, tmp_path: Path) -> None:
-        FooDataset(root=tmp_path, download=True)
 ```
 
-- 100% coverage required; check with `pytest --cov=torchgeo.datasets.foo tests/datasets/test_foo.py`
+- 100% coverage required; check with `pytest --cov=torchgeo.datasets tests/datasets/test_foo.py`
 - Fake data in `tests/data/<dataset>/` (never real data); `data.py` generates it
 - `@pytest.mark.slow` for downloads/network
-- `pytest.importorskip('pkg')` for optional deps
+- `pytest.importorskip('pkg', minversion='X.X')` for optional deps
 - `plt.close()` at end of plot tests
 
 ## Dataset Implementation
@@ -138,9 +131,8 @@ class MyDataset(RasterDataset):
     is_image = True
 ```
 
-- `root="data"` default; `_verify()` for existence checks
+- `root='data'` default; `_verify()` for existence checks
 - Sample dict: `{"image": tensor, "mask": tensor}`
-- Use `torchvision.datasets.utils` for downloads
 - PIL for non-geo images, rasterio for geo images
 
 ### New Dataset Checklist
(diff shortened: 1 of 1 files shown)


adamjstewart:
Surprisingly human-readable. Any reason we can't put this information in https://torchgeo.readthedocs.io/en/stable/user/contributing.html so we don't duplicate ourselves? We can use a symlink if copilot only searches a specific file.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Commands
+
+```bash
+# Install
+pip install -e ".[all]"
+
+# Lint (run from repo root)
+pre-commit run --all-files
+# or: ruff format && ruff check && mypy && prettier --write .
```
adamjstewart: Would prefer this since pre-commit rapidly becomes outdated but the situation is better now

.github/copilot-instructions.md
```diff
@@ -0,0 +1,207 @@
...
+
+# Test
+pytest --cov=torchgeo tests/                                                        # all (skip slow)
+pytest --cov=torchgeo.datasets.foo tests/datasets/test_foo.py                       # single file
+pytest --cov=torchgeo.datasets.foo tests/datasets/test_foo.py::TestFooDataset::test_getitem  # single method
```
adamjstewart: `--cov=torchgeo.datasets.foo` doesn't work due to known incompatibilities between coverage.py and numpy/torchvision. You have to use `--cov=torchgeo.datasets` instead.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,207 @@
...
+pytest -m "" --cov=torchgeo tests/                                                  # include slow
+
+# Docs (requires Pandoc + pip install -e ".[docs]")
```
adamjstewart: No longer requires `-e`

.github/copilot-instructions.md
```diff
@@ -0,0 +1,207 @@
...
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+```
+
+### Formatting (Ruff)
+
+- Single quotes, no trailing commas
+- Imports: isort order (stdlib → third-party → local)
+- Local imports (`from .utils`) in `torchgeo/`; absolute (`from torchgeo`) in `tests/`
+- `# noqa: E501` only for unsplittable URLs
```
adamjstewart: Ruff automatically ignores these, we don't use `# noqa: E501` anywhere

.github/copilot-instructions.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Type Hints (strict mypy)
+
+- All functions require full annotations
+- Union: `X | Y` (not `Union[X, Y]`)
+- Avoid `Any`; use `Sequence` over `List`/`Tuple`
```
adamjstewart: Sequence is preferred for parameters, list/tuple is preferred for return values

.github/copilot-instructions.md
```diff
@@ -0,0 +1,207 @@
...
+    Returns:
+        Output tensor.
+
+    Raises:
+        ValueError: If x is empty.
+
+    Warns:
+        UserWarning: If weights are outdated.
+
+    .. versionadded:: 0.6
```
adamjstewart: Hoping this won't start inserting this text everywhere

.github/copilot-instructions.md
```diff
@@ -0,0 +1,207 @@
...
+- Period after first line; blank line after first sentence
+- Don't document default values in Args (Sphinx adds them)
+- `.. versionadded::` for new features; `.. deprecated::` for deprecations
+- Inline comments only for "why", not "what"
+
+### RST Conventions (docs/docstrings)
+
+- Monospace: double backticks ` ``code`` `
+- Italics: `*text*`
+- Single backticks create links (not monospace)
```
adamjstewart: Only when followed by an underscore, without they create italics

.github/copilot-instructions.md
```diff
@@ -0,0 +1,207 @@
...
+- Include `Args`, `Returns`, `Raises`, `Warns` sections as needed
+- Period after first line; blank line after first sentence
+- Don't document default values in Args (Sphinx adds them)
+- `.. versionadded::` for new features; `.. deprecated::` for deprecations
+- Inline comments only for "why", not "what"
+
+### RST Conventions (docs/docstrings)
+
+- Monospace: double backticks ` ``code`` `
+- Italics: `*text*`
```
adamjstewart, line 86: Also worth adding bold to the list

.github/copilot-instructions.md
```diff
@@ -0,0 +1,207 @@
...
+
+    .. versionadded:: 0.5
+    """
+    url = 'https://...'
+    classes = ['background', 'building']
+
+    def __init__(self, root: Path = 'data', download: bool = False) -> None:
+```
+
+- Class attributes for constants (url, classes, filename_glob, etc.)
```
adamjstewart: Only needed for mutables (dicts, lists), we should strive to use tuples instead

.github/copilot-instructions.md
```diff
@@ -0,0 +1,207 @@
...
+        return FooDataset(root='tests/data')
+
+    def test_getitem(self, dataset: FooDataset) -> None:
+        x = dataset[0]
+        assert x['image'].ndim == 3
+        assert x['mask'].ndim == 2
+
+    @pytest.mark.slow
+    def test_download(self, tmp_path: Path) -> None:
+        FooDataset(root=tmp_path, download=True)
```
adamjstewart: We don't use this, it would flood CI storage

.github/copilot-instructions.md
```diff
@@ -0,0 +1,207 @@
...
+```
+
+- 100% coverage required; check with `pytest --cov=torchgeo.datasets.foo tests/datasets/test_foo.py`
```
adamjstewart: .foo not supported

.github/copilot-instructions.md
```diff
@@ -0,0 +1,207 @@
...
+- Fake data in `tests/data/<dataset>/` (never real data); `data.py` generates it
+- `@pytest.mark.slow` for downloads/network
+- `pytest.importorskip('pkg')` for optional deps
```
adamjstewart: With minversion preferably

.github/copilot-instructions.md
```diff
@@ -0,0 +1,207 @@
...
+```python
+class MyDataset(RasterDataset):
+    filename_glob = '*.tif'
+    filename_regex = r'^(?P<date>\d{8})_.*\.tif$'
+    date_format = '%Y%m%d'
+    is_image = True
+```
+
+- `root="data"` default; `_verify()` for existence checks
+- Sample dict: `{"image": tensor, "mask": tensor}`
```
adamjstewart: Single quotes

.github/copilot-instructions.md
```diff
@@ -0,0 +1,207 @@
...
+- Use `torchvision.datasets.utils` for downloads
```
adamjstewart: Not for long

.github/copilot-instructions.md
```diff
@@ -0,0 +1,207 @@
...
+2. Import in `torchgeo/datasets/__init__.py`
+3. `tests/data/foo/data.py` (fake data matching real structure)
+4. `tests/datasets/test_foo.py` (100% coverage)
+5. `docs/api/datasets.rst`
+6. `docs/api/datasets/[non_]geo_datasets.csv`
+
+## Error Handling
+
+- Never catch bare `Exception`; be specific
+- Include context: `f"Dataset not found in {self.root}"`
```
adamjstewart: We have builtin types for this

.github/copilot-instructions.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Lightning
+
+- `prepare_data`: downloads only, NO state
+- `setup`: create dataset objects
+- `on_after_batch_transfer`: GPU augmentations
+
+## Performance
+
+- Compile regex outside loops; glob once, reuse result
```
adamjstewart: Compilation actually doesn't matter unless we are using hundreds of regexes, Python caches this automatically.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,207 @@
...
+- Class attributes for reusable objects (not per-call instantiation)
+- `torch.view` > reshape; specify `dim` in `squeeze()`
```
adamjstewart: We should really prefer einops.reshape

.github/copilot-instructions.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Dependencies
+
+- Find actual minimum working version; document WHY
+- Sync versions across requirements files
+- Lazy import optional deps inside using function; helpful ImportError message
+
+## Git
+
+- Conventional commits: `feat|fix|refactor|docs|test|chore: description`
```
adamjstewart: We have not agreed to this. I personally prefer PR titles matching past release notes. I don't care about commit messages. I think we should use milestones instead of PR titles to decide backports.

adamjstewart (changes requested):
Of course I have nits for my nits.

isaaccorley:
> Surprisingly human-readable. Any reason we can't put this information in [torchgeo.readthedocs.io/en/stable/user/contributing.html](https://torchgeo.readthedocs.io/en/stable/user/contributing.html) so we don't duplicate ourselves? We can use a symlink if copilot only searches a specific file.

Copilot unfortunately doesn't follow references to other files. The goal of this instructions file is to provide a super minimal context before development/review happens by the agent. I'll follow up with another PR with a "Best Practices" section in CONTRIBUTING.md

PR #3308, opened 2026-01-20
Fix CopernicusFM dtype handling

isaaccorley:
Respect configured dtypes in CopernicusFM tensor creation and resizing.

@wangyi111 can I get your blessing on this?

## AI Usage Disclosure

### DIFF
--- a/tests/models/conftest.py
+++ b/tests/models/conftest.py
@@ -1,23 +1,10 @@
 # Copyright (c) TorchGeo Contributors. All rights reserved.
 # Licensed under the MIT License.
 
-from collections.abc import Iterator
-
 import pytest
-import torch
 from _pytest.fixtures import SubRequest
 
 
 @pytest.fixture(params=[True, False])
 def features_only(request: SubRequest) -> bool:
     return bool(request.param)
-
-
-@pytest.fixture
-def use_float16() -> Iterator[None]:
-    dtype = torch.get_default_dtype()
-    torch.set_default_dtype(torch.float16)
-    try:
-        yield
-    finally:
-        torch.set_default_dtype(dtype)
--- a/tests/models/conftest.py
+++ b/tests/models/conftest.py
@@ -1,10 +1,23 @@
 # Copyright (c) TorchGeo Contributors. All rights reserved.
 # Licensed under the MIT License.
 
+from collections.abc import Iterator
+
 import pytest
+import torch
 from _pytest.fixtures import SubRequest
 
 
 @pytest.fixture(params=[True, False])
 def features_only(request: SubRequest) -> bool:
     return bool(request.param)
+
+
+@pytest.fixture
+def use_bfloat16() -> Iterator[None]:
+    dtype = torch.get_default_dtype()
+    torch.set_default_dtype(torch.bfloat16)
+    try:
+        yield
+    finally:
+        torch.set_default_dtype(dtype)
--- a/tests/models/conftest.py
+++ b/tests/models/conftest.py
@@ -14,9 +14,9 @@ def features_only(request: SubRequest) -> bool:
 
 
 @pytest.fixture
-def use_bfloat16() -> Iterator[None]:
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yaml
```diff
@@ -44,7 +44,11 @@ jobs:
         run: pip list
       - name: Run pytest checks
         run: |
-          pytest --cov=torchgeo --cov-report=xml
+          if [ "${{ runner.os }}" = "macOS" ]; then
```
adamjstewart: Could condition this directly in conftest.py to keep this file simple. Also makes it easier to debug locally if users don't have to toggle env vars to reproduce CI.

tests/models/conftest.py
```diff
@@ -1,10 +1,27 @@
...
 @pytest.fixture(params=[True, False])
 def features_only(request: SubRequest) -> bool:
     return bool(request.param)
+
+
+@pytest.fixture(autouse=True)
+def set_precision() -> Iterator[None]:
+    previous_dtype = torch.get_default_dtype()
+
+    if os.environ.get('TORCHGEO_USE_BFLOAT16') == 'true':
```
adamjstewart: Why bfloat16 instead of float16? Won't both use the same memory?

Why only macOS? Shouldn't this make our tests faster and use less memory on all OSes?

tests/models/test_btc.py
```diff
@@ -23,8 +23,8 @@ def test_btc_sizes(
     ) -> None:
         model = BTC(backbone=backbone, backbone_pretrained=False)
         model.eval()
-        with torch.no_grad():
-            x = torch.randn(2, 2 * 3, 256, 256)
+        with torch.inference_mode():
```
adamjstewart, line 26: Would like to split "float16" and "inference_mode" into separate PRs. They have the same goal, but I want to measure the speed impact of each individually. When I tried using inference_mode in more places, things got slower on average. Also would prefer throwing this in conftest.py so that _all_ tests run in inference mode, not just a few.

tests/models/test_btc.py
```diff
@@ -23,8 +23,8 @@ def test_btc_sizes(
...
+            x = torch.randn(1, 2 * 3, 64, 64)
```
adamjstewart: Dimension changes could be a third PR, there are a lot more places where this could be used to save time/memory.

torchgeo/models/copernicusfm.py
```diff
@@ -64,10 +64,13 @@ def resize_abs_pos_embed(
 
     # Interpolate position embedding
     pos_embed = pos_embed.reshape(1, old_size[0], old_size[1], -1).permute(0, 3, 1, 2)
+    dtype = pos_embed.dtype
```
adamjstewart, line 66: Model bug fixes should be a fourth PR so they can be reviewed more carefully by model authors and backported to a bugfix release.

adamjstewart (commented):
Smaller, more focused PRs make things easier to review

adamjstewart:
@wangyi111

tests/models/conftest.py
isaaccorley: bfloat16 is faster and also reduces risk of underflow/overflow compared to float16

torchgeo/models/copernicusfm.py
```diff
@@ -115,7 +118,7 @@ def calculate_pinv(
     ) -> Tensor:
         mat = []
         for i in range(np.prod(old_shape)):
-            basis_vec = torch.zeros(old_shape)
+            basis_vec = torch.zeros(old_shape, dtype=torch.float32)
```
isaaccorley, line 121: torch.linalg.pinv doesn't have a half precision implementation so keep it in float32 then cast back to whatever dtype after the operation.

calebrob6 (approved):

calebrob6 (approved):

wangyi111:
just get email notified.. but good👍

Issue #2905, opened 2025-07-30
pip -> uv


adamjstewart:
### Summary

We should consider switching from pip to uv.

### Rationale

We already switched to ruff, uv is made by the same people and is allegedly much faster.

### Implementation

There are a few places we should switch from pip to uv:

- [ ] `.github/workflows/*.yaml`
- [ ] `.devcontainer/devcontainer.json`?
- [ ] `docs/tutorials/*.ipynb`?

There are also a few places worth mentioning both pip and uv:

- [ ] `docs/user/installation.rst`
- [ ] `README.md`?
- [ ] `torchgeo/datasets/utils.py`?

These could each be separate PRs.

### Alternatives

See the following documentation for suggestions on integration:

* https://docs.astral.sh/uv/guides/integration/github/
* https://docs.astral.sh/uv/guides/integration/dependency-bots/

### Additional information

I'm most interested in if uv is actually faster than pip for our specific use cases. We should benchmark speeds before and after the change, excluding caching. I'm also curious if uv is any better at caching than pip is.

amruthjakku:
Hi @adamjstewart , I’d like to take this issue as a first-time contributor. I’ll start by replacing pip with uv in the workflows and documentation as outlined. Let me know if that’s okay!

adamjstewart:
Please do! I think separate PRs for each would be good.

orbin123:
Thankyou for your feedback. my lap was broke that is why I couldn't respond. I will make the seperate PRs.

orbin123:
Sir, even if I try uv packages are showing dependency issues

adamjstewart:
What kind of dependency issues?

orbin123:
from pydantic import Schema
ImportError: cannot import name 'Schema' from 'pydantic' (/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/pydantic/__init__.cpython-311-x86_64-linux-gnu.so)

adamjstewart:
We'll drop 3.11 support soon, so that issue may go away. Otherwise I would need to see the full stacktrace to know which dependency requires older pydantic. I'm guessing this is only an issue in the minimum tests?

kshitijrajsharma:
I agree ! uv works perfectly with existing setup you have from the .toml ! all you need is CI to use uv ! I tested uv setup while working on this [PR](https://github.com/torchgeo/torchgeo/pull/3309/)  ! and uv supports switching python version like its piece of cake !

[...]

PR #3309, opened 2026-01-21
Add Dataset : OpenAerialMap

### LINKED ISSUE #3287: Add OpenAerialMap as new Dataset
### Summary

OpenAerialMap has quite a lot of crowdsource drone images ( CC by 4.0)  , There is a new STAC endpoint available , it would be nice to have it as dataset in torchgeo which will let people train geoAI models easily and builld a dataset. I will be happy to raise PR for this ! 

https://openaerialmap.org/ 
https://api.imagery.hotosm.org/ 

### Rationale

To easily train AI models based on openstreetmap data as labels and images from openaerialmap as OAM is heavily used during the disaster and lot of people contribute their drone images or high resolution image from companies donated for humanitarian purpose and primarily being used in tasking manager to create openstreetmap data which can be used as labels for tochgeo models ! 

### Implementation

Using TMS endpoint , similar to OpenStreetMap dataset in torchgeo takes bbox as input , queries the STAC get the features available in OAM and get the TMS of the drone image , based on the zoom level specified we fetch the relevant tile directly from the TMS and save it as tiff and save it in the folder that can later combined with openstreetmap for labels or others dataset ! Can be further extended to raster dataset ! 

### Alternatives

we can also download the geotiff directly of the drone image but that would be heavy i would rather follow the same dataset structure openstreetmap has ! 

### Additional information

https://openaerialmap.org/ 
https://api.imagery.hotosm.org/ 
 Here is a standalone implementation : https://github.com/kshitijrajsharma/geoml-toolkits/blob/master/src/geomltoolkits/downloader/tms.py

kshitijrajsharma:
### What does this PR do ? 

- Resolves #3287 

Adds an [OpenAerialMap](https://openaerialmap.org/) dataset that discovers imagery via OAM's STAC search API and downloads georeferenced tiles . Mirrors the OpenStreetMap dataset structure so OAM and OSM can be used in parallel.

### Implementation 
- Query: call OAM STAC search endpoint to get imagery for a bbox (or pick latest).
- Tiling: compute tiles with `mercantile.tiles(*bbox, zoom, truncate=True)`. I intentionally kept truncate to true to make sure tiles are returned only inside the bounding box 
- Sizes: supports 256 and 512 tiles; lower zooms are resampled by the tiler.
- Download & georef: download TMS tiles, georeference using mercantile bounds and save as `OAM-{x}-{y}-{z}.tif.`

### Usage 
```python
from torchgeo.datasets import OpenAerialMap
ds = OpenAerialMap(root="data/oam", bbox=(minx, miny, maxx, maxy), zoom=18, size=256)

```

I am aware of the STAC implementation : we can switch to a STAC baseclass / shared STAC client when available to torchgeo , I will update it after it is available , rest of the logic remains the same ! 

### Related resources : 

- Here is the sample my  low quality dummy model notebook to use this dataset in conjunction with OpenStreetMap : https://gist.github.com/kshitijrajsharma/20250180cc67edee6326597aabea7956 

I have tired my best to review this guideline : https://torchgeo.readthedocs.io/en/stable/user/contributing.html , Your comments and reviews are valuable and welcome ! 

### Test 

For testing , You can use this Drone image : https://map.openaerialmap.org/#/85.528564453125,27.635025869728384,12/square/12313122310120/62d86c65d8499800053796c4?_k=zd2ks6 

image_id is = `62d86c65d8499800053796c4`

Mapping in OSM was done using this drone image so you can build your dataset accurately 

for my testing I use this bounding box : `BBOX = [85.514668, 27.628367, 85.528875, 27.638514]`

You can also get it from the UI of OpenAerialMap or from the search_items from following bbox as I implemented in torchgeo. 

If there are multiple images available in same area then you might need to filter which image you wanna use , by default it will use the recent one, but that might not always match the labels in OSM or of low resolution , always visually check if required in openaerialmap

Example of dataset from `dataset.plot(sample)` 

image

## Summary by CodeRabbit

* **New Features**
  * Added OpenAerialMap dataset: downloadable tile-based aerial imagery with configurable bbox, zoom, tile sizing, on-disk georeferencing, and plotting.

* **Documentation**
  * Added API docs for the OpenAerialMap dataset.

* **Tests**
  * Added comprehensive tests covering tile math, grid generation, download flows, georeferencing, error handling, and plotting, plus a test utility that generates dummy GeoTIFF tiles.


adamjstewart:
Tagging @ashnair1 to review since it's currently similar to OSM, but ideally we would have a new STACDataset base class that handles all complexity for most STAC-isms and only a small subclass for OAM.

kshitijrajsharma:
Thanks @adamjstewart  , wow , you are very efficient to review  , let me update the PR documentation , fix the test cases and make the PR ready and yes you are right , I actually looked at current openstreetmap implementation and followed similar structure ! OAM and OSM will go parallel most of the time by the nature of those two dataset, I fully agree with you when STAC implementation is there  we can simply import STAC as baseclass and remove the API call from here ! Rest can remain the same ! I will continue contributing for sure to update this after STAC implementation !

kshitijrajsharma:
@adamjstewart  on the documentation it says : 
>All files with a `.txt` extension are managed by dependabot and should not be manually edited unless you need to add or remove a dependency.

Which means I am not supposed to change the requirements for dataset.txt  , but when I look at test cases CI it is using that .txt so how would I make the github action run with the new deps I added in .toml ? I added mercantile !

kshitijrajsharma:
I use [uv](https://docs.astral.sh/uv/getting-started/installation/)  most of the time , it is great for deps resolution and saves heck of time , well maintained , great community.  I think on your test CI , since you already have .toml which is compatible with uv already , I used uv for this PR as well to setup the env , you can easily generate requirements.txt within the CI from .toml and then test it for those PR which involves the deps changes , uv doesn't have to be in the production or in your workflows rightaway ( you can continue using what you have , unless need for change ) it can only be used for testing CI for this kind of PR , that would save you hassle , I can work on the testing CI if you think it is worth doing ( aah i see there is already PR drafted by another contributor , so i will wait ) !  
- I see you already have the issue : #2905

kshitijrajsharma:
This PR is ready from my side  ! cc : @ashnair1 @adamjstewart

docs/api/datasets/geo_datasets.csv
```diff
@@ -29,6 +29,7 @@ Dataset,Type,Source,License,Size (px),Resolution (m)
 `NCCM`_,Masks,Sentinel-2,"CC-BY-4.0",-,10
 `NLCD`_,Masks,Landsat,"public domain",-,30
 `Open Buildings`_,Geometries,"Maxar, CNES/Airbus","CC-BY-4.0 OR ODbL-1.0",-,-
+`OpenAerialMap`_,Imagery,OpenAerialMap,CC-BY-4.0,-,0.03--50
```
adamjstewart: Confirmed the license: https://openaerialmap.org/legal/

tests/datasets/test_openaerialmap.py
```diff
@@ -0,0 +1,319 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+
+
+import pytest
+
+try:
+    import mercantile
+except ImportError:
+    pytest.skip('mercantile is not installed', allow_module_level=True)
```
adamjstewart: This should use `pytest.importorskip`

tests/datasets/test_openaerialmap.py
```diff
@@ -0,0 +1,319 @@
...
+
+import asyncio
+import os
+import shutil
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock, patch
```
adamjstewart: We use pytest MonkeyPatch instead of unittest

tests/datasets/test_openaerialmap.py
```diff
@@ -0,0 +1,319 @@
...
+                tile = mercantile.Tile(x=1, y=1, z=1)
+
+                with patch.object(dataset, '_georeference_tile') as mock_geo:
+                    await dataset._download_tiles_async(
+                        'http://tms/{z}/{x}/{y}', [tile]
+                    )
+                    assert mock_geo.called
+                    assert (tmp_path / 'OAM-1-1-1.tif').exists()
+
+        asyncio.run(wrapper())
```
adamjstewart: Haven't played around with asyncio before. What is the advantage of this?

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,425 @@
...
+"""OpenAerialMap dataset."""
+
+import asyncio
+import os
+import warnings
+from collections.abc import Callable, Iterable
+from concurrent.futures import ThreadPoolExecutor
+from typing import Any, ClassVar, cast
+
+import aiohttp
```
adamjstewart: We don't currently have a direct dependency on aiohttp, is it possible to do this using a different library?

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,425 @@
...
+import matplotlib.pyplot as plt
+import pandas as pd
+import rasterio
+import requests
```
adamjstewart, line 17: This is currently an optional dependency, although I wouldn't mind promoting it to required

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,425 @@
...
+from matplotlib.figure import Figure
+from pyproj import CRS
+from rasterio.crs import CRS as RioCRS
+from rasterio.transform import from_bounds
+from tqdm.asyncio import tqdm_asyncio
```
adamjstewart: We don't currently have a direct dependency on tqdm, is this required?

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,425 @@
...
+    * https://openaerialmap.org/
+    """
+
+    _stac_api_url: ClassVar[str] = 'https://api.imagery.hotosm.org/stac'
+    _tile_source_url: ClassVar[str] = (
+        'https://titiler.hotosm.org/cog/tiles/WebMercatorQuad/{z}/{x}/{y}@{scale}x?url={source}'
+    )
+
+    filename_glob = 'OAM-*.tif'
+    filename_regex = r'^OAM-.*\.tif$'
```
adamjstewart: Regex isn't needed if the filename doesn't include things like timestamp or band number.

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,425 @@
...
+        res: float | tuple[float, float] | None = None,
+        bbox: tuple[float, float, float, float] | None = None,
+        zoom: int = 19,
+        max_items: int = 1,
+        transforms: Callable[[Sample], Sample] | None = None,
+        cache: bool = True,
+        download: bool = False,
+        search: bool = False,
+        image_id: str | None = None,
+        tile_size: int = 256,
```
adamjstewart: If there are only specific accepted values, you can use:

```suggestion
        tile_size: Literal[256, 512] = 256,
```

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,425 @@
...
+                    tile_z=str(tile.z),
+                )
+        except rasterio.errors.RasterioIOError:
+            warnings.warn(
+                f'Could not georeference {filepath}. Not a valid raster file.',
+                UserWarning,
+                stacklevel=2,
+            )
+
+    def plot(
```
adamjstewart, line 472: Can you add an example figure to your PR description so we can see what the plots look like with real data?

pyproject.toml
```diff
@@ -95,6 +95,8 @@ datasets = [
     "h5py>=3.10",
     # laspy 2.5.3+ required for wheels
     "laspy>=2.5.3",
+    # mercantile 1.2+ required for OpenAerialMap
+    "mercantile>=1.2.1",
```
adamjstewart: My biggest concern at the moment is new dependencies (aiohttp, mercantile, requests, tqdm). We try really hard to make it as easy as possible to install TorchGeo. If every PR adds 3 new required deps and 1 new optional dep, it will be impossible to install TorchGeo after a year. If it's at all possible to get this dataset working without these new dependencies, that would make this easier to merge.

kshitijrajsharma: Thanks @adamjstewart  I fully understand ! Only [mercantile](https://pypi.org/project/mercantile/) is needed for the web tile calculation to be able to fetch tiles as 256/256 from the drone images ! its kinda lite library , rest I think I can avoid !

adamjstewart: What math is mercantile doing that we can't do ourselves with pyproj?

kshitijrajsharma: its to get the xyz webmercator tile id's based within bbox , i think I can get the function for the tile computation ! we should be able to write it !

kshitijrajsharma: this math should work : https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames

torchgeo/datasets/openaerialmap.py
kshitijrajsharma: its not required for the core functionality of the dataset , its mainly for people to see the progress , i saw some notebooks in torchgeo docs that uses tqdm hence i added it ! you have indirect deps on tqdm , I can remove the tqdm as its not hard requirement

torchgeo/datasets/openaerialmap.py
kshitijrajsharma: aah thanks

tests/datasets/test_openaerialmap.py
kshitijrajsharma: asyncio lets me download multiple tiles ( I/O ) asynchronously on single thread , for eg when i compute within bbox i have 500 tiles to download then with asyncio await and run i would be able to download multiple tiles at a same time without waiting for network request to finish !

torchgeo/datasets/openaerialmap.py
kshitijrajsharma: Yes , i can use asyncio threads ! Thanks

tests/datasets/test_openaerialmap.py
kshitijrajsharma: okay !

torchgeo/datasets/openaerialmap.py
kshitijrajsharma: Yes

pyproject.toml
kshitijrajsharma: Okay , Thanks @adamjstewart for your beautiful review !  tqdm is out , aiohttp is out as well , I guess request : you would need for upcoming STAC perhaps and i am not sure how to avoid  , I will look into replacement for mercantile if I can !

tests/datasets/test_openaerialmap.py
adamjstewart: Doesn't sound necessary for our unit tests...

kshitijrajsharma: without asyncio.run how would you recommend to add test case for download tiles async function ? should i remove this test function ?

kshitijrajsharma:
I have resolved the comments ! Except  for the replacement of mercantile , I will do debugging a little if we can avoid mercantile ( Kindly let me know if you have suggestions ) , I have removed rest of the deps ! 

Update : I have removed mercantile as deps as well so no more deps changes !

tests/datasets/test_openaerialmap.py
adamjstewart: No idea, maybe @ashnair1 has an idea.

pyproject.toml
kshitijrajsharma: I have added the necessary tile computation logic from mercantile !

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,501 @@
...
+from pyproj import CRS
+from rasterio.crs import CRS as RioCRS
+from rasterio.transform import from_bounds
+
+from .geo import RasterDataset
+from .utils import Path, Sample
+
+
+class TileUtils:
+    """Compact mercantile-compatible tile utilities. This is to avoid deps on mercantile as we are only using few functions. Credit of this code block goes to contributors of mercantile. Source & Cite : # https://github.com/mapbox/mercantile/blob/5975e1c0e1ec58e99f8e5770c975796e44d96b53/mercantile/__init__.py."""
```
ashnair1: This line is too long. Could you wrap it like this?

```suggestion                                                                                                                                                          
class TileUtils:                                                                                                                                                       
    """Compact mercantile-compatible tile utilities.                                                                                                                   
                                                                                                                                                                       
    This is to avoid dependencies on mercantile as we are only using a few                                                                                             
    functions. Credit for this code block goes to contributors of mercantile.                                                                                          
                                                                                                                                                                       
    Source: https://github.com/mapbox/mercantile/blob/5975e1c0e1ec58e99f8e5770c975796e44d96b53/mercantile/__init__.py                                                  
    """                                                                                                                                                                
```

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,501 @@
...
+        Args:
+            paths: one or more root directories to search or files to load
+            crs: :term:`coordinate reference system (CRS)` to warp to
+                (defaults to EPSG:3857)
+            res: resolution of the dataset in units of CRS
+                (defaults to resolution of first file found)
+            bbox: bounding box for STAC query as (xmin, ymin, xmax, ymax) in EPSG:4326.
+                Same format as OpenStreetMap for easy dataset combination.
+            zoom: zoom level for tiles (1-22), only used when download=True.
+                Higher zoom = more detail. Typical values: 18-20 for high-res drone imagery, higher the zoom , higher the resolution on same tile but it would be more zoomed in. Usually it is advised to increase the tilesize if you want better quality at same zoom. it is advised to check gsd of the image itself before trying to download it.
```
ashnair1: Docstring is too long for `zoom`. Needs wrapping 

```suggestion                                                                                                                                                          
            zoom: zoom level for tiles (6-22), only used when download=True.                                                                                           
                Higher zoom = more detail. Typical values: 18-20 for high-res                                                                                          
                drone imagery. Higher zoom gives higher resolution but covers                                                                                          
                less area per tile. Consider increasing tile_size for better                                                                                           
                quality at the same zoom. Check the GSD of the image before                                                                                            
                downloading.                                                                                                                                           
```

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,501 @@
...
+
+        if self.bbox is None:
+            warnings.warn(
+                'Bounding box (bbox) is required to calculate tiles, even when image_id is provided.',
+                UserWarning,
+                stacklevel=2,
+            )
+            return
+
+        # we use truncate=True to avoid tiles outside the bbox , just to make sure there won't be corning tiles
```
ashnair1:
```suggestion
        # we use truncate=True to avoid tiles outside the bbox, just to make sure there won't be corner tiles
```

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,501 @@
...
+
+        return self._tile_source_url.format(
+            z='{z}',
+            x='{x}',
+            y='{y}',
+            scale=int(self.tile_size / 256),
+            source=visual_source,
+        )
+
+    async def _download_tiles_async(self, tms_url: str, tiles: list[Any]) -> None:
```
ashnair1: `list[TileUtils.Tile]` would be a more accurate type-hint than `list[Any]`

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,501 @@
...
+            tiles: List of mercantile tiles to download
+        """
+        tasks = [self._download_single_tile(tms_url, tile) for tile in tiles]
+        total = len(tasks)
+        print(f'Starting download of {total} tiles...')
+
+        for i, task in enumerate(asyncio.as_completed(tasks), 1):
+            await task
+
+    async def _download_single_tile(self, tms_url: str, tile: Any) -> None:
```
ashnair1:
```diff
-tile:Any
+tile:TileUtils.Tile
```

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,501 @@
...
+
+            with open(filepath, 'wb') as f:
+                f.write(response.content)
+
+            self._georeference_tile(filepath, tile)
+
+        except (requests.RequestException, OSError) as e:
+            warnings.warn(f'Error downloading tile {tile}: {e}', UserWarning)
+
+    def _georeference_tile(self, filepath: str, tile: Any) -> None:
```
ashnair1:
```diff
-tile:Any
+tile:TileUtils.Tile
```

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,501 @@
...
+from matplotlib.figure import Figure
+from pyproj import CRS
+from rasterio.crs import CRS as RioCRS
+from rasterio.transform import from_bounds
+
+from .geo import RasterDataset
+from .utils import Path, Sample
+
+
+class TileUtils:
```
ashnair1, line 26: Line 48 (case where `if abs(sinlat) >= 1.0`)  and line 58 (if truncate case in tile class method) are currently not covered by tests.

torchgeo/datasets/openaerialmap.py
adamjstewart: Mercantile is BSD-3-Clause, we can't copy any code from there to TorchGeo without changing our license to MIT AND BSD-3-Clause.

kshitijrajsharma: Aah you are right , I initially thought BSD and MIT would be compatible but i did bit digging , turns out we would need to partially change the license of this file right ? What do you suggest ? I can add attribution from bsd here on top of the file if that would be okay else , I would look for another path ?

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,501 @@
...
+        )
+
+    If you use this dataset in your research, please cite OpenAerialMap:
+
+    * https://openaerialmap.org/
+    """
+
+    _stac_api_url: ClassVar[str] = 'https://api.imagery.hotosm.org/stac'
+    _tile_source_url: ClassVar[str] = (
+        'https://titiler.hotosm.org/cog/tiles/WebMercatorQuad/{z}/{x}/{y}@{scale}x?url={source}'
```
rbavery: a recent issue on the hotosm page points to server burden.

https://github.com/hotosm/openaerialmap/issues/248

I don't think it is the correct path to add to Humanitarian OpenStreet Maps server burden by trying to use it's titiler service in a dataloader.

It will get used/misused by ML hobbyists or researchers without understanding the impact their requests are having on this nonprofit's infrastructure, which they rely on to support their mission.

instead, the open aerial map dataset could be hosted on torchgeo's own huggingface or other s3 compatible storage so it can be accessed without a server.

torchgeo/datasets/openaerialmap.py
adamjstewart: We don't want to change the license of TorchGeo. We either need to write the code from scratch or add the dependency.

torchgeo/datasets/openaerialmap.py
adamjstewart: TorchGeo is even less than a non-profit, we are running out of storage space on HF and I'm concerned about our ability to keep unofficial copies synced with upstream. I also don't want to steal credit from the HOSM team.

tests/datasets/test_openaerialmap.py
ashnair1: > asyncio lets me download multiple tiles ( I/O ) asynchronously on single thread , for eg when i compute within bbox i have 500 tiles to download then with asyncio await and run i would be able to download multiple tiles at a same time without waiting for network request to finish !

That isn't what's happening here. You're using `requests` which is blocking. To do this truly async, you'd need to use `aiohttp` or `httpx` which is async aware. The concurrency comes from `asyncio.to_thread(...)`, i.e., running downloads in a thread pool while `asyncio` coordinates them.

So you don't really need `asyncio` at all here. Just use a `ThreadPoolExecutor` directly for downloading tiles.

kshitijrajsharma: Thanks, @ashnair1 !   I was using aiohttp before , but in a purpose of not adding new deps I have removed it, Now I am using asyncio.to_thread now like you have suggested ; I wrapped requests inside the asyncio_to_thread so it should enable the parallel download !

torchgeo/datasets/openaerialmap.py
kshitijrajsharma: okay ! Thanks @adamjstewart  I will work on the custom implementation !

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,505 @@
...
+            response = await asyncio.to_thread(requests.get, url, timeout=30)
+            if response.status_code != 200:
+                warnings.warn(
+                    f'Failed to download tile {tile}: HTTP {response.status_code}',
+                    UserWarning,
+                )
+                return
+
+            with open(filepath, 'wb') as f:
+                f.write(response.content)
```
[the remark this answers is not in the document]
kshitijrajsharma: fixes by [50e1ba3](https://github.com/torchgeo/torchgeo/pull/3309/commits/50e1ba37bd47ea96944d415354d3230599ceffad)

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,505 @@
...
+            params['bbox'] = list(self.bbox)
+
+        try:
+            response = requests.post(
+                f'{self._stac_api_url}/search', json=params, timeout=30
+            )
+            response.raise_for_status()
+            data = response.json()
+        except Exception as e:
+            raise RuntimeError(f'Failed to query STAC API: {e}') from e
```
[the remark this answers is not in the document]
kshitijrajsharma: fixes by [50e1ba3](https://github.com/torchgeo/torchgeo/pull/3309/commits/50e1ba37bd47ea96944d415354d3230599ceffad), i am checking the json response now

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,505 @@
...
+    -------------
+
+    The dataset can be used in two modes:
+
+    1. **Search mode**: Query available imagery in a bbox.
+    2. **Download mode**: Query STAC API and download TMS tiles.
+
+    For search mode::
+
+        oam = OpenAerialMap(bbox=bbox, search=True)
```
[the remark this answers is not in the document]
kshitijrajsharma: i don't think this is needed, it is kinda straightforward !

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,505 @@
...
+            loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
+            try:
+                loop.run_until_complete(self._download_tiles_async(tms_url, tiles))
+            finally:
+                loop.close()
+
+        with ThreadPoolExecutor() as executor:
+            executor.submit(run_in_thread).result()
+
```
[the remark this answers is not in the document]
kshitijrajsharma: nice catch , fixed by [50e1ba3](https://github.com/torchgeo/torchgeo/pull/3309/commits/50e1ba37bd47ea96944d415354d3230599ceffad)

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,505 @@
...
+                f.write('Download attempted but no TMS URLs found.\n')
+            return
+
+        if self.bbox is None:
+            warnings.warn(
+                'Bounding box (bbox) is required to calculate tiles, even when image_id is provided.',
+                UserWarning,
+                stacklevel=2,
+            )
+            return
```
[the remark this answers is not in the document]
kshitijrajsharma: fixed by [50e1ba3](https://github.com/torchgeo/torchgeo/pull/3309/commits/50e1ba37bd47ea96944d415354d3230599ceffad)

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,505 @@
...
+from rasterio.crs import CRS as RioCRS
+from rasterio.transform import from_bounds
+
+from .geo import RasterDataset
+from .utils import Path, Sample
+
+
+class TileUtils:
+    """Compact mercantile-compatible tile utilities. This is to avoid deps on mercantile as we are only using few functions. Credit of this code block goes to contributors of mercantile. Source & Cite : # https://github.com/mapbox/mercantile/blob/5975e1c0e1ec58e99f8e5770c975796e44d96b53/mercantile/__init__.py."""
+
```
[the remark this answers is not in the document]
kshitijrajsharma: outdated by [50e1ba3](https://github.com/torchgeo/torchgeo/pull/3309/commits/50e1ba37bd47ea96944d415354d3230599ceffad) as now we moved to custom implementation !

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,505 @@
...
+        if not visual_source:
+            return None
+
+        return self._tile_source_url.format(
+            z='{z}',
+            x='{x}',
+            y='{y}',
+            scale=int(self.tile_size / 256),
+            source=visual_source,
+        )
```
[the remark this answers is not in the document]
kshitijrajsharma: not needed ! if visual source is none , function gonna return anyway !

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,505 @@
...
+            DatasetNotFoundError: If dataset is not found and download=False.
+            ValueError: If download=True but bbox is not provided.
+        """
+        self.paths = paths
+        self.bbox = bbox
+        self.zoom = zoom
+        self.max_items = max_items
+        self.download = download
+        self.image_id = image_id
+        self.search_results: pd.DataFrame | None = None
```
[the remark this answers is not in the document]
kshitijrajsharma: no need ,there is static type check directly in the param Literal [256,512,768,1024]  !

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,505 @@
...
+            cache: if True, cache file handle to speed up repeated sampling
+            download: if True, download imagery from STAC API based on bbox
+            search: if True, query STAC API for available imagery and return results in
+                self.search_results. Skips dataset initialization if download=False.
+            image_id: optional STAC item ID to download specific imagery
+            tile_size: size of the tiles to download (supported : 256 , 512, 768, 1024 ; Do verify they exists in the remote image)
+
+        Raises:
+            DatasetNotFoundError: If dataset is not found and download=False.
+            ValueError: If download=True but bbox is not provided.
```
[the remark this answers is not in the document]
kshitijrajsharma: addressed in [50e1ba3](https://github.com/torchgeo/torchgeo/pull/3309/commits/50e1ba37bd47ea96944d415354d3230599ceffad)

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,505 @@
...
+            res: resolution of the dataset in units of CRS
+                (defaults to resolution of first file found)
+            bbox: bounding box for STAC query as (xmin, ymin, xmax, ymax) in EPSG:4326.
+                Same format as OpenStreetMap for easy dataset combination.
+            zoom: zoom level for tiles (6-22), only used when download=True.                                                                                           
+                Higher zoom = more detail. Typical values: 18-20 for high-res                                                                                          
+                drone imagery. Higher zoom gives higher resolution but covers                                                                                          
+                less area per tile. Consider increasing tile_size for better                                                                                           
+                quality at the same zoom. Check the GSD of the image before                                                                                            
+                downloading.
```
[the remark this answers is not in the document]
kshitijrajsharma: outdated

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,505 @@
...
+        image = sample['image']
+        #  C, H, W -> H, W, C
+        rgb = image[0:3, :, :].permute(1, 2, 0)
+
+        if rgb.is_floating_point() and rgb.max() > 1:
+            rgb = rgb / 255.0
+            rgb = rgb.clamp(0, 1)
+
+        fig, ax = plt.subplots(nrows=1, ncols=1, figsize=(4, 4))
+        ax.imshow(rgb, cmap='gray' if rgb.ndim == 2 else None)
```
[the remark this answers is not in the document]
kshitijrajsharma: fixes by :  [8b4dc05](https://github.com/torchgeo/torchgeo/pull/3309/commits/8b4dc059d603c08950733ae73ee6882911accc26)

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,505 @@
...
+            tms_url: TMS URL template with {z}, {x}, {y} placeholders
+            tiles: List of mercantile tiles to download
+        """
+        tasks = [self._download_single_tile(tms_url, tile) for tile in tiles]
+        total = len(tasks)
+        print(f'Starting download of {total} tiles...')
+
+        for i, task in enumerate(asyncio.as_completed(tasks), 1):
+            await task
+
```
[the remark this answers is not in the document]
kshitijrajsharma: fixes by [8b4dc05](https://github.com/torchgeo/torchgeo/pull/3309/commits/8b4dc059d603c08950733ae73ee6882911accc26)

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,505 @@
...
+            paths, crs or CRS.from_epsg(3857), res, transforms=transforms, cache=cache
+        )
+
+    def _search_stac(self) -> None:
+        """Query and display available imagery as a DataFrame."""
+        assert self.bbox is not None
+        try:
+            resp = requests.post(
+                f'{self._stac_api_url}/search',
+                json={'bbox': list(self.bbox), 'limit': max(self.max_items, 20)},
```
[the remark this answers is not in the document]
kshitijrajsharma: nice , fixes by : [8b4dc05](https://github.com/torchgeo/torchgeo/pull/3309/commits/8b4dc059d603c08950733ae73ee6882911accc26)

torchgeo/datasets/openaerialmap.py
kshitijrajsharma: it's a valid concern @rbavery  thanks , I have added `user-agent`  [67fafab](https://github.com/torchgeo/torchgeo/pull/3309/commits/67fafab62a936e29f0194da975969a87b1a6cc84) on the requests that is going out from torchgeo for HOTOSM team to be able to track when needed , if it gets overwhelming then perhaps we can look for alternatives.

spwoodcock: Really appreciate you trying to keep our costs down @rbavery 🙏

The issue you link to was actually more hypothetical. Federation of our catalogue **would** provide the benefit of load reduction, but that isn't to say the current load is unsustainable.

We are in the process of tweaking our autoscaling capabilities to better handle load peaks, but overall I still think we can use the catalogue.

@kshitijrajsharma is currently tweaking things to use a different production endpoint (and perhaps a different zoom threshold), but overall I'm happy for this process to go ahead 😄 

And I agree @adamjstewart that it's not realistic for you to keep a mirror in sync with our catalogue 👍

kshitijrajsharma: Thanks @spwoodcock  , I have switched the tiles endpoint to : https://api.imagery.hotosm.org/raster/api.html#/ , also placed limit on zoom level from 15-23

tests/datasets/test_openaerialmap.py
ashnair1: That's fine but `asyncio.to_thread()` uses a `ThreadPoolExecutor` internally. So this creates an event loop just to schedule work onto another thread pool. The event loop does no async I/O so it's just coordination overhead.

This could be a simpler alternative without using `asyncio`.                                                                                                                          
                                                                                                                                                                         
```python                                                                                                                                                 
from concurrent.futures import ThreadPoolExecutor, as_completed                                                                                                        
                                                                                                                                                                       
def _download_tiles(self, tiles_url: str, tiles: list[TileUtils.Tile]) -> None:                                                                                        
    print(f'Starting download of {len(tiles)} tiles...')                                                                                                               
    with ThreadPoolExecutor() as executor:                                                                                                                             
        futures = {                                                                                                                                                    
            executor.submit(self._download_single_tile, tiles_url, tile): tile                                                                                         
            for tile in tiles                                                                                                                                          
        }                                                                                                                                                              
        for future in as_completed(futures):                                                                                                                           
            future.result()                                                                                                                                            
    print(f'Finished downloading {len(tiles)} tiles.')                                                                                                                 
                                                                                                                                                                       
def _download_single_tile(self, tiles_url: str, tile: TileUtils.Tile) -> None:                                                                                         
    # Same logic, just synchronous — no async/await needed                                                                                                             
    response = requests.get(url, headers={'User-Agent': 'torchgeo'}, timeout=30)                                                                                       
    # ... rest of the method    
```

spwoodcock: Yeah true, if `aiohttp` was used asyncio would make sense, but with `requests` sync code in a thread pool is best 👍

tests/datasets/test_openaerialmap.py
```diff
@@ -0,0 +1,442 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+
+import os
+import shutil
+from pathlib import Path
+from typing import Any
+from unittest.mock import MagicMock
```
adamjstewart: Can we use pytest.MonkeyPatch instead of unittest.mock.MagicMock?

tests/data/openaerialmap/data.py
```diff
@@ -0,0 +1,48 @@
...
+    os.makedirs(root, exist_ok=True)
+
+    # Using roughly the area around Banepa, Nepal
+    minx, miny, maxx, maxy = 85.516, 27.631, 85.523, 27.637
+
+    for x, y, z in tiles:
+        path = os.path.join(root, f'OAM-{x}-{y}-{z}.tif')
+
+        # Create a 3-channel dummy image (RGB)
+        height, width = 256, 256
```
adamjstewart, line 27: Can we make the test data smaller, maybe 36x36? This makes git clone a little faster.

EDIT: depends on the answer to my question about `tile_size` only supporting 256 and 512.

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,565 @@
...
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from typing import Any, ClassVar, Literal, cast
+
+import matplotlib.pyplot as plt
+import pandas as pd
+import rasterio
+import requests
+from matplotlib.figure import Figure
+from pyproj import CRS
+from rasterio.crs import CRS as RioCRS
```
adamjstewart: A bit nervous about mixing and matching these. So far, rasterio seems to be 99% compatible with pyproj CRS, so we may be able to get away with only using pyproj. Of course, once rasterio adds type hints, it will start to complain, but this is true for all other datasets too.

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,565 @@
...
+        search: bool = False,
+        image_id: str | None = None,
+        tile_size: Literal[256, 512] = 256,
+    ) -> None:
+        """Initialize a new OpenAerialMap dataset instance.
+
+        Args:
+            paths: one or more root directories to search or files to load
+            crs: :term:`coordinate reference system (CRS)` to warp to
+                (defaults to EPSG:3857)
```
adamjstewart: The default CRS and the bbox don't use the same CRS? Can we make them match?

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,565 @@
...
+            res: resolution of the dataset in units of CRS
+                (defaults to resolution of first file found)
+            bbox: bounding box for STAC query as (xmin, ymin, xmax, ymax) in EPSG:4326.
+                Same format as OpenStreetMap for easy dataset combination.
```
adamjstewart: All GeoDatasets can be automatically combined, and CRS/res differences are handled behind the scenes. So we don't need to take other datasets into account when thinking about this one.

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,565 @@
...
+            transforms: a function/transform that takes an input sample
+                and returns a transformed version. Note: CRS transformation is handled
+                automatically via the crs parameter.
+            cache: if True, cache file handle to speed up repeated sampling
+            download: if True, download imagery from STAC API based on bbox
+            search: if True, query STAC API for available imagery and return results in
+                self.search_results. Skips dataset initialization if download=False.
+            image_id: optional STAC item ID to download specific imagery
+            tile_size: size of the tiles to download (supported : 256 , 512 );
+                Do verify they exists in the remote image source.
```
adamjstewart: Only 256 and 512 are supported? Is this an API limitation, or something you imposed?

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,565 @@
...
+                less area per tile. Consider increasing tile_size for better
+                quality at the same zoom. Check the GSD of the image before
+                downloading.
+            max_items: maximum number of STAC items to query.
+            transforms: a function/transform that takes an input sample
+                and returns a transformed version. Note: CRS transformation is handled
+                automatically via the crs parameter.
+            cache: if True, cache file handle to speed up repeated sampling
+            download: if True, download imagery from STAC API based on bbox
+            search: if True, query STAC API for available imagery and return results in
```
adamjstewart: None of our other datasets have this feature. It's nice for debugging, but I'm not sure if it will confuse the average user...

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,565 @@
...
+            zoom: zoom level for tiles (15-23), only used when download=True.
+                Higher zoom = more detail. Typical values: 18-20 for high-res
+                drone imagery. Higher zoom gives higher resolution but covers
+                less area per tile. Consider increasing tile_size for better
+                quality at the same zoom. Check the GSD of the image before
+                downloading.
+            max_items: maximum number of STAC items to query.
+            transforms: a function/transform that takes an input sample
+                and returns a transformed version. Note: CRS transformation is handled
+                automatically via the crs parameter.
```
adamjstewart: Not sure if this is worth explaining here, this is true for every GeoDataset in TorchGeo. Maybe we should add this to https://torchgeo.readthedocs.io/en/stable/user/glossary.html in a separate PR.

adamjstewart (commented):
Still relying on @ashnair1 for the low-level review.

Just adding some high-level comments on the user-facing API changes and imports.

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,565 @@
...
+            paths='data/openaerial',
+            bbox=(lon_min, lat_min, lon_max, lat_max),
+            zoom=19,
+            download=True,
+        )
+
+    If you use this dataset in your research, please cite OpenAerialMap:
+
+    * https://openaerialmap.org/
+    """
```
ashnair1, line 195: Needs `..versionadded` tag

```suggestion
    .. versionadded:: 0.9
    """
```

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,565 @@
...
+        if not result:
+            warnings.warn(
+                f'No imagery found for bbox {self.bbox} or ID {self.image_id}. '
+                'Try a different area or check OpenAerialMap coverage.',
+                UserWarning,
+                stacklevel=2,
+            )
+            with open(os.path.join(root, '.downloaded'), 'w') as f:
+                f.write('Download attempted but no imagery found.\n')
+            return
```
ashnair1: What is the purpose of this `.downloaded` file? If query returns no imagery, this file is written. But this doesn't seem to be used anywhere.

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,565 @@
...
+        self.search_results: pd.DataFrame | None = None
+        self.tile_size = tile_size
+
+        if search:
+            if self.bbox is None:
+                raise ValueError('bbox must be provided when search=True')
+            self._search_stac()
+            # If user only wants to search, return early, because dataset will raise error if it is empty and is instantiated by super init
+            if not download:
+                return
```
ashnair1: So there are two flows in this dataset. A user searching for files in a given bbox and a user downloading available data in a given bbox. 

I'm not sure why this searching is required and especially why it's part of the dataset. Typical flow is 

```
1. User specifies a region/bbox
2. Dataset queries endpoint for that region
3. If data is available, it is downloaded
4. If no data is available, an error is raised saying that data is unavailable in that region
```

Additionally if `search=True` and `download=False`, `super().__init()` is never called and you're left with a broken class.

docs/api/datasets.rst
```diff
@@ -186,6 +186,11 @@ Open Buildings
 
 .. autoclass:: OpenBuildings
 
+OpenAerialMap
+^^^^^^^^^^^^^
+
+.. autoclass:: OpenAerialMap
+
```
ashnair1: We've changed how docs are structured and organized - Refer to #3385.

torchgeo/datasets/openaerialmap.py
kshitijrajsharma: Do you suggest to only use pyproj CRS ?

torchgeo/datasets/openaerialmap.py
kshitijrajsharma: More of API limitation , higher the image size slower the API gets so 256 512 is kinda sweet spot ! it starts from 256 though !

torchgeo/datasets/openaerialmap.py
kshitijrajsharma: okay , it was helpful for me while implementing so i thought it would be useful if there are multiple images available in the area, I can remove it !

torchgeo/datasets/openaerialmap.py
kshitijrajsharma: ok

torchgeo/datasets/openaerialmap.py
kshitijrajsharma: Removed search functionality !  I was mainly thinking to have search if there are multiple images available in the area so i can pick one ! by default it would pick the recent one , so it would be okay

torchgeo/datasets/openaerialmap.py
adamjstewart: Yes, as long as it works

torchgeo/datasets/openaerialmap.py
adamjstewart: So 256, 257, 258 are all valid? I would just support all ints then. You can add an assert for > 255 if you want.

kshitijrajsharma: aah no , it should be valid tile size so 256x256 , 512x512 , 768x768 , like this ! Hence I included only 256 and 512

torchgeo/datasets/openaerialmap.py
kshitijrajsharma: ok

torchgeo/datasets/openaerialmap.py
kshitijrajsharma: ok

ashnair1:
Test coverage is pending. Docs are failing due to missing `docs/api/datasets/openaerialmap.rst`. Content of file should look like this. 

```rst
.. _OpenAerialMap:

OpenAerialMap
=============

.. currentmodule:: torchgeo.datasets
.. autoclass:: OpenAerialMap
```

kshitijrajsharma:
Thanks for the detailed review , I have tried to resolve all the comments !

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,486 @@
...
+            bbox=(lon_min, lat_min, lon_max, lat_max),
+            zoom=19,
+            download=True,
+        )
+
+    If you use this dataset in your research, please cite OpenAerialMap:
+
+    * https://openaerialmap.org/
+
+    .. versionadded:: 0.9
```
ashnair1:
```suggestion
    .. versionadded:: 0.10
```
0.9 was recently released so this will be included in the next release

torchgeo/datasets/openaerialmap.py
ashnair1: @kshitijrajsharma 

Technically 256 (`@1x`), 512 (`@2x`), 768 (`@3x`) and 1024 (`@4x`) are supported. `@5x` and beyond are not. So you can update `tile_size` to support 768 and 1024 as well

Can be validated with this script

```python
import requests

# Known working tile from OpenAerialMap
URL = "https://api.imagery.hotosm.org/raster/collections/openaerialmap/items/6708341a4a0ab60001b0b94d/tiles/WebMercatorQuad/17/96629/55042"

for scale in range(1, 6):
    r = requests.get(f"{URL}@{scale}x?assets=visual", timeout=30)
    if r.status_code == 200:
        print(f"@{scale}x: HTTP 200 OK")
    else:
        msg = r.json()["detail"][0]["msg"]
        print(f"@{scale}x: HTTP {r.status_code} - {msg}")
```

Output
```bash
@1x: HTTP 200 OK
@2x: HTTP 200 OK
@3x: HTTP 200 OK
@4x: HTTP 200 OK
@5x: HTTP 422 - Input should be less than or equal to 4
```

ashnair1:
Only one minor comment pending from me - https://github.com/torchgeo/torchgeo/pull/3309#discussion_r2900248468. After that, PR is good to go.

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,489 @@
...
+
+            self._georeference_tile(filepath, tile)
+
+        except (requests.RequestException, OSError) as e:
+            warnings.warn(f'Error downloading tile {tile}: {e}', UserWarning)
+
+    def _georeference_tile(self, filepath: str, tile: TileUtils.Tile) -> None:
+        """Add georeferencing metadata to a downloaded tile.
+
+        This sets the CRS to EPSG:4326 because mercantile bounds are lat/lon.
```
calebrob6, line 439: mercantile isn't used anymore

calebrob6 (approved):
Read through this all and only noticed this (also I noted down to benchmark a couple of ways to write out the tiles we are loading from remote tileserver -- unsure if the current way is fastest, but we should do this after this is merged)

torchgeo/datasets/openaerialmap.py
kshitijrajsharma: Included

docs/api/datasets/images.csv
```diff
@@ -3,5 +3,6 @@ Dataset,Type,Source,Size (px),Resolution (m),License
 :ref:`EnMAP`,Imagery,EnMAP,"1,200x1,200",30,`EnMAP Data License <https://www.enmap.org/data/resources/EnMAP_Data_License.pdf>`_
 :ref:`Landsat`,Imagery,Landsat,"8,900x8,900",30,public domain
 :ref:`NAIP`,Imagery,Aerial,"6,100x7,600",0.3--2,public domain
+:ref:`OpenAerialMap`,Imagery,Aerial,"256x256 OR 512x512",0.03--50,CC-BY-4.0
```
adamjstewart, line 6: Confirmed the license: https://openaerialmap.org/legal/

tests/datasets/test_openaerialmap.py
```diff
@@ -0,0 +1,658 @@
...
+                raise post_exception('Fail')
+            return MockPostResponse()
+
+        monkeypatch.setattr('torchgeo.datasets.openaerialmap.requests.post', mock_post)
+        monkeypatch.setattr(
+            'torchgeo.datasets.openaerialmap.requests.get',
+            lambda *args, **kwargs: MockGetResponse(),
+        )
+
+        result = dataset._fetch_item_id()  # type: ignore[reportPrivateUsage]
```
adamjstewart: Can we remove these unused ignores?

tests/datasets/test_openaerialmap.py
```diff
@@ -0,0 +1,658 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+
+import os
+import shutil
+from pathlib import Path
+from typing import Any
```
adamjstewart: Would like to avoid Any if possible as it basically disables the type checker. Maybe we can use TypedDict here?

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,490 @@
...
+
+    If you use this dataset in your research, please cite OpenAerialMap:
+
+    * https://openaerialmap.org/
+
+    .. versionadded:: 0.10
+    """
+
+    _stac_url: ClassVar[str] = 'https://api.imagery.hotosm.org/stac'
+    _tiles_url: ClassVar[str] = 'https://api.imagery.hotosm.org/raster'
```
adamjstewart: ClassVar isn't needed here, str's are immutable:

```suggestion
    _stac_url = 'https://api.imagery.hotosm.org/stac'
    _tiles_url = 'https://api.imagery.hotosm.org/raster'
```

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,490 @@
...
+import requests
+from matplotlib.figure import Figure
+from pyproj import CRS
+from rasterio.transform import from_bounds
+
+from .geo import RasterDataset
+from .utils import Path, Sample
+
+
+class TileUtils:
```
adamjstewart, line 26: Class and namedtuples feel like overkill. This could be replaced by a couple helper functions. The more complexity we have, the more difficult it is to maintain.

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,490 @@
...
+
+    def _fetch_item_id(self) -> str | None:
+        """Query STAC API and extract tiles URL.
+
+        Combines logic for ID-based and BBox-based queries.
+
+        Returns:
+            tiles_url_template, or None if not found
+        """
+        params: dict[str, Any] = {'limit': self.max_items}
```
adamjstewart: Avoid Any. For example, this could be:
```python
from typing import NotRequired, TypedDict

class JSON(TypedDict):
    limit: int
    ids: NotRequired[list[str]]
    bbox: NotRequired[list[float]]

params: JSON = {'limit': self.max_items}
```
Could also skip type hints entirely on local params, I think they're only required for function signatures?

torchgeo/datasets/openaerialmap.py
```diff
@@ -0,0 +1,490 @@
...
+                raise ValueError('bbox must be provided when download=True')
+            if not 15 <= zoom <= 23:
+                raise ValueError(f'zoom must be between 15 and 23, got {zoom}')
+            self._download()
+            print('Download complete.')
+
+        if crs is None:
+            crs = CRS.from_epsg(4326)
+
+        super().__init__(paths, crs, res, transforms=transforms, cache=cache)
```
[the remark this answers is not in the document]
kshitijrajsharma: resolved by #[494d1c1](https://github.com/torchgeo/torchgeo/pull/3309/commits/494d1c124c8344cb0b0c98e946b315f11cfb4d2b)

tests/datasets/test_openaerialmap.py
kshitijrajsharma: Resolved by #[494d1c1](https://github.com/torchgeo/torchgeo/pull/3309/commits/494d1c124c8344cb0b0c98e946b315f11cfb4d2b)

torchgeo/datasets/openaerialmap.py
kshitijrajsharma: Resolved by [494d1c1](https://github.com/torchgeo/torchgeo/pull/3309/commits/494d1c124c8344cb0b0c98e946b315f11cfb4d2b)

tests/datasets/test_openaerialmap.py
kshitijrajsharma: Resolved by [2d351cf](https://github.com/torchgeo/torchgeo/pull/3309/commits/2d351cfc0463a3b7c59def3d046aed321472b93e)

PR #3311, opened 2026-01-21
Update pandas dependency

adamjstewart:
Hoping this fixes RtD CI. My current best guess is that these classes were moved internally, so although the classes and docs exist for both pandas 2 and 3, the stable docs (3) don't match the location of the tested API (2).

### DIFF
--- a/torchgeo/datasets/cropharvest.py
+++ b/torchgeo/datasets/cropharvest.py
@@ -230,7 +230,7 @@ class CropHarvest(NonGeoDataset):
         ]
         properties = row.to_dict(orient='records')[0]
         label = 'None'
-        if properties['properties.label']:
+        if isinstance(properties['properties.label'], str):
             label = properties['properties.label']
         elif properties['properties.is_crop'] == 1:
             label = 'Other'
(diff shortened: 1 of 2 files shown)


torchgeo/datasets/cropharvest.py
```diff
@@ -230,7 +230,7 @@ def _load_label(self, idx: str, dataset: str) -> Tensor:
         ]
         properties = row.to_dict(orient='records')[0]
         label = 'None'
-        if properties['properties.label']:
+        if isinstance(properties['properties.label'], str):
```
adamjstewart, line 233: Bug caused by new dedicated string data type.

In pandas 2, empty values are bool(None) == False. In pandas 3, empty values are bool(nan) == True.

Check for type instead of "truthiness" of value.

torchgeo/datasets/splits.py
```diff
@@ -165,8 +166,7 @@ def random_bbox_splitting(
                             new_geometry, geometry_remaining = geom1, geom2
                         else:
                             geometry_remaining, new_geometry = geom1, geom2
-
-            new_datasets[j].index.iloc[i].geometry = new_geometry
+            new_datasets[j].index.iloc[i, i_geom] = new_geometry
```
adamjstewart, line 169: Bug caused by new copy-on-write behavior.

It is no longer possible to chain multiple loc/iloc, they must now be combined.

isaaccorley (approved):

PR #3312, opened 2026-01-22
Add project instructions for Claude Code

hkristen:
As Claude Code is very popular right now, I believe this is a worthwile addtion following up with #3306. 

This PR enables Claude Code to read the same instructions as Copliot. This way we only need to maintain one instructions file for LLMs.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -1,0 +1,3 @@
+# TorchGeo Claude Code Instructions
+
+@.github/copilot-instructions.md


adamjstewart:
My only hesitation is one more top-level directory that users have to scroll through to get to the README. Is there an open feature request for Claude to recognize the `.github` directory? Or for Copilot to recognize `CONTRIBUTING.md` or https://torchgeo.readthedocs.io/en/stable/user/contributing.html?

robmarkcole:
Is not `AGENTS.md` supported by copilot/claude/codex? Seems undesirable to have a seperate file for each system

hkristen:
> My only hesitation is one more top-level directory that users have to scroll through to get to the README. Is there an open feature request for Claude to recognize the .github directory?

I had the same hesitation. The other option is to put the CLAUDE.md file into the root dir. Don't know if that is prettier :P 

> Is not AGENTS.md supported by copilot/claude/codex? Seems undesirable to have a seperate file for each system

I was not aware of this, would be amazing IF all LLM companies could agree that their CLI tools actually read this. At least Claude Code doesnt read it automatically. I need to direct it to it at the start of every session. It only reads CLAUDE.md files automatically at session start.

isaaccorley:
I'm actually leaning towards splitting up the current instructions into AGENTS.md and a .github/instructions/reviewer/instructions.md file which the copilot bot can read for reviews only.

The current instructions incorrectly mixes copilot chat and copilot review instructions.

isaaccorley:
@hkristen I've made a PR to move to using AGENTS.md instead so I think this solve your problem in this PR.

hkristen:
Replaced by #3336

PR #2917, opened 2025-08-04
feat(ci): Migrate GitHub Actions workflows to uv for package manageme…

orbin123:
This pull request migrates the core GitHub Actions workflows to use uv for Python dependency management. This change aims to leverage uv's performance improvements to achieve faster and more efficient CI/CD pipeline execution.

This PR specifically addresses the workflow files located in the .github/workflows/ directory as a focused first step towards the larger migration outlined in issue #2905.

Key Changes:
Files Modified: All .yaml files within the .github/workflows/ folder (deploy.yaml, release.yaml, style.yaml, tests.yaml, and tutorials.yaml).

Package Manager: Replaced all instances of pip install and pip list with uv pip install and uv pip list.

Installation: Added the astral-sh/setup-uv@v1 action to each job to ensure uv is available on the runner.

Caching: Updated actions/cache paths and keys to correctly cache uv's dependency cache (~/.cache/uv), which is more efficient for uv's operation.

Cleanup: Removed the unnecessary pip cache purge commands.

PR #3318, opened 2026-01-22
Migrate pip builds to UV : CI's only 

kshitijrajsharma:
### What does this PR do ? 
- It migrates the pip builds to uv build for test CI 
- Contributes to #2905  

This PR migrates the pip build to uv & also adds the local install of the package to allow deps changes in PR and test them in the PR itself ! 

### Things to note : 

- I had to pinpoint fastapi and pydantic in minimum builds .txt in order to make it build using uv

I added workflow dispatch to the ci ( this is for me to test the CI manually when i was doing changes , if this is not needed then I can remove it ) 

### Moving forward 
After the PR is merged i think next thing to do would be to migrate the min deps to pyproject.toml and then we can build all of the CI using uv sync , currently only the step of minimum build is using uv pip install , rest uses uv sync ! 

### How to test ? 
Well; if everything passes on the CI , then I guess job done , Beer on me :P took quite some hit and trials ! 

Closes #3030
Closes #2917
Closes #2907

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -20,7 +20,7 @@ jobs:
       - name: List dependencies
         run: uv pip list
       - name: Build project
-        run: python3 -m build
+        run: python -m build
       - name: Upload artifacts
         uses: actions/upload-artifact@v6.0.0
         with:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -28,8 +28,7 @@ jobs:
             requirements/datasets.txt
             requirements/tests.txt
       - name: Install dependencies
-        run: |
-          uv pip install .[datasets,tests]
+        run: uv pip install .[datasets,tests]
       - name: List dependencies
         run: uv pip list
       - name: Run integration checks
@@ -56,8 +55,7 @@ jobs:
             requirements/docs.txt
             requirements/tests.txt
       - name: Install dependencies
-        run: |
-          uv pip install .[docs,tests] planetary_computer pystac tensorboard h5py scikit-learn
+        run: uv pip install .[docs,tests] planetary_computer pystac tensorboard h5py scikit-learn
       - name: List dependencies
         run: uv pip list
       - name: Run notebook checks
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -34,14 +34,13 @@ jobs:
             requirements/models.txt
             requirements/tests.txt
       - name: Install dependencies
-        run: |
-          uv pip install -r requirements/required.txt -r requirements/datasets.txt -r requirements/models.txt -r requirements/tests.txt
+        run: uv pip install -r requirements/required.txt -r requirements/datasets.txt -r requirements/models.txt -r requirements/tests.txt
       - name: List installed dependencies
         run: uv pip list
       - name: Run checks
         run: |
           pytest --cov=torchgeo --cov-report=xml
-          python3 -m torchgeo --help
+          python -m torchgeo --help
       - name: Report coverage
         uses: codecov/codecov-action@v5.5.2
         with:
@@ -61,14 +60,13 @@ jobs:
           cache-dependency-glob: |
             requirements/min-reqs.old
       - name: Install dependencies
-        run: |
-          uv pip install -r requirements/min-reqs.old
+        run: uv pip install -r requirements/min-reqs.old
       - name: List installed dependencies
         run: uv pip list
       - name: Run checks
         run: |
           pytest --cov=torchgeo --cov-report=xml
-          python3 -m torchgeo --help
+          python -m torchgeo --help
       - name: Report coverage
         uses: codecov/codecov-action@v5.5.2
         with:
@@ -89,14 +87,13 @@ jobs:
             requirements/required.txt
             requirements/tests.txt
       - name: Install dependencies
-        run: |
-          uv pip install -r requirements/required.txt -r requirements/tests.txt
+        run: uv pip install -r requirements/required.txt -r requirements/tests.txt
       - name: List installed dependencies
         run: uv pip list
       - name: Run checks
         run: |
           pytest --cov=torchgeo --cov-report=xml
-          python3 -m torchgeo --help
+          python -m torchgeo --help
       - name: Report coverage
         uses: codecov/codecov-action@v5.5.2
         with:
--- a/.github/workflows/tutorials.yaml
+++ b/.github/workflows/tutorials.yaml
@@ -31,10 +31,8 @@ jobs:
             requirements/required.txt
             requirements/docs.txt
             requirements/tests.txt
-
       - name: Install dependencies
-        run: |
-          uv pip install -r requirements/required.txt -r requirements/docs.txt -r requirements/tests.txt planetary_computer pystac tensorboard h5py scikit-learn .
+        run: uv pip install -r requirements/required.txt -r requirements/docs.txt -r requirements/tests.txt planetary_computer pystac tensorboard h5py scikit-learn .
       - name: List dependencies
         run: uv pip list
       - name: Run notebook checks
--- a/.github/workflows/style.yaml
+++ b/.github/workflows/style.yaml
@@ -32,8 +32,7 @@ jobs:
             requirements/tests.txt
             requirements/style.txt
       - name: Install dependencies
-        run: |
-          uv pip install -r requirements/required.txt -r requirements/datasets.txt -r requirements/models.txt -r requirements/style.txt -r requirements/tests.txt
+        run: uv pip install -r requirements/required.txt -r requirements/datasets.txt -r requirements/models.txt -r requirements/style.txt -r requirements/tests.txt
       - name: List pip dependencies
         run: uv pip list
       - name: Run mypy checks
@@ -54,8 +53,7 @@ jobs:
           cache-dependency-glob: |
             requirements/style.txt
       - name: Install dependencies
-        run: |
-          uv run pip install -r requirements/style.txt
+        run: uv run pip install -r requirements/style.txt
       - name: List dependencies
         run: uv pip list
       - name: Run ruff checks


pyproject.toml
adamjstewart, line 1: Why are these changes required?

.github/workflows/tests.yaml
```diff
@@ -8,108 +8,62 @@ on:
...
+          fi
+          UV_RESOLUTION=highest ${{ matrix.install_cmd }}
+      - name: List installed dependencies
         run: |
-          pip install -r requirements/required.txt -r requirements/tests.txt
-          pip cache purge
-      - name: List pip dependencies
-        run: pip list
-      - name: Run pytest checks
+          echo "=== Installed packages ==="
```
adamjstewart: This line can be removed

.github/workflows/tests.yaml
```diff
@@ -8,108 +8,62 @@ on:
...
-          python-version: '3.13'
-      - name: Cache dependencies
-        uses: actions/cache@v5.0.2
-        id: cache
-        with:
-          path: ${{ env.pythonLocation }}
-          key: ${{ runner.os }}-${{ runner.arch }}-Python-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('requirements/required.txt', 'requirements/tests.txt') }}
-      - name: Install pip dependencies
-        if: steps.cache.outputs.cache-hit != 'true'
+          # Create venv if using pip install mode (uv sync creates it automatically)
```
adamjstewart: What is all of this?

.github/workflows/tests.yaml
```diff
@@ -8,108 +8,62 @@ on:
...
       - main
       - release**
+  workflow_dispatch:
 defaults:
   run:
     shell: bash
 jobs:
-  latest:
-    name: latest
+  tests:
```
adamjstewart: Why do we need to rename all of our tests? We would need to change all of our required tests to merge this.

.github/workflows/tests.yaml
```diff
@@ -8,108 +8,62 @@ on:
     branches:
       - main
       - release**
+  workflow_dispatch:
```
adamjstewart: Why is this required?

kshitijrajsharma: I added this before to be able to run the CI manually without having to commit for testing , removed in  #[5f8ab70](https://github.com/torchgeo/torchgeo/pull/3318/commits/5f8ab70a729aaab6dd76b6c81b8ede3a0fb1e753)

.github/workflows/tests.yaml
kshitijrajsharma: Reverted  !

.github/workflows/tests.yaml
kshitijrajsharma: Verbose , Reverted to keep minimal changes !

pyproject.toml
kshitijrajsharma: I had to limit the version of netcd5 and h5py because I was facing this problem on latest ubuntu https://github.com/torchgeo/torchgeo/actions/runs/21262626310/job/61193713267 , and then i took this reference https://github.com/torchgeo/torchgeo/blob/0a1cf1a87c332659551c33af25a3841560e5df04/requirements/datasets.txt#L5 so limited to <1.8

adamjstewart: netcdf4 1.7.4 is the latest version, there is not 1.8, so that is definitely not the reason the tests were failing.

We need to use pinned versions for CI (requirements.txt) and loose versions for installation (pyproject.toml). We shouldn't be using pyproject.toml in CI.

.github/workflows/tests.yaml
```diff
@@ -18,34 +18,23 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ['3.12', '3.13', '3.14', '3.14t']
+        python-version: ['3.12', '3.13']
```
adamjstewart: Undo this change

kshitijrajsharma: ok, was testing if 3.12 3.13 runs through without 3.14 , let me re-enable  !

pyproject.toml
kshitijrajsharma: alright !

pyproject.toml
```diff
@@ -298,7 +298,6 @@ filterwarnings = [
     # Expected warnings
     # Lightning warns us about using num_workers=0, but it's faster on macOS
     "ignore:The .*dataloader.* does not have many workers which may be a bottleneck:UserWarning:lightning",
-    "ignore:The .*dataloader.* does not have many workers which may be a bottleneck:lightning.fabric.utilities.warnings.PossibleUserWarning:lightning",
```
adamjstewart, line 303: Undo this change

.github/workflows/tests.yaml
```diff
@@ -88,28 +74,22 @@ jobs:
...
-        uses: actions/cache@v5.0.2
-        id: cache
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v7
         with:
-          path: ${{ env.pythonLocation }}
-          key: ${{ runner.os }}-${{ runner.arch }}-Python-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('requirements/required.txt', 'requirements/tests.txt') }}
-      - name: Install pip dependencies
-        if: steps.cache.outputs.cache-hit != 'true'
+          python-version: '3.13'
```
adamjstewart: Should be 3.14

.github/workflows/tests.yaml
```diff
@@ -22,30 +22,22 @@ jobs:
     steps:
       - name: Clone repo
         uses: actions/checkout@v6.0.2
-      - name: Set up python
-        id: setup-python
-        uses: actions/setup-python@v6.2.0
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v7
```
adamjstewart:
```suggestion
        uses: astral-sh/setup-uv@v7.2.0
```
Same below, this will give us more stable CI

.github/workflows/tests.yaml
```diff
@@ -22,30 +22,22 @@ jobs:
...
-      - name: Cache dependencies
-        uses: actions/cache@v5.0.2
-        id: cache
-        with:
-          path: ${{ env.pythonLocation }}
-          key: ${{ runner.os }}-${{ runner.arch }}-Python-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('requirements/required.txt', 'requirements/datasets.txt', 'requirements/models.txt', 'requirements/tests.txt') }}
-        if: ${{ runner.os != 'macOS' }}
-      - name: Install pip dependencies
-        if: steps.cache.outputs.cache-hit != 'true'
+          enable-cache: true
```
adamjstewart, line 29: I think we also need to set `cache-dependency-glob`, otherwise it won't know when to update the cache. Should be the same list of requirements files used below.

.github/workflows/tests.yaml
```diff
@@ -22,30 +22,22 @@ jobs:
...
+          uv pip install -e . --no-deps
+          uv pip install -r requirements/required.txt -r requirements/datasets.txt -r requirements/models.txt -r requirements/tests.txt
+      - name: List installed dependencies
+        run: uv pip list
+      - name: Run checks
         run: |
-          pytest --cov=torchgeo --cov-report=xml
-          python3 -m torchgeo --help
+          uv run --no-sync pytest --cov=torchgeo --cov-report=xml
+          uv run --no-sync torchgeo --help
```
adamjstewart: What does `--no-sync` do?

.github/workflows/tests.yaml
```diff
@@ -22,30 +22,22 @@ jobs:
...
+          enable-cache: true
+      - name: Install dependencies
         run: |
-          pip install -r requirements/required.txt -r requirements/datasets.txt -r requirements/models.txt -r requirements/tests.txt
-          pip cache purge
-      - name: List pip dependencies
-        run: pip list
-      - name: Run pytest checks
+          uv venv
+          uv pip install -e . --no-deps
```
adamjstewart: Is this local editable install needed?

.github/workflows/tests.yaml
kshitijrajsharma: it will avoid uv syncing via .toml and use the lib that we installed on above steps !

.github/workflows/tests.yaml
kshitijrajsharma: local installation is needed to checkout the PR changes when CI runs  but not the editable version , should i remove `-e .`  ? we can do uv pip install .

.github/workflows/tests.yaml
kshitijrajsharma: Should we define a key then ?         
 path: /tmp/.uv-cache
 key: uv-${{ runner.os }}  
 
 may be with minimal optional job name ? 
          
 ref here :  https://docs.astral.sh/uv/guides/integration/github/#caching , or the default should be sufficient ?

.github/workflows/tests.yaml
```diff
@@ -56,28 +48,22 @@ jobs:
...
-        if: steps.cache.outputs.cache-hit != 'true'
+          enable-cache: true
+      - name: Install dependencies
         run: |
-          pip install -r requirements/min-reqs.old
-          pip cache purge
-      - name: List pip dependencies
-        run: pip list
-      - name: Run pytest checks
+          uv venv
```
kshitijrajsharma: uv has this env variable : env:
  UV_SYSTEM_PYTHON: 1 
if we use ti then venv is not required   it would install it similar to --system, I don't see much difference though ! what do you suggest ? @adamjstewart

.github/workflows/tests.yaml
adamjstewart: We need to set `cache-dependency-glob`: https://github.com/astral-sh/setup-uv

.github/workflows/tests.yaml
adamjstewart: We've never installed before, is our pwd being changed somehow?

.github/workflows/tests.yaml
adamjstewart: I would rather use the `activate-environment` setting in https://github.com/astral-sh/setup-uv

kshitijrajsharma: Nice , let me do that !

kshitijrajsharma: Done !

.github/workflows/tests.yaml
kshitijrajsharma: aah then previously may be you were only relying on the pytest ! How would you check if the local installation of the lib works for PRs for eg ? I see you were doing torchgeo --help , it would have gone before that step !

adamjstewart: We only use `pip install` and `torchgeo --help` in `release.yaml`, in all the others we rely on the fact that we checkout the files to the local directory. It is a pure-Python library, so nothing needs to be built.

.github/workflows/tests.yaml
```diff
@@ -22,30 +22,24 @@ jobs:
...
-        with:
-          path: ${{ env.pythonLocation }}
-          key: ${{ runner.os }}-${{ runner.arch }}-Python-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('requirements/required.txt', 'requirements/datasets.txt', 'requirements/models.txt', 'requirements/tests.txt') }}
-        if: ${{ runner.os != 'macOS' }}
-      - name: Install pip dependencies
-        if: steps.cache.outputs.cache-hit != 'true'
+          enable-cache: true
+          activate-environment: true
+          cache-dependency-glob: |
+            requirements/*.txt
```
adamjstewart: Technically not all .txt files, we don't need to include docs or style deps

.github/workflows/tests.yaml
```diff
@@ -56,28 +50,24 @@ jobs:
...
-        id: cache
-        with:
-          path: ${{ env.pythonLocation }}
-          key: ${{ runner.os }}-${{ runner.arch }}-Python-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('requirements/min-reqs.old') }}
-      - name: Install pip dependencies
-        if: steps.cache.outputs.cache-hit != 'true'
+          enable-cache: true
+          activate-environment: true
+          cache-dependency-glob: |
+            **/min-reqs.old
```
adamjstewart:
```suggestion
            requirements/min-reqs.old
```

.github/workflows/tests.yaml
```diff
@@ -88,28 +78,25 @@ jobs:
...
-        with:
-          path: ${{ env.pythonLocation }}
-          key: ${{ runner.os }}-${{ runner.arch }}-Python-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('requirements/required.txt', 'requirements/tests.txt') }}
-      - name: Install pip dependencies
-        if: steps.cache.outputs.cache-hit != 'true'
+          enable-cache: true
+          activate-environment: true
+          cache-dependency-glob: |
+            **/required.txt
+            **/tests.txt
```
adamjstewart:
```suggestion
            requirements/required.txt
            requirements/tests.txt
```

.github/workflows/tests.yaml
kshitijrajsharma: aah then it makes sense !

adamjstewart: I think we can remove this line now then?

.github/workflows/tests.yaml
```diff
@@ -22,30 +22,27 @@ jobs:
...
-        run: pip list
-      - name: Run pytest checks
+          uv pip install -e . --no-deps
+          uv pip install -r requirements/required.txt -r requirements/datasets.txt -r requirements/models.txt -r requirements/tests.txt
+      - name: List installed dependencies
+        run: uv pip list
+      - name: Run checks
         run: |
-          pytest --cov=torchgeo --cov-report=xml
-          python3 -m torchgeo --help
```
adamjstewart, line 48: Does the previous syntax no longer work?

kshitijrajsharma: same command : with slightly different way of calling them,  i added them inside uv run 

```code
  uv run --no-sync pytest --cov=torchgeo --cov-report=xml
  uv run --no-sync torchgeo --help
 ```

File: .github/workflows/tests.yaml (hunk 11 above)
Comment (kshitijrajsharma) [reply .github/workflows/tests.yaml]: let me check  !

Comment (kshitijrajsharma) [reply .github/workflows/tests.yaml]: Done !

**Comment by kshitijrajsharma** [comment]
Thanks and Yes, do you want me to include those in same PR or different for each one ? @adamjstewart

**Comment by adamjstewart** [comment]
Yes, same PR please.

**Comment by adamjstewart** [comment]
Couple more remaining.

File: .github/workflows/release.yaml [hunk 17]
```diff
@@ -16,56 +16,52 @@ jobs:
...
       - name: Run integration checks
         run: |
-          pytest -m slow --durations=10 tests/samplers
-          pytest -m slow --durations=10 tests/trainers
-          python -m torchgeo --help
-          torchgeo --help
+          uv run --no-sync pytest -m slow --durations=10 tests/samplers
+          uv run --no-sync pytest -m slow --durations=10 tests/trainers
+          uv run --no-sync python -m torchgeo --help
+          uv run --no-sync torchgeo --help
```
adamjstewart: Curious if we can remove `uv run --no-sync` from all tests and have them still work. It's a lot of extra code and it's not clear why it's necessary.

kshitijrajsharma: From my understanding : 
this method is added specific to the TorchGeo implementation. Since we are using fixed, pinned versions from requirements.txt, we want to prevent uv from automatically syncing the virtual environment with the pyproject.toml file. Typically, when uv run is executed, it checks for updates in the toml file and attempts to reconcile the environment; we need to bypass this behavior so that it respects our requirements file instead. To achieve this without repeatedly using the --no-sync flag, we could either omit the pyproject.toml file from the CI environment so uv finds nothing to sync, or we could pass the configuration via an environment variable to streamline the command execution Or we keep it as it is for now while we look for an option to optimize ! 

Reference : 
 https://docs.astral.sh/uv/concepts/projects/sync/#automatic-lock-and-sync   , wdyt

adamjstewart: I'm not talking about removing `--no-sync`, I'm talking about removing `uv run --no-sync`. Assuming these tools are in PATH, there's no reason to use uv to run them.

kshitijrajsharma: gotcha ~!

.github/workflows/deploy.yaml
```diff
@@ -10,16 +10,17 @@ jobs:
...
+          activate-environment: true
       - name: Install pip dependencies
-        run: pip install build
+        run: uv pip install build
       - name: List pip dependencies
-        run: pip list
+        run: uv pip list
       - name: Build project
-        run: python3 -m build
+        run: uv run python -m build
```
adamjstewart:
```suggestion
        run: python3 -m build
```

.github/workflows/tests.yaml
```diff
@@ -56,28 +52,23 @@ jobs:
...
-        run: pip list
-      - name: Run pytest checks
+          uv pip install -r requirements/min-reqs.old
+      - name: List installed dependencies
+        run: uv pip list
+      - name: Run checks
         run: |
           pytest --cov=torchgeo --cov-report=xml
-          python3 -m torchgeo --help
+          python -m torchgeo --help
```
adamjstewart:
```suggestion
          python3 -m torchgeo --help
```

.github/workflows/tests.yaml
```diff
@@ -88,28 +79,24 @@ jobs:
...
-        run: pip list
-      - name: Run pytest checks
+          uv pip install -r requirements/required.txt -r requirements/tests.txt
+      - name: List installed dependencies
+        run: uv pip list
+      - name: Run checks
         run: |
           pytest --cov=torchgeo --cov-report=xml
-          python3 -m torchgeo --help
+          python -m torchgeo --help
```
adamjstewart:
```suggestion
          python3 -m torchgeo --help
```

.github/workflows/tutorials.yaml
```diff
@@ -20,24 +20,23 @@ jobs:
...
-          path: ${{ env.pythonLocation }}
-          key: ${{ runner.os }}-${{ runner.arch }}-Python-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('requirements/required.txt', 'requirements/docs.txt', 'requirements/tests.txt') }}-tutorials
+          enable-cache: true
+          activate-environment: true
+          cache-dependency-glob: |
+            requirements/required.txt
+            requirements/docs.txt
+            requirements/tests.txt
+
       - name: Install pip dependencies
```
adamjstewart: Optional: could convert "Install pip dependencies" and "List pip dependencies" to something like "Install dependencies" and "List dependencies". We used to also install apt/brew deps as well but removed those.

adamjstewart:
Needs a rebase for tests to run.

.github/workflows/tutorials.yaml
kshitijrajsharma: aah okay ; Included in #[f060dc7](https://github.com/torchgeo/torchgeo/pull/3318/commits/f060dc7c8f6df10a74b478107311b03725d27520)

.github/workflows/tests.yaml
```diff
@@ -22,30 +22,26 @@ jobs:
...
-        run: pip list
-      - name: Run pytest checks
+          uv pip install -r requirements/required.txt -r requirements/datasets.txt -r requirements/models.txt -r requirements/tests.txt
+      - name: List installed dependencies
+        run: uv pip list
+      - name: Run checks
         run: |
           pytest --cov=torchgeo --cov-report=xml
-          python3 -m torchgeo --help
+          python -m torchgeo --help
```
adamjstewart:
```suggestion
          python3 -m torchgeo --help
```

kshitijrajsharma:
Awesome , Thanks @adamjstewart ! ! 

>We should also add uv to the installation instructions and possibly a few other places. But let's do those in follow-up PRs.

Sure !

adamjstewart:
Did my last change somehow break this? It seems like `python` is installed but not `python3`? First time seeing that, but I have heard of this problem on Windows. Let me undo.

adamjstewart (approved):

kshitijrajsharma:
> Did my last change somehow break this? It seems like `python` is installed but not `python3`? First time seeing that, but I have heard of this problem on Windows. Let me undo.

Aah strange, weird problem in windows, I see ! Nice to know

PR #3324, opened 2026-01-24
Docs: reorganize model architectures

adamjstewart:
Unlike torchvision where all models have the same input dimensions and models are organized by output dimensions, I have organized these by input dimensions to highlight the ongoing work to add time series support to TorchGeo.

There may be some models that belong in multiple categories. This might be tricky to make Sphinx happy, as items are not allowed to appear in multiple toctrees. I would vote for defaulting to the non-2D section since that section is most crowded.

### DIFF
--- a/docs/api/models.rst
+++ b/docs/api/models.rst
@@ -10,8 +10,8 @@ Model Architectures
 
 TorchGeo contains a number of model architectures depending on the task you are trying to solve and your model inputs.
 
-1D Time Series (:math:`B \times T \times C`)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+1D Time Series (:math:`\scriptstyle B \times T \times C`)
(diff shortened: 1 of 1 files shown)


adamjstewart:
Wondering whether we should place the math below the section header, some lines get long.

adamjstewart:
Could also add links to timm, SMP, and torchange for more recommended model architectures...

docs/api/models.rst
```diff
@@ -8,32 +8,69 @@ This section provides an overview of all models available in ``torchgeo.models``
 Model Architectures
 -------------------
 
+TorchGeo contains a number of model architectures depending on the task you are trying to solve and your model inputs.
+
+1D Time Series (:math:`B \times T \times C`)
```
ashnair1: > Wondering whether we should place the math below the section header, some lines get long.

I agree that the dimensions are a bit loud and makes the line too long. Could reduce the size by specifying `\scriptstyle`

```suggestion
1D Time Series (:math:`\scriptstyle B \times T \times C`)
```

PR #3334, opened 2026-01-28
Type checking: mypy -> ty

adamjstewart:
Mostly just want to see how far we are from fixing all remaining ty errors. Won't merge this until we fix everything.

Closes #3170

### DIFF
--- a/.github/workflows/style.yaml
+++ b/.github/workflows/style.yaml
@@ -80,7 +80,6 @@ jobs:
         run: npm ls --all
       - name: Run prettier formatting
         run: npx prettier --check .
-
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.head.label || github.head_ref || github.ref }}
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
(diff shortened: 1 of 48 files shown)


adamjstewart:
Frustratingly, CI has things flagged by ty that I can't reproduce locally, and vice versa. This makes it really difficult to debug and fix. Could be version differences, but almost all of my versions are the same.

isaaccorley:
Ty is happy!

.pre-commit-config.yaml
```diff
@@ -12,10 +12,12 @@ repos:
           - python
           - jupyter
 
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.18.2
+  - repo: local
```
adamjstewart, line 15: There is not yet a pre-commit hook for ty, GitHub suggests using this format, haven't tested myself.

isaaccorley: I've been using this method for awhile in Wherobots and personal repos and it works well.

adamjstewart:
We now only have 70 type ignores in TorchGeo. Most previous type ignores were due to non-type checked dependencies like torchvision. We should continue to try to remove any type ignores, casts, or Any usage in follow-up PRs.

isaaccorley (approved):

PR #3316, opened 2026-01-22
Enhance  copilot instructions

jonlwowski012:
Expanded guidelines for handling copilot instructions, including boundaries, code quality standards, common pitfalls, and error handling.

PR #3336, opened 2026-01-28
docs: copilot-instructions -> AGENTS.mds

isaaccorley:
Copilot Reviewer now is able to reference custom instructions from `AGENTS.md`. [AGENTS.md](https://agents.md) is currently the standard instructions location for all coding agent providers (not github copilot specific). This should fix the issue in this PR where Claude Code isn't automatically aware of the `~/.github/copilot-instructions.md` https://github.com/torchgeo/torchgeo/pull/3312.

https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions#creating-custom-instructions

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,4 +1,4 @@
-# TorchGeo Instructions
+# TorchGeo LLM Instructions
 
 Datasets, samplers, transforms, and pre-trained models for geospatial data.
 


AGENTS.md
```diff
@@ -1,4 +1,4 @@
-# TorchGeo Copilot Instructions
+# TorchGeo Instructions
```
adamjstewart: Should we name this "LLM Instructions" or something so humans don't stumble on this?

adamjstewart:
@hkristen

adamjstewart:
If we review #3316 first it would mean less rebasing for @jonlwowski012

AGENTS.md
hkristen: TorchGeo LLM instructions sounds good

isaaccorley: Does this matter? I don't think so

isaaccorley:
> If we review #3316 first it would mean less rebasing for @jonlwowski012

I think that PR needs more revision since we want to reduce number of tokens and much of it is repetitive of what's already in the instructions.

AGENTS.md
adamjstewart: If it's a top-level file, people are going to look at it. It isn't obviously clear what `AGENTS.md` is for.

PR #3338, opened 2026-01-29
download_url: replace torchvision utility

adamjstewart:
Drop-in replacement for torchvision's utility with support for os.PathLike and secure hashing algorithms. Doesn't yet support Google Drive or redirection. Hoping to get a minimal working solution first and expand features in follow-up PRs.

Depends on #3302

### DIFF
--- a/torchgeo/datasets/utils.py
+++ b/torchgeo/datasets/utils.py
@@ -309,7 +309,7 @@ def check_integrity(fpath: Path, md5: str | None = None, **kwargs: str | None) -
     Args:
         fpath: File path to check.
         md5: Expected MD5 checksum.
-        **kwargs: Expected checksum for any valid :module:`hashlib` algorithm.
+        **kwargs: Expected checksum for any valid :mod:`hashlib` algorithm.
 
     Returns:
         True if file exists and checksum is None or matches, else False.
@@ -383,7 +383,7 @@ def download_url(
         filename: File path to save to. Defaults to the basename of the URL.
         md5: Expected MD5 checksum.
         max_redirect_hops: Maximum number of allowed redirection attempts.
-        **kwargs: Expected checksum for any valid :module:`hashlib` algorithm.
+        **kwargs: Expected checksum for any valid :mod:`hashlib` algorithm.
 
     Raises:
         RuntimeError: If checksum of downloaded file does not match.


adamjstewart:
Could also use https://docs.pytorch.org/docs/stable/hub.html#torch.hub.download_url_to_file

torchgeo/datasets/utils.py
```diff
@@ -361,6 +362,50 @@ def extract_archive(
...
+    if not filename:
+        filename = os.path.basename(url)
+
+    root = os.path.expanduser(root)
+    os.makedirs(root, exist_ok=True)
+
+    fpath = os.path.join(root, filename)
+    if not check_integrity(fpath, md5, **kwargs):
+        # TODO: use fsspec if we want AWS/Azure/GCS support
+        # TODO: use gdown if we want Google Drive support
```
adamjstewart, line 401: At the moment, this PR will break downloads for:

* ETCI 2021: Unknown license, I'm in touch with the authors to see if we can redistribute
* FAIR1M: CC-BY-NC-SA-3.0, Google Earth imagery, unclear if we can legally redistribute
* GID15: Unknown license, Google Earth imagery, unclear if we can legally redistribute
* LEVIRCD: Unknown license, Google Earth imagery, unclear if we can legally redistribute: #3351

Given that there are only 4 datasets in jeopardy, I'm personally okay with breaking these until we get legal advice. Don't want a new dep just to support these given that we may not legally be allowed to download these anyway.

torchgeo/datasets/utils.py
```diff
@@ -361,6 +362,50 @@ def extract_archive(
...
+        # TODO: use requests if we want redirect support
```
adamjstewart, line 402: At the very least, simple redirects like https://hf.co -> https://huggingface.co seem to work?

torchgeo/datasets/utils.py
```diff
@@ -361,6 +362,50 @@ def extract_archive(
...
+        # TODO: use tqdm if we want a progress bar
```
adamjstewart, line 403: We do seem to have some simple progress bar, may not need tqdm.

torchgeo/datasets/utils.py
adamjstewart: @robmarkcole this is our current list of GDrive datasets we would like to change.

isaaccorley: Agree on this

isaaccorley (dismissed):

isaaccorley:
> Could also use https://docs.pytorch.org/docs/stable/hub.html#torch.hub.download_url_to_file

I'd be for using this but I'm okay with the current solution since it's working and fixes our needs.

torchgeo/datasets/utils.py
```diff
@@ -361,6 +362,50 @@ def extract_archive(
     return to_path
 
 
+def download_url(
+    url: str,
+    root: Path,
+    filename: Path | None = None,
+    md5: str | None = None,
+    max_redirect_hops: int = 3,
```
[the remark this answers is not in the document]
isaaccorley: @adamjstewart decent catch by copilot. Do we want to remove this arg?

torchgeo/datasets/utils.py
```diff
@@ -361,6 +362,50 @@ def extract_archive(
...
+        download_url(url, root, md5='...')
+        download_url(url, root, sha256='...')
+
+    Args:
+        url: URL to download.
+        root: Root directory to save downloaded file to.
+        filename: File path to save to. Defaults to the basename of the URL.
+        md5: Expected MD5 checksum.
+        max_redirect_hops: Maximum number of allowed redirection attempts.
+        **kwargs: Expected checksum for any valid :module:`hashlib` algorithm.
```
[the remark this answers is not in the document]
isaaccorley: @adamjstewart I looked at sphinx docs and this seems to be correct

adamjstewart:
> > Could also use https://docs.pytorch.org/docs/stable/hub.html#torch.hub.download_url_to_file
> 
> I'd be for using this but I'm okay with the current solution since it's working and fixes our needs.

Looked at the source code, it's quite complicated but doesn't have too many features we don't already have: https://github.com/pytorch/pytorch/blob/v2.10.0/torch/hub.py#L694. Uses requests and tqdm behind the scenes.

I only tested this on a couple HF datasets, would be nice to test downloading from more diverse hosts.

torchgeo/datasets/utils.py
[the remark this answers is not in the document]
adamjstewart: I want to keep it for a little while for compatibility with torchvision but we can slowly deprecate and remove this.

PR #3339, opened 2026-01-29
download_and_extract_archive: replace torchvision utility

adamjstewart:
Drop-in replacement for torchvision's utility with support for os.PathLike and secure hashing algorithms.

Depends on #3338, #3307, #3302

### DIFF
--- a/tests/datasets/conftest.py
+++ b/tests/datasets/conftest.py
@@ -6,7 +6,6 @@ import shutil
 from typing import Any
 
 import pytest
-import torchvision.datasets.utils
 from _pytest.fixtures import SubRequest
 from pytest import MonkeyPatch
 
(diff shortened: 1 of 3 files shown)


tests/datasets/conftest.py
```diff
@@ -21,7 +20,6 @@ def copy(url: str, root: Path, *args: Any, **kwargs: Any) -> None:
 
 @pytest.fixture(autouse=True)
 def download_url(monkeypatch: MonkeyPatch, request: SubRequest) -> None:
```
adamjstewart, line 22: If we convert all of our fake test URLs to file:// we can someday remove this fixture.

tests/datasets/test_utils.py
```diff
@@ -379,6 +380,16 @@ def test_download_url(tmp_path: Path) -> None:
         download_url(url, tmp_path, md5=md5 + '2')
 
 
+def test_download_and_extract_archive(tmp_path: Path) -> None:
+    url = Path('tests/data/vhr10/NWPU VHR-10 dataset.zip')
```
adamjstewart, line 384: We can't use a real URL here because of our autouse fixture.

PR #3342, opened 2026-01-29
FarSeg: do not require internet access during tests

adamjstewart:
Attempting to finally solve some of the CI issues reported in #1088. No TorchGeo code should require internet access to test.

Alternatively, could switch to a timm backbone or use the deprecated `pretrained=True` syntax in the model loader. Would be a shame if we can no longer use pretrained models.

Alternative to #3348
Closes #3348

### DIFF
--- a/tests/models/test_changestar.py
+++ b/tests/models/test_changestar.py
@@ -17,9 +17,7 @@ SF = [4, 8, 1]
 class TestChangeStar:
     @torch.no_grad()
     def test_changestar_farseg_classes(self) -> None:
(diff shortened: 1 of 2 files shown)


isaaccorley:
I don't think this is desirable because this would mean users loads an inferior model which is fully random init instead of one that has a pretrained encoder. I'd be in favor of simply changing the backbone to use timm which should solve this issue though.

adamjstewart:
I think #3348 is the better way to go here, take a look and see what you think.

PR #3344, opened 2026-01-29
UNet: do not require internet access during tests

adamjstewart:
Part of #1088 

A bit confusing because technically our API only supports custom weights or None, it doesn't support True/False/etc. But technically None meant that imagenet weights were loaded, which is no longer the case. So technically backwards-incompatible. Not sure if there's a better way to handle this such that imagenet weights are still supported.

### DIFF
--- a/torchgeo/models/unet.py
+++ b/torchgeo/models/unet.py
@@ -200,9 +200,9 @@ def unet(
         A U-Net model.
     """
     kwargs['arch'] = 'Unet'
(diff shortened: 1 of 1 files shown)


adamjstewart:
> Not sure if there's a better way to handle this such that imagenet weights are still supported.

I guess they are if you pass them in as a WeightsEnum. So this is backwards-incompatible, but it's still possible to get the previous behavior. Should be safe to merge then.

PR #3345, opened 2026-01-30
Geospatial Tutorial: fix hyperspectral image link

adamjstewart:
Previous image link is dead.


adamjstewart:
We should probably add copyright info to more of our images...

PR #3346, opened 2026-01-30
Pretrained Weights Tutorial: use GPU runtime by default

adamjstewart:


adamjstewart:
Oh wait, we removed training from this tutorial, so now all we do is load the weights. Nevermind, this doesn't need a GPU.

PR #3348, opened 2026-01-30
FarSeg: do not require internet access during tests

adamjstewart:
Alternative to #3342 that keeps the ability to easily use pre-trained backbones but brings FarSeg and ChangeStar more inline with our other models. The user can now specify any weights they want, including models pre-trained on Sentinel imagery.

Closes #3342


adamjstewart:
Could add additional mocked_weights and real weights tests if we want to but it doesn't affect test coverage.

isaaccorley:
This LGTM

PR #3349, opened 2026-01-30
InstanceSegmentationTask: do not require internet access during tests

adamjstewart:
Working towards #1088 and #3343. This puts our InstanceSegmentationTask more inline with other tasks, with support for arbitrary WeightsEnum inputs. We probably should also support bool and str like our other tasks, but I was too tired to add tests for this and my goal of avoiding downloads is all I care about right now. Perhaps in a follow-up PR?

If this looks good, I'll do the same thing for ObjectDetectionTask.

### DIFF
--- a/tests/trainers/test_instance_segmentation.py
+++ b/tests/trainers/test_instance_segmentation.py
@@ -71,10 +71,6 @@ class TestInstanceSegmentationTask:
         with pytest.raises(ValueError, match=match):
             InstanceSegmentationTask(backbone='invalid_backbone')
 
-    def test_weights(self) -> None:
-        InstanceSegmentationTask(weights=True, num_classes=3)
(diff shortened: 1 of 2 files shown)


adamjstewart:
> This lgtm although I wonder if we should put an example in the docstring that shows how to use the weights and weights_backbone args. Most users I imagine will get stumped on what weights to use for this task.

We have an entire tutorial on using the weights API. I agree that it's a bit confusing that you have to get your weights from torchvision to use with torchgeo. Other than that, it's quite similar to all other trainers.

> Seems like this is making it more complex for the user

Also more powerful since you can use any weights you want, not just the ImageNet weights we used to default to.

Any suggestions for improvement that would bring this trainer closer to our other trainers? I don't want torchvision-based trainers to be any different from timm/smp-based trainers.

Issue #3340, opened 2026-01-29
Move LEVIR-CD to HF


robmarkcole:
### Summary

As discussed on slack. https://justchenhao.github.io/LEVIR/

Q1: Keep train/val/test as 3 seperate zips or merged into one? -> as original split
Q2: Add Levir-cd+ also/instead? Same but more samples
Q3: Add levir-cd-256? Pre-chipped from https://github.com/AndreaCodegoni/Tiny_model_4_CD -> not doing

https://huggingface.co/datasets/satellite-image-deep-learning/LEVIR-CD

adamjstewart:
Note that we can't add LEVIR-CD to the _TorchGeo_ HF because:

1. It contains Google Earth imagery that may have been illegally scraped
2. It does not list an open source license that the dataset can be redistributed under

Perhaps the authors can shed some light on this? @justchenhao

Answers to your questions:

1. I would prefer to keep things the same as the original source, makes it easier to download only a subset
2. We should add both. Could be separate repos I suppose.
3. We don't yet have a data loader for this one, but we could.

PR #3351, opened 2026-01-31
LEVIRCD: update to use Hugging Face

robmarkcole:
Close https://github.com/torchgeo/torchgeo/issues/3340


adamjstewart (commented):
Any idea why the md5s changed?

robmarkcole:
I zipped the folders myself, having unzipped the originals when I downloaded them at some point in the past

adamjstewart:
I would be more comfortable re-downloading the originals and uploading those so that the checksums don't change.

Do you also want to include LEVIR-CD+?

robmarkcole:
Yes can do

PR #3353, opened 2026-02-01
Datasets: fix plot docstring reference to getitem

adamjstewart:
Many of our `plot` method docstring references to `__getitem__` are either overly constrained or incorrect due to copy-n-paste errors.

### DIFF
--- a/torchgeo/datasets/cdl.py
+++ b/torchgeo/datasets/cdl.py
@@ -341,7 +341,7 @@ class CDL(RasterDataset):
         """Plot a sample from the dataset.
 
         Args:
-            sample: a sample returned by :meth:`RasterDataset.__getitem__`
+            sample: a sample returned by :meth:`__getitem__`
             show_titles: flag indicating whether to show titles above each panel
             suptitle: optional string to use as a suptitle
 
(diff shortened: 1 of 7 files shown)


torchgeo/datasets/cdl.py
```diff
@@ -341,7 +341,7 @@ def plot(
         """Plot a sample from the dataset.
 
         Args:
-            sample: a sample returned by :meth:`RasterDataset.__getitem__`
+            sample: a sample returned by :meth:`__getitem__`
```
adamjstewart, line 344: Incorrect, CDL has its own `__getitem__` that is different from the base class

torchgeo/datasets/skyscript.py
```diff
@@ -164,7 +164,7 @@ def plot(
         """Plot a sample from the dataset.
 
         Args:
-            sample: a sample returned by :meth:`RasterDataset.__getitem__`
+            sample: a sample returned by :meth:`__getitem__`
```
adamjstewart, line 167: Incorrect, SkyScript isn't even a RasterDataset

PR #3356, opened 2026-02-02
Add Embedded Seamless Data (ESD) dataset

heiyuxiaokai:
- [x]  Basic data loader
- [x]  Tests
- [x]  Docs

@adamjstewart 

Depends on [ESD Code](https://github.com/shuangchencc/ESD) and [IEarth](https://data-starcloud.pcl.ac.cn/iearthdata/64)

图片1


tests/data/esd/data.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
```
adamjstewart: Could remove the encoding line

tests/data/esd/data.py
```diff
@@ -0,0 +1,90 @@
...
+# Licensed under the MIT License.
+
+import os
+import numpy as np
+import rasterio
+from rasterio.transform import from_bounds
+from rasterio.crs import CRS
+
+WIDTH = 3600
+HEIGHT = 3600
```
adamjstewart: Let's do 36x36. This is just for testing, we want the tests to run as fast as possible.

tests/data/esd/data.py
```diff
@@ -0,0 +1,90 @@
...
+BANDS = 13
+PIXEL_SIZE = 30.0  # meters per pixel
+np.random.seed(0)
+
+# EPSG 32602 / UTM zone 2N using PROJ string (bypass proj.db issues)
```
adamjstewart, line 19: What kind of proj.db issues do you have? I usually just use `CRS.from_epsg(32602)` since it's less code.

tests/data/esd/data.py
```diff
@@ -0,0 +1,90 @@
...
+CRS_32602 = CRS.from_string("+proj=utm +zone=2 +north +datum=WGS84 +units=m +no_defs")
+
+def write_mock_geotiff(path: str, origin_x: float, origin_y: float,
+                       width: int = WIDTH, height: int = HEIGHT,
+                       bands: int = BANDS, crs: CRS = CRS_32602):
```
adamjstewart:
```suggestion
                       bands: int = BANDS, crs: CRS = CRS_32602) -> None:
```

tests/data/esd/data.py
```diff
@@ -0,0 +1,90 @@
...
+
+# 2024 image (02VMN)
+write_mock_geotiff(
+    path=os.path.join("tests", "data", "esd", "SDC30_EBD_V001", "2024",
+                      "SDC30_EBD_V001_02VMN_2024_mock.tif"),
+    origin_x=399945.0,
+    origin_y=6800055.0
+)
+
+# 2005 image (02WMV)
```
adamjstewart: Technically only need 1 image for testing. Trying to keep the test data from getting too big.

tests/datasets/test_esd.py
```diff
@@ -0,0 +1,66 @@
...
+class TestEmbeddedSeamlessData:
+    @pytest.fixture(
+        params=[
+            '2005/SDC30_EBD_V001_02WMV_2005_mock.tif',
+            '2024/SDC30_EBD_V001_02VMN_2024_mock.tif'
+        ]
+    )
+    def dataset(self, request: SubRequest) -> EmbeddedSeamlessData:
+        paths = os.path.join('tests', 'data', 'esd', 'SDC30_EBD_V001', request.param)
+        return EmbeddedSeamlessData(paths)
```
adamjstewart: I think we'll also need a fake `nn.Identity()` transform to get full test coverage.

tests/datasets/test_esd.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def test_len(self, dataset: EmbeddedSeamlessData) -> None:
+        assert len(dataset) == 1
+
+    def test_getitem(self, dataset: EmbeddedSeamlessData) -> None:
+        x = dataset[dataset.bounds]
+        assert isinstance(x, dict)
+        assert isinstance(x['image'], torch.Tensor)
+        # Rasterio may return nodata for upside down rasters
```
adamjstewart: Is this true for your dataset? I think only Google's embeddings suffer from this problem.

tests/datasets/test_esd.py
```diff
@@ -0,0 +1,66 @@
...
+            EmbeddedSeamlessData(tmp_path)
+
+    def test_invalid_query(self, dataset: EmbeddedSeamlessData) -> None:
+        with pytest.raises(
+            IndexError, match=r'index: .* not found in dataset with bounds:'
+        ):
+            dataset[0:0, 0:0, pd.Timestamp.min : pd.Timestamp.min]
+
+if __name__ == '__main__':
+    pytest.main()
```
adamjstewart: Can remove this, this file is already run by pytest automatically.

torchgeo/datasets/esd.py
```diff
@@ -0,0 +1,269 @@
+import pathlib
```
adamjstewart: Needs copyright comment and docstring description.

torchgeo/datasets/esd.py
```diff
@@ -0,0 +1,269 @@
...
+
+    .. note::
+        The output retains the channel dimension corresponding to embedding levels.
+        Users can further convert embeddings to visualizations or aggregate them
+        for downstream tasks.
+    
+    """
+    def __init__(
+        self,
+        levels: List[int] = [8, 8, 8, 5, 5, 5],
```
adamjstewart: If tuples are also supported, could also use Sequence here.

torchgeo/datasets/esd.py
```diff
@@ -0,0 +1,269 @@
...
+from typing import List
+from .utils import (
+    GeoSlice,
+    Path,
+    Sample,
+    disambiguate_timestamp
+)
+
+
+class ESDQuantizer():
```
adamjstewart: Interesting to see this documentation. So the original data is uint16 and you decode to float32 or float64 during usage?

We didn't bother with this for other datasets because TorchGeo automatically casts everything to float32 and most people will apply normalization in their data module. But I'm okay with adding this to the dataset instead.

torchgeo/datasets/esd.py
```diff
@@ -0,0 +1,269 @@
...
+
+        basis_tensor = torch.cumprod(
+            torch.tensor([1] + levels[:-1], dtype=torch.int32),
+            dim=0
+        )
+        self._basis = basis_tensor
+
+    @staticmethod
+    def _exists(v) -> bool:
+        return v is not None
```
adamjstewart: Could remove this, the parameter is required to be of type Tensor, so no need to check whether or not it is defined.

torchgeo/datasets/esd.py
```diff
@@ -0,0 +1,269 @@
...
+
+    Produced by the ESDNet framework, ESD provides an ultra-lightweight, globally
+    consistent representation of surface conditions, enabling flexible, high-resolution
+    analysis of land surface dynamics over decades.
+
+    If you use this dataset in your research, please cite:
+
+    * Chen, S. et al. (2026). "Democratizing planetary-scale analysis: An ultra-lightweight
+      Earth embedding database for accurate and flexible global land monitoring." 
+      arXiv preprint arXiv:2601.11183. <https://arxiv.org/abs/2601.11183>
```
adamjstewart: We usually just add the link, I figure most people will use a different program to generate the citation format.

torchgeo/datasets/esd.py
```diff
@@ -0,0 +1,269 @@
...
+
+    .. note::
+       The dataset and code are available at
+       `GitHub: shuangchencc/ESD <https://github.com/shuangchencc/ESD>`.
```
adamjstewart:
```suggestion
       `GitHub: shuangchencc/ESD <https://github.com/shuangchencc/ESD>`__.
```
Otherwise this won't be a link

torchgeo/datasets/esd.py
```diff
@@ -0,0 +1,269 @@
...
+       dataset : https://data-starcloud.pcl.ac.cn/iearthdata/64
+
+    .. versionadded:: 0.9
+    """
+    quantizer = ESDQuantizer()
+
+    def __getitem__(self, index: GeoSlice) -> Sample:
```
adamjstewart, line 152: This can be replaced by:
```python
def __getitem__(self, index: GeoSlice) -> Sample:
    """..."""
    sample = super().__getitem__(index)
    sample['image'] = self.quantizer.quantize(sample['image'])
    return sample
```
or this could be a separate transform that is applied in a datamodule like other datasets.

torchgeo/datasets/esd.py
```diff
@@ -0,0 +1,269 @@
...
+          
+        # ESD quantize
+        sample['image'] = self.quantizer.quantize(sample['image'])
+
+        if self.transforms is not None:
+            sample = self.transforms(sample)
+
+        return sample
+
+    def _filepath_to_timestamp(self, filepath: Path) -> tuple[datetime, datetime]:
```
adamjstewart, line 168: Not needed if the filename contains all necessary information. You could instead use:
```python
filename_glob = "SDC30_EBD_*"
filename_regex = r".*_(?P<date>\d{4})\."
date_format = '%Y'
```
See https://torchgeo.readthedocs.io/en/stable/tutorials/custom_raster_dataset.html for docs.

torchgeo/datasets/esd.py
```diff
@@ -0,0 +1,269 @@
...
+        Args:
+            vectors: ESD embedding tensor of shape (months, channels, height, width)
+            show_titles: whether to show a title above the image
+            suptitle: optional figure-level title
+
+        Returns:
+            A matplotlib Figure containing the RGB(A) visualization.
+        """
+        vectors = sample['image']
+        print(vectors.shape)
```
adamjstewart: Can remove debugging print statements

adamjstewart:
> Let us know if/when you release the training data. I can also help figure out the licenses of the underlying datasets if you need help with that.

Here are my findings so far:

Screenshot 2026-02-02 at 12 05 13 PM

You can re-release public domain data under a different license if you want. For the CC-BY-NC-3.0-IGO and CC-BY-4.0 datasets, you have to keep the original licenses, but you can also include them and use a more strict license for the collection as long as it satisfies the restrictions of the original licenses (e.g., non-commercial). Personally, I usually just release the data under the same license or most strict license of the original to make things simpler.

tests/data/esd/data.py
heiyuxiaokai: Thanks for reminding. I originally used `CRS.from_epsg(32602)`, but encountered PROJ database version errors in the test environment due to a PROJ / rasterio mismatch.

Because this code is only used to generate mock test data, and to avoid modifying TorchGeo’s dependency constraints or environment setup, the CRS is defined using an explicit PROJ string instead. This provides the same spatial reference without relying on EPSG lookups via proj.db.

Wrong message as:
`Using PROJ DB at: \.conda\envs\torchgeo\Lib\site-packages\pyproj\proj_dir\share\proj ERROR 1: PROJ: proj_create_from_database: \.conda\envs\torchgeo\Lib\site-packages\pyproj\proj_dir\share\proj\proj.db contains DATABASE.LAYOUT.VERSION.MINOR = 4 whereas a number >= 6 is expected. It comes from another PROJ installation.`

torchgeo/datasets/esd.py
heiyuxiaokai: By definition, it is not a method of the transform class but part of the dataset. Only after applying the quantizer does it become the actual dataset, so the former approach may be more suitable.

torchgeo/datasets/esd.py
heiyuxiaokai: Yes, we not only reduce the image size through embedding, but also through quantization. As stated in the description:
‘By transforming multi-sensor satellite observations into compact, quantized latent vectors, ESD reduces the original data volume (~1 PB for a full year of global land surfaces) to approximately 2.4 TB, enabling decadal-scale analysis on standard workstations.

heiyuxiaokai:
I have addressed the issues mentioned earlier and applied fixes using `pre-commit`. Some comments now appear as outdated; I hope you can check them with the changes. For the training data, we may consider releasing it in the future, but it is not publicly available at the moment.
@adamjstewart

tests/data/esd/data.py
adamjstewart: Ah interesting, looks like some kind of conda bug.

adamjstewart:
Can you fix the test failures? Then I'll re-review. Let me know if any of them are confusing and I can explain how to fix them.

heiyuxiaokai:
> Can you fix the test failures? Then I'll re-review. Let me know if any of them are confusing and I can explain how to fix them.

I have addressed most of the test failures, but there are still two remaining that I’m unsure how to fix. Would you be able to offer some advice or guidance? @adamjstewart

docs/api/datasets/geo_datasets.csv
```diff
@@ -13,6 +13,7 @@ Dataset,Type,Source,License,Size (px),Resolution (m)
 `EDDMapS`_,Points,Citizen Scientists,-,-,-
 `EnMAP`_,Imagery,EnMAP,`EnMAP Data License <https://www.enmap.org/data/resources/EnMAP_Data_License.pdf>`_,"1,200x1,200",30
 `EnviroAtlas`_,"Imagery, Masks","NAIP, NLCD, OpenStreetMap","CC-BY-4.0",-,1
+`Embedded Seamless Data`_,Embedding,"Landsat 5/7/8/9, MODIS(Terra)",-,"3,600x3,600",30
```
adamjstewart: Should actually be above EnMAP to be sorted alphabetically.

I believe NASADEM data was also used to produce the embeddings?

License should be CC-BY-4.0?

```suggestion
`Embedded Seamless Data`_,Embedding,"Landsat 5/7/8/9, MODIS Terra, NASADEM",CC-BY-4.0,"3,600x3,600",30
```

docs/api/datasets.rst
```diff
@@ -96,6 +96,11 @@ EnviroAtlas
 
 .. autoclass:: EnviroAtlas
 
+Embedded Seamless Data
+^^^^^^^^
```
adamjstewart:
```suggestion
Embedded Seamless Data
^^^^^^^^^^^^^^^^^^^^^^
```
Same here, should be above EnMAP.

This will fix the failing documentation tests.

tests/datasets/test_esd.py
```diff
@@ -0,0 +1,51 @@
...
+    EmbeddedSeamlessData,
+    IntersectionDataset,
+    UnionDataset,
+)
+
+
+class TestEmbeddedSeamlessData:
+    @pytest.fixture(params=['2024/SDC30_EBD_V001_02VMN_2024_mock.tif'])
+    def dataset(self, request: SubRequest) -> EmbeddedSeamlessData:
+        paths = os.path.join('tests', 'data', 'esd', 'SDC30_EBD_V001', request.param)
```
adamjstewart: No need to parametrize if there's only a single value:

```suggestion
    def dataset(self) -> EmbeddedSeamlessData:
        paths = os.path.join('tests', 'data', 'esd')
```

tests/datasets/test_esd.py
```diff
@@ -0,0 +1,51 @@
...
+        return EmbeddedSeamlessData(paths, transforms=torch.nn.Identity())
+
+    def test_len(self, dataset: EmbeddedSeamlessData) -> None:
```
adamjstewart, line 31: Can we also test the `__getitem__` to make sure everything has the right type? Something like:
```python
    def test_getitem(self, dataset: GoogleSatelliteEmbedding) -> None:                   
        x = dataset[dataset.bounds]                                                      
        assert isinstance(x, dict)                                                       
```

torchgeo/datasets/esd.py
```diff
@@ -0,0 +1,196 @@
...
+from matplotlib import pyplot as plt
+from matplotlib.figure import Figure
+from torch import Tensor
+
+from .geo import RasterDataset
+from .utils import GeoSlice, Path, Sample, disambiguate_timestamp
+
+
+class ESDQuantizer:
+    r"""Decode ESD-encoded quantized indices into continuous embedding vectors.
```
adamjstewart:
```suggestion
    """Decode ESD-encoded quantized indices into continuous embedding vectors.
```
I don't think you have any raw characters in the docstring?

torchgeo/datasets/esd.py
```diff
@@ -0,0 +1,196 @@
...
+        decoded = decoder.apply_transform(torch.from_numpy(ESD_codes.astype(np.int32)))
+
+    .. note::
+        The output retains the channel dimension corresponding to embedding levels.
+        Users can further convert embeddings to visualizations or aggregate them
+        for downstream tasks.
+
+    """
+
+    def __init__(self, levels: tuple[int, ...] = (8, 8, 8, 5, 5, 5)) -> None:
```
adamjstewart:
```suggestion
    def __init__(self, levels: Sequence[int, ...] = (8, 8, 8, 5, 5, 5)) -> None:
```
Both lists and tuples should be fine here

torchgeo/datasets/esd.py
```diff
@@ -0,0 +1,196 @@
...
+        """Initialize the quantization levels for the embedding."""
```
adamjstewart: Can we document the Args?

torchgeo/datasets/esd.py
```diff
@@ -0,0 +1,196 @@
...
+        levels_tensor = torch.tensor(levels, dtype=torch.int32)
+        self._levels = levels_tensor
+
+        basis_tensor = torch.cumprod(
+            torch.tensor([1, *levels[:-1]], dtype=torch.int32), dim=0
+        )
+        self._basis = basis_tensor
+
+    def indices_to_codes(self, indices: Tensor) -> Tensor:
+        """Convert embedding indices to quantized codes."""
```
adamjstewart: Can we document Args and Returns?

torchgeo/datasets/esd.py
```diff
@@ -0,0 +1,196 @@
...
+        indices = rearrange(indices, '... -> ... 1')
+        level_indices = (indices // self._basis) % self._levels
+
+        half = self._levels // 2
+        codes = (level_indices - half) / half
+
+        return codes
+
+    def quantize(self, input: Tensor) -> Tensor:
+        """Quantize the input tensor using predefined levels."""
```
adamjstewart: Args and Returns

torchgeo/datasets/esd.py
```diff
@@ -0,0 +1,196 @@
...
+        return self.indices_to_codes(input).movedim(-1, -3)
+
+
+class EmbeddedSeamlessData(RasterDataset):
+    r"""Embedded Seamless Data (ESD).
+
+    The `Embedded Seamless Data (ESD) <https://arxiv.org/abs/2601.11183>`
```
adamjstewart:
```suggestion
    `Embedded Seamless Data (ESD) <https://arxiv.org/abs/2601.11183>`__
```
Without underscores, this is not a link

torchgeo/datasets/esd.py
```diff
@@ -0,0 +1,196 @@
...
+    designed to overcome the computational and storage challenges of planetary-scale
+    Earth system science. By transforming multi-sensor satellite observations into
+    compact, quantized latent vectors, ESD reduces the original data volume (~1 PB for
+    a full year of global land surfaces) to approximately 2.4 TB, enabling decadal-scale
+    analysis on standard workstations.
+
+    Key features:
+
+    * **Longitudinal Consistency**: Provides a continuous record from 2000 to 2024,
+      harmonized from Landsat 5, 7, 8, 9 and MODIS Terra imagery.
```
adamjstewart: And NASADEM?

torchgeo/datasets/esd.py
```diff
@@ -0,0 +1,196 @@
...
+    * Code: https://github.com/shuangchencc/ESD
+    * Dataset: https://data-starcloud.pcl.ac.cn/iearthdata/64
+
+    .. versionadded:: 0.9
+    """
+
+    quantizer = ESDQuantizer()
+
+    def __getitem__(self, index: GeoSlice) -> Sample:
+        """..."""
```
adamjstewart: Can you fill out the docstring? Can just copy from the base class.

torchgeo/datasets/esd.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def plot(
+        self, sample: Sample, show_titles: bool = True, suptitle: str | None = None
+    ) -> Figure:
+        """Plot a sample from the dataset.
+
+        Args:
+            sample: Sample dict containing 'image' tensor and metadata.
+            show_titles: Whether to show titles on subplots.
+            suptitle: Optional figure suptitle.
```
adamjstewart, line 195: Returns is missing

adamjstewart (changes requested):
Mostly minor documentation changes remaining.

P.S. We're planning a 0.9 release this week and are planning to include this PR. If you get busy and need me to take over and finish the PR, just let me know.

torchgeo/datasets/esd.py
adamjstewart: Removing this will also fix the coverage tests.

docs/api/datasets/geo_datasets.csv
heiyuxiaokai: Yes, NASADEM was also used to produce the embeddings, and the license is CC-BY-4.0.

heiyuxiaokai:
I’ve pushed an updated version addressing the feedback and cleaning up a few docstrings.
Since the remaining changes should be minimal and I may be a bit busy this week, I might need your help to take over and wrap up the PR, especially with the 0.9 release coming up. @adamjstewart

adamjstewart:
@heiyuxiaokai I'm having trouble making changes directly on your main branch. So either you'll have to finish this PR up, or make a new feature branch that I have access to.

I had to resolve merge conflicts, so we need to re-add the documentation changes. Also need to replace `_filepath_to_timestamp` as suggested above.

heiyuxiaokai:
@adamjstewart I've opened a new PR using a separate branch, so you should now have the necessary access to make changes directly. I’ve also replaced the `_filepath_to_timestamp` function as suggested.

PR #3357, opened 2026-02-02
Return dict from SemanticSegmentationTask.predict_step

hkristen:
## Summary
- Changes `predict_step` return type from `Tensor` to `dict[str, Any]`
- Returns `probabilities`, `bounds`, and `transform` keys
- Enables callbacks to access spatial metadata for tiled inference

### DIFF
--- a/torchgeo/trainers/segmentation.py
+++ b/torchgeo/trainers/segmentation.py
@@ -381,7 +381,7 @@ class SemanticSegmentationTask(BaseTask):
 
     def predict_step(
         self, batch: Any, batch_idx: int, dataloader_idx: int = 0
-    ) -> dict[str, Any]:
+    ) -> dict[str, Tensor]:
         """Compute the predicted class probabilities.
 
         Args:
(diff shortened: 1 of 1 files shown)


torchgeo/trainers/segmentation.py
```diff
@@ -390,7 +390,15 @@ def predict_step(
...
 
         Returns:
-            Output predicted probabilities.
+            Dictionary containing:
+                - probabilities: Model predictions (B, C, H, W) with softmax/sigmoid
+                  applied
+                - bounds: Patch bounds tensor (B, 9) if available
+                - transform: Per-patch affine transform (B, 6) if available
+
+        .. versionchanged:: 0.8
```
```suggestion
        .. versionchanged:: 0.9
```

torchgeo/trainers/segmentation.py
```diff
@@ -381,7 +381,7 @@ def test_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) -> None
 
     def predict_step(
         self, batch: Any, batch_idx: int, dataloader_idx: int = 0
-    ) -> Tensor:
+    ) -> dict[str, Any]:
```
```suggestion
    ) -> dict[str, Tensor]:
```

hkristen:
>I think we should probably do the same thing for PixelwiseRegressionTask and ChangeDetectionTask. Those should have very similar pixel-level predictions that could be stitched together. Stitching might look a little different for regression than for classification.

Good point. We could add those in a follow-up PR. ChangeDetectionTask is straightforward copy-paste I believe, for PixelwiseRegressionTask stitching would most likely use averaging instead of probability accumulation. @blaz-r care to jump on this one? ;)

PR #3360, opened 2026-02-03
Add time dimension to TreeSatAI dataset

yichiac:
The dataset doesn't directly have time-series images. It has one timestamp per location. 

Current shape: C x H x W
New shape: 1 x C x H x W


adamjstewart:
[TreeSatAI](https://essd.copernicus.org/articles/15/681/2023/) is not a time series dataset, so there is no reason to make changes here.

[TreeSatAI-TS](https://huggingface.co/datasets/IGNF/TreeSatAI-Time-Series) is a time series version of the dataset. This is worth adding, although not as high priority as converting existing time series datasets.

yichiac:
Okay, I'll close this PR and prioritize other time-series datasets first.

PR #3362, opened 2026-02-03
XarrayDataset: add warning about experimental features

adamjstewart:
@calebrob6 @isaaccorley some kind of warning like this would make me more comfortable about adding experimental features like TileDataset or SedonaDataset. I want to reserve the right to modify, rename, or completely remove these experimental features if we find a better way to integrate them in the future.


isaaccorley:
Sounds good to me. We could also later add an experimental decorator that is printed but this is good enough imo.

adamjstewart:
That's actually a pretty cool idea. Last time I looked into this, decorators were easy to add for functions but kinda hard to add for classes? I tried to recreate the `warnings.deprecated` decorator in older Python versions and it was non-trivial.

PR #3363, opened 2026-02-03
BigEarthNetV2: Correctly pass filename and md5 arguments to download_url()

alecsandrei:
Hi,

For BigEarthNetv2, md5 was passed as the filename argument to download_url. This caused the downloaded files to be named after the md5 hash values. BigEarthNetV2._extract method is failing as it expects the right filenames. This silently passed the tests as torchgeo/tests/data/bigearthnet/v2/ contains the right filenames.

### DIFF
--- a/torchgeo/datasets/bigearthnet.py
+++ b/torchgeo/datasets/bigearthnet.py
@@ -894,7 +894,12 @@ class BigEarthNetV2(NonGeoDataset):
             for fname, md5 in meta['files'].items():
                 target_path = os.path.join(self.root, fname)
                 if not os.path.exists(target_path):
-                    download_url(self.url.format(fname), self.root, md5)
+                    download_url(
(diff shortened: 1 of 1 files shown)


adamjstewart:
Duplicate of #3210. @goergen95 what do you want to do here? We could merge this PR as a short-term fix for the md5 rename bug and then rebase your PR with the bigger fix for separate downloading of s1/s2. Order doesn't matter to me, but want to get the fix in soon before our next release.

goergen95:
It seems reasonable to me to first merge this PR as its quite a minimal change and adresses the main issue. I could then rebase and update my PR to include the separate downloads of s1/s2. The other issue I am addressing is improved handling of interrupted downloads, e.g. when only some but not all required files were downloaded which in the current implementation is not handled correctly.

PR #3367, opened 2026-02-03
Tests: don't verify checksum of fake files

adamjstewart:
In the process of converting all md5 -> sha256, we need to either:

1. Convert all md5 -> sha256 for our test data, or
2. Don't verify the checksum of our test data

I opted for the latter, as:

1. We're able to reduce our test code by 470 lines
2. These lines aren't required for coverage
3. This makes it simpler to add new datasets (less testing required)
4. We monkeypatch download_url, so the checksum check is skipped anyway

We can also remove a similar number of SLOCs from `tests/data/*/data.py`, but that's lower priority. Can be done in a separate PR.

### DIFF
--- a/tests/datasets/test_advance.py
+++ b/tests/datasets/test_advance.py
@@ -23,12 +23,10 @@ class TestADVANCE:
             os.path.join(data_dir, 'ADVANCE_vision.zip'),
             os.path.join(data_dir, 'ADVANCE_sound.zip'),
         ]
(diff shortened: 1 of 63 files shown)


isaaccorley:
I'm cool with the latter as well. It was always super tedious to create checksums for the test data actually, especially if test data gets updated, git tracks the change anyway.

isaaccorley (approved):

PR #3368, opened 2026-02-03
SouthAfricaCropType: add time series support

yichiac:
TODO
- [x] Rename datamodules/southafricacroptype.py to datamodules/south_africa_crop_type.py to keep naming consistent with its dataset.
- [x] Add sample plots
- [x] Check if the test data could be smaller

sample_00_wide


torchgeo/datamodules/south_africa_crop_type.py
```diff
@@ -48,22 +48,34 @@ def __init__(
             **kwargs,
         )
 
-        self.train_aug = K.AugmentationSequential(
+        self._train_video_aug = K.VideoSequential(
             K.Normalize(mean=self.mean, std=self.std),
             K.RandomResizedCrop(_to_tuple(self.patch_size), scale=(0.6, 1.0)),
             K.RandomVerticalFlip(p=0.5),
             K.RandomHorizontalFlip(p=0.5),
```
adamjstewart:
```suggestion
        self.train_aug = K.AugmentationSequential(K.VideoSequential(
K.Normalize(mean=self.mean, std=self.std),
K.RandomResizedCrop(_to_tuple(self.patch_size), scale=(0.6, 1.0)),
K.RandomVerticalFlip(p=0.5),
K.RandomHorizontalFlip(p=0.5)), data_keys=None, keepdim=True,
```

torchgeo/datamodules/south_africa_crop_type.py
```diff
@@ -48,22 +48,34 @@ def __init__(
...
         )
 
         self.aug = K.AugmentationSequential(
-            K.Normalize(mean=self.mean, std=self.std), data_keys=None, keepdim=True
+            K.VideoSequential(K.Normalize(mean=self.mean, std=self.std)),
+            data_keys=None,
+            keepdim=True,
         )
 
+    def on_after_batch_transfer(self, batch: Sample, dataloader_idx: int) -> Sample:
```
adamjstewart: Remove

torchgeo/datasets/south_africa_crop_type.py
```diff
@@ -287,31 +290,35 @@ def plot(
...
-        ncols = 2
+        num_timesteps = images.shape[0]
+        ncols = num_timesteps + 1
 
         showing_prediction = 'prediction' in sample
         if showing_prediction:
             pred = sample['prediction'].squeeze()
             ncols += 1
 
+        fig: Figure
```
adamjstewart: Is this needed?

tests/datasets/test_south_africa_crop_type.py
```diff
@@ -44,9 +44,12 @@ def test_getitem(self, dataset: SouthAfricaCropType) -> None:
         assert isinstance(x, dict)
         assert isinstance(x['image'], torch.Tensor)
         assert isinstance(x['mask'], torch.Tensor)
+        assert x['image'].ndim == 4
+        assert x['image'].shape[0] == 2
+        assert x['image'].shape[1] == len(dataset.bands)
 
     def test_len(self, dataset: SouthAfricaCropType) -> None:
-        assert len(dataset) == 10
+        assert len(dataset) == 20
```
adamjstewart: Does this length make sense? From what I understand, we have 10 locations with 2 timestamps. Shouldn't the length be 10? Otherwise we will duplicate the same location multiple times per epoch.

Any chance we can reduce the number of fields/locations to 1 or 2 to make the fake data smaller?

torchgeo/datasets/south_africa_crop_type.py
adamjstewart, line 284: Can you add an example plot from the real data to the PR description so we can see what this looks like? I'm worried that the plot will be very wide or tall.

adamjstewart (changes requested):
One major bug in dataset length and a few general questions.

torchgeo/datasets/south_africa_crop_type.py
yichiac: It was for mypy. Let me work to see if this can be skipped. It seems redundant.

yichiac:
The updated train_aug and the removal of on_after_batch_transfer cause the test failures

adamjstewart:
I would check the dimensions of image and mask before and after augmentation and see if they make sense. We use VideoSequential for all other time series datasets, I don't see anything special about this one that would explain why it wouldn't work.

torchgeo/datasets/south_africa_crop_type.py
yichiac: The plots are indeed too wide to be readable. I tried plotting them on grids, but they still look a bit too much. Not sure how to properly handle this 75 timestamps vs 1 mask visualization. 

Original
sample_00_wide

Grids
sample_00_grid

tests/datasets/test_south_africa_crop_type.py
yichiac: I reduced the amount of fake data. We can't do 1 or 2 since the train/val/test splits are 0.8, 0.1, 0.1. So I did 5 locations.

torchgeo/datamodules/southafricacroptype.py
```diff
@@ -48,20 +46,10 @@ def __init__(
             **kwargs,
         )
 
-        self.train_aug = K.AugmentationSequential(
-            K.Normalize(mean=self.mean, std=self.std),
-            K.RandomResizedCrop(_to_tuple(self.patch_size), scale=(0.6, 1.0)),
-            K.RandomVerticalFlip(p=0.5),
-            K.RandomHorizontalFlip(p=0.5),
```
adamjstewart, line 55: Can we keep these augs to minimize backwards-incompatible changes?

torchgeo/datasets/south_africa_crop_type.py
```diff
@@ -72,10 +72,10 @@ class SouthAfricaCropType(RasterDataset):
 
     url = 'https://radiantearth.blob.core.windows.net/mlhub/ref-south-africa-crops-competition-v1'
 
-    filename_glob = '*_07_*_{}_10m.*'
+    filename_glob = '*_{}_10m.*'
```
adamjstewart, line 75: ~I'm confused about one thing. SouthAfricaCropType is just a normal RasterDataset. The `__len__` method should return the length of the index. With this change, the index should not contain all files for all timestamps. But we instead want a list of all unique locations, not a list of all files. Won't this change cause the length of the dataset to increase? How are the dataset length tests passing? Did we change the number of test files to only have a single image per timestamp?~

EDIT: I understand now. We need to change the tests to assert that the length of the dataset should be 5. But we also need to make changes in the dataset class to ensure that this is true. If the months are always the same, we can just pick a single month to be in the glob. If not, we will have to dynamically populate the index (override init) or something more complex (override len).

tests/datasets/test_south_africa_crop_type.py
adamjstewart: 5 sounds good. But you also need to change the expected length of the dataset to 5. That's why the tests are passing even though they should fail.

torchgeo/datasets/south_africa_crop_type.py
adamjstewart: For now let's keep plotting a single image like we did before. I have some plans for better handling plotting of time series datasets. I want to explore animated GIFs for these. But that's for a future PR.

PR #3369, opened 2026-02-03
Substation: return time series by default, correct S2 plotting

yichiac:
## Summary

This dataset already supports time series with different temporal aggregations. This PR restructures the return format to T×C×H×W and fixes the plotting function.

## Changes

- **Return shape T×C×H×W**: Enables compatibility with temporal learning pipelines that expect an explicit time dimension.
- **Fix plot RGB order**: Corrected channel ordering for proper RGB display.
- **Fix plot S2 normalization**: The original plotting function normalized with `/ 255.0`, but Sentinel-2 values range from 0–10,000, resulting in mostly blank plots. Updated to use the correct scale.

## Before / After

**Before** (blank due to incorrect normalization):

substation_sample_0

**After**:

substation_sample_0
substation_ts_sample_2

## Checklist

- [x] Returns T×C×H×W
- [x] Fix plot: RGB order
- [x] Fix plot: S2 scale

### DIFF
--- a/torchgeo/datasets/substation.py
+++ b/torchgeo/datasets/substation.py
@@ -66,7 +66,7 @@ class Substation(NonGeoDataset):
         mask_2d: bool = True,
         num_of_timepoints: int = 4,
         timepoint_aggregation: Literal['concat', 'median', 'first', 'random']
-        | None = 'concat',
+        | None = None,
         transforms: Callable[[Sample], Sample] | None = None,
         download: bool = False,
         checksum: bool = False,


torchgeo/datasets/substation.py
```diff
@@ -75,6 +76,7 @@ def __init__(
             mask_2d: Whether to use a 2D mask.
             num_of_timepoints: Number of timepoints to use for each image.
             timepoint_aggregation: How to aggregate multiple timepoints.
+                If None, returns time-series as T x C x H x W.
```
adamjstewart, line 82: I wonder if we should remove this parameter and always return a full time series. All other aggregation strategies are more of a research decision, not a common dataset usage. Unless the official dataset paper/code also has these comparisons?

torchgeo/datasets/substation.py
```diff
@@ -131,6 +133,9 @@ def __getitem__(self, index: int) -> Sample:
                 image = image[0]
             case 'random':
                 image = image[np.random.randint(image.shape[0])]
+            case None:
+                # preserve T x C x H x W
+                pass
```
adamjstewart: This code doesn't do anything, could be removed.

torchgeo/datasets/substation.py
adamjstewart, line 164: Can you include an example plot in the PR description so we can see what changed?

torchgeo/datasets/substation.py
yichiac: The official paper and code support different temporal aggregations. 
https://github.com/Lindsay-Lab/substation-seg/blob/fad4a97a2b5060d1aa1bc702f5c66d2af649566a/dataloader.py#L390

torchgeo/datasets/substation.py
```diff
@@ -161,49 +166,89 @@ def plot(
...
+                .transpose(0, 2, 3, 1)
+            )
+            num_images = min(len(images), 2)
+            ncols = num_images + 1
         else:
-            image = sample['image'][:3].permute(1, 2, 0).cpu().numpy()
-        image = image / 255.0
+            image = (
+                torch.clamp(sample['image'][rgb_indices] / 4000, min=0, max=1)
+                .permute(1, 2, 0)
```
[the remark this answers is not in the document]
yichiac: Setting `/ 10000` makes the RGB images dim. The paper used ` / 4000` as a scaling constant. For visualization, `/ 4000` is brighter and more readable.

torchgeo/datasets/substation.py
adamjstewart: Let's default to None for consistency with all other datasets:

```suggestion
        | None = None,
```

torchgeo/datasets/substation.py
```diff
@@ -161,49 +166,89 @@ def plot(
     ) -> Figure:
         """Plot a sample from the dataset.
 
+        When the image is 4D (T x C x H x W), the first two timepoints are plotted.
```
adamjstewart, line 169: Whatever is probably fine for now.

In the long-term, a neat idea I'm considering is plotting all time series as animated GIFs.

torchgeo/datasets/substation.py
adamjstewart: Could move this to the datamodule, but I'm also fine with keeping things simple for now.

adamjstewart:
Next bug: the data module does not support `timepoint_aggregation = None`, and we never actually test this.

PR #3372, opened 2026-02-03
SSL4EO: add time series support, update trainers

yichiac:
The existing `__getitem__` returns `TCxHxW`. It's updated to `TxCxHxW`. We also need to update pretraining trainer, including `moco`, `byol`, and `simclr`. I will update `moco` first in this PR, and we can update others later.

- [x] update SSL4EOS12 dataset to support time series
- [x] update SSL4EOL dataset to support time series
- [x] update MoCo, BYOL, SimCLR trainer to support time series


yichiac:
It seems that only updating `moco` makes the tests for `byol` and `simclr` fail. The `training_step` in `byol` and `simclr` are identical to `moco`'s. If `moco` update is okay, we can just implement the changes in `byol` and `simclr`

yichiac:
tagging SSL4EOS12 author @wangyi111

torchgeo/trainers/moco.py
```diff
@@ -185,6 +186,12 @@ def __init__(
             gather_distributed: Gather negatives from all GPUs during distributed
                 training (ignored if memory_bank_size > 0).
             size: Size of patch to crop.
+            season: Season selection mode for temporal (5D) input.
```
adamjstewart: New parameter not required. Definitely _could_ add something like this, but I think the prior seasonal contrast strategy should work best for any multi-time series datasets.

torchgeo/trainers/moco.py
```diff
@@ -380,14 +389,32 @@ def training_step(
...
+            elif season == 'fixed':
+                g = torch.Generator()
+                g.manual_seed(42)
+                idx = torch.randint(t, (1,), generator=g).item()
+                x1 = x2 = x[:, idx]
+            elif season == 'random':
+                idx = torch.randint(t, (1,)).item()
+                x1 = x2 = x[:, idx]
+
+        else:  # (B, TC, H, W)
```
adamjstewart: I think we should assume that the input is either BxTxCxHxW (for time series datasets) or BxCxHxW (for non-time series datasets). TorchGeo will no longer support BxTCxHxW after all of your PRs are merged.

torchgeo/trainers/moco.py
```diff
@@ -380,14 +389,32 @@ def training_step(
...
+            if season == 'augment':
+                x1 = x[:, torch.randint(t, (1,)).item()]
+                x2 = x[:, torch.randint(t, (1,)).item()]
+            elif season == 'fixed':
+                g = torch.Generator()
+                g.manual_seed(42)
+                idx = torch.randint(t, (1,), generator=g).item()
+                x1 = x2 = x[:, idx]
+            elif season == 'random':
+                idx = torch.randint(t, (1,)).item()
```
adamjstewart: Not reproducible without a random seed.

adamjstewart (changes requested):
Let's try to make this simpler. Once MoCo looks good, we can make the same changes to the other SSL trainers and datasets. All will need to happen in a single PR to make the tests pass.

torchgeo/trainers/moco.py
yichiac: If the random seed is given, 'random' will look exactly like 'fixed', which is not really random. The original code doesn't add a random seed to the random mode: https://github.com/zhu-xlab/SSL4EO-S12/blob/2156913c5d8e5a2c572a5b000f0d5eaed6fc3192/src/benchmark/pretrain_ssl/pretrain_moco_v2_s2c.py#L144

adamjstewart: We don't want the same random seed during every step, but at least once when the trainer is first initialized. Maybe lightning already takes care of this for us, I don't know.

torchgeo/trainers/byol.py
```diff
@@ -360,22 +360,17 @@ def training_step(
...
         batch_size = x.shape[0]
 
-        in_channels = self.hparams['in_channels']
-        assert x.size(1) == in_channels or x.size(1) == 2 * in_channels
-
-        if x.size(1) == in_channels:
+        if x.ndim == 5:  # (B, T, C, H, W)
+            t = x.shape[1]
+            x1 = x[:, torch.randint(t, (1,)).item()]
+            x2 = x[:, torch.randint(t, (1,)).item()]
```
adamjstewart: This seems to have the possibility that x1 and x2 could have the same timestep. We want random sampling without replacement here.

yichiac:
Changing trainers (byol, moco, and simclr) would also affect `seco.py`, so `elif x.ndim == 4` was added for seco. Once we update seco to 5D in the next PR, we can remove the condition.

torchgeo/datasets/ssl4eo.py
```diff
@@ -277,7 +277,7 @@ def __getitem__(self, index: int) -> Sample:
                 wavelengths.extend(self.wavelengths)
 
         sample = {
-            'image': torch.cat(images),
+            'image': torch.stack(images),
```
adamjstewart: If seasons == 1, I think we should return CxHxW instead of 1xCxHxW. All of our currently SSL trainers are designed for time series, but there are plans to add MAE, which would probably require CxHxW.

torchgeo/trainers/moco.py
```diff
@@ -378,16 +378,24 @@ def training_step(
...
+            else:
+                idx = torch.randperm(t)[:2]
+            x1 = x[:, idx[0]]
+            x2 = x[:, idx[1]]
+        elif x.ndim == 4 and x.shape[1] > in_channels:  # (B, T*C, H, W)
+            t = x.shape[1] // in_channels
+            idx = torch.randperm(t)[:2]
+            x1 = x[:, idx[0] * in_channels : (idx[0] + 1) * in_channels]
+            x2 = x[:, idx[1] * in_channels : (idx[1] + 1) * in_channels]
+        else:  # (B, C, H, W)
```
adamjstewart: I would like to remove support for these, so the trainer assumes (or asserts) that the input is always Bx2xCxHxW. We will later add other trainers like MAE that support BxCxHxW. This should greatly simplify testing.

adamjstewart (changes requested):
Let's keep hacking on SSL4EO and MoCo until they look good. Don't worry about whether the tests pass. Once they look good, we'll make the same changes to SeCo and BYOL/SimCLR in the same PR so that all tests pass. We can remove the BYOL/MoCo/SimCLR seasons=1 tests.

torchgeo/trainers/moco.py
adamjstewart: @MathiasBaumgartinger @DarthReca since you're actually using these trainers, do you have any opinions on this? Are you always using temporal datasets with Bx2xCxHxW, or do you have your own datasets where you want support for BxCxHxW?

MathiasBaumgartinger: Currently, I am not at all using temporal datasets.

adamjstewart: For MAE or for MoCo/SimCLR/BYOL? I think the latter benefits a lot from seasonal augmentations, but maybe it shouldn't be required. There are plenty of pretraining datasets without time series...

DarthReca: I am using the plain BxCxHxW right now, and I do not know how common this could be

adamjstewart: Okay, then in that case let's support both BxCxHxW and Bx2xCxHxW. I don't think we should support Bx2CxHxW or any other arbitrary BxTxCxHxW.

torchgeo/datasets/ssl4eo.py
yichiac: supporting both CxHxW and TxCxHxW would make things complicated...

yichiac:
not sure why `ty check` picked up errors from files I didn't change.

adamjstewart:
Don't worry about the warnings, those can be solved in a separate PR. The errors are in files you modified though.

adamjstewart:
Can you rebase? We renamed trainers to tasks.

PR #3374, opened 2026-02-04
TorchGeo Tutorial: add more slicing options

adamjstewart:
Advertises the flexibility and power of our new slicing syntax. I don't think this is documented anywhere else, but all 3 of these should be supported.


calebrob6 (approved):

calebrob6:
Conditionally approved on "did you run this to check to see if it works even though it totally should"

adamjstewart:
It passes CI 🙂

PR #3377, opened 2026-02-04
Updating ESW notebook with new tutorial content

calebrob6:
Shows how to save/load model and run gridded prediction on a new Sentinel-2 scene. This pretty much does everything we don't like in a notebook:
- [x] Requires network access
- [x] Longish running computations
- [x] Saves big files
- [x] pip install all the things

### DIFF
--- a/docs/tutorials/earth_surface_water.ipynb
+++ b/docs/tutorials/earth_surface_water.ipynb
@@ -1133,6 +1133,23 @@
   {
    "cell_type": "code",
    "execution_count": null,
+   "metadata": {
+    "nbmake": {
+     "mock": {
+      "inference_batch_size": 4,
+      "max_tiles": 4
+     }
+    }
+   },
+   "outputs": [],
+   "source": [
+    "inference_batch_size = 16\n",
+    "max_tiles = None"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -1146,7 +1163,7 @@
    "outputs": [],
    "source": [
     "s2_dataloader = DataLoader(\n",
-    "    s2_dataset, sampler=grid_sampler, batch_size=16, collate_fn=stack_samples\n",
+    "    s2_dataset, sampler=grid_sampler, batch_size=inference_batch_size, collate_fn=stack_samples\n",
     ")\n",
     "\n",
     "model.eval()\n",
@@ -1156,6 +1173,9 @@
     "\n",
     "with torch.no_grad():\n",
     "    for batch_idx, batch in enumerate(tqdm(s2_dataloader)):\n",
+    "        if max_tiles is not None and len(predictions) >= max_tiles:\n",
+    "            break\n",
+    "\n",
     "        X = tfms(batch['image']).to(device)\n",
     "\n",
     "        # Run model\n",
--- a/docs/tutorials/earth_surface_water.ipynb
+++ b/docs/tutorials/earth_surface_water.ipynb
@@ -74,7 +74,7 @@
    },
    "outputs": [],
    "source": [
-    "%pip install torchgeo scikit-learn"
+    "%pip install torchgeo scikit-learn planetary-computer pystac"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


docs/tutorials/earth_surface_water.ipynb
```diff
@@ -965,6 +984,302 @@
...
+    "print(f'  Total tiles: {len(grid_sampler)}')"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "model = load_model(\"deeplabv3_surface_water.pth\")"
```
adamjstewart: This model is already loaded, correct? Why do we need to load from disk?

docs/tutorials/earth_surface_water.ipynb
```diff
@@ -965,6 +984,302 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "item_url = 'https://planetarycomputer.microsoft.com/api/stac/v1/collections/sentinel-2-l2a/items/S2C_MSIL2A_20260201T130241_R095_T23KPQ_20260201T153209'\n",
+    "item = pystac.Item.from_file(item_url)\n",
```
adamjstewart, line 1023: If we redistribute these files on HF, we won't need planetary_computer/pystac and we won't be dependent on PC for our tutorial tests to pass. We can also use a smaller cropped file so we don't run out of time/memory during testing.

docs/tutorials/earth_surface_water.ipynb
```diff
@@ -965,6 +984,302 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Download each band and organize into R10m/ and R20m/ subdirectories in\n",
+    "# a name expected by the Sentinel2 dataset\n",
```
adamjstewart, line 1061: I don't think this is actually required anymore after #3043

docs/tutorials/earth_surface_water.ipynb
```diff
@@ -965,6 +984,302 @@
...
+    "    if dst_row_start >= 0 and dst_col_start >= 0:\n",
+    "        mask_crop = mask[src_row_start:src_row_end, src_col_start:src_col_end]\n",
+    "        prediction_sum[dst_row_start:dst_row_end, dst_col_start:dst_col_end] += mask_crop\n",
+    "        prediction_count[dst_row_start:dst_row_end, dst_col_start:dst_col_end] += 1\n",
+    "\n",
+    "# Average predictions and threshold at 0.5\n",
+    "prediction_count = np.maximum(prediction_count, 1)  # Avoid division by zero\n",
+    "full_prediction = (prediction_sum / prediction_count >= 0.5).astype(np.uint8)\n",
+    "\n",
+    "print(f'Reconstructed prediction mosaic: {full_prediction.shape}')"
```
adamjstewart, line 1266: Hopefully all of this will be obsolete once #3214 is merged.

adamjstewart (commented):
Main comments:

1. A bunch of non-TorchGeo dependencies loaded, we should decide if we actually require these or we can get away without them.
2. Timeout likely due to large Sentinel-2 tiles, slow downloads, and limited storage/RAM. We could store a cropped Sentinel-2 tile on HF to solve all of these problems and reduce dependencies.

@nilsleh would you have time to work on this? I'm swamped reviewing other PRs for this release.

docs/tutorials/earth_surface_water.ipynb
calebrob6: So you can train once and come back and play with inference in different places later

docs/tutorials/earth_surface_water.ipynb
calebrob6: I know but showing how to do it on PC is important because it shows people how to run the water model for wherever they want (I originally had a full search example here). I don't _hate_ having this commented out and some precut S2 scene on HF, but its kind of silly!

docs/tutorials/earth_surface_water.ipynb
calebrob6: It was definitely broken in whatever version I was using

docs/tutorials/earth_surface_water.ipynb
calebrob6: Hopefully :)

calebrob6:
installing sklearn, planetary-computer and pystac isn't a very big lift

calebrob6:
@copilot can you fix CI? The tutorials/notebooks check is failing because of a cell timeout. Look at https://raw.githubusercontent.com/torchgeo/torchgeo/refs/heads/main/docs/tutorials/custom_segmentation_trainer.ipynb as an example of how to use nbmake mock to override variables in a cell during CI, copying the relevant bit to look at here, too: "nbmake": {
     "mock": {
      "batch_size": 1,
      "fast_dev_run": true,
      "max_epochs": 1,
      "num_workers": 0,
      "use_pretrained": false
     }
    }

isaaccorley (approved):

PR #3379, opened 2026-02-05
Swap Sentinel-2 TOA and SR number of channels in SSL4EO-S12

DarthReca:
I spotted this error in the documentation of SSL4EO-S12. 

I think SR should be 12 bands and TOA 13 bands. The band lists seems correctly assigned instead.


adamjstewart:
Is this still a draft or can I merge?

DarthReca:
It is ready. I was just double-checking if I am understanding correctly

PR #3385, opened 2026-02-08
Docs: restructure datasets API into individual pages

ashnair1:


docs/api/datasets/aboveground-woody-biomass.rst
```diff
@@ -0,0 +1,7 @@
+.. _Aboveground Woody Biomass:
```
adamjstewart, line 1: Curious why we didn't need this before and we still don't need it for the models but we do need this now for the datasets?

docs/api/datasets/copernicus-bench.rst
```diff
@@ -0,0 +1,29 @@
+Copernicus-Bench
+================
+
+Copernicus-Bench is a comprehensive evaluation benchmark with 15 downstream tasks hierarchically organized across preprocessing (e.g., cloud removal), base applications (e.g., land cover classification), and specialized applications (e.g., air quality estimation). This benchmark enables systematic assessment of foundation model performances across various Sentinel missions on different levels of practical applications.
+
+.. csv-table:: C = classification,  R = regression, S = semantic segmentation, T = time series, CD = change detection, E = embedding
```
adamjstewart: Would like to keep this table on the main page similar to nongeo and geo tables. Will eventually add more tables myself later.

docs/api/datasets/spacenet.rst
```diff
@@ -0,0 +1,21 @@
+SpaceNet
+========
+
+The `SpaceNet Dataset <https://spacenet.ai/datasets/>`_ is hosted as an Amazon Web Services (AWS) `Public Dataset <https://registry.opendata.aws/spacenet/>`_. It contains ~67,000 square km of very high-resolution imagery, >11M building footprints, and ~20,000 km of road labels to ensure that there is adequate open source data available for geospatial machine learning research. SpaceNet Challenge Dataset's have a combination of very high resolution satellite imagery and high quality corresponding labels for foundational mapping features such as building footprints or road networks.
+
+.. csv-table:: I = instance segmentation
```
adamjstewart: Same here, let's keep this table and possibly the above description in the main datasets page and only document the classes here. Or duplicate them?

docs/api/datasets/copernicus-embed.rst
```diff
@@ -0,0 +1,7 @@
+.. _Copernicus-Embed:
+
+Copernicus-Embed
```
adamjstewart, line 3: Can't decide whether Copernicus-{Bench,Embed,Pretrain} should all be on a single documentation page... We group them in a sub-directory, so it's a borderline case for sure.

docs/api/datasets/non_geo_datasets.csv
```diff
@@ -1,77 +1,77 @@
...
+:ref:`RESISC45`,C,Google Earth,"CC-BY-NC-4.0","31,500",45,256x256,0.2--30,RGB
+:ref:`Rwanda Field Boundary`,S,Planetscope,"NICFI AND CC-BY-4.0",70,2,256x256,4.7,RGB + NIR
+:ref:`SatlasPretrain`,"C, R, S, I, OD","NAIP, Landsat, Sentinel",ESA AND CC0-1.0 AND ODbL-1.0 AND CC-BY-4.0,302M,137,512,0.6--30,"SAR, MSI"
+:ref:`Seasonal Contrast`,T,Sentinel-2,"CC-BY-4.0",100K--1M,-,264x264,10,MSI
+:ref:`SeasoNet`,S,Sentinel-2,"CC-BY-4.0","1,759,830",33,120x120,10,MSI
+:ref:`SEN12MS`,S,"Sentinel-1/2, MODIS","CC-BY-4.0","180,662",33,256x256,10,"SAR, MSI"
+:ref:`SKIPP'D`,R,"Fish-eye","CC-BY-4.0","363,375",-,64x64,-,RGB
+:ref:`SkyScript`,IC,"NAIP, orthophotos, Planet SkySat, Sentinel-2, Landsat 8--9",MIT,5.2M,-,100--1000,0.1--30,RGB
+:ref:`So2Sat`,C,Sentinel-1/2,"CC-BY-4.0","400,673",17,32x32,10,"SAR, MSI"
+:ref:`SODA`,OD,Aerial,"CC-BY-NC-4.0","2513",9,"~2700x~4800",-,"RGB"
```
adamjstewart, line 61: Good catch. This seems to happen quite often. Trying to brainstorm ways to prevent this. Ideas so far:

* Unit test that reads CSV files with `csv` and asserts the same length for each line
* Unit test that reads CSV files with `pandas`, writes with pandas, and asserts identical files. Also catches things like unnecessary quotes
* Adopt an auto-formatter similar to prettier or ruff to check for this

docs/api/datasets/aboveground-woody-biomass.rst
ashnair1: These labels are necessary for the clickable `:ref:` links from the CSV tables. I could remove it along with the `:ref:` syntax from the CSVs but then it'll be just plain text.

adamjstewart: Nah, we definitely need links to access the pages. Why do we need `:ref:` instead of `foo_` like before?

ashnair1: `foo_` doesn't work across files,  `:ref:` does. Using `foo_` throws the following error

```
api/datasets/geo_datasets.csv:1: ERROR: Unknown target name: "aboveground woody biomass". [docutils]
```
Refer to https://www.sphinx-doc.org/en/master/usage/referencing.html#role-ref

adamjstewart: Ah, and the toctree auto-generates links for the models, so that's why they don't need it. Thanks!

docs/api/datasets/non_geo_datasets.csv
ashnair1: Unit tests could work. Alternatively we could use [CleverCSV](https://github.com/alan-turing-institute/CleverCSV). It has a CLI and we could use the [standardize](https://clevercsv.readthedocs.io/en/latest/_readme.html#standardize) command for auto-formatting but it'll be one more dependency.

adamjstewart: Let's worry about this another day, there are enough things to worry about already.

adamjstewart:
Going to merge before we merge several other datasets and end up in rebasing hell.

Would still like to deal with the above question in a follow-up PR:

> One question: do people like the left-hand menu that lists all datasets? I find it really annoying because if you want to see models, you either have to minimize datasets or scroll forever. Can we either minimize by default or remove datasets from the left-hand menu?

adamjstewart:
Can someone check https://torchgeo.readthedocs.io/en/latest/user/contributing.html#datasets and https://torchgeo.readthedocs.io/en/latest/tutorials/contribute_non_geo_dataset.html#Documentation-Entries and update them if needed?

PR #3399, opened 2026-02-09
ty: ignore incorrect Path type suggestions

adamjstewart:
We're getting into random type hint bug fix territory. I'm not convinced that pathlib.Path is even possible here, but I don't have the energy to fight ty on this one.


adamjstewart:
Seems like mypy and ty disagree on this one. I'm leaning towards ignoring instead of fighting.

isaaccorley (dismissed):

PR #3402, opened 2026-02-09
Add ACE2 climate model

cnagda:


adamjstewart:
Naming-wise, I would like the files to be named `ace.py` and the model to remain ACE2. We probably won't, but if anyone ever wants to add ACE1, it would go in the same file.

PR #3405, opened 2026-02-10
Release tests: correct socket flag

adamjstewart:
What's the opposite of `--disable-socket`?

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -59,7 +59,7 @@ jobs:
       - name: List dependencies
         run: uv pip list
       - name: Run notebook checks
-        run: pytest --nbmake --durations=10 docs/tutorials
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -33,8 +33,8 @@ jobs:
         run: uv pip list
       - name: Run integration checks
         run: |
-          pytest -m slow --enable-socket --durations=10 tests/samplers
-          pytest -m slow --enable-socket --durations=10 tests/trainers
+          pytest -m slow --force-enable-socket --durations=10 tests/samplers
+          pytest -m slow --force-enable-socket --durations=10 tests/trainers
```
adamjstewart, line 37: I'm not even sure if these need to download, maybe only tests/models? Maybe keep the flag here and figure out a way to re-add slow model tests to CI without running out of storage.

isaaccorley (approved):

PR #3409, opened 2026-02-12
Documentation: redesign and reorganize dataset tables

adamjstewart:
TorchGeo now has many different kinds of datasets (140+ in total), and our previous tables do not well capture this diversity. This PR completely reorganizes all tables to make it easier to understand which datasets to use for which purposes.

### Before

We previously only differentiated between:

* NonGeoDatasets
* GeoDatasets
* Copernicus-Bench
* SpaceNet

### After

We now differentiate between:

* Benchmark datasets
  * Copernicus-Bench
  * SpaceNet
* Pre-training datasets
* Embeddings datasets
* Image sources
* Mask labels
* Toy datasets

Each table can have its own columns and captions. This should make it easier to expand these tables in the future. Tables are now consistently sorted alphabetically, and the license column has been moved to the end.

Other "nice to have" features I'm thinking about for follow-up PRs:

- [ ] Add a "domain" column for listing datasets for "agriculture", "forestry", etc.
- [ ] Use | instead of , as a separator so we never need to escape commas
- [ ] Add autoformatting or unit testing to ensure CSVs remain sorted, with equal length rows, without extraneous quoting
- [ ] Add ability to sort tables by any column

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -4,7 +4,7 @@ API Reference
 Complete API documentation for all TorchGeo modules.
 
 .. toctree::
-   :maxdepth: 0
+   :maxdepth: 2
 
    datamodules
    datasets


docs/api/datasets/pretraining.csv
```diff
@@ -0,0 +1,11 @@
+Dataset,Task,Source,License,# Samples,# Classes,Size (px),Resolution (m),Bands
```
adamjstewart: Pretraining datasets are likely the most controversial, as many can be used for either benchmarking or pretraining. I tried to limit this to curated datasets with no labels (e.g., SeCo, SSL4EO) or with 1M+ images (e.g., SatlasPretrain, MMEarth). Completely open to suggestions to add/remove from this list.

docs/api/datasets.rst
```diff
@@ -3,187 +3,95 @@ torchgeo.datasets
 
 .. module:: torchgeo.datasets
 
-In :mod:`torchgeo`, we define two types of datasets: :ref:`Geospatial Datasets` and :ref:`Non-geospatial Datasets`. These abstract base classes are documented in more detail in :ref:`Base Classes`.
+.. toctree::
+   :maxdepth: 0
```
adamjstewart, line 7: Hoping this will prevent a massive list from appearing in the left-hand column, making it difficult to scroll to models without first minimizing datasets.

docs/api/datasets.rst
```diff
@@ -3,187 +3,95 @@ torchgeo.datasets
...
+   :hidden:
+   :glob:
```
adamjstewart, line 9: Praying that this works and we no longer need to update this list.

docs/api/datasets/embeddings.csv
```diff
@@ -0,0 +1,11 @@
+Dataset,Kind,Spatial Extent,Spatial Resolution,Temporal Extent,Temporal Resolution,Dimensions,Dtype,License
```
adamjstewart, line 1: Unlike our other tables that are sorted alphabetically, I tried to sort this similar to our table in https://arxiv.org/abs/2601.13134, first by "kind" and then by date of release. This list happens to be complete, and we are reasonably confident in release dates. I'm open to suggestions to change this to alphabetical if we want.

Is Dtype useful? Would image source be more useful? Tried to match Table 1 from the above paper, but open to suggestions to change columns too.

docs/api/datasets.rst
adamjstewart: Did not work. Any other suggestions?

docs/api/datasets.rst
adamjstewart: It worked!

docs/api/datasets.rst
adamjstewart: Fixed by changing settings in `conf.py`.

PR #3425, opened 2026-02-15
Docs: add changelog page using sphinx-github-changelog

ashnair1:
## Summary by CodeRabbit

* **Documentation**
  * Added a new Changelog page to the documentation, providing an auto-generated changelog linked to GitHub Releases. Users can now easily access a comprehensive release history and track all updates in one convenient location.

* **Chores**
  * Added sphinx-github-changelog dependency to enable automatic changelog generation from GitHub release data.


adamjstewart:
For the record, I like the idea if you can get it working. We could also copy-n-paste everything in a .md file if needed.

ashnair1:
@adamjstewart This PR will require setting up a Github token (`SPHINX_GITHUB_CHANGELOG_TOKEN`) to read the releases. I've created a token under torchgeo but it needs approval

image

.readthedocs.yaml
```diff
@@ -18,6 +18,8 @@ build:
       - uv venv "${READTHEDOCS_VIRTUALENV_PATH}"
     install:
       - UV_PROJECT_ENVIRONMENT="${READTHEDOCS_VIRTUALENV_PATH}" uv sync --locked --extra docs
+    pre_build:
+      - 'echo "Token present: ${SPHINX_GITHUB_CHANGELOG_TOKEN:+yes}"'
```
adamjstewart: Does this expose secrets in the CI logs?

pyproject.toml
```diff
@@ -126,6 +126,8 @@ docs = [
     # sphinx 5.3+ required to allow section titles inside autodoc class docstrings
     # https://github.com/sphinx-doc/sphinx/pull/10887
     "sphinx>=5.3",
+    # sphinx-github-changelog 1.0+ required for viewing release notes in docs
```
adamjstewart: These comments are to explain why 1.0.0+, not why sphinx-github-changelog. Did you test locally with 1.0.0 to confirm it works? Seems quite old.

docs/changelog.rst
adamjstewart, line 1: Maybe we could instead put this under User Documentation? Doesn't seem important enough for its own personal top-level page.

adamjstewart (commented):
Very cool! A few images don't render, but maybe we could fix those in the old release notes. Math also doesn't render, but we could remove that if we need to, it would just be unfortunate. Overall I'm surprised the rendering isn't better, it's scraping HTML to build HTML, right?

.readthedocs.yaml
ashnair1: Nope. It just prints `Token present:yes` if token exists or `Token present:` if it doesn't

adamjstewart: If this was just for debugging maybe we can remove it now.

adamjstewart (dismissed):
Only remaining bugs are images and math that won't render. I would suggest manually fixing the images in the release notes and reporting the math bug to the developers. If they decide not to support math, we can consider replacing the math in the release notes.

Will leave open in case others want to review, but feel free to merge after a week if the above is fixed and you're reasonably happy.

ashnair1:
> Only remaining bugs are images and math that won't render. I would suggest manually fixing the images in the release notes and reporting the math bug to the developers. If they decide not to support math, we can consider replacing the math in the release notes.
> 
> Will leave open in case others want to review, but feel free to merge after a week if the above is fixed and you're reasonably happy.

The problem is that images uploaded via GitHub's drag-and-drop (`github.com/user-attachments/assets/`) get converted to signed URLs with JWT tokens that expire after 5 minutes. When `sphinx-github-changelog` fetches releases via API, GitHub returns these time-limited `private-user-images.githubusercontent.com` URLs instead of permanent ones. 

Let me see if I can fix this

adamjstewart:
Some images can probably be found externally. Is it only images uploaded in the release notes or also images uploaded in PRs?

ashnair1:
> Some images can probably be found externally. Is it only images uploaded in the release notes or also images uploaded in PRs?

You mean the drag and drop images getting converted to time limited URLs? Think this might be the case for both. 

But we only care about images in release notes right? `sphinx-github-changelog` only fetches GitHub Releases content, not PR descriptions. 

I've added a post-build script that replaces the expiring signed URLs with permanent `github.com/user-attachments/assets/{uuid}` URLs which should fix the issue with images. You can check in the latest build of the docs.

It's either this or adding all images ever used in release notes to `docs/_static/images` folder.

adamjstewart:
Could this URL replacement be submitted upstream so we don't have to maintain it?

ashnair1:
> Could this URL replacement be submitted upstream so we don't have to maintain it?

Most likely. Will revert once I have an update.

ashnair1:
Was able to upstream the fix and it made it into the latest release - https://github.com/ewjoachim/sphinx-github-changelog/releases/tag/1.7.2

Images should be displayed properly now.

adamjstewart (approved):
Looks great now! There are still things that aren't perfect, like no syntax highlighting for code, no colored background for notes/warnings, no math formatting, etc. But I think we can open issues on GitHub if we care about those, I don't think it should hold up this PR.

Not sure if anyone else has strong feelings about adding a new dependency. I'm fine with it since it isn't user facing, but we can easily drop it if it ever becomes unmaintained.

PR #3432, opened 2026-02-16
Trainers: batch Any -> Sample

adamjstewart:
Mypy used to complain about this not being compatible with the superclass. Let's see if that is still the case, ty is happy with this and it reduces a lot of our Any usage.

### DIFF
--- a/tests/trainers/test_segmentation.py
+++ b/tests/trainers/test_segmentation.py
@@ -12,6 +12,7 @@ import torch
 import torch.nn as nn
 from lightning.pytorch import Trainer
 from pytest import MonkeyPatch
+from torch import Tensor
 from torch.nn.modules import Module
 from torchvision.models._api import WeightsEnum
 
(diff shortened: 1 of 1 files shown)


torchgeo/trainers/segmentation.py
```diff
@@ -319,8 +320,8 @@ def test_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) -> None
         self.log('test_loss', loss, batch_size=batch_size)
 
     def predict_step(
-        self, batch: Any, batch_idx: int, dataloader_idx: int = 0
-    ) -> dict[str, Tensor]:
+        self, batch: Sample, batch_idx: int, dataloader_idx: int = 0
+    ) -> dict[str, Tensor | None]:
```
adamjstewart, line 324: @hkristen FYI. `batch.get('bounds')` returns Tensor for GeoDataset and None for NonGeoDataset. So we always need to check whether the value is None before using it in your callback.

PR #3434, opened 2026-02-16
Collate fn: use Sample type hints

adamjstewart:
This PR restricts our collation functions to only support TorchGeo Samples.

### Alternative

If we want to make these more general, we should really be using Generics everywhere instead of Any.

### DIFF
--- a/torchgeo/datasets/utils.py
+++ b/torchgeo/datasets/utils.py
@@ -530,7 +530,7 @@ def _list_dict_to_dict_list(samples: Iterable[Sample]) -> dict[str, list[Any]]:
     return collated
 
 
-def _dict_list_to_list_dict(sample: Mapping[str, Sequence[Tensor]]) -> list[Sample]:
+def _dict_list_to_list_dict(sample: Mapping[str, Sequence[Any]]) -> list[Sample]:
     """Convert a dictionary of lists to a list of dictionaries.
 
     Args:


torchgeo/datasets/utils.py
```diff
@@ -545,18 +541,16 @@ def _dict_list_to_list_dict(
...
-    ]
+    uncollated = [{} for _ in range(max(map(len, sample.values())))]
     for key, values in sample.items():
         for i, value in enumerate(values):
             uncollated[i][key] = value
     return uncollated
 
 
 def pad_across_batches(
-    batch: list[dict[str, Tensor]], padding_length: int, padding_value: float = 0.0
```
adamjstewart, line 558: Add support for tuples

torchgeo/datasets/utils.py
```diff
@@ -689,10 +683,13 @@ def unbind_samples(sample: MutableMapping[Any, Any]) -> list[dict[Any, Any]]:
 
     .. versionadded:: 0.2
     """
+    uncollated = {}
     for key, values in sample.items():
         if isinstance(values, Tensor):
-            sample[key] = torch.unbind(values)
```
adamjstewart, line 694: Previously using dynamic typing here: `Tensor -> list[Tensor]`

torchgeo/datasets/utils.py
```diff
@@ -689,10 +683,13 @@ def unbind_samples(sample: MutableMapping[Any, Any]) -> list[dict[Any, Any]]:
...
-    return _dict_list_to_list_dict(sample)
+            uncollated[key] = torch.unbind(values)
+        else:
+            uncollated[key] = values
```
adamjstewart, line 691: This is new, we no longer skip non-Tensor values in the output.

PR #3441, opened 2026-02-19
Separated sample logic from vector dataset getitem

BastienKovac:
This PR refactors the vector dataset by decoupling sample creation logic from the __getitem__ method. The goal is to make the sampling process more modular and extensible for other potential vector use cases such as keypoint prediction

## Summary by CodeRabbit

* **Documentation**
  * Resolved documentation configuration issues to properly handle external library type references.

* **Tests**
  * Added test coverage for error handling when invalid task values are used during dataset indexing.

* **Refactor**
  * Improved dataset sample preparation logic for better code organization and maintainability.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -69,6 +69,7 @@ nitpick_ignore = [
     ('py:class', 'torchvision.models._api.WeightsEnum'),
     ('py:class', 'torchvision.models.resnet.ResNet'),
     ('py:class', 'torchvision.models.swin_transformer.SwinTransformer'),
+    ('py:class', 'numpy.int32'),
 ]
 
 
--- a/torchgeo/datasets/geo.py
+++ b/torchgeo/datasets/geo.py
@@ -26,7 +26,8 @@ import rasterio.merge
 import rasterio.warp
 import shapely
 import torch
-from geopandas import GeoDataFrame, GeoSeries
+from geopandas import GeoDataFrame
+from shapely.geometry.base import BaseGeometry
 from pyproj import CRS
 from rasterio.enums import Resampling
 from rasterio.io import DatasetReader
@@ -1030,7 +1031,7 @@ class VectorDataset(GeoDataset):
 
     def _semantic_segmentation_sample(
         self,
-        shapes: list[tuple[GeoSeries, np.int32]],
+        shapes: list[tuple[BaseGeometry, np.int32]],
         width: float,
         height: float,
         transform: Affine,
@@ -1046,7 +1047,7 @@ class VectorDataset(GeoDataset):
 
     def _object_detection_sample(
         self,
-        shapes: list[tuple[GeoSeries, np.int32]],
+        shapes: list[tuple[BaseGeometry, np.int32]],
         width: float,
         height: float,
         transform: Affine,
@@ -1078,7 +1079,7 @@ class VectorDataset(GeoDataset):
 
     def _instance_segmentation_sample(
         self,
-        shapes: list[tuple[GeoSeries, np.int32]],
+        shapes: list[tuple[BaseGeometry, np.int32]],
         width: float,
         height: float,
         transform: Affine,
@@ -1129,7 +1130,7 @@ class VectorDataset(GeoDataset):
 
     def prepare_sample(
         self,
-        shapes: list[tuple[GeoSeries, np.int32]],
+        shapes: list[tuple[BaseGeometry, np.int32]],
         width: float,
         height: float,
         transform: Affine,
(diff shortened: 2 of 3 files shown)


torchgeo/datasets/geo.py
```diff
@@ -30,7 +30,9 @@
 from pyproj import CRS
 from rasterio.enums import Resampling
 from rasterio.io import DatasetReader
+from rasterio.transform import Affine
 from rasterio.vrt import WarpedVRT
+from shapely.geometry.base import BaseGeometry
```
isaaccorley:
```suggestion
from shapely import Geometry
```

torchgeo/datasets/geo.py
```diff
@@ -30,7 +30,9 @@
 from pyproj import CRS
 from rasterio.enums import Resampling
 from rasterio.io import DatasetReader
+from rasterio.transform import Affine
```
isaaccorley:
```suggestion
from rasterio import Affine
```

torchgeo/datasets/geo.py
```diff
@@ -1027,6 +1029,175 @@ def __init__(
         index = pd.IntervalIndex.from_tuples(datetimes, closed='both', name='datetime')
         self.index = GeoDataFrame(data, index=index, geometry=geometries, crs=crs)
 
+    def _semantic_segmentation_sample(
+        self,
+        shapes: list[tuple[BaseGeometry, np.int32]],
```
isaaccorley:
```suggestion
        shapes: list[tuple[Geometry, np.int32]],
```

torchgeo/datasets/geo.py
```diff
@@ -1027,6 +1029,175 @@ def __init__(
...
+                shapes, out_shape=(round(height), round(width)), transform=transform
+            )
+        else:
+            masks = np.zeros((round(height), round(width)), dtype=np.uint8)
+
+        return {'mask': array_to_tensor(masks).to(self.dtype)}
+
+    def _object_detection_sample(
+        self,
+        shapes: list[tuple[BaseGeometry, np.int32]],
```
isaaccorley:
```suggestion
        shapes: list[tuple[Geometry, np.int32]],
```

torchgeo/datasets/geo.py
```diff
@@ -1027,6 +1029,175 @@ def __init__(
...
+            labels = np.empty((0,), dtype=np.int32)
+
+        return {
+            'bbox_xyxy': torch.from_numpy(boxes_xyxy),
+            'label': torch.from_numpy(labels),
+        }
+
+    def _instance_segmentation_sample(
+        self,
+        shapes: list[tuple[BaseGeometry, np.int32]],
```
isaaccorley:
```suggestion
        shapes: list[tuple[Geometry, np.int32]],
```

torchgeo/datasets/geo.py
```diff
@@ -1027,6 +1029,175 @@ def __init__(
...
+
+        return {
+            'mask': array_to_tensor(masks),
+            'bbox_xyxy': torch.from_numpy(boxes_xyxy),
+            'label': torch.from_numpy(labels),
+        }
+
+    def prepare_sample(
+        self,
+        shapes: list[tuple[BaseGeometry, np.int32]],
```
isaaccorley:
```suggestion
        shapes: list[tuple[Geometry, np.int32]],
```

isaaccorley (changes requested):
This is pretty close imo. Just need to fix typing things with my suggestions and then let's add `('py:class', 'numpy.int32'),` to the `nitpick_ignore` section in `torchgeo/docs/conf.py`

torchgeo/datasets/geo.py
BastienKovac: Seems like the Geometry type hint doesn't work here, as it's used as input of shape() method, which expects `dict[str, Any] | SupportsGeoInterface`

calebrob6:
I like this change, agree with @isaaccorley that it's really close. Thanks for the contribution @BastienKovac!

isaaccorley:
@BastienKovac I can fix the docs and type hints if you want me to

BastienKovac:
> @BastienKovac I can fix the docs and type hints if you want me to

If you have time to do it, I would love to, I have to admit I'm pretty inexperienced with strict type hinting in Python (and ty doesn't work on my PC so that doesn't help).

What do you need me to do ?

docs/conf.py
```diff
@@ -69,6 +69,7 @@
     ('py:class', 'torchvision.models._api.WeightsEnum'),
     ('py:class', 'torchvision.models.resnet.ResNet'),
     ('py:class', 'torchvision.models.swin_transformer.SwinTransformer'),
+    ('py:class', 'numpy.int32'),
```
adamjstewart, line 72: Optional: sort alphabetically

adamjstewart (commented):
Not opposed to this, but note that #3333 will also edit these lines, so let's make sure we don't lose that bug fix in the process.

adamjstewart:
Is this PR still a WIP?

PR #3444, opened 2026-02-21
chore: migrate from requirements.txt to uv

isaaccorley:


.github/workflows/tests.yaml
```diff
@@ -58,9 +56,9 @@ jobs:
           enable-cache: true
           activate-environment: true
           cache-dependency-glob: |
-            requirements/min-reqs.old
+            pyproject.toml
       - name: Install dependencies
-        run: uv pip install -r requirements/min-reqs.old
+        run: uv pip install --resolution lowest-direct '.[datasets,models,tests]'
```
adamjstewart, line 61: I love this feature, been dreaming of pip and spack adding it since before uv even existed.

docs/user/contributing.rst
```diff
@@ -132,7 +132,7 @@ All of our documentation is hosted on `Read the Docs <https://about.readthedocs.
 
 .. code-block:: console
 
-   $ pip install .[docs]
+   $ uv pip install .[docs]
```
adamjstewart: Let's undo this change. I think we should add a `uv` section to `installation.rst` but we can do that in a separate PR. I still think most people are using pip, and if they are using uv then they know how to go from pip to uv.

tests/test_requirements.py
adamjstewart, line 1: This test was originally designed to ensure that `min-reqs.old` stayed in sync with `pyproject.toml`. Now that `min-reqs.old` no longer exists, I'm not so sure that this test is actually useful. What is it even testing? I think we may be able to remove this file.

pyproject.toml
```diff
@@ -49,14 +49,14 @@ dependencies = [
     # lightly 1.4.26 is incompatible with the version of timm required by smp
     # https://github.com/torchgeo/torchgeo/issues/1824
     "lightly>=1.4.5,!=1.4.26",
-    # lightning 2+ required for LightningCLI args + sys.argv support
+    # lightning 2.4+ required for lowest-direct compatibility and LightningCLI args + sys.argv support
```
adamjstewart: Explanation?

pyproject.toml
```diff
@@ -92,8 +92,8 @@ dynamic = ["version"]
 
 [project.optional-dependencies]
 datasets = [
-    # h5py 3.10+ required for Python 3.12 wheels
-    "h5py>=3.10",
+    # h5py pinned for pre-3.14 due to wheel support constraints
+    "h5py==3.14.0; python_version<'3.14'",
```
adamjstewart: Explanation?

pyproject.toml
```diff
@@ -120,13 +120,17 @@ docs = [
     "ipywidgets>=7",
     # myst-parser 0.18.0+ required for sphinx 5.0+ compatibility
     "myst-parser>=0.18.0",
+    # nbmake 1.3.3+ required for notebook metadata mocking in docs/tutorials
+    "nbmake>=1.3.3",
```
adamjstewart: Why move this out of tests? It isn't needed to build the docs, only to test the docs.

pyproject.toml
```diff
@@ -120,13 +120,17 @@ docs = [
...
     # nbsphinx 0.8.5 fixes bug with nbformat attributes
     "nbsphinx>=0.8.5",
     # pydata-sphinx-theme 0.14.0+ required for sphinx 5.0+ compatibility
     "pydata-sphinx-theme>=0.14.0",
     # sphinx 5.3+ required to allow section titles inside autodoc class docstrings
     # https://github.com/sphinx-doc/sphinx/pull/10887
     "sphinx>=5.3",
+    # setuptools<81 required for pkg_resources used by nbmake plugin
+    "setuptools>=77.0.1,<81",
```
adamjstewart: Would rather avoid this. We can bump the min version of nbmake if necessary, 1.4.5+ removed the pkg_resources dependency.

pyproject.toml
```diff
@@ -259,7 +261,6 @@ filterwarnings = [
     # Expected warnings
     # Lightning warns us about using num_workers=0, but it's faster on macOS
     "ignore:The .*dataloader.* does not have many workers which may be a bottleneck:UserWarning:lightning",
-    "ignore:The .*dataloader.* does not have many workers which may be a bottleneck:lightning.fabric.utilities.warnings.PossibleUserWarning:lightning",
```
adamjstewart: Seems unrelated? You reminded me that I've been meaning to minimize these ignored warnings, most are ancient. But let's do that in a separate PR.

uv.lock
adamjstewart, line 1: So if I understand correctly, this file lists the latest version of every dep, so we avoid CI failures when transitive deps get new releases. But it also only gets updated by dependabot when a direct dep gets a new release?

adamjstewart (changes requested):
Also need to update `.readthedocs.yaml`

pyproject.toml
isaaccorley: I was having build errors with lightning-cloud for the min tests but I didn't try debugging it too hard before I just bumped the version

pyproject.toml
isaaccorley: Yes but tests implies it's needed for pytest tests right? which it's not used for. In CI it's not used except for the docs step

pyproject.toml
isaaccorley: This is actually defined in the requirements/datasets.txt. I just moved it into pyproject.toml

uv.lock
isaaccorley: Yup so when you run uv sync --lock it will basically install the exact versions defined in the lock file. This means it's less likely to fail due to resolution errors and it also speeds install up because it doesn't have to figure out resolving dependencies.

.github/workflows/tests.yaml
isaaccorley: Agree, another reason to love uv

pyproject.toml
adamjstewart: Ah, I think we should at least have a range. Basically this is because h5py 3.15+ isn't compatible with the netcdf4 wheels. For Python 3.14+, we aren't testing h5py at all because of lack of wheels.

pyproject.toml
adamjstewart: It's used for `pytest --nbmake`, it isn't used for `sphinx-build`. So it's more of a testing thing than a documentation thing.

pyproject.toml
isaaccorley: Ahhh I accidentally copy/pasted from the txt file without using the old range. I'll fix

pyproject.toml
isaaccorley: Gotcha okay I'll revert

isaaccorley:
For the minimum tests I'm getting this error everywhere `AttributeError: module 'fiona' has no attribute 'path'. Did you mean: 'Path'?` it's probably best to just bump geopandas to 1.0 so we don't have to deal with fiona.

.github/workflows/release.yaml
```diff
@@ -24,11 +24,10 @@ jobs:
...
           activate-environment: true
           cache-dependency-glob: |
-            requirements/required.txt
-            requirements/datasets.txt
-            requirements/tests.txt
+            pyproject.toml
+            uv.lock
       - name: Install dependencies
-        run: uv pip install .[datasets,tests]
+        run: uv sync --locked --extra datasets --extra tests
```
adamjstewart: For the release tests, we specifically weren't using pinned deps to make sure the latest versions work fine.

.readthedocs.yaml
```diff
@@ -13,8 +13,10 @@ build:
 # Configuration of the Python environment to be used
 python:
   install:
-    - requirements: requirements/required.txt
-    - requirements: requirements/docs.txt
+    - method: pip
+      path: .
+      extra_requirements:
+        - docs
```
adamjstewart: Can we still use pinned versions here? https://docs.readthedocs.com/platform/stable/build-customization.html#install-dependencies-with-uv is a bit ugly, but it's better than failing tests.

pyproject.toml
```diff
@@ -92,8 +92,8 @@ dynamic = ["version"]
 
 [project.optional-dependencies]
 datasets = [
-    # h5py 3.10+ required for Python 3.12 wheels
```
adamjstewart, line 95: Let's keep this lower bound comment too

pyproject.toml
adamjstewart: Would still like to know more here.

isaaccorley: I've reverted and we can see what happens in CI

isaaccorley: this is the error in CI for the minimum tests:

```bash
_______________ ERROR collecting tests/datamodules/test_caffe.py _______________
ImportError while importing test module '/home/runner/work/torchgeo/torchgeo/tests/datamodules/test_caffe.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/datamodules/test_caffe.py:9: in <module>
    from torchgeo.datamodules import CaFFeDataModule
torchgeo/datamodules/__init__.py:6: in <module>
    from .agrifieldnet import AgriFieldNetDataModule
torchgeo/datamodules/agrifieldnet.py:15: in <module>
    from .geo import GeoDataModule
torchgeo/datamodules/geo.py:11: in <module>
    from lightning.pytorch import LightningDataModule
.venv/lib/python3.12/site-packages/lightning/__init__.py:18: in <module>
    from lightning.app import storage  # noqa: E402
.venv/lib/python3.12/site-packages/lightning/app/__init__.py:24: in <module>
    from lightning.app import components  # noqa: E402, F401
.venv/lib/python3.12/site-packages/lightning/app/components/__init__.py:1: in <module>
    from lightning.app.components.database.client import DatabaseClient
.venv/lib/python3.12/site-packages/lightning/app/components/database/__init__.py:1: in <module>
    from lightning.app.components.database.client import DatabaseClient
.venv/lib/python3.12/site-packages/lightning/app/components/database/client.py:22: in <module>
    from lightning.app.components.database.utilities import _GeneralModel
.venv/lib/python3.12/site-packages/lightning/app/components/database/utilities.py:20: in <module>
    from fastapi import Response, status
.venv/lib/python3.12/site-packages/fastapi/__init__.py:5: in <module>
    from .applications import FastAPI
.venv/lib/python3.12/site-packages/fastapi/applications.py:3: in <module>
    from fastapi import routing
.venv/lib/python3.12/site-packages/fastapi/routing.py:6: in <module>
    from fastapi import params
.venv/lib/python3.12/site-packages/fastapi/params.py:4: in <module>
    from pydantic import Schema
E   ImportError: cannot import name 'Schema' from 'pydantic' (/home/runner/work/torchgeo/torchgeo/.venv/lib/python3.12/site-packages/pydantic/__init__.py)"
```

isaaccorley: The the minimum tests install `lightning==2.0.0` which installs `fastapi==0.1.17` with pydantic v2 but `fastapi` imports `pydantic.Schema` which was removed in pydantic v2.

We have to bump to lightning 2.4 though because previous lightning versions were failing because of setuptools (pkg_resources) errors

.github/workflows/release.yaml
```diff
@@ -51,11 +52,12 @@ jobs:
...
-            requirements/required.txt
-            requirements/docs.txt
-            requirements/tests.txt
+            pyproject.toml
+            uv.lock
       - name: Install dependencies
-        run: uv pip install .[docs,tests] planetary_computer pystac tensorboard h5py scikit-learn
+        run: |
+          uv sync --locked --extra docs --extra tests
+          uv pip install planetary_computer pystac tensorboard h5py scikit-learn
```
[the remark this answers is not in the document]
adamjstewart: This works because we aren't using netcdf4 in combination with h5py. h5py 3.15+ has Python 3.14 wheels, but isn't compatible with netcdf4.

.github/workflows/release.yaml
```diff
@@ -51,11 +52,12 @@ jobs:
...
           cache-dependency-glob: |
-            requirements/required.txt
-            requirements/docs.txt
-            requirements/tests.txt
+            pyproject.toml
+            uv.lock
       - name: Install dependencies
-        run: uv pip install .[docs,tests] planetary_computer pystac tensorboard h5py scikit-learn
+        run: |
+          uv sync --locked --extra docs --extra tests
```
adamjstewart, line 60: This one should also have `uv lock --upgrade`

pyproject.toml
adamjstewart: Alright, good enough for me, 2.4 has been out for a while anyway.

adamjstewart (approved):
I think we should merge this as it's a huge improvement to prevent new releases of transitive deps from breaking CI.

Only remaining concern I will leave is the lack of a separate file to denote conflicts for minimum requirements. This means we need to put all conflicts in pyproject.toml, which may be overly strict. For example, torchmetrics still hasn't removed pkg_resources usage in main even though it's been deprecated for 5+ years: https://github.com/Lightning-AI/torchmetrics/pull/3324. This time we aren't affected because wheels exist, but next time we may not be so lucky. I think a year or two from now, we will come back to this and will need to find a way to document testing constraints that are different than run-time constraints, but we can also wait until more features are added to uv.

PR #3446, opened 2026-02-21
docs: NAIP road segmentation tutorial

isaaccorley:
Adds a tutorial for the ChesapeakeRSC road segmentation model with sample NAIP imagery

Here's the output plot

image


robmarkcole:
Will this work for historical models, or just those trained in newer versions of pytorch? 
```
preprocess = weights.transforms
inputs = preprocess(image)
```

isaaccorley:
All the torchvision models support this syntax now. There's no more resnet50(pretrained=True).

calebrob6:
I told claude to make the code style consistent with the other tutorials.

Only comment I have is about the vsicurl prefix, pretty sure that's not needed with rasterio and you can just pass https://....

Confirmed that the notebook works for me and I think this is a great tutorial to have.

calebrob6 (approved):

PR #3450, opened 2026-02-23
API docs: don't resolve type aliases

adamjstewart:
Previously, an annotation like:
```python
def __getitem__(index: GeoSlice) -> Sample:
```
would be rendered in the docs as:
```python
def __getitem__(index: slice | tuple[slice] | tuple[slice, slice] | tuple[slice, slice, slice]) -> dict[str, Any]:
```
This will only get worse if we start documenting the generic slice types (#3294) or if Sample becomes a TypedDict. This PR is designed to instead use the type alias, then refer the reader to the docs for that alias.

### DIFF
--- a/torchgeo/datasets/mmflood.py
+++ b/torchgeo/datasets/mmflood.py
@@ -3,8 +3,6 @@
 
 """MMFlood dataset."""
 
-from __future__ import annotations
-
 import os
 from collections.abc import Callable
 from glob import glob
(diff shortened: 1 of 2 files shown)


adamjstewart:
1. Types added to the docs, but not docstrings without using `.. autotype::`
2. Type aliases not used in function signatures, still using expanded form
3. Seems to require `from __future__ import annotations`, whereas other features don't work with this

adamjstewart:

PR #3451, opened 2026-02-23
Datasets: percentile_normalization -> quantile_normalization

adamjstewart:
### Motivation

While working on converting to `Sample = dict[str, Tensor]`, I noticed that `percentile_normalization` only claims to work for numpy arrays, but we often feed it torch tensors. This went unnoticed because 1) tensors happen to work, and 2) we were using Any for the type hint.

### Implementation

Since all images are guaranteed to be tensors at the start, I think it makes more sense to keep them as tensors during normalization. I rewrote `percentile_normalization` (numpy) as `quantile_normalization` (torch) to have fewer type conversions. There are a few key differences:

* numpy arrays now explicitly fail
* upper/lower are in the range 0-1, not 0-100
* axis is replaced by dim, does not support multiple dims

### Alternatives

The least amount of code changes required to make ty happy would be to simply support both numpy array and torch tensor inputs. Another alternative would be to strictly require numpy arrays and cast tensors to numpy when needed. Personally I think it makes most sense to stay within the torch ecosystem here.

### DIFF
--- a/tests/datasets/test_utils.py
+++ b/tests/datasets/test_utils.py
@@ -28,6 +28,7 @@ from torchgeo.datasets.utils import (
     lazy_import,
     merge_samples,
     pad_across_batches,
+    percentile_normalization,
     quantile_normalization,
     stack_samples,
     unbind_samples,
@@ -540,6 +541,15 @@ def test_nonexisting_directory(tmp_path: Path) -> None:
 
     with working_dir(str(subdir), create=True):
         assert subdir.cwd() == subdir
+
+
+def test_percentile_normalization() -> None:
+    img = np.array([[1, 2], [98, 100]])
+    match = 'Use torchgeo.datasets.utils.quantile_normalization instead'
+    with pytest.warns(DeprecationWarning, match=match):
+        img = percentile_normalization(img, 2, 98)
+    assert img.min() == 0
+    assert img.max() == 1
 
 
 def test_quantile_normalization() -> None:
--- a/torchgeo/datasets/utils.py
+++ b/torchgeo/datasets/utils.py
@@ -773,6 +773,41 @@ def rgb_to_mask(
     return mask
 
 
+@deprecated('Use torchgeo.datasets.utils.quantile_normalization instead')
+def percentile_normalization(
+    img: np.typing.NDArray[np.int_],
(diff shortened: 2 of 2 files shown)


isaaccorley:
I agree with staying with the torch torch ecosystem over numpy, I just wish it was easier to enforce checks that define dim order for input, hwc vs chw

adamjstewart:
> I just wish it was easier to enforce checks that define dim order for input, hwc vs chw

I agree. @rgommers have you seen any adoption for [PEP 646](https://peps.python.org/pep-0646/) in numpy or torch? The closest things I've seen are https://github.com/patrick-kidger/torchtyping and https://github.com/patrick-kidger/jaxtyping.

rgommers:
The canonical issue is https://github.com/numpy/numpy/issues/16544. @jorenham may have something more informative to say about the specific question above

jorenham:
Thanks for the ping :)

For NumPy, the sad truth is that PEP 646 isn't usable. That's because `TypeVarTuple` cannot be declared covariant, and does not support upper type bounds. Covariance is needed so that it will accept subtypes of integer-shapes such as `tuple[Literal[1]]`, and the bound is needed to prevent non-integer axis from being passed, as that would be type-unsafe. So we would need `ShapeT = TypeVarTuple("ShapeT", covariant=True, bound=int)`, but that's just not possible.

But luckily we don't need PEP 646 in order to statically enforce a specific rank of arrays, as we can just rely on tuple types instead. For example, describe a 1-d f64 array, you could write `np.ndarray[tuple[int], np.dtype[np.float64]]`.
However, this should not be used on `numpy<2.2`. I'll spare you the details as to why that is exactly, but let's just say that statically enforcing shape-types should only be done if the minimum supported numpy version is at least 2.2.

isaaccorley:
@adamjstewart could we deprecate `percentile_normalization` for a release first, some of my libraries use it as a utility so I imagine others might.

adamjstewart:
Done. Note that I don't consider functions to be public unless they appear in the documentation, regardless of underscore prefix. So I wouldn't recommend on relying on this function. We can make it public if you think it would be useful though.

isaaccorley (approved):

PR #3458, opened 2026-02-26
Copernicus-Pretrain: remove JSON metadata

adamjstewart:
We've been trying to slowly remove non-Tensor values from our sample dictionaries to ensure that our datasets are compatible with both PyTorch's default collate_fn and Lightning's default transfer_to_device function. This PR removes the JSON dict returned by Copernicus-Pretrain and only returns the images needed for training.

@wangyi111 is this change okay with you?


wangyi111:
Is it possible to add a note somewhere mentioning "there is metadata available and if you want to get them do this and that"? If people only use the images it's fine, if they wanna use metadata then a guide somewhere would be good.

isaaccorley:
Can probably add a code example directly in the docstring for this. Wdyt @adamjstewart?

adamjstewart:
Not sure about code since it isn't easy to access anymore, but we can at least mention that metadata exists.

PR #3472, opened 2026-02-27
pre-commit: fix ty hook breaking local commits

hkristen:
When committing locally with pre-commit, the ty hook fails because --config-file only accepts ty.toml -> passing the pyproject.toml causes a parse error. 

The CI is unaffected since it runs ty check directly without the flag. Removing the arg fixes local commits and ty auto-discovers pyproject.toml on its own.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,7 +19,6 @@ repos:
         entry: ty check
         language: python
         pass_filenames: false
-        args: [--config-file, pyproject.toml]
         additional_dependencies:
           - einops>=0.6.0
           - kornia>=0.7.4


adamjstewart:
@isaaccorley is there any other way to use shared config or do we need to manually copy all settings? I don't want to use ty.toml.

isaaccorley:
> @isaaccorley is there any other way to use shared config or do we need to manually copy all settings? I don't want to use ty.toml. 

According to the ty [docs](https://docs.astral.sh/ty/configuration/#configuration-files) it will auto search for pyproject.toml so I think merging this as-is is the fix

adamjstewart (approved):

adamjstewart:
FYI, `--config pyproject.toml` may also work, but if we haven't seen any problems, then we're fine.

PR #3476, opened 2026-02-28
Habitalp2 dataset

hkristen:
Adds the [Habitalp2](https://huggingface.co/datasets/JR-DIGITAL/habitalp2.0) ecological habitat mapping dataset for Gesäuse National Park, Austria.

Consits of two datasets & the corresponding HabitAlp2DataModule
- HabitAlp2: For Semantic segmentation over three epochs (2003, 2013, 2020) with 23 habitat classes, RGB/CIR imagery and LiDAR-derived terrain layers
- HabitAlp2CD: For bi-temporal change detection with task="binary" (0/1) or  task="multiclass" (0–8 change classes) for (2003->2013 and 2013->2020)

## Summary by CodeRabbit

* **New Features**
  * Added HabitAlp2 dataset support with semantic segmentation capabilities across multiple temporal epochs (2003, 2013, 2020)
  * Added HabitAlp2 change detection dataset for bi-temporal pair analysis
  * Introduced HabitAlp2DataModule for streamlined data loading and preprocessing with configurable batch sizes, patch sizes, and task-specific augmentations


calebrob6:
Could you post some example plots from the dataset? Also, have you tested the datamodule? How is the throughput?

hkristen:
@calebrob6 ofc I have tested the HabitAlp2DataModule for both `task='segmentation'` and` task='change_detection'`.

### Sample Plots
(I know those labels are weird and fuzzy sometimes, that makes working with this data extra fun)

habitalp2_seg_samples

habitalp2_cd_samples

### Throughput Benchmark

My System: Intel i9-14900K, 96 GB RAM, RTX 4090 wiht a very fast SSD
Measured as: data loading + CPU→GPU transfer (pin_memory=True), 3 warmup batches discarded, 30 measured batches of 8 patches @ 256×256px, RandomBatchGeoSampler

habitalp2_throughput

torchgeo/datamodules/habitalp2.py
```diff
@@ -0,0 +1,109 @@
...
+from ..samplers.utils import _to_tuple
+from .geo import GeoDataModule
+
+
+class HabitAlp2DataModule(GeoDataModule):
+    """LightningDataModule implementation for the HabitAlp2 dataset.
+
+    Supports both semantic segmentation and change detection tasks.
+
+    .. versionadded:: 0.9.1
```
calebrob6: should be 0.10 I think

torchgeo/datamodules/habitalp2.py
```diff
@@ -0,0 +1,109 @@
...
+        if task == 'segmentation':
+            self.train_aug = K.AugmentationSequential(
+                K.Normalize(mean=self.mean, std=self.std),
+                K.RandomResizedCrop(_to_tuple(self.patch_size), scale=(0.6, 1.0)),
+                K.RandomVerticalFlip(p=0.5),
+                K.RandomHorizontalFlip(p=0.5),
+                data_keys=None,
+                keepdim=True,
+                extra_args={
+                    DataKey.MASK: {'resample': Resample.NEAREST, 'align_corners': None}
```
calebrob6, line 77: Flagging that I'm not familiar with this pattern, is it used elsewhere?

torchgeo/datasets/habitalp2.py
```diff
@@ -0,0 +1,765 @@
...
+    * masks are single-band GeoTIFFs with class IDs 1-23
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://doi.org/10.48550/arXiv.2511.00073
+
+    .. versionadded:: 0.9.1
+    """
+
+    url = 'https://huggingface.co/datasets/JR-DIGITAL/habitalp2.0/resolve/main/'
```
calebrob6: We usually pin to a commit hash I think

torchgeo/datamodules/habitalp2.py
hkristen: I saw the pattern being used here https://github.com/torchgeo/torchgeo/blob/deeced59999e96baf1ac06e5aba2e856ff6c55a0/torchgeo/datamodules/l7irish.py#L59

Kinda needed as RandomResizedCrop defaults to bilinear interpolation, which would blend class label integers into invalid values.

torchgeo/datasets/habitalp2.py
```diff
@@ -0,0 +1,765 @@
...
+                    download_url(
+                        self.url + year_files[band], self.root, year_files[band]
+                    )
+
+        mask_file = self.mask_files[self.year]
+        mask_path = os.path.join(self.root, mask_file)
+        if not os.path.exists(mask_path):
+            download_url(self.url + mask_file, self.root, mask_file)
+
+    def plot(
```
nilsleh, line 413: perhaps also show labels in legend of segmentation classes as part of the plot?

torchgeo/datamodules/habitalp2.py
```diff
@@ -0,0 +1,109 @@
...
+
+        Args:
+            batch_size: Size of each mini-batch.
+            patch_size: Size of each patch, either ``size`` or ``(height, width)``.
+            length: Length of each training epoch.
+            num_workers: Number of workers for parallel data loading.
+            task: One of 'segmentation' or 'change_detection'.
+            **kwargs: Additional keyword arguments passed to
+                :class:`~torchgeo.datasets.HabitAlp2` or
+                :class:`~torchgeo.datasets.HabitAlp2CD`.
```
nilsleh, line 47: could add a "Raises" doc string here for the AssertionError

tests/datamodules/test_habitalp2.py
adamjstewart, line 1: Would like to remove this and only test via the actual trainers if possible, see https://github.com/torchgeo/torchgeo/blob/main/tests/datamodules/README.md

tests/datasets/test_habitalp2.py
```diff
@@ -0,0 +1,217 @@
...
+    UnionDataset,
+)
+
+DATA_DIR = os.path.join('tests', 'data', 'habitalp')
+
+
+def _copy_file(url: str, root: Path, filename: str, **kwargs: Any) -> None:
+    dst = os.path.join(str(root), filename)
+    os.makedirs(os.path.dirname(dst), exist_ok=True)
+    shutil.copy(url, dst)
```
adamjstewart, line 32: This should already be mocked automatically by conftest.py?

tests/datasets/test_habitalp2.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+class TestHabitAlp2:
+    @pytest.fixture(params=['2003', '2013', '2020'])
```
adamjstewart: Do you expect these 3 years to hit different code paths, or is this overkill for getting test coverage?

tests/datasets/test_habitalp2.py
```diff
@@ -0,0 +1,217 @@
...
+    def test_getitem(self, dataset: HabitAlp2) -> None:
+        x = dataset[dataset.bounds]
+        assert isinstance(x, dict)
+        assert isinstance(x['image'], torch.Tensor)
+        assert isinstance(x['mask'], torch.Tensor)
+        assert x['image'].dtype == torch.float32
+        assert x['mask'].dtype == torch.int64
+
+    def test_len(self, dataset: HabitAlp2) -> None:
+        assert len(dataset) >= 1
```
adamjstewart: Can we assert a specific length?

torchgeo/datamodules/habitalp2.py
```diff
@@ -0,0 +1,109 @@
...
+    .. versionadded:: 0.10
+    """
+
+    def __init__(
+        self,
+        batch_size: int = 8,
+        patch_size: int | tuple[int, int] = 256,
+        length: int | None = None,
+        num_workers: int = 0,
+        task: str = 'segmentation',
```
adamjstewart:
```suggestion
        task: Literal['segmentation', 'change_detection'] = 'segmentation',
```
Not all `str` are valid, only an infinitesimally small subset.

torchgeo/datamodules/habitalp2.py
```diff
@@ -0,0 +1,109 @@
...
+    def setup(self, stage: str) -> None:
+        """Set up datasets.
+
+        Args:
+            stage: Either 'fit', 'validate', 'test', or 'predict'.
+        """
+        dataset = cast(GeoDataset, self.dataset_class(**self.kwargs))
+        self.train_dataset = dataset
+        self.val_dataset = dataset
+        self.test_dataset = dataset
```
adamjstewart: We are currently testing on train, which is a big no-no. We should either split the dataset or split the sampler based on geographic bounds. Is there any officially published splitting strategy?

torchgeo/datamodules/habitalp2.py
```diff
@@ -0,0 +1,109 @@
...
+import kornia.augmentation as K
+from kornia.constants import DataKey, Resample
+
+from ..datasets import GeoDataset, HabitAlp2, HabitAlp2CD
+from ..samplers import GridGeoSampler, RandomBatchGeoSampler
+from ..samplers.utils import _to_tuple
+from .geo import GeoDataModule
+
+
+class HabitAlp2DataModule(GeoDataModule):
```
adamjstewart, line 20: Wonder if we should have multiple datamodules to match the multiple datasets?

torchgeo/datasets/habitalp2.py
```diff
@@ -0,0 +1,765 @@
...
+    is_image = True
+
+
+class _HabitAlp2Mask(RasterDataset):
+    """Mask component for HabitAlp2 dataset."""
+
+    is_image = False
+
+
+class HabitAlp2(GeoDataset):
```
adamjstewart, line 55: I prefer to use IntersectionDataset as a base class instead of GeoDataset, see L7Irish, L8Biome, I/O Bench, and MMFlood for examples of this style.

torchgeo/datasets/habitalp2.py
```diff
@@ -0,0 +1,765 @@
...
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://doi.org/10.48550/arXiv.2511.00073
+
+    .. versionadded:: 0.10
+    """
+
+    url = 'https://huggingface.co/datasets/JR-DIGITAL/habitalp2.0/resolve/df01fe8ae22df182e7bf1c2e3e713dcfd92c0c81/'
+
+    valid_years: ClassVar[tuple[str, ...]] = ('2003', '2013', '2020')
```
adamjstewart:
```suggestion
    valid_years = ('2003', '2013', '2020')
```
tuples are immutable, no need to mark with ClassVar, only dicts need this treatment. Same below.

torchgeo/datasets/habitalp2.py
```diff
@@ -0,0 +1,765 @@
...
+        'Old broad-leaved forest CC<80',
+        'Old broad-leaved forest CC>=80',
+        'Clearcut areas',
+        'Mountain dwarf forest (Krummholz)',
+        'Grassland, buffer strip',
+        'Alpine grassland, heath',
+        'Low importance/small extent',
+    )
+
+    def __init__(
```
adamjstewart, line 197: May want to consider adding time_series parameter, we recently added this to all RasterDataset classes.

torchgeo/datasets/habitalp2.py
```diff
@@ -0,0 +1,765 @@
...
+            )
+
+        self.bands = bands
+
+        self._verify()
+
+        year_files = self.data_files[year]
+        mask_path = os.path.join(root, self.mask_files[year])
+
+        needs_rgb = any(b in ['R', 'G', 'B'] for b in self.bands)
```
adamjstewart: Optional: another way I like to write this is using set notation:
```suggestion
        needs_rgb = set(self.bands) & {'R', 'G', 'B'}
```
I'll let you decide which is more readable.

torchgeo/datasets/habitalp2.py
```diff
@@ -0,0 +1,765 @@
...
+            if show_titles:
+                axs[2].set_title('Prediction')
+
+        if suptitle is not None:
+            plt.suptitle(suptitle)
+
+        return fig
+
+
+class HabitAlp2CD(GeoDataset):
```
adamjstewart, line 481: A shared base class could probably cut down on code duplication. If merged as is, this will become one of the longest dataset.py files, but the dataset itself doesn't actually seem that complicated compared to others.

adamjstewart (changes requested):
Needs docs

torchgeo/datasets/habitalp2.py
hkristen: This would be quite a lot of effort and I don't see the big benefit here. HabitAlp2 is structurally different from L7Irish/L8Biome where a single glob covers images and another covers masks. Would rather leave it as is to avoid the effort of refactoring a lot of code here.

torchgeo/datasets/habitalp2.py
hkristen: This would probably save 40-50 lines total and add a third class to understand, not sure if this is worth it.

hkristen: To the comment on the length of the dataset: (fair point!)
The file is long because it has two complete dataset classes (segmentation + change detection), each with their own init, plot, download, and verify logic plus quite some definitions on top. Don't know how I could make this much shorter tbh.

torchgeo/datamodules/habitalp2.py
hkristen: Done

torchgeo/datasets/habitalp2.py
hkristen: As there are 24 classes the legend would be rather cluttered ... so I would rather not do that if you are ok.

tests/datamodules/test_habitalp2.py
hkristen: Removed it -> testing through `habitalp2_seg.yaml` and `habitalp2_cd.yaml` in the trainer tests now.

tests/datasets/test_habitalp2.py
hkristen: The conftest mock prevents real HTTP downloads, but `_copy_file `replaces `download_url `with a local file copy so the test data actually gets moved into `tmp_path`. Without it the download test has nothing to load from.

tests/datasets/test_habitalp2.py
hkristen: 2003 hits the RGB-only path (no CIR/terrain), 2013 hits the full path. 2020 is identical to 2013 code-path-wise, so dropped it. Reduced to `['2003', '2013']`.

tests/datasets/test_habitalp2.py
hkristen: Done, changed to `== 1`

torchgeo/datamodules/habitalp2.py
hkristen: There's no official split published for this dataset. Which is the same pattern as some other datamodules (e.g. SouthAfricaCropType). We could add a geographic split later?

torchgeo/datamodules/habitalp2.py
hkristen: A single datamodule with task parameter keeps things simpler and mirrors how the dataset classes share the same data on disk. I would rather keep it like this.

torchgeo/datasets/habitalp2.py
hkristen: Done, removed ClassVar from all tuple attributes. Kept it on dicts only.

torchgeo/datasets/habitalp2.py
hkristen: Would be a nice addition but feels out of scope for this PR since it would be a new feature across the codebase. Happy to add in a follow-up.

torchgeo/datasets/habitalp2.py
hkristen: Went with a list comprehension` (rgb_requested = tuple(b for b in self.bands if b in ('R', 'G', 'B'))) `instead because I needed the filtered list for fixing the band subsetting fix (passing exactly the requested RGB subset to `_HabitAlp2RGB`).

torchgeo/datamodules/habitalp2.py
```diff
@@ -0,0 +1,112 @@
...
+    """LightningDataModule implementation for the HabitAlp2 dataset.
+
+    Supports both semantic segmentation and change detection tasks.
+
+    .. versionadded:: 0.10
```
nilsleh: should be updated to 0.11

torchgeo/datamodules/habitalp2.py
nilsleh: I think if we don't introduce a geographic split here already, it at least needs to be clearly documented, otherwise people might mistakenly overlook this

torchgeo/datasets/habitalp2.py
```diff
@@ -0,0 +1,790 @@
...
+    Dataset format:
+
+    * images are multi-band GeoTIFFs
+    * masks are single-band GeoTIFFs with class IDs 1-23
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://doi.org/10.48550/arXiv.2511.00073
+
+    .. versionadded:: 0.10
```
nilsleh: needs to be updated

torchgeo/datasets/habitalp2.py
```diff
@@ -0,0 +1,790 @@
...
+
+        Returns:
+            sample containing image and mask at that index
+
+        Raises:
+            IndexError: if index is not found in the index
+        """
+        sample = self.dataset[index]
+        sample['image'] = sample['image'].float()
+        sample['mask'] = sample['mask'].long().squeeze(0)
```
nilsleh, line 346: if the labels are 1 indexed, they probably need to be shifted to be 0 indexed to work with the common loss functions, or 0 can be introduced as an explicit background class if there is one?

torchgeo/datasets/habitalp2.py
nilsleh: I am not familiar with the dataset, but an option could be to just show the ones that are present in that plot (under the assumption that not always 24 classes show up in each mask) but also I am fine with how you prefer.

torchgeo/datasets/habitalp2.py
```diff
@@ -0,0 +1,790 @@
...
+    Dataset format:
+
+    * images are multi-band GeoTIFFs
+    * masks are single-band GeoTIFFs with class IDs 0-8 (binary task binarizes to 0/1)
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://doi.org/10.48550/arXiv.2511.00073
+
+    .. versionadded:: 0.10
```
nilsleh: needs to be updated

torchgeo/datasets/habitalp2.py
hkristen: A background class = 0 actually already exists in the dataset, 1-23 are the habitat classes. I made that clearer in the docstring.

hkristen:
> I am not familiar with the dataset, but an option could be to just show the ones that are present in that plot (under the assumption that not always 24 classes show up in each mask) but also I am fine with how you prefer.

That is actually a good idea @nilsleh -> applied it to both datasets.

torchgeo/datamodules/habitalp2.py
hkristen: I heard you guys-> added a geographic split similar to how it is done in the Sentinel2+CDL and Sentinel2+EuroCrops datamodules. 

The `setup()` now clips the dataset index to the study area outline for the target year, then uses `random_grid_cell_assignment `to split into train/val/test (70/15/15, grid_size=10, fixed seed similar to what we did in our preprint). The outlines GPKG is hosted alongside the data on HF and downloaded automatically if missing.

PR #3498, opened 2026-03-08
Adding DLRSD dataset

calebrob6:
Adding the DLRSD dataset from https://sites.google.com/view/zhouwx/dataset

This was used recently in the https://research.google/blog/accelerating-scientific-discovery-with-ai-powered-empirical-software/ and I thought it'd be cool if it was in torchgeo.

There are two datasets here, DLRSD is a version of UCM with semantic segmentation labels. DLRSD_multilabel is a version of UCM with multi-labels.

- [x] Will upload to hf here https://huggingface.co/datasets/torchgeo/dlrsd/tree/main when I get better internet.

E.g.:
```
from torchgeo.datasets.dlrsd import DLRSD, DLRSDMultilabel

ds = DLRSD(root="data/dlrsd")
ds.plot(ds[100])
plt.show()
plt.close()

ds = DLRSDMultilabel(root="data/dlrsd")
ds.plot(ds[100])
plt.show()
plt.close()
```

image

image

 

## Release Notes

* **New Features**
  * Added DLRSD semantic segmentation dataset supporting 2100 RGB images with 17 segmentation labels.
  * Added DLRSDMultilabel multi-label classification dataset with 17 scene classes.
  * Added data modules for both dataset variants with configurable train/validation/test splits and augmentation pipelines.
  * Added documentation and comprehensive test coverage for new datasets.


calebrob6:
Other things:
- There isn't a defined train/val/test split. The google paper I linked above uses random 80/20 train/test. Original paper "Performance Evaluation of Single-Label and Multi-Label Remote Sensing Image Retrieval Using a Dense Labeling Dataset" says random 80%/20% train/test by category.
- In the hf upload I've included the multilabel file (as csv instead of excel) as the images are identical between the original two dataset downloads
- ~I will add a fixed train/val/test split of 70/10/20~
  - I will add a datamodule that lets you do the 70/10/20 split with a seed parameter.

isaaccorley:
@coderabbitai review

adamjstewart:
> There are two datasets here, DLRSD is a version of UCM with semantic segmentation labels. DLRSD_multilabel is a version of UCM with multi-labels.

This is cool, I'm glad people are building on existing datasets. Similar to the instance segmentation polygons added to VHR-10 after it was released.

docs/api/datasets/benchmark.csv
```diff
@@ -14,6 +14,8 @@ Dataset,Task,Source,# Samples,# Classes,Size (px),Resolution (m),Bands,License
 :ref:`DFC2022`,S,Aerial,"3,981",15,"2,000x2,000",0.5,RGB,CC-BY-4.0
 :ref:`DIOR`,OD,Aerial,"23,463",20,"800x800",0.5,RGB,CC-BY-NC-4.0
 :ref:`DL4GAM`,S,"Sentinel-2","2,251 or 11,440","2","256x256","10","MSI",CC-BY-4.0
+:ref:`DLRSD`,S,USGS National Map,"2,100",17,256x256,0.3,RGB,public domain
+:ref:`DLRSD Multilabel`,C,USGS National Map,"2,100",17,256x256,0.3,RGB,public domain
```
adamjstewart: Were you able to find the license information, or is this just copied from UC Merced? I confirmed the license of UC Merced with the author, but it might be worth double checking with the authors of DLRSD too.

docs/api/datasets/dlrsd-multilabel.rst
adamjstewart, line 1: Wondering if we should combine both classes in a single documentation page since they share the same `dlrsd.py` file. You can still have multiple rows in `benchmark.csv` if you want, just use:
```rst
:ref:`DLRSD` Multilabel
```
for one of them.

tests/data/dlrsd/data.py
```diff
@@ -0,0 +1,95 @@
...
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+
+import csv
+import os
+import shutil
+
+import numpy as np
+from PIL import Image
+from torchvision.datasets.utils import calculate_md5
```
adamjstewart: We no longer test the hashes of fake data, this could probably be removed.

tests/data/dlrsd/data.py
```diff
@@ -0,0 +1,95 @@
...
+    ]
+    csv_path = os.path.join(folder_path, 'multilabels.csv')
+    with open(csv_path, 'w', newline='') as f:
+        writer = csv.writer(f)
+        writer.writerow(header)
+        writer.writerows(csv_rows)
+
+    # Create archive
+    shutil.make_archive(os.path.join(root, 'DLRSD'), 'zip', root, 'DLRSD')
+    shutil.rmtree(folder_path)
```
adamjstewart: Not sure if we want this. When you run the trainer tests, does it unpack the zip file and leave extra files that aren't tracked by git? If so, we should add those to git.

tests/datasets/test_dlrsd.py
```diff
@@ -0,0 +1,124 @@
...
+from torch.utils.data import ConcatDataset
+
+from torchgeo.datasets import DLRSD, DatasetNotFoundError, DLRSDMultilabel
+
+
+class TestDLRSD:
+    @pytest.fixture
+    def dataset(self, monkeypatch: MonkeyPatch, tmp_path: Path) -> DLRSD:
+        url = os.path.join('tests', 'data', 'dlrsd') + os.sep
+        md5 = '6472a8d525ab11aa682f0fea86cbeb3a'
```
adamjstewart: Can ignore md5 patching, we no longer do this for any other datasets.

torchgeo/datamodules/dlrsd.py
```diff
@@ -0,0 +1,161 @@
...
+                    self.val_split_pct,
+                    self.test_split_pct,
+                ],
+                generator,
+            )
+        else:
+            self.train_dataset, self.test_dataset = random_split(
+                dataset, [1 - self.test_split_pct, self.test_split_pct], generator
+            )
+            self.val_dataset = self.train_dataset
```
adamjstewart, line 89: This seems like a bad idea. Why not always assert a proper train/val/test split?

torchgeo/datamodules/dlrsd.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+class DLRSDMultilabelDataModule(NonGeoDataModule):
```
adamjstewart: Could probably be a subclass of DLRSDDataModule if you want to save some lines of code.

torchgeo/datasets/dlrsd.py
```diff
@@ -0,0 +1,553 @@
...
+    16. water
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://doi.org/10.3390/rs10060964
+
+    .. versionadded:: 0.10
+    """
+
+    url = 'https://hf.co/datasets/torchgeo/dlrsd/resolve/073a50e4ec8f531510f0ee72f2cf953816e01fd7/'
```
adamjstewart: All new datasets in the TorchGeo HF require majority approval by the TorchGeo TSC before they can be added. In particular, this repo is missing any reference to the original source and licensing info. I would suggest moving to your private account before we figure out the licensing question and get approval.

torchgeo/datasets/dlrsd.py
```diff
@@ -0,0 +1,553 @@
...
+
+    filename = 'DLRSD.zip'
+    md5 = '2df826d636fb68bf36c0250ae62a9161'
```
adamjstewart: We could start using sha256 if you want, we no longer have to use md5. At some point I'll update all the HF datasets to sha256 since it's reported via the web interface.

torchgeo/datasets/dlrsd.py
```diff
@@ -0,0 +1,553 @@
...
+        labels_dir = os.path.join(root, self.directory, 'Labels')
+
+        for cls in sorted(os.listdir(images_dir)):
+            cls_images_dir = os.path.join(images_dir, cls)
+            cls_labels_dir = os.path.join(labels_dir, cls)
+            for img in sorted(os.listdir(cls_images_dir)):
+                if img.endswith('.tif'):
+                    self.image_fns.append(os.path.join(cls_images_dir, img))
+                    mask_fn = img.replace('.tif', '.png')
+                    self.mask_fns.append(os.path.join(cls_labels_dir, mask_fn))
```
adamjstewart: Could this whole thing be replaced by a glob?

torchgeo/datasets/dlrsd.py
```diff
@@ -0,0 +1,553 @@
...
+        Args:
+            index: index to return
+
+        Returns:
+            the image
+        """
+        path = self.image_fns[index]
+        with Image.open(path) as img:
+            array: np.typing.NDArray[np.int_] = np.array(img.convert('RGB'))
+            tensor = torch.from_numpy(array).permute((2, 0, 1)).to(torch.float32)
```
adamjstewart: Optional: einops.rearrange is preferred

torchgeo/datasets/dlrsd.py
```diff
@@ -0,0 +1,553 @@
...
+            if showing_predictions:
+                axs[2].set_title('Predictions')
+
+        if suptitle is not None:
+            plt.suptitle(suptitle)
+
+        return fig
+
+
+class DLRSDMultilabel(NonGeoDataset):
```
adamjstewart: Could have a shared mixin to avoid a lot of code duplication.

adamjstewart (changes requested):
Bunch of minor comments and major licensing question, everything else looks good!

torchgeo/datasets/dlrsd.py
adamjstewart: This also hits the question of whether or not datasets scraped from Google Earth high-res imagery are legal.

docs/api/datasets/benchmark.csv
calebrob6: The dataset is from here https://sites.google.com/view/zhouwx/dataset -- they don't provide a license, but use the UCM imagery.

tests/data/dlrsd/data.py
calebrob6: Out of curiosity, why not?

tests/datasets/test_dlrsd.py
calebrob6: Out of curiosity, why not?

torchgeo/datamodules/dlrsd.py
calebrob6: https://research.google/blog/accelerating-scientific-discovery-with-ai-powered-empirical-software/ uses an undefined 80/20 train/test split

torchgeo/datasets/dlrsd.py
calebrob6: Out of curiosity, why not? (I'm documenting this)

torchgeo/datasets/dlrsd.py
calebrob6: Maybe but I like to be explicit about the sorting

torchgeo/datasets/dlrsd.py
calebrob6: Why? permute should be faster

docs/api/datasets/benchmark.csv
adamjstewart: In that case the dataset is unlicensed and therefore not open source. We can't redistribute it via our HF, but you can use it at your own discretion. I would change the license to unknown until we get confirmation from the authors:
```suggestion
:ref:`DLRSD`,S,USGS National Map,"2,100",17,256x256,0.3,RGB,public domain
:ref:`DLRSD Multilabel`,C,USGS National Map,"2,100",17,256x256,0.3,RGB,
```

tests/data/dlrsd/data.py
adamjstewart: See #3367 and #3428

tests/datasets/test_dlrsd.py
adamjstewart: See #3367 and #3428

torchgeo/datamodules/dlrsd.py
adamjstewart: We can follow better practices than some silly _start-up_ 🙂

torchgeo/datasets/dlrsd.py
adamjstewart: MD5 is cryptographically insecure and can be broken. SHA256 has not yet been broken, which is why HF prefers it. It's also easier to calculate since you don't actually need to download the file and calculate it yourself, you can simply trust HF.

torchgeo/datasets/dlrsd.py
adamjstewart: Yes, you can sort a glob too. I'm talking about the doubly-nested os.listdir for-loop that could be removed.

torchgeo/datasets/dlrsd.py
adamjstewart: Compared to file I/O, both are $$\mathcal{O}(0)$$. Rearrange is more clear and more maintainable.

torchgeo/datamodules/dlrsd.py
calebrob6: I don't want to change this as the standard has been set by prior publications. Creating a fixed train/val/test might give worse results. The dataset is so small people should do multiple random runs.

isaaccorley (dismissed):
This LGTM and we probably do want to use the same split protocol that was proposed in the paper so that it's closer to the original

docs/api/datasets/benchmark.csv
```diff
@@ -14,6 +14,8 @@ Dataset,Task,Source,# Samples,# Classes,Size (px),Resolution (m),Bands,License
 :ref:`DFC2022`,S,Aerial,"3,981",15,"2,000x2,000",0.5,RGB,CC-BY-4.0
 :ref:`DIOR`,OD,Aerial,"23,463",20,"800x800",0.5,RGB,CC-BY-NC-4.0
 :ref:`DL4GAM`,S,"Sentinel-2","2,251 or 11,440","2","256x256","10","MSI",CC-BY-4.0
+:ref:`DLRSD`,S,USGS National Map,"2,100",17,256x256,0.3,RGB,public domain
+:ref:`DLRSD` Multilabel,C,USGS National Map,"2,100",17,256x256,0.3,RGB,public domain
```
adamjstewart: If I recall you confirmed the public domain license via email with the original dataset authors? Just trying to leave a paper trail in case anyone asks where this came from.

tests/datasets/test_dlrsd.py
```diff
@@ -0,0 +1,124 @@
...
+import pytest
+import torch
+import torch.nn as nn
+from pytest import MonkeyPatch
+from torch.utils.data import ConcatDataset
+
+from torchgeo.datasets import DLRSD, DatasetNotFoundError, DLRSDMultilabel
+
+
+class TestDLRSD:
```
adamjstewart, line 19: If the tests are basically identical for both classes, we could parametrize this like we do for EuroSAT and cut down total # lines by half.

torchgeo/datamodules/dlrsd.py
```diff
@@ -0,0 +1,100 @@
...
+    """LightningDataModule implementation for the DLRSD dataset.
+
+    Uses random train/val/test splits.
+
+    .. versionadded:: 0.10
+    """
+
+    std = torch.tensor(255)
+
+    _dataset_cls: type[NonGeoDataset] = DLRSD
```
adamjstewart: Could be:

```suggestion
    _dataset_cls: type[DLRSDBase] = DLRSD
```

torchgeo/datamodules/dlrsd.py
```diff
@@ -0,0 +1,100 @@
...
+    ) -> None:
+        """Initialize a new DLRSDDataModule instance.
+
+        Args:
+            batch_size: Size of each mini-batch.
+            num_workers: Number of workers for parallel data loading.
+            val_split_pct: Fraction of dataset to use for validation. If 0, the
+                train set is also used for validation.
+            test_split_pct: Fraction of dataset to use for testing.
+            seed: Random seed for reproducible train/val/test splits.
```
adamjstewart, line 45: We technically don't expose this in other datamodules, but probably should. NeurIPS is now getting more specific and asking for seed experiments which parts were random and which weren't.

torchgeo/datasets/dlrsd.py
```diff
@@ -0,0 +1,515 @@
...
+    def _extract(self) -> None:
+        """Extract the dataset."""
+        filepath = os.path.join(self.root, self.filename)
+        extract_archive(filepath)
+
+
+class DLRSD(DLRSDBase):
+    """DLRSD semantic segmentation dataset.
+
+    The `DLRSD <https://sites.google.com/view/zhouwx/dataset>`__
```
adamjstewart: A lot of this description is duplicated below, can we move this to the base class instead since it's shared? I realize this may confuse users who have to look in two places though. Especially for shorter things like "cite this paper" you could probably convince me to duplicate it.

torchgeo/datasets/dlrsd.py
```diff
@@ -0,0 +1,515 @@
...
+        Returns:
+            a matplotlib Figure with the rendered sample
+        """
+        image = einops.rearrange(sample['image'].numpy(), 'c h w -> h w c')
+        if image.max() > 1:
+            image = image.astype(np.float32) / 255.0
+
+        # Use a uint8 0-255 copy of the image for the segmentation overlay so
+        # that user transforms which normalize the tensor to [0, 1] don't
+        # produce a near-black overlay (draw_semantic_segmentation_masks
```
adamjstewart, line 312: We fixed this in datamodules (by reverting the normalization) but it's true that this doesn't help for user transforms. Another win for `quantile_normalization` 😉

torchgeo/datasets/dlrsd.py
```diff
@@ -0,0 +1,515 @@
...
+        super().__init__(root, transforms, download, checksum)
+
+        images_dir = os.path.join(root, self.directory, 'Images')
+        self.image_fns: list[str] = sorted(
+            glob.glob(os.path.join(images_dir, '*', '*.tif'))
+        )
+
+        self.multilabels: dict[str, list[int]] = {}
+        csv_path = os.path.join(root, self.directory, 'multilabels.csv')
+        with open(csv_path) as f:
```
adamjstewart: I used to use csv for this but switched to pandas after we made it required by default, may be a bit faster for larger files (but doesn't matter too much for this specific dataset).

adamjstewart (dismissed):
A few more comments on how we could get this from 900 lines to a more maintainable 700 lines but my main concerns have been addressed so I don't want you to wait too much longer for my review!

torchgeo/datasets/dlrsd.py
calebrob6: Yeah cite this paper definitely should stay

tests/datasets/test_dlrsd.py
calebrob6: Okay I copied the EuroSAT pattern

docs/api/datasets/benchmark.csv
```diff
@@ -14,8 +14,8 @@ Dataset,Task,Source,# Samples,# Classes,Size (px),Resolution (m),Bands,License
 :ref:`DFC2022`,S,Aerial,"3,981",15,"2,000x2,000",0.5,RGB,CC-BY-4.0
 :ref:`DIOR`,OD,Aerial,"23,463",20,"800x800",0.5,RGB,CC-BY-NC-4.0
 :ref:`DL4GAM`,S,"Sentinel-2","2,251 or 11,440","2","256x256","10","MSI",CC-BY-4.0
-:ref:`DLRSD`,S,USGS National Map,"2,100",17,256x256,0.3,RGB,public domain
-:ref:`DLRSD` Multilabel,C,USGS National Map,"2,100",17,256x256,0.3,RGB,public domain
+:ref:`DLRSD`,S,USGS National Map,"2,100",17,256x256,0.3,RGB,CC-BY-4.0
```
calebrob6, line 17: The email from the author actually says CC-BY-4.0

isaaccorley (approved):

adamjstewart:
@calebrob6 can you fix these warnings in CI?
```
tests/trainers/test_classification.py::TestClassificationTask::test_trainer[True-dlrsd_multilabel]
tests/trainers/test_classification.py::TestClassificationTask::test_trainer[True-dlrsd_multilabel]
tests/trainers/test_classification.py::TestClassificationTask::test_trainer[True-dlrsd_multilabel]
tests/trainers/test_segmentation.py::TestSemanticSegmentationTask::test_trainer[True-dlrsd]
  /home/runner/work/torchgeo/torchgeo/torchgeo/datamodules/dlrsd.py:76: UserWarning: Length of split at index 2 is 0. This might result in an empty dataset.
    self.train_dataset, self.val_dataset, self.test_dataset = random_split(

tests/trainers/test_classification.py::TestClassificationTask::test_trainer[True-dlrsd_multilabel_no_val]
tests/trainers/test_classification.py::TestClassificationTask::test_trainer[True-dlrsd_multilabel_no_val]
tests/trainers/test_classification.py::TestClassificationTask::test_trainer[True-dlrsd_multilabel_no_val]
tests/trainers/test_segmentation.py::TestSemanticSegmentationTask::test_trainer[True-dlrsd_no_val]
  /home/runner/work/torchgeo/torchgeo/torchgeo/datamodules/dlrsd.py:86: UserWarning: Length of split at index 1 is 0. This might result in an empty dataset.
    self.train_dataset, self.test_dataset = random_split(
```

PR #3506, opened 2026-03-11
 Use TypedDict for OpenStreetMap classes parameter

ashnair1:
## Summary by CodeRabbit

* **Refactor**
  * Strengthened type safety for the OpenStreetMap dataset with improved type annotations, ensuring more robust class configuration handling.

* **Tests**
  * Updated OpenStreetMap dataset tests to use explicit typing for class configurations.

* **Chores**
  * Updated documentation configuration to suppress internal type warnings.


isaaccorley (approved):

adamjstewart:
Definitely a step in the right direction, but there are plenty of Any remaining we should still squash. My end goal is to remove all uses of Any other than `*args: Any, **kwargs: Any`.

PR #3499, opened 2026-03-08
Add tessera embeddings notebook

enomis-dev:
This PR introduces an exploratory notebook demonstrating a complete workflow for training TorchGeo models using Tessera embeddings for pixel-wise classification tasks.

The notebook covers:
- download of tessera embeddings
- data preparation with TesseraEmbeddings torchgeo RasterDataset and samplers
- training classical ML linear model for crop type classification
- training simple deep learning model for crop type classification
- evaluation and visual analysis

Closes #3469
Blocked by https://github.com/torchgeo/torchgeo/pull/3828

PR #3515, opened 2026-03-15
Add PixelDatasetWrapper

Isabelle-Wittmann:
Introduces a `PixelDatasetWrapper` to support pixel-based workflows, particularly useful for embedding-based pipelines.

`PixelDatasetWrapper` is a lightweight wrapper around an existing GeoDataset that:
* returns pixel-level samples (C,) instead of patch samples (C, H, W)
* renames image / mask to pixel / label to better reflect pixel-based learning workflows
* optionally handles `ignore_index` by retrying sampling when ignored labels are encountered (bounded by `max_attempts`)

While similar workflows are technically possible using existing datasets together with e.g. RandomGeoSampler(size=1), you still need to manually extract pixels from (C,1,1) patches, implement ignore logic, and often write a custom collate_fn. This wrapper should provide a clearer and more convenient interface for pixel-based workflows.

For example, in workflows such as #3499 this simplifies the pipeline to:

```
from torchgeo.datasets import TesseraEmbeddings, CDL, PixelDatasetWrapper

dataset = TesseraEmbeddings(path) & CDL(paths="./data", years=[2024])
pixels = PixelDatasetWrapper(dataset)

loader = DataLoader(
    pixels,
    sampler=RandomGeoSampler(dataset, size=1, length=5000),
    batch_size=64,
)

for batch in loader:
    x = batch["pixel"]  # (B, C)
    y = batch["label"]  # (B,)
```

Happy to also add docs if this is considered a useful addition :) 

## Summary by CodeRabbit

* **New Features**
  * PixelDatasetWrapper added to the public API: allows extracting pixel-level samples from geospatial datasets, returns a single pixel and label, warns when patches are larger than 1x1, supports ignoring specific label values and configurable retry behavior.

* **Tests**
  * Added comprehensive tests covering normal and edge cases: non-1x1 patches, ignore-index handling, retry behavior, length passthrough, 3D mask handling, and error conditions when retries fail.


adamjstewart:
Thanks for the PR!

Can't say I fully understand how it works. It seems like it just converts a dataset from returning an image to only returning a single pixel per image? So we would still have heavy I/O and load entire images at a time, but we wouldn't use most pixels in the image and therefore miss a lot of locations during GridGeoSampler?

I also thought about this, specifically for using models like Presto for semantic segmentation. I think there are 3 different places we could handle this:

1. In the dataset: similar to this PR, but iterate over all pixels and yield instead of return
2. In the sampler: basically same as `size=1` with our existing samplers, kinda slow unless you use RandomBatchGeoSampler
3. In the trainer: iterate over all pixels during model usage

I like 3 because it's fast, but would require more custom code. 2 and 1 are also nice because they would be compatible with the Classification1DTask I'm planning on writing. We may need logic in the trainer that basically says "if BxCxHxW and HxW are both 1, then use BxC. I think that's what you're actually going for here?

Isabelle-Wittmann:
Hi, thanks! In its current form the Wrapper is intended to be combined with a Sampler that returns 1x1 patches otherwise it throws a warning to make users aware of exactly such I/O inefficiencies. Originally, I also had a PixelSampler added, but as that really boils down to setting` size=1` on any patch sampler, I didn't see value in adding that.

I tested it in a use case where I used samplers with `size=1` for train and val (to train on _sparse_ pixel points), and then I used a large patch without the PixelWrapper for dense inference (If we end up pushing something like this, we could also unify to have a similar dense inference pixel wrapper).

The main value I see in the current PR is: a) Clearer interface, b) handles the [c,1,1] > [c] reshape + filtering for invalid labels, which makes train loops simpler (no reshapes, or varying batch sizes, see the Tessera sample notebook).

I think,
1) has the disadvantage that you want to use most embedding datasets for both patch and pixel-style training. We can add a second wrapper that does what you are describing: Take dense patches and return them as pixels. I think a wrapper is cleaner for this than doing it on dataset level
2) That's essentially what I'm doing right now, sampler size=1 + wrapper for additional convenience. You're right about the speed though, maybe we can cache sparse embedding vectors for such trainings?
3) That also makes a lot of sense, the two issues I see here: do most people use Trainers right now, or rather the dataset and samplers standalone? The advantage of not having it in a trainer, is that you can much simpler combine it with random forests, linear probes etc. on top, and don't have to use trainers.

... to sum up again, I think its important to distinguish two workflows here:
- sparse pixel training: My PR addresses this, but this is generally slower because of pixel_count-times raster reads 
- "dense" pixel training: take larger image patches and transform to pixel-label pairs, less raster reads, but maybe less diverse training sets, def. useful for dense inference

would be nice if both would be covered, or what do you think?

adamjstewart:
> handles the [c,1,1] > [c] reshape + filtering for invalid labels, which makes train loops simpler (no reshapes, or varying batch sizes, see the Tessera sample notebook).

To counter this, instead of maintaining these 250 lines of code, we could instead maintain 1 line of code in the trainer:
```
image = einops.reshape(image, 'c 1 1 -> c')
```

> do most people use Trainers right now, or rather the dataset and samplers standalone?

I have no idea. I suspect that trainers are more common for benchmark datasets and less common for geospatial datasets, but don't know the overall user body. We should consider having annual user surveys and include this in a questionnaire.

> The advantage of not having it in a trainer, is that you can much simpler combine it with random forests, linear probes etc. on top, and don't have to use trainers.

This is a very good point. I was considering writing trainers for these, but 1) it may not make sense to have sklearn support in _Torch_ Geo, and 2) it defeats the purpose since Lightning is kinda like an sklearn wrapper anyway. @calebrob6 may have opinions here.

Isabelle-Wittmann:
Yes, I get that, and happy to solve this like this/with less new code 👍 Just to mention two other points I've found  convenient in my workflow: 
- You can pass a Sampler to the wrapper + plus an ignore label, and it gives you consistent batch-sized, valid-label samples, especially in real-world dataset settings this is useful as you avoid impact of non-valid label datapoints on the training due to a varying number of datapoints that contribute to your loss in each gradient step. (Which I would argue is more important for low sample number pixel workflows, than dense patch training) 
- If we end up adding more functionalities for this, it could be nice to establish a pixel/label return instead of image/mask

adamjstewart:
> it could be nice to establish a pixel/label return instead of image/mask

This reminds me, "pixel" is not compatible with any Kornia augmentations, that's why we use "image" instead. "C" isn't compatible either, only "C x 1 x 1". Of course, there aren't too many augmentations you would want to use in single-pixel workflows. Maybe just normalization and occasionally color jitter.

Isabelle-Wittmann:
Ok, good to know! How about I instead add the Wrapper logic as a tutorial notebook for now?

adamjstewart:
Let me think on it for a couple days. I'm currently in the process of redesigning our samplers and trainers for time series support. There's a chance that this could be part of that refactor.

PR #3521, opened 2026-03-19
Added MAE Trainer

DarthReca:
This is a classical MAE training for ViTs, based on Lightly and Lightning. Some important notes are:

 1. extract_backbone is currently commented, because it seems to work only for ConvNets
 2. I have introduced two tests on SSL4EO-S12, but I can extend it if needed
 3. I have added some metrics, but I am open to suggestions


adamjstewart:
@wangyi111 @xiong-zhitong can you review?

torchgeo/trainers/mae.py
```diff
@@ -0,0 +1,245 @@
...
+# Licensed under the MIT License.
+
+"""MAE trainer for self-supervised learning (SSL)."""
+from typing import Any
+
+import torch
+from lightly.models import utils
+
+try:
+    from lightly.models.modules import MAEDecoderTIMM, MaskedVisionTransformerTIMM
```
DarthReca: There are some issues with this import, due to the changes made in lighly. In older versions, they were using different apis rather than TIMM. [Example old version](https://github.com/lightly-ai/lightly/blob/053327dac516cf118e71fc9ae9128dfcae3b9b94/lightly/models/modules/__init__.py).

I am not sure how to proceed

torchgeo/trainers/mae.py
```diff
@@ -2,16 +2,12 @@
...
 
 import torch
 from lightly.models import utils
-
-try:
-    from lightly.models.modules import MAEDecoderTIMM, MaskedVisionTransformerTIMM
-except ImportError:
-    from lightly.models import MAEDecoderTIMM, MaskedVisionTransformerTIMM
-
+from lightly.models.modules import MAEDecoderTIMM, MaskedVisionTransformerTIMM
```
DarthReca, line 10: There are some issues with this import, due to the changes made in lighly. In older versions, they were using different apis rather than TIMM. [Example old version](https://github.com/lightly-ai/lightly/blob/053327dac516cf118e71fc9ae9128dfcae3b9b94/lightly/models/modules/__init__.py).

I am not sure how to proceed

tests/trainers/test_mae.py
```diff
@@ -0,0 +1,116 @@
...
+
+class TestMAETask:
+    @pytest.mark.parametrize('name', ['ssl4eo_s12_mae_1', 'ssl4eo_s12_mae_2'])
+    def test_trainer(
+        self, monkeypatch: MonkeyPatch, name: str, fast_dev_run: bool
+    ) -> None:
+        config = os.path.join('tests', 'conf', name + '.yaml')
+
+        if name.startswith('seco'):
+            monkeypatch.setattr(SeasonalContrastS2, '__len__', lambda self: 2)
```
adamjstewart: Technically not needed yet, but I would be fine with adding SeCo tests too. Not really necessary though.

torchgeo/trainers/mae.py
```diff
@@ -0,0 +1,261 @@
...
+            decoder_dim: The embedding dimension of the MAE decoder. Typically 512 is a good choice for ViT-Base encoders.
+            lr: _Should be 1.5e-4 * batch_size / 256.
+            decoder_num_heads: Number of attention heads in the MAE decoder.
+            decoder_depth: Number of layers in the MAE decoder. Typically 1-4 layers is sufficient for good performance.
+            weight_decay: Weight decay for the AdamW optimizer.
+            mask_ratio: The ratio of tokens to mask during training. Typically 0.75 is a good choice.
+        """
+        self.transform = (
+            transform if transform is not None else MAETransform(normalize=False)  # type: ignore
+        )
```
adamjstewart:
```suggestion
        self.transform: nn.Module = (
            transform if transform is not None else MAETransform(normalize=False) 
        )
```
See if this works

torchgeo/trainers/mae.py
```diff
@@ -0,0 +1,261 @@
...
+                'scheduler': scheduler,  # type: ignore
+                'interval': 'epoch',
+            },
+        }
+
+    def lr_scheduler_step(
+        self, scheduler: torch.optim.lr_scheduler.LRScheduler, metric: Any | None
+    ) -> None:
+        """Step the learning rate scheduler."""
+        scheduler.step(epoch=self.current_epoch)
```
adamjstewart: Is this necessary? I haven't seen this before

torchgeo/trainers/mae.py
```diff
@@ -0,0 +1,261 @@
...
+        optim = torch.optim.AdamW(
+            self.parameters(),
+            lr=self.hparams['lr'],
+            weight_decay=self.hparams['weight_decay'],
+            betas=(0.9, 0.95),
+        )
+        max_epochs = 800
+        if self.trainer and self.trainer.max_epochs is not None:
+            max_epochs = self.trainer.max_epochs
+        scheduler = CosineLRScheduler(
```
adamjstewart:
```suggestion
        scheduler: LRScheduler = CosineLRScheduler(
```
There is a bug in ty that prevents it from understanding the type hints well here. Take a look at the latest version of other SSL trainers to see what hacks I needed to make ty happy.

adamjstewart (commented):
Just a few minor comments on type hints, otherwise just waiting for someone more knowledgeable than me to review.

torchgeo/trainers/mae.py
adamjstewart: I would be completely fine with bumping the min supported version of lightly to 1.5.1 if you want. Just make the change to `pyproject.toml` and run `uv sync` to update `uv.lock` and commit both. Alternative is to use try-except import branches for different versions of lightly.

torchgeo/trainers/mae.py
```diff
@@ -0,0 +1,241 @@
...
+
+        # Load weights
+        if weights and weights is not True:
+            if isinstance(weights, WeightsEnum):
+                state_dict = weights.get_state_dict(progress=True)
+            # elif os.path.exists(weights):
+            #    _, state_dict = extract_backbone(weights)
+            else:
+                state_dict = get_weight(weights).get_state_dict(progress=True)
+            load_state_dict(vit, state_dict)  # type: ignore[invalid-argument-type]
```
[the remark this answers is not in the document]
adamjstewart: Agreed, does this just require an extra line of coverage, or was there some issue with our extract_backbone function for ViTs?

torchgeo/trainers/mae.py
DarthReca: It is a necessary "workaround" because it is a TIMM scheduler, not a PyTorch one. It needs the current epoch to make the warmup

tests/trainers/test_mae.py
DarthReca: About the need for more training tests I am open to add more config if needed

torchgeo/trainers/mae.py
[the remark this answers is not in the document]
DarthReca: extract_backbone seems not working properly for ViTs. Can @isaaccorley confirm this?

torchgeo/trainers/mae.py
DarthReca: If we can go for upgrading without breaking everything, it should be more robust, probably.

torchgeo/trainers/mae.py
```diff
@@ -0,0 +1,261 @@
...
+        decoder_depth: int = 1,
+        weight_decay: float = 0.05,
+        mask_ratio: float = 0.75,
+        size: int = 224,
+    ) -> None:
+        """Initialize the MAE task.
+
+        Args:
+            model: The ViT architecture to use for the encoder. Must be compatible with timm's create_model function.
+            weights: Pretrained weights to initialize the encoder with. Can be a timm WeightsEnum, a path to a checkpoint file, a string identifier for a timm weight, True to use default pretrained weights, or None for random initialization.
```
adamjstewart: Each line should be wrapped to 88 characters

pyproject.toml
```diff
@@ -48,7 +48,7 @@ dependencies = [
     # lightly 1.4.5+ required for LARS optimizer
     # lightly 1.4.26 is incompatible with the version of timm required by smp
     # https://github.com/torchgeo/torchgeo/issues/1824
```
```suggestion
    # lightly 1.5.1+ required for lightly.models.modules.MAEDecoderTIMM
```

adamjstewart (commented):
Couple minor comments, otherwise this is close to being ready.

torchgeo/trainers/mae.py
```diff
@@ -0,0 +1,261 @@
...
+        idx_keep, idx_mask = utils.random_token_mask(
+            size=(batch_size, self.sequence_length),
+            mask_ratio=self.hparams['mask_ratio'],
+            device=images.device,
+        )
+        x_pred = self.forward(images, idx_keep, idx_mask)
+        # get image patches for masked tokens
+        patches = utils.patchify(images, self.patch_size)
+        # must adjust idx_mask for missing class token
+        target = utils.get_at_index(patches, idx_mask - 1)
```
xiong-zhitong: The - 1 assumes exactly one prefix token, but timm ViTs also support class_token=False and register tokens. Could we use vit.num_prefix_tokens instead to stay robust?

torchgeo/trainers/mae.py
```diff
@@ -0,0 +1,261 @@
...
+
+        # per-sample loss for std logging
+        loss_per_sample = self.criterion(x_pred, target)
+        loss_per_sample = loss_per_sample.mean(dim=list(range(1, loss_per_sample.ndim)))
+        loss = loss_per_sample.mean()
```
xiong-zhitong, line 243: The original MAE paper uses per-patch normalized pixels as the reconstruction target by default, and the ablation shows it meaningfully improves downstream performance. Could we add norm_pix_loss: bool = True and normalize target over the last dim (per-patch) before the MSE?

torchgeo/trainers/mae.py
```diff
@@ -0,0 +1,261 @@
...
+
+        # Load weights
+        if weights and weights is not True:
+            if isinstance(weights, WeightsEnum):
+                state_dict = weights.get_state_dict(progress=True)
+            # elif os.path.exists(weights):
+            #    _, state_dict = extract_backbone(weights)
+            else:
+                state_dict = get_weight(weights).get_state_dict(progress=True)
+            load_state_dict(vit, state_dict)  # type: ignore[invalid-argument-type]
```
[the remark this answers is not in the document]
DarthReca: Currently unsupported; let's see when we update the extract_backbone

torchgeo/trainers/mae.py
```diff
@@ -0,0 +1,287 @@
...
+def mae_augmentation(size: int = 224) -> K.AugmentationSequential:
+    """Get the default MAE augmentations as a Kornia AugmentationSequential module."""
+    return K.AugmentationSequential(
+        K.RandomResizedCrop(
+            size=(size, size),
+            scale=(0.2, 1.0),
+            ratio=(3 / 4, 4 / 3),
+            resample='bicubic',
+        ),
+        K.RandomHorizontalFlip(p=0.5),
```
adamjstewart, line 32: We probably also want random vertical flip by default? There is no up and down in satellite imagery.

pyproject.toml
adamjstewart, line 1: Can you undo all changes to this file other than bumping the min lightly version?

DarthReca: Yup I can try. I think it is ruff doing things along with uv

adamjstewart: Ruff should only modify python files, and uv should only modify uv.lock, I don't think either ever changes toml files.

DarthReca: Ok, so let me double check what is acting automatically on the file

adamjstewart (approved):
Looks good to me now! Let's get this in so people can start playing with it. We've needed this for a long time, this will be a great addition.

P.S. If anyone wants to add other SSL trainers, DINO is probably the next biggest priority.

PR #3526, opened 2026-03-22
Contributing: document files to modify when adding new models

adamjstewart:
@isaaccorley @piperwolters can you double check this and make sure I didn't miss anything? I didn't include things like pyproject.toml/uv.lock because we don't often add new deps when adding new models.

Pretty sure @calebrob6 was looking for something like this in our contributing docs, we already have one for datasets.


adamjstewart:
Let's actually move this to a PR template.

adamjstewart:
Actually, multiple PR templates is currently more confusing than useful. Let's still go ahead with this PR and convert to PR templates once the selector option is in public preview.

isaaccorley (approved):
Looks like it covers all the files

PR #3537, opened 2026-03-25
[FIX] regex idx retrieval from filename in inria dataset

### LINKED ISSUE #3533: Inria Dataset Regex Bug
### Description

Dear torchgeo community,

first of all, thank you for maintaining this fantastic package! :)

When scaling code on an HPC system, I encountered an issue with the `InriaDataset` (`datasets/inria.py`, around line 119): the regex used to extract the image number can match a number from the full path instead of the actual filename.  

In particular, the default extraction process for the Inria dataset can produce directory names that contain numbers (e.g. `"2"`), so the regex will always return that directory number rather than the intended image number. This becomes especially problematic on HPC systems where usernames or project directories often include numeric IDs.

I have a fix for this and would be happy to open a PR. Please let me know if you'd be interested in a contribution, or how you’d prefer this to be addressed.
Happy hacking
Moritz

### Steps to reproduce

1. Download Inria Dataset and unpack using the commands from the Inria website.

```
dm = InriaAerialImageLabelingDataModule(batch_size=2, root=<basedir>)  # basedir contains a number such as user123/...
```

### Version

0.9.0

mw3iss:
Fixes #3533

regex matching a number somewhere in the path to the data dir instead of the number from the filename, which was the intended behaviour for filenames such as austin10.tif matches "10".

### DIFF
--- a/torchgeo/datasets/inria.py
+++ b/torchgeo/datasets/inria.py
@@ -115,7 +115,8 @@ class InriaAerialImageLabeling(NonGeoDataset):
             labels = sorted(labels)
 
             for img, lbl in zip(images, labels):
-                if match := pattern.search(img):
+                fname = os.path.basename(img)
+                if match := pattern.search(fname):
                     idx = int(match.group(2))
                     # For validation, use the first 5 images of every location
                     if self.split == 'train' and idx > 5:


adamjstewart (approved):
Makes sense to me

PR #3539, opened 2026-03-25
[FIX] change augmentation order to improve performance in selected datamodules

mw3iss:
I guess, that the given augmentations in the inria, potsdam and vaihingen datamodules can be swapped to majorly improve performance. The image size is decreased first to reduce the computational cost of all following augmentations and reduce memory footprint ASAP, which is imporrtant for high-resolution image datasets.

### DIFF
--- a/torchgeo/datamodules/l8biome.py
+++ b/torchgeo/datamodules/l8biome.py
@@ -49,10 +49,10 @@ class L8BiomeDataModule(GeoDataModule):
         )
 
         self.train_aug = K.AugmentationSequential(
-            K.Normalize(mean=self.mean, std=self.std),
             K.RandomResizedCrop(_to_tuple(self.patch_size), scale=(0.6, 1.0)),
             K.RandomVerticalFlip(p=0.5),
             K.RandomHorizontalFlip(p=0.5),
+            K.Normalize(mean=self.mean, std=self.std),
             data_keys=None,
             keepdim=True,
             extra_args={
--- a/torchgeo/datamodules/resisc45.py
+++ b/torchgeo/datamodules/resisc45.py
@@ -36,12 +36,12 @@ class RESISC45DataModule(NonGeoDataModule):
         super().__init__(RESISC45, batch_size, num_workers, **kwargs)
 
         self.train_aug = K.AugmentationSequential(
-            K.Normalize(mean=self.mean, std=self.std),
             K.RandomRotation(p=0.5, degrees=90),
             K.RandomHorizontalFlip(p=0.5),
             K.RandomVerticalFlip(p=0.5),
             K.RandomSharpness(p=0.5),
             K.RandomErasing(p=0.1),
+            K.Normalize(mean=self.mean, std=self.std),
             K.ColorJitter(p=0.5, brightness=0.1, contrast=0.1, saturation=0.1, hue=0.1),
             data_keys=None,
             keepdim=True,
--- a/torchgeo/datamodules/spacenet.py
+++ b/torchgeo/datamodules/spacenet.py
@@ -122,12 +122,12 @@ class SpaceNet1DataModule(SpaceNetBaseDataModule):
         )
 
         self.train_aug = K.AugmentationSequential(
-            K.Normalize(mean=self.mean, std=self.std),
             K.PadTo((448, 448)),
             K.RandomRotation(p=0.5, degrees=90),
             K.RandomHorizontalFlip(p=0.5),
             K.RandomVerticalFlip(p=0.5),
             K.RandomSharpness(p=0.5),
+            K.Normalize(mean=self.mean, std=self.std),
             K.ColorJitter(p=0.5, brightness=0.1, contrast=0.1, saturation=0.1, hue=0.1),
             data_keys=None,
             keepdim=True,
--- a/torchgeo/datamodules/treesatai.py
+++ b/torchgeo/datamodules/treesatai.py
@@ -91,9 +91,9 @@ class TreeSatAIDataModule(NonGeoDataModule):
         self.sensors = kwargs.get('sensors', TreeSatAI.all_sensors)
 
         self.train_aug = K.AugmentationSequential(
+            K.Resize(self.patch_size),
             K.RandomVerticalFlip(p=0.5),
             K.RandomHorizontalFlip(p=0.5),
-            K.Resize(self.patch_size),
             data_keys=None,
             keepdim=True,
         )
--- a/torchgeo/datamodules/ucmerced.py
+++ b/torchgeo/datamodules/ucmerced.py
@@ -31,8 +31,8 @@ class UCMercedDataModule(NonGeoDataModule):
         super().__init__(UCMerced, batch_size, num_workers, **kwargs)
 
         self.aug = K.AugmentationSequential(
-            K.Normalize(mean=self.mean, std=self.std),
             K.Resize(size=(256, 256)),
+            K.Normalize(mean=self.mean, std=self.std),
             data_keys=None,
             keepdim=True,
         )
--- a/torchgeo/datamodules/vhr10.py
+++ b/torchgeo/datamodules/vhr10.py
@@ -52,10 +52,10 @@ class VHR10DataModule(NonGeoDataModule):
         self.collate_fn = collate_fn_detection
 
         self.train_aug = K.AugmentationSequential(
-            K.Normalize(mean=self.mean, std=self.std),
             K.RandomHorizontalFlip(),
-            K.ColorJiggle(0.1, 0.1, 0.1, 0.1, p=0.7),
             K.RandomVerticalFlip(),
+            K.Normalize(mean=self.mean, std=self.std),
+            K.ColorJiggle(0.1, 0.1, 0.1, 0.1, p=0.7),
             data_keys=None,
             keepdim=True,
         )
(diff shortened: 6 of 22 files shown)


adamjstewart:
Can we do this for all datamodules with CenterCrop/RandomCrop? There are quite a lot:
```console
> git grep '\(Center\|Random\)Crop' torchgeo/datamodules/*.py
torchgeo/datamodules/chesapeake.py:            K.CenterCrop(patch_size), data_keys=None, keepdim=True
torchgeo/datamodules/deepglobelandcover.py:            K.CenterCrop(self.patch_size),
torchgeo/datamodules/deepglobelandcover.py:            K.RandomCrop(self.patch_size, pad_if_needed=True),
torchgeo/datamodules/gid15.py:            K.RandomCrop(self.patch_size, pad_if_needed=True),
torchgeo/datamodules/gid15.py:            K.CenterCrop(self.patch_size),
torchgeo/datamodules/inria.py:            K.RandomCrop(self.patch_size, pad_if_needed=True),
torchgeo/datamodules/inria.py:            K.CenterCrop(self.patch_size),
torchgeo/datamodules/inria.py:            K.CenterCrop(self.patch_size),
torchgeo/datamodules/levircd.py:                K.RandomCrop(self.patch_size, pad_if_needed=True),
torchgeo/datamodules/levircd.py:                K.Normalize(mean=self.mean, std=self.std), K.CenterCrop(self.patch_size)
torchgeo/datamodules/levircd.py:                K.Normalize(mean=self.mean, std=self.std), K.CenterCrop(self.patch_size)
torchgeo/datamodules/levircd.py:                K.RandomCrop(self.patch_size, pad_if_needed=True),
torchgeo/datamodules/levircd.py:                K.Normalize(mean=self.mean, std=self.std), K.CenterCrop(self.patch_size)
torchgeo/datamodules/levircd.py:                K.Normalize(mean=self.mean, std=self.std), K.CenterCrop(self.patch_size)
torchgeo/datamodules/oscd.py:            K.VideoSequential(K.RandomCrop(self.patch_size)),
torchgeo/datamodules/oscd.py:            K.VideoSequential(K.RandomCrop(self.patch_size)),
torchgeo/datamodules/potsdam.py:            K.CenterCrop(self.patch_size),
torchgeo/datamodules/potsdam.py:            K.RandomCrop(self.patch_size, pad_if_needed=True),
torchgeo/datamodules/reforestree.py:            K.RandomCrop(self.patch_size, pad_if_needed=True),
torchgeo/datamodules/reforestree.py:            K.CenterCrop(self.patch_size),
torchgeo/datamodules/solar_plants_brazil.py:            K.RandomCrop(self.patch_size, pad_if_needed=True),
torchgeo/datamodules/solar_plants_brazil.py:            K.CenterCrop(size=self.patch_size),
torchgeo/datamodules/ssl4eo_benchmark.py:            K.CenterCrop(self.patch_size),
torchgeo/datamodules/ssl4eo_benchmark.py:            K.CenterCrop(self.patch_size),
torchgeo/datamodules/vaihingen.py:            K.RandomCrop(self.patch_size, pad_if_needed=True),
torchgeo/datamodules/vaihingen.py:            K.CenterCrop(self.patch_size),
```

mw3iss:
Absolutely, yes! I just highlighted it for the datasets I am planning to use in the next days. I can apply the changes and update the PR in the next hours.

torchgeo/datamodules/resisc45.py
```diff
@@ -36,12 +36,12 @@ def __init__(
         super().__init__(RESISC45, batch_size, num_workers, **kwargs)
 
         self.train_aug = K.AugmentationSequential(
-            K.Normalize(mean=self.mean, std=self.std),
             K.RandomRotation(p=0.5, degrees=90),
             K.RandomHorizontalFlip(p=0.5),
             K.RandomVerticalFlip(p=0.5),
             K.RandomSharpness(p=0.5),
             K.RandomErasing(p=0.1),
+            K.Normalize(mean=self.mean, std=self.std),
```
adamjstewart, line 44: Why change this? This seems unrelated to speed. Our goal isn't to make Normalize last, it's to make Crop first.

torchgeo/datamodules/spacenet.py
```diff
@@ -122,12 +122,12 @@ def __init__(
         )
 
         self.train_aug = K.AugmentationSequential(
-            K.Normalize(mean=self.mean, std=self.std),
             K.PadTo((448, 448)),
             K.RandomRotation(p=0.5, degrees=90),
             K.RandomHorizontalFlip(p=0.5),
             K.RandomVerticalFlip(p=0.5),
             K.RandomSharpness(p=0.5),
+            K.Normalize(mean=self.mean, std=self.std),
```
adamjstewart, line 130: This actually looks slower than the previous implementation?

torchgeo/datamodules/treesatai.py
```diff
@@ -91,9 +91,9 @@ def __init__(
         self.sensors = kwargs.get('sensors', TreeSatAI.all_sensors)
 
         self.train_aug = K.AugmentationSequential(
+            K.Resize(self.patch_size),
             K.RandomVerticalFlip(p=0.5),
             K.RandomHorizontalFlip(p=0.5),
-            K.Resize(self.patch_size),
```
adamjstewart, line 96: This also looks slower?

torchgeo/datamodules/ucmerced.py
```diff
@@ -31,8 +31,8 @@ def __init__(
         super().__init__(UCMerced, batch_size, num_workers, **kwargs)
 
         self.aug = K.AugmentationSequential(
-            K.Normalize(mean=self.mean, std=self.std),
             K.Resize(size=(256, 256)),
+            K.Normalize(mean=self.mean, std=self.std),
```
adamjstewart, line 35: No change?

torchgeo/datamodules/vhr10.py
```diff
@@ -52,10 +52,10 @@ def __init__(
         self.collate_fn = collate_fn_detection
 
         self.train_aug = K.AugmentationSequential(
-            K.Normalize(mean=self.mean, std=self.std),
```
adamjstewart, line 55: No change?

adamjstewart (changes requested):
Current PR moves Normalize to the end, but our goal should actually be to move Crop to the beginning. Augmentations like Resize and RandomResizedCrop may increase or decrease image size. I would check the original image size in https://torchgeo.readthedocs.io/en/stable/api/datasets.html#benchmark-datasets and the default resize shape and check whether or not the image size will increase or decrease before changing any of those. For datasets that don't involve any crop/resize, I wouldn't change those without justification.

torchgeo/datamodules/l8biome.py
```diff
@@ -49,10 +49,10 @@ def __init__(
         )
 
         self.train_aug = K.AugmentationSequential(
-            K.Normalize(mean=self.mean, std=self.std),
```
adamjstewart, line 52: Not sure if anything changes here, it is `self.patch_size` before and after RandomResizedCrop. You changed L8 Biome but not L7 Irish which is identical.

torchgeo/datamodules/resisc45.py
mw3iss: check, just wanted to refactor everything cleanly, but I got your point

torchgeo/datamodules/spacenet.py
mw3iss: slower? because the input images are smaller than 448? in this case, yes.. but yes, skip this changes. Not the goal of this PR

torchgeo/datamodules/vhr10.py
mw3iss: skip these changes as well, no cropping here

torchgeo/datamodules/ucmerced.py
mw3iss: first the resize here, isn't it?

torchgeo/datamodules/treesatai.py
mw3iss: only when self.patch_size > image_size.. but is this going to happen?

torchgeo/datamodules/l8biome.py
mw3iss: I may have missed one file, but still the randomResizedCrop should be done before the normalization to safe compute

torchgeo/datamodules/spacenet.py
adamjstewart: Yes, because the input images are smaller and are padded. Changing the order of normalization also changes the padding values, which we probably don't want.

torchgeo/datamodules/treesatai.py
adamjstewart: Yes, this is usually going to happen. The original image sizes are 6x6, 20x20, and 304x304. There is never a case when resizing makes the image smaller, only the case where it becomes bigger.

torchgeo/datamodules/ucmerced.py
adamjstewart: The original images are 256x256, the Resize doesn't change anything

torchgeo/datamodules/l8biome.py
adamjstewart: We aren't saving compute if the patch size is the same before and after normalization.

PR #3541, opened 2026-03-25
[FIX] Potsdam dataset folder structure 

mw3iss:
(sorry for PR bombing, but I guess you prefer more small changes instead of one large commit)
fixes wrong/deprecated folder structure of potsdam dataset. The mask are located in the subfolder '5_Label_all'


mw3iss:
Still working on it.. seems to be different for Windows/Linux.

adamjstewart:
We do prefer more small PRs, thanks for all the fixes!

PR #3546, opened 2026-03-25
docs: add uv installation instructions

nuglifeleoji:
## Summary

Add documentation for installing TorchGeo with uv, the fast Python package manager from Astral.

## Rationale

Issue #2905 tracks the migration from pip to uv. While uv is already used in CI workflows, the installation documentation only mentioned pip, conda, and spack. This PR adds uv instructions so users are aware of this faster option.

## Implementation

Added a new "uv" section to  that:
- Describes uv as a fast Rust-based alternative to pip
- Provides equivalent commands for all common installation scenarios
- Follows the same structure as existing pip section for consistency

## Checklist

- [x] Follows existing documentation style and conventions
- [x] Maintains the same structure as pip/conda/spack sections
- [x] All installation commands tested and valid

## Related Issues

Resolves the documentation portion of #2905

### DIFF
--- a/docs/user/contributing.rst
+++ b/docs/user/contributing.rst
@@ -29,6 +29,12 @@ From there, you can make any changes you want. Once you are satisfied with your 
    $ git commit -m "descriptive commit message"
    $ git push
 
+For developers using `uv <https://docs.astral.sh/uv/>`_, you can sync the project dependencies with:
+
+.. code-block:: console
+
+   $ uv sync
+
 
 For changes to Python code, you'll need to ensure that your code is :ref:`well-tested <tests>` and all :ref:`linters <linters>` pass. When you're ready, you can `open a pull request on GitHub <https://github.com/torchgeo/torchgeo/compare>`_. All pull requests should be made against the ``main`` branch. If it's a bug fix, we will backport it to a release branch for you.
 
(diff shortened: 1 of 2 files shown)


adamjstewart:
@isaaccorley can you review and see if there are any other uv-specific things we should document? I think it's worth documenting `uv sync` at least, it's useful for developers, not just users.

ashnair1:
Should we be documenting "uv-native" (`uv add`) or "pip-based" (`uv pip install`) approach? I suppose this would depend on whether a user is working on torchgeo itself vs adding torchgeo as a dependency to their projects.

isaaccorley (dismissed):
Should use `uv add` throughout and not `uv pip install`

nuglifeleoji:
Thanks for all the comments, I have added the uv pip install as well as the condition for developers!

adamjstewart:
I'm also in favor of uv-native commands everywhere.

adamjstewart (approved):
This now looks quite nice to me, @isaaccorley can you double check that all the commands work?

nuglifeleoji:
Updated — the uv section now uses `uv add` throughout with no `uv pip install` commands remaining. @isaaccorley could you take another look?

Issue #2501, opened 2025-01-05
pytest: improve branch coverage AND fix the tests


adamjstewart:
By default, coverage only reports which lines were executed during unit testing. By enabling branch coverage, we can also see whether if-statements are tested for both True and False conditions.

To enable this, simply add the following to `pyproject.toml`:
```toml
[tool.coverage.run]
branch = true
```
https://app.codecov.io/gh/microsoft/torchgeo/pull/2276/indirect-changes gives an idea of places where this coverage is missing. The real work involves adding new tests to add the missing coverage.

Reference: https://coverage.readthedocs.io/en/latest/branch.html

@isaaccorley edit -- We've had several PRs simply changing the above line in `pyproject.toml`, however the end goal of this issue is to not only change this line but also fix the missing branch tests so we obtain 100% code coverage. Please consider this effort when contributing a PR for this.

modaribrahim:
I'm working on this. PR #3990 adds branch coverage to the MMFlood and BioMassters plot() tests as a first contribution; I'll keep submitting similar small coverage PRs for the remaining datasets.

PR #3550, opened 2026-03-26
test: add branch coverage for _to_tuple function

nuglifeleoji:
Add branch coverage tests for the _to_tuple function in torchgeo/samplers/utils.py.

This is the first small individual PR to improve branch coverage as part of #2501.


nuglifeleoji:
I updated the branch coverage with style similar to #3550 !

PR #3551, opened 2026-03-26
test: add branch coverage for check_integrity function

nuglifeleoji:
Add branch coverage tests for the check_integrity function in torchgeo/datasets/utils.py.

This is another small individual PR to improve branch coverage as part of #2501.


nuglifeleoji:
Closing this PR as the test failures require more investigation than anticipated. Will revisit with a simpler approach.

PR #3552, opened 2026-03-26
Samplers: refactor and add time series support

### LINKED ISSUE #1630: GridGeoSampler on areas smaller than patch size
### Discussed in https://github.com/microsoft/torchgeo/discussions/1624

<div type='discussions-op-text'>

<sup>Originally posted by **mpelchat04** October  5, 2023</sup>
We use the GridGeoSampler to create patches. Although, we have cases where the portion covered by the training data is small (smaller than patch_size), as seen in pink in the image:  
bcf04b15-c161-45de-b9f8-d2e60ca1f8c4
The image and data showed here are taken from the Spacenet-7 dataset.  
Using the GridGeoSampler, these patches are discarded and it seems like an intended feature [ref](https://github.com/microsoft/torchgeo/issues/431#issuecomment-1049503005).  
Is there a logical way to overcome this limitation, where it discards useful patches?</div>  

### Intended behavior is up for discussion: 
**Suggestion:** The minimum value of the intersection bounds could be the patch_size X patch_size.
**Q1:** What if the area of intersection is 0? A: Then no need to have a sampler.
**Q2:** What if we're using a point dataset? A: What is the result of the intersection between a single point and an image tile?  
**Q3:** What if the image tile is also smaller than patch_size?

### LINKED ISSUE #3548: [FR] More explicitly communicate `RandomBatchGeoSampler` same-tile sampling behavior
## Context

I am trying to pretrain a classifier using SimCLR and a geo-based extension. As dataset to learn from, I am using FLAIR#2, a georeferenced, yet tiled dataset. Although I tried numerous different configurations and approaches, my SimCLR behaved undesirably.

As some last resort I tried to manually inspect the batches of data and given the extraordinary image similarity and "scene-equality", I knew something had to go on.

Image

The SimCLR-classifier, based on learning the similarity of two augmentations while increasing the distance of all other images could not learn: The underlying distribution is so simillar that pushing negatives also pushes positives and pulling positives also pulls negatives.

--- 

## Problem

Torchgeo provides custom samplers which give the opportunity to exploit some of the advantages of continuous geodata, e.g. random sampling during training: it diversifies and enlarges dataset while at the same time acting as a regularizer - the same pixels may appear in different images.

This is great in theory, there is an important implementation detail: Iterating over this sampler causes sampling from the same tile over and over for one batch: each batch is essentially one semi-randomly chosen leading to very little diversity per batch. More importantly, these tiles may even be overlapping and thus semi-equal

https://github.com/torchgeo/torchgeo/blob/86f71dc382bbcfc61b2875d59474b2e9c4913ccc/torchgeo/samplers/batch.py#L169-L189

## Fix

I think the documentation is a bit unfortunate. A `RandomGeoSampler` which would not be prone to this behavior directly recommend instead using `RandomBatchGeoSampler` for tile-based datasets. 

> This sampler is not recommended for use with tile-based datasets. Use [RandomBatchGeoSampler](https://torchgeo.readthedocs.io/en/stable/api/samplers.html#torchgeo.samplers.RandomBatchGeoSampler) instead.

The `Random` in the name is very misleading to what it actually does. I understand that the goal were performance gains of this approach, [@adamjstewart mentioned that this might not even be the case](https://github.com/torchgeo/torchgeo/discussions/3516#discussioncomment-16294827).

I would recommend at least one of the two:

1. changing the documentation to clearly communicate this problem
2. add a flag to the `RandomBatchGeoSampler` that can disable same-tile sampling 

### Implementation

```py
    def __iter__(self) -> Iterator[list[tuple[slice, slice, slice]]]:  # type: ignore[override]
        """Return the indices of a dataset.

        Yields:
            Batch of [xmin:xmax, ymin:ymax, tmin:tmax] coordinates to index a dataset.
        """
        for _ in range(len(self)):
            batch = []
            for _ in range(self.batch_size):
                # Choose a random tile, weighted by area
                idx = torch.multinomial(self.areas, 1)
                hit = self.hits[idx]


                # Choose random index within that tile
                bounding_box = get_random_bounding_box(
                    bounds, self.size, self.res, self.generator
                )
                batch.append(bounding_box)


            yield batch

### PR DESCRIPTION
**Pull request by adamjstewart**
This is a ground-up rewrite of TorchGeo's GeoSamplers with improved spatial sampling and new support for time series sampling strategies.

### Motivation

TorchGeo's existing file-based samplers have a number of known limitations and biases:

* No support for time series sampling, time is completely ignored
* RandomGeoSampler: sampling bias for locations/times with more files than others
* GridGeoSampler: locations with multiple files are repeated

Instead of file-based sampling, what we really want is separate spatial and temporal sampling. Not all datasets have space or time, and for datasets with both, there are many possible combinations of sampling strategies one might want to perform.

### Design

The following [UML diagram](https://mermaid.js.org/syntax/classDiagram.html) illustrates the base class hierarchy:
```mermaid
classDiagram
  GeoSampler <|-- SpatialSampler
  GeoSampler <|-- TemporalSampler
  SpatialSampler <|.. SpatioTemporalSampler
  GeoSampler <|-- SpatioTemporalSampler
  TemporalSampler <|.. SpatioTemporalSampler
  class GeoSampler {
    -int _length
    +\_\_iter__() Iterator~GeoSlice~
    +\_\_len__() int
  }
  class SpatialSampler {
    +Literal~random, sequential~ strategy
    +\_\_init__(GeoDataset dataset, Polygon | None roi)
    +\_\_matmul__(TemporalSampler other) SpatioTemporalSampler
    +plot() FuncAnimation
  }
  class TemporalSampler {
    +Literal~random, sequential~ strategy
    +\_\_init__(GeoDataset dataset, Interval | None toi)
    +plot() FuncAnimation
    -_init_subset(GeoSlice location) IntervalIndex
    -_iter_subset(GeoSlice location) Iterator~GeoSlice~
  }
  class SpatioTemporalSampler {
    +\_\_init__(SpatialSampler sampler1, TemporalSampler sampler2)
    +plot() FuncAnimation
  }
```
Each `SpatialSampler` handles spatial sampling independently of time. Similarly, each `TemporalSampler` handles temporal sampling independent of (or optionally dependent on) location. These samplers can be combined using the following syntax:
```python
spatial_sampler = GriddedPatchSampler(dataset, size=224, stride=112)
temporal_sampler = SequentialPeriodSampler(dataset, freq='Y')
spatiotemporal_sampler = spatial_sampler @ temporal_sampler
```
The resulting sampler will then iterate over both the spatial and temporal samplers to combine these sampling strategies. The flexibility of this design allows each sampler subclass to be fairly simple and easy to write.

#### Spatial samplers

RandomPatchSamplerGriddedPatchSampler

#### Temporal samplers

RandomTimestampSamplerSequentialTimestampSampler
RandomTimedeltaSamplerSequentialTimedeltaSampler
RandomPeriodSamplerSequentialPeriodSampler

### Performance

All samplers are generally significantly faster at initialization time because we no longer pre-compute the length. The following table benchmarks the number of seconds it takes to iterate over our I/O Bench dataset:

| state | raw (random) | raw (grid) | preprocessed (random) | preprocessed (grid) |
| --- | --- | --- | --- | --- |
| v0.7 | 11.799 | 7.8161 | 6.8362 | 3.1881 |
| main | **broken** | **broken** | 8.3264 | 4.8549 |
| PR | 8.5464 | 5.2481 | 8.017 | 5.2172 |

The plan is to deprecate and eventually remove all of TorchGeo's previous file-based samplers. These samplers have a number of flaws and can mostly be captured by the new and improved samplers. However, note that there will not be a drop-in replacement for these existing samplers. In particular, there are currently no plans to replace BatchGeoSampler with an equivalent. If your use case requires control over batch sampling, please let us know.

### Progress

- base.py
  - [x] GeoSampler
  - [x] SpatialSampler
  - [x] TemporalSampler
  - [x] SpatioTemporalSampler
- spatial.py
  - A single patch/chip
    - [x] RandomPatchSampler
    - [x] GriddedPatchSampler
    - [ ] WeightedPatchSampler?
  - A single point
    - [ ] RandomPointSampler?
    - [ ] SequentialPointSampler?
    - [ ] WeightedPointSampler?
  - etc.
- temporal.py
  - A single image ([`pd.Timestamp`](https://pandas.pydata.org/docs/reference/api/pandas.Timestamp.html))
    - [x] RandomTimestampSampler
    - [x] SequentialTimestampSampler
  - A fixed-length sliding window ([`pd.Timedelta`](https://pandas.pydata.org/docs/reference/api/pandas.Timedelta.html))
    - [x] RandomTimedeltaSampler
    - [x] SequentialTimedeltaSampler
  - A fixed-length fixed window ([`pd.Period`](https://pandas.pydata.org/docs/reference/api/pandas.Period.html))
    - [x] RandomPeriodSampler
    - [x] SequentialPeriodSampler
  - Rolling window sampler
    - [ ] Something similar to sklearn's TimeSeriesSplit?
  - First image before/after a specific date
    - [ ] RandomChangeSampler?
    - [ ] FixedChangeSampler?
    - [ ] SequentialChangeSampler?
  - etc.
- single.py, batch.py, utils.py
  - [x] Deprecate all existing file-based samplers
- datamodules
  - [x] Update all existing GeoDataModules
- tests
  - [x] base
  - [x] spatial
  - [x] temporal
- documentation
  - [x] API docs
  - [x] tutorials

## AI Usage Disclosure

Closes #3548 (deprecates all batch samplers) @MathiasBaumgartinger
Closes #3033 (supersedes) @sfalkena 
Closes #1630, #1624 (small patches included in GriddedPatchSampler) @mpelchat04


torchgeo/samplers/spatial.py
```diff
@@ -0,0 +1,190 @@
...
+
+        Yields:
+            [xmin:xmax, ymin:ymax] coordinates to index a dataset.
+        """
+        # Ensure a new set of random points for each epoch
+        series = GeoSeries([self.geometry])
+        points = series.sample_points(size=self.length)
+
+        for point in points:
+            # TODO: snap to pixel grid? How? Can use outer geometry, but not file-specific, users will have to use TAP more
```
Moritz-Langer: I think for users to have to TAP might be a pitfall.
Since dataset is of type GeoDataset, the coordinates, crs and resolution should be known. Therefore it should be possible to snap to pixel origin coordinates. I think this makes it more intuitiv to use and less error prone for downstream data products.
However for the Random Sampler this might also be a non issue.

torchgeo/samplers/spatial.py
```diff
@@ -0,0 +1,192 @@
...
+        self.stride = _to_tuple(stride or size)
+
+        # Convert from pixel units to CRS units
+        if units == Units.PIXELS:
+            self.size = (self.size[0] * dataset.res[1], self.size[1] * dataset.res[0])
+            self.stride = (
+                self.stride[0] * dataset.res[1],
+                self.stride[1] * dataset.res[0],
+            )
+
```
Moritz-Langer, line 180: See https://github.com/torchgeo/torchgeo/compare/main...Moritz-Langer:torchgeo:feature/roi_fixed for __init__ that precalculates the valid chips and stores them so the __len__ can be known.

torchgeo/samplers/spatial.py
adamjstewart: CRS is known, but we need the transform, which is not known until sampling time. We could also store that I suppose. But it is currently removed during `union_all`, so no easy way to access.

torchgeo/samplers/spatial.py
adamjstewart: Yeah, we could just move all loops into `__init__` and store the bounds so we can calculate `__len__`. But I'm not sure how to get `__len__` for SpatioTemporalSampler because the length of the temporal sampler depends on the spatial location.

Moritz-Langer: Ah true, I neglected the temporal dimension. I havent used that before.
Only taking the spatial dimension into account I ran a test for the `GridSpatialSampler` using a real world example and it works just as I would expect it to, producing the same amount of chips. 
Could the loops for calculating the number of chips / length iterable also be calculated when using the `__len__` function/property? I thinkt in most cases it wouldnt be to expensive and it would improve backwards compatibility, as the len of dataloader could be used for the iteration over the dataloader. This is only a minor thing. If the cleaner code outweighs this benefit its also only `for batch_idx, batch in enumerate(dataloader_new):` but then maybe communicate the change in the release notes.

adamjstewart: Also a good idea. I definitely only want the code to exist in one place, but a helper function could be reused in `__len__` and `__getitem__` if we want to avoid memory limits. I'm leaning more and more towards making `__len__` required for all `GeoSamplers` just like the old design. At the very least, people will want to be able to report the epoch length during training.

adamjstewart: One problem: temporal length depends on spatial location unless using aggregated mosaics. For `RandomSpatialSampler`, since the spatial locations are random each time, the temporal length will be different each time. This means that unless users manually specify a `length` for `SpatioTemporalSampler`, they will get a different epoch length each time. Or we set the default length to be the first location sampled. Or we document this limitation carefully. All ideas worth considering.

adamjstewart:
Still working on fixing two last checks, but this should finally be ready for review!

torchgeo/datamodules/sentinel2_cdl.py
adamjstewart, line 1: @yichiac many of these GeoDataModules should probably be converted to time series versions (add `time_series=True` and use spatiotemporal samplers). But let's do that after the paper deadline, no rush.

torchgeo/samplers/base.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def __iter__(self) -> Iterator[tuple[slice, slice, slice]]:
+        """Iterate over generated sample locations for each epoch.
+
+        Yields:
+            [:, :, tmin:tmax] coordinates to index a dataset.
+        """
+        yield from self._iter_subset()
+
+    def _init_subset(
```
adamjstewart, line 143: I hate these helper function names, but at least they're private, so could change them later. Actually, should they be public so it's easier to write your own custom TemporalSampler that is compatible with SpatioTemporalSampler?

torchgeo/samplers/base.py
```diff
@@ -0,0 +1,226 @@
...
+        self, location: tuple[slice, slice] = (slice(None), slice(None))
+    ) -> IntervalIndex:
+        """Narrow down index to a specific location.
+
+        Args:
+            location: A specific location.
```
sfalkena, line 149: I think it would be better to explicitly rename this to `spatial_location`, so that no ambiguity between spatial and temporal location exists.

torchgeo/samplers/temporal.py
```diff
@@ -0,0 +1,359 @@
...
+        length = math.ceil((right - left - self.delta) / self.stride) + 1
+
+        x, y = location
+        for _ in range(length):
+            # TODO: ensure this doesn't escape our TOI
+            interval = Interval(left, left + self.delta)
+            if intervals.overlaps(interval).any():
+                t = slice(interval.left, interval.right)
+                yield x, y, t
+            left += self.delta
```
sfalkena, line 231: We should advance with the stride
```suggestion
            left += self.stride
```

torchgeo/samplers/temporal.py
```diff
@@ -0,0 +1,359 @@
...
+        *,
+        delta: Timedelta,
+        stride: Timedelta | None = None,
+        toi: Interval | None = None,
+    ) -> None:
+        """Initialize a new SequentialTimedeltaSampler instance.
+
+        Args:
+            dataset: Dataset to sample from.
+            delta: Duration of time.
```
sfalkena, line 196: From the first read it wasnt really clear what `delta` meant. I understand that you named it delta as you follow the `timeDelta` naming convention, but wouldn't it be more intuitive to call it `duration` or `window`? Also, I think it would be helpful for users to have some indication on the impact of this parameter in the docstring: "Depending on the parameter `time_series`, all aqcuisitions within this window will be merged into a mosaic, or returned as timeseries imagery".

torchgeo/samplers/temporal.py
```diff
@@ -0,0 +1,359 @@
...
+        Yields:
+            [:, :, tmin:tmax] coordinates to index a dataset.
+        """
+        intervals = self._init_subset(location)
+
+        left = intervals.left.min()
+        right = intervals.right.max() - self.delta
+
+        # TODO: make tile_to_chips more generic, support 1D inputs
+        length = math.ceil((right - left - self.delta) / self.stride) + 1
```
sfalkena, line 222: You are double subtracting `self.delta`, as `right` already has subtracted `self.delta`

torchgeo/samplers/base.py
```diff
@@ -0,0 +1,226 @@
...
+        """Iterate over generated sample locations for each epoch.
+
+        Yields:
+            [xmin:xmax, ymin:ymax, tmin:tmax] coordinates to index a dataset.
+        """
+        match self.spatial_sampler.strategy, self.temporal_sampler.strategy:
+            case 'random', 'random':
+                for _ in range(len(self.spatial_sampler) * len(self.temporal_sampler)):
+                    location = next(iter(self.spatial_sampler))
+                    yield next(iter(self.temporal_sampler._iter_subset(location)))
```
sfalkena, line 213: `iter(self.spatial_sampler)` calls `__iter__` which, for `RandomSpatialSampler`, calls `GeoSeries.sample_points(size=self.length)` generating ALL points but then discards all but the first. This runs N × M times per epoch, making the whole iteration O(N²M) in point generation. The sequential, random case has the same issue for the temporal side. Perhaps for all the random cases we could sample N points and then introduce a `shuffle()` method which is called after N `__iter__` calls?

torchgeo/samplers/base.py
```diff
@@ -0,0 +1,226 @@
...
+            case 'sequential', 'sequential':
+                for location in iter(self.spatial_sampler):
```
sfalkena, line 215: `iter` is not strictly needed here and for the other cases of sequential sampling.

torchgeo/samplers/base.py
```diff
@@ -0,0 +1,226 @@
...
+
+        Args:
+            location: Region of interest to sample from.
+
+        Yields:
+            [:, :, tmin:tmax] coordinates to index a dataset.
+        """
+
+
+class SpatioTemporalSampler(GeoSampler):
```
sfalkena, line 179: `SpatioTemporalSampler` has no length attribute, so `__len__` falls back to iterating the sampler to completion.

This has two concrete consequences:

PRNG state is consumed by `len()`. In the random × random case, `__iter__` drives `spatial_sampler.generator`. Any call to `len(sampler)` from `DataLoader.__init__`, Lightning's setup, a progress bar, or user code, advances the generator by a full epoch's worth of draws. The samples produced by the subsequent actual iteration are therefore shifted, so a fixed seed does not guarantee reproducibility if `len()` is called a different number of times before iteration begins.

Length is non-deterministic across processes. Because the length is computed by running the random iterator, two workers with different generator states can report different lengths, breaking DataLoader's distributed accounting.

torchgeo/samplers/base.py
```diff
@@ -0,0 +1,226 @@
...
+        Args:
+            location: A specific location.
+
+        Returns:
+            A subset of *self.index* at *location*.
+        """
+        index = self.index
+
+        # TODO: ensure we aren't modifying dataset.index too, may need to deepcopy
+        if location:
```
sfalkena, line 157: This always evaluates to `True`  through its default values as `bool(slice(None)) = True`

torchgeo/samplers/base.py
```diff
@@ -0,0 +1,226 @@
...
+            dataset: Dataset to sample from.
+            roi: Region of interest to sample from
+                (defaults to the bounds of ``dataset.index``).
+        """
+        # Create one single MultiPolygon of all objects
+        # Allows all locations to be equally weighted, regardless of # time stamps
+        # TODO: ensure we aren't modifying dataset.index too, may need to deepcopy
+        self.geometry = dataset.index.geometry.union_all()
+
+        if roi:
```
sfalkena, line 85: Shapely deprecated geometry truthiness in 2.x (ShapelyDeprecationWarning), and an empty Polygon() passed as ROI would silently skip the intersection.

```suggestion
        if roi is not None:
```

torchgeo/samplers/base.py
```diff
@@ -0,0 +1,226 @@
...
+        """Initialize a new TemporalSampler instance.
+
+        Args:
+            dataset: Dataset to sample from.
+            toi: Time of interest to sample from
+                (defaults to the bounds of ``dataset.index``).
+        """
+        self.index = dataset.index
+
+        if toi:
```
sfalkena, line 124:
```suggestion
        if toi is not None:
```

torchgeo/samplers/base.py
```diff
@@ -0,0 +1,226 @@
...
+    ) -> IntervalIndex:
+        """Narrow down index to a specific location.
+
+        Args:
+            location: A specific location.
+
+        Returns:
+            A subset of *self.index* at *location*.
+        """
+        index = self.index
```
sfalkena, line 154: I think you indeed need to copy

torchgeo/samplers/temporal.py
```diff
@@ -0,0 +1,359 @@
...
+            [:, :, tmin:tmax] coordinates to index a dataset.
+        """
+        intervals = self._init_subset(location)
+
+        left = intervals.left.min()
+        right = intervals.right.max() - self.delta
+
+        i = 0
+        x, y = location
+        while i < self.length:
```
sfalkena, line 165: For all `while` loops we could introduce an upper bound on the number of trials which triggers an error:

```
raise RuntimeError(
    f'Could not sample a valid interval after {max_attempts} attempts. '
    'The dataset may have insufficient temporal coverage for the requested delta/toi.'
)
```

torchgeo/samplers/temporal.py
```diff
@@ -0,0 +1,359 @@
...
+            generator: Pseudo-random number generator (PRNG).
+        """
+        super().__init__(dataset, toi=toi)
+        self.freq = freq
+        # TODO: should these be moved to _iter_subset? Length will change each epoch
+        left = self.index.index.left.min()
+        right = self.index.index.right.max()
+        period = Period(left, freq=freq)
+        self.length = length or (right - left) // (period.end_time - period.start_time)
+        self.generator = np.random.default_rng(generator)
```
sfalkena, line 276: This is incorrect behavior. The `generator` argument is set to be one of 4 different generators, or an `int`. In here we only support the default rng, which would raise an error when the actual generators are passed. I am not sure which behavior you were targetting, but I think you probably want the default_rng as default parameter to the argument.

sfalkena: agrifieldnet assumes that the generator support an `int` in the current implementation.

torchgeo/samplers/temporal.py
```diff
@@ -0,0 +1,359 @@
...
+
+    .. versionadded:: 0.10
+    """
+
+    strategy = 'sequential'
+
+    def __init__(
+        self, dataset: GeoDataset, *, freq: str, toi: Interval | None = None
+    ) -> None:
+        """Initialize a new SequentialTimedeltaSampler instance.
```
sfalkena, line 323: Docstring inconsistancy

torchgeo/samplers/spatial.py
```diff
@@ -0,0 +1,192 @@
...
+        # Default to approximate number of non-overlapping patches
+        patch_area = self.size[0] * self.size[1]
+        self.length = length or int(shapely.area(self.geometry) // patch_area)
+
+        # Erosion to avoid out-of-bounds sampling
+        # Purposefully conservative radius calculation
+        # TODO: this operation removes Point and LineString, should we keep these?
+        distance = math.sqrt((self.size[0] / 2) ** 2 + (self.size[1] / 2) ** 2)
+        self.geometry = shapely.buffer(self.geometry, -distance)
```
sfalkena, line 86: Should we do a check after this to ensure that the geometry is still valid and not empty?

torchgeo/samplers/spatial.py
```diff
@@ -0,0 +1,192 @@
...
+        # Convert from pixel units to CRS units
+        if units == Units.PIXELS:
+            self.size = (self.size[0] * dataset.res[1], self.size[1] * dataset.res[0])
+
+        # Default to approximate number of non-overlapping patches
+        patch_area = self.size[0] * self.size[1]
+        self.length = length or int(shapely.area(self.geometry) // patch_area)
+
+        # Erosion to avoid out-of-bounds sampling
+        # Purposefully conservative radius calculation
```
sfalkena, line 83: What is the reason for the conserviatism? I assume to not sample any `nodata` pixels? However, can we actually guarantee this? In general, what behavior do we want when we have sampled a spatial location on the border of 2 sentinel tiles, and due to temporal sampling, we can only return pixels from 1 of them. This would also cause `nodata` pixels. Earlier in my custom implementation, I introduced a `drop_nodata` parameter, which would ignore the entire patch so that I dont get jibberish predictions with my model.

sfalkena (changes requested):
I love that we are able to use SpatialSampler and TemporalSampler independently through combination with `_disambiguate_slice`!

I would still be interested to see the option to precompute the spatial bboxes, but I will open a PR for that after this is merged in, since the current version of this PR is in line with the old samplers that generate samples on the fly.

I think something worth discussing is if you think that every sampler should have its length computed or not. In the current implementation `SpatioTemporalSampler`, `SequentialTimedeltaSampler` and `SequentialPeriodSampler` have no length attribute. You know my opinion: I would advocate for computing samples up front and make length known before starting the sampling.

I will take a look at datamodules/docs/tests in the next round.

torchgeo/samplers/base.py
```diff
@@ -0,0 +1,226 @@
...
+
+        Args:
+            dataset: Dataset to sample from.
+            roi: Region of interest to sample from
+                (defaults to the bounds of ``dataset.index``).
+        """
+        # Create one single MultiPolygon of all objects
+        # Allows all locations to be equally weighted, regardless of # time stamps
+        # TODO: ensure we aren't modifying dataset.index too, may need to deepcopy
+        self.geometry = dataset.index.geometry.union_all()
```
adriantre, line 83: `union_all()` should return a new Polygon object, so no deepcopy should be needed here.

torchgeo/samplers/base.py
```diff
@@ -0,0 +1,226 @@
...
+        """
+        self.index = dataset.index
+
+        if toi:
+            tmin = np.maximum(toi.left.to_datetime64(), self.index.index.left)
+            tmax = np.minimum(toi.right.to_datetime64(), self.index.index.right)
+            valid = tmax >= tmin
+            tmin = tmin[valid]
+            tmax = tmax[valid]
+            self.index = self.index[valid]
```
adriantre, line 130: This, too, does not alter the original `dataset.index` since it is a boolean index that returns a new object. But if anyone adds e.g. `self.index.drop(..., inplace=True)` then the original `dataset.index` will be changed. In this case we might want a shallow copy

torchgeo/samplers/spatial.py
```diff
@@ -0,0 +1,192 @@
...
+        Yields:
+            [xmin:xmax, ymin:ymax] coordinates to index a dataset.
+        """
+        # Ensure a new set of random points for each epoch
+        series = GeoSeries([self.geometry])
+        points = series.sample_points(size=self.length, rng=self.generator).explode()
+
+        for point in points:
+            # TODO: snap to pixel grid? How? Can use outer geometry, but not file-specific, users will have to use TAP more
+            # IDEA: snap to pixel grid in RasterDataset, not here. Or maybe that would be incredibly confusing...
```
adriantre, line 100: I like this idea. Return points (in wgs84?) from samplers and construct box in `__getitem__` when CRS and transform is decided. Enables minimal re-projection.

adriantre: The method of creating the boxes can possibly live in the samplers and be called with crs and transform in `__getitem__`, if that makes it tidy?

torchgeo/samplers/spatial.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def __iter__(self) -> Iterator[tuple[slice, slice]]:
+        """Iterate over generated sample locations for each epoch.
+
+        Yields:
+            [xmin:xmax, ymin:ymax] coordinates to index a dataset.
+        """
+        # Ensure a new set of random points for each epoch
+        series = GeoSeries([self.geometry])
+        points = series.sample_points(size=self.length, rng=self.generator).explode()
```
adriantre, line 96: I suggest putting this in its own method, allowing users to override point generation, e.g. for weighting.

torchgeo/samplers/spatial.py
darkblue-b: posit:  an abstraction that holds the CRS and transform is the right place to query for those values, not lower level data parts. Consider returning crs and transform in one way, at the highest level of abstraction that applies to the data structures.. function clients that ask for CRS and transform must know that, so they could reasonably be expected to not make a query to individual elements.. maybe that could be a preference, but I'll argue that the converse is not just a preference.. if you leave the ambiguity that individual elements can have their own CRS and transform, then unrelated parts of the sfwr system might ask about them or behave conditionally.. definitely not desirable..

torchgeo/samplers/__init__.py
darkblue-b, line 1: this file shows that the old `GeoSampler` will remain while the new two-part samplers are made available? so a "deprecated" warning or similar ?

adamjstewart: GeoSampler is still the name of the top-level base class, see the UML diagram in the PR description.

torchgeo/samplers/base.py
adamjstewart: I think "location" is clearly spatial, but I'm thinking of using `roi` to match our samplers instead.

torchgeo/samplers/base.py
adamjstewart: I don't think so? `index.cx[]` should return a copy or subset, it shouldn't modify the original.

torchgeo/samplers/base.py
adamjstewart: > `SpatioTemporalSampler` has no length attribute, so `__len__` falls back to iterating the sampler to completion.

In the case of random/random, this is easy to fix. In all other cases, iteration is indeed required. Perhaps the length attribute should be tied to the random strategy? This would require `RandomMixin` and `SequentialMixin` classes, but is otherwise easy to do.

> PRNG state is consumed by `len()`... so a fixed seed does not guarantee reproducibility if `len()` is called a different number of times before iteration begins.

Yep. This is unfortunate, but I could probably live with this. The important thing is that a static code snippet will always result in the same samples (given a seed) unless modified.

> Any call to len(sampler) from `DataLoader.__init__`, ...

Double checked the source code. `DataLoader.__init__` does not call `len(sampler)`, although `DataLoader.__len__` does.

> Lightning's setup, a progress bar, ...

Confirmed that even with `--trainer.enable_progress_bar=false`, something somewhere is calling sampler/dataloader length on every epoch. So at the very least, we are currently doubling the sampler iteration time with my PR. I can help by caching length of sequential samplers after the first iteration, but we need to iterate at least once to calculate length regardless. We _could_ cache the sampled locations, which is the feature you want, but this probably results in high memory usage.

> Length is non-deterministic across processes. Because the length is computed by running the random iterator, two workers with different generator states can report different lengths, breaking DataLoader's distributed accounting.

To clarify, length is only non-deterministic in the case of random/sequential, it is deterministic in all other combinations: random/random, sequential/sequential, sequential/random, sequential, and random. But yes, this will likely break a lot of stuff. We could explicitly forbid random/sequential (and possibly sequential/random) if we are concerned.

Let me keep hacking on this, but I agree this is currently the biggest pitfall of the current implementation. I think clever caching will help but not entirely resolve our concerns.

adamjstewart: Update: our `BaseDataModule` class is also calling `if sampler:`. I think truthiness of an iterator is determined by length != 0, so that may be another place that hits this.

torchgeo/samplers/spatial.py
adamjstewart: > What is the reason for the conserviatism? I assume to not sample any `nodata` pixels?

Yes.

> However, can we actually guarantee this? In general, what behavior do we want when we have sampled a spatial location on the border of 2 sentinel tiles, and due to temporal sampling, we can only return pixels from 1 of them. This would also cause nodata pixels.

We can guarantee it if we only have spatial samplers, but we can't guarantee it with all combinations of spatial and temporal samplers.

My basic guiding design philosophy so far has been:

* random samplers: never sample out of bounds
* sequential samplers: sample out of bounds if needed to get full coverage

> Earlier in my custom implementation, I introduced a `drop_nodata` parameter, which would ignore the entire patch so that I dont get jibberish predictions with my model.

Does this rely on loading the data with rasterio first and checking for nodata pixels? We tried this one time and it didn't work with multiprocessing. Perhaps rasterio has fixed this, but I'm worried about the slowdown of doubling I/O times.

torchgeo/samplers/spatial.py
adamjstewart: We can do that in a follow-up PR. My main sticking point for weighted sampling is I don't have a good idea for the API. It feels like most of our users are afraid of subclassing and are more likely to instead copy-n-paste the entire class. I would rather create a new WeightedSpatialSampler class with a stable API and well-defined input. But I don't know what that input could be.

torchgeo/samplers/spatial.py
adamjstewart: The problem with doing this in the dataset is that there may be situations where both image and mask are in the same CRS but with different res or with same res but half a pixel shift. We don't want two datasets to independently snap to their own grids. So I think we really do need to do this in the sampler somehow.

torchgeo/samplers/spatial.py
adamjstewart: I wrote a `__len__` base class implementation that checks for the presence of `self.length` or brute force iterates. So now all samplers have a length. I think we still need to cache this though so as to minimize iteration.

torchgeo/samplers/temporal.py
adamjstewart: We could. I'm not super worried, but maybe I should be?

torchgeo/samplers/temporal.py
adamjstewart: Fixed.

torchgeo/samplers/temporal.py
adamjstewart: I can't remember why I used `default_rng` here. I do remember why I allowed several possible inputs (I wish numpy was as simple as torch). I think we can change this or do whatever we want. I also wonder if we should default to some seed instead of None...

torchgeo/samplers/spatial.py
adriantre: Interesting, I like to subclass. But often I end up copy-n-pasting the whole init-method etc, which is what I would have to do here, instead of something like

```python
def MyWeightedRS(RandomSpatialSampler):
    def sampled_points(self):
        return myhand_picked_points(self,length)
```

I feel like this is a great way for people to explore weightning, to later make suggestions towards an api for TorchGeo.

torchgeo/samplers/spatial.py
adriantre: I lean towards always snapping to the first dataset, and require this to be the raster.
```python
dataset = landsat & cdl  # will snap overlapping cdl to each landsat rasters crs
```

torchgeo/samplers/spatial.py
adamjstewart: If it's just a list of hand-picked points, then we can easily support that via an `__init__` parameter. I'm not sure how useful that is though.

I think you like subclassing because you are coming from a CS background? A lot of users from Geo backgrounds have no idea what OOP is and are more used to the numpy API.

torchgeo/samplers/spatial.py
adamjstewart: Oof, I didn't think about whether or not the first dataset was raster, that's yet another complication.

Since we merge all geometries into one, we can easily check if we are snapped to the overall grid, but if each file differs by half a pixel, we would have to store and re-query the raw unmerged geometries to snap to each file grid. Not impossible, just painful.

I vote we snap to the overall grid for now (easy) and revisit this later. I have a student working on ways of not requiring all rasters to be reprojected, so she can take a look at this at the same time.

adriantre: Sounds good. And good to hear that avoiding reprojection is worked on. Especially for global datasets, the reprojection is to be avoided both for training speed and data quality / non-distortion, but you know that already :)

calebrob6 (changes requested):
Can you create a script that downloads a time series of Sentinel 2 data (can/should be much smaller than a full scene-- e.g. you can do a bunch of 512 x 512 patches over an area you know is interesting), creates a Sentinel 2 dataset, and runs these samplers on that dataset and visualizes the resulting patches? This will let me really understand what is going on and check correctness (I know there are unit tests!)

adamjstewart:
> Can you create a script that...

I think @isaaccorley is actually working on this right now?

I'm currently lost in the details of performance. Basically, I _suspect_ that things are correct and _know_ that things are slower than they can be, so trying to fix the latter before finalizing the former. 

We definitely need to do both before merging though.

torchgeo/samplers/utils.py
```diff
@@ -120,3 +124,38 @@ def tile_to_chips(
...
+        kernel_size: Size of each output window.
+        stride: Stride with which to sample (defaults to *input_size*).
+
+    Returns:
+        The number of output windows that can be sampled.
+
+    .. versionadded:: 0.10
+    """
+    stride = stride or kernel_size
+    return math.ceil((input_size - kernel_size) / stride) + 1  # ty: ignore[no-matching-overload]
```
adamjstewart, line 163: Found a bug in ty: https://github.com/astral-sh/ty/issues/3323

adamjstewart:
@calebrob6 I created a visualization for our spatial samplers (not yet for temporal or spatiotemporal). If you download our I/O Bench dataset and run:
```python
import einops
import matplotlib
import numpy as np
from matplotlib.patches import Rectangle

from torchgeo.datasets import Landsat9
from torchgeo.datasets.utils import quantile_normalization
from torchgeo.samplers import GridSpatialSampler, RandomSpatialSampler

dataset = Landsat9('data/io/preprocessed/landsat', res=300)
random_sampler = RandomSpatialSampler(dataset, size=100)
grid_sampler = GridSpatialSampler(dataset, size=100, stride=90)

fig, ax = plt.subplots(1, 2)
cmap = matplotlib.colormaps['viridis']

image = dataset[dataset.bounds]['image']
image = image[[3, 2, 1]]
image = einops.rearrange(image, 'c h w -> h w c')
image = quantile_normalization(image, 0.35, 0.98)

x, y, t = dataset.bounds
extent = (x.start, x.stop, y.start, y.stop)


def plot_samples(ax, sampler):
    ax.imshow(image, extent=extent)
    colors = cmap(np.linspace(0, 1, len(sampler)))
    for i, (x, y) in enumerate(sampler):
        xy = (x.start, y.start)
        width = x.stop - x.start
        height = y.stop - y.start
        color = colors[i]
        patch = Rectangle(xy, width, height, edgecolor=color, facecolor='none')
        ax.add_patch(patch)

    return ax


ax[0].set_title('RandomSpatialSampler')
ax[0] = plot_samples(ax[0], random_sampler)

ax[1].set_title('GridSpatialSampler')
ax[1] = plot_samples(ax[1], grid_sampler)

fig = fig.tight_layout()
plt.show()
```
you'll see:
samplers

This actually revealed a bug in geopandas! `geopandas.GeoSeries.sample_points(100, rng=0)` will generate 100 random samples, but they are sorted by x-value: https://github.com/geopandas/geopandas/issues/3772.

Update: shuffled all points and things look good now, added to the ever-growing PR description.

torchgeo/samplers/spatial.py
adamjstewart: Implemented a basic snap-to-grid feature. Seems to be quite fast, even without preprocessing!

torchgeo/samplers/temporal.py
adamjstewart: > In here we only support the default rng, which would raise an error when the actual generators are passed.

`default_rng` supports all of these inputs: https://numpy.org/devdocs/reference/random/generator.html. Existing generators are simply passed through or coerced into the same type. The idea is that regardless of what kind of seed/generator is passed, the underlying code remains the same.

torchgeo/samplers/spatial.py
adamjstewart: Got the snap to grid working, things are surprisingly much faster than expected!

torchgeo/samplers/spatial.py
adamjstewart: Caching added.

torchgeo/samplers/base.py
adamjstewart: Good catch! Moved `iter()` outside the loop so it is only called once per epoch. Confirmed via warning that it is indeed called once and exactly once per epoch. This made things an order of magnitude faster.

torchgeo/samplers/base.py
adamjstewart: Removed

torchgeo/samplers/base.py
adamjstewart: Cached.

docs/_static/animations/GridSpatialSampler.gif
adriantre, line 1: This is so nice

adamjstewart: Just wait until you see the 3D animations I'm working on for spatiotemporal samplers 🙂

adriantre: Love it. My mind went to the 3D-conv animations from [this article](https://medium.com/apache-mxnet/1d-3d-convolutions-explained-with-ms-excel-5f88c0f35941) that I remember from a few years ago
3D-conv

adamjstewart: These are way better: https://github.com/vdumoulin/conv_arithmetic

adamjstewart:
GriddedPatchSampler

I'm not 100% sure this is right. For example, the very first grid cell does indeed intersect with the polygon, but the one after it covers the exact same parts of the polygon, so the first feels redundant. There may be situations where it is nice to include anyway (e.g., patch intersection would be near edge without it), but this is also true if we made our overall grid larger as well. Not sure if we should 1) add more grid cells so that we never sample near the edge, or 2) remove more grid cells so we never compute redundant information.

EDIT: I changed this from `intersects` to `intersects and not touches` and the very first grid cell is now removed. It's still not 100% what I want, but it's much better now.

torchgeo/samplers/spatial.py
```diff
@@ -0,0 +1,210 @@
...
+            self.size = (self.size[0] * dataset.res[1], self.size[1] * dataset.res[0])
+
+        # Default to approximate number of non-overlapping patches
+        total_area = shapely.area(self.geometry)
+        patch_area = self.size[0] * self.size[1]
+        self.length = length or convolution_arithmetic(total_area, patch_area)
+
+        # Erosion to avoid out-of-bounds sampling
+        # Purposefully conservative radius calculation
+        # TODO: this operation removes Point and LineString, should we keep these?
```
adriantre: I do think we should keep these. If you are training e.g. road detection you might want to pass roi as a linestring.

torchgeo/samplers/spatial.py
```diff
@@ -0,0 +1,210 @@
...
+            [xmin:xmax, ymin:ymax] coordinates to index a dataset.
+        """
+        xmin, ymin, xmax, ymax = self.geometry.bounds
+
+        # TODO: adjust xmin/ymin to have equal spacing in case of non-integer multiple
+        # xmid, ymid = self.geometry.centroid
+        # TODO: but ensure we remain snapped to the pixel grid...
+
+        rows = convolution_arithmetic(ymax - ymin, self.size[0], self.stride[0])
+        cols = convolution_arithmetic(xmax - xmin, self.size[1], self.stride[1])
```
adriantre, line 190: Regarding redundant cells that contain the exact same pixels; is it an option to compute the intersection for all cells outside the loop? Then you could pre-filter them using drop_duplicates(intersection) in geopandas.

adriantre (approved):

adriantre:
Sorry, did not mean to approve 🤠 stupid big thumbs on a small phone.

You are a beast for this!

calebrob6:
> Can you create a script that downloads a time series of Sentinel 2 data (can/should be much smaller than a full scene-- e.g. you can do a bunch of 512 x 512 patches over an area you know is interesting), creates a Sentinel 2 dataset, and runs these samplers on that dataset and visualizes the resulting patches? This will let me really understand what is going on and check correctness (I know there are unit tests!)

I did this if anyone wants it! Using this I was able to verify most of the functionality. The only weird thing I hit was how to get the time dimension working with the DataLoader (stack_samples doesn't work by default here). Is there any guidance for that?

Also the sampler plot methods are cool :)

image

adamjstewart:
> The only weird thing I hit was how to get the time dimension working with the DataLoader (stack_samples doesn't work by default here).

Can you elaborate on this? Are you using `time_series=True` with the dataset? `stack_samples` should not be necessary at all since #3138, PyTorch's built-in `default_collate` should suffice.

torchgeo/samplers/spatial.py
adamjstewart: Dug into this and the behavior is quite surprising. Basically, there are no errors anywhere, but nothing happens. `shapely.buffer()` will return an empty polygon, which `gpd.GeoSeries()` will happily accept, which `gpd.GeoSeries().sample_points()` will happily ignore. So the length of the epoch will become 0 even if the length of the sampler is non-zero.

adamjstewart: So:
```python
len(sampler) == 1
```
but
```python
for _ in sampler:
    pass
```
does 0 loops and:
```python
next(iter(sampler))
```
raises a StopIteration. Do you think it's worth raising an error instead? What error class?

torchgeo/samplers/spatial.py
adamjstewart: I think we'll want a separate sampler for this, could call it PointSpatialSampler.

torchgeo/samplers/base.py
adamjstewart: Let's keep this a private method for now, we can decide to change the name later.

torchgeo/samplers/spatial.py
adamjstewart: It's an option, although it quickly leads to OOM errors if you scale up to global.

adamjstewart:
Discovered another bug: https://github.com/matplotlib/matplotlib/issues/31665

torchgeo/samplers/spatial.py
adriantre: What about pre-computing and storing to e.g. a geopackage? Would the IO be slow?

In general, I see so many upsides with pre-computing the patches (for all samples). Dataset statistics, balancing etc

adamjstewart: * GridSpatialSampler: I don't think those upsides apply since you always want to sample the full region
* RandomSpatialSampler: you would have to re-compute each epoch anyway, so pre-computation is less advantageous

I would vote for the current implementation for now and then we can consider pre-computation in a follow-up PR. Let me also reach out to the geopandas devs to see if they have any clever ideas for weighted sampling.

adamjstewart: FWIW, @sfalkena has also been begging for pre-computation for many of the same reasons. And I've been resisting for many of the same reasons lol.

adamjstewart:
Now I'm second-guessing some of the naming conventions. For `SpatialSampler`, how do people feel about:

- spatial.py
  - A single patch
    - [x] RandomPatchSampler
    - [x] GriddedPatchSampler
    - [ ] WeightedPatchSampler
  - A single point
    - [ ] RandomPointSampler
    - [ ] SequentialPointSampler
    - [ ] WeightedPointSampler

Downside is we start to get into the patch vs. chip vs. tile naming debate.

calebrob6:
Have you tried this with a Sentinel2 dataset object that contains samples over time?

adamjstewart:
I think you did this above? @yichiac can you also try this since you need it for your crop type mapping tutorial?

docs/_static/animations/GridSpatialSampler.gif
adriantre: They're nice, but does not show 3D conv layers 😎

torchgeo/samplers/spatial.py
adriantre: Then I know who to team up with against you in the future 😏

calebrob6:
I know I did it, but have you done it? Argument 1: You've put so much effort into this so far, why not actually test it in a real-world use case? Argument 2: This is a huge change, you need to test with a real-world use case. I wouldn't dream of shipping a dataset/trainer/etc. without actually trying it out.

yichiac:
> I think you did this above? @yichiac can you also try this since you need it for your crop type mapping tutorial?

okay, testing on PASTIS. will report the sampler tests.

yichiac:
@adamjstewart PASTIS is a `NonGeoDataset`. `NonGeoDataset` is not covered by the current sampler PR as it's based on `GeoSampler`.

adamjstewart:
@yichiac yes, but our stretch goal was to train/test a model on PASTIS and then use the same model for inference on a new region. So you could test this to make sure it works for your inference.

adamjstewart:
> I know I did it, but have you done it?

Nope.

> Argument 1: You've put so much effort into this so far, why not actually test it in a real-world use case? 

I haven't tested this on real-world data _because_ I've already put so much effort into this. After 1.5 years of planning and 2 months of painstaking implementation, I'm no longer the one who needs to be convinced that the design or code are correct. I'm also barely keeping up with reviewer requests and don't have time to do even more experiments.

> Argument 2: This is a huge change, you need to test with a real-world use case. I wouldn't dream of shipping a dataset/trainer/etc. without actually trying it out.

This is the benefit of a team effort. We've already tested with both real data (my I/O Bench, your experiments) and synthetic data (my animations, my unit tests). I'm also hoping the time-series tutorial team can test this since they need it for their own tutorials, but haven't seen any progress there yet.

I'm happy to fix any bugs people spot, but I'm no longer capable of finding any more bugs, it's pretty bug-free (famous last words).

yichiac:
@adamjstewart 

> @yichiac yes, but our stretch goal was to train/test a model on PASTIS and then use the same model for inference on a new region. So you could test this to make sure it works for your inference.

Thanks for the clarification. Just to confirm, we want to train a model on PASTIS (`NonGeoDataset`) and perform inference on a new, raw, continuous S2 `GeoDataset` using the new sampler in this PR. Is this correct? If so, I will start preparing a random region for testing, maybe somewhere in France, as PASTIS only covers France.

adamjstewart:
That is correct, although I would specifically pick somewhere not France so it isn't an area in the training data.

yichiac:
> That is correct, although I would specifically pick somewhere not France so it isn't an area in the training data.

@adamjstewart sounds good. We can perform inference on Germany using the EuroCrops labels for evaluation. Meanwhile, let me test the raw S2 with the new sampler first.

yichiac:
I downloaded 12 Sentinel-2 L2A scenes over a ~7×7 km AOI in NRW, Germany, from 2019-04-08 to 2019-06-04. When `GriddedPatchSampler` @ `SequentialPeriodSampler('M')` runs over an unevenly-sampled S2 time series, each per-sample image tensor has a different T (e.g. 7 for April, 3 for May), so `DataLoader` with `stack_samples` raises 
`RuntimeError: stack expects each tensor to be equal size`. The root cause is that `stack_samples` uses `torch.stack`, which requires identical shapes, while Period/Timedelta samplers don't pin the number of timestamps per window.


```
from torchgeo.datasets.utils import stack_samples
from torchgeo.samplers import GriddedPatchSampler, SequentialPeriodSampler

PASTIS_BANDS = ['B02', 'B03', 'B04', 'B05', 'B06', 'B07', 'B08', 'B8A', 'B11', 'B12']
ds = Sentinel2(paths=tmp, bands=PASTIS_BANDS, res=10, time_series=True)

sp = GriddedPatchSampler(ds, size=128, stride=128)
tp = SequentialPeriodSampler(ds, freq='M')
comp = sp @ tp

for i, idx in enumerate(comp):
    if i >= 4:
        break
    img = ds[idx]['image']
    print(f'  sample {i}: T={img.shape[0]}, full shape={tuple(img.shape)}')

loader = DataLoader(ds, sampler=comp, batch_size=2, collate_fn=stack_samples)
try:
    batch = next(iter(loader))
    print(f'  unexpected success: {tuple(batch["image"].shape)}')
except RuntimeError as e:
    print(f'  RuntimeError: {e}')
```

yichiac:
I also tested other individual samplers: two spatial samplers and six temporal samplers can generate correct `len()` values.

adamjstewart:
Yep, you'll need to use the padding collation function you wrote.

calebrob6:
What's the expected behavior of?
```
spatial = RandomPatchSampler(ds, size=PATCH, length=4, generator=0)
temporal = SequentialTimestampSampler(ds)
st = spatial @ temporal
num_samples = len(st)
num_samples_2 = len(list(st))
assert num_samples == num_samples_2
```

calebrob6:
> Yep, you'll need to use the padding collation function you wrote.

What is this custom padding?

adamjstewart:
> What's the expected behavior of?
> 
> ```
> 
> spatial = RandomPatchSampler(ds, size=PATCH, length=4, generator=0)
> 
> temporal = SequentialTimestampSampler(ds)
> 
> st = spatial @ temporal
> 
> num_samples = len(st)
> 
> num_samples_2 = len(list(st))
> 
> assert num_samples == num_samples_2
> 
> ```
> 
> 

This should be the same, although some specific combinations have a different length from one epoch to the next. I haven't figured out a good rule of thumb for this. I'm considering disallowing random @ sequential for this reason.

calebrob6:
It wasn't the same in my test case, was 4 then 7 as the code that calculates the length dynamically consumes the iterator IIRC

adamjstewart:
They should be the same for random, sequential, random @ random, sequential @ sequential, and sequential @ random, but may be different for random @ sequential (your case above) depending on the input data. Do you think we should disallow random @ sequential since it can result in epochs of different length under certain situations? I'm not even sure how useful random @ sequential and sequential @ random are anyway.

calebrob6:
I'd strongly expect `num_samples == num_samples_2` to be the same in my example above (and any example -- calling `len(obj)` shouldn't change `obj` -- that's a Schrodinger's box type situation 🙂), but I don't really have a strong opinion about the design. I can post the data for reproducing if that's helpful.

adamjstewart:
Will discuss at today's time-series meeting. I'm leaning towards a warning but still allowing it.

calebrob6:
Thinking about this for a few more minutes -- my most frequent use case here would be a single S2 MGRS tile over time, and I'd strongly expect random @ sequential to work for this.

adamjstewart:
> Thinking about this for a few more minutes -- my most frequent use case here would be a single S2 MGRS tile over time, and I'd strongly expect random @ sequential to work for this.

Assuming all tiles are perfectly aligned (not shifted by a few pixels) then this _should_ result in the same number of samples per epoch. It's only when different random locations have different temporal length that the problem occurs.

torchgeo/samplers/spatial.py
```diff
@@ -0,0 +1,222 @@
...
+from .base import SpatialSampler
+from .constants import Units
+from .utils import _to_tuple, convolution_arithmetic
+
+
+class RandomPatchSampler(SpatialSampler):
+    """Sample locations from a region of interest randomly.
+
+    This is particularly useful during training when you want to maximize the size of
+    the dataset and return as many random :term:`chips <chip>` as possible. Note that
```
nilsleh, line 26: What is the difference between a patch and a chip? But we are returning chips but sampling patches?

torchgeo/samplers/spatial.py
```diff
@@ -0,0 +1,222 @@
...
+            self.size = (self.size[0] * dataset.res[1], self.size[1] * dataset.res[0])
+
+        # Default to approximate number of non-overlapping patches
+        total_area = shapely.area(self.geometry)
+        patch_area = self.size[0] * self.size[1]
+        self._length = length or convolution_arithmetic(total_area, patch_area)
+
+        # Erosion to avoid out-of-bounds sampling
+        # Purposefully conservative radius calculation
+        distance = math.sqrt((self.size[0] / 2) ** 2 + (self.size[1] / 2) ** 2)
```
nilsleh, line 86: What happens here if the requested patch size is close to the geometry? The length has already been set and self.size at this point has already been converted from pixels to CRS units and distance is the half-diagonal of the patch.. For a
  100×100 CRS-unit geometry with patch=100 CRS units, that's distance = sqrt(50² + 50²) ≈ 70.71. But then buffering that geometry 100×100 box inward by 70.71 produces an empty geometry. I guess it is an edge case but perhaps still relevant if you have a random global collection of tiles that are not huge tiles.

torchgeo/samplers/temporal.py
```diff
@@ -0,0 +1,353 @@
...
+            location: Region of interest to sample from.
+
+        Yields:
+            [:, :, tmin:tmax] coordinates to index a dataset.
+        """
+        intervals = self._init_subset(location)
+
+        left = intervals.left.min()
+        right = intervals.right.max()
+        length = convolution_arithmetic(right - left, self.delta)
```
nilsleh: do we also need to pass the stride here? Especially if stride < delta? Because stride is also use down below to compute left

torchgeo/datamodules/iobench.py
```diff
@@ -6,7 +6,7 @@
 from typing import Any
 
 from ..datasets import IOBench
-from ..samplers import GridGeoSampler, RandomGeoSampler
+from ..samplers import GriddedPatchSampler, RandomPatchSampler
```
nilsleh, line 9: Do we have an IO Bench for RandomBatchGeoSampler vs the new RandomPatchSampler since a bunch of datasets switched to this one and if I understood correctly:

For a Landsat dataset for example with 1000 scenes, if I specify Batch size 32 and Length 10000 patches per epoch.
  - RandomBatchGeoSampler: 10000 / 32 = 313 batches → 313 file opens per epoch.
  - RandomPatchSampler: 10000 patches → up to 10000 file opens per epoch.

So somehow a tradeoff between sample drawing efficiency and batch diversity, but perhaps good to know whether "too diverse" ends up being "too costly" or a bottleneck.

nilsleh (commented):
I have read some of the comments etc to try and understand as much as possible, but it is hard to review a PR this big. However, it looks very powerful!

I would generally agree that precomputing indices would be a very nice feature because it can help with filtering/visualization or general data exploration for setting up large training runs (for example this work https://arxiv.org/abs/2501.12535) as well as reproducibility. 

I think what could furthermore be helpful is to have a docs/tutorial page that explains the different use cases and combinations of samplers in both space and time for common applications.

torchgeo/samplers/spatial.py
adamjstewart: https://docs.torchgeo.org/en/stable/user/glossary.html#term-chip

torchgeo/samplers/spatial.py
adamjstewart: This actually isn't an edge case, it's a common problem. We are currently deprecating `PrechippedGeoSampler` and telling people to use `RandomPatchSampler` instead. However, if they try this, they will find that they immediately end up with an empty geometry. See the discussion in https://github.com/torchgeo/torchgeo/pull/3552#discussion_r3072991153 for what will happen (non-zero sampler length, but zero samples). Not really sure how to properly handle this honestly, thanks for bringing this up again.

torchgeo/datamodules/iobench.py
adamjstewart: We used to, see Figure 3a from https://dl.acm.org/doi/10.1145/3707459. Basically, there never was a substantial speed boost from RandomBatchGeoSampler, at least with typical settings and COGs. There are however major disadvantages, especially for SSL: #3548.

There is no fundamental reason we can't have batch versions of the samplers in this PR, although it would double the length of the PR. I just don't see a strong incentive. 

It's possible this changes with Zarr where users can take advantage of the structure of how the data was stored (by location, by time, or by variable), but there seem to be few agreed upon standards in Zarr at this point so it would be difficult to optimize anything ahead of time.

torchgeo/samplers/temporal.py
adamjstewart: Good catch, fixed!

adamjstewart:
> I would generally agree that precomputing indices would be a very nice feature because it can help with filtering/visualization or general data exploration for setting up large training runs (for example this work https://arxiv.org/abs/2501.12535) as well as reproducibility.

I agree that it would help with filtering, although I think there are other ways to do this. How would it help with visualization and reproducibility?

My main concern is that the memory requirements blow up, and all attempts to resolve this will result in the PR blowing up in size as well. Let's discuss in a follow-up PR and keep this PR simple(r).

adamjstewart:
> I think what could furthermore be helpful is to have a docs/tutorial page that explains the different use cases and combinations of samplers in both space and time for common applications.

Agreed. I was thinking about updating https://torchgeo--3552.org.readthedocs.build/en/3552/tutorials/torchgeo.html#Samplers to use time series data and show how to choose samplers for this specific case. We could also have a separate basic usage page just for samplers and how to select them and several examples where they would be helpful. I'm willing to write this, but before I do, are you sure you want this PR to be larger? 😅

calebrob6:
> Will discuss at today's time-series meeting. I'm leaning towards a warning but still allowing it.

What was the decision here?

adamjstewart:
We decided to add a warning but not disallow it.

robmarkcole:
Suggest a final review from @nilsleh or @calebrob6

calebrob6:
@robmarkcole @adamjstewart -- I've reviewed this as much as I can review a "+2,006, -179" PR. I implemented a bunch of tests with Sentinel-2 data and the only large problem I found is how the dynamic computation of length can unexpectedly change the length of the dataset itself. Specifically, the below assert does *not* pass whereas I would expect it to.

```
spatial = RandomPatchSampler(ds, size=PATCH, length=4, generator=0)
temporal = SequentialTimestampSampler(ds)
st = spatial @ temporal
num_samples = len(st)
num_samples_2 = len(list(st))
assert num_samples == num_samples_2
```

@adamjstewart flagged that he'd add a warning for this.

Aside from that, I'd love to see results about speed of this sampling, using these samplers in "real world" trainers, etc. but nothing in my tests jumped out as problematic s.t. these results are urgent (i.e. nothing was crazy slow).

adamjstewart:
> I'd love to see results about speed of this sampling

This is in the "Performance" section of the PR description. Once #3705 is merged, we can go from epoch time to sampling rate more directly, which will offer a more fair comparison if epoch length differ between branches.

calebrob6 (approved):

robmarkcole:
Just a couple of comments to address from @nilsleh and @sfalkena - possibly resolved already

nilsleh (approved):

sfalkena (approved):

adamjstewart:
99% of review requests have now been taken care of. There are a couple that remain involving the intricacies of error messages and details of documentation, but let's not hold up the PR for those. With 4 approvals, I will now merge and we can finally get these samplers into the hands of users so they can report bugs or suggest changes. The important thing is the high-level API, which I think is quite solid now. Thanks everyone for reviewing!

torchgeo/datamodules/geo.py
```diff
@@ -228,8 +228,8 @@ def setup(self, stage: str) -> None:
                     **self.kwargs,
                 ),
             )
-            self.train_batch_sampler = RandomBatchGeoSampler(
-                self.train_dataset, self.patch_size, self.batch_size, self.length
+            self.train_sampler = RandomPatchSampler(
+                self.train_dataset, size=self.patch_size, length=self.length
```
adriantre, line 232: Hi @adamjstewart! Here, you switched from a spatio-temporal sampler to a spatial-only sampler, correct? Some users could experience unintended behaviour without realising it. Worth adding a note about this in the release notes

adamjstewart: RandomBatchGeoSampler is a file-based sampler, not really spatial or temporal. It iterates over each file and randomly samples patches. We could use a spatio-temporal sampler here, but I'm not actually sure what makes most sense as a default.

I'll definitely have a detailed section of the release notes just for the sampler changes, as they are very much backwards-incompatible. I think they make more sense now (you can guarantee things about spatial or temporal coverage) but people trying to reproduce old studies will need old TorchGeo.

adriantre: True! If it was RandomGeoSampler, then it would have been like I said. I overlooked the Batch-part.

My opinion is that the default sampler (if not overridden) should be the one that merges pixels across files the least. Which would be the most similar to the old behaviour.

adamjstewart: RandomGeoSampler and RandomBatchGeoSampler are quite similar in how sampling is performed, only difference is mini-batch organization. Both are file-based. I agree though that we are now sampling across all time, which we weren't doing before.

PR #3528, opened 2026-03-23
[DRAFT] Add SpatioTemporalPixelwiseRegressionTask

robmarkcole:
- Partners https://github.com/torchgeo/torchgeo/pull/3280
- Pending https://github.com/torchgeo/torchgeo/pull/3086 or https://github.com/torchgeo/torchgeo/pull/3554

PR #3554, opened 2026-03-26
Add BioMasstersDataModule

robmarkcole:
Requires #3086


tests/datamodules/test_biomassters.py
adamjstewart, line 1: We are trying to get rid of this directory. I would instead suggest combining this PR and your SpatioTemporalPixelwiseRegressionTask PR into one so you can immediately test the datamodule works.

torchgeo/datamodules/biomassters.py
```diff
@@ -0,0 +1,236 @@
...
+            plot_sample['label'] = sample['mask']
+        if 'prediction' in sample:
+            plot_sample['prediction'] = sample['prediction']
+
+        return self.dataset.plot(
+            plot_sample, show_titles=show_titles, suptitle=suptitle
+        )
+
+
+def _pad_regression_time_series(
```
adamjstewart: A helper function for a helper function? Can't we directly use `pad_across_batches`?

torchgeo/datamodules/biomassters.py
```diff
@@ -0,0 +1,236 @@
...
+from torch import Tensor
+from torch.utils.data import random_split
+
+from ..datasets import BioMassters
+from ..datasets.geo import NonGeoDataset
+from ..datasets.utils import Sample, pad_across_batches
+from .geo import NonGeoDataModule
+
+
+class _BioMasstersSpatioTemporalRegressionDataset(NonGeoDataset):
```
adamjstewart: Can we instead change the dataset to match all other datasets? That will make this data module much simpler. Not backwards compatible, but now is the time to make such changes.

torchgeo/datamodules/biomassters.py
```diff
@@ -0,0 +1,236 @@
...
+    return collated
+
+
+class BioMasstersDataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the BioMassters dataset.
+
+    This datamodule adapts BioMassters to a fused spatiotemporal regression format
+    with samples shaped like ``{'image': (T, C, H, W), 'mask': (1, H, W)}``.
+
+    .. versionadded:: 0.9
```
adamjstewart:
```suggestion
    .. versionadded:: 0.10
```

torchgeo/datamodules/biomassters.py
```diff
@@ -0,0 +1,236 @@
...
+        dtype=torch.long,
+    )
+    return collated
+
+
+class BioMasstersDataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the BioMassters dataset.
+
+    This datamodule adapts BioMassters to a fused spatiotemporal regression format
+    with samples shaped like ``{'image': (T, C, H, W), 'mask': (1, H, W)}``.
```
adamjstewart: I think the mask should be HxW

torchgeo/datamodules/biomassters.py
```diff
@@ -0,0 +1,236 @@
...
+        sensors: Sequence[Literal['S1', 'S2']] = ('S1', 'S2'),
+        **kwargs: Any,
+    ) -> None:
+        """Initialize a new BioMasstersDataModule instance.
+
+        Args:
+            batch_size: Size of each mini-batch.
+            num_workers: Number of workers for parallel data loading.
+            val_split_pct: Percentage of the labeled train split used for validation.
+            test_split_pct: Percentage of the labeled train split used for testing.
```
adamjstewart, line 42: The dataset already has a test split, does it not have public labels?

torchgeo/datamodules/biomassters.py
```diff
@@ -0,0 +1,236 @@
...
+            _BioMasstersSpatioTemporalRegressionDataset,
+            batch_size=batch_size,
+            num_workers=num_workers,
+            sensors=sensors,
+            **kwargs,
+        )
+        self.val_split_pct = val_split_pct
+        self.test_split_pct = test_split_pct
+        self.padding_length = padding_length
+        self.collate_fn = lambda batch: _pad_regression_time_series(
```
adamjstewart: I personally prefer `functools.partial` over `lambda`

adamjstewart (dismissed):
Would like to make this simpler by making the fixes directly in the original dataset instead of subclassing.

robmarkcole:
To be addressed in upcoming SpatioTemporalPixelwiseRegressionTask PR

torchgeo/datamodules/biomassters.py
robmarkcole: Yes. The helper has been removed and collate_fn now directly uses functools.partial(pad_across_batches, ...).

torchgeo/datamodules/biomassters.py
```diff
@@ -0,0 +1,232 @@
...
+
+                image = self.dataset._load_input(sensor_df['filename'].tolist())
+                sensor_images[sensor] = image.squeeze(0)
+
+            reference = next(iter(sensor_images.values()))
+            _, height, width = reference.shape
+
+            channels = []
+            for sensor in self.sensors:
+                if sensor in sensor_images:
```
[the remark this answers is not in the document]
robmarkcole: BioMassters now explicitly validates that each sensor has exactly one acquisition per month and indexes the resulting tensor with [0]. A regression test covers duplicate acquisitions.

torchgeo/datamodules/biomassters.py
robmarkcole: Done. BioMassters itself now returns fused image and mask samples, including time-series fusion and zero-filled channels for missing sensor acquisitions. The datamodule is now a thin wrapper around the public dataset.

torchgeo/datamodules/biomassters.py
robmarkcole: Updated. BioMassters now squeezes the single target band and returns mask with shape H x W; batched masks have shape B x H x W.

torchgeo/datamodules/biomassters.py
robmarkcole: The competition test split has no public AGBM labels. The datamodule therefore splits the labeled train data for train/validation/test evaluation and reserves the official unlabeled test split for prediction.

torchgeo/datamodules/biomassters.py
```diff
@@ -0,0 +1,232 @@
...
+        self.dataset = BioMassters(
+            root=root, split=split, sensors=self.sensors, as_time_series=True
+        )
+        self.sample_groups = [
+            sample_df.copy()
+            for _, sample_df in self.dataset.df.groupby('num_index', sort=True)
+        ]
+
+    def __getitem__(self, index: int) -> Sample:
+        """Return an index within the dataset.
```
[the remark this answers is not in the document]
robmarkcole: Removed the adapter and its copied DataFrame groups. BioMassters now performs the fusion directly from its own metadata on demand.

torchgeo/datamodules/biomassters.py
robmarkcole: Updated to functools.partial(pad_across_batches, ...) and removed both lambdas.

tests/datamodules/test_biomassters.py
robmarkcole: The earlier task PR #3528 was closed, and SpatioTemporalRegression is currently being reviewed in #3086. This PR now uses the standard image/mask dataset contract and directly tests every dataloader; end-to-end task coverage remains with the task PR.

torchgeo/datamodules/biomassters.py
```diff
@@ -0,0 +1,232 @@
...
+        self.test_split_pct = test_split_pct
+        self.padding_length = padding_length
+        self.collate_fn = lambda batch: _pad_regression_time_series(
+            batch, padding_length=self.padding_length
+        )
+        self.aug = lambda batch: batch
+
+    def setup(self, stage: str) -> None:
+        """Set up datasets.
+
```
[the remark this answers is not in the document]
robmarkcole: Replaced the collate lambda with functools.partial and the augmentation lambda with torch.nn.Identity, both of which are picklable.

torchgeo/datamodules/biomassters.py
```diff
@@ -0,0 +1,232 @@
...
+            'length': torch.tensor(len(frames), dtype=torch.long),
+        }
+        if self.split == 'train':
+            sample['mask'] = self.dataset._load_target(
+                sample_df['corresponding_agbm'].iloc[0]
+            )
+
+        return sample
+
+    def __len__(self) -> int:
```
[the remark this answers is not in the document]
robmarkcole: Removed the adapter entirely. BioMassters now exposes the fused image and mask through its public __getitem__ interface, and the datamodule instantiates it directly.

robmarkcole:
still pending https://github.com/torchgeo/torchgeo/pull/3086

robmarkcole:
@adamjstewart removed that test file

torchgeo/datamodules/biomassters.py
```diff
@@ -0,0 +1,87 @@
...
+from .geo import NonGeoDataModule
+
+
+class BioMasstersDataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the BioMassters dataset.
+
+    Samples have the fused spatiotemporal regression format
+    ``{'image': (T, C, H, W), 'mask': (H, W)}``.
+
+    .. versionadded:: 0.10
```
nilsleh: 0.10 has already been released

robmarkcole:
@adamjstewart any blockers on this?

torchgeo/datasets/biomassters.py
```diff
@@ -146,24 +147,70 @@ def __getitem__(self, index: int) -> Sample:
...
+            month_df = sample_df[sample_df['num_month'] == month]
+            sensor_images: dict[str, Tensor] = {}
+            for sensor in self.sensors:
+                filenames = month_df[month_df['satellite'] == sensor][
+                    'filename'
+                ].tolist()
+                if filenames:
+                    if len(filenames) != 1:
+                        msg = f'Expected one {sensor} acquisition per month.'
+                        raise ValueError(msg)
```
adamjstewart, line 192: Does this situation actually occur in the dataset or is this just defensive coding for something that doesn't happen?

tests/datasets/test_biomassters.py
```diff
@@ -37,8 +51,20 @@ def test_plot(self, dataset: BioMassters) -> None:
...
+        with pytest.raises(ValueError, match='Expected image tensor'):
+            dataset.plot({'image': torch.zeros(1)})
+
+    def test_duplicate_monthly_acquisition(self) -> None:
+        root = os.path.join('tests', 'data', 'biomassters')
+        dataset = BioMassters(root, split='train', as_time_series=True)
+        dataset.df = pd.concat([dataset.df, dataset.df.iloc[[0]]])
+
+        with pytest.raises(ValueError, match='Expected one S1 acquisition per month'):
+            dataset[0]
```
adamjstewart, line 70: Delete

torchgeo/datamodules/biomassters.py
adamjstewart, line 24:
```suggestion
    .. versionadded:: 0.11
```

torchgeo/datamodules/biomassters.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+
+    target_mean = 0
+    target_std = 1
```
adamjstewart, line 28: These are unused?

torchgeo/datamodules/biomassters.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def __init__(
+        self,
+        batch_size: int = 32,
+        num_workers: int = 0,
+        val_split_pct: float = 0.2,
+        test_split_pct: float = 0.2,
+        padding_length: int = 12,
+        sensors: Sequence[Literal['S1', 'S2']] = ('S1', 'S2'),
```
adamjstewart, line 37: Remove this argument, it can be included in kwargs

torchgeo/datasets/biomassters.py
```diff
@@ -146,24 +147,70 @@ def __getitem__(self, index: int) -> Sample:
...
+                    sensor_images.get(
+                        sensor,
+                        torch.zeros(
+                            self.channel_counts[sensor],
+                            height,
+                            width,
+                            dtype=reference.dtype,
+                        ),
+                    )
+                )
```
adamjstewart, line 209: What in the world is going on here?

torchgeo/datasets/biomassters.py
```diff
@@ -236,18 +283,30 @@ def plot(
...
-        fig, axs = plt.subplots(1, ncols=ncols, figsize=(5 * ncols, 10))
+        fig, axs_array = plt.subplots(
+            1, ncols=ncols, figsize=(5 * ncols, 10), squeeze=False
+        )
+        axs = axs_array[0]
+        image = sample['image'].float()
+        expected_ndim = 4 if self.as_time_series else 3
+        if image.ndim != expected_ndim:
+            msg = f'Expected image tensor with {expected_ndim} dimensions.'
+            raise ValueError(msg)
```
adamjstewart, line 301: Why would this ever happen?

adamjstewart (changes requested):
Code is basically illegible, see if you can get this down to +50 lines of code at most.

torchgeo/datamodules/biomassters.py
robmarkcole: unused, removed

torchgeo/datamodules/biomassters.py
robmarkcole: done

torchgeo/datasets/biomassters.py
robmarkcole: removed

tests/datasets/test_biomassters.py
robmarkcole: done

torchgeo/datasets/biomassters.py
robmarkcole: Handle missing S2 data with zero filled channels

torchgeo/datasets/biomassters.py
robmarkcole: overly defensive, removed

robmarkcole:
Reduced to ~ 50 datamodule lines.

- Removed _load_time_series.
- Load each sensor once and place acquisitions directly into a fixed 12-month tensor.
- Removed custom padding/collation.
- Simplified plotting with torch.split.
- Trimmed redundant test configuration.
- Restored target_mean/std; the integration test proved the task uses them.

Couple of other points:
- The competition defined a train split (test had no labels), should we use that rather than random splitting? The [reference implementation](https://github.com/quqixun/BioMassters/blob/main/split.py) actually uses cross validation
- Normalisation: currently none applied, the reference implementation uses z-score norm, use that?
Both those changes would increase LOC

robmarkcole:
Updated to use the dataset train/test splits. Took around 12 hours to download the full dataset so planning a biomasters-100 dataset. 

Using the full dataset exposed a couple more issues:

- S1 float32 dB data is cast to int32, losing precision.
- The plotting normalization assumes positive linear S1 values, making S1 panels almost black.
- Some target masks contain extreme outliers: sampled maxima reached 3,190 and 6,079, while most values are below approximately 500. This could dominate MSE training.

### Mask outliers
The [reference implementation](https://github.com/quqixun/BioMassters/tree/main) provides a clear precedent: its default “plain” preprocessing derives label bounds from the training-set 0.1/99.9 percentiles, then clips and min–max normalizes every training label. Its published upper bound is 425.7 -> the clip is now applied

## S1 dtype
Casting to int32 is an error, the ref implementation and other torchgeo S1 datasets use float32, so use that.

## Normalisation
IMO the datamodule is only useful if it also performs normalisation as per the reference implementation. Now norm the imagery and target using the reference implementation values. Plots of norm and un-norm data sample below:

biomassters_test_0_normalized
biomassters_test_0_raw

robmarkcole:
@adamjstewart this is ready for another look. Also created https://github.com/torchgeo/torchgeo/pull/4132

PR #3557, opened 2026-03-27
Add Landslide4Sense dataset, loaders & tests

ArsalaanAhmad:
## Summary

This PR adds support for the Landslide4Sense benchmark dataset.

### Included
- `Landslide4Sense` dataset class in `torchgeo/datasets/landslide4sense.py`
- import/export in `torchgeo/datasets/__init__.py`
- fake HDF5 test data generator
- dataset unit tests
- docs stub for autodoc

## Dataset structure

This implementation follows the public Landslide4Sense release structure:

- `train`: images + masks
- `val`: images only
- `test`: images only

Only the training split includes masks in the public release, so `val` and `test` samples return image-only dictionaries.

## Notes

- each image sample is loaded from an HDF5 file using the `img` key
- training masks are loaded from HDF5 files using the `mask` key
- tests use small synthetic HDF5 fixtures matching the expected dataset structure

## Validation

- added unit tests for split handling, sample shapes, mask presence/absence, invalid splits, and band subsetting
- `py -m pytest tests/datasets/test_landslide4sense.py` passes locally


tests/data/landslide4sense/data.py
```diff
@@ -0,0 +1,39 @@
+from __future__ import annotations
```
adamjstewart: Missing copyright header. Why import annotations?

tests/data/landslide4sense/data.py
```diff
@@ -0,0 +1,39 @@
...
+
+from pathlib import Path
+
+import numpy as np
+
+
+def _write_image(path: Path, shape: tuple[int, int, int] = (32, 32, 14)) -> None:
+    import h5py
```
adamjstewart: Let's move this import to the top of the file

tests/datasets/test_landslide4sense.py
```diff
@@ -0,0 +1,88 @@
+from __future__ import annotations
```
adamjstewart: Missing copyright, no need for annotations import

tests/datasets/test_landslide4sense.py
```diff
@@ -0,0 +1,88 @@
...
+
+from pathlib import Path
+
+import pytest
+import torch
+
+pytest.importorskip('h5py')
+
+from tests.data.landslide4sense.data import create
```
adamjstewart: This should not be run during testing, it should be run once and the sample files should be saved to the repo.

tests/datasets/test_landslide4sense.py
adamjstewart, line 1: Doesn't match the structure of any of our existing dataset tests

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,175 @@
...
+import numpy as np
+import torch
+
+from .geo import NonGeoDataset
+
+
+class Landslide4Sense(NonGeoDataset):
+    """Landslide4Sense benchmark dataset.
+
+    The Landslide4Sense dataset is a benchmark for landslide detection from
```
adamjstewart, line 23: URL?

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,175 @@
...
+         Weile and Peng, Weihang and Ghamisi, Pedram},
+         journal={IEEE Journal of Selected Topics in Applied Earth Observations and
+         Remote Sensing},
+         title={The Outcome of the 2022 Landslide4Sense Competition: Advanced
+         Landslide Detection From Multisource Satellite Imagery},
+         year={2022},
+         volume={15},
+         pages={9927-9942},
+         doi={10.1109/JSTARS.2022.3220845}
+       }
```
adamjstewart: Can we instead link to a paper? Then people can use Google Scholar to generate any format they want, not just bibtex.

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,175 @@
...
+    )
+
+    def __init__(
+        self,
+        root: str | Path,
+        split: str = 'train',
+        transforms: Callable | None = None,
+        bands: list[int] | None = None,
+    ) -> None:
+        """Initialize a new Landslide4Sense dataset instance."""
```
adamjstewart: Need to document all parameters.

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,175 @@
...
+        'B11',
+        'B12',
+        'Slope',
+        'DEM',
+    )
+
+    def __init__(
+        self,
+        root: str | Path,
+        split: str = 'train',
```
adamjstewart:
```suggestion
        split: Literal['train', 'val', 'test'] = 'train',
```

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,175 @@
...
+        transforms: Callable | None = None,
+        bands: list[int] | None = None,
```
adamjstewart:
```suggestion
        bands: Sequence[int] | None = None,
```

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,175 @@
...
+        'B12',
+        'Slope',
+        'DEM',
+    )
+
+    def __init__(
+        self,
+        root: str | Path,
+        split: str = 'train',
+        transforms: Callable | None = None,
```
adamjstewart: Callable isn't very specific, it should be:

```suggestion
        transforms: Callable[[Sample], Sample] | None = None,
```

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,175 @@
...
+        if split not in self.splits:
+            raise ValueError(f"Invalid split '{split}'. Expected one of {self.splits}.")
+
+        self.root = Path(root)
+        self.split = split
+        self.transforms = transforms
+        self.bands = bands
+        self.samples = self._load_samples()
+
+        if len(self.samples) == 0:
```
adamjstewart: Is it possible to add automatic download support?

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,175 @@
...
+            raise FileNotFoundError(
+                f"No samples found for split='{split}' under root='{self.root}'."
+            )
+
+    def _split_dirs(self) -> tuple[Path, Path | None]:
```
adamjstewart, line 108: Needs a docstring

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,175 @@
...
+
+        return samples
+
+    def __len__(self) -> int:
+        """Return the number of samples in the dataset."""
+        return len(self.samples)
+
+    def __getitem__(self, index: int) -> dict[str, torch.Tensor]:
+        """Return a single sample from the dataset."""
+        import h5py
```
adamjstewart: This should use `lazy_import`, see the other datasets that use h5py for examples.

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,175 @@
...
+
+            if mask.ndim != 2:
+                raise ValueError(f'Expected mask with 2 dims, got shape {mask.shape}')
+
+            sample['mask'] = torch.from_numpy(mask).unsqueeze(0).float()
+
+        if self.transforms is not None:
+            sample = self.transforms(sample)
+
+        return sample
```
adamjstewart, line 194: Missing a plot method.

adamjstewart (changes requested):
Before submitting a new dataset PR, please read:

* https://torchgeo.readthedocs.io/en/stable/user/contributing.html
* https://torchgeo.readthedocs.io/en/stable/tutorials/contribute_non_geo_dataset.html

You are currently missing some documentation and expected methods. I would recommend basing your PR off of an existing dataset. There are already several that use h5py.

tests/data/landslide4sense/data.py
ArsalaanAhmad: The problem i was facing with that was that the CI/CD Tests kept crashing when the import was on top, is there any other way to fix that while moving the import to the top?

adamjstewart: If you don't run this code inside your tests then you'll be fine

tests/data/landslide4sense/data.py
ArsalaanAhmad: Hello! Thank you for your comments, I've added the fixes and would love another review. I've closely followed other h5py using databases to tweak my test file.

tests/data/landslide4sense/data.py
ArsalaanAhmad: Thank you, that did it!

torchgeo/datasets/landslide4sense.py
ArsalaanAhmad: I'm not exactly sure unfortunately, I'll have a look.

torchgeo/datasets/landslide4sense.py
ArsalaanAhmad: Thank you, I've now added one

tests/datasets/test_landslide4sense.py
ArsalaanAhmad: my bad, I've edited the test file to resemble the other h5py dataset test files, it's working as of now but let me know if there is any additional problem.

ArsalaanAhmad:
> Before submitting a new dataset PR, please read:
> 
> * https://torchgeo.readthedocs.io/en/stable/user/contributing.html
> * https://torchgeo.readthedocs.io/en/stable/tutorials/contribute_non_geo_dataset.html
> 
> You are currently missing some documentation and expected methods. I would recommend basing your PR off of an existing dataset. There are already several that use h5py.

Hello, once again, thanks for the comment. I'm actually an undergrad just dipping my toes into Open-Source contribution :) 
I've made the changes you've requested, if there is anything left, do let me know. About the link to the dataset, i'll link it within the description when I get it.

torchgeo/datasets/landslide4sense.py
ArsalaanAhmad: These are the ones I have:
https://arxiv.org/abs/2206.00515 
https://github.com/iarai/Landslide4Sense-2022
https://www.kaggle.com/datasets/tekbahadurkshetri/landslide4sense

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,248 @@
...
+
+            samples.append((image_path, mask_path))
+
+        return samples
+
+    def __len__(self) -> int:
+        """Return the number of samples in the dataset."""
+        return len(self.samples)
+
+    def __getitem__(self, index: int) -> Sample:
```
nilsleh, line 160: need a docstring for the argument and return

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,248 @@
...
+
+        if self.transforms is not None:
+            sample = self.transforms(sample)
+
+        return sample
+
+    def plot(
+        self, sample: Sample, show_titles: bool = True, suptitle: str | None = None
+    ) -> Figure:
+        """Plot a sample from the dataset."""
```
nilsleh: missing docstring for argument and return

nilsleh (changes requested):

nilsleh:
@ArsalaanAhmad Can you rebase this PR, meaning that you merge the latest updated main version into this branch, such that we can clearly see the difference of your PR branch with the main version?

.readthedocs.yaml
```diff
@@ -7,6 +7,8 @@ version: 2
 # Set the version of Python
 build:
   os: ubuntu-lts-latest
+  apt_packages:
```
nilsleh: why is this change necessary in this PR?

ArsalaanAhmad:
> @ArsalaanAhmad Can you rebase this PR, meaning that you merge the latest updated main version into this branch, such that we can clearly see the difference of your PR branch with the main version?

Yes I can do that @nilsleh, it should be done now. There were some rebase conflicts due to upstream conflicts but hopefully they're fixed now.

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,248 @@
...
+    )
+    rgb_bands = ('B4', 'B3', 'B2')
+
+    def __init__(
+        self,
+        root: str | Path,
+        split: Literal['train', 'val', 'test'] = 'train',
+        transforms: Callable[[Sample], Sample] | None = None,
+        bands: Sequence[int] | None = None,
+    ) -> None:
```
nilsleh, line 83: Missing docstrings for all arguments

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,248 @@
...
+         title={The Outcome of the 2022 Landslide4Sense Competition: Advanced
+         Landslide Detection From Multisource Satellite Imagery},
+         year={2022},
+         volume={15},
+         pages={9927-9942},
+         doi={10.1109/JSTARS.2022.3220845}
+       }
+
+    Furthermore, you can also cite the original dataset paper: https://arxiv.org/abs/2206.00515
+
```
nilsleh, line 54: missing versionadded parameter for documentation

nilsleh (changes requested):

nilsleh:
@ArsalaanAhmad As Adam said, you might find it helpful to go through the checklist here: https://torchgeo.readthedocs.io/en/stable/tutorials/contribute_non_geo_dataset.html#Documentation-Entries to make sure you aren't missing things.

torchgeo/datasets/landslide4sense.py
ArsalaanAhmad: This should be fixed, I assume this pertains to the milestone 0.10.0?

torchgeo/datasets/landslide4sense.py
ArsalaanAhmad: Should be fixed now

torchgeo/datasets/landslide4sense.py
ArsalaanAhmad: Fixed.

torchgeo/datasets/landslide4sense.py
ArsalaanAhmad: Fixed this by adding the necessary docstrings now.

.readthedocs.yaml
ArsalaanAhmad: Hello, thanks for the review. 
I added pandoc to the RTD apt packages because the docs use nbsphinx which requires system pandoc for notebook conversion. Prior to adding this I was having test failures with RTD. That said, if there's any alternative or workaround I might not be aware of, please let me know!

ArsalaanAhmad:
> @ArsalaanAhmad As Adam said, you might find it helpful to go through the checklist here: https://torchgeo.readthedocs.io/en/stable/tutorials/contribute_non_geo_dataset.html#Documentation-Entries to make sure you aren't missing things.

@nilsleh Hello Nils, thanks for the review. I did forget some things from this checklist previously, I followed it closely and have added the necessary documentation and changes. 
I have two enquiries regarding the pandoc apt package and the versionadded docstring, do let me know in their respective comments.

.readthedocs.yaml
nilsleh: Mhm interesting. Pandoc was not necessary before for the entire torchgeo documentation, so it is worth investigating what is specific about this case that would need pandoc. Can you show the test failures you are encountering?

nilsleh:
The tests seem to be missing some coverage and need to be extended. You can see common use patterns of tests in other datasets and can test code coverage locally as instructed here: https://torchgeo.readthedocs.io/en/stable/user/contributing.html#tests

.readthedocs.yaml
ArsalaanAhmad: nevermind about this, I removed it and the test still succeeded. I assume the problem was due to the previous conflict with the head branch(?)
eitherway, we're good!

ArsalaanAhmad:
> The tests seem to be missing some coverage and need to be extended. You can see common use patterns of tests in other datasets and can test code coverage locally as instructed here: https://torchgeo.readthedocs.io/en/stable/user/contributing.html#tests

Hello, I noticed this earlier and saw that it was the lines with "Raise" that were not being covered in the tests. I commited the extra tests to cover these. It should broadly be similar to other datasets that also use h5py but there are a few tests that are specific to this dataset due to the structuring etc. Some of these include: 
1. Split-dependent sample schema tests: train has mask, val/test are image-only.
2. Directory layout and file discovery failure tests: missing mask dir, no h5 files, missing paired mask file
3. Specific band indexing tests.
4. Constructor validation paths are a lot more explicit

Let me know if anything else is needed from this PR. @nilsleh

torchgeo/datasets/landslide4sense.py
nilsleh: Second that, just having the link to the paper is sufficient

torchgeo/datasets/landslide4sense.py
nilsleh: Do you know more about whether this is possible? This makes the dataset significantly more accessible to users.

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,175 @@
...
+        self.split = split
+        self.transforms = transforms
+        self.bands = bands
+        self.samples = self._load_samples()
+
+        if len(self.samples) == 0:
+            raise FileNotFoundError(
+                f"No samples found for split='{split}' under root='{self.root}'."
+            )
+
```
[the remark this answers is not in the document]
nilsleh: Copilot has a point here, you are already checking this in the _load_samples() function

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,286 @@
...
+        """
+        h5py = lazy_import('h5py')
+
+        image_path, mask_path = self.samples[index]
+
+        with h5py.File(image_path, 'r') as f:
+            image = f['img'][:]
+
+        if image.ndim != 3:
```
nilsleh: Under what circumstances do you expect there to be images with differing number of dimensions?

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,175 @@
...
+        sample: dict[str, torch.Tensor] = {'image': torch.from_numpy(image).float()}
+
+        if mask_path is not None:
+            with h5py.File(mask_path, 'r') as f:
+                mask = f['mask'][:]
+
+            if mask.ndim != 2:
+                raise ValueError(f'Expected mask with 2 dims, got shape {mask.shape}')
+
+            sample['mask'] = torch.from_numpy(mask).unsqueeze(0).float()
```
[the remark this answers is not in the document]
nilsleh: If the mask is an integer based segmentation mask, it should be long datatype

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,286 @@
...
+        if self.bands is not None:
+            image = image[self.bands, :, :]
+
+        sample: Sample = {'image': torch.from_numpy(image).float()}
+
+        if mask_path is not None:
+            with h5py.File(mask_path, 'r') as f:
+                mask = f['mask'][:]
+
+            if mask.ndim != 2:
```
nilsleh: similar question here, is the dataset a mix of different masks with varying number of dimensions or why is this check necessary?

torchgeo/datasets/landslide4sense.py
ArsalaanAhmad: Hello, that's a good catch. I don’t expect the official Landslide4Sense release to contain mixed-dimensional masks or images. The check was intended as more of a defensive validation at the file boundary so that malformed or corrupted .h5 contents fail with a clear dataset-specific error rather than later with a less obvious tensor/indexing failure. If you’d prefer to keep the loader leaner and rely on the fixed dataset contract or if torchgeo has failsafes in place for this, I’m happy to remove it.

torchgeo/datasets/landslide4sense.py
ArsalaanAhmad: This too is simply just defensive validation for malformed or nonconforming files, not an expectation that valid dataset samples vary in dimensionality. 
the masks are 2d arrays while images/returned tensors are 3d arrays.

torchgeo/datasets/landslide4sense.py
ArsalaanAhmad: I believe this is slightly tricky. I assume that at some point it **WAS** possible since if you have a look at the official repository of the dataset, it had clear and clean download links for all splits of the dataset. 

That said, those links don't work anymore and seem to show that the server is unavailable. The dataset is conversely available to download publicly from Kaggle, which I don't think supports automatic download support since it involves: Kaggle account/authentication, Kaggle API setup and a fairly brittle user experience.

docs/api/datasets/benchmark.csv
```diff
@@ -33,6 +33,7 @@ Dataset,Task,Source,# Samples,# Classes,Size (px),Resolution (m),Bands,License
 :ref:`Kenya Crop Type`,S,Sentinel-2,"4,688",7,"3,035x2,016",10,MSI,CC-BY-SA-4.0
 :ref:`L7 Irish`,S,Landsat 7,206,5,"8,400x7,500","15, 30",MSI,CC0-1.0
 :ref:`L8 Biome`,S,Landsat 8,96,5,"8,900x8,900","15, 30",MSI,CC0-1.0
+:ref:`Landslide4Sense`,S,"Sentinel-2, ALOS PALSAR",2,14,"128x128",-,"MSI, SAR",
```
nilsleh: can you add the resolution range for the two sensors? You can see a couplt of datasets that have multiple Sources in the table.

torchgeo/datasets/landslide4sense.py
nilsleh: Could you take a look at simplifying this

torchgeo/datasets/landslide4sense.py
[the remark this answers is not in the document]
nilsleh: Can you also address this? This file check is happening twice essentially.

torchgeo/datasets/landslide4sense.py
nilsleh: I assume you have the dataset locally, so perhaps you can iterate through the whole dataset to make sure all files are valid. But since this is a curated dataset, I think this is not necessary, and also will simplify your tests.

torchgeo/datasets/landslide4sense.py
nilsleh: Same comment as above, I don't think it is necessary, and will simplify tests.

torchgeo/datasets/landslide4sense.py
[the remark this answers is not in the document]
nilsleh: It seems this is still a float in the code?

nilsleh (commented):
Can you also plot a sample and add it to the PR comment? That checks the plotting function and also gives a sense of the dataset.

docs/api/datasets/benchmark.csv
ArsalaanAhmad: Yes I'll do that, I haven't committed the other changes yet.

ArsalaanAhmad: The native sensor resolutions are available in the paper: Sentinel-2 bands are provided at 10/20/60 m, and ALOS PALSAR DEM is 12.5 m. The benchmark itself resamples all bands to 10 m per pixel. So for the table, I think the most appropriate entry is 10-60, 12.5.

torchgeo/datasets/landslide4sense.py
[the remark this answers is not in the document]
ArsalaanAhmad: I've fixed this now

ArsalaanAhmad:
@nilsleh I've added the requested changes. Here is the plot method picture too:
image

left shows a RGB composite built from channels B4, B3, B2 (the dataset’s red/green/blue bands). The code transposes the 14-channel array to (C,H,W), selects the RGB indices, permutes to (H,W,C) and normalizes before calling. The right is the binary mask. 

hopefully this is fine?

docs/api/datasets/benchmark.csv
ArsalaanAhmad: @nilsleh @adamjstewart could i have a re-review to confirm if everything is ok?

nilsleh: yes that seems good to me

nilsleh:
@ArsalaanAhmad That example plot looks wrong to me, like that cannot be the expected input image to the model? If I look at the original repo for this dataset https://github.com/iarai/Landslide4Sense-2022#data-description (correct me if I am wrong) the images and masks look very different, so I don't think this is correct. Either the plotting function or the data itself. Or are you plotting the random test data?

torchgeo/datasets/landslide4sense.py
nilsleh: You can add the github repo here in the docstring, and later mention the arxiv paper to be cited

ArsalaanAhmad:
> @ArsalaanAhmad That example plot looks wrong to me, like that cannot be the expected input image to the model? If I look at the original repo for this dataset https://github.com/iarai/Landslide4Sense-2022#data-description (correct me if I am wrong) the images and masks look very different, so I don't think this is correct. Either the plotting function or the data itself. Or are you plotting the random test data?

@nilsleh Hey, The current plot output is generated from the synthetic .h5 test fixtures (i assume that's what you mean by random test data) used for CI/testing rather than real Landslide4Sense samples, which is why it looks visually unrealistic compared to the examples in the original repository. The plotting logic itself is using the expected RGB bands and mask structure from the dataset implementation.

I can improve the synthetic fixtures to produce more realistic spatial patterns if that would make the visualization less misleading, or add real images from the dataset with their respective masks. From the original repo of the dataset, this is how it would look like ideally: 
image

adamjstewart:
> I can improve the synthetic fixtures to produce more realistic spatial patterns if that would make the visualization less misleading, or add real images from the dataset with their respective masks.

No need to make the fake data look pretty, we just want example plots of the real data. We can't add the real data to the repo due to licensing.

ArsalaanAhmad:
@adamjstewart @nilsleh Hello, I assume this is what you were thinking of? Hopefully it's fine now

image
image

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,259 @@
...
+        Secondary source:
+        https://github.com/iarai/Landslide4Sense-2022/tree/main
+
+        Additional dependencies:
+
+        * ``h5py`` for reading the dataset files.
+
+    If you use this dataset in your research, please cite it using the following
+    format:
+
```
calebrob6, line 50: Missing something here

ArsalaanAhmad: @calebrob6 I've fixed this in the recent push, essentially it's just that i was suggested to cite using the arxiv paper instead of the bibtex. 

also do you think it would be even better to additionally cite the paper (that i authored) where I created the loaders for this specific dataset? not sure if it would be useful, but let me know.

Is there anything else?

tests/datasets/test_landslide4sense.py
```diff
@@ -0,0 +1,165 @@
...
+
+        assert isinstance(sample, dict)
+        assert isinstance(sample['image'], torch.Tensor)
+        assert isinstance(sample['mask'], torch.Tensor)
+
+        assert sample['image'].shape == (14, 32, 32)
+        assert sample['mask'].shape == (1, 32, 32)
+        assert sample['mask'].dtype == torch.int64
+
+    def test_len_train(self, dataset: Landslide4Sense) -> None:
```
adamjstewart, line 36: These tests could be greatly simplified. You can parametrize the fixture over train/val/test splits and remove most of the code duplication.

tests/datasets/test_landslide4sense.py
```diff
@@ -0,0 +1,165 @@
...
+        root = os.path.join('tests', 'data', 'landslide4sense')
+        dataset = Landslide4Sense(root=root, split='val')
+        assert len(dataset) == 1
+
+    def test_len_test(self) -> None:
+        root = os.path.join('tests', 'data', 'landslide4sense')
+        dataset = Landslide4Sense(root=root, split='test')
+        assert len(dataset) == 1
+
+    def test_invalid_split(self, dataset: Landslide4Sense) -> None:
```
adamjstewart, line 49: Could probably remove this, we usually only test valid inputs and let type checkers flag invalid inputs

tests/datasets/test_landslide4sense.py
```diff
@@ -0,0 +1,165 @@
...
+        with pytest.raises(FileNotFoundError, match='Mask directory not found'):
+            Landslide4Sense(root=root, split='train')
+
+    def test_missing_h5_files(self, tmp_path: Path) -> None:
+        root = tmp_path / 'landslide4sense'
+        (root / 'TrainData' / 'img').mkdir(parents=True)
+        (root / 'TrainData' / 'mask').mkdir(parents=True)
+
+        with pytest.raises(FileNotFoundError, match=r'No \.h5 files found'):
+            Landslide4Sense(root=root, split='train')
```
adamjstewart, line 71: These seem like they are all the same test?

tests/datasets/test_landslide4sense.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def test_no_samples_found(self, monkeypatch: pytest.MonkeyPatch) -> None:
```
adamjstewart, line 73: No need to mock anything, the tests should be simple

tests/datasets/test_landslide4sense.py
```diff
@@ -0,0 +1,165 @@
...
+        assert sample['image'].shape == (14, 32, 32)
+
+    def test_band_subset(self, dataset: Landslide4Sense) -> None:
+        dataset = Landslide4Sense(root=dataset.root, split='train', bands=[0, 3, 13])
+        sample = dataset[0]
+
+        assert sample['image'].shape == (3, 32, 32)
+        assert sample['mask'].shape == (1, 32, 32)
+
+    def test_missing_train_mask(self, tmp_path: Path) -> None:
```
adamjstewart, line 110: Delete

tests/datasets/test_landslide4sense.py
```diff
@@ -0,0 +1,165 @@
...
+        ) -> list[tuple[Path, Path | None]]:
+            return []
+
+        monkeypatch.setattr(Landslide4Sense, '_load_samples', _mock_load_samples)
+
+        root = os.path.join('tests', 'data', 'landslide4sense')
+        dataset = Landslide4Sense(root=root, split='train')
+        assert len(dataset) == 0
+
+    def test_val_sample_contains_only_image(self) -> None:
```
adamjstewart, line 85: Only need a single test for getitem parameterized across all splits

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,258 @@
...
+    The public release contains the following splits:
+
+    * ``train``: images and masks
+    * ``val``: images only
+    * ``test``: images only
+
+     .. note::
+
+         In the public release, only the training split includes masks.
```
adamjstewart, line 36: Duplicate of previous paragraph

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,258 @@
...
+
+         The public dataset can be downloaded from:
+         https://www.kaggle.com/datasets/tekbahadurkshetri/landslide4sense
+
+         Secondary source:
+         https://github.com/iarai/Landslide4Sense-2022/tree/main
```
adamjstewart, line 42: Doesn't seem to contain the dataset?

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,258 @@
...
+         Additional dependencies:
+
+         * ``h5py`` for reading the dataset files.
+
+     If you use this dataset in your research, please cite it using the following
+     link:
+
+     * https://arxiv.org/abs/2206.00515
+
+    .. versionadded:: 0.10.0
```
adamjstewart, line 53:
```suggestion
    .. versionadded:: 0.10
```

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,258 @@
...
+        'B11',
+        'B12',
+        'Slope',
+        'DEM',
+    )
+    rgb_bands = ('B4', 'B3', 'B2')
+
+    def __init__(
+        self,
+        root: str | Path,
```
adamjstewart, line 79:
```suggestion
        root: str | Path = 'data',
```

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,258 @@
...
+
+        Raises:
+            ValueError: If ``split`` is not one of the supported split names.
+            FileNotFoundError: If required split files cannot be found.
+
+        """
+        lazy_import('h5py')
+
+        if split not in self.splits:
+            raise ValueError(f"Invalid split '{split}'. Expected one of {self.splits}.")
```
adamjstewart, line 100: Can remove, type checkers will handle this

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,258 @@
...
+
+        self.root = pathlibPath(root)
+        self.split = split
+        self.transforms = transforms
+        self.bands = bands
```
adamjstewart, line 105:
```suggestion
        self.bands = bands or self.band_names
```

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,258 @@
...
+        'DEM',
+    )
+    rgb_bands = ('B4', 'B3', 'B2')
+
+    def __init__(
+        self,
+        root: str | Path,
+        split: Literal['train', 'val', 'test'] = 'train',
+        transforms: Callable[[Sample], Sample] | None = None,
+        bands: Sequence[int] | None = None,
```
adamjstewart, line 82:
```suggestion
        bands: Sequence[str] | None = None,
```

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,258 @@
...
+            if full_index in available_band_indices:
+                rgb_indices.append(available_band_indices.index(full_index))
+            else:
+                raise RGBBandsMissingError()
+
+        has_mask = 'mask' in sample
+        has_prediction = 'prediction' in sample
+
+        ncols = 1 + int(has_mask) + int(has_prediction)
+        fig, axs = plt.subplots(1, ncols, figsize=(8 * ncols, 8))
```
adamjstewart, line 228: If you use `squeeze=False` you won't need the next line

torchgeo/datasets/landslide4sense.py
```diff
@@ -0,0 +1,258 @@
...
+
+        if ncols == 1:
+            axs = [axs]
+
+        image = sample['image'][rgb_indices].permute(1, 2, 0).numpy()
+        image = (image - image.min()) / (image.max() - image.min() + 1e-8)
```
adamjstewart, line 234: Use quantile_normalization

adamjstewart (changes requested):

PR #3549, opened 2026-03-26
test: enable branch coverage in pytest

nuglifeleoji:
Enable branch coverage to ensure if-statements are tested for both True and False conditions.

This is the first part of #2501. The configuration change is minimal, but will reveal missing branch coverage that needs to be addressed in follow-up PRs.

Reference: https://coverage.readthedocs.io/en/latest/branch.html

Part of #2501

PR #3559, opened 2026-03-27
test: add branch coverage for get_random_bounding_box

nuglifeleoji:
Add tests for get_random_bounding_box function.

Covers branches:
- float size/res input
- tuple size/res input
- with explicit generator
- without generator (None)

Relates to #3549


adamjstewart:
This function is being deprecated in #3552 and will be removed in the 1.0 release, I'm not sure if it's worth adding more tests for this. Regardless, they would need to be integrated into the existing test file if we want them.

adamjstewart:
Is this PR still a WIP or can we close?

PR #3560, opened 2026-03-27
test: add branch coverage for collate_fn_detection

nuglifeleoji:
Add branch coverage tests for collate_fn_detection in datamodules/utils.py.

Tests cover all branches:
- image only (no optional fields)
- with bbox_xyxy and label
- with bbox_xyxy but no label (triggers auto-label elif branch)
- with mask

Relates to #3549


tests/samplers/test_utils_random_bbox.py
ashnair1, line 1: I'm not sure why the float and tuple size tests are needed as these are already covered by the `test_get_random_bounding_box`

That aside, `get_random_bounding_box` has recently been deprecated (see [here](https://github.com/torchgeo/torchgeo/blob/2f3172bd32bd0262dae3f3cc7cd90802cbfdd648/torchgeo/samplers/utils.py#L40)) so I don't see a point in adding more tests for it now. This file `test_utils_random_bbox.py` can probably be removed.

adamjstewart:
Is this PR still a WIP or can we close?

PR #3563, opened 2026-03-27
Deps: torchmetrics[detection] required for pycocotools dep

adamjstewart:
pycocotools was removed in #3500. However, our Object Detection and Instance Segmentation trainers rely on pycocotools for mAP calculation. As a result, these tests were being skipped in CI and coverage dropped to _only_ 99%. This should get our coverage back up to 100%.

EDIT: turns out that 1% coverage was enough to hide multiple bugs in VHR-10, fixed now.

### DIFF
--- a/torchgeo/datasets/vhr10.py
+++ b/torchgeo/datasets/vhr10.py
@@ -341,7 +341,6 @@ class VHR10(NonGeoDataset):
             scores = sample['prediction_score']
             labels = sample['prediction_label']
             boxes = sample['prediction_bbox_xyxy']
-            masks = sample['prediction_mask']
             for i in range(len(labels)):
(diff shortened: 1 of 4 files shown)


torchgeo/datasets/vhr10.py
```diff
@@ -391,6 +391,4 @@ def plot(
         if suptitle is not None:
             fig.suptitle(suptitle)
 
-        fig.tight_layout()
```
ashnair1, line 394: Why remove `fig.tight_layout()`? Was this causing issues with plotting?

adamjstewart: Just a warning message that it couldn't fit the plot.

PR #3568, opened 2026-03-30
Add GitHub markdown alert styling for changelog

ashnair1:
Follow up to #3425. This PR adds colored background for notes/warnings

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -19,3 +19,61 @@ html {
 .nboutput .prompt {
   display: none;
 }
+
+/* GitHub Markdown Alert Styling for Changelog */
+.markdown-alert {
+  padding: 8px 16px;
+  margin-bottom: 16px;
+  border-left: 4px solid #d0d7de;
+  background-color: #f6f8fa;
+}
+
+.markdown-alert > :first-child { margin-top: 0; }
+.markdown-alert > :last-child { margin-bottom: 0; }
+
+.markdown-alert-title {
+  display: flex;
+  font-weight: 500;
+  align-items: center;
+  line-height: 1;
+  margin-bottom: 4px;
+}
+
+.markdown-alert-title svg {
+  margin-right: 8px;
+}
+
+/* Note - Blue */
+.markdown-alert-note {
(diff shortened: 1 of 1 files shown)


adamjstewart:
Does not look good on dark backgrounds:
Screenshot 2026-03-30 at 12 33 22 PM
Is this also something we could contribute upstream?

ashnair1:
> Does not look good on dark backgrounds: Screenshot 2026-03-30 at 12 33 22 PM Is this also something we could contribute upstream?

Should look good now. 

image

adamjstewart:
> Is this also something we could contribute upstream?

What I'm thinking is that Sphinx (and likely pydata-sphinx-theme) already has styling for all valid docutils [admonitions](https://docutils.sourceforge.io/docs/ref/rst/directives.html#admonitions). And all GitHub markdown [alerts](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts) have corresponding admonitions. So in sphinx-github-changelog, we could map between GitHub alerts and Sphinx admonitions. Then we don't need to maintain any of these lines of code. What do you think?

ashnair1:
> > Is this also something we could contribute upstream?
> 
> What I'm thinking is that Sphinx (and likely pydata-sphinx-theme) already has styling for all valid docutils [admonitions](https://docutils.sourceforge.io/docs/ref/rst/directives.html#admonitions). And all GitHub markdown [alerts](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts) have corresponding admonitions. So in sphinx-github-changelog, we could map between GitHub alerts and Sphinx admonitions. Then we don't need to maintain any of these lines of code. What do you think?

Fair point and that mapping should work. I'll look into pushing this upstream and revert when I have an update.

ashnair1:
ewjoachim/sphinx-github-changelog#189 should address this

ashnair1:
Latest release of `sphinx-github-changelog` now supports this. Closing in favor of #3708

Issue #3042, opened 2025-10-16
Convert time series datasets


yichiac:
We have completed converting one dataset (PASTIS) to support time series in #2382. This issue lists and tracks the remaining datasets that need to be converted to support time series. If anyone wants to help with this effort, the basic goal is that all time series datasets return images of the form TxCxHxW or sequences of the form TxC.

### Benchmark datasets

- [x] Benin Cashew Plantations
- [x] BioMassters
- [x] BRIGHT: #2862
- [x] CaBuAr: #2863
- [x] ChaBuD: #2878
- [x] Copernicus-Bench
- [ ] CropHarvest: #3190
- [x] DigitalTyphoon
- [x] Kenya Crop Type
- [x] LEVIR-CD: #2879
- [x] OSCD: #2422
- [x] PASTIS: #2921
- [ ] QuakeSet: @DarthReca 
- [x] Rwanda Field Boundary
- [x] SeasoNet: #3371 
- [ ] SKIPP'D: #3479
- [ ] South Africa Crop Type: #3368 
- [x] Substation: #3369
- [x] SustainBench Crop Yield
- [ ] Western USA Live Fuel Moisture
- [x] xBD
- [x] ZueriCrop

### Pretraining datasets

- [ ] CopernicusPretrain: ignore for now, still need to figure out how to handle this in WebDataset
- [ ] SeasonalContrastS2: #3372 
- [ ] SSL4EO-S12 #3372 
- [ ] SSL4EO-L: #3372

DarthReca:
@adamjstewart @yichiac I can take care of QuakeSet if you are not actively working on it

adamjstewart:
Note that for QuakeSet, we may need to add new VideoClassification and VideoRegression (names TBD) trainers at the same time in order to test the new datamodules.

madhavcodez:
Opened #3770 converting **Western USA Live Fuel Moisture** to the T x C time-series format (4 time steps by 34 variables), following the pattern in #2921. Static lat/lon are returned under their own keys. Happy to adjust the sample-key naming or add a datamodule if you'd prefer.

PR #3591, opened 2026-04-04
SpatioTemporalClassifier, SpatioTemporalRegressor and refactored QuakeSet

DarthReca:
This is the draft implementation of the SpatioTemporalClassifier and the refactored QuakeSet. The regressor is currently WIP.

I have taken inspiration from #3280 but with some modifications (I still need to fully align the code with this pull request).

This is linked to both #3042 and #2382.


tests/tasks/test_spationtemporal_classification.py
adamjstewart, line 1: Typo in filename

tests/conf/quakeset_classification.yaml
adamjstewart, line 1: Maybe call this `quakeset_classification.yaml` since we will also add a regression version in the future?

torchgeo/datamodules/quakeset.py
```diff
@@ -34,9 +34,14 @@ def __init__(
         """
         super().__init__(QuakeSet, batch_size, num_workers, **kwargs)
         self.train_aug = K.AugmentationSequential(
-            K.Normalize(mean=self.mean, std=self.std),
```
adamjstewart, line 37: We still want normalization during training time, that should stay

torchgeo/trainers/spatiotemporal_classification.py
```diff
@@ -0,0 +1,242 @@
...
+
+from ..datamodules import BaseDataModule
+from ..datasets import RGBBandsMissingError, unbind_samples
+from ..datasets.utils import Sample
+from ..models import ConvLSTM
+from .base import BaseTask
+from .mixins import ClassificationMixin
+
+
+class _ConvLSTMClassifier(nn.Module):
```
adamjstewart: This is now removed from the other PR, let's wait until that gets merged and then update

DarthReca: I have also made a change (uncomitted for now) using the TIMM [ClassifierHead](https://github.com/huggingface/pytorch-image-models/blob/main/timm/layers/classifier.py#L77); probably it's more stable than reinventing the wheel

adamjstewart:
FYI, https://github.com/torchgeo/torchgeo/pull/3671 is finally merged! Can copy-n-paste this as the foundation for this PR.

torchgeo/trainers/spatiotemporal_classification.py
DarthReca: I have made a small change to the ConvLSTM head. If you believe it is fine or if you suggest any changes, I will update the tests, etc., once stable.

adamjstewart: Does this work for regression too?

DarthReca: Yep, it is simply the standard head builder of timm that can be customized a bit (for example, using a conv or linear). There are also some other versions based on MLP if I see well in the timm file.

DarthReca:
Mostly done, I need to investigate what is wrong with ReadTheDocs

torchgeo/trainers/spatiotemporal_regression.py
```diff
@@ -0,0 +1,182 @@
...
+import torch.nn as nn
+from torch import Tensor
+from torchmetrics import MeanAbsoluteError, MeanSquaredError, MetricCollection
+
+from ..datasets.utils import Sample
+from ..models import ConvLSTM
+from .base import BaseTask
+
+
+class SpatioTemporalRegressionTask(BaseTask):
```
robmarkcole: Just to confirm, this differs from SpatioTemporalPixelwiseRegressionTask in https://github.com/torchgeo/torchgeo/pull/3086 right? I wonder if we can make this clearer, i.e. this task is for regressing a single patch level value

torchgeo/tasks/spatiotemporal_regression.py
```diff
@@ -0,0 +1,182 @@
...
+        num_filters: int = 3,
+        loss: Literal['mse', 'mae'] = 'mse',
+        lr: float = 1e-3,
+        patience: int = 10,
+        **kwargs: Any,
+    ) -> None:
+        """Initialize a new SpatioTemporalRegressionTask instance.
+
+        Args:
+            model: Video model name. Only ``'convlstm'`` is currently supported.
```
robmarkcole: @adamjstewart didn't we decide agains this `video` naming approach?

torchgeo/trainers/spatiotemporal_regression.py
```diff
@@ -0,0 +1,182 @@
...
+
+        * :class:`~torchmetrics.MeanSquaredError`: The average of the squared
+        differences between the predicted and actual values (MSE) and its
+        square root (RMSE). Lower values are better.
+        * :class:`~torchmetrics.MeanAbsoluteError`: The average of the absolute
+        differences between the predicted and actual values (MAE).
+        Lower values are better.
+        """
+        metrics = MetricCollection(
+            # https://github.com/astral-sh/ty/issues/2985
```
robmarkcole: This issue is closed so why is it referenced?

torchgeo/models/convlstm.py
```diff
@@ -134,6 +136,8 @@ def __init__(
             num_classes: Optional number of segmentation classes for an attached
                 prediction head.
             head_kernel_size: Kernel size for the optional segmentation head.
+            convolutional_head: If ``True``, uses a convolutional head for
+                segmentation instead of a fully connected head.
```
robmarkcole: `a fully connected head` is the appropriate choice for image-level prediction? could be a useful comment to include

torchgeo/tasks/spatiotemporal_classification.py
```diff
@@ -0,0 +1,167 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+
+"""Trainers for spatiotemporal regression."""
```
robmarkcole: I think it would be useful to clarify 'at image level' to avoid confusion with pixel level tasks

robmarkcole:
This LGTM - would be nice to see a full training result with this new trainer on the dataset?

torchgeo/tasks/spatiotemporal_regression.py
adamjstewart: Yes, we typically refer to "video" input as "spatiotemporal" input. As long as it isn't a user-facing class/parameter name, it doesn't matter that much though.

torchgeo/trainers/spatiotemporal_regression.py
adamjstewart: Probably copy-n-paste. Shouldn't need to add any metrics if you inherit from RegressionMixin.

torchgeo/trainers/spatiotemporal_regression.py
adamjstewart: #3749 will probably help make this more clear.

DarthReca: Yes, correct. They are both creating a single real or categorical value per timeseries

torchgeo/tasks/spatiotemporal_regression.py
DarthReca: Let me copy the new naming from the accepted MR (probably this was taken from the old version)

torchgeo/trainers/spatiotemporal_regression.py
DarthReca: Ok, thanks. I have missed the RegressionMixin

adamjstewart: That should also fix the RtD test failure?

adamjstewart:
Can you rebase? We moved all trainers to tasks.

DarthReca:
Yup, after coming back from vacation I will finish this PR

robmarkcole:
I think there could be some confusion between SpatioTemporalPixelwiseRegression added in https://github.com/torchgeo/torchgeo/pull/3086 and SpatioTemporalRegression. Perhaps SpatioTemporalPatchwiseRegression would be clearer @adamjstewart ?

adamjstewart:
There hasn't been confusion between our existing Regression and PixelwiseRegression has there?

robmarkcole:
@adamjstewart I guess not, as you would know if there has :-)

adamjstewart:
I would know if someone complained to me. That's why we're planning a user survey, to get feedback on stuff like this.

torchgeo/trainers/spatiotemporal_regression.py
DarthReca: seems solved

DarthReca:
@adamjstewart @robmarkcole Meanwhile, I am testing the training, I am not sure if the update to convlstm sounds to you and the uncovered line can be removed or exists for a reason.

Issue #3325, opened 2026-01-24
Rasteret-backed RasterDataset


isaaccorley:
### Summary

Rasteret appears to improve reads from COGs over rasterio by a significant percentage by being smart about caching the headers. 

Cross-posting from https://github.com/terrafloww/rasteret/issues/1

print-sid8:
Hi @isaaccorley @adamjstewart, thanks for this cross-post.

Sorry about the delayed reply, cause we were working on a new release of Rasteret v0.3.0, now that its out, here's my update to TorchGeo team. 

License is now **Apache-2.0**, and v0.3.0 is a substantial rewrite.

**Repo**: https://github.com/terrafloww/rasteret | **Docs**: https://terrafloww.github.io/rasteret 

---

## What changed since the prototype

Isaac may remember the earlier codebase — that version queried a STAC API, built a STAC-GeoParquet with cached TIFF header columns, and used that to async-read COGs into xarray or GeoDataFrame. That core idea (cache headers, skip GDAL in the read path) hasn't changed.

What changed in v0.3.0 is that the STAC-specific GeoParquet has generalized into a **Collection** - a Parquet file with a minimal schema (`id`, `datetime`, `geometry`, `assets`) plus optional COG acceleration columns. STAC is one way to build a Collection, but `build_from_table()` accepts any Parquet/GeoParquet with those four columns (Source Cooperative exports, academic lab registries, custom catalogs). A **dataset catalog** (`build("earthsearch/sentinel-2-l2a", ...)`) with 12 pre-registered entries abstracts the source entirely. The I/O layer added Obstore dependency for multi-cloud: S3, Azure, GCS, the core IO planning and math is still Rasteret, we just need AioHTTP to fetch tiff data, so Obstore is not dependency we cannot exist without, its just convenience for multi-cloud. 

Because a Collection is Parquet, people can filter, join, enrich with arbitrary columns, version, and share it with standard Arrow tools, while pixels stay in COGs and stream on demand via byte-range reads.

---

## TorchGeo integration

We chose to subclass `GeoDataset` rather than `RasterDataset` because Rasteret's architecture differs at every step -remote COGs via async HTTP, metadata from the Collection's Parquet index, no GDAL file handles.

[...]

PR #3597, opened 2026-04-07
'Rasteret' backed RasterDataset

print-sid8:
## Summary

Following up on this open [issue 3325](https://github.com/torchgeo/torchgeo/issues/3325) for Rasteret by @isaaccorley

to create `torchgeo.datasets.RasteretDataset` to hopefully benefit TorchGeo users with much faster IO and throughput in training and inference workloads.

## Intended Mental Model

The intended usage is:

1. Build or Load a Rasteret `Collection`
2. Construct `RasteretDataset(collection=..., bands=...)`
3. Continue with standard TorchGeo usage

After dataset construction, normal TorchGeo workflows stay the same.
The difference is primarily **how raster images are prepared and read**, not how TorchGeo training code is written downstream.

## Why This Exists

TorchGeo's `RasterDataset` is filepath-first: it globs the filesystem, builds an rtree index, and calls `rasterio.open()` per scene at chip-read time. On cloud-hosted COGs that means an HTTP round-trip for every scene header, every chip.

Rasteret is Collection-table-first: COG/TIFF file headers are parsed once at build time and stored in a Parquet table. All TorchGeo reads can then use the `Collection` to fetch pixel bytes fast and async via Rasteret's IO path.

`RasteretDataset` is a thin bridge that gives TorchGeo users access to that read path without changing anything about how they write TorchGeo code.

## What This PR Adds

- `torchgeo/datasets/rasteret.py` — `RasteretDataset(RasterDataset)`
- `tests/datasets/test_rasteret.py` — 14 tests
- `docs/api/datasets/rasteret.rst` — reference page (`autoclass`)
- `experiments/rasteret/benchmark.py` — reproduces the numbers below

## Usage

```python
import rasteret
from torchgeo.datasets import RasteretDataset

collection = rasteret.build(
    "earthsearch/sentinel-2-l2a",
    name="s2-bangalore",
    bbox=(77.5, 12.9, 77.7, 13.1),
    date_range=("2024-01-01", "2024-06-30"),
)

spring = collection.subset(
    date_range=("2024-03-01", "2024-05-31"),
    cloud_cover_lt=15,
)

dataset = RasteretDataset(collection=spring, bands=["B04", "B03", "B02"])
sampler = RandomGeoSampler(dataset, size=256, length=100)
```

Time-series mode — all matched scenes stacked as `[T, C, H, W]`, all reads
fired concurrently:

```python
dataset = RasteretDataset(collection=collection, bands=["B04"], time_series=True)
# sample["image"].shape == [T, 1, 256, 256]
```

Pre-built read-ready Rasteret collections also work directly:

```python
# AlphaEarth Foundation Embeddings — no build step needed
collection = rasteret.load("aef/v1-annual")
dataset = RasteretDataset(collection=collection, bands=["A00", "A63"])
```

For multiprocessing (`DataLoader(num_workers>0)`), pass a collection opened with
`rasteret.load(...)` (a persisted collection); an in-memory collection cannot be
sent to worker processes.

## Benchmark

Sentinel-2 L2A, `B04`, single MGRS tile (43PGP), `chip=256`, `batch=4`.
Python 3.14, `rasteret 0.4.1`, `rasterio 1.4.4`, on an 8 vCPU AWS `us-west-2` machine.

**Methodology (fair + cold).** Each scenario reads with a full temporal-range query
so *both* datasets load **all N overlapping scenes** (equal work, matching output
shapes) — otherwise a sampler's instant temporal slice makes date-aware Rasteret
select fewer scenes than date-blind `RasterDataset`. Each scenario runs in its own
fresh process so torchgeo's GDAL/`vsicurl` cache is empty and it opens every COG for
the first time (the real per-epoch cost). Rasteret's `Collection` is prebuilt; its
build is a one-time step and is **excluded** from the numbers (Rasteret `init` below
is just loading the prebuilt index). `scenes` is the count both sides read.

### 1. Same-cloud — Earth Search (`sentinel-cogs` S3, `us-west-2`) from a `us-west-2` box, 29 scenes

| Scenario | Rasteret (init + first batch) | TorchGeo (init + first batch) | Speedup |
|---|---:|---:|---:|
| spatial / random | 81 ms + 2892 ms | 5254 ms + 17410 ms | **7.6×** |
| spatial / grid | 19 ms + 1204 ms | 4059 ms + 2008 ms | **5.0×** |
| timeseries / random | 19 ms + 2134 ms | 2848 ms + 11608 ms | **6.7×** |
| timeseries / grid | 18 ms + 865 ms | 2595 ms + 1708 ms | **4.9×** |

### 2. Cross-cloud — Planetary Computer (Azure COGs) from a `us-west-2` box, 56 scenes

| Scenario | Rasteret (init + first batch) | TorchGeo (init + first batch) | Speedup |
|---|---:|---:|---:|
| spatial / random | 34 ms + 14213 ms | 29576 ms + 51616 ms | **5.7×** |
| spatial / grid | 30 ms + 12381 ms | 29812 ms + 27183 ms | **4.6×** |
| timeseries / random | 61 ms + 12866 ms | 28769 ms + 51253 ms | **6.2×** |
| timeseries / grid | 30 ms + 12606 ms | 28584 ms + 26817 ms | **4.4×** |

Time-series uses `read_window(group_by="id")`, one timestep per scene, matching
`RasterDataset(time_series=True)` (shapes match). Speedups are largest same-cloud;
cross-cloud (Azure↔AWS) inflates both sides' per-read latency, compressing the ratio.

More scenarios (multi-AOI, cross-UTM-zone): [Rasteret benchmark notebook](https://terrafloww.github.io/rasteret/tutorials/05_torchgeo_benchmark_rasteret_vs_rasterio/).

## What This PR Does Not Do

- Does **not** touch rasterio/GDAL anywhere in TorchGeo
- Does **not** auto-convert existing datasets

## Note about CRS

Works the same as `RasterDataset` — pick an output CRS, and all chips come
out in that CRS regardless of what projection the underlying COGs are in.
Cross-CRS collections (e.g. Sentinel-2 spanning multiple UTM zones) just work:
Rasteret reprojects each chip to the output grid at read time, the same way
rasterio does today.

The one difference: pass `crs=` when you create the dataset rather than
reassigning it afterwards. That's it.

## What Changed Since April 2026

The original PR was built against Rasteret 0.3.7, which shipped a built-in
TorchGeo adapter (`to_torchgeo_dataset()`). That adapter has since been removed
in favor of a small, general integration surface on `Collection`. This rewrite:

- Keeps `RasteretDataset` as a `RasterDataset` subclass, so it drops into
  standard TorchGeo raster workflows (samplers, datamodules, `&`/`|`).
- Builds the spatial index from `collection.footprints()` and reads the
  sampling resolution from `collection.native_res()` — no raster is opened at
  construction.
- Reads pixels via `collection.read_window()`, using `group_by="id"` for
  `time_series=True` so it stacks one timestep per scene, matching
  `RasterDataset(time_series=True)`.

> Requires **rasteret >= 0.4.1**, the first release with the integration
> boundary `footprints()`, `native_res()`, `read_window()`, and the
> `read_window(group_by="id")` time-series mode (pinned in pyproject + CI).

## AI Usage Disclosure


print-sid8:
Can someone please rerun the git actions? It was my mistake not publishing 0.3.7 Rasteret version to pypi. It should not fail on that step anymore now. 

Thanks.

.github/workflows/tests.yaml
adamjstewart, line 1: Undo changes to this file and instead make rasteret dep conditional on python version in pyproject.toml

docs/api/datasets/images.csv
```diff
@@ -5,4 +5,5 @@ Dataset,Type,Source,Size (px),Resolution (m),License
 :ref:`NAIP`,Imagery,Aerial,"6,100x7,600",0.3--2,public domain
 :ref:`OpenAerialMap`,Imagery,Aerial,"256x256 OR 512x512",0.03--50,CC-BY-4.0
 :ref:`PRISMA`,Imagery,PRISMA,512x512,5--30,
+:ref:`Rasteret`,Imagery,Cloud-Optimized GeoTIFFs (COGs),variable,variable,depends on underlying data source
```
adamjstewart: No really an image dataset, maybe a base class?

experiments/rasteret/.gitignore
adamjstewart, line 1: Remove the experiments directory

tests/datasets/test_rasteret.py
```diff
@@ -0,0 +1,353 @@
...
+pa = pytest.importorskip('pyarrow')
+pads = pytest.importorskip('pyarrow.dataset')
+gpd = pytest.importorskip('geopandas')
+shapely_geometry = pytest.importorskip('shapely.geometry')
+pytest.importorskip('rasteret')
+
+if TYPE_CHECKING:
+    from rasteret import Collection
+
+Collection = import_module('rasteret').Collection
```
adamjstewart: Could be combined into one import

tests/datasets/test_rasteret.py
```diff
@@ -0,0 +1,353 @@
...
+
+
+def _non_epsg_crs() -> CRS:
```
adamjstewart: Why a function instead of a variable?

tests/datasets/test_rasteret.py
```diff
@@ -0,0 +1,353 @@
...
+            raise KeyError('assets')
+
+        monkeypatch.setattr(ds._collection, 'to_table', _no_assets_column)
+        assert ds.files == []
+
+    def test_files_skips_rows_without_the_band(
+        self, collection: Collection, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        """files skips rows whose assets are null or lack an href for the band."""
+        from torchgeo.datasets import RasteretDataset
```
adamjstewart: All imports should be moved to the top of the file

tests/datasets/test_rasteret.py
adamjstewart, line 1: Almost double the length of our RasterDataset tests. RasterDataset is far more complex and supports hundreds of file formats. Can we make this significantly simpler? We already have plenty of raster test data you can reuse instead of creating new datasets.

torchgeo/datasets/rasteret.py
```diff
@@ -0,0 +1,294 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+
+"""Rasteret dataset."""
+
+from __future__ import annotations
```
adamjstewart: Why?

torchgeo/datasets/rasteret.py
```diff
@@ -0,0 +1,294 @@
...
+
+    .. warning::
+       This dataset is experimental and its API may change.
+
+    .. versionadded:: 0.10
+    """
+
+    def __init__(
+        self,
+        collection: Any,
```
adamjstewart: Why Any, can we be more specific?

torchgeo/datasets/rasteret.py
```diff
@@ -0,0 +1,294 @@
...
+        bands: Sequence[str],
```
adamjstewart: Can we make this optional and load all bands by default like RasterDataset?

torchgeo/datasets/rasteret.py
```diff
@@ -0,0 +1,294 @@
...
+        crs: CRS | None = None,
+        res: float | tuple[float, float] | None = None,
+        transforms: Callable[[Sample], Sample] | None = None,
+        cache: bool = True,
+        time_series: bool = False,
+        is_image: bool = True,
+        max_concurrent: int = 50,
+        backend: Any = None,
```
adamjstewart: Would like to always avoid Any

torchgeo/datasets/rasteret.py
```diff
@@ -0,0 +1,294 @@
...
+        """
+        lazy_import('rasteret')
+
+        if not bands:
+            raise ValueError('At least one band is required')
+        for method in ('footprints', 'native_res', 'read_window'):
+            if not callable(getattr(collection, method, None)):
+                raise TypeError(
+                    f'collection must be a rasteret.Collection exposing {method}(...)'
+                )
```
adamjstewart: If correctly typed, the type checker can handle this part

torchgeo/datasets/rasteret.py
```diff
@@ -0,0 +1,294 @@
...
+        self.bands = tuple(bands)
+        self.all_bands = tuple(bands)
+        self.transforms = transforms
+        self.cache = cache
+        self.time_series = time_series
+        self.is_image = is_image
+        self.separate_files = False
+        self.band_indexes = None
+        self._max_concurrent = max_concurrent
+        self._backend = backend
```
adamjstewart: Why are some underscore and some not?

torchgeo/datasets/rasteret.py
```diff
@@ -0,0 +1,294 @@
...
+        footprints = collection.footprints(crs=target_crs, band=self.bands[0])
+        if footprints.empty:
+            raise ValueError('Rasteret collection produced no footprints')
+        if footprints.crs is None:
+            raise ValueError('Collection has mixed CRS; pass crs= to unify.')
+        epsg = CRS.from_user_input(footprints.crs).to_epsg()
+        if epsg is None:
+            raise ValueError(
+                'Collection CRS is not EPSG-resolvable; pass an EPSG crs=.'
+            )
```
adamjstewart: I dislike defensive coding, if you make the code simple enough, none of this is needed

torchgeo/datasets/rasteret.py
```diff
@@ -0,0 +1,294 @@
...
+            AttributeError: Always raised. Rasteret binds the read-time CRS at
+                construction; use ``RasteretDataset(..., crs=...)`` instead.
+        """
+        raise AttributeError(
+            'RasteretDataset CRS is fixed after construction; '
+            'create a new dataset with crs=...'
+        )
+
+    @property
+    def res(self) -> tuple[float, float]:
```
adamjstewart: A lot of this is simply copy-n-pasted from RasterDataset, no need to duplicate if nothing changed

torchgeo/datasets/rasteret.py
```diff
@@ -0,0 +1,294 @@
...
+        for assets in table['assets'].to_pylist():
+            if not isinstance(assets, dict):
+                continue
+            for band in self.bands:
+                asset = assets.get(band)
+                if isinstance(asset, dict) and asset.get('href'):
+                    hrefs.append(str(asset['href']))
+        return list(dict.fromkeys(hrefs))
+
+    def close(self) -> None:
```
adamjstewart: Will this ever be used?

torchgeo/datasets/rasteret.py
```diff
@@ -0,0 +1,294 @@
...
+        """Release Rasteret background resources if the collection supports it."""
+        close = getattr(self._collection, 'close', None)
+        if callable(close):
+            close()
```
adamjstewart: Defensive coding

pyproject.toml
```diff
@@ -134,6 +131,12 @@ models = [
     # olmoearth-pretrain-minimal 0.0.2+ required for normalization support
     "olmoearth-pretrain-minimal>=0.0.2",
 ]
+rasteret = [
```
adamjstewart: Add to the datasets extra, no reason to have its own extra

adamjstewart (changes requested):

tests/datasets/test_rasteret.py
print-sid8: Thanks for the review, yes much of this is AI slop i didnt check on. Have simplified it a lot myself.

.github/workflows/tests.yaml
print-sid8: Done

print-sid8:
@adamjstewart 

thanks for your time on all the reviews. ive tried to resolve all of them. CI is failing ill look into that. 

But one open question remains and i didnt want to make a decision by myself . its on this point/question you made  -
> No really an image dataset, maybe a base class?

You are right and I agree its hard to put `RasteretDataset` as image dataset coz its source, size, and resolution are all "whatever collection you hand it," so it's an access mechanism, not a dataset in the Landsat/NAIP sense. 

The way I think about it, it's the rasteret-backed sibling of `RasterDataset`: same job ie., turn a set of rasters into a sampleable GeoDataset, different input (a Collection index instead of filesystem paths)

So where this lives might set a precedent. 

If "dataset backed by an external index/reader/engine" becomes its own category, RasteretDataset is probably the first of a few (`STAC-geoparquet`, other cloud-native backends), and it'd be cleaner to give that family a home now than to retrofit it later? 

I've removed the line out of the imagery table csv for now. 

Happy to go whichever way you think the taxonomy should grow. 

Other maintainers could also comment on this @calebrob6 @robmarkcole @isaaccorley @rosielickorish

torchgeo/datasets/rasteret.py
print-sid8: Done. bands is now optional and defaults to every band in the collection (collection.bands), mirroring RasterDataset's bands or self.all_bands. but can be overriden. 

  - RasteretDataset(collection) → all bands (collection.bands)
  - RasteretDataset(collection, bands=['B04', 'B03', 'B02']) → exactly those three, in that order

experiments/rasteret/.gitignore
print-sid8: done

adamjstewart:
If it's a sibling to RasterDataset, we can put it in the same place in the docs. Long-term, I'm hoping to refactor more of our base classes to be backend-agnostic. This includes spatiotemporal indexing backends (e.g., geopandas, STAC, sedona) and I/O backends (e.g., rasterio, rasteret, xarray). However, this is a huge refactor, so I'm not sure if this will happen in 1.0 or 2.0.

print-sid8:
Got it. I would love to be involved in backend-agnostic part if the team picks it up. 

And for this ive simply added it just as a BaseClass, and removed the standalone docs piece i had before.

Do let me know if anything else is required to be reviewed or discussed. Will work on it.

PR #3611, opened 2026-04-12
Fix incorrect mask dimension handling in InriaAerialImageLabelingDataModule

ohhan777:
## Summary
Fixes incorrect handling of mask dimensions in `InriaAerialImageLabelingDataModule.on_after_batch_transfer`.

## Problem
The current implementation assumes that `mask.ndim == 3` corresponds to a batch_size=1 case. However, this is also true for standard batched masks with shape [B, H, W].

This leads to incorrect reshaping and breaks augmentation consistency.

## Solution
- Restrict special handling to true single-mask case ([H, W])
- Leave batched masks ([B, H, W]) unchanged

## Impact
- Prevents shape mismatch errors during augmentation
- Preserves existing behavior for valid batched inputs

## Notes
This change preserves the original docstring and keeps the modification minimal.


ohhan777:
Closing in favor of #3612  which includes a more complete fix with additional test coverage.

PR #3612, opened 2026-04-12
fix(inria): reorder augmentation pipeline and fix mask handling in on_after_batch_transfer

ohhan777:
## Summary

This PR fixes two bugs in `InriaAerialImageLabelingDataModule`:

1. **CUDA kernel error with `batch_size >= 8`** caused by applying spatial augmentations
   (`RandomHorizontalFlip`, `RandomVerticalFlip`) on full-resolution 5000x5000 images
   before cropping.
2. **Incorrect mask shape handling in `on_after_batch_transfer`** that silently corrupts
   mask dimensions when `batch_size > 1`.

### Problem

The Inria dataset contains 5000x5000 images. The current `train_aug` pipeline applies
`RandomHorizontalFlip` and `RandomVerticalFlip` on full-resolution GPU tensors before
cropping:

```python
# Current
    K.RandomHorizontalFlip(p=0.5),          # on 5000x5000
    K.RandomVerticalFlip(p=0.5),            # on 5000x5000
    K.RandomCrop(self.patch_size, ...),
    ...
)
```

When `batch_size >= 8`, kornia's internal `compute_transformation` CUDA kernel exceeds
its launch configuration limit on the large tensor (e.g. `(8, 3, 5000, 5000)`):

```
torch.AcceleratorError: CUDA error: invalid configuration argument
  File "kornia/augmentation/_2d/base.py", line 106, in generate_transformation_matrix
    trans_matrix_B = self.compute_transformation(in_tensor[to_apply], ...)
```

<details>
<summary>Minimal reproduction</summary>

```python
import torch

aug = K.AugmentationSequential(
    data_keys=None, keepdim=True,
)

# batch_size=4: OK
batch = {'image': torch.randn(4, 3, 5000, 5000, device='cuda'),
         'mask': torch.randint(0, 2, (4, 5000, 5000), device='cuda')}
aug(batch)  # works

# batch_size=8: CUDA error
batch = {'image': torch.randn(8, 3, 5000, 5000, device='cuda'),
         'mask': torch.randint(0, 2, (8, 5000, 5000), device='cuda')}
aug(batch)  # CUDA error: invalid configuration argument
```

Tested on NVIDIA H100 80GB HBM3, PyTorch 2.9.0+cu128, kornia 0.8.2.
</details>

### Fix

Move crop operations to the beginning of each pipeline:

```python
# After fix
    K.RandomCrop(self.patch_size, pad_if_needed=True),  # 5000x5000 -> patch_size
    K.RandomHorizontalFlip(p=0.5),                       # on patch_size
    K.RandomVerticalFlip(p=0.5),                          # on patch_size
    ...
)
```

The same reordering is applied to `aug` and `predict_aug` (`CenterCrop` before `Normalize`).

### Why this is safe

- `Normalize` is a per-pixel linear operation. Normalizing before or after cropping
  produces identical results for the cropped region.
- `RandomHorizontalFlip` / `RandomVerticalFlip` applied after a random crop provides
  equivalent augmentation diversity, since the crop location is already random.
- This is consistent with other torchgeo datamodules (Potsdam2D, Vaihingen2D,
  DeepGlobeLandCover, GID15, LEVIRCDPlus) which all crop before or immediately after
  normalization, without spatial transforms in between.

### Problem

The current code checks `batch['mask'].ndim == 3` and applies `unsqueeze(0)` +
`squeeze(dim=1)`:

```python
# Current
if 'mask' in batch and batch['mask'].ndim == 3:
    batch['mask'] = batch['mask'].unsqueeze(0)       # (B, H, W) -> (1, B, H, W)
    batch = super().on_after_batch_transfer(batch, dataloader_idx)
    batch['mask'] = batch['mask'].squeeze(dim=1)     # squeeze dim=1 (=B)
    return batch
```

For `batch_size=1`, the mask shape is `(1, H, W)` -> `(1, 1, H, W)` -> squeeze dim=1
-> `(1, H, W)`. This works by coincidence.

For `batch_size > 1` (e.g. 32), the mask shape is `(32, H, W)` -> `(1, 32, H, W)` ->
`squeeze(dim=1)` does nothing (dim=1 has size 32) -> `(1, 32, H, W)`. This causes a
shape mismatch between predictions `(B, C, H, W)` and target `(1, B, H, W)` during
loss computation or metric evaluation.

### Fix

The intent of this code is to handle the edge case where a **single sample** without a
batch dimension is loaded (mask shape `(H, W)` instead of `(B, H, W)`). The fix checks
for `ndim == 2` instead:

```python
# After fix
if 'mask' in batch and batch['mask'].ndim == 2:
    batch['mask'] = batch['mask'].unsqueeze(0)       # (H, W) -> (1, H, W)
    batch = super().on_after_batch_transfer(batch, dataloader_idx)
    batch['mask'] = batch['mask'].squeeze(0)          # (1, H, W) -> (H, W)
    return batch
```

This correctly handles the single-sample edge case while leaving normal batched masks
`(B, H, W)` untouched.

## Test

Full training run completed successfully with `batch_size=32` on NVIDIA H100:

```python
import lightning as L

datamodule = InriaAerialImageLabelingDataModule(
    root='./data/Inria', batch_size=32, num_workers=8, patch_size=512,
)
    model='unet', backbone='resnet34', weights=True,
    in_channels=3, task='binary', loss='bce', lr=1e-3,
)
trainer = L.Trainer(
    accelerator='gpu', devices=1, max_epochs=10, precision='bf16-mixed',
)
# Trainer.fit stopped: max_epochs=10 reached.
```

## Environment

- torchgeo 0.7.0 / main branch (both affected)
- PyTorch 2.9.0+cu128
- kornia 0.8.2
- Lightning 2.6.1
- GPU: NVIDIA H100 80GB HBM3


adamjstewart (changes requested):
Thanks for the PR! As discussed in #3535, the correct fix is to delete `on_after_batch_transfer`. Can you make that change here or in a separate PR?

ohhan777:
Done — removed on_after_batch_transfer entirely and updated the test accordingly. All tests pass with both batch_size=1 and batch_size>1.

adamjstewart:
Can you also undo all other changes?

ohhan777:
Done — reverted test changes. The PR now only contains changes to torchgeo/datamodules/inria.py:                                                                                                                                                                        
 1. Reorder augmentation pipeline (crop before normalize/flip)                                                                                                                                                                                                         
 2. Delete on_after_batch_transfer

adamjstewart:
Actually, it looks like this is a duplicate of #3538 as well.

PR #3619, opened 2026-04-15
Air Quality dataset

f-schi:
This PR takes over PR #3255 from a different fork.

"This PR adds the Air Quality time series data set from the [UCI Machine Learning Repository](https://archive.ics.uci.edu/dataset/360/air+quality). This is split off from PR https://github.com/torchgeo/torchgeo/pull/2639 to be merged separately."

aq

## AI Usage Disclosure


nilsleh:
@f-schi Can you rebase the PR and resolve the conflicts, such that the tests can run?

torchgeo/datamodules/air_quality.py
```diff
@@ -0,0 +1,114 @@
...
+from .geo import NonGeoDataModule
+
+
+class AirQualityDataModule(NonGeoDataModule):
+    """LightningDataModule implementation for the AirQuality dataset.
+
+    Uses the user provided splits to divide the dataset into
+    train/val/test sets.
+
+    .. versionadded:: 0.9
```
nilsleh: I think this will be 0.10

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,162 @@
...
+    def plot(self, sample: Sample) -> Figure:
+        """Plot a sample from the dataset.
+
+        Args:
+            sample: a sample returned by :meth:`__getitem__`
+
+        Returns:
+            a matplotlib Figure with the plotted sample
+        """
+        import matplotlib.pyplot as plt
```
nilsleh: This should be an import at top of the file

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,162 @@
...
+            df['Date'] + ' ' + df['Time'],
+            format='%m/%d/%Y %H:%M:%S',  # month/day/year
+        )
+        df.drop(columns=['Date', 'Time'], inplace=True)
+
+        # Convert datetime64 to float (Unix timestamp in seconds) so it can become a Tensor
+        df['datetime'] = df['datetime'].astype('int64') / 1e9
+
+        # Drop rows with any remaining NaNs
+        df.dropna(inplace=True)
```
nilsleh: If you are dropping NaNs here, is there ever a risk of an empt dataframe being returned that would cause issues for batches?

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,162 @@
...
+        """
+        past_targets = self.data.iloc[index : index + self.num_past_steps]
+        future_targets = self.data.iloc[
+            index + self.num_past_steps : index
+            + self.num_past_steps
+            + self.num_future_steps
+        ]
+
+        return {
+            'past_targets': torch.tensor(past_targets.values, dtype=torch.float32),
```
nilsleh: I find the naming a bit confusing, what are "past_targets"? I understand the task to be a forecasting task, from "historical" data to "future" data, but am then confused why past data would be a target?

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,162 @@
...
+    Dataset Features:
+
+    * hourly averaged sensor responses and reference analyzer ground truth over one year (2004-2005)
+    * has missing features
+
+    If you use this dataset in your research, please cite:
+
+    * https://doi.org/10.1016/J.SNB.2007.09.060
+
+    .. versionadded:: 0.9
```
nilsleh: also 0.10 here

nilsleh (changes requested):

torchgeo/datasets/air_quality.py
f-schi: Yes, in theory that’s possible as an edge case if too many rows are dropped, but in this dataset it’s unlikely given the structure. Still, I agree it’s not ideal to enforce that behavior here. I’d remove the dropna() for now, which also means we don’t impose any specific missing-data handling strategy inside the dataset class.

torchgeo/datasets/air_quality.py
f-schi: Yes, I agree. I inherited the terminology from the existing PR. Personally, I would prefer something like:
`x_context` for past observed values, `y_target` for future values to predict, and `y_pred` for the model output.
If that aligns better with the codebase, I’m happy to adjust.

nilsleh: Yes, I like that as well, but check with @adamjstewart he has opinions on the batch item naming

torchgeo/datasets/air_quality.py
nilsleh: Okay, yeah I get your point. I suppose there is arguments for both sides, just dropping all NaNs might also cause some silent bugs, but ideally, a user does data cleaning before starting a ML run anyways.

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,162 @@
...
+
+        # Combine Date and Time into a single numeric column
+        df['datetime'] = pd.to_datetime(
+            df['Date'] + ' ' + df['Time'],
+            format='%m/%d/%Y %H:%M:%S',  # month/day/year
+        )
+        df.drop(columns=['Date', 'Time'], inplace=True)
+
+        # Convert datetime64 to float (Unix timestamp in seconds) so it can become a Tensor
+        df['datetime'] = df['datetime'].astype('int64') / 1e9
```
nilsleh: Could make the above slight more succinct:

```python
df['datetime'] = pd.to_datetime(df['Date'] + ' ' + df['Time']).view('int64') / 1e9
df.drop(columns=['Date', 'Time'], inplace=True)
```

I think pandas is pretty good at figuring out the format, or is it ambiguous?

torchgeo/datasets/air_quality.py
adamjstewart: Unfortunately no standards from Kornia we can follow here, so I'm fine with whatever y'all decide. Could be inputs/targets, X/y, or whatever.

adamjstewart:
Seems like a bad rebase? Current PR deletes a lot of documentation.

adamjstewart:
Needs another rebase now

torchgeo/datasets/air_quality.py
f-schi: I added linear interpolation and removed the NMHC channel for now, since it consists mostly of NaN values. With the current dataset/datamodule design, existing NaN values could otherwise break both the sliding-window sampling and the normalization logic (mean/std propagation to NaNs).
So in practice one of the following is possible:
1) drop invalid windows / rows,
2) impute missing values,
3) implement explicit NaN-aware handling throughout sampling, normalization, and potentially the model itself.

I went with 2. here as compromise to keep the temporal structure intact.

adamjstewart (changes requested):
Needs a new line in `docs/api/datasets/benchmark.csv`.

We should add checksum support for `data.csv`.

Can you add an example plot of the real data to the PR description?

Does not yet have 100% test coverage, you may need to move the datamodule to the TemporalRegressionTask PR.

docs/api/datasets/benchmark.csv
```diff
@@ -1,6 +1,7 @@
 Dataset,Task,Source,# Samples,# Classes,Size (px),Resolution (m),Bands,License
 :ref:`ADVANCE`,C,"Google Earth, Freesound","5,075",13,512x512,0.5,RGB,CC-BY-4.0
 :ref:`AgriFieldNet`,S,Sentinel-2,"7,081",14,"256x256",10,MSI,CC-BY-4.0
+:ref:`Air Quality`,R,"Pirelli Labs multi-sensor device","9,358",-,-,-,"15 (sensor features)",CC-BY-4.0
```
adamjstewart: Confirmed the license from https://archive.ics.uci.edu/dataset/360/air+quality

adamjstewart (dismissed):
I think this is a good start. Features worth adding in folllow-up PRs:

* Ability to select which features you want, I've only ever used the GT features since they're more reliable
* Ability to plot future predictions, likely needed for the tutorial
* Inclusion of sin/cos of HOD and DOY, important input features for modeling, especially in a time-series paper

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,166 @@
...
+        self.data = self._load_data()
+        self.feature_names = [c for c in self.data.columns if c != 'datetime']
+
+    def __len__(self) -> int:
+        """Return the number of data points in the dataset.
+
+        Returns:
+            length of the dataset
+        """
+        return len(self.data) - (self.num_input_steps + self.num_target_steps)
```
nilsleh: I think this needs a +1 because 0 is also a valid start index right?

if len(data) = 10, num_input_steps = 3, and num_target_steps = 2, then each sample needs a window of length 5.

Valid start indices would be: 0, 1, 2, 3, 4, 5

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,166 @@
...
+            df = pd.read_csv(self.url, na_values=['-200'])
+
+        # Drop Date and Time, not yet using these inputs
+        df.drop(columns=['Date', 'Time'], inplace=True)
+
+        # Drop NMHC(GT) column which has mostly missing values
+        df.drop(columns=['NMHC(GT)'], inplace=True)
+
+        # Interpolate missing values
+        df.interpolate(inplace=True)
```
nilsleh: I think this should be documented somewhere because that is a pretty definitive preprocessing step that people should be aware of.  I think maybe in a public method like the init even.

nilsleh (dismissed):

adamjstewart (dismissed):

f-schi:
I just noticed the API CSV is (comma-delimited vs semicolons, empty rows in download version). The dataset currently only supports the API format therefore.

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,167 @@
...
+
+        Raises:
+            DatasetNotFoundError: If dataset is not found and *download* is False.
+        """
+        self.root = root
+        self.download = download
+        self.num_input_steps = num_input_steps
+        self.num_target_steps = num_target_steps
+        self.data = self._load_data()
+        self.feature_names = [c for c in self.data.columns if c != 'datetime']
```
f-schi: This line can be removed, as datetime column is deprecated.

adamjstewart: Please do, this is your PR lol.

adamjstewart:
> I just noticed the API CSV is (comma-delimited vs semicolons, empty rows in download version). The dataset currently only supports the API format therefore.

This is fine with me, just make a note of this in the class docstring.

adamjstewart (dismissed):

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,226 @@
...
+    def _parse_datetime(self, data: pd.DataFrame) -> None:
+        """Parse datetime columns into cyclical features.
+
+        Args:
+            data: Raw data.
+        """
+        if {'Date', 'Time'} <= set(data.columns):
+            dt = pd.to_datetime(data['Date'] + ' ' + data['Time']).dt
+            doy = 2 * np.pi * dt.dayofyear / 365.25  # ty: ignore[unsupported-operator]
+            hod = 2 * np.pi * dt.hour / 24  # ty: ignore[unsupported-operator]
```
adamjstewart, line 123: These are false-positives by ty, likely an issue in pandas-stubs, will try to remember to report later

torchgeo/datasets/air_quality.py
```diff
@@ -0,0 +1,226 @@
...
+            pass
+        elif self.download:
+            filepath = self.url
+        else:
+            raise DatasetNotFoundError(self)
+
+        # Load twice in case target_features is not a subset of input_features
+        kwargs = {'na_values': -200}
+        self.input_data = pd.read_csv(filepath, usecols=self.input_features, **kwargs)  # ty: ignore[no-matching-overload]
+        self.target_data = pd.read_csv(filepath, usecols=self.target_features, **kwargs)  # ty: ignore[no-matching-overload]
```
adamjstewart, line 144: These are false-positives by ty, likely an issue in pandas-stubs, will try to remember to report later

adamjstewart (approved):
For the sake of time (deadline tomorrow), let's merge once tests pass. There is a pretty good chance that I will end up moving a lot of the preprocessing (imputation, cyclic encoding) from the dataset to the datamodule, so that will hopefully take care of one of @nilsleh's concerns.

adamjstewart:
> I think this is a good start. Features worth adding in folllow-up PRs:
> 
> * Ability to select which features you want, I've only ever used the GT features since they're more reliable
> * Ability to plot future predictions, likely needed for the tutorial
> * Inclusion of sin/cos of HOD and DOY, important input features for modeling, especially in a time-series paper

These are all supported now 🎉

PR #3625, opened 2026-04-18
Add rf-detr to ObjectDetectionTask

### LINKED ISSUE #3422: New model: RF-DETR
### Summary

- Source: [roboflow/rf-detr](https://github.com/roboflow/rf-detr)
- Pypi: https://pypi.org/project/rfdetr/ 
- License: Apache-2.0

### Rationale

_No response_

### Implementation

_No response_

### Alternatives

_No response_

### Additional information

_No response_

robmarkcole:
Closes https://github.com/torchgeo/torchgeo/issues/3422

CC @isaaccorley @Borda


isaaccorley:
+1 for this

robmarkcole:
@adamjstewart CI is failing in the ubuntu-latest matrix because the 3.14t free-threaded job installs the models extra, which pulls in rfdetr -> supervision -> opencv-python. The locked `opencv-python==4.13.0.92` release is wheel-only and does not appear installable for the free-threaded interpreter, so `uv sync --locked --extra datasets --extra models --extra tests` fails on that matrix leg even though standard Linux wheels exist for regular CPython.

robmarkcole:
Training is proceeding on full DOTA dataset

image

image

robmarkcole:
Lots of experimentation with dependencies to try get around this `opencv-python==4.13.0.92` issue..

robmarkcole:
@adamjstewart the windows test fails with `Segmentation fault` - any ideas?

robmarkcole:
- Created https://github.com/roboflow/rf-detr/pull/999
- Be nice to support multispectral since https://github.com/roboflow/rf-detr/pull/180 (thanks @isaaccorley !)

.github/workflows/tests.yaml
```diff
@@ -18,7 +18,9 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ['3.12', '3.13', '3.14', '3.14t']
+        # NumPy does not yet publish cp314t wheels, so free-threaded Python 3.14
+        # falls back to an sdist build in CI.
```
adamjstewart: Not true? https://pypi.org/project/numpy/#files

robmarkcole: Removed the stale cp314t/NumPy comment

robmarkcole:
After much back and forth..
```
Run python -X faulthandler -u -c "print('import pytest', flush=True); import pytest; print('ok pytest', flush=True)"
  python -X faulthandler -u -c "print('import pytest', flush=True); import pytest; print('ok pytest', flush=True)"
  python -X faulthandler -u -c "print('import coverage', flush=True); import coverage; print('ok coverage', flush=True)"
  python -X faulthandler -u -c "print('import torchgeo', flush=True); import torchgeo; print('ok torchgeo', flush=True)"
  python -X faulthandler -u -c "print('import matplotlib', flush=True); import matplotlib; print('ok matplotlib', flush=True)"
  python -X faulthandler -u -c "print('import torch', flush=True); import torch; print('ok torch', torch.__version__, flush=True)"
  python -X faulthandler -u -c "print('import torchvision', flush=True); import torchvision; print('ok torchvision', torchvision.__version__, flush=True)"
  python -X faulthandler -u -c "print('import tests.conftest deps', flush=True); import matplotlib; import pytest; import torch; import torchvision; from pytest import MonkeyPatch; print('ok tests.conftest deps', flush=True)"
  python -X faulthandler -u -c "print('exec tests/conftest.py', flush=True); exec(compile(open('tests/conftest.py').read(), 'tests/conftest.py', 'exec')); print('ok tests/conftest.py', flush=True)"
  python -X faulthandler -u -m pytest --version
  set +e
  for target in \
    tests/conftest.py \
    tests/datamodules \
    tests/datasets \
    tests/losses \
    tests/models \
    tests/samplers \
    tests/trainers \
    tests/transforms \
    docs/tutorials
  do
    echo "::group::collect ${target}"
    PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -X faulthandler -m pytest -o addopts='' --collect-only -vv "${target}"
    echo "exit code: $?"
    echo "::endgroup::"
  done
  echo "::group::trace config"
  python -X faulthandler -m pytest --trace-config
  echo "exit code: $?"
  echo "::endgroup::"
  exit 1
  shell: C:\Program Files\Git\bin\bash.EXE --noprofile --norc -e -o pipefail {0}
  env:
    UV_TOOL_BIN_DIR: D:\a\_temp\uv-tool-bin-dir
    UV_TOOL_DIR: D:\a\_temp\uv-tool-dir
    UV_PYTHON_INSTALL_DIR: D:\a\_temp\uv-python-dir
    UV_PYTHON: 3.14
    VIRTUAL_ENV: D:\a\torchgeo\torchgeo\.venv
    UV_CACHE_DIR: D:\a\_temp\setup-uv-cache
import pytest
ok pytest
import coverage
ok coverage
import torchgeo
ok torchgeo
import matplotlib
<frozen importlib._bootstrap>:491: Warning: Numpy built with MINGW-W64 on Windows 64 bits is experimental, and only available for 
testing. You are advised not to use it for production. 

CRASHES ARE TO BE EXPECTED - PLEASE REPORT THEM TO NUMPY DEVELOPERS
Windows fatal exception: access violation

Current thread 0x000015d4 (most recent call first):
  File "<frozen importlib._bootstrap>", line 491 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 1053 in create_module
  File "<frozen importlib._bootstrap>", line 816 in module_from_spec
  File "<frozen importlib._bootstrap>", line 924 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1342 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1371 in _find_and_load
  File "D:\a\torchgeo\torchgeo\.venv\Lib\site-packages\numpy\_core\overrides.py", line 7 in <module>
  File "<frozen importlib._bootstrap>", line 491 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 759 in exec_module
  File "<frozen importlib._bootstrap>", line 938 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1342 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1371 in _find_and_load
  File "<frozen importlib._bootstrap>", line 491 in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1434 in _handle_fromlist
  File "D:\a\torchgeo\torchgeo\.venv\Lib\site-packages\numpy\_core\multiarray.py", line 10 in <module>
  File "<frozen importlib._bootstrap>", line 491 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 759 in exec_module
  File "<frozen importlib._bootstrap>", line 938 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1342 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1371 in _find_and_load
  File "<frozen importlib._bootstrap>", line 491 in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1434 in _handle_fromlist
  File "D:\a\torchgeo\torchgeo\.venv\Lib\site-packages\numpy\_core\__init__.py", line 23 in <module>
  File "<frozen importlib._bootstrap>", line 491 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 759 in exec_module
  File "<frozen importlib._bootstrap>", line 938 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1342 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1371 in _find_and_load
  File "<frozen importlib._bootstrap>", line 491 in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1314 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1371 in _find_and_load
  File "D:\a\torchgeo\torchgeo\.venv\Lib\site-packages\numpy\__config__.py", line 4 in <module>
  File "<frozen importlib._bootstrap>", line 491 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 759 in exec_module
  File "<frozen importlib._bootstrap>", line 938 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1342 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1371 in _find_and_load
  File "D:\a\torchgeo\torchgeo\.venv\Lib\site-packages\numpy\__init__.py", line 114 in <module>
  File "<frozen importlib._bootstrap>", line 491 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 759 in exec_module
  File "<frozen importlib._bootstrap>", line 938 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1342 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1371 in _find_and_load
  File "D:\a\torchgeo\torchgeo\.venv\Lib\site-packages\matplotlib\cbook.py", line 24 in <module>
  File "<frozen importlib._bootstrap>", line 491 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 759 in exec_module
  File "<frozen importlib._bootstrap>", line 938 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1342 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1371 in _find_and_load
  File "<frozen importlib._bootstrap>", line 491 in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1434 in _handle_fromlist
  File "D:\a\torchgeo\torchgeo\.venv\Lib\site-packages\matplotlib\__init__.py", line 161 in <module>
  File "<frozen importlib._bootstrap>", line 491 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 759 in exec_module
  File "<frozen importlib._bootstrap>", line 938 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1342 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1371 in _find_and_load
  File "<string>", line 1 in <module>

Current thread's C stack trace (most recent call first):
  <cannot get C stack on this system>
D:\a\_temp\942733ad-66d7-4fdf-a427-dd9ba1197427.sh: line 4:  1319 Segmentation fault      python -X faulthandler -u -c "print('import matplotlib', flush=True); import matplotlib; print('ok matplotlib', flush=True)"
```
Windows Python 3.14 segfaults while importing matplotlib because that imports NumPy, and NumPy prints its MINGW-W64 experimental warning immediately before the access violation.

adamjstewart:
Can you reproduce by only importing numpy? Or is this specific combination of numpy/matplotlib versions to blame? @rgommers may be interested in this segfault.

robmarkcole:
@adamjstewart I am on mac so limited to running tests on windows on CI. Isolated the error on https://github.com/torchgeo/torchgeo/actions/runs/24995108284/job/73190066997?pr=3625

.github/workflows/tests.yaml
```diff
@@ -18,7 +18,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ['3.12', '3.13', '3.14', '3.14t']
+        python-version: ['3.12', '3.13', '3.14']
+        exclude:
+          # NumPy wheels for Python 3.14 on Windows are currently MINGW-W64
+          # builds that can segfault at import, breaking pytest collection.
```
adamjstewart: @rgommers is this a known issue?

docs/conf.py
```diff
@@ -64,6 +64,9 @@
     ('py:class', 'lightning.pytorch.utilities.types.OptimizerConfig'),
     ('py:class', 'lightning.pytorch.utilities.types.OptimizerLRSchedulerConfig'),
     ('py:class', 'numpy.uint8'),
+    ('py:class', 'pandas._libs.interval.Interval'),
+    ('py:class', 'pandas._libs.tslibs.timedeltas.Timedelta'),
+    ('py:class', 'pandas.core.series.Series'),
```
adamjstewart: Not just pandas, but a lot of deps seem to have transient failures with sphinx. I don't know how to make these more stable unfortunately.

robmarkcole:
Place in draft to focus on time-series

.github/workflows/tests.yaml
robmarkcole: @adamjstewart don't think we are getting a reply. 3.14t is experimental anyway?

robmarkcole:
Hitting https://github.com/decisionintelligence/K2VAE/issues/4

- https://github.com/torchgeo/torchgeo/pull/3625/commits/b7697b7e5ff83e6a650179d61b7340d552dd8ae6 -> pinning "jsonargparse[signatures]>=4.35,<4.46" does not resolve
- https://github.com/torchgeo/torchgeo/pull/3625/commits/fe048935406fc01f0fd83120b39c8fd62ec8b5c4 -> try `"lightning..!=2.6.4",` [fails](https://github.com/torchgeo/torchgeo/actions/runs/26443438362/job/77843569948?pr=3625)
- https://github.com/torchgeo/torchgeo/pull/3625/commits/1850f667ce4a3ba821fef208a6d3666565c31f90 -> Raise the jsonargparse floor to >=4.39 and revert the incorrect !=2.6.4 lightning exclusion (2.6.1 has the same _InstantiatorFn signature 

Address: https://github.com/torchgeo/torchgeo/issues/3741

robmarkcole:
The CICD didn't complete when I had removed `3.14t`, but adding it back tests are again failing due to 
```
error: Distribution `opencv-python==4.13.0.92 @ registry+[https://pypi.org/simple`](https://pypi.org/simple%60) can't be installed because it doesn't have a source distribution or wheel for the current platfo
```
Overall I don't feel confident adding `rf-detr` as a dependency, closing

Borda:
> Overall I don't feel confident adding rf-detr as a dependency, closing

Is the openCV only blocker?

robmarkcole:
@Borda correct, thanks

robmarkcole:
@Borda created 

- https://github.com/roboflow/rf-detr/issues/1073
- https://github.com/roboflow/rf-detr/issues/1074

An alternative approach would be to move the core model functionality to a standalone repo, similar to https://github.com/allenai/olmoearth_pretrain_minimal

robmarkcole:
Reopening as making progress on the blockers

robmarkcole:
- Noting https://github.com/roboflow/rf-detr/pull/1090 which makes supervision, and hence open-cv, an optional install
- In progress https://github.com/roboflow/rf-detr/pull/1112

robmarkcole:
No movement on rf-detr PRs, moving on

Borda:
I believe, the supervison limitation shall be resolved by https://github.com/roboflow/supervision/pull/2320

robmarkcole:
https://github.com/roboflow/rf-detr/pull/1112 was merged, now awaiting https://github.com/roboflow/rf-detr/pull/1090 which is resolved by https://github.com/roboflow/supervision/pull/2320 ?

@Borda still blocked by 
```
Distribution `opencv-python==4.13.0.92 @ registry+[https://pypi.org/simple`](https://pypi.org/simple%60) can't be installed because it doesn't have a source distribution or wheel for the current platform
```

Borda:
Supervison will have opencv as optional dependency...

robmarkcole:
@Borda thanks is there a PR/release date planned?

robmarkcole:
Tracking https://github.com/roboflow/supervision/pull/2443 then will need new release of supervision & rf-detr. Looks like @Borda is making rapid progress towards https://github.com/roboflow/supervision/pull/2462

adamjstewart:
Can you rebase now that we renamed trainers to tasks?

Borda:
All needed release on our side, let me know if anything breaks... 🦝

-
adamjstewart, line 1: I think this file was added by accident?

pyproject.toml
```diff
@@ -133,6 +133,11 @@ models = [
     "microsoft-aurora>=1.6",
     # olmoearth-pretrain-minimal 0.0.2+ required for normalization support
     "olmoearth-pretrain-minimal>=0.0.2",
+    # RF-DETR 1.9.1+ required for object detection support
+    # RF-DETR depends on tokenizers, which does not provide Python 3.14t wheels
+    "rfdetr>=1.9.1; python_version<'3.14'",
+    # supervision 0.30+ required for RF-DETR 1.9.1 compatibility
+    "supervision>=0.30.0",
```
adamjstewart: I don't think you're using supervision?

torchgeo/tasks/detection.py
adamjstewart, line 1: Would be great if we can get this file down to ~10 lines changed. All of our current detection models only require ~10 lines of code, I don't know why RFDETR should require ~400.

tests/tasks/test_detection.py
adamjstewart, line 1: This file should require 0 lines changed, just add a new conf YAML file to test this new model.

adamjstewart (changes requested):
Currently 2k lines of code, shouldn't be more than 50. This is the definition of AI slop and violates the "conciseness" part of our AI policy.

pyproject.toml
robmarkcole: this is the latest release which removes open-cv dependency, which is a dependency of rfdetr

adamjstewart: And? Uv will install the latest version by default, we don't need to track dependencies of dependencies in pyproject.toml.

-
robmarkcole: artefact of uv pip compile, removed

pyproject.toml
robmarkcole: removed as comes in via rfdetr, resolves to supervision==0.30.0.

tests/tasks/test_detection.py
robmarkcole: replace with yaml

torchgeo/tasks/detection.py
robmarkcole: moved the RF-DETR specific code to seperate file and got this one down to 14 lines. RF-DETR is a more complex model

adamjstewart:
Let me know when this is no longer a draft and I'll review again.

robmarkcole:
Noting rf-detr is not supported on Python 3.14 and free-threaded 3.14t, but the actual incompatibility is Python 3.14t:

- RF-DETR depends on _transformers_, which depends on the compiled _tokenizers_ package.
- tokenizers has stable-ABI wheels usable by regular Python 3.14, but no free-threaded 3.14t wheel.
- Dependency markers cannot distinguish 3.14 from 3.14t—both report python_version == '3.14'.

Therefore TorchGeo uses python_version < '3.14', conservatively excluding both. Regular 3.14 is collateral exclusion; 3.14t is the genuinely unsupported environment.

robmarkcole:
@adamjstewart ready for review

docs/conf.py
adamjstewart: Undo this change

torchgeo/models/rfdetr.py
```diff
@@ -0,0 +1,98 @@
...
+
+from typing import Any
+
+import torch
+from torch import Tensor, nn
+from torchvision.ops import box_convert
+
+
+class RFDETR(nn.Module):
+    """Adapt RF-DETR to the torchvision object detection model interface."""
```
adamjstewart, line 16: Could we instead do this in the `forward` method of `ObjectDetection` and delete 150 lines of code?

adamjstewart (changes requested):

torchgeo/models/rfdetr.py
robmarkcole: Done, net reduction: 99 lines.

tests/tasks/test_detection.py
```diff
@@ -3,6 +3,7 @@
 
 import os
 from typing import Any
+from unittest.mock import Mock
```
adamjstewart: Use pytest's MonkeyPatch instead

tests/tasks/test_detection.py
```diff
@@ -33,11 +34,22 @@ def plot(*args: Any, **kwargs: Any) -> None:
...
+        if name == 'vhr10_rf_detr':
+            pytest.importorskip('rfdetr')
+            import rfdetr.models
+
+            load_pretrain_weights = Mock()
+            apply_lora = Mock()
+            monkeypatch.setattr(
+                rfdetr.models, 'load_pretrain_weights', load_pretrain_weights
+            )
+            monkeypatch.setattr(rfdetr.models, 'apply_lora', apply_lora)
```
adamjstewart: Are these just no-ops? Seems like we won't have test coverage with this.

torchgeo/tasks/detection.py
```diff
@@ -95,8 +104,48 @@ def __init__(
            renamed to *weights*, *lr*, and *patience*.
         """
         self.weights = weights
+        self.model_kwargs = kwargs
+        self.rf_detr_criterion: Any | None = None
+        self.rf_detr_postprocess: Any | None = None
```
adamjstewart: Seems like these are only used in `forward` and could be local vars instead of attributes?

torchgeo/tasks/detection.py
```diff
@@ -109,6 +158,43 @@ def configure_models(self) -> None:
...
+                RFDETRNanoConfig,
+                RFDETRSmallConfig,
+                TrainConfig,
+            )
+            from rfdetr.models import (
+                apply_lora,
+                build_criterion_from_config,
+                build_model_from_config,
+                load_pretrain_weights,
+            )
```
adamjstewart: I think these need to use `lazy_import`?

tests/tasks/test_detection.py
robmarkcole: done

tests/tasks/test_detection.py
robmarkcole: Replaced with an assertion

torchgeo/tasks/detection.py
robmarkcole: Removed as set in `configure_models()` - storing it on self registers it for device movement

torchgeo/tasks/detection.py
robmarkcole: done

tests/tasks/test_detection.py
```diff
@@ -33,11 +33,29 @@ def plot(*args: Any, **kwargs: Any) -> None:
...
     @pytest.mark.parametrize(
-        'name', ['nasa_marine_debris', 'reforestree', 'vhr10_obj_det']
+        'name', ['nasa_marine_debris', 'reforestree', 'vhr10_obj_det', 'vhr10_rf_detr']
     )
     def test_trainer(
         self, monkeypatch: MonkeyPatch, name: str, fast_dev_run: bool
     ) -> None:
+        if name == 'vhr10_rf_detr':
+            pytest.importorskip('rfdetr')
+            import rfdetr.models
```
adamjstewart:
```suggestion
            rfdetr = pytest.importorskip('rfdetr')
```
Does this work, or is importing the submodule also required?

adamjstewart (approved):
I think this is about as compact as it can be given the differences between torchvision- and timm-style APIs. Would like at least one more maintainer approval before merging. It's yet another dependency, but it's optional and it's useful. Will be cool to test this on some rotated bounding box datasets. I think the blocker there was getting support for that in Kornia?

tests/tasks/test_detection.py
```diff
@@ -33,11 +33,29 @@ def plot(*args: Any, **kwargs: Any) -> None:
...
+
+            # Avoid checkpoint downloads and the optional PEFT dependency.
```
adamjstewart: Does this mean that the default behavior of:
```python
ObjectDetection(model='rf-detr-nano', weights=None)
```
is to download and load pretrained model weights? Can we change this so that `weights=None` downloads nothing?

adamjstewart (changes requested):

tests/tasks/test_detection.py
adamjstewart: What is PEFT? Does this model not work without it? If so we should require the extra too.

tests/tasks/test_detection.py
robmarkcole: appears fine yes

tests/tasks/test_detection.py
robmarkcole: Re weights, I've updated so that `weights=None` is respected. 

[PEFT](https://huggingface.co/docs/peft/en/index) is Hugging Face’s Parameter-Efficient Fine-Tuning library. RF-DETR uses it only for LoRA/DoRA adapters when `backbone_lora=True`. The default is False, so I've removed `backbone_lora` from the test configuration along with the PEFT mock and assertion.

tests/tasks/test_detection.py
```diff
@@ -33,11 +33,24 @@ def plot(*args: Any, **kwargs: Any) -> None:
...
+
+            # Avoid checkpoint downloads.
+            called: set[str] = set()
+
+            def record_load_pretrain_weights(*args: Any, **kwargs: Any) -> None:
+                called.add('load_pretrain_weights')
+
+            monkeypatch.setattr(
+                rfdetr.models, 'load_pretrain_weights', record_load_pretrain_weights
+            )
```
adamjstewart: Can we remove all of this now that nothing should be downloaded?

tests/tasks/test_detection.py
```diff
@@ -76,6 +92,16 @@ def test_invalid_backbone(self) -> None:
...
 
+    def test_rf_detr_no_weights(self, monkeypatch: MonkeyPatch) -> None:
+        rfdetr = pytest.importorskip('rfdetr')
+
+        def fail(*args: Any, **kwargs: Any) -> None:
+            pytest.fail('Pretrained weights should not be loaded')
+
+        monkeypatch.setattr(rfdetr.models, 'load_pretrain_weights', fail)
+        with pytest.warns(UserWarning, match='initialised from scratch'):
+            ObjectDetection(model='rf-detr-nano', weights=None, num_classes=2)
```
adamjstewart: Can be removed, we already test this using pytest-socket.

torchgeo/tasks/detection.py
```diff
@@ -109,6 +156,36 @@ def configure_models(self) -> None:
...
+                self.model_kwargs.setdefault('pretrain_weights', None)
+            model_config = variants[model](
+                num_classes=num_classes - 1, **self.model_kwargs
+            )
+            train_config = rfdetr_config.TrainConfig(dataset_dir='.', output_dir='.')
+            self.model = rfdetr_models.build_model_from_config(
+                model_config, train_config
+            )
+            if model_config.pretrain_weights is not None:
+                rfdetr_models.load_pretrain_weights(self.model, model_config)
```
adamjstewart: Not sure I like how weights are handled, would be nice if `weights` and `pretrain_weights` were forced to match and used as such. That would also keep things simpler. Only way to test this is to create model weights on the fly, cache them, then reload them. That's how we get test coverage for other models.

tests/tasks/test_detection.py
robmarkcole: removed

tests/tasks/test_detection.py
robmarkcole: removed

torchgeo/tasks/detection.py
robmarkcole: Updated so loading is controlled solely by `weights`, which now accepts RF-DETR checkpoint paths.
Added test creates, saves, and reloads temporary RF-DETR weights.

tests/tasks/test_detection.py
```diff
@@ -76,6 +80,37 @@ def test_invalid_backbone(self) -> None:
...
+        def apply_lora(model: torch.nn.Module) -> None:
+            called.append(model)
+
+        monkeypatch.setattr(rfdetr.models, 'apply_lora', apply_lora)
+
+        model = ObjectDetection(
+            model='rf-detr-nano', weights=None, num_classes=2, backbone_lora=True
+        )
+
+        assert called == [model.model]
```
adamjstewart: We can't remove all this and treat rf-detr like all other models?

torchgeo/tasks/detection.py
```diff
@@ -95,8 +104,44 @@ def __init__(
            renamed to *weights*, *lr*, and *patience*.
         """
         self.weights = weights
+        self.model_kwargs = kwargs
         super().__init__()
 
+    def forward(
+        self, images: Tensor, targets: list[dict[str, Tensor]] | None = None
+    ) -> dict[str, Tensor] | list[dict[str, Tensor]]:
+        """Run a forward pass."""
```
adamjstewart: Needs documentation of inputs/outputs

torchgeo/tasks/detection.py
```diff
@@ -109,6 +154,35 @@ def configure_models(self) -> None:
...
+            self.model_kwargs['pretrain_weights'] = self.weights
+            model_config = variants[model](
+                num_classes=num_classes - 1, **self.model_kwargs
+            )
+            train_config = rfdetr_config.TrainConfig(dataset_dir='.', output_dir='.')
+            self.model = rfdetr_models.build_model_from_config(
+                model_config, train_config
+            )
+            if self.weights is not None:
+                rfdetr_models.load_pretrain_weights(self.model, model_config)
```
adamjstewart: Could manually load weights just like all other tasks/models.

torchgeo/tasks/detection.py
```diff
@@ -109,6 +154,35 @@ def configure_models(self) -> None:
...
+            if model_config.backbone_lora:
+                rfdetr_models.apply_lora(self.model)
```
adamjstewart: We don't support this for any other model, this could possibly be a new callback?

adamjstewart (changes requested):
Let's save this one for the next release so other maintainers have time to review, especially since it adds yet another dependency.

tests/tasks/test_detection.py
robmarkcole: You mean remove lora support, or also loading pretrained weights? I think lora will be niche usage, pretrained weights essential

adamjstewart: Remove lora, and make loading pretrained weights the same as all other models (`model.load_state_dict()`)

adamjstewart: And remove all the new tests

torchgeo/tasks/detection.py
robmarkcole: Per `Remove lora` lets just remove this functionality, can be added in another PR, although I'm not sure who would even need it

tests/tasks/test_detection.py
robmarkcole: Done

torchgeo/tasks/detection.py
robmarkcole: done

torchgeo/tasks/detection.py
robmarkcole: added

torchgeo/tasks/detection.py
```diff
@@ -95,8 +104,57 @@ def __init__(
            renamed to *weights*, *lr*, and *patience*.
         """
         self.weights = weights
+        self.model_kwargs = kwargs
         super().__init__()
 
+    def forward(
```
calebrob6: I've seen a ton of back-and-forth on this PR, so apologies if this has already been hashed out.

The forward(...) seems to be custom to rf-detr (lines 127 and 128 just forward into the individual model), so it would make sense to me that this be part of the rf-detr model (and not the object detection trainer). Thinking ahead, if there is another object detection model that needs a custom forward, this would get messy fast.

torchgeo/tasks/detection.py
```diff
@@ -109,6 +167,41 @@ def configure_models(self) -> None:
         num_classes: int = self.hparams['num_classes']
         freeze_backbone: bool = self.hparams['freeze_backbone']
 
+        if model.startswith('rf-detr'):
```
calebrob6, line 117: Seeing this (along with my previous comment), it strikes me that RF-DETR should likely just be a subclass of the object detectiont ask?

pyproject.toml
```diff
@@ -135,6 +135,9 @@ models = [
     "microsoft-aurora>=1.6",
     # olmoearth-pretrain-minimal 0.0.2+ required for normalization support
     "olmoearth-pretrain-minimal>=0.0.2",
+    # RF-DETR 1.9.1+ required for object detection support
+    # RF-DETR depends on tokenizers, which does not provide Python 3.14t wheels
+    "rfdetr>=1.9.1; python_version<'3.14'",
```
calebrob6, line 140: Following current patterns, this is fine with me. We need to sit down and have a strategic conversation about this direction though because there are going to be unlimited opportunities to keep adding new model/dataset dependencies. This essentially shifts complexity from "user time" to "maintainer time". E.g. assume this PR is merged -- as a user that's great for me because I can theoretically just "use" rf-detr now with any torchgeo dataset (whereas previously I need to do some custom work to interface the two). As a maintainer then this is yet another thing that gets dependabot updates, breaks things, gets broken by other things, etc.

calebrob6:
I could do a real training run on VHR-10 using random init RF-DETR. I **couldn't** start training from the official pre-trained weights though. @robmarkcole is this something you want to do?

pyproject.toml
robmarkcole: @calebrob6 I agree, more dependencies need to be weighted against maintenance. FWIW I worked with rf-detr maintainers to clean up their own dependencies ahead of this pr, to make opencv-python an optional for example. Currently object detection doesn't have many model options, so I think this is a meaningful addition, and justifies maintenance.

torchgeo/tasks/detection.py
robmarkcole: @calebrob6 previously this was a standalone class, see https://github.com/torchgeo/torchgeo/blob/6f0a634/torchgeo/models/rfdetr.py @adamjstewart advise to merge it into the trainer to reduce LOC but I agree this approach could get messy and I prefer the seperation

Claude makes a good observation: The reason `training_step/validation_step/test_step/predict_step` are model-agnostic today is that **FasterRCNN/FCOS/RetinaNet all already speak the same interface**: `forward(images, targets=None)` returns a loss dict in train mode and a list of xyxy/1-indexed prediction dicts in eval mode. That uniformity is what lets the task stay generic. RF-DETR breaks that interface (it wants cxcywh normalized boxes, 0-indexed labels, and separate criterion/postprocess objects), and instead of absorbing that mismatch behind an adapter, the current code leaks it straight into the shared class.

Claude also made a new suggestion: **Option 3: Backend registry/plugin pattern (most scalable, biggest change)**
Instead of hardcoded `if model.startswith(...)` branches, define a small internal interface (e.g. a `DetectionBackend` protocol with `build_model()` and `forward()`) and a registry dict mapping model-name prefixes to backend implementations. torchvision and rf-detr each become one registered backend; adding a future model family (YOLO-style, another DETR variant) means registering a new backend, not editing configure_models()/forward() again.

torchgeo/tasks/detection.py
robmarkcole: as comment above

robmarkcole:
@calebrob6 was there an error with the pretrained weights? I actually can't see anyone wanting to use rf-detr _without_ pretrained weights, but this approach was taken to align with the existing behaviour for OD.

torchgeo/tasks/detection.py
calebrob6: Having a model that the existing object detection task can use seems much more appropriate here. That way the custom forward logic can all live in the model? @adamjstewart these lines of code are worth keeping.

pyproject.toml
calebrob6: thanks :) -- the above is just me ranting in general, should probably be an issue or a discussion rather than a review thread on the >80 comment rf-detr PR hahaha

calebrob6:
Yes, the RF-DETR pre-trained weights don't work by default because they are full training checkpoints that contain _model_, optimizer, scheduler, epoch, and an `argparse.Namespace`. Some things:

1. Current method of `weights_only=True` rejects everything. Very secure 👍 
2. If we attempt to load `checkpoint['model']` directly, it still fails because its COCO head has 90 classes, while VHR10 has 10.
3. RF-DETR has a `load_pretrain_weights` method that unwraps the checkpoint**, loads compatible pretrained parameters, and reinitializes the incompatible detection head.

I used the RF-DETR loader to create TorchGeo-compatible raw state dicts and compared it vs. random init on VHR-10 for 20-epochs. This changed Nano's test mAP from 0.023 scratch to 0.572 pretrained. 

** I don't know if it does this safely or not.

robmarkcole:
@calebrob6 I've restored standalone RFDETR adapter and removed RF-DETR logic from ObjectDetection.

- weights maps to pretrain_weights.
- Uses RF-DETR’s load_pretrain_weights.

Please let me know how you get on with this?

PR #3629, opened 2026-04-21
add DEO model

wolfilip:
Added DEO model with tests and docs.


torchgeo/models/deo.py
```diff
@@ -0,0 +1,132 @@
...
+        """Get multi-stage swin features.
+
+        Args:
+            x: input image tensor (b, c, h, w).
+
+        Returns:
+            list of swin feature tensors list[(b, c, h', w')].
+        """
+        features = []
+        # apply the appropriate conv layer based on the number of input channels
```
isaaccorley, line 73: It would be better to make the class have an `in_channels` arg like `def __init__(self, model: str = 'swin_b', in_channels: int = 10)` and then add a `self.neck = feat_extr.conv_ms if in_channels = 10 else feat_extr.conv_rgb` to be explicit instead of having if/else logic in the forward pass

torchgeo/models/deo.py
```diff
@@ -0,0 +1,132 @@
...
+    """Pretrained Distillation for EO (DEO) model implementation.
+
+    If you use this model in your research, please cite the following paper:
+
+    * https://arxiv.org/abs/2602.19863
+
+    .. versionadded:: 0.9
+    """
+
+    def __init__(self, model: str = 'swin_b') -> None:
```
isaaccorley: model should be `model: Literal[str] = ['swin_b', ...]` with the possible combinations you expect a user to be able to use

docs/api/weights/agnostic.csv
```diff
@@ -12,3 +12,4 @@ OlmoEarthV1_Weights.TINY,`link <https://github.com/allenai/olmoearth_pretrain>`_
 OlmoEarthV1_Weights.BASE,`link <https://github.com/allenai/olmoearth_pretrain>`__,`link <https://arxiv.org/abs/2511.13655>`__,OlmoEarth Artifact License,explicit,explicit,explicit,,,,,,,,,,,
 OlmoEarthV1_Weights.LARGE,`link <https://github.com/allenai/olmoearth_pretrain>`__,`link <https://arxiv.org/abs/2511.13655>`__,OlmoEarth Artifact License,explicit,explicit,explicit,,,,,,,,,,,
 Tessera_Weights.TESSERA,`link <https://github.com/ucam-eo/tessera>`__,`link <https://arxiv.org/abs/2506.20380>`__,CC0-1.0,-,explicit,-,,,,,,,,,,,
+DEO_Weights.DEO_SWIN,`link <https://github.com/wolfilip/DEO-FM>`__,`link <https://arxiv.org/abs/2602.19863>`__,CC-BY-4.0,-,-,-,58.43,,,,52.23,97.0,94.41,80.21,,68.56,37.96,72.47
```
isaaccorley:
```suggestion
DEO_Weights.DEO_SWIN,`link <https://github.com/wolfilip/DEO-FM>`__,`link <https://arxiv.org/abs/2602.19863>`__,CC-BY-4.0,-,-,explicit,58.43,,,,52.23,97.0,94.41,80.21,,68.56,37.96,72.47
```

torchgeo/models/deo.py
```diff
@@ -0,0 +1,132 @@
...
+                3,
+                self.feat_extr.features[0][0].norm1.normalized_shape[0],
+                kernel_size=(4, 4),
+                stride=(4, 4),
+            ),
+            Permute([0, 2, 3, 1]),
+            norm_layer_rgb(self.feat_extr.features[0][0].norm1.normalized_shape[0]),
+        )
+
+    def forward_swin(self, x: torch.Tensor) -> list[torch.Tensor]:
```
isaaccorley: Seems like we could save a few lines just by renaming this method to `forward` and cutting the other one

torchgeo/models/deo.py
```diff
@@ -0,0 +1,132 @@
...
+    """DEO base model weights.
+
+    .. versionadded:: 0.9
+    """
+
+    DEO_SWIN = Weights(
+        url='https://huggingface.co/SolaireTheSun/DEO/resolve/main/DEO_swin_b.pth',
+        transforms=nn.Identity(),
+        meta={
+            'dataset': 'fMoW',
```
isaaccorley: Need some more info here like what bands it expects, what license the weights are under, what the pretraining input shape was, etc. Please reference some other weights. Anything that feels relevant for a user to know what inputs are "in-distribution" or "out-of-distribution"

torchgeo/models/deo.py
```diff
@@ -0,0 +1,132 @@
...
+
+class DEO_Weights(WeightsEnum):
+    """DEO base model weights.
+
+    .. versionadded:: 0.9
+    """
+
+    DEO_SWIN = Weights(
+        url='https://huggingface.co/SolaireTheSun/DEO/resolve/main/DEO_swin_b.pth',
+        transforms=nn.Identity(),
```
isaaccorley: Is it really `nn.Identity()`?

isaaccorley (dismissed):

torchgeo/models/deo.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class DEO(nn.Module):
+    """Pretrained Distillation for EO (DEO) model implementation.
+
+    If you use this model in your research, please cite the following paper:
+
+    * https://arxiv.org/abs/2602.19863
+
+    .. versionadded:: 0.9
```
nilsleh: should be 0.10, since 0.9 is already the current release.

torchgeo/models/deo.py
```diff
@@ -0,0 +1,132 @@
...
+        """Forward function for DEO."""
+        features = self.forward_swin(x)
+
+        return features
+
+
+class DEO_Weights(WeightsEnum):
+    """DEO base model weights.
+
+    .. versionadded:: 0.9
```
nilsleh: same version comment as above.

nilsleh (changes requested):

torchgeo/models/deo.py
wolfilip: The model was pretrained on high-res RGB, normalized with Imagenet weights, and low-res Sentinel-2, normalized with Sentinel-2 values. The normalization values for evaluation are based on the dataset used in my experiments (since normalizing Sentinel-2 data with ImageNet values would yield very large values). I don't know how to handle this. Are normalization values, backbone dependent?

wolfilip: I have currently implemented it as dynamic normalization in the forward function based on in_channels.

nilsleh: The main intention for the transforms parameter in the Weights class is to document the normalization scheme used during pretraining of those weights so users have an idea of what normalization range the model expects data to be in.

adamjstewart: I'm fine with no normalization in the weights and dynamic normalization within the model, as long as it's documented somewhere. @isaaccorley is this a big no-no in the STAC MLM world?

torchgeo/models/deo.py
wolfilip: Done.

torchgeo/models/deo.py
wolfilip: Added.

docs/api/weights/agnostic.csv
wolfilip: Added.

torchgeo/models/deo.py
wolfilip: Done.

torchgeo/models/deo.py
wolfilip: I checked some other methods and added this info to the best of my knowledge.

torchgeo/models/deo.py
wolfilip: Done

torchgeo/models/deo.py
wolfilip: Done

torchgeo/models/deo.py
isaaccorley: Better to detach the normalization not to be in the forward function of a model. It's better to be explicit because only knowing number of input channels isn't enough to determine what imagery is being fed to the model.

wolfilip: I have removed normalizations from the forward pass and added information on how the bands were normalized for pretraining.

wolfilip:
I have addressed all posted comments.

blaz-r:
@adamjstewart @isaaccorley any update on this?

docs/api/models/deo.rst
```diff
@@ -0,0 +1,7 @@
+DEO
+=====
```
adamjstewart:
```suggestion
===
```

docs/api/weights/agnostic.csv
```diff
@@ -12,3 +12,4 @@ OlmoEarthV1_Weights.TINY,`link <https://github.com/allenai/olmoearth_pretrain>`_
 OlmoEarthV1_Weights.BASE,`link <https://github.com/allenai/olmoearth_pretrain>`__,`link <https://arxiv.org/abs/2511.13655>`__,OlmoEarth Artifact License,explicit,explicit,explicit,,,,,,,,,,,
 OlmoEarthV1_Weights.LARGE,`link <https://github.com/allenai/olmoearth_pretrain>`__,`link <https://arxiv.org/abs/2511.13655>`__,OlmoEarth Artifact License,explicit,explicit,explicit,,,,,,,,,,,
 Tessera_Weights.TESSERA,`link <https://github.com/ucam-eo/tessera>`__,`link <https://arxiv.org/abs/2506.20380>`__,CC0-1.0,-,explicit,-,,,,,,,,,,,
+DEO_Weights.DEO_SWIN,`link <https://github.com/wolfilip/DEO-FM>`__,`link <https://arxiv.org/abs/2602.19863>`__,CC-BY-4.0,-,-,explicit,58.43,,,,52.23,97.0,94.41,80.21,,68.56,37.96,72.47
```
adamjstewart: Can you sort in alphabetical order?

adamjstewart: HF says the weights are under the MIT license, where did CC-BY-4.0 come from?

tests/models/test_deo.py
```diff
@@ -0,0 +1,63 @@
...
+
+from pathlib import Path
+
+import pytest
+import torch
+from _pytest.fixtures import SubRequest
+from pytest import MonkeyPatch
+
+from torchgeo.models import deo_base
+from torchgeo.models.deo import DEO_Weights
```
adamjstewart:
```suggestion
from torchgeo.models import DEO_Weights
```

torchgeo/models/deo.py
```diff
@@ -0,0 +1,167 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+
+"""Pretrained Distillation for EO (DEO) model implementation."""
+
+from functools import partial
+from typing import Any, Literal
+
+import torch
+import torch.nn as nn
```
adamjstewart: Latest version of ruff will complain about this line

torchgeo/models/deo.py
```diff
@@ -0,0 +1,167 @@
...
+
+    If you use this model in your research, please cite the following paper:
+
+    * https://arxiv.org/abs/2602.19863
+
+    .. versionadded:: 0.10
+    """
+
+    def __init__(
+        self, model: Literal['swin_b'] = 'swin_b', in_channels: Literal[3, 10] = 10
```
adamjstewart: The model doesn't work for any other # channels?

torchgeo/models/deo.py
```diff
@@ -0,0 +1,167 @@
...
+    ) -> None:
+        """Initialise DEO model.
+
+        Args:
+            model: backbone type (for now swin_b).
```
adamjstewart, line 31: No other backbones work?

torchgeo/models/deo.py
```diff
@@ -0,0 +1,167 @@
...
+            self.feat_extr.patch_embed = nn.Sequential(
+                nn.Conv2d(
+                    3,
+                    self.feat_extr.features[0][0].norm1.normalized_shape[0],
+                    kernel_size=(4, 4),
+                    stride=(4, 4),
+                ),
+                Permute([0, 2, 3, 1]),
+                norm_layer_rgb(self.feat_extr.features[0][0].norm1.normalized_shape[0]),
+            )
```
adamjstewart:
```suggestion
        norm_layer = partial(nn.LayerNorm, eps=1e-5)
        self.feat_extr.patch_embed = nn.Sequential(
                nn.Conv2d(
                    in_channels,
                    self.feat_extr.features[0][0].norm1.normalized_shape[0],
                    kernel_size=(4, 4),
                    stride=(4, 4),
                ),
                Permute([0, 2, 3, 1]),
                norm_layer(self.feat_extr.features[0][0].norm1.normalized_shape[0]),
        )
```

torchgeo/models/deo.py
```diff
@@ -0,0 +1,167 @@
...
+
+import torch
+import torch.nn as nn
+from torchvision import models as torchvision_models
+from torchvision import transforms
+from torchvision.models._api import Weights, WeightsEnum
+from torchvision.ops.misc import Permute
+
+
+class DEO(nn.Module):
```
adamjstewart, line 17: Perhaps I'm missing something, but how does this model architecture differ from any of our other swin models? It just looks like you're replacing the first layer with the correct number of channels, which timm already does automatically.

torchgeo/models/deo.py
```diff
@@ -0,0 +1,167 @@
...
+            )
+
+    def forward(self, x: torch.Tensor) -> list[torch.Tensor]:
+        """Get multi-stage swin features.
+
+        Args:
+            x: input image tensor (b, c, h, w).
+
+        Returns:
+            list of swin feature tensors list[(b, c, h', w')].
```
adamjstewart, line 70: Why do we want the feature tensors instead of the output prediction?

torchgeo/models/deo.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+class DEO_Weights(WeightsEnum):
+    """DEO base model weights.
+
+    .. versionadded:: 0.10
+    """
+
+    DEO_SWIN = Weights(
+        url='https://huggingface.co/SolaireTheSun/DEO/resolve/main/DEO_swin_b.pth',
```
adamjstewart: main should be replaced by the exact git commit hash, filename should include sha256 prefix so we can use `check_hash=True` in `torch.hub.load_state_dict_from_url`

torchgeo/models/deo.py
```diff
@@ -0,0 +1,167 @@
...
+        transforms=_deo_transforms,
+        meta={
+            'dataset': 'fMoW, fMoW-Sentinel',
+            'model': 'Swin_b',
+            'publication': 'https://arxiv.org/abs/2602.19863',
+            'repo': 'https://github.com/wolfilip/DEO-FM',
+            'license': 'MIT',
+            'ssl_method': 'DEO',
+            'bands': ['R', 'G', 'B', 'B5', 'B6', 'B7', 'B8', 'B8A', 'B11', 'B12'],
```
adamjstewart:
```suggestion
            'bands': ['B4', 'B3', 'B2', 'B5', 'B6', 'B7', 'B8', 'B8A', 'B11', 'B12'],
```

torchgeo/models/deo.py
```diff
@@ -0,0 +1,167 @@
...
+        *args: Additional arguments to pass to :class:`DEO`.
+        **kwargs: Additional keyword arguments to pass to :class:`DEO`.
+
+    Returns:
+        DEO Swin model.
+    """
+    model = DEO(*args, **kwargs)
+
+    if weights:
+        state_dict = weights.get_state_dict(progress=True)
```
adamjstewart:
```suggestion
        state_dict = weights.get_state_dict(progress=True, weights_only=True)
```
Security vulnerability, especially if we had kept using the tip of the main branch.

adamjstewart (changes requested):

torchgeo/models/deo.py
wolfilip: The model has a separate 3- (RGB) and 10-channel input encoding layer. Other combinations can only be handled by zeroing bands in the 10-channel encoder. The model was specifically optimised for these two combinations of input bands.

torchgeo/models/deo.py
wolfilip: During pretraining, the model trained two separate input encoding layers: one for 3 input channels (RGB) and the other for 10 input channels. The performance depends greatly on the trained input encoding weights being loaded.

torchgeo/models/deo.py
wolfilip: This forward function was designed for segmentation heads such as UPerNet. I can also just return the last-layer patch features, the pooled features, or whatever is correct.

torchgeo/models/deo.py
wolfilip: We added ViT support, but the main version is Swin. I can add the ViT version if needed, but I prefer users to use the Swin one.

docs/api/models/deo.rst
wolfilip: Addressed.

docs/api/weights/agnostic.csv
wolfilip: Addressed.

tests/models/test_deo.py
wolfilip: Addressed.

torchgeo/models/deo.py
wolfilip: Addressed.

torchgeo/models/deo.py
wolfilip: Addressed.

torchgeo/models/deo.py
wolfilip: Addressed.

torchgeo/models/deo.py
adamjstewart: How is that different from https://arxiv.org/abs/2401.07527?

torchgeo/models/deo.py
adamjstewart: I only see a single input encoding layer in this model. In fact, after the recent changes, I don't see any difference from this model and normal Swin. Am I missing something, or is this just a Swin model pretrained on 10 channels of imagery? If so, we can move it to `torchgeo/models/swin.py`.

wolfilip: it is a swin model  that is pretrained simultaneously on both 10 and 3 channel imagery so it explicitly sees both, and therefore has 2 separate encoding layers that are loaded during model initialization. If this can be just moved to the standard Swin, then its fine.

torchgeo/models/deo.py
wolfilip: In terms of pretraining, our model is pretrained using dual branch contrastive self-distillation for MS and VFM distillation fro RGB insetad of masked image modeling. In terms of just the input encodings, i don't see a difference. The student is a Swin with two separate input encoding layers loaded that change based on the input data.

torchgeo/models/deo.py
adamjstewart: Perhaps forward_features is more appropriate for each layer and forward is more appropriate for the last layer?

torchgeo/models/deo.py
adamjstewart: Where are the two separate encoding layers? I only see one.

adamjstewart:
If a single model/weights is supposed to have 2 different encoders, then this is fine as a separate class. If each model only has a single encoder at a time (one model for 3 channels and a separate model for 10 channels) then we can add this to `torchgeo/models/swin.py`, it would only be 2 new weights.

wolfilip:
It is a single model with two different encoders/input layers.

torchgeo/models/deo.py
wolfilip: Addressed.

adamjstewart:
Moved this to 0.11 because it doesn't seem like this will be ready today, but let me know if you have time to finish this today.

Current problem is that the model is supposed to have two encoder heads but currently only has one. I can't easily test this since it's on your main branch instead of a separate branch, but I suspect the model weights don't even load with the current implementation as the rgb and ms encoder heads are missing.

wolfilip:
I have now re-implemented the two separate input encoders.

torchgeo/models/deo.py
```diff
@@ -0,0 +1,189 @@
...
+    """
+
+    def __init__(
+        self, model: Literal['swin_b'] = 'swin_b', in_channels: Literal[3, 10] = 10
+    ) -> None:
+        """Initialise DEO model.
+
+        Args:
+            model: backbone type (for now swin_b).
+            in_channels: Number of input channels.
```
adamjstewart: in_channels is no longer used and can be removed

torchgeo/models/deo.py
```diff
@@ -0,0 +1,189 @@
...
+        if x.shape[1] == 10:
+            x = self.feat_extr.conv_ms(x)
+        else:
+            x = self.feat_extr.conv_rgb(x)
+
+        # extract intermediate swin layers
+        for i, layer in enumerate(self.feat_extr.features):
+            x = layer(x)
+
+        return x
```
adamjstewart: Could we replace this with:
```suggestion
        return self.forward_features(x)[-1]
```
This would also fix test coverage for `forward_features`.

adamjstewart (changes requested):

torchgeo/models/deo.py
wolfilip: Resolved.

torchgeo/models/deo.py
wolfilip: Resolved.

adamjstewart (dismissed):
LGTM now, just needs approval from @nilsleh and @isaaccorley

torchgeo/models/deo.py
```diff
@@ -0,0 +1,176 @@
...
+        *args: Additional arguments to pass to :class:`DEO`.
+        **kwargs: Additional keyword arguments to pass to :class:`DEO`.
+
+    Returns:
+        DEO Swin model.
+    """
+    model = DEO(*args, **kwargs)
+
+    if weights:
+        state_dict = weights.get_state_dict(progress=True, weights_only=True)
```
adamjstewart:
```suggestion
        state_dict = weights.get_state_dict(progress=True, check_hash=True, weights_only=True)
```
One last change, although your download filename doesn't contain the sha256 hash so this won't actually be used yet. You could consider renaming the download file.

wolfilip: Will I not get a torchgeo link? Otherwise, addressed.

nilsleh (approved):
@isaaccorley Can you kick of torchgeo-bench runs for this after being merged? Curious to see what the numbers are!

torchgeo/models/deo.py
adamjstewart: You can, although we need to vote on it first. See the Organization Card for https://huggingface.co/torchgeo for instructions.

adamjstewart (approved):

torchgeo/models/deo.py
adamjstewart: This would result in the model only having a single feature extractor instead of two, so we would basically be reinventing a Swin transformer. The point of DEO is that it has two simultaneous feature extractors. Reverted so that model weights can actually load.

blaz-r:
> @isaaccorley Can you kick of torchgeo-bench runs for this after being merged? Curious to see what the numbers are!

That'd be awesome to see.

nilsleh:
@blaz-r @wolfilip  I am working on checking the benchmark results and, one thing I saw is: a potential band-order discrepancy for DEO’s 10-channel S2 input. TorchGeo’s [DEO_Weights.DEO_SWIN metadata](https://github.com/torchgeo/torchgeo/blob/main/torchgeo/models/deo.py#L138-L150) specifies B04,B03,B02,B05,B06,B07,B08,B8A,B11,B12, while DEO-FM’s [fMoWMSDataset loader](https://github.com/wolfilip/DEO-FM/blob/main/DEO/utils/datasets.py#L75-L88) selects B02,B03,B04,B05,B06,B07,B08,B8A,B11,B12. Could you confirm which ordering the released Swin-B checkpoint expects at inference? This affects the first three channels and their normalization.

wolfilip:
The model was pretrained and tested with bands B04,B03,B02,B05,B06,B07,B08,B8A,B11,B12, so the first three (non-atmospheric) S2 bands are in an RGB order. I will fix this in my local repo and specify the expected order.

PR #3637, opened 2026-04-25
Add S2-100k dataset

adamjstewart:
Includes several improvements over the [original](https://github.com/microsoft/satclip/blob/main/satclip/datamodules/s2geo_dataset.py) implementation:

- [x] Fixes a couple type hinting mistakes
- [x] Clips images before plotting to fix warning
- [x] Adds automatic download and checksum support
- [x] API documentation
- [x] Unit tests

Let me know if you want me to submit the first 2 fixes upstream.

Could also add the data module, but we don't yet have a trainer for this task to test it with.

Example plot:

s2_100k

As a sanity check, I also confirmed that these lon/lat match this image via Google Maps:

Screenshot 2026-04-25 at 3 01 38 PM

## AI Usage Disclosure

@konstantinklemmer @calebrob6 @estherrolf


docs/api/datasets/pretraining.csv
```diff
@@ -4,6 +4,7 @@ Dataset,Task,Source,# Samples,# Classes,Size (px),Resolution (m),Bands,License
 :ref:`HySpecNet-11k`,-,EnMAP,11k,-,128,30,HSI,CC0-1.0
 :ref:`MMEarth`,"C, S","Aster, Sentinel, ERA5","100K--1M",,"128x128 or 64x64",10,MSI,CC-BY-4.0
 :ref:`Million-AID`,C,Google Earth,1M,51--73,,0.5--153,RGB,
+:ref:`S2-100k`,LE,Sentinel-2,100k,-,256x256,10,MSI,
```
adamjstewart: Unknown license: https://huggingface.co/datasets/davanstrien/satclip/discussions/4

torchgeo/datasets/s2_100k.py
```diff
@@ -0,0 +1,207 @@
...
+    Sample,
+    download_and_extract_archive,
+    download_url,
+    extract_archive,
+)
+
+CHECK_MIN_FILESIZE = 10000  # 10kb
+
+
+class S2100k(NonGeoDataset):
```
adamjstewart, line 31: Original source calls it S2Geo, I went with S2100k to match the dataset name from the paper.

torchgeo/datasets/s2_100k.py
```diff
@@ -0,0 +1,207 @@
...
+                n_skipped_files += 1
+                continue
+
+            self.filenames.append(filename)
+            self.points.append((df.iloc[i]['lon'], df.iloc[i]['lat']))
+
+        print(
+            f'skipped {n_skipped_files}/{len(df)} images because they were smaller '
+            f'than {CHECK_MIN_FILESIZE} bytes... they probably contained nodata pixels'
+        )
```
adamjstewart: This is pretty slow, can we instead:

1. Keep all data?
2. Permanently remove invalid from index.csv and/or data.tar.xz?
3. Create a separate index_valid.csv?

torchgeo/datasets/utils.py
```diff
@@ -388,7 +388,7 @@ def download_url(
     filename: Path | None = None,
     md5: str | None = None,
     max_redirect_hops: int = 3,
-    **kwargs: str,
+    **kwargs: str | None,
```
adamjstewart, line 391: I fixed this in `check_integrity` but forgot to fix this for functions that call this helper function.

torchgeo/datasets/s2_100k.py
adamjstewart: Removing this also makes testing significantly easier...

torchgeo/datasets/s2_100k.py
```diff
@@ -0,0 +1,207 @@
...
+
+        Args:
+            sample: A sample returned by :meth:`__getitem__`.
+            show_titles: Flag indicating whether to show titles above each panel.
+            suptitle: Optional string to use as a suptitle.
+
+        Returns:
+            A matplotlib Figure with the rendered sample.
+        """
+        image = sample['image'][[3, 2, 1]]
```
adamjstewart, line 159: What should `plot` do when `mode='points'`?

docs/api/datasets/pretraining.csv
isaaccorley: @konstantinklemmer what license should we put for the dataset?

torchgeo/datasets/s2_100k.py
```diff
@@ -0,0 +1,190 @@
...
+        '1c419e0a44877844c79eaedf227b13c4854f1acddd959ac7109c73627311b39f',
+        'bc923a3420623925afb024edcadcb8295ebeecb8337726b05493068951462cde',
+        'a3399adc2b99528d2e64ebb74cefef74f70226f9392fa29cd9214c2ffbfcdd8b',
+    )
+
+    def __init__(
+        self,
+        root: Path = 'data',
+        *,
+        mode: Literal['both', 'points'] = 'both',
```
isaaccorley, line 55: Should this be "all" to match other datasets?

```suggestion
        mode: Literal['all', 'points'] = 'all',
```

torchgeo/datasets/s2_100k.py
```diff
@@ -0,0 +1,190 @@
...
+
+        self.root = pathlib.Path(root)
+        self.transforms = transforms
+        self.mode = mode
+        self.download = download
+        self.checksum = checksum
+
+        self._verify()
+
+        self.df = pd.read_csv(self.root / 'index.csv')
```
isaaccorley: Nit: Should we call this index instead so it's more self explanatory than df?

```suggestion
        self.index = pd.read_csv(self.root / 'index.csv')
```

torchgeo/datasets/s2_100k.py
isaaccorley: Maybe could use geopandas to plot a red dot on top of a basemap

isaaccorley (changes requested):
A couple nits but overall lgtm

torchgeo/datasets/s2_100k.py
adamjstewart: I think basemap requires cartopy?

torchgeo/datasets/s2_100k.py
adamjstewart: What other datasets call it "all"? This just matches the original implementation, I don't know of any other standards.

torchgeo/datasets/s2_100k.py
adamjstewart: Only downside is that it might get confused with our GeoDataset.index, but yeah this is definitely more clear.

adamjstewart:
Note that there are some missing files in the dataset: https://github.com/microsoft/satclip/issues/30. @calebrob6 is looking into this.

Update: full version re-uploaded to https://huggingface.co/datasets/kklmmr/s2-100k, will update the download.

docs/api/datasets/pretraining.csv
adamjstewart: MIT was chosen for https://huggingface.co/datasets/kklmmr/s2-100k, will update

adamjstewart:
Can anyone with more storage than me confirm that download/checksum/extract works as expected?

calebrob6:
yeah I'll look

calebrob6:
had a couple of HTTP errors as the download was big, I'd put a hf cli example in the docstring in case people want to download it offline

torchgeo/datasets/s2_100k.py
```diff
@@ -0,0 +1,174 @@
...
+    download_and_extract_archive,
+    download_url,
+    extract_archive,
+)
+
+
+class S2100k(NonGeoDataset):
+    """S2-100K dataset.
+
+    The `S2-100k dataset <https://huggingface.co/datasets/davanstrien/satclip>`__
```
calebrob6: Should likely be https://huggingface.co/datasets/kklmmr/s2-100k

calebrob6 (changes requested):

torchgeo/datasets/s2_100k.py
adamjstewart: Yes, good catch

adamjstewart:
> had a couple of HTTP errors as the download was big, I'd put a hf cli example in the docstring in case people want to download it offline

I'm wondering if we should start using `hf_hub_download()` across TorchGeo instead of urllib/requests. It's one more dep, but I'm guessing it's probably faster/more stable?

calebrob6:
Potentially, I'd leave that to a larger refactor of how to download everything that happens to be on HF

calebrob6 (approved):

isaaccorley (approved):

PR #3640, opened 2026-04-26
feat: SatCLIP location encoder

isaaccorley:
- Adds the SatCLIP location encoder ([Klemmer et al., arXiv:2311.17179](https://arxiv.org/abs/2311.17179)) — the first location encoder model in torchgeo.
- This is an **optimized** implementation: the spherical-harmonics positional encoding uses a stable analytic three-term recurrence on the associated Legendre polynomials in place of the SymPy-generated closed-form expressions in the original

cc @estherrolf @konstantinklemmer @marccoru

Closes #3580

## AI Usage Disclosure

I used AI to help profile the original SatCLIP encoder to find the largest bottlenecks and to validate parity with the original codebase


konstantinklemmer:
Yay and thanks! 🎉

torchgeo/models/satclip.py
```diff
@@ -0,0 +1,429 @@
...
+    """
+
+    def __init__(
+        self,
+        legendre_polys: int = 40,
+        capacity: int = 512,
+        embed_dim: int = 256,
+        num_hidden_layers: int = 2,
+    ) -> None:
+        """Initialize a new SatCLIP instance.
```
nilsleh, line 393: Wondering if we can add some more info/context about the arguments here, like some kind of intuition about what the effect of increasing/decreasing that parameter does? I think that would be really helpful for users

torchgeo/models/satclip.py
```diff
@@ -0,0 +1,429 @@
...
+        self.bias: nn.Parameter | None = (
+            nn.Parameter(bias) if bias is not None else None
+        )
+        self.activation = SineActivation(w0) if activation is None else activation
+
+    def forward(self, x: Tensor) -> Tensor:
+        """Forward pass.
+
+        Args:
+            x: Input tensor.
```
nilsleh, line 292: perhaps also helpful to have description of shapes included here as well

adamjstewart:
I think the weights are missing documentation, can add a new table/csv file if needed.

torchgeo/models/satclip.py
isaaccorley: Expanded the SatCLIP/SIREN argument docs with shape info and parameter-effect intuition

torchgeo/models/satclip.py
isaaccorley: Added explicit input/output shapes for the SIREN layer/network docstring

docs/api/weights/location.csv
```diff
@@ -0,0 +1,2 @@
+Weight,Source,Citation,License,Training data,Image encoder,Embedding dimension
+SatCLIP_Weights.SATCLIP_VIT16_L40,`link <https://github.com/microsoft/satclip>`__,`link <https://arxiv.org/abs/2311.17179>`__,MIT,S2-100K,ViT-16,256
```
adamjstewart: Confirmed the license at https://huggingface.co/microsoft/SatCLIP-ViT16-L40

torchgeo/models/satclip.py
```diff
@@ -0,0 +1,466 @@
...
+    SATCLIP_VIT16_L40 = Weights(
+        url='https://huggingface.co/microsoft/SatCLIP-ViT16-L40/resolve/0ef2acc0b91d3c8cfdf4a0cc03207095989287ab/satclip-vit16-l40.ckpt',
+        transforms=nn.Identity(),
+        meta={
+            'dataset': 'S2-100K',
+            'model': 'satclip',
+            'publication': 'https://arxiv.org/abs/2311.17179',
+            'repo': 'https://github.com/microsoft/satclip',
+            'ssl_method': 'clip',
+            'image_encoder': 'vit16',
```
adamjstewart: Should this be vit16-l40? Not sure what the difference is.

adamjstewart (changes requested):
Still missing stuff like hubconf.py, see the full list from https://torchgeo.readthedocs.io/en/latest/user/contributing.html#models and let me know when it's complete.

The current stuff all looks good to me, need to run the slow tests locally to confirm though.

torchgeo/models/satclip.py
isaaccorley: Good catch — renamed to `vit16-l40` to match the released `microsoft/SatCLIP-ViT16-L40` model name. Note this `image_encoder` field is metadata only (the factory doesn't consume it), so it's purely descriptive.

isaaccorley:
Thanks for the review! Addressed the outstanding items:
- Added `satclip` to `hubconf.py` (imports + `__all__`) to complete the models checklist.
- Renamed the weight `image_encoder` metadata to `vit16-l40` to match the released model name.
- Rebrought the branch up to date with `main` (clean merge, no conflicts).

@nilsleh's docstring/shape feedback was addressed earlier in e9c34ae8. CI is green; ready for another look.

docs/api/weights/location.csv
adamjstewart:
```suggestion
SatCLIP_Weights.SATCLIP_VIT16_L40,`link <https://github.com/microsoft/satclip>`__,`link <https://arxiv.org/abs/2311.17179>`__,MIT,S2-100K,ViT16-L40,256
```

torchgeo/models/satclip.py
```diff
@@ -0,0 +1,466 @@
...
+            embed_dim: Number of output embedding features. Must be positive.
+            num_hidden_layers: Number of hidden SIREN layers. Higher values increase
+                network depth and coordinate-expression capacity. Must be positive.
+
+        Raises:
+            ValueError: If any dimension argument is not positive.
+        """
+        super().__init__()
+        if legendre_polys < 1:
+            raise ValueError('legendre_polys must be positive')  # pragma: no cover
```
adamjstewart: Nice try, but we need test coverage for these.

torchgeo/models/satclip.py
```diff
@@ -0,0 +1,466 @@
...
+        kwargs['legendre_polys'] = weights.meta['legendre_polys']
+        kwargs['capacity'] = weights.meta['capacity']
+        kwargs['embed_dim'] = weights.meta['embed_dim']
+        kwargs['num_hidden_layers'] = weights.meta['num_hidden_layers']
+
+    model = SatCLIP(*args, **kwargs)
+
+    if weights:
+        # SatCLIP releases ship as Lightning checkpoints with the location encoder
+        # weights under the 'model.location.nnet.*' namespace.
```
adamjstewart, line 492: Wonder if we should create our own redistribution of only the satclip weights.

adamjstewart (changes requested):

torchgeo/models/satclip.py
calebrob6: @konstantinklemmer correct me if I'm wrong, but I don't think there's a situation where these could be 0 or negative. That wouldn't make sense. Can we fix this with like a "positive int" type or something?

calebrob6:
Would be good to have this for @isaaccorley and friends ongoing work

torchgeo/models/satclip.py
```diff
@@ -0,0 +1,466 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+# Adapted from https://github.com/microsoft/satclip. Copyright (c) Microsoft Corporation.
+
+"""SatCLIP location encoder.
+
+Reference:
```
calebrob6: We usually have this in the docstring

torchgeo/models/satclip.py
```diff
@@ -0,0 +1,466 @@
...
+            Activated tensor.
+        """
+        return torch.sin(self.w0 * x)
+
+
+class Siren(nn.Module):
+    """SIREN linear layer with sine activation.
+
+    Reference:
+        Sitzmann et al., Implicit Neural Representations with Periodic
```
calebrob6: Can you add a doi?

torchgeo/models/satclip.py
```diff
@@ -0,0 +1,466 @@
...
+            Output tensor of shape ``(B, dim_out)``.
+        """
+        for layer in self.layers:
+            x = layer(x)
+        return self.last_layer(x)
+
+
+class SatCLIP(nn.Module):
+    """SatCLIP location encoder.
+
```
calebrob6, line 378: Cite Satclip here

calebrob6 (dismissed):

torchgeo/models/satclip.py
konstantinklemmer: Yes, you'd want at least 1

torchgeo/models/satclip.py
```diff
@@ -0,0 +1,465 @@
...
+    def __init__(self, legendre_polys: int) -> None:
+        """Initialize a new SphericalHarmonics instance.
+
+        Args:
+            legendre_polys: Number of Legendre polynomials. Higher values add
+                higher-frequency spatial basis functions and increase the output
+                feature dimension quadratically as ``legendre_polys ** 2``.
+        """
+        super().__init__()
+        self.L = int(legendre_polys)
```
adamjstewart: Seems like this cast shouldn't be required:

```suggestion
        self.L = legendre_polys
```

torchgeo/models/satclip.py
```diff
@@ -0,0 +1,465 @@
...
+        phi = torch.deg2rad(lon + 180)
+        theta = torch.deg2rad(lat + 90)
+        cos_theta = torch.cos(theta)
+        sin_theta = torch.sqrt(torch.clamp(1 - cos_theta * cos_theta, min=0))
+
+        dtype = x.dtype
+        device = x.device
+        batch_size = x.shape[0]
+        orders = self.orders.to(device=device)
+        order_values = orders.to(dtype=dtype)
```
adamjstewart: Does this result in `orders` and `self.orders` being on different devices? Could these two lines be combined into one like on line 95?

torchgeo/models/satclip.py
```diff
@@ -0,0 +1,465 @@
...
+        """
+        super().__init__()
+        if legendre_polys < 1:
+            raise ValueError('legendre_polys must be positive')
+        if capacity < 1:
+            raise ValueError('capacity must be positive')  # pragma: no cover
+        if embed_dim < 1:
+            raise ValueError('embed_dim must be positive')  # pragma: no cover
+        if num_hidden_layers < 1:
+            raise ValueError('num_hidden_layers must be positive')  # pragma: no cover
```
adamjstewart: All of these need test coverage unless you want to convert them to asserts. Some of these could be moved to the actual components that use them so that they benefit other future models that also reuse these components.

adamjstewart (commented):
I have a ton of other nitpicks (e.g., many lines are very long and end up being wrapped, making the math almost impossible to follow) but I don't want our implementation to diverge too much from the original. As soon as we actually have 100% test coverage, I think this will be ready.

torchgeo/models/satclip.py
isaaccorley: I'll change to asserts

torchgeo/models/satclip.py
```diff
@@ -0,0 +1,503 @@
...
+
+
+class SatCLIP_Weights(WeightsEnum):
+    """SatCLIP location encoder weights.
+
+    .. versionadded:: 0.10
+    """
+
+    SATCLIP_VIT16_L40 = Weights(
+        url='https://huggingface.co/microsoft/SatCLIP-ViT16-L40/resolve/0ef2acc0b91d3c8cfdf4a0cc03207095989287ab/satclip-vit16-l40.ckpt',
```
adamjstewart, line 445: Would be nice to get a checksum in this URL, we could rehost in our TorchGeo HF if needed.

torchgeo/models/satclip.py
```diff
@@ -0,0 +1,503 @@
...
+        # weights under the 'model.location.nnet.*' namespace.
+        checkpoint = weights.get_state_dict(progress=True, map_location='cpu')
+        prefix = 'model.location.'
+        state_dict = {
+            key[len(prefix) :]: value
+            for key, value in checkpoint['state_dict'].items()
+            if key.startswith(prefix + 'nnet.')
+        }
+        model.load_state_dict(state_dict, strict=True)
+        model.eval()
```
adamjstewart, line 501: Haven't seen this for other models, isn't this something the user should do instead?

adamjstewart (approved):
Only question is if we want to redistribute the weights in the correct format with the correct checksum in the filename.

torchgeo/models/satclip.py
isaaccorley: I'd prefer to rehost later because it requires voting and all that and don't want to hold up this PR

PR #3646, opened 2026-04-29
Normalize PASTIS plot RGB frames

robmarkcole:
## AI Usage Disclosure


adamjstewart (approved):
Can you share an example before and after plot to see how things compare?

robmarkcole:
pastis_comparison
In this spot check it doesn't make much of a difference. As claude just pointed out however, `quantile_normalization` could instead be called _per-frame_ rather than across the full time series. Too much computing?

adamjstewart:
I think per-time series is more stable than per-frame, but don't have strong opinions. Seems fine to me!

PR #3647, opened 2026-04-29
Support custom PASTIS Sentinel-2 bands

robmarkcole:
## AI Usage Disclosure

### DIFF
--- a/tests/datasets/test_pastis.py
+++ b/tests/datasets/test_pastis.py
@@ -52,6 +52,8 @@ class TestPASTIS:
         root = tmp_path
         bands = params['bands']
         mode = params['mode']
+        assert isinstance(bands, str)
+        assert isinstance(mode, str)
         transforms = nn.Identity()
 
         url = os.path.join('tests', 'data', 'pastis', 'PASTIS-R.zip')
--- a/tests/datasets/test_pastis.py
+++ b/tests/datasets/test_pastis.py
@@ -18,28 +18,14 @@ from torchgeo.datasets import PASTIS, PASTIS100, DatasetNotFoundError
 
 
 class TestPASTIS:
-    olmoearth_bands = (
-        'B02',
-        'B02',
-        'B03',
-        'B04',
-        'B05',
-        'B06',
-        'B07',
-        'B08',
-        'B8A',
-        'B8A',
-        'B11',
-        'B12',
-    )
-
     @pytest.fixture(
         params=product(
             [PASTIS, PASTIS100],
             [
-                {'folds': (1, 2), 'bands': 's2', 'mode': 'semantic'},
-                {'folds': (1, 2), 'bands': 's1a', 'mode': 'semantic'},
-                {'folds': (1, 2), 'bands': 's1d', 'mode': 'instance'},
+                {'folds': (1, 2), 'bands': PASTIS.s2_bands, 'mode': 'semantic'},
+                {'folds': (1, 2), 'bands': ('B04', 'B03', 'B02'), 'mode': 'semantic'},
+                {'folds': (1, 2), 'bands': PASTIS.s1a_bands, 'mode': 'semantic'},
+                {'folds': (1, 2), 'bands': PASTIS.s1d_bands, 'mode': 'instance'},
             ],
         )
     )
@@ -52,7 +38,6 @@ class TestPASTIS:
         root = tmp_path
         bands = params['bands']
         mode = params['mode']
-        assert isinstance(bands, str)
         assert isinstance(mode, str)
         transforms = nn.Identity()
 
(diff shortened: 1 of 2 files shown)


torchgeo/datasets/pastis.py
```diff
@@ -126,12 +126,13 @@ class PASTIS(NonGeoDataset):
...
         'instance': os.path.join('INSTANCE_ANNOTATIONS', 'INSTANCES_'),
     }
+    s2_bands = ('B02', 'B03', 'B04', 'B05', 'B06', 'B07', 'B08', 'B8A', 'B11', 'B12')
 
     def __init__(
         self,
         root: Path = 'data',
         folds: Sequence[int] = (1, 2, 3, 4, 5),
-        bands: str = 's2',
+        bands: str | Sequence[str] = 's2',
```
adamjstewart: I know it breaks backwards-compatibility, but I would personally prefer only supporting a list of band names instead of supporting s1a/s1d/s2a inputs. This would match the majority of our existing datasets. In this case I'm fine with both straight removal of the old feature or slow deprecation and removal later. Just not sure how to handle s1a and s1d having the same band names...

tests/datasets/test_pastis.py
```diff
@@ -18,6 +18,21 @@
 
 
 class TestPASTIS:
+    olmoearth_bands = (
```
adamjstewart: Not sure if we want/need to hard-code this to OlmoEarth

torchgeo/datasets/pastis.py
robmarkcole: Proposal: replace the existing s2_bands tuple and add two new ones:
```python
s2_bands:  ClassVar[tuple[str, ...]] = ('B02', 'B03', 'B04', 'B05', 'B06', 'B07', 'B08', 'B8A', 'B11', 'B12')
s1a_bands: ClassVar[tuple[str, ...]] = ('S1A_VV', 'S1A_VH', 'S1A_VVVH')
s1d_bands: ClassVar[tuple[str, ...]] = ('S1D_VV', 'S1D_VH', 'S1D_VVVH')
```
VVVH (no slash) keeps names as plain identifiers. The `S1A_/S1D_` prefix is the only way to resolve the naming collision without a separate sensor param.

__init__ signature — drop the str option entirely:
```python
bands: Sequence[str] = s2_bands,
```

Sensor inference in __init__ — replace the current `isinstance(bands, str)` block:
```python
s2_set  = set(self.s2_bands)
s1a_set = set(self.s1a_bands)
s1d_set = set(self.s1d_bands)
bands_set = set(bands)

if bands_set <= s2_set:
    self.image_key = 's2'
    all_bands = self.s2_bands
elif bands_set <= s1a_set:
    self.image_key = 's1a'
    all_bands = self.s1a_bands
elif bands_set <= s1d_set:
    self.image_key = 's1d'
    all_bands = self.s1d_bands
else:
    raise ValueError(f"bands must be a subset of s2_bands, s1a_bands, or s1d_bands; got {bands}")

self.bands = tuple(bands)
self.band_indices = [all_bands.index(b) for b in self.bands]
```

Drop the `band_indices = None` fast-path — always compute indices. `_load_image` then always slices, which is a no-op cost for the full-band case.

**What this doesn't support (intentionally):** mixing bands across sensors (e.g. S1A + S2 in one call) — that would require returning two tensors with different time axes (65 vs 43 timesteps), which is a much bigger design change and not currently supported anyway.

**Removal vs deprecation:** given the dataset has been around since 0.5 and the old 's1a'/'s1d'/'s2' strings are prominently documented, a one-version deprecation warning (DeprecationWarning + point to the new band constants) is safer than a hard break. But if you want clean removal, the blast radius is small — just tests and docs.

tests/datasets/test_pastis.py
robmarkcole: prefer a more generic name, `duplicated_bands` or something similar?

torchgeo/datasets/pastis.py
adamjstewart: `S1A_` prefix seems fine with me. I don't fully understand `VVVH` though.

tests/datasets/test_pastis.py
adamjstewart: Sure, or just a parametrize with RGB-only, S1A, S1D, and S2

torchgeo/datasets/pastis.py
robmarkcole: VVVH is the VV/VH backscatter ratio expressed in dB, alternatively we could use `VV_VH` e.g. `S1A_VV_VH`

adamjstewart: Yeah I like that better

calebrob6:
Can you post some example plots from the different band sets?

robmarkcole:
@calebrob6 is the objective to compare how similar or not the bands are? For context this PR is to support olmo which requires only a subset of the bands

calebrob6 (approved):

calebrob6:
Nevermind, I tested and it works like I expected :)

PR #3651, opened 2026-04-29
Plotting normalization: ignore nodata pixels

adamjstewart:
For images containing nodata pixels, these zeros completely throw off our math during normalization. For reproducibility, the below figures are from our I/O Bench dataset.

### Before

before

### After

after

## AI Usage Disclosure

### DIFF
--- a/torchgeo/datasets/utils.py
+++ b/torchgeo/datasets/utils.py
@@ -28,6 +28,7 @@ import pandas as pd
 import rasterio
 import shapely.affinity
 import torch
+from numpy.typing import NDArray
 from pandas import Timedelta, Timestamp
 from rasterio import Affine
 from shapely import Geometry
@@ -797,11 +798,11 @@ def rgb_to_mask(
 
 @deprecated('Use torchgeo.datasets.utils.quantile_normalization instead')
 def percentile_normalization(
-    img: np.typing.NDArray[np.int_],
+    img: NDArray,
     lower: float = 2,
     upper: float = 98,
     axis: int | Sequence[int] | None = None,
-) -> np.typing.NDArray[np.int_]:
+) -> NDArray:
     """Applies percentile normalization to an input image.
 
     Specifically, this will rescale the values in the input such that values <= the
@@ -818,9 +819,15 @@ def percentile_normalization(
     Returns:
         normalized version of ``img``
 
+    Raises:
+        AssertionError: If *lower* is higher than *upper*.
+
     .. versionadded:: 0.2
     .. versiondeprecated:: 0.10
     """
+    if not np.any(img):
+        return img
+
     assert lower < upper
     lower_percentile = np.percentile(img[img != 0], lower, axis=axis)
     upper_percentile = np.percentile(img[img != 0], upper, axis=axis)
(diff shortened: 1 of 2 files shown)


adamjstewart:
Question: should we error on all zero images (like the one failing test) or gracefully return all zeros?

robmarkcole:
@adamjstewart which failing test? An all zero image could be confusing for users, an error feels appropriate

adamjstewart:
It was the Copernicus-Pretrain test that was failing, I guess our fake image for that one is all zeros.

All zero image isn't uncommon (especially because we currently sample from nodata regions). I didn't want an error in the middle of an epoch (we do plot during validation in our trainers).

robmarkcole:
Maybe just a warning/info on plotting then, dont want to terminate training over that

calebrob6:
You can't assume that nodata is 0 and doing a mask like this can take a relatively long time at S2 scene level. Would suggest allowing a `nodata_mask` or something if you want to subset which pixels the stats are computed over.

Also, this is why fixed RGB normalization for different platforms is useful. E.g. if I have the S2 RGB bands I can just divide by 3000 or 4000 and clip to get something that looks good regardless of nodata, scene composition, etc.

adamjstewart:
> You can't assume that nodata is 0

This is a valid point. We've talked about adding a `nodata` parameter to `RasterDataset` (default to 0) that users could override. For XarrayDataset, we already hardcode `nodata=0`, but I don't think we yet do this in `RasterDataset`.

> doing a mask like this can take a relatively long time at S2 scene level.

This is just a random example, I don't expect users to do this very often. This method is only used during plotting, never during data augmentation.

> Would suggest allowing a nodata_mask or something if you want to subset which pixels the stats are computed over.

I think this would just move the complexity elsewhere, right?

calebrob6:
> I think this would just move the complexity elsewhere, right?

Yeah I guess this doesn't help as its used in all the plot methods -- each would just need to compute masks there. My thought was that making it explicit forces users to realize that there is a mask happening. If this is just on patches it won't be an issue.


We can propagate nodatas later. Fine to merge

calebrob6 (dismissed):

torchgeo/datasets/utils.py
```diff
@@ -818,12 +819,18 @@ def percentile_normalization(
...
     .. versionadded:: 0.2
     .. versiondeprecated:: 0.10
     """
+    if not np.any(img):
+        return img
+
     assert lower < upper
-    lower_percentile = np.percentile(img, lower, axis=axis)
-    upper_percentile = np.percentile(img, upper, axis=axis)
+    lower_percentile = np.percentile(img[img != 0], lower, axis=axis)
```
robmarkcole: Could we not pass in an arg for the nodata value, default to 0?

robmarkcole (dismissed):

torchgeo/datasets/utils.py
adamjstewart: Done, hopefully this also addresses one of @calebrob6's concerns as well.

robmarkcole (approved):

PR #3652, opened 2026-04-29
Plotting: use 2–98% normalization instead of 0–100%

adamjstewart:
0–100% tends to be too sensitive to outliers (both nodata pixels and clouds). Use `quantile_normalization` with it's 2–98% normalization to match QGIS and other tools.

## AI Usage Disclosure


torchgeo/datasets/idtrees.py
```diff
@@ -507,13 +514,11 @@ def plot(
...
 
-        def normalize(x: Tensor) -> Tensor:
-            return (x - x.min()) / (x.max() - x.min())
-
         ncols = 3
 
-        hsi = normalize(sample['hsi'][hsi_indices, :, :]).permute((1, 2, 0)).numpy()
-        chm = normalize(sample['chm']).permute((1, 2, 0)).numpy()
+        hsi = sample['hsi'][hsi_indices, :, :].permute((1, 2, 0))
+        hsi = quantile_normalization(hsi)
```
robmarkcole, line 520: why is this not a 1 liner?

robmarkcole:
LGTM

torchgeo/datasets/idtrees.py
adamjstewart: Too long to fit on 1 line without wrapping onto 3 lines

calebrob6:
Is there a particular scenario where the current normalization isn't working for you?

adamjstewart:
No specific examples, but 2–98% tends to be more robust to outliers than 0–100%. Slower, but this is only used occasionally via plotting, never during data augmentation.

robmarkcole:
I've also tended towards 2/98. Ideally you have a cloud mask available so these are computed over clear pixels only, since it tends to be clouds which mess up the contrast

adamjstewart:
Not a bad idea, especially in combination with the nodata mask proposed by Caleb in #3651. But let's keep this PR simple and add complexity later. I have a lot of long-term plans/ideas for visualization if I can find someone to work on it.

calebrob6:
> No specific examples, but 2–98% tends to be more robust to outliers than 0–100%. 

I'm right on the border of:
- "if it isn't broke then don't fix it" -- I'd ideallllllllllllly like to go visualize examples of each of these datasets to make sure that this change makes sense, what are the before/after examples from this change. Did we make choose min/max for a specific reason in any of these dataset? (rhetorical question -- don't spend time actually looking this up). The sensors, dtypes, dynamic ranges, and contents of the datasets we support vary enormously -- there's no reason why we need the same approach for plotting all of these, especially if people have done work while implementing the plot() methods to make sure the visualizations already work. 
- 2nd-98th percentile is _probablyyy_ fine.

As I'm on the borderline I'm fine to merge :)

adamjstewart:
I'm also on the fence. For Landsat in particular, I wanted this in combination with #3651 to get that pretty picture. I'm fine with removing any datasets we have concerns about. 

I'm also fine with saving plotting for a bigger refactor later. I want to try to make a generic `plot` method at the base class level that grabs min/max or mean/std from the dataset or 2–98% from the data and works for most datasets. That would allow us to remove most code duplication, but would require a ton of manual data download and visualization to check that things are okay.

robmarkcole (approved):

robmarkcole:
I vote that we move ahead with this - at the very least it simplifies the code

ashnair1 (approved):

PR #3656, opened 2026-04-30
GeoCLIP: add new model

adamjstewart:
Includes several improvements over the [original implementation](https://github.com/VicenteVivan/geo-clip):

- [x] Does not download anything without user consent: https://github.com/VicenteVivan/geo-clip/issues/22
- [x] Does not store binary data/weight files in the repo
- [x] Complete static type hints
- [ ] Complete unit test coverage
- [ ] Complete API documentation

@VicenteVivan

## AI Usage Disclosure

### DIFF
--- a/torchgeo/models/geoclip.py
+++ b/torchgeo/models/geoclip.py
@@ -1,0 +1,533 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+
+# Adapted from https://github.com/VicenteVivan/geo-clip/tree/main/geoclip/model.
+# Copyright (c) 2024 Vicente Vivanco
+
+"""GeoCLIP models."""
+
+import os
+
+import numpy as np
+import pandas as pd
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from PIL import Image
+from torch import Tensor
+from transformers import AutoProcessor, CLIPModel
+
+# Constants
+A1 = 1.340264
+A2 = -0.081106
+A3 = 0.000893
+A4 = 0.003796
+SF = 66.50336
+
+
+def load_gps_data(csv_file: str) -> Tensor:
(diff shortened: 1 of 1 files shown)


torchgeo/models/geoclip.py
```diff
@@ -0,0 +1,588 @@
...
+    """
+    model = GeoCLIP(*args, **kwargs)
+
+    if weights:
+        model.image_encoder.mlp.load_state_dict(
+            GeoCLIP_Weights.GEOCLIP_IMAGE_ENCODER_MLP.get_state_dict(progress=True),
+            strict=True,
+        )
+        model.image_encoder.CLIP = CLIPModel.from_pretrained(
+            'openai/clip-vit-large-patch14'
```
adamjstewart, line 573: These will have to be added to the enum, there's no way to test this without internet download.

torchgeo/models/geoclip.py
```diff
@@ -0,0 +1,588 @@
...
+        Args:
+            v: Input tensor of shape :math:`(N, *, \text{input_size})`.
+
+        Returns:
+            Mapped tensor of shape :math:`(N, *, 2 \cdot m \cdot \text{input_size})`.
+        """
+        return positional_encoding(v, self.sigma, self.m)
+
+
+class GeoCLIP_Weights(WeightsEnum):
```
adamjstewart, line 493: Similar to Tessera, the model weights are split into several components. My current strategy is to make `weights` a boolean instead of an enum, but this doesn't match any of our other models. Alternative is to create a new combined weight for everything, although some parts are under different licenses. @isaaccorley what do you think? Would love to move from the torchvision API to the timm API sooner rather than later if that helps.

adamjstewart:
Going to close this, I think it would be better to contribute the features we want to https://github.com/VicenteVivan/geo-clip and add that as a dependency instead.

PR #3660, opened 2026-05-03
Add TimeSen2Crop Sentinel-2 time-series crop classification dataset

### LINKED ISSUE #3642: [FR] TimeSen2Crop 1D time-series dataset
### Summary

A Million Labeled Samples Dataset of Sentinel 2 Image Time Series for Crop Type Classification

Available in a few locations:
- https://huggingface.co/datasets/monster-monash/TimeSen2Crop (needs verification)
- https://zenodo.org/records/4715631 (official zenodo)
- https://github.com/elliotvincent/agriitsc (official Gdrive

Image

Image

omkar-334:
## Description 
Related issue - #3642 - I was looking for issues to contribute and this seemed interesting...

The dataset has 3 sources - huggingface, zenodo and google drive.  
The problem with huggingface source is that it relied on a python loading script, which is now not supported in the latest version of `datasets`. It also drops the `Other Crops` class and it linearly interpolates each per-tile series to a fixed 365-day grid, collapsing the irregular acquisition dates.

So I went ahead with the zenodo source. There is one issue here - Variable-length sequences break default DataLoader collation. This seems the issue that the huggingface source was fixing, but im not sure. I've added a docstring to convey the choices a user can take - The docstring shows two fixes: pass `pad_to=<int>` for static padding, or pass a custom  
  `collate_fn` (e.g., `torch.nn.utils.rnn.pad_sequence`) for per-batch padding.  
  
  This is a comparison of huggingface and our approaches - 
  
image

## Real-format discrepancies vs. the description PDF (verified by downloading)

After downloading the real Zenodo zip, three things differ from the PDF figure and have been corrected in the implementation:

1. Top-level extracted folder is `TimeSen2Crop/`, not `Dataset/` as the PDF figure shows.
2. Class subfolders are bare integer-named directories (`0`, `1`, ..., `15`), not `Class 0`..`Class 15`.
3. Both `dates.csv` and per-sample CSVs include a header row (`acquisition_date` for dates, `B1,B2,...,B9,Flag` for samples). The CSV column labels are sequential placeholders; the actual S2 band identities (B2, B3, B4, B5, B6, B7, B8A, B11, B12) come from position per the PDF text.

## References

- Issue: https://github.com/torchgeo/torchgeo/issues/3642
- Paper / DOI: https://doi.org/10.1109/JSTARS.2021.3073965
- Zenodo: https://zenodo.org/records/4715631
- Description PDF: https://zenodo.org/api/records/4715631/files/TimeSen2Crop_Description.pdf/content
- Author downstream loader: https://github.com/elliotvincent/agriitsc/blob/main/src/datasets/ts2c.py

## AI Usage Disclosure

Closes #3642


robmarkcole:
@omkar-334 variable length sequences are supported in https://github.com/torchgeo/torchgeo/pull/3280 suggest after that is merged, to evaluate this dataset using it

omkar-334:
hey @robmarkcole I trained ConvLSTM (#3280) on tiles 33TUN + 33TWN, 80k subsample,
hidden_dim=32, batch size 256, 8 epochs, Adam lr=1e-3, CPU.

| epoch | train loss | train acc | val loss | val acc | wall (s) |
| --- | --- | --- | --- | --- | --- |
| 0 | 2.107 | 0.306 | 1.806 | 0.386 | 167 |
| 1 | 1.641 | 0.476 | 1.354 | 0.547 | 168 |
| 2 | 1.229 | 0.579 | 1.235 | 0.565 | 173 |
| 3 | 1.086 | 0.621 | 1.025 | 0.645 | 172 |
| 4 | 0.991 | 0.657 | 0.951 | 0.672 | 168 |
| 5 | 0.934 | 0.675 | 0.943 | 0.664 | 169 |
| 6 | 0.882 | 0.693 | 0.846 | 0.708 | 175 |
| 7 | 0.845 | 0.708 | 0.829 | 0.715 | 169 |

robmarkcole:
Nice! You could checkout https://github.com/torchgeo/torchgeo/pull/3671 and report some test metrics to compare against literature? Could also add a datamodule to this PR if there is a standard train/val/test split published

torchgeo/datasets/timesen2crop.py
```diff
@@ -0,0 +1,392 @@
...
+            RGBBandsMissingError: if any of B4, B3, B2 are missing from
+                ``self.bands``.
+        """
+        try:
+            rgb_indices = [self.bands.index(b) for b in self.rgb_bands]
+        except ValueError as exc:
+            raise RGBBandsMissingError() from exc
+
+        image = sample['image'].numpy()
+        rgb = np.clip(image[:, rgb_indices] / 3000.0, 0, 1)
```
robmarkcole, line 378: Not familiar with the data, why norm to 3000?

omkar-334:
> Could also add a datamodule to this PR if there is a standard train/val/test split published  

https://zenodo.org/records/4715631 - the zenodo version only includes the raw data, no splits. The HF mirror includes splits, but includes the caveats i mentioned above.  

> Nice! You could checkout #3671 and report some test metrics to compare against literature?

I didn;t understand this.. could you explain a bit more?  #3671 is a segmentation task and this dataset is for classification right?  
Do you mean something like this - 
1. Add a SpatioTemporalClassificationTask (siimiliar to #3671's segmentation task) that takes (B, T, C) and batch['label'], returns (B,  num_classes).
  2. Add a TimeSen2CropDataModule parameterized by `held_out_tile`. The held-out tile is test; the other 14 tiles form train, with 10% as eval
  3. Run the full 15-fold sweep externally (one Lightning run per fold) and report mean / per-fold OA and macro-F1 alongside the paper's published numbers

robmarkcole:
- Sorry which `caveats`? 
- My mistake SpatioTemporalClassificationTask cannot be used then. Which trainer did you use for your ConvLSTM experiment? Comparison to the  paper's published numbers would be nice validation!

omkar-334:
> The dataset has 3 sources - huggingface, zenodo and google drive. The problem with huggingface source is that it relied on a python loading script, which is now not supported in the latest version of `datasets`. It also drops the `Other Crops` class and it linearly interpolates each per-tile series to a fixed 365-day grid, collapsing the irregular acquisition dates.
> 
> So I went ahead with the zenodo source. There is one issue here - Variable-length sequences break default DataLoader collation. This seems the issue that the huggingface source was fixing, but im not sure. I've added a docstring to convey the choices a user can take - The docstring shows two fixes: pass `pad_to=<int>` for static padding, or pass a custom `collate_fn` (e.g., `torch.nn.utils.rnn.pad_sequence`) for per-batch padding.
> 
> This is a comparison of huggingface and our approaches -

@robmarkcole

omkar-334:
> * Which trainer did you use for your ConvLSTM experiment?

```python
from torchgeo.datasets import TimeSen2Crop
from torchgeo.models import ConvLSTM
```  
I loaded the model, dataset and just used a custom pytorch training loop for my experiment ... I think `ClassificationTask` is not suitable for this. `SpatioTemporalSegmentationTask` is the closest but it is for segmentation... do you think i should create a new task for this?

torchgeo/datasets/timesen2crop.py
omkar-334: I dont think it is specific to this data. `3000` is what most of the other Sentinel-2 datasets in the repo use for plotting (eurosat.py, cropharvest.py, benin_cashews.py, ftw.py, ssl4eo.py, chesapeake.py, cloud_cover.py) so i just followed that convention

robmarkcole:
RE trainer I think you want https://github.com/torchgeo/torchgeo/pull/2639

omkar-334:
hey @robmarkcole i looked at the time series issue roadmap in #2382, i think the approprate trainers for this dataset would be `TemporalClassificationTask` or `SpatioTemporalClassificationTask`(#3591).... 
I can try to contribute towards `TemporalClassificationTask` based on the structure of other trainers, what do you think?

robmarkcole:
@omkar-334 my suggestion was rather that you could use a torchgeo trainer rather than bespoke code for the evaluation, nice to close the loop!

docs/api/datasets/benchmark.csv
```diff
@@ -63,6 +63,7 @@ Dataset,Task,Source,# Samples,# Classes,Size (px),Resolution (m),Bands,License
 :ref:`South Africa Crop Type`,S,Sentinel-2,,10,"256x256",10,MSI,CC-BY-4.0
 :ref:`Substation`,S,OpenStreetMap & Sentinel-2,27K,2,228x228,10,MSI,CC-BY-4.0
 :ref:`SustainBench Crop Yield`,R,MODIS,11k,-,32x32,-,MSI,CC-BY-SA-4.0
+:ref:`TimeSen2Crop`,"C, T",Sentinel-2,"1.13M",16,1x1,10,MSI,CC-BY-4.0
```
adamjstewart, line 66: Confirmed the license from https://zenodo.org/records/4715631

torchgeo/datasets/timesen2crop.py
```diff
@@ -0,0 +1,392 @@
...
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://doi.org/10.1109/JSTARS.2021.3073965
+
+    .. versionadded:: 0.10
+    """
+
+    url = 'https://zenodo.org/records/4715631/files/TimeSen2Crop.zip'
+    filename = 'TimeSen2Crop.zip'
+    md5 = 'b5b7aad3fef192e78252e11c9a0e5cb8'
```
adamjstewart, line 89: If you happen to have this file download locally, can you switch this from MD5 to SHA256? We're trying to do this for all datasets as MD5 is not cryptographically secure.

torchgeo/datasets/timesen2crop.py
```diff
@@ -0,0 +1,392 @@
...
+
+    def __init__(
+        self,
+        root: Path = 'data',
+        tiles: Sequence[str] = valid_tiles,
+        bands: Sequence[str] = all_bands,
+        pad_to: int | None = None,
+        transforms: Callable[[Sample], Sample] | None = None,
+        download: bool = False,
+        checksum: bool = False,
```
adamjstewart, line 146: Let's default to True, I'm working on changing all existing datasets in a different PR.

tests/datasets/test_timesen2crop.py
```diff
@@ -0,0 +1,169 @@
...
+TEST_TILES = ('33TUN', '2019_33UVP')
+TEST_TILE_T = {'33TUN': 5, '2019_33UVP': 7}
+
+
+class TestTimeSen2Crop:
+    @pytest.fixture
+    def dataset(self, monkeypatch: MonkeyPatch, tmp_path: Path) -> TimeSen2Crop:
+        url = os.path.join('tests', 'data', 'timesen2crop', 'TimeSen2Crop.zip')
+        monkeypatch.setattr(TimeSen2Crop, 'url', url)
+        monkeypatch.setattr(TimeSen2Crop, 'md5', '63729536a368e643b5819d1cdf101a37')
```
adamjstewart, line 25: Technically don't need to patch md5 if you test `checksum=False` instead

tests/datasets/test_timesen2crop.py
```diff
@@ -0,0 +1,169 @@
...
+        rogue = os.path.join(
+            dataset.root, TimeSen2Crop.extracted_dirname, '33TUN', '99'
+        )
+        os.makedirs(rogue)
+        with open(os.path.join(rogue, '0.csv'), 'w') as f:
+            f.write('B1,B2\n0,0\n')
+        monkeypatch.setattr(TimeSen2Crop, 'valid_tiles', TEST_TILES)
+        TimeSen2Crop(root=dataset.root, tiles=TEST_TILES)
+
+    def test_build_cache_skips_empty_class_dir(
```
adamjstewart, line 161: This is a lot more tests that most other datasets. Most dataset tests are under 100 lines of code. Can we reduce these to the bare minimum required to achieve coverage?

torchgeo/datasets/timesen2crop.py
```diff
@@ -0,0 +1,392 @@
...
+    which requires identical shapes and raises
+    ``RuntimeError: stack expects each tensor to be equal size``). Two
+    options:
+
+    1. Pass ``pad_to=<int>`` to right-pad every sample to a fixed length.
+       Padded time steps are marked in ``mask`` with
+       :attr:`PADDING_VALUE` so loss masking is easy.
+    2. Pass a custom ``collate_fn`` to the DataLoader, for example using
+       :func:`torch.nn.utils.rnn.pad_sequence`, which only pads to the
+       longest sequence in each batch.
```
adamjstewart, line 78: Probably too much detail, we have `torchgeo.datasets.utils.pad_across_batches` for this.

torchgeo/datasets/timesen2crop.py
```diff
@@ -0,0 +1,392 @@
...
+    * https://doi.org/10.1109/JSTARS.2021.3073965
+
+    .. versionadded:: 0.10
+    """
+
+    url = 'https://zenodo.org/records/4715631/files/TimeSen2Crop.zip'
+    filename = 'TimeSen2Crop.zip'
+    md5 = 'b5b7aad3fef192e78252e11c9a0e5cb8'
+    extracted_dirname = 'TimeSen2Crop'
+    cache_dirname = 'cache'
```
adamjstewart, line 91: Why do we need a separate cache dir?

torchgeo/datasets/timesen2crop.py
```diff
@@ -0,0 +1,392 @@
...
+        '33UUP',
+        '33UVP',
+        '33UWP',
+        '33UWQ',
+        '33UXP',
+    )
+
+    #: Mask value used for time steps that fall in the right-padded region
+    #: when ``pad_to`` is set, distinct from the four condition codes (0..3).
+    PADDING_VALUE: ClassVar[int] = 4
```
adamjstewart, line 136: Remove this, this should happen in a collation function so we don't duplicate code

torchgeo/datasets/timesen2crop.py
```diff
@@ -0,0 +1,392 @@
...
+
+    def __init__(
+        self,
+        root: Path = 'data',
+        tiles: Sequence[str] = valid_tiles,
+        bands: Sequence[str] = all_bands,
+        pad_to: int | None = None,
```
adamjstewart, line 143: Remove padding

torchgeo/datasets/timesen2crop.py
```diff
@@ -0,0 +1,392 @@
...
+        'Rapeseed',
+        'Beet',
+        'Spring Cereals',
+        'Winter Wheat',
+        'Winter Triticale',
+        'Permanent Plantation',
+        'Other Crops',
+    )
+
+    valid_tiles: ClassVar[tuple[str, ...]] = (
```
adamjstewart, line 115: ClassVar not needed

torchgeo/datasets/timesen2crop.py
```diff
@@ -0,0 +1,392 @@
...
+
+    #: Mask value used for time steps that fall in the right-padded region
+    #: when ``pad_to`` is set, distinct from the four condition codes (0..3).
+    PADDING_VALUE: ClassVar[int] = 4
+
+    def __init__(
+        self,
+        root: Path = 'data',
+        tiles: Sequence[str] = valid_tiles,
+        bands: Sequence[str] = all_bands,
```
adamjstewart, line 142: Latest version of ruff will probably complain about these lines

torchgeo/datasets/timesen2crop.py
```diff
@@ -0,0 +1,392 @@
...
+                False.
+        """
+        for tile in tiles:
+            assert tile in self.valid_tiles, (
+                f'Invalid tile {tile!r}. Valid tiles: {list(self.valid_tiles)}.'
+            )
+        for band in bands:
+            assert band in self.all_bands, (
+                f'Invalid band {band!r}. Valid bands: {list(self.all_bands)}.'
+            )
```
adamjstewart, line 179: Could simplify:
```python
assert set(tiles) <= set(self.valid_tiles), '...'
assert set(bands) <= set(self.all_bands), '...'
```

torchgeo/datasets/timesen2crop.py
```diff
@@ -0,0 +1,392 @@
...
+        return len(self.index)
+
+    def __getitem__(self, index: int) -> Sample:
+        """Return the sample at the given flat index.
+
+        Args:
+            index: index of the sample.
+
+        Returns:
+            Dictionary with ``image`` (T, B), ``mask`` (T,), and ``label`` ().
```
adamjstewart, line 227: What is "B" here?

torchgeo/datasets/timesen2crop.py
```diff
@@ -0,0 +1,392 @@
...
+        self._build_cache()
+
+    def _download(self) -> None:
+        """Download the dataset archive."""
+        download_url(
+            self.url,
+            self.root,
+            filename=self.filename,
+            md5=self.md5 if self.checksum else None,
+        )
```
adamjstewart, line 292: There is a `download_and_extract_archive` function that could combine these

torchgeo/datasets/timesen2crop.py
```diff
@@ -0,0 +1,392 @@
...
+
+    def _build_cache(self) -> None:
```
adamjstewart, line 294: Why is this needed, and how long does it take? I would rather directly read and use CSV files using pandas to make things simpler and easier to maintain.

torchgeo/datasets/timesen2crop.py
```diff
@@ -0,0 +1,392 @@
...
+        rgb = np.clip(image[:, rgb_indices] / 3000.0, 0, 1)
+
+        fig, ax = plt.subplots(figsize=(8, 2))
+        ax.imshow(rgb[None, ...], aspect='auto')
+        ax.set_yticks([])
+        ax.set_xlabel('Acquisition step')
+
+        if show_titles:
+            label_idx = int(sample['label'])
+            ax.set_title(f'Crop: {self.classes[label_idx]}')
```
adamjstewart, line 387: Should we also plot the predicted class?

torchgeo/datasets/timesen2crop.py
```diff
@@ -0,0 +1,392 @@
...
+
+                arrays[f'bands_{class_id}'] = np.stack(bands_stack, axis=0)
+                arrays[f'mask_{class_id}'] = np.stack(mask_stack, axis=0)
+                arrays[f'count_{class_id}'] = np.asarray(len(csv_paths))
+
+            # ty narrows **kwargs against the typed `allow_pickle` parameter,
+            # which is unrelated to the array kwargs we actually pass.
+            np.savez(cache_path, **arrays)  # ty: ignore[invalid-argument-type]
+
+    def plot(
```
adamjstewart, line 355: Can you add an example plot of the real data to the PR description so we can see what it looks like?

adamjstewart (changes requested):

PR #3663, opened 2026-05-04
Add utilities for computing valid-data footprints from raster masks

adriantre:
Adds `clean_binary_mask`, `calculate_valid_footprint_from_binary_mask`, and `get_valid_footprint_from_datasource` to `torchgeo.datasets.utils`. These allow a RasterDataset subclass to index each file by its precise valid-data footprint rather than its bounding box, which matters when scenes have significant nodata regions (e.g. satellite swath edges, cloud masks or after re-projection). Intended for use with RasterDataset._footprint_from_datasource in #2903 

For large datasets, footprints can be pre-computed once and stored in a GeoPackage, avoiding the per-run overhead of reading and processing every raster mask. Suggested implementation is found in docstring of `get_valid_footprint_from_datasource`.

## AI Usage Disclosure

### DIFF
--- a/tests/datasets/test_utils.py
+++ b/tests/datasets/test_utils.py
@@ -15,6 +15,9 @@ import pytest
 import torch
 from numpy.typing import NDArray
 from pytest import MonkeyPatch
+from rasterio import MemoryFile
+from rasterio.transform import from_origin
+from shapely import MultiPolygon, Polygon, box
 from torch import Tensor
 
 from torchgeo.datasets import BoundingBox, DependencyNotFoundError
@@ -23,12 +26,14 @@ from torchgeo.datasets.utils import (
     Sample,
     array_to_tensor,
     check_integrity,
+    clean_binary_mask,
     concat_samples,
     disambiguate_timestamp,
     download_and_extract_archive,
     download_url,
     extract_archive,
     find_files,
+    get_valid_footprint_from_datasource,
     lazy_import,
     merge_samples,
     pad_across_batches,
@@ -692,3 +697,83 @@ class TestFindFiles:
         """A path that does not exist inside an archive resolves to nothing."""
         _, archive = temp_archive
         assert find_files(f'/vsizip/{archive}/non_existing.tif') == []
+
+
+def create_test_raster_file(
+    width: int = 100,
+    height: int = 100,
+    nodata_value: float = 0,
+    crs: str = 'EPSG:32633',
+    num_bands: int = 1,
+    with_half_nodata: bool = False,
+) -> MemoryFile:
+    """Creates a synthetic raster file for testing."""
+    base_data = np.ones((height, width), dtype=np.uint8)
+
+    if with_half_nodata:
+        for row in range(height):
+            for col in range(width):
+                if row + col < width:
+                    base_data[row, col] = nodata_value
+
+    data = np.stack([base_data] * num_bands, axis=0)
+    # Realistic Sentinel-2-like origin and 10 m resolution in UTM zone 32N
+    transform = from_origin(600000, 5700000, 10, 10)
+
+    memfile = MemoryFile()
+    with memfile.open(
+        driver='GTiff',
+        height=height,
+        width=width,
+        count=num_bands,
+        dtype=data.dtype,
+        transform=transform,
+        crs=crs,
+        nodata=nodata_value,
+    ) as dataset:
+        dataset.write(data)
+
+    return memfile
+
+
+@pytest.mark.parametrize('num_bands', [1, 3])
+@pytest.mark.parametrize('with_half_nodata', [True, False])
+def test_calc_valid_data_footprint_from_raster_mask(
+    num_bands: int, with_half_nodata: bool
+) -> None:
+    with (
+        create_test_raster_file(
+            num_bands=num_bands, with_half_nodata=with_half_nodata
+        ) as memfile,
+        memfile.open() as dataset,
+    ):
+        # Verify clean_binary_mask's 3D OR logic matches rasterio's own mask combination
+        assert np.array_equal(
+            clean_binary_mask(dataset.dataset_mask()),
+            clean_binary_mask(dataset.read_masks()),
+        )
+
+        footprint = get_valid_footprint_from_datasource(dataset)
+
+        assert isinstance(footprint, Polygon | MultiPolygon)
+        assert footprint.is_valid
+
+        raster_extent = box(*dataset.bounds)
+        expected_area = raster_extent.area
+
+        if with_half_nodata:
+            expected_area /= 2
+
+            relative_error = abs(footprint.area - expected_area) / expected_area
+            # Theoretical bound for triangle: ~2√2/N where N is the raster width in pixels.
+            # For N=100 this is ~2.83%; we allow 4% for floating-point margin.
+            assert relative_error < 0.04
+        else:
+            assert footprint.area == expected_area
+
+
+def test_get_valid_footprint_all_nodata_warns() -> None:
+    with create_test_raster_file(nodata_value=1) as memfile, memfile.open() as dataset:
+        with pytest.warns(UserWarning, match='All pixels are nodata'):
+            footprint = get_valid_footprint_from_datasource(dataset)
+        assert footprint.is_empty
--- a/tests/datasets/test_utils.py
+++ b/tests/datasets/test_utils.py
@@ -704,6 +704,8 @@ def create_test_raster_file(
     height: int = 100,
     nodata_value: float = 0,
     crs: str = 'EPSG:32633',
+    origin: tuple[float, float] = (600000, 5700000),
+    res: float = 10,
     num_bands: int = 1,
     with_half_nodata: bool = False,
 ) -> MemoryFile:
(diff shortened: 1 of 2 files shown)


tests/datasets/test_utils.py
```diff
@@ -627,3 +632,99 @@ def test_pad_across_batches() -> None:
     assert out['image'].shape[1] == 5
     assert out['bbox_xyxy'].shape[0] == len(batch)
     assert out['label'].shape[0] == len(batch)
+
+
+def create_test_raster_file(
```
adriantre: Can now use sentinel2 test data-folder as these contain nodata-region.

torchgeo/datasets/utils.py
```diff
@@ -990,3 +992,149 @@ def convert_poly_coords(
...
+    simplification_tolerance = 2 * raster_resolution_x
+    return cast(
+        Polygon | MultiPolygon, vector_footprint.simplify(simplification_tolerance)
+    )
+
+
+def get_valid_footprint_from_datasource(
+    src: DatasetReader | WarpedVRT,
+) -> MultiPolygon | Polygon:
+    """Compute the valid data footprint of a raster dataset.
```
adriantre: @adamjstewart I could also see this as a part of the docs/tutorials guides, alongside what we discussed in #2903

adamjstewart: As long as the tutorials remain simple, I don't want to scare away users who think they need to understand all of this code to contribute.

adriantre: I meant this docstring, consider it an MVP what I would put in the tutorial. Did you get a chance to look at it?

adamjstewart: Ah okay. Docstring is quite long. I would add the first sentence and some example code specific to Sentinel-2 .SAFE files. The other notes/description could be shortened and added in markdown. The examples aren't necessary, we only need one example in the tutorial, not 3.

calebrob6:
I love this! I have a similar script that I use all the time for damage assessments https://github.com/microsoft/building-damage-assessment/blob/main/scripts/extract_data_mask_from_geotiff.py and would love to just have this functionality in torchgeo

torchgeo/datasets/utils.py
```diff
@@ -990,3 +992,149 @@ def convert_poly_coords(
...
+        src: An open raster dataset, either a `DatasetReader` (from `rasterio.open`)
+            or a `WarpedVRT` instance.
+
+    Returns:
+        A `Polygon` or `MultiPolygon` representing the footprint of valid data
+        in the raster's CRS.
+
+    .. versionadded:: 0.9
+    """
+    import shapely
```
calebrob6: Can this be at the top of the file?

calebrob6 (commented):
Does this work with epsg:4326 and epsg:3857 inputs (projected vs. not)?

How does this work when there are multiple valid data masks?

Is there an option to simplify the mask?

torchgeo/datasets/utils.py
adriantre: Definitely! Forgot to fix this when copying this out from the other PR

adriantre:
That's great @calebrob6!

> Does this work with epsg:4326 and epsg:3857 inputs (projected vs. not)?

Yes! If I remember correctly I also addet tests for this (or intended to)

> How does this work when there are multiple valid data masks?

Do you mean multi-band mask? It combines them as one mask.

> Is there an option to simplify the mask?

Yes, it currently simplifies based on the raster resolution to avoid jagged lines following the pixel grid. I might handle this part differently when I have time to revisit this.

What do you think about my docstring, where I suggest a caching approach in RasterDataset?

torchgeo/datasets/utils.py
```diff
@@ -990,3 +992,149 @@ def convert_poly_coords(
...
+        threshold: pixel values >= threshold are considered valid.
+            This is needed when/if the mask is based on alpha channel.
+            Defaults to 1 as we assume pixels outside the valid footprint will
+            also have 0 alpha.
+
+    Returns:
+        Binary mask of dtype uint8 with 255 = valid pixels, 0 = invalid.
+        If input is 3D, masks are combined (OR) before thresholding.
+
+    .. versionadded:: 0.9
```
calebrob6: needs bump

torchgeo/datasets/utils.py
```diff
@@ -990,3 +992,149 @@ def convert_poly_coords(
...
+def clean_binary_mask(
+    mask: np.typing.NDArray[np.number], threshold: int = 1
+) -> np.typing.NDArray[np.uint8]:
+    """Convert any rasterio mask to a clean binary mask (uint8 0 or 255).
+
+    Args:
+        mask: input mask array from `DatasetReader.dataset_mask()`
+              or `DatasetReader.read_masks()`.
+              Can be 2D or 3D with values between 0 and 255.
+        threshold: pixel values >= threshold are considered valid.
```
calebrob6, line 1064: Are alpha values that are different than 0 or 255 common? (just curious, not saying anything about this choice)

torchgeo/datasets/utils.py
```diff
@@ -990,3 +992,149 @@ def convert_poly_coords(
...
+    Args:
+        mask: binary mask where 255 representing valid pixels
+            and 0 representing nodata pixels
+        transform: affine transform for the raster
+        raster_resolution_x: pixel size in the x direction in CRS units
+
+    Returns:
+        A `Polygon` or `MultiPolygon` representing the valid data footprint of the raster
+
+    .. versionadded:: 0.9
```
calebrob6: Need bump

torchgeo/datasets/utils.py
calebrob6: Ping

calebrob6 (changes requested):
Just a few things, else looks good to me

torchgeo/datasets/utils.py
adriantre: I'm not sure how common it is, but I believe they often follow the dtype of the raster itself. The [sentinel driver in gdal](https://gdal.org/en/stable/drivers/raster/sentinel2.html#performance-issues-for-l1c-and-l2a) says:

> ALPHA=[YES​/​NO]: Defaults to NO. Whether to expose an alpha band. If set, an extra band is added after the Sentinel2 bands with an alpha channel. Its value are:
> - 0 on areas with no tiles, or when the tile data is set to the NODATA or SATURATED special values,
> - 4095 on areas with valid data.

adriantre:
Done! Bumped the versionadded to 0.10. Made the non-public methods private, and removed the versionadded for these.

I also rewrote the tests, trying to make them communicate intent of behaviour, and for coverage.

I compacted the docstring by removing the "example usage" to avoid bloating the docstring.

For future reference, this is one way of using it:
```python
def pre_compute_footprints(root:str = 'data/')
    rows = []
    for path in Path(root).rglob('*.tif'):
            rows.append({
                'filename': path.name,
                'geometry': get_valid_footprint_from_datasource(src)}
            )
    gpd.GeoDataFrame(rows, crs=src.crs).to_file('footprints.gpkg')

# Use in a RasterDataset subclass
class MyDataset(RasterDataset):
    def __init__(self, root, footprints='footprints.gpkg', **kwargs):
        gdf = gpd.read_file(footprints).set_index('filename')
        self._footprints = gdf['geometry'].to_dict()
        super().__init__(root, **kwargs)

    def _footprint_from_datasource(self, src):
        name = os.path.basename(src.name)
        return self._footprints.get(name) or super()._footprint_from_datasource(src)
```

torchgeo/datasets/utils.py
adriantre: Note that I currently do not expose threshold from the public method that calls this. Maybe I should? I doubt anyone would want to override it, but you never know

calebrob6:
I verified this works for me:

```
import geopandas as gpd
import rasterio
from rasterio.vrt import WarpedVRT
from torchgeo.datasets.utils import get_valid_footprint_from_datasource

with rasterio.open('B040001100075810_clip_warped.tif') as raster, \
        WarpedVRT(raster, src_nodata=0) as src:
    footprint = get_valid_footprint_from_datasource(src)
    crs = src.crs

gpd.GeoDataFrame(geometry=[footprint], crs=crs).to_crs(epsg=4326) \
   .to_file('B040001100075810_clip_warped_footprint.geojson', driver='GeoJSON')
```

Had to set the nodata explicitly with WarpedVRT as my rraster didn't have nodata set.

calebrob6 (approved):

calebrob6 (approved):
It might be nice to have a little wrapper function that lets me pass what value I want the nodata to be, but this is great for my purposes already

adriantre:
> It might be nice to have a little wrapper function that lets me pass what value I want the nodata to be, but this is great for my purposes already

Awesome of you to test it! 🙌🏻

And, I agree #3847 😎

calebrob6:
Would that help this implementation though? The actual function is taking a rasterio object if I understand correctly

adriantre:
Indeed! I keep forgetting what is merged and not.
When this too #2903 gets merged, you can do:

```python
from torchgeo.datasets.utils import get_valid_footprint_from_datasource

class MyDataset(RasterDataset):
    nodata_value = 0.0

    def footprint_from_datasource(
        self, dataset: DatasetReader | WarpedVRT
    ) -> MultiPolygon | Polygon:
        # Add cache to file here
        return get_valid_footprint_from_datasource(dataset)
```

Which overrides the default box from bounds

isaaccorley (approved):

PR #3665, opened 2026-05-06
docs: Skip case study notebooks from pytest collection

hkristen:
As agreed within the TSC, we want to enable more sophisticated and long running case studies within TorchGeo. A pragmatic way to do this, without overwhelming the CI runners is disabling tests for the case studies. 

Skips testing for the existing case studies. With this approach every new case study needs to be added to the `collect_ignore` list. The upside is, that the URLs to existing case studies stay the same.

The alternative is moving them into a subdirectory and excluding it from pytest, but that would've changed the URLs for those pages on the docs site and broken any existing links.

## AI Usage Disclosure

### DIFF
--- a/docs/tutorials/conftest.py
+++ b/docs/tutorials/conftest.py
@@ -1,0 +1,8 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+
+collect_ignore = [
+    'earth_surface_water.ipynb',
+    'earthquake_detection.ipynb',
+    'naip_road_segmentation.ipynb',
+]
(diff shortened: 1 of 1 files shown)


docs/tutorials/conftest.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+
+collect_ignore = [
+    'earth_surface_water.ipynb',
```
adamjstewart, line 6:
```suggestion
    'change_detection.ipynb',
    'earth_surface_water.ipynb',
```
You're missing your own tutorial which is now merged on main 😆

adamjstewart (approved):
Looks good, this is significantly better than what I was thinking.

PR #3666, opened 2026-05-06
feat: Add GeoTIFFWriter callback for chunked COG writing

hkristen:
This is the first of three PRs splitting up the tiled-inference-callabck #3214 (check out the archtitecture overview there for more context about how this will be used). 

Adds `GeoTIFFWriter`, a rasterio-based utility for writing Cloud-Optimized GeoTIFFs in chunks without loading the full scene into memory. It uses LZW compression, tiled layout, and optional overview building via GDAL if available. 

The writer is a context manager, meaning you open it, call `write_chunk()` for each spatial tile, then `finalize()` builds the overviews and closes the file. No torchgeo internal dependencies, fully standalone. 

PR2 (blending utilities) and PR3 (TiledInferenceCallback) build on top of this.

## AI Usage Disclosure

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -55,6 +55,7 @@ needs_sphinx = '5.3'
 nitpicky = True
 nitpick_ignore = [
     # Undocumented classes
+    ('py:class', 'affine.Affine'),
     ('py:class', 'kornia.augmentation._2d.intensity.base.IntensityAugmentationBase2D'),
     ('py:class', 'kornia.augmentation._3d.geometric.base.GeometricAugmentationBase3D'),
     ('py:class', 'kornia.augmentation.base._AugmentationBase'),
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -69,6 +69,7 @@ nitpick_ignore = [
     ('py:class', 'torchmetrics.Recall'),
     ('py:class', 'torchmetrics.detection.mean_ap.MeanAveragePrecision'),
     # Undocumented classes
+    ('py:class', 'affine.Affine'),
     ('py:class', 'kornia.augmentation._2d.intensity.base.IntensityAugmentationBase2D'),
     ('py:class', 'kornia.augmentation._3d.geometric.base.GeometricAugmentationBase3D'),
     ('py:class', 'kornia.augmentation.base._AugmentationBase'),
--- a/torchgeo/callbacks/__init__.py
+++ b/torchgeo/callbacks/__init__.py
@@ -2,7 +2,3 @@
 # Licensed under the MIT License.
 
 """TorchGeo callbacks."""
-
-from .writer import GeoTIFFWriter
-
-__all__ = ['GeoTIFFWriter']
--- a/torchgeo/callbacks/__init__.py
+++ b/torchgeo/callbacks/__init__.py
@@ -1,0 +1,8 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""TorchGeo callbacks."""
+
+from .writer import GeoTIFFWriter
+
+__all__ = ['GeoTIFFWriter']
--- a/torchgeo/callbacks/__init__.py
+++ b/torchgeo/callbacks/__init__.py
@@ -1,4 +1,4 @@
-# Copyright (c) Microsoft Corporation. All rights reserved.
+# Copyright (c) TorchGeo Contributors. All rights reserved.
 # Licensed under the MIT License.
 
 """TorchGeo callbacks."""
--- a/torchgeo/callbacks/writer.py
+++ b/torchgeo/callbacks/writer.py
@@ -10,6 +10,7 @@ from typing import Any
 
 import numpy as np
 import rasterio
+import rasterio.shutil
 from rasterio.transform import Affine
 from rasterio.windows import Window
 
@@ -68,6 +69,7 @@ class GeoTIFFWriter:
         self.dtype = dtype
         self.cog_config = cog_config or {}
 
+        self.tmp_path = self.output_path.with_suffix('.tmp.tif')
         self.dataset: Any = None
 
     def __enter__(self) -> GeoTIFFWriter:
@@ -86,9 +88,10 @@ class GeoTIFFWriter:
             'transform': self.transform,
             'tiled': True,
             'compress': self.cog_config.get('compress', 'lzw'),
+            'BIGTIFF': 'IF_SAFER',
         }
 
-        self.dataset = rasterio.open(self.output_path, 'w', **kwargs)
+        self.dataset = rasterio.open(self.tmp_path, 'w', **kwargs)
         return self
 
     def write_chunk(
@@ -120,29 +123,18 @@ class GeoTIFFWriter:
             self.dataset.close()
 
     def finalize(self) -> None:
-        """Build overview pyramids for Cloud-Optimized GeoTIFF.
+        """Translate the written GTiff into a Cloud-Optimized GeoTIFF.
 
-        Builds internal overviews at the specified levels using the configured
-        resampling method. Overviews enable efficient visualization at multiple
-        zoom levels without loading the full resolution image.
+        Streams the temporary GTiff through the GDAL COG driver, which builds
+        overviews so it validates as a COG, without loading the full resolution
+        image into memory.
         """
-        if self.cog_config.get('overviews'):
-            overview_levels = self.cog_config['overviews']
-            resampling = self.cog_config.get('overview_resampling', 'nearest')
-
-            from rasterio.enums import Resampling
-
-            resampling_map = {
-                'nearest': Resampling.nearest,
-                'bilinear': Resampling.bilinear,
-                'cubic': Resampling.cubic,
-                'average': Resampling.average,
-                'mode': Resampling.mode,
-            }
-
-            resampling_method = resampling_map.get(
-                resampling.lower(), Resampling.nearest
-            )
-
-            with rasterio.open(self.output_path, 'r+') as dst:
-                dst.build_overviews(overview_levels, resampling_method)
+        rasterio.shutil.copy(
+            self.tmp_path,
+            self.output_path,
+            driver='COG',
+            compress=self.cog_config.get('compress', 'lzw'),
+            overview_resampling=self.cog_config.get('overview_resampling', 'nearest'),
+            BIGTIFF='IF_SAFER',
+        )
+        self.tmp_path.unlink()
--- a/torchgeo/callbacks/writer.py
+++ b/torchgeo/callbacks/writer.py
@@ -164,5 +164,8 @@ class GeoTIFFWriter(contextlib.AbstractContextManager['GeoTIFFWriter']):
                 BIGTIFF='IF_SAFER',
                 **extra,
             )
+        except Exception:
+            self.output_path.unlink(missing_ok=True)
+            raise
         finally:
             self.tmp_path.unlink(missing_ok=True)
--- a/torchgeo/callbacks/writer.py
+++ b/torchgeo/callbacks/writer.py
@@ -42,6 +42,7 @@ class GeoTIFFWriter:
         crs: Any,
         transform: Affine,
         dtype: str = 'uint8',
+        nodata: float | None = None,
         cog_config: dict[str, Any] | None = None,
     ) -> None:
         """Initialize writer.
@@ -54,6 +55,7 @@ class GeoTIFFWriter:
             crs: Coordinate reference system.
             transform: Affine transform.
             dtype: Output data type.
+            nodata: Value to use for nodata pixels.
             cog_config: Optional COG configuration.
         """
         self.output_path = Path(output_path)
@@ -63,6 +65,7 @@ class GeoTIFFWriter:
         self.crs = crs
         self.transform = transform
         self.dtype = dtype
+        self.nodata = nodata
         self.cog_config = cog_config or {}
 
         self.tmp_path = self.output_path.with_suffix('.tmp.tif')
@@ -82,6 +85,7 @@ class GeoTIFFWriter:
             'dtype': self.dtype,
             'crs': self.crs,
             'transform': self.transform,
+            'nodata': self.nodata,
             'tiled': True,
             'compress': self.cog_config.get('compress', 'lzw'),
             'BIGTIFF': 'IF_SAFER',
--- a/torchgeo/callbacks/writer.py
+++ b/torchgeo/callbacks/writer.py
@@ -50,6 +50,7 @@ class GeoTIFFWriter(contextlib.AbstractContextManager['GeoTIFFWriter']):
         transform: Affine,
         dtype: str = 'uint8',
         nodata: float | None = None,
+        overwrite: bool = False,
         **kwargs: Any,
     ) -> None:
         """Initialize writer.
@@ -63,6 +64,7 @@ class GeoTIFFWriter(contextlib.AbstractContextManager['GeoTIFFWriter']):
             transform: Affine transform.
             dtype: Output data type.
             nodata: Value to use for nodata pixels.
+            overwrite: If True, overwrite an existing file at *output_path*.
             **kwargs: Additional keyword arguments passed to the GDAL COG
                 driver (e.g. ``compress``, ``overview_resampling``).
         """
@@ -74,6 +76,7 @@ class GeoTIFFWriter(contextlib.AbstractContextManager['GeoTIFFWriter']):
         self.transform = transform
         self.dtype = dtype
         self.nodata = nodata
+        self.overwrite = overwrite
         self.kwargs = kwargs
 
         self.tmp_path = self.output_path.with_suffix('.tmp.tif')
@@ -85,6 +88,10 @@ class GeoTIFFWriter(contextlib.AbstractContextManager['GeoTIFFWriter']):
         Returns:
             GeoTIFFWriter instance.
         """
+        if self.output_path.exists() and not self.overwrite:
+            msg = f'Output path already exists: {self.output_path}'
+            raise FileExistsError(msg)
+
         self.dataset = rasterio.open(
             self.tmp_path,
             'w',
--- a/torchgeo/callbacks/writer.py
+++ b/torchgeo/callbacks/writer.py
@@ -1,0 +1,148 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+
+"""GeoTIFF writer with Cloud-Optimized GeoTIFF support."""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+
+import numpy as np
+import rasterio
+from rasterio.transform import Affine
+from rasterio.windows import Window
+
+
+class GeoTIFFWriter:
+    """GeoTIFF writer with chunked writing and COG support.
+
+    Example::
+
+        writer = GeoTIFFWriter(
+            output_path='output.tif',
+            width=1024,
+            height=1024,
+            num_bands=1,
+            crs='EPSG:32631',
+            transform=Affine(...),
+        )
+
+        with writer:
+            for chunk_y, chunk_x, chunk_data in chunks:
+                writer.write_chunk(chunk_data, chunk_y, chunk_x)
+
+        writer.finalize()
+
+    """
+
+    def __init__(
+        self,
+        output_path: str | Path,
+        width: int,
+        height: int,
+        num_bands: int,
+        crs: Any,
+        transform: Affine,
+        dtype: str = 'uint8',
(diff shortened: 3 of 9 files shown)


calebrob6 (commented):
This is really cool! Some thoughts:
- Can you use https://github.com/OSGeo/gdal/blob/master/swig/python/gdal-utils/osgeo_utils/samples/validate_cloud_optimized_geotiff.py or `rio cog validate` (I _think_ that's the command, see https://github.com/cogeotiff/rio-cogeo) to check to make sure the output of this is indeed a COG.
- What are some other places that we could put this other than a new callbacks module?
- How does this compare speed-wise to writing the whole COG at once? I'm trying to figure out at what size input you'd want to change over from writing the whole COG vs. writing piece-wise.

adamjstewart:
> Can you use https://github.com/OSGeo/gdal/blob/master/swig/python/gdal-utils/osgeo_utils/samples/validate_cloud_optimized_geotiff.py or rio cog validate (I think that's the command, see https://github.com/cogeotiff/rio-cogeo) to check to make sure the output of this is indeed a COG.

Or `gdal driver cog validate` if using a new-enough version.

> What are some other places that we could put this other than a new callbacks module?

We could have a new hidden utils module, but currently this is only scheduled to be used in a new callback.

hkristen:
I will have a look which of the proposed ways to validate the COGs would work best and integrate nicely. 

The main reason to write smaller COGs is to be independent from the available RAM. It should always be faster to write a single large COG instead of many small ones as you only create, open and close the file once. If you want I can write a small benchmark script to look into this though @calebrob6 

Idc too much where we put this tbh. As Adam said, it will only be used in the callback as of now, hence my reasoning to put it there.

adamjstewart:
> It should always be faster to write a single large COG instead of many small ones

I have a feeling that proper parallelization could change this, but also that this will end up being very nuanced. GDAL does a good job of parallelization entirely (assuming we are setting the right env vars/flags) so it may not actually be a huge difference.

calebrob6:
You don't have to build n the validation -- was just curious if you run this on your side, does the output validate?

Also was just curious about the difference between writing full file vs. piece-wise, not a blocker.

I do think this should go into a utils type location until we have more than one callback that necessitates its own top-level organization.

adamjstewart:
In addition to the stitching callback, there are also plans to add a logging callback for tensorboard and friends.

hkristen:
> You don't have to build n the validation -- was just curious if you run this on your side, does the output validate?
> 

@calebrob6 Good instinct on your side, the COG didn't validate: I was appending the overviews after the image creation, which isn't ok by the COG spec :P

I changed the logic to write a plain GeoTIFF tmpfile to disk and then do a COG translation afterwards (incl. building the overviews). Now it validates with validate_cloud_optimized_geotiff.py script you pointed me to. It is still memory efficient, but of course it temporarily needs double the disk space. Since users usually have enough disk but not enough memory, that seems like a good tradeoff.


> Also was just curious about the difference between writing full file vs. piece-wise, not a blocker.
> 

I tried this out in a small benchmark (based on the changes above): writing the whole file in one shot is actually ~2x faster on an 8k x 8k single-band image. As the chunked writing does an extra read+write pass for the COG translation. But since we want to write rasters that are potentially bigger than memory, chunked writing is the only option that works there, and disk is cheaper than RAM.

> I do think this should go into a utils type location until we have more than one callback that necessitates its own top-level organization.

Any decision between @calebrob6 and @adamjstewart and where we should put this now?

adamjstewart:
I'm fine with `torchgeo.callbacks.utils` since we know more callbacks are coming. Other option is to quickly add the other tensorboard plotting callback we wanted which definitely justifies a new namespace.

calebrob6:
Cool thanks for this @hkristen, interesting work!

I'm fine with what Adam said.

torchgeo/callbacks/__init__.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
ashnair1:
```diff
- # Copyright (c) Microsoft Corporation. All rights reserved.
+ # Copyright (c) TorchGeo Contributors. All rights reserved.
```

ashnair1 (changes requested):
One minor copyright issue. Otherwise LGTM

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,140 @@
...
+            'height': self.height,
+            'width': self.width,
+            'count': self.num_bands,
+            'dtype': self.dtype,
+            'crs': self.crs,
+            'transform': self.transform,
+            'tiled': True,
+            'compress': self.cog_config.get('compress', 'lzw'),
+            'BIGTIFF': 'IF_SAFER',
+        }
```
adriantre: Consider specifying a nodata value, such that future blending will ignore these pixels and not interpret e.g. value `0` as an actual pixel value. Nodata will happen near the edges due to reprojection or due to ROI-sampling.

```python
'nodata': self.nodata
```

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,140 @@
...
+            data: Chunk data of shape (H, W) for single band.
+            y_offset: Row offset in output.
+            x_offset: Column offset in output.
+        """
+        if self.dataset is None:
+            raise RuntimeError('Writer not opened. Use with statement.')
+
+        h, w = data.shape
+        window = Window(x_offset, y_offset, w, h)  # ty: ignore[too-many-positional-arguments]
+        self.dataset.write(data, 1, window=window)
```
adriantre:
```suggestion
        if data.ndim == 2:
            data = data[np.newaxis]
        _, h, w = data.shape
        window = Window(x_offset, y_offset, w, h)  # ty: ignore[too-many-positional-arguments]
        self.dataset.write(data, window=window)
```

Allows writing multi-band (by first adding a 1-channel dim to 2D data.

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,140 @@
...
+            'tiled': True,
+            'compress': self.cog_config.get('compress', 'lzw'),
+            'BIGTIFF': 'IF_SAFER',
+        }
+
+        self.dataset = rasterio.open(self.tmp_path, 'w', **kwargs)
+        return self
+
+    def write_chunk(
+        self, data: np.typing.NDArray[np.uint8], y_offset: int, x_offset: int
```
adriantre:
```suggestion
        self, data: np.typing.NDArray[Any], y_offset: int, x_offset: int
```
Predictions are often float too

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,140 @@
...
+    ) -> None:
+        """Write a chunk to the output GeoTIFF.
+
+        Args:
+            data: Chunk data of shape (H, W) for single band.
```
adriantre:
```suggestion
            data: Chunk data of shape (C, H, W) for single band.
```

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,140 @@
...
+
+    """
+
+    def __init__(
+        self,
+        output_path: str | Path,
+        width: int,
+        height: int,
+        num_bands: int,
+        crs: Any,
```
adriantre:
```suggestion
        crs: CRS | str,
```
`from pyproj import CRS`. rasterio accepts both

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,140 @@
...
+            crs='EPSG:32631',
+            transform=Affine(...),
+        )
+
+        with writer:
+            for chunk_y, chunk_x, chunk_data in chunks:
+                writer.write_chunk(chunk_data, chunk_y, chunk_x)
+
+        writer.finalize()
+
```
adriantre:
```suggestion
    .. versionadded:: 0.10
```
:)

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        rasterio.shutil.copy(
+            self.tmp_path,
+            self.output_path,
+            driver='COG',
+            compress=self.cog_config.get('compress', 'lzw'),
+            overview_resampling=self.cog_config.get('overview_resampling', 'nearest'),
+            BIGTIFF='IF_SAFER',
+        )
+        self.tmp_path.unlink()
```
adriantre: When finalize is called outside the context-manager any exception inside the `with`-clause will cause finalize() to not run. This will leave an .tmp.tif and no COG.

Suggesting doing this inside `__exit__`, which both handles errors and the user does not have to call it manually

```suggestion
        if self.dataset is not None:
            self.dataset.close()
            self.dataset = None
        if exc_type is not None:
            # writing failed — don't emit a half-written COG
            self.tmp_path.unlink(missing_ok=True)
            return
        rasterio.shutil.copy(
            self.tmp_path, self.output_path, driver='COG',
            compress=self.cog_config.get('compress', 'lzw'),
            overview_resampling=self.cog_config.get('overview_resampling', 'nearest'),
            BIGTIFF='IF_SAFER',
        )   
        self.tmp_path.unlink(missing_ok=True)
```

torchgeo/callbacks/writer.py
calebrob6: @adriantre, (C, H, W) is incorrect as this specifies "for single band"

torchgeo/callbacks/writer.py
calebrob6: Oh never mind my above comment, you want to change this to write multi-band. I agree with this :)

calebrob6 (changes requested):
I agree with @adriantre's review here

torchgeo/callbacks/writer.py
adriantre: Yeah, then the restriction moves to the combiner instead :)

torchgeo/callbacks/writer.py
hkristen: Makes sense, added a nodata parameter (defaults to None).

torchgeo/callbacks/writer.py
hkristen: Good idea, we can support multi-band in the writer. The single-band and uint8 limitations were just carried over from the initial PR to keep scope small.

torchgeo/callbacks/writer.py
hkristen: Good catch!! I checked the related branch #3667 and it shows exactly this pattern (finalize called after the with block) -> will implement it your way :)

hkristen: FYI I also wrapped the COG translation in try/finally so a failed translation can't leave the tmp file behind

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,156 @@
...
+        ) as writer:
+            for chunk_y, chunk_x, chunk_data in chunks:
+                writer.write_chunk(chunk_data, chunk_y, chunk_x)
+
+    .. versionadded:: 0.10
+    """
+
+    def __init__(
+        self,
+        output_path: str | Path,
```
adamjstewart: Should except any os.PathLike, use `..datasets.utils.Path` instead.

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,156 @@
...
+        Returns:
+            GeoTIFFWriter instance.
+        """
+        kwargs = {
+            'driver': 'GTiff',
+            'height': self.height,
+            'width': self.width,
+            'count': self.num_bands,
+            'dtype': self.dtype,
+            'crs': self.crs,
```
adamjstewart: rasterio CRS != pyproj CRS, see https://pyproj4.github.io/pyproj/stable/crs_compatibility.html#rasterio, the new rasterio-types type hints will complain about this

torchgeo/callbacks/__init__.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+
+"""TorchGeo callbacks."""
+
+from .writer import GeoTIFFWriter
+
+__all__ = ['GeoTIFFWriter']
```
adamjstewart: Is this something we want to expose to users or a utility we plan to use internally? Personally I think the less we document, the less we have to worry about breaking backwards compatibility in the future.

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,156 @@
...
+        self,
+        output_path: str | Path,
+        width: int,
+        height: int,
+        num_bands: int,
+        crs: CRS | str,
+        transform: Affine,
+        dtype: str = 'uint8',
+        nodata: float | None = None,
+        cog_config: dict[str, Any] | None = None,
```
adamjstewart: Can we use `**kwargs` instead?

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,156 @@
...
+
+import numpy as np
+import rasterio
+import rasterio.shutil
+from pyproj import CRS
+from rasterio.transform import Affine
+from rasterio.windows import Window
+
+
+class GeoTIFFWriter:
```
adamjstewart: Should this be a subclass of one of https://docs.python.org/3/library/contextlib.html?

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,156 @@
...
+        self.height = height
+        self.num_bands = num_bands
+        self.crs = crs
+        self.transform = transform
+        self.dtype = dtype
+        self.nodata = nodata
+        self.cog_config = cog_config or {}
+
+        self.tmp_path = self.output_path.with_suffix('.tmp.tif')
+        self.dataset: Any = None
```
adamjstewart: Can we get a little more specific about the type?

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def __enter__(self) -> GeoTIFFWriter:
+        """Open GeoTIFF for writing.
+
+        Returns:
+            GeoTIFFWriter instance.
+        """
+        kwargs = {
```
adamjstewart: Can you add a comment pointing to the suggested COG settings we are using so it's clear where they came from? For example, `torchgeo/main.py` points to https://github.com/pangeo-data/cog-best-practices

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,156 @@
...
+            y_offset: Row offset in output.
+            x_offset: Column offset in output.
+        """
+        if self.dataset is None:
+            raise RuntimeError('Writer not opened. Use with statement.')
+
+        if data.ndim == 2:
+            data = data[np.newaxis]
+        _, h, w = data.shape
+        window = Window(x_offset, y_offset, w, h)  # ty: ignore[too-many-positional-arguments]
```
adamjstewart: Might be fixed by rasterio-types

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,156 @@
...
+            'tiled': True,
+            'compress': self.cog_config.get('compress', 'lzw'),
+            'BIGTIFF': 'IF_SAFER',
+        }
+
+        self.dataset = rasterio.open(self.tmp_path, 'w', **kwargs)
+        return self
+
+    def write_chunk(
+        self, data: np.typing.NDArray[Any], y_offset: int, x_offset: int
```
adamjstewart, line 113: When/where does the predicted Tensor get converted to a numpy array?

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,156 @@
...
+        """Write a chunk to the output GeoTIFF.
+
+        Args:
+            data: Chunk data of shape (H, W) for single band or (C, H, W) for
+                multi-band.
+            y_offset: Row offset in output.
+            x_offset: Column offset in output.
+        """
+        if self.dataset is None:
+            raise RuntimeError('Writer not opened. Use with statement.')
```
adamjstewart: If we make this a private helper context manager, then we don't need so much defensive coding, as users can't use it incorrectly.

adamjstewart (dismissed):
See https://github.com/torchgeo/torchgeo/pull/3861 for a few more files that need to be modified when adding a new subpackage.

Need to refer back to the original PR so I remember why we designed things this way, but this is definitely easier to review one function at a time.

torchgeo/callbacks/__init__.py
hkristen: I second you that this is utility is used only internally.

torchgeo/callbacks/writer.py
hkristen: Sure!

torchgeo/callbacks/writer.py
hkristen: Sure :)

torchgeo/callbacks/writer.py
hkristen: Honestly those are settings I arrived at, after trial and error, couldn't find an official source for best practice COG-writing?

torchgeo/callbacks/writer.py
hkristen: Agreed and removed.

torchgeo/callbacks/writer.py
hkristen: The conversion happens in #3667

torchgeo/callbacks/writer.py
calebrob6: These settings make sense to me. They seem to be more/less default for COGs based on https://gdal.org/en/stable/drivers/raster/cog.html#creation-options

calebrob6 (dismissed):

calebrob6:
I've confirmed this works as expected in the context manager format and writes valid COGs quickly. This is a good addition -- let's merge!

calebrob6:
@adamjstewart you have requested changes here that I think have been addressed

adriantre (approved):
Could add a deliberate behaviour for when the output path already exists, I don't think that is covered, other than that it looks good to me!

hkristen:
> Could add a deliberate behaviour for when the output path already exists, I don't think that is covered, other than that it looks good to me!

Good idea @adriantre -> Now raises `FileExistsError `if the output file already exists, with an `overwrite=True` flag to opt in to replacing it.

@calebrob6 sorry, I accidentally dismissed your review with a rebase+force push

calebrob6:
@adamjstewart could you review this please or handoff the reviewer batton

calebrob6:
@hkristen the copilot review above is good -- can you take a look at the comments and we can get this merged

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,178 @@
...
+                self.tmp_path,
+                self.output_path,
+                driver='COG',
+                compress=self.kwargs.get('compress', 'lzw'),
+                overview_resampling=self.kwargs.get('overview_resampling', 'nearest'),
+                BIGTIFF='IF_SAFER',
+                **extra,
+            )
+        except Exception:
+            self.output_path.unlink(missing_ok=True)
```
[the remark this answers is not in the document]
hkristen: Good catch, the COG copy now goes to a staging file first and only os.replaces onto the output on success. If the translation fails the original is untouched.

hkristen:
@calebrob6 I factored in all the remaining review comments. 

Good to go from my side 🚀

docs/conf.py
```diff
@@ -86,6 +87,9 @@
     ('py:class', 'torchvision.models.swin_transformer.SwinTransformer'),
     # Internal type aliases we don't yet want to expose
     ('py:class', 'torchgeo.datasets.openstreetmap.OSMClassConfig'),
+    # No intersphinx inventory for these classes
+    ('py:class', 'tokenizers.models.BPE'),
```
calebrob6: Why did these change?

torchgeo/callbacks/writer.py
```diff
@@ -0,0 +1,184 @@
...
+            width=1024,
+            height=1024,
+            num_bands=1,
+            crs='EPSG:32631',
+            transform=Affine(...),
+        ) as writer:
+            for chunk_y, chunk_x, chunk_data in chunks:
+                writer.write_chunk(chunk_data, chunk_y, chunk_x)
+
+    .. versionadded:: 0.10
```
calebrob6: Needs to be updated to 0.11

calebrob6 (changes requested):
Sorry, noticed a few more details!

Also, can you go through https://github.com/torchgeo/torchgeo/pull/3861/changes and make sure you have everything that PR added to create the profilers. At least one change I see there that is missing here is a CODEOWNERS update.

docs/conf.py
calebrob6: oh I see -- this is for a new change that broke CI. You can remove this from this PR

hkristen: ok, i got rid of this commit again

PR #3667, opened 2026-05-06
feat: weighted blending utilities for patch stitching

hkristen:
Second of three PRs splitting up #3214 (check out the archtitecture overview there for more context about how this will be used).

Depends on PR1 (GeoTIFFWriter #3666) -> set base branch to `feat/callbacks-geo-tiff-writer` for now. I will rebase onto main once #3666 merges.

Adds the blending and stitching layer used to reconstruct a full scene from overlapping inference patches. The core function is `weighted_merge()`, which reads patches from disk, applies per-patch cosine or linear blend masks (Hann window approach from https://doi.org/10.1371/journal.pone.0229839), and writes the merged result as a COG via `GeoTIFFWriter`. Patches are processed in chunks with a spatial grid index to keep memory usage flat regardless of scene size.

A few things worth noting: 
- patches at scene boundaries get their edge crop suppressed to avoid black borders (`_get_edge_deltas`), and the scene extent can either be inferred from patch transforms or pinned to the original dataset bounds. 
- The regression tests for both of those edge cases are in `TestBlackBorder` and `TestDatasetBoundsMode`.
- Also introduces PatchMetadata, the TypedDict that carries per-patch geo bounds and transform info between the callback and the merge step.

**ToDos for further versions / not included to reduce scope**
- [ ] BUG: When using delta=0, blending is fully disabled
- [ ] Widen output dtype to `uint16` when `num_classes > 255` to avoid label truncation

## AI Usage Disclosure


torchgeo/callbacks/blending.py
adriantre, line 1: Here you go! https://github.com/hkristen/torchgeo/pull/1

torchgeo/callbacks/blending.py
```diff
@@ -0,0 +1,531 @@
...
+
+    top = 0 if abs(geo_bbox[3] - maxy) < tolerance else delta
+    bottom = 0 if abs(geo_bbox[1] - miny) < tolerance else delta
+    left = 0 if abs(geo_bbox[0] - minx) < tolerance else delta
+    right = 0 if abs(geo_bbox[2] - maxx) < tolerance else delta
+
+    return top, bottom, left, right
+
+
+def _reconstruct_scene_from_patches(
```
calebrob6, line 106: This won't work for "south-up" rasters like the AlphaEarth ones on source.coop. It always anchors the output at the maximum y-coordinate and assigns rows downward. This assumes a negative y-resolution.

torchgeo/callbacks/blending.py
hkristen: Thanks @adriantre - I included your suggested refactor in this branch :)

torchgeo/callbacks/blending.py
hkristen: Good catch @calebrob6! I added support for "south-up" rasters -> it auto-detects the y_res sign and handles origin/row placement accordingly.

PR #412, opened 2022-02-17
Add STACAPI dataset

nilsleh:
This PR closes #403 by adding a STACAPI dataset. I will try to explain my thinking for this proposal:

Here are some explanations for choices I made:
- constructor takes the usual root, crs, res, and transforms argument like other raster datasets
- additionally, there is a bands, api_endpoint and **query_parameters argument. I thought the bands argument should be separate because other raster datasets also specifically have this argument

When the __getitem__ method receives a bbox query:
- a single item that is part of the hit objects will be merged together to a single raster and computed resulting in a 4D xarray of (time x channel x height x width)
- since it can happen that multiple hits are returned for a particular bbox query, these have to be merged. Usually, this happens with rasterio because they are datareader objects or filepaths and not xarray resulting from stackstac. I only found the rioxarray library, which does merging for xarrays so this library is currently also required
- however, since rioxarray.merge only supports 2D or 3D tensors and potentially stac returns many samples over a time period, I loop over the time dimension, merge with rioxarray per time step and afterwords stack back together into one final tensor which will be returned as an image in the sample
- squeeze out the time dimension if it is only a single time instance

Questions / Left to do:
- several new libraries needed, maybe some can be avoided like rioxarray?
- computing the aoi can potentially take fairly long albeit it is much faster than I expected
- handling the resolution self.res parameter with each separate band having a different resolution
- deciding whether an image will be under image or mask key because StacAPI has both images and labels, so maybe another argument for the constructor?
- given the last point, need to implement a plot method that can handle both scenarios of the dataset containing images or labels
testing with sampler and dataloader

Issue #403, opened 2022-02-16
STAC API dataset


calebrob6:
SpatioTemporal Asset Catalogs ([STACs](https://stacspec.org/)) are a way to organize geospatial datasets. STAC APIs let users query huge STAC Catalogs by date, time, and other metadata. 

For example, the Microsoft [Planetary Computer](https://planetarycomputer.microsoft.com/) runs a STAC API that lets users search over catalogs containing _all_ of Sentinel 2 imagery, _all_ Landsat 8, etc. The following code uses the [pystac_client](https://pystac-client.readthedocs.io/en/latest/quickstart.html) library to query the Planetary Computer STAC API and returns metadata, and links to GeoTIFFs, for relevant Sentinel 2 scenes:
```
from pystac_client import Client

area_of_interest = {
    "type": "Polygon",
    "coordinates": [
        [
            [-148.56536865234375, 60.80072385643073],
            [-147.44338989257812, 60.80072385643073],
            [-147.44338989257812, 61.18363894915102],
            [-148.56536865234375, 61.18363894915102],
            [-148.56536865234375, 60.80072385643073],
        ]
    ],
}
time_of_interest = "2019-06-01/2019-08-01"

catalog = Client.open("https://planetarycomputer.microsoft.com/api/stac/v1")

search = catalog.search(
    collections=["sentinel-2-l2a"],
    intersects=area_of_interest,
    datetime=time_of_interest,
    query={"eo:cloud_cover": {"lt": 10}},
)

items = list(search.get_items())
print(f"Returned {len(items)} Items")
```

We'd like to build a `STACAPIDataset` object that essentially wraps `catalog.search(...)`, creates a RasterDataset from the returned items, and otherwise behaves as a normal PyTorch dataset (signing assets as needed, etc.).

As a detailed note, it may be a good idea to cache accessed data in a local directory.

nilsleh:
I would be really interested in taking on this task!

calebrob6:
All yours :) (I had you in mind writing this actually, it is a bit more interesting than the other dataset stuff!) -- feel free to message me if you want to discuss details

@adamjstewart -- this would involve taking on some dependencies (pystac_client, planetary-computer, maybe stackstac)

adamjstewart:
We can make those deps optional if we need to.

metazool:
Nice potential feature, is there still intention to work on it?

adamjstewart:
[...]

PR #3670, opened 2026-05-07
STAC Geoparquet TorchGeo dataset

print-sid8:
This PR adds a `STACDataset` for STAC GeoParquet-backed raster workflows in TorchGeo.

This work follows prior discussions and attempts around STAC support in TorchGeo, including:

- STAC meeting notes: [G-doc](https://docs.google.com/document/d/1Y-R4BGxNlVp3NwOzjlklMS_yezdQXnZXzog_a-hNsZ0/edit?tab=t.0#heading=h.rjnqg1n0389o)
- STAC discussion PR #3523
- OpenAerialMap PR #3309
- older STAC PRs #403 and #412

Some discussion may continue here and in Slack: [thread](https://torchgeo.slack.com/archives/C0A0CE1J4D6/p1776241448078329).

The design targets STAC GeoParquet workflows where live STAC API calls during training or inference are undesirable because they can be slow, rate limited, and hard to reproduce. Instead, `STACDataset` uses a STAC GeoParquet item table as the dataset index. It builds a TorchGeo spatiotemporal index from matching STAC items, then reads raster pixels lazily through TorchGeo's existing `RasterDataset` path.

## Features

- Provider-neutral `STACDataset(index_path=..., asset_keys=...)`
- Local, HTTP(S), and fsspec-backed STAC GeoParquet reads
- bbox, time, and PyArrow-style filters for item selection
- strict STAC GeoParquet contract validation
- relative asset href resolution
- lazy raster asset signing via `sign_href`
- deterministic offline tests with committed tiny GeoTIFF fixtures

## User API

```python
import os
from torchgeo.datasets import STACDataset

os.environ['AWS_NO_SIGN_REQUEST'] = 'YES'   # anonymous reads of the public COGs

# (1) Meta Canopy Height v2 — the dataset from #3757, read via STAC GeoParquet
chm = STACDataset(
    index_path='s3://us-west-2.opendata.source.coop/tge-labs/meta-chm-v2/stac/items.parquet',
    asset_keys=('chm',),
    intersects_bbox=(-122.5, 47.4, -122.0, 47.8),                 # Seattle area
    time_range=('2019-01-01', '2020-12-31'),
    storage_options={'anon': True},
)
# len(chm.index) == 6 ; chm.crs → EPSG:3857 ; chm.res ≈ (1.19, 1.19)
# sample['image'].shape == torch.Size([1, 64, 64])

# (2) AlphaEarth Foundation embeddings — 64-band COGs  (registry.opendata.aws/aef-source)
aef = STACDataset(
    index_path='s3://us-west-2.opendata.source.coop/tge-labs/aef/v1/annual/aef_index_stac_geoparquet.parquet',
    asset_keys=('data',),
    intersects_bbox=(178.5, -17.8, 179.3, -17.0),                 # a tile over Fiji
    time_range=('2025-01-01', '2025-12-31'),
    storage_options={'anon': True},
)
# len(aef.index) == 7 ; aef.crs → EPSG:32760 ; aef.res == (10.0, 10.0)
# sample['image'].shape == torch.Size([64, 64, 64])
```

## Notes

- `STACDataset` can index a full GeoParquet table or a filtered selection.
- `intersects_bbox` selects STAC items whose geometry intersects the bbox. It does not clip raster pixels.
- Pixel extents are controlled by TorchGeo samplers or `__getitem__` slices.
- `max_index_items` is a memory-safety guard for the in-memory TorchGeo index; set `max_index_items=None` to intentionally index all matching items.
- Cloud URI support uses fsspec. HTTP(S) support is included through `fsspec[http]`; object-store URIs may require user-installed backends such as `s3fs`, `gcsfs`, or `adlfs`.
- MPC collection discovery/signing, CQL2 parsing, STAC API querying, and STAC API-to-GeoParquet materialization are intentionally out of scope for this MVP.

## Dependencies

Adds one optional dataset dependency:

- `fsspec[http]>=2024.6`

## Tests

- `tests/datasets/test_stac.py`
- `tests/data/stac/data.py`

## AI Usage Disclosure


print-sid8:
I force-pushed this PR to reduce the scope and make the review path clearer.

The earlier version tried to include too much in one change: provider-specific MPC helpers, extra optional dependencies, custom PyArrow/STAC I/O utilities, and more aggressive performance-oriented filtering code. That made the PR harder to review and much bigger as well.

~2700 lines have reduced to ~1000 lines now.

print-sid8:
@rosielickorish Would like your review here if you have some time. @adamjstewart had mentioned you in Torchgeo Slack. 

Hopefully this can help terratorch and terrakit as well coz torchgeo will be the foundation for them ofcourse in many ways.

The focus here in this PR, is not to support live STAC API endpoints rather only STAC Geoparquets to avoid API calls in live training and inference workflows. Many previous attempts have been on getting STAC API support. 

Also, I have avoided all dependencies like Xarray and more.

torchgeo/datasets/stac.py
```diff
@@ -0,0 +1,392 @@
...
+            raise ValueError('STAC GeoParquet is missing geometry.')
+        if 'assets' not in cols:
+            raise ValueError('STAC GeoParquet is missing required assets column.')
+        if 'datetime' not in cols and not {'start_datetime', 'end_datetime'} <= cols:
+            raise ValueError(
+                'STAC GeoParquet must include datetime or start_datetime/end_datetime.'
+            )
+
+        # normalize datetime columns to a single start/end pair, filling missing values
+        has_interval = {'start_datetime', 'end_datetime'} <= cols
```
lakshay-sharma-instacart: Nit: for readability, I might suggest making this `{'start_datetime', 'end_datetime'}.issubset(cols)`

torchgeo/datasets/stac.py
```diff
@@ -0,0 +1,392 @@
...
+            )
+            if dt_col is not None:
+                start = start.fillna(dt_col)
+                end = end.fillna(dt_col)
+        else:
+            start = dt_col
+            end = dt_col
+
+        start = cast(pd.Series, start)
+        end = cast(pd.Series, end)
```
lakshay-sharma-instacart: **Nit: the `cast()` calls hide a type-checker lie.**

  In the `else` branch (`has_interval=False`), `start = end = dt_col`, but `dt_col`
  is typed `Series | None`. The `cast(pd.Series, ...)` at L198–199 tells the
  checker it's a `Series` without proving it. Runtime is safe only because the
  column-validation block at L171–174 guarantees that whenever `has_interval` is
  `False`, `'datetime'` must be present (and therefore `dt_col` is a `Series`) —
  but that invariant lives too far away for the type checker to follow.

  Suggest replacing the casts with an `assert` inside the branch that actually
  needs it:

  ```python
  if has_interval:
      start = pd.to_datetime(df['start_datetime'], utc=True, format='ISO8601', errors='coerce')
      end   = pd.to_datetime(df['end_datetime'],   utc=True, format='ISO8601', errors='coerce')
      if dt_col is not None:
          start = start.fillna(dt_col)
          end = end.fillna(dt_col)
  else:
      # Guaranteed by the column-validation block above: if the interval columns
      # are absent then 'datetime' must be present, so dt_col is a Series.
      assert dt_col is not None
      start = end = dt_col
  ```
  Both cast() calls can then be deleted — the has_interval=True branch is
  already Series-typed via pd.to_datetime / fillna.

torchgeo/datasets/stac.py
```diff
@@ -0,0 +1,392 @@
...
+            )
+            if t_end < t_start:
+                raise ValueError(
+                    'time_range stop must be greater than or equal to start.'
+                )
+
+            time_mask = (start <= t_end) & (end >= t_start)
+            df, start, end = df[time_mask], start[time_mask], end[time_mask]
+
+        if df.empty:
```
lakshay-sharma-instacart, line 232: When the code hits if df.empty at line 226, the DataFrame could be empty because:

  - intersects_bbox matched nothing, but if time_range is also set, the error still blames time_range.
  - The pushdown filters matched nothing; same misattribution.
  - The bbox fallback matched nothing; same.
  - Only the last case (time_mask filtered everything out) actually points at time_range.

  For simplicity and to not provide a misleading message, can just do

```
    if df.empty:
      raise ValueError('No STAC items matched the requested filters.')
```

torchgeo/datasets/stac.py
```diff
@@ -0,0 +1,392 @@
...
+                f'STAC filters matched more than {len(df)} items, exceeding max_index_items={max_index_items}. Narrow the filters or increase max_index_items.'
+            )
+
+        # extract and validate asset hrefs for all requested keys
+        hrefs = {}
+        valid_rows = pd.Series(True, index=df.index)
+        for key in self.asset_keys:
+            col = df['assets'].str.get(key).str.get('href')
+            hrefs[key] = col
+            valid_rows &= col.notna() & (col != '')
```
lakshay-sharma-instacart: Can increase null/whitespace check coverage by changing this line to

  ```python
  valid_rows &= col.fillna('').str.strip() != ''
```
 This covers the cases above plus `' '`

File: torchgeo/datasets/stac.py (hunk 1 above)
Comment (print-sid8) [reply torchgeo/datasets/stac.py]: Done. yes its more readable

File: torchgeo/datasets/stac.py (hunk 2 above)
Comment (print-sid8) [reply torchgeo/datasets/stac.py]: good catch. Replaced both with an assert dt_col is not None in the else branch and dropped the casts. ty is happy with the narrowing. Left a short comment too.

File: torchgeo/datasets/stac.py (hunk 3 above)
Comment (print-sid8) [reply torchgeo/datasets/stac.py:232]: yep, done.

File: torchgeo/datasets/stac.py (hunk 4 above)
Comment (print-sid8) [reply torchgeo/datasets/stac.py]: Done

**Comment by rosielickorish** [comment]
Thanks @print-sid8 for making me aware of this feature. From what I can see, a possible future update to Terrakit could be to include an export utility function that reads a .tacozip and writes a STAC GeoParquet file — extracting the embedded rasters and pointing asset hrefs at them — so the result could be consumed directly by STACDataset in torchgeo/Terratorch. The STAC metadata already embedded by Terrakit in each sample (CRS, geotransform, spatial bounds, timestamps) is exactly what the GeoParquet table would need, so the main work would be the extraction and table-writing step. Integration with TerraTorch looks simple and straightforward. Nice work!

**Comment by print-sid8** [comment]
@rosielickorish Thank you!

Just one note, im reminded of this after seeing your comment. Might be worth discussing too @lakshay-sharma-instacart @adamjstewart @isaaccorley 

On CRS/resolution fetching: keep the raster as source of truth (my preference)

Worth being explicit about why STACDataset reads CRS/res from the raster rather than STAC metadata. 

Two roles:
1. Init — one file is opened to pick a default sampling grid (this is skipped entirely if crs/res are passed as params to STACDataset).

2. Read — every sampled raster is opened anyway, and rasterio takes each source's CRS/transform from the file header (_load_warp_file) to warp it. This open is unavoidable for reading pixels.

So STAC proj:* metadata could only save the single `init` open — the read-time opens happen regardless. 

Given how often proj:code/proj:epsg/proj:transform are missing or wrong in real-world STACs, I'd rather not make that STAC metadata the authority for a one init call saving. 

If the STAC metadata is wrong, the raster is still the truth and rasterio reads it anyway.

Users can still query/filter on crs,res,bounds on a large stac-geoparquet like this today in this PR already - 

```python
  # asset-level resolution, nested inside the assets struct
  ds = STACDataset(
      "items.parquet",
      ("B04", "B08"),
      filters=[("assets.B04.gsd", "<=", 10.0)],     # dotted path into the struct
  )

  # combine top-level CRS (Item property) with nested asset resolution
  ds = STACDataset(
      "items.parquet",
      ("B04",),
      filters=[
          ("proj:code", "==", "EPSG:32632"),        # top-level: properties promote up
          ("assets.B04.gsd", "<=", 10.0),           # nested: asset object stays a sub-struct
      ],
  ) 
```
So I'd propose keeping raster-as-truth for CRS and Resolution as the default rather than trusting STAC metadata.
 
Does that sound reasonable?

adamjstewart:
Can you modify the "User API" in the PR description to be a working example? I'm STAC illiterate and don't know where to find a catalog to play with.

adamjstewart:
Also curious how this implementation compares to #3757, which makes it seem like it only takes 25 lines of code to convert a STAC GeoParquet to a TorchGeo-compatible dataset.

print-sid8:
@adamjstewart 

I read #3757,  and that PR is similar to how we already have different landsat/NAIP datasets in torchgeo but we dont have a generic dataloader for STAC.

MetaCHM PR stac reading is compact precisely because everything is hardcoded for that one product:

- the catalog URL — [`meta_chm.py#L47`](https://github.com/torchgeo/torchgeo/blob/c9654636c08451b9e09c68647919933497c809df/torchgeo/datasets/meta_chm.py#L47)
- CRS, resolution, band, and image/mask flag — [`#L49-L52`](https://github.com/torchgeo/torchgeo/blob/c9654636c08451b9e09c68647919933497c809df/torchgeo/datasets/meta_chm.py#L49-L52) and [`to_crs('EPSG:3857')`](https://github.com/torchgeo/torchgeo/blob/c9654636c08451b9e09c68647919933497c809df/torchgeo/datasets/meta_chm.py#L76)
- the single asset key `chm` and an S3→HTTPS href rewrite — [`#L79`](https://github.com/torchgeo/torchgeo/blob/c9654636c08451b9e09c68647919933497c809df/torchgeo/datasets/meta_chm.py#L79)
- the datetime→interval handling — [`#L86-L89`](https://github.com/torchgeo/torchgeo/blob/c9654636c08451b9e09c68647919933497c809df/torchgeo/datasets/meta_chm.py#L86-L89)


Its `__init__` takes only `transforms` and `cache` ([`#L54-L56`](https://github.com/torchgeo/torchgeo/blob/c9654636c08451b9e09c68647919933497c809df/torchgeo/datasets/meta_chm.py#L54-L56)) — no bbox, time range, filters, CRS/res, or storage options — and it pulls the **entire** 213k-row table into memory over `urllib` on construction ([`#L71-L75`](https://github.com/torchgeo/torchgeo/blob/c9654636c08451b9e09c68647919933497c809df/torchgeo/datasets/meta_chm.py#L71-L75)). That's probably alright for one curated product, and when its known that its `stac-geoparquet` is probably small few MBs. 

It doesn't generalize to "point me at any STAC GeoParquet," and also doesn't scale to filtering a large catalog stac-geoparquet like Sentinel2 in Microsoft planetary computer which is over 2GB in size. 

`STACDataset` is fully parameterized and adds bbox + time + Apache Arrow predicate pushdown (ie., filter into parquet sitting in object store without copying it to local for reading). 

And it already reads MetaCHM's same catalog easily.

Here's a code snippet for it, and I also added a example for accessing AlphaEarth embeddings from its stac-geoparquet too:

```python
import os
from torchgeo.datasets import STACDataset

# (1) Meta Canopy Height v2 — the dataset from #3757, read via STAC GeoParquet

URL = "s3://us-west-2.opendata.source.coop/tge-labs/meta-chm-v2/stac/items.parquet"
chm = STACDataset(
    index_path=URL,
    asset_keys=("chm",),                          # Meta canopy height COG
    intersects_bbox=(-122.5, 47.4, -122.0, 47.8), # Seattle area
    time_range=("2019-01-01", "2020-12-31"),      # tz-naive -> exercises the fix
    storage_options={"anon": True},
)
print("items:", len(chm.index), "| crs:", chm.crs.to_epsg(), "| res:", chm.res)
s = chm[next(iter(RandomGeoSampler(chm, size=64, length=1)))]
print("image shape:", tuple(s['image'].shape), "| dtype:", s['image'].dtype)
```

`items: 6 | crs: 3857 | res: (1.1943285669558463, 1.1943285669558747)`
`image shape: (1, 64, 64) | dtype: torch.float32`

```python
# (2) AlphaEarth Foundation embeddings — 64-band COGs  (registry.opendata.aws/aef-source)

URL = "s3://us-west-2.opendata.source.coop/tge-labs/aef/v1/annual/aef_index_stac_geoparquet.parquet"
aef = STACDataset(
    index_path=URL,
    asset_keys=("data",),
    intersects_bbox=(178.5, -17.8, 179.3, -17.0),   # one AEF tile over Fiji
    time_range=("2025-01-01", "2025-12-31"),
    storage_options={"anon": True},
)
print("index items after filter:", len(aef.index))
print("crs:", aef.crs, "| res:", aef.res)
sampler = RandomGeoSampler(aef, size=64, length=1)
q = next(iter(sampler))
sample = aef[q]
print("sample image shape:", tuple(sample['image'].shape), "| dtype:", sample['image'].dtype)
```
`index items after filter: 7`
`crs: PROJCS["WGS 84 / UTM zone 60S", ...AUTHORITY["EPSG","32760"]] | res: (10.0, 10.0)`
`sample image shape: (64, 64, 64) | dtype: torch.float32`

---

I have update PR description also with same example

fmigneault:
Could a class name like `STACGeoParquetDataset` be considered?

The fact that it doesn’t rely on a STAC API or directory/file structure of STAC Collection/Items might confuse users reading `STACDataset` expecting an input "URL" to point at a STAC live/static endpoint. 

A GeoParquet file embedding STAC items should be clearer directly with a proper name.

print-sid8:
@fmigneault Maybe. But ideally this same Dataset can extend in future to support STAC API too, while being clear that APIs can be way less performant than GPQ files. The PR description above points to this already and also tags previous PRs and discussions in issues and Slack.

Its been a long discussion about STAC API vs STAC geoparquet, stac gpq wins today because of the nature of TorchGeo workloads where feeding GPUs shouldn't ideally be bottle-necked by live API calls in training or inference runs.

fmigneault:
@print-sid8 
Maybe it could be a later addition?

I could see some kind of specialization of an abstract `STACDataset()` where the `STACGeoParquetDataset` provides its concrete `read_index()` https://github.com/torchgeo/torchgeo/pull/3670/changes#diff-ab8028d30bde5e0458e6e72d176d25dd25d751b71b0c4f941c7aec0242b238a6R149-R159 method. Other implementations based on APIs, a directory hierarchy or any other index formats could then easily be added with their own `read_index`. 

The API implemention could be as simple as loading the FeatureCollection obtained from the STAC API search and produce the equivalent GeoDataFrame content that the GeoParquet uses. I think the API could be just as efficient as GeoParquet if it uses proper caching.

print-sid8:
Yes i meant the same thing, we can add STAC API reads to this same Dataset in future. 

about caching and performance - only large servers (like aws element84 earthsearch) can return JSON API + CQL filters results fast, but there can be many servers maintained by academicians or small orgazniations who may either rate limit, or crash if many TorchGeo users try to access the API at once.  

This is also another reason to not support STAC APIs in first iteration.

PR #3671, opened 2026-05-07
Add spatiotemporal segmentation task

robmarkcole:
## Summary

- Added `SpatioTemporalSegmentationTask` for (B, T, C, H, W) semantic segmentation with ConvLSTM.
- Registered the new task in `torchgeo.trainers`
- Updated the PASTIS test config to use SpatioTemporalSegmentationTask + ConvLSTM instead of flattening timesteps into a standard SemanticSegmentationTask.
- Updated `pad_across_batches` to return a `length` tensor so models can select the correct final valid timestep after padding/truncation.
- Updated PASTIS datamodules to use `functools.partial` for the collate function so it remains picklable with multiprocessing workers (i.e. with `num_workers > 0`)
- Added dedicated spatiotemporal trainer tests covering config execution, ConvLSTM kwargs, sequence lengths, losses, prediction behavior, and classwise metrics.
- Removed PASTIS/PASTIS100 from the generic semantic segmentation trainer config test list.

## AI


torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,157 @@
...
+                self.model = ConvLSTM(
+                    input_dim=in_channels,
+                    hidden_dim=hidden_dim,
+                    kernel_size=kernel_size,
+                    num_layers=num_layers,
+                    bias=bias,
+                    return_all_layers=return_all_layers,
+                    num_classes=num_classes,
+                    head_kernel_size=head_kernel_size,
+                )
```
calebrob6: Should we raise a Value error if nothing matches?

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,157 @@
...
+
+        loss: Tensor = self.criterion(y_hat, y)
+        self.log(f'{stage}_loss', loss, batch_size=batch_size)
+        return loss
+
+    def training_step(
+        self, batch: Any, batch_idx: int, dataloader_idx: int = 0
+    ) -> Tensor:
+        """Compute the training loss and additional metrics."""
+        del batch_idx, dataloader_idx
```
calebrob6: Why delete these?

tests/datasets/test_utils.py
```diff
@@ -605,11 +605,13 @@ def test_pad_across_batches() -> None:
     out = pad_across_batches(batch, padding_value=0.0, padding_length=3)
     assert out['image'].shape[1] == 3
     assert out['mask'].shape[0] == len(batch)
+    assert torch.equal(out['length'], torch.tensor([2, 3], device=out['length'].device))
```
calebrob6, line 608: What is this testing?

calebrob6 (commented):
More generally, does this work with PASTIS?

calebrob6:
Probably needs to go in the docs too

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: yes will now raise

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: removed

tests/datasets/test_utils.py
robmarkcole: comment added

robmarkcole:
@calebrob6 re docs this should be automatic https://github.com/torchgeo/torchgeo/blob/main/docs/api/trainers.rst

Below, results of 10 epoch train on pastis100:

image

calebrob6 (dismissed):

yichiac:
> image

@robmarkcole

Does the black color mean no data?

Thanks for creating the trainer and updating the PASTIS datamodule!

tests/trainers/test_segmentation.py
```diff
@@ -67,8 +67,6 @@ class TestSemanticSegmentationTask:
             'loveda',
             'mmflood',
             'naipchesapeake',
-            'pastis',
-            'pastis100',
```
adamjstewart, line 73: This results in decreased test coverage for SemanticSegmentationTask, we no longer test that time series input works.

tests/datamodules/test_pastis.py
adamjstewart, line 1: See https://github.com/torchgeo/torchgeo/blob/main/tests/datamodules/README.md

tests/trainers/test_spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,200 @@
...
+        try:
+            main(['predict', *args])
+        except MisconfigurationException:
+            pass
+
+    @pytest.fixture
+    def create_spatiotemporal_model(
+        self,
+    ) -> Callable[..., SpatioTemporalSegmentationTask]:
+        def _create_spatiotemporal_model(
```
adamjstewart: This is quite complex: a closure inside of a fixture. The name is also quite confusing because it returns a trainer/task/lightningmodule, not a model. Is this complexity really needed? Can't this be replaced by:
```python
model = SpatioTemporalSegmentationTask()
model.model = ...
```

tests/trainers/test_spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,200 @@
...
+            **kwargs: Any,
+        ) -> SpatioTemporalSegmentationTask:
+            model = SpatioTemporalSegmentationTask(hidden_dim=8, num_layers=1, **kwargs)
+            # Avoid Lightning warnings when calling step hooks without a Trainer.
```
adamjstewart: Why are we calling step hooks? Don't we already get coverage from the above unit test?

tests/trainers/test_spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,200 @@
...
+import torch.nn as nn
+
+from torchgeo.datamodules import MisconfigurationException
+from torchgeo.main import main
+from torchgeo.models import ConvLSTM
+from torchgeo.trainers import SpatioTemporalSegmentationTask
+
+
+class TestSpatioTemporalSegmentationTask:
+    def test_trainer_with_pastis_config(self, fast_dev_run: bool) -> None:
```
adamjstewart: This should not be specific to pastis, we will eventually add other datasets to this test. Let's parametrize it like all other trainer tests.

tests/trainers/test_spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,200 @@
...
+        ) -> SpatioTemporalSegmentationTask:
+            model = SpatioTemporalSegmentationTask(hidden_dim=8, num_layers=1, **kwargs)
+            # Avoid Lightning warnings when calling step hooks without a Trainer.
+            setattr(model, 'log', lambda *args, **kwargs: None)
+            setattr(model, 'log_dict', lambda *args, **kwargs: None)
+            return model
+
+        return _create_spatiotemporal_model
+
+    def test_spatiotemporal_forward_defaults_to_convlstm(self) -> None:
```
adamjstewart: Vast majority of these tests are redundant and don't add any additional coverage. I think most of these can be removed.

tests/trainers/test_spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,200 @@
...
+        assert model.model.head is not None
+
+    def test_spatiotemporal_forward_supports_direct_convlstm_kwargs(self) -> None:
+        model = SpatioTemporalSegmentationTask(
+            in_channels=3, num_classes=5, hidden_dim=8, num_layers=1
+        )
+        assert isinstance(model.model, ConvLSTM)
+        assert model.model.hidden_dim == [8]
+
+    def test_spatiotemporal_invalid_model(self) -> None:
```
adamjstewart: Could be removed if we don't care about testing invalid inputs.

tests/trainers/test_spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,200 @@
...
+        invalid_model: Any = 'invalid'
+        match = "Invalid model type 'invalid'. Supported model: 'convlstm'"
+        with pytest.raises(ValueError, match=match):
+            SpatioTemporalSegmentationTask(model=invalid_model)
+
+    def test_spatiotemporal_direct_kwargs_are_saved_in_hparams(self) -> None:
```
adamjstewart: We don't need to test Lightning, just TorchGeo.

tests/trainers/test_spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,200 @@
...
+        assert isinstance(model.criterion, nn.CrossEntropyLoss)
+        torch.testing.assert_close(
+            model.criterion.weight, torch.tensor([1.0, 2.0], dtype=torch.float32)
+        )
+
+    @pytest.mark.parametrize(
+        ('loss', 'expected_type'),
+        [('jaccard', smp.losses.JaccardLoss), ('focal', smp.losses.FocalLoss)],
+    )
+    def test_alternate_losses(
```
adamjstewart: These could be placed in `pastis_jaccard.yaml`, etc. so that this file is only 1 line longer

tests/trainers/test_spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,200 @@
...
+        )
+        batch = {'image': torch.randn(2, 4, 3, 16, 16), 'length': torch.tensor([4, 3])}
+
+        probabilities = model.predict_step(batch, 0)
+        assert probabilities.shape == (2, 4, 16, 16)
+        torch.testing.assert_close(
+            probabilities.sum(dim=1), torch.ones((2, 16, 16)), atol=1e-5, rtol=1e-5
+        )
+
+    def test_multiclass_classwise_metrics(
```
adamjstewart: Don't really understand the purpose of this test

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,160 @@
...
+    """
+
+    def __init__(
+        self,
+        model: Literal['convlstm'] = 'convlstm',
+        in_channels: int = 3,
+        task: Literal['binary', 'multiclass', 'multilabel'] = 'multiclass',
+        num_classes: int | None = None,
+        num_labels: int | None = None,
+        labels: list[str] | None = None,
```
adamjstewart, line 29: Does it have to be a list or will any Sequence work?

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,160 @@
...
+                    kernel_size=kernel_size,
+                    num_layers=num_layers,
+                    bias=bias,
+                    return_all_layers=return_all_layers,
+                    num_classes=num_classes,
+                    head_kernel_size=head_kernel_size,
+                )
+            case _:
+                msg = f"Invalid model type '{model}'. Supported model: 'convlstm'"
+                raise ValueError(msg)
```
adamjstewart: I would personally remove this, that's what we have type hints for

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,160 @@
...
+                kernel_size = cast(
+                    int | tuple[int, int] | list[int | tuple[int, int]],
+                    self.hparams.get('kernel_size', 3),
+                )
+                num_layers = cast(int, self.hparams.get('num_layers', 1))
+                head_kernel_size = cast(int, self.hparams.get('head_kernel_size', 1))
+                bias = cast(bool, self.hparams.get('bias', True))
+                return_all_layers = cast(
+                    bool, self.hparams.get('return_all_layers', False)
+                )
```
adamjstewart: Can we get rid of all of this and instead use `**kwargs`? Or does it not work that way? I don't like setting default values both here and in ConvLSTM, they are bound to diverge.

adamjstewart (changes requested):
Current tests appear to be auto-generated and are quite different from our existing tests. I think a lot of the testing could be made drastically simpler without losing any coverage or rigor.

tests/trainers/test_segmentation.py
robmarkcole: added back

tests/datamodules/test_pastis.py
robmarkcole: deleted

tests/trainers/test_spatiotemporal_segmentation.py
robmarkcole: Replaced with a simple method

tests/trainers/test_spatiotemporal_segmentation.py
robmarkcole: done

tests/trainers/test_spatiotemporal_segmentation.py
robmarkcole: Removed

tests/trainers/test_spatiotemporal_segmentation.py
robmarkcole: Removed

tests/trainers/test_spatiotemporal_segmentation.py
robmarkcole: Done

tests/trainers/test_spatiotemporal_segmentation.py
robmarkcole: Was testing `labels` is passed along, removed

tests/trainers/test_spatiotemporal_segmentation.py
robmarkcole: Removed 

- test_spatiotemporal_forward_defaults_to_convlstm
- test_spatiotemporal_forward_supports_direct_convlstm_kwargs
- test_ce_class_weights_from_sequence 
- test_convlstm_timeseries_forward_and_step

tests/trainers/test_spatiotemporal_segmentation.py
robmarkcole: Removed now

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: [ClasswiseWrapper](https://lightning.ai/docs/torchmetrics/stable/wrappers/classwise_wrapper.html) requires `list[str]`

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: Removed

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: Yes, changed

robmarkcole:
@yichiac black is `background`, white is `nodata`

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: [Tests then fail](https://github.com/torchgeo/torchgeo/actions/runs/25911621913/job/76157975528?pr=3671) with 
```
>       self.model = ConvLSTM(input_dim=in_channels, num_classes=num_classes, **kwargs)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       TypeError: ConvLSTM.__init__() missing 1 required positional argument: 'kernel_size'
```
Also  jsonargparse injects `_class_path` into `self.hparams` which passes into ConvLSTM and causes another error

Could read defaults from ConvLSTM, but it is a bit ugly. Suggest the simplest approach is just to accept the duplication and hard code defaults here. Another option would be to add a ConvLSTM config dataclass to be shared by both. WDYT?

tests/conf/pastis.yaml
```diff
@@ -1,16 +1,41 @@
 model:
-  class_path: SemanticSegmentationTask
+  class_path: SpatioTemporalSegmentationTask
```
adamjstewart, line 2: If this is being tested in `tests/trainers/test_segmentation.py`, shouldn't it use SemanticSegmentationTask?

tests/trainers/test_spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,79 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+
+import os
+from unittest.mock import MagicMock
```
adamjstewart: You should instead be using `pytest.MonkeyPatch`, none of our other unit tests rely on `unittest.mock`.

torchgeo/datamodules/pastis.py
adamjstewart, line 1: This could technically be a separate PR but I'm okay with keeping it here if you aren't in a rush to merge this bug fix.

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,148 @@
...
+            return_all_layers=self.hparams['return_all_layers'],
+            num_classes=num_classes,
+            head_kernel_size=self.hparams['head_kernel_size'],
+        )
+
+    def _shared_step(self, batch: Any, stage: str) -> Tensor:
+        """Compute the loss and metrics for the given stage."""
+        x = batch['image']
+        y = batch['mask']
+        lengths = batch.get('length')
```
adamjstewart: Are all datasets guaranteed to have this and all models guaranteed to support this? I'm worried about how specific this is to PASTIS, what if I use a different dataset that doesn't contain variable-length sequences and doesn't need padding?

tests/trainers/test_spatiotemporal_segmentation.py
adamjstewart: Looks much simpler now, thanks!

torchgeo/trainers/spatiotemporal_segmentation.py
adamjstewart: > Tests then fail with...

Just need to add a default `kernel_size` to fix that.

Have you tried adding `**kwargs: Any` and `self.kwargs = kwargs` to `__init__` and then using `**self.kwargs` when creating the model? Then you don't have to go through `self.hparams` where all kwargs are mixed in with args.

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: Defaults to `none` if not present, in which case ConvLSTM falls back to the last timestep. Fixed-length datasets just work without it. Currently only DigitalTyphoon dataset provides `length`

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: Implemented your suggestion, much neater!

tests/trainers/test_spatiotemporal_segmentation.py
robmarkcole: Done

tests/conf/pastis.yaml
robmarkcole: SemanticSegmentationTask doesn't support conlstm, removed from test_segmentation

tests/trainers/test_spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,82 @@
...
+        model = SpatioTemporalSegmentationTask(
+            in_channels=3, task='binary', loss='bce', hidden_dim=8, num_layers=1
+        )
+
+        # Avoid Lightning warnings when calling step hooks without a Trainer.
+        def noop(*_args: object, **_kwargs: object) -> None:
+            pass
+
+        monkeypatch.setattr(model, 'log', noop)
+        monkeypatch.setattr(model, 'log_dict', noop)
```
adamjstewart: Can we instead use a trainer? This is what we do in other tests: https://github.com/torchgeo/torchgeo/blob/v0.9.0/tests/trainers/test_classification.py#L236

tests/trainers/test_spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,82 @@
...
+
+        assert model.validation_step(batch, 0) is None
+        assert model.test_step(batch, 0) is None
+
+        probabilities = model.predict_step(batch, 0)
+        assert probabilities.shape == (2, 1, 16, 16)
+        assert torch.all(probabilities >= 0)
+        assert torch.all(probabilities <= 1)
+
+    def test_multiclass_predict_step(self) -> None:
```
adamjstewart: I'm guessing this one will need the same treatment? Shouldn't this test multilabel, not multiclass? Multiclass is already tested by PASTIS.

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,133 @@
...
+            class_weights: Optional rescaling weight given to each
+                class and used with 'ce' loss.
+            ignore_index: Optional integer class index to ignore in the loss and
+                metrics.
+            lr: Learning rate for optimizer.
+            patience: Patience for learning rate scheduler.
+            **kwargs: Additional keyword arguments passed to the model constructor.
+                For ``'convlstm'``: ``hidden_dim``, ``kernel_size``, ``num_layers``,
+                ``bias``, ``return_all_layers``, ``head_kernel_size``.
+
```
adamjstewart: I would remove this, users can look up the ConvLSTM docs on their own and we will eventually have many more models supported.

```suggestion
```

torchgeo/trainers/spatiotemporal_segmentation.py
adamjstewart: This explains that datasets without length are okay, but what about models without a length parameter? Will OlmoEarth be okay for example?

tests/trainers/test_spatiotemporal_segmentation.py
robmarkcole: Replaced with a trainer

tests/trainers/test_spatiotemporal_segmentation.py
robmarkcole: replaced with untested multilabel

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: Olmo performs padding internally but doesn't support a `lengths` arg and I don't think we can assume that all other models would. So we will have to (1) pass only when not none and (2) check the model supports it first.

robmarkcole: @adamjstewart replaced with `self.model(x, **kwargs)`

robmarkcole:
10 epoch validation on pastis100:

image

tests/trainers/test_spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,86 @@
...
+import torch
+from lightning.pytorch import Trainer
+from torch.utils.data import DataLoader, Dataset
+
+from torchgeo.datamodules import MisconfigurationException
+from torchgeo.main import main
+from torchgeo.trainers import SpatioTemporalSegmentationTask
+
+
+class BinarySegDataset(Dataset):  # type: ignore[type-arg]
```
adamjstewart:
```suggestion
class BinarySegDataset(NonGeoDataset):
```

tests/trainers/test_spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,86 @@
...
+    def __len__(self) -> int:
+        return 4
+
+    def __getitem__(self, index: int) -> dict[str, Any]:
```
adamjstewart:
```suggestion
    def __getitem__(self, index: int) -> Sample:
```

torchgeo/models/convlstm.py
```diff
@@ -234,6 +234,7 @@ def forward(
         input_tensor: torch.Tensor,
         lengths: torch.Tensor | None = None,
         hidden_state: list[tuple[torch.Tensor, torch.Tensor]] | None = None,
+        **kwargs: Any,
```
adamjstewart: This should not be here, this will result in mispelled args being ignored instead of an error.

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,135 @@
...
+        self.model = ConvLSTM(
+            input_dim=in_channels, num_classes=num_classes, **self.kwargs
+        )
+
+    def _shared_step(self, batch: Any, stage: str) -> Tensor:
+        """Compute the loss and metrics for the given stage."""
+        x = batch['image']
+        y = batch['mask']
+        batch_size = x.shape[0]
+        kwargs: dict[str, Any] = {}
```
adamjstewart: Can we assert that these will always be Tensors?

torchgeo/trainers/spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def test_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) -> None:
+        """Compute the test loss and additional metrics."""
+        self._shared_step(batch, 'test')
+
+    def predict_step(
+        self, batch: Any, batch_idx: int, dataloader_idx: int = 0
+    ) -> Tensor:
+        """Compute the predicted class probabilities."""
+        kwargs: dict[str, Any] = {}
```
adamjstewart: Same here

torchgeo/models/convlstm.py
```diff
@@ -243,10 +244,10 @@ def forward(
                 Values larger than the available sequence length use the final
                 timestep.
             hidden_state: An optional initial hidden state.
+            **kwargs: Additional keyword arguments (ignored).
```
adamjstewart: Same here, can remove

tests/trainers/test_spatiotemporal_segmentation.py
```diff
@@ -0,0 +1,86 @@
...
+            in_channels=3, task='binary', loss='bce', hidden_dim=8, num_layers=1
+        )
+        trainer = Trainer(
+            accelerator='cpu',
+            fast_dev_run=fast_dev_run,
+            log_every_n_steps=1,
+            max_epochs=1,
+        )
+        trainer.fit(model, train_dataloaders=dataloader, val_dataloaders=dataloader)
+        trainer.test(model, dataloaders=dataloader)
```
adamjstewart: Actually, now that I look at this, `test_multilabel_predict_step` doesn't have a Trainer and doesn't have any warnings. Maybe this isn't needed and we can simplify this to match?

adamjstewart (changes requested):
Getting closer!

torchgeo/models/convlstm.py
robmarkcole: removed

torchgeo/models/convlstm.py
robmarkcole: removed

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: now typed as tensor

torchgeo/trainers/spatiotemporal_segmentation.py
robmarkcole: now typed as tensor

tests/trainers/test_spatiotemporal_segmentation.py
robmarkcole: Simplified to match

adamjstewart (approved):
Now _this_ is a good PR! Sorry it took so many iterations, and thanks for keeping up with my review requests!

robmarkcole:
Quality takes time, thanks @adamjstewart and @calebrob6  !

PR #3674, opened 2026-05-07
GeoDataset: ensure axis-aligned bounding boxes after reprojection

calebrob6:
Fixes a regression introduced when GeoDataset migrated from rtree to geopandas (#2747). Running

```
python -m torchgeo fit --config tests/conf/io_raw.yaml
```

fails with a spurious `IndexError` from CDL.

When reprojecting in the IntersectionDataset, geopandas turns the index geometry from a bbox into an actual polygon (which is curved in the IoBench case). `RandomGeoSampler` then samples from .bounds, and some samples will be outside the intersection area.

This adds tests so this doesn't happen again. Would also love IOBench to be an integration test.

I verified that IOBench runs after this

### DIFF
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -129,6 +129,19 @@ class TestGeoDataset:
     def test_crs(self, dataset: GeoDataset, crs: CRS) -> None:
         dataset.crs = crs
 
+    def test_crs_geometry_envelope(self) -> None:
+        # Reprojection densifies edges and turns axis-aligned bounding boxes
+        # into curved polygons. The crs setter must restore them to envelopes
+        # so samplers (which sample within geometry.bounds) never produce
+        # queries that fall outside the dataset's geometry.
+        ds = CustomGeoDataset(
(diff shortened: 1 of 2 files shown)


isaaccorley (approved):

adamjstewart (changes requested):
There are actually two places we could fix this:

1. **GeoDataset**: go back to axis-aligned bounding boxes (partially reverts #2747)
2. **GeoSampler**: ensure samples overlap with geometry before yielding

I opted for 1 in #3552 which also fixes this same issue. If this PR is merged as is, it won't fix the issue for arbitrary Polygon `roi`s in the samplers or splitters. It will also need to be immediately reverted once #3552 is merged, resulting in the old samplers remaining broken. Note that @adriantre is also working on several PRs to purposefully add non-axis-aligned Polygons to the index so we can properly avoid nodata pixels: #2903, #3663. I don't want to render his hard work obsolete.

So my preference would be to fix this in the (soon-to-be-deprecated) samplers instead of in the dataset.

calebrob6:
Sounds great! I'm not up-to-date on either of those efforts (other than trying to test the samplers for correctness and speed) so definitely defer to your and @adriantre 's work here!

PR #1559, opened 2023-09-12
Add command line script

adamjstewart:
This PR replaces our `train.py` script with a `torchgeo` script installed by pip. It can be invoked in two ways:
```bash
# If torchgeo has been installed
torchgeo
# If torchgeo has been installed, or if it has been cloned to the current directory
python3 -m torchgeo
```
It is based on LightningCLI, and supports command-line configuration or YAML/JSON config files. Valid options can be found from the help messages:
```bash
# See valid stages
torchgeo --help
# See valid trainer options
torchgeo fit --help
# See valid model options
torchgeo fit --model.help ClassificationTask
# See valid data options
torchgeo fit --data.help EuroSAT100DataModule
```
An example of the script in action:
```bash
# Train and validate a model
torchgeo fit --config tests/conf/eurosat100.yaml --trainer.max_epochs=1
# Validate-only
torchgeo validate --config tests/conf/eurosat100.yaml
# Calculate and report test accuracy
torchgeo test --config tests/conf/eurosat100.yaml
```
It can also be imported and used in a Python script if you need to extend it to add new features:
```python
from torchgeo.main import main

main(["fit", "--config", "tests/conf/eurosat100.yaml"])
```

### Documentation

Setuptools packaging:

* https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html
* https://docs.python.org/3/library/__main__.html
* https://setuptools.pypa.io/en/latest/userguide/entry_point.html

LightningCLI:

* https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.cli.LightningCLI.html
* https://lightning.ai/docs/pytorch/stable/cli/lightning_cli.html

Closes #228 
Closes #1352

PR #3675, opened 2026-05-07
Replace IOBench with a script

calebrob6:
Our current IOBench is way too complicated and not useful. This PR replaces 226 lines of code across 6 files with a single 116 line (overly-verbose!) script.

## Motivation

If, as a new user to torchgeo, I want to understand what is going on in `python -m torchgeo fit --config tests/conf/io_raw.yaml`, then I will go to that yaml conf:

```
model:
  class_path: IOBenchTask
data:
  class_path: IOBenchDataModule
  dict_kwargs:
    root: 'data/io'
    split: 'raw'
    download: true
    checksum: true
trainer:
  max_epochs: 1
  num_sanity_val_steps: 0
  profiler: 'simple'
```

This redirects me to an 80 line no-op file (IOBenchTask) and a 65 line datamodule.

When I actually run `torchgeo fit ...` command on my machine, the trainer tries to start using all my GPUs, which hangs. I then need to figure out how to run it on CPU only, `--trainer.accelerator cpu`. When I get there, the output looks something like:

```
Epoch 0/0  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 31/31 0:01:39 • 0:00:00 0.35it/s v_num: 6.000
`Trainer.fit` stopped: `max_epochs=1` reached.
Epoch 0/0  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 31/31 0:01:39 • 0:00:00 0.35it/s v_num: 6.000
FIT Profiler Report

--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
|  Action                                                                                                                                                               |  Mean duration (s)    |  Num calls            |  Total time (s)       |  Percentage %     |
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
|  Total                                                                                                                                                                |  -                    |  1879                 |  213.79               |  100 %            |
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
|  run_training_epoch                                                                                                                                                   |  213.39               |  1                    |  213.39               |  99.816           |
|  [_EvaluationLoop].val_next                                                                                                                                           |  3.6015               |  31                   |  111.65               |  52.222           |
|  [_TrainingEpochLoop].train_dataloader_next                                                                                                                           |  3.1316               |  31                   |  97.079               |  45.409
...
this goes on for quite awhile
...
```

Now, I'm really unclear what number I should be looking at/reporting unless I've read the contributing docs (which I guess I need to do if I want to figure out how to run this in the first place 🙂). The total time reported by the tqdm bar is 99 seconds, the training epoch is 213 seconds, train_dataloader_next is 97 seconds. It is unintuitive to map train/val to random/gridgeosampler. How can I change to use a different sampler? How can I rerun with different numbers of workers? Will this change with the next version of lightning? All questions I don't know!

Also, the tests for IOBench weren't helpful anyway as they didn't catch the fact that it hasn't been able to be run for 3 versions!!

## Proposed Change

`python scripts/iobench.py` you run this and it downloads and runs the IOBench benchmarking. Pros:
- Single file
- Very easy to understand what is happening and change it if you want
- Very easy to chain different samplers
- Very easy to maintain
- Very hard to misinterpret the output
- Doesn't depend on the whims of lightning trainer defaults
- Will never try to run on all my GPUs

Cons:
- Having a script in `scripts/` might seem scary, but this is actually a pro

Output looks like this:
```
$ python scripts/iobench.py --root data/io --num-workers 8
IOBench benchmark
root=data/io batch_size=32 patch_size=256 num_workers=8

=== split=raw ===
Converting CDL CRS from PROJCS["NAD83 / Conu..........
Converting CDL res from (30.000255453565156, 30.00162369548359) to (30.0, 30.0)
raw: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 31/31 [00:09<00:00,  3.26it/s]
samples:       992
batches:       31
total time:    9.503 s
samples/sec:   104.38
batches/sec:   3.26

=== split=preprocessed ===
preprocessed: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 31/31 [00:05<00:00,  5.83it/s]
samples:       992
batches:       31
total time:    5.318 s
samples/sec:   186.53
batches/sec:   5.83
```

If you want it to use the test data you can do `python scripts/iobench.py --root tests/data/iobench/ --split preprocessed`


adamjstewart:
How do I use this after `pip install torchgeo`?

calebrob6:
You obviously can't? What kind of question is that?

Here's a counter rhetorical question: in what scenarios are you running the IOBench command without having a local copy of torchgeo cloned (where you *can* run this script)? The point of IOBench is to benchmark the performance impact of changes in torchgeo.

adamjstewart:
> in what scenarios are you running the IOBench command without having a local copy of torchgeo cloned 

The scenario in which a user complains that sampling performance is slow and I ask them to measure _how_ slow.

calebrob6:
How often does that come up? Do those users have `git` installed or are there scenarios where you are trying to `pip install torchgeo` on different envs and benchmark and gather results? Are you actually asking for this to be subset under the `torchgeo` command? Let's be constructive.

adamjstewart:
> Let's be constructive.

Ignoring all of the [obviously inflammatory rhetoric](https://github.com/torchgeo/governance/blob/main/CODE-OF-CONDUCT.md), here are some constructive review comments:

1. The whole purpose of our current CLI was to unify all of our old `train.py`, `eval.py`, `bench.py` scripts under one shared CLI (#1559). I would prefer not to go back to having multiple separate scripts with duplicated logic and subtle differences in flag names.
2. LightningCLI is very convenient for this as it comes with hundreds of built-in features, documentation, help messages, and support (one less thing _we_ need to maintain). It also provides an identical interface for all users and developers, regardless of if you are benchmarking dataset I/O, sampler speed, or model performance. For new users already familiar with Lightning, it's one less thing they need to learn.
3. This PR doesn't allow me to measure distributed I/O performance, this is much easier with LightningCLI.
4. This PR deletes datasets, data modules, and trainers without first deprecating them. In #2578 we decided to first deprecate anything before removal (within reason of course). At the moment, this is an immediate backwards-incompatible break that means I need to use different syntax to measure I/O performance across different versions of TorchGeo.
5. We don't yet have a direct dependency on tqdm, this would need to be added to `pyproject.toml`/`uv.lock`. I think it's an indirect dependency anyway, so I'm fine with this.
6. You're missing your AI disclosure.

To be completely fair, I also have complaints about I/O Bench that are partially addressed in this PR or could reasonably be fixed in our current implementation instead:

1. Detailed profiler output is sometimes nice (not just _is_ it slow, but _why_ is it slow), but most of the time too verbose. We could write a custom profiler subclass that grabs only the things we care about, renames them to be more clear, and formats them as a valid Markdown table.
2. I have to run the script twice (raw and preprocessed) per branch to get the metrics I want. This makes it a little harder to format the output as a valid table. I don't have a great workaround for this in the Lightning ecosystem, this is definitely easier to do in a custom script.
3. I actually care about sampling _rate_ AND sampling _time_. One of the things I noticed in #3552 was that some of the performance differences are due to different epoch length, not different I/O rates. Both are important, it would be nice to have both like in this PR.
4. The fact that GeoDataset broke and we didn't notice for multiple releases is obviously bad, we should add more unit tests.
5. The current implementation also requires a git clone of the repo to get the config.yaml (does Lightning support URLs?). I would like to figure out how to make this work without any git clone.

I think starting off with a discussion of pain points and proposed solutions before immediately going to a PR would be more helpful. I'm not yet convinced we need to throw the baby out with the bath water just to make the output simpler.

calebrob6:
1. Sorry! I'm not trying to be obviously inflammatory. I was trying to figure out if "How do I use this after pip install torchgeo?" was your main concern (I'm still not entirely sure). I know you knew immediately it _wouldn't_ usable without a clone, and I imagined you had much more relevant/pressing concerns, so was confused about why you are asking that question.

2. Great points, thanks -- I wasn't considering deprecation at all! Feel free to be like "nah, this isn't aligned" instead of spending actual valuable time. I'll definitely open an issue next time for discussion.

3. I'm going to close this as the underlying problem that I'm actually trying to solve here is benchmarking the new samplers. That led me to the GeoDataset bug, then fixing that led me directly to IOBench trying to use all GPUs, then fixing that led me to not knowing how to actually change which sampler IOBench uses (you need a new datamodule), then changing that led me to having to parse like 50 lines of profiler output -- none of which I know for sure how to interpret (what all exactly is measured in `train_dataloader_next`?). This trace through torchgeo's guts, to do something that should be relatively simple, makes me ask, "is there a better way to do this?" As I show above, you _can_ do this with a standalone ~100 line script where it is clear what exactly is being timed and clear how to extend to support different combinations of samplers/datasets. I'll admit I have **zero** sense of whether there is value in: distributed I/O performance, using benchmark scripts after pip installing torchgeo, or the contribution of LightningCLI to the IO benchmarking experience (running `torchgeo fit --help` gives me like 3 pages of output but none mention `num_workers` or `patch_size` -- it seems like I would need to know to go to the configs to change these and separately know how to change them).

adamjstewart:
No worries, I was still recovering from NeurIPS and defensive about AI undoing hard work. I do think many of the points raised here are valid, and there are certainly ways I/O Bench could be improved. Let me open a new issue and we can see which of these are actionable.

PR #3699, opened 2026-05-14
Build backend: setuptools -> uv-build

adamjstewart:
We've successfully dodged plenty of new hype build tools in the past (flit, poetry, hatch) but uv feels different. Since we are actively using it as a build _frontend_, we should consider using it as a build _backend_ too.

### Pros

* 8x faster than setuptools (1.4 s -> 170 ms)

### Cons

* Does not yet support dynamic version metadata: https://github.com/astral-sh/uv/issues/11718
* Lacks a lot of other setuptools features, but we don't use those right now anyway
* A bit buggy in my experience: https://github.com/astral-sh/uv/issues/19403, https://github.com/astral-sh/uv/issues/19404

Any other pros/cons I'm missing that could help us decide?

## AI Usage Disclosure

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,13 +1,14 @@
 [build-system]
 requires = [
-    # setuptools 77.0.1+ required for PEP 639 support
-    "setuptools>=77.0.1",
+    # uv-build 0.7+ required for bdist reorganization
+    "uv_build>=0.7,<0.12",
 ]
-build-backend = "setuptools.build_meta"
+build-backend = "uv_build"
 
 # https://packaging.python.org/en/latest/specifications/declaring-project-metadata/
 [project]
 name = "torchgeo"
+version = "0.10.0.dev0"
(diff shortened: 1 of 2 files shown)


adamjstewart:
> Does not yet support dynamic version metadata

I think they intend for us to use `uv version --bump` instead of manually editing multiple files. For example:

* 0.10.0.dev0 to 0.10.0: `uv version --bump stable`
* 0.10.0 to 0.10.1: `uv version --bump patch`
* 0.10.0 to 0.11.0.dev0: `uv version --bump minor --bump dev`

This seems okay in my mind, we can hard-code this in the release instructions.

robmarkcole:
This looks like a welcome simplification/standardisation on the `uv` ecosystem, but I am not familiar with the tools it is replacing so defer to others for opinions

robmarkcole:
@adamjstewart who would be a good reviewer?

isaaccorley (approved):
This lgtm and TIL about `uv version bump`!

nilsleh (approved):

PR #3702, opened 2026-05-15
Documentation: fix uv installation instructions

adamjstewart:
Previous instructions don't work at all?
```console
> uv add --editable .
error: Requirement name `torchgeo` matches project name `torchgeo`, but self-dependencies are not permitted without the `--dev` or `--optional` flags. If your project name (`torchgeo`) is shadowing that of a third-party dependency, consider renaming the project.
```

@nuglifeleoji

## AI Usage Disclosure

### DIFF
--- a/docs/user/installation.rst
+++ b/docs/user/installation.rst
@@ -66,7 +66,8 @@ To develop TorchGeo itself, clone the repository, create a new virtual environme
    $ git clone https://github.com/torchgeo/torchgeo.git
    $ cd torchgeo
    $ uv venv
-   $ . .venv/bin/activate
+   $ source .venv/bin/activate  # macOS/Linux
+   $ .venv\Scripts\activate     # Windows
    $ uv sync
 
 See the `uv documentation <https://docs.astral.sh/uv/>`_ for more details.
--- a/docs/user/installation.rst
+++ b/docs/user/installation.rst
@@ -66,7 +66,7 @@ To develop TorchGeo itself, clone the repository, create a new virtual environme
    $ git clone https://github.com/torchgeo/torchgeo.git
    $ cd torchgeo
    $ uv venv
-   $ . .venv/bin/activate.sh
+   $ . .venv/bin/activate
    $ uv sync
 
 See the `uv documentation <https://docs.astral.sh/uv/>`_ for more details.
--- a/docs/user/installation.rst
+++ b/docs/user/installation.rst
@@ -57,7 +57,7 @@ Optional dependencies can be added using extras:
 
 .. code-block:: console
 
-   $ uv add torchgeo[datasets,models]
+   $ uv add torchgeo --extra datasets --extra models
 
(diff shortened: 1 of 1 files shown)


docs/user/installation.rst
```diff
@@ -59,18 +59,14 @@ Optional dependencies can be added using extras:
...
 
    $ git clone https://github.com/torchgeo/torchgeo.git
    $ cd torchgeo
-   $ uv add --editable .
-
-To sync all dependencies after making changes:
-
-.. code-block:: console
-
+   $ uv venv
```
robmarkcole: Not sure which version of python you get with this? My boilerplate is

```
uv venv --python 3.12
source .venv/bin/activate
uv sync
```

robmarkcole (dismissed):

docs/user/installation.rst
adamjstewart: You get 3.13 by default right now, I manually switch to 3.14. Doesn't really matter as long as it's compatible.

docs/user/installation.rst
```diff
@@ -59,18 +59,14 @@ Optional dependencies can be added using extras:
...
+   $ . .venv/bin/activate.sh
```
ashnair1: Should be `activate` right? Ref: https://docs.astral.sh/uv/pip/environments/#using-a-virtual-environment

```diff
- $ . .venv/bin/activate.sh
+ $ . .venv/bin/activate
```

adamjstewart: Oh you're right, good catch. I use activate.fish on my system so just assumed they would keep the extension.

ashnair1 (dismissed):

docs/user/installation.rst
```diff
@@ -66,7 +66,7 @@ To develop TorchGeo itself, clone the repository, create a new virtual environme
    $ git clone https://github.com/torchgeo/torchgeo.git
    $ cd torchgeo
    $ uv venv
-   $ . .venv/bin/activate.sh
+   $ . .venv/bin/activate
```
robmarkcole: I think the exact command may depend on platform

adamjstewart: I suppose we could link to https://docs.astral.sh/uv/pip/environments/#using-a-virtual-environment or provide both with a comment?

adamjstewart: Technically it also depends on shell, and I'm one of the crazy people not using bash, but even the uv instructions don't go into that much detail.

robmarkcole (dismissed):
LGTM, linking UV docs would be a bonus

adamjstewart:
We do link to the general uv docs above and below, just not specific to activation.

adamjstewart:
I removed the venv stuff, I don't think it's required and just makes things more complicated. Users can and should read the uv docs if they would like to become power users.

ashnair1 (approved):

PR #3703, opened 2026-05-15
Devcontainer: use uv

adamjstewart:
I have no idea how to test this. Does anyone know how to actually use a devcontainer? Does anyone still use these things or is agentic AI the solution nowadays?

## AI Usage Disclosure

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -10,7 +10,7 @@
   },
 
   // "postCreateCommand": "",
-  "postStartCommand": "uv sync",
+  "postStartCommand": "uv sync --extra all",
 
   // Enable GPU support
   "hostRequirements": {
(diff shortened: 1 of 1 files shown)


robmarkcole:
If you use a codespace, devcontainer is used. I do use this 
image

No issues with this branch so approving:
image

.devcontainer/devcontainer.json
```diff
@@ -10,7 +10,7 @@
   },
 
   // "postCreateCommand": "",
-  "postStartCommand": "pip install -e .[all]",
+  "postStartCommand": "uv sync",
```
robmarkcole: I think this is equivalent to `uv sync --extra all` but might be better to be explicit

robmarkcole (dismissed):

.devcontainer/devcontainer.json
adamjstewart: I think it's actually not equivalent, let's be more explicit.

robmarkcole (approved):

Issue #3684, opened 2026-05-09
Revamping I/O Bench


adamjstewart:
### Summary

We use an I/O Bench dataset/datamodule/trainer to [measure I/O performance](https://docs.torchgeo.org/en/stable/user/contributing.html#i-o-benchmarking). While this works okay, it has a number of limitations that make it less than fun/easy/intuitive to use. This issue is to track actionable improvements that could be made to I/O Bench.

### Rationale

Current limitations include:
 
1. **Overly verbose output**: Lightning's "simple" profiler output is far from simple. Users are bombarded with a massive table of profiler timings for every hook and helper function. Users then have to figure out which functions correspond to I/O vs. other internal methods.
2. **No rate**: Currently, the profiler output only contains number of calls and total time. While total time is important, it isn't fully reflective of sampling rate due to changes in epoch length. I would really like to see both sampling rate and total time.
3. **Double the fun**: Currently, in order to measure I/O performance for a particular branch, you need to run the script twice, once for raw data and preprocessed data. Ideally we could compute both with one command.
4. **What tests?**: Raw data sampling broke in #2747 and no one noticed FOR AN ENTIRE YEAR! Our current tests could clearly use an improvement.
5. **Git requirement**: Ideally the script would be useful even if torchgeo is pip installed and not git cloned. We should figure out how to make this work.

### Implementation

Suggested solutions include:
 
1. **Less verbose output**: Lightning allows you to create [custom profilers](https://lightning.ai/docs/pytorch/1.6.3/advanced/profiler.html#custom-profiling). We could create our own profiler that directly prints a copy-n-pastable Markdown table that could be added to issues/PRs.
2. **Rate and time**: With a custom profiler, we could easily report rate, time, or both.
3. **Once is enough**: I don't have great ideas to solve this problem, but it's important if we want to be able to generate a single table in one go.
4. **Testing FTW**: We should either add more fake data (raw and preprocessed) or add integration tests that ensure that dataset + sampler + data loader actually works with raw data requiring reprojection. The actual bug will be fixed in #3552.
5. **Batteries included**: We should see if the CLI supports a URL to a YAML file. If not, perhaps JSON would work better.

### Alternatives

We could go back to a custom script (#88, #115, #3675) but this has a number of limitations. The main thing I care about is having a single consistent interface for I/O, sampler, and model benchmarking.

### Additional information

[...]

Issue #3376, opened 2026-02-04
[FR] Log confusion matrix


robmarkcole:
### Summary

To complement metrics. Use torchmetrics implementation? How to handle many classes?

adamjstewart:
My suggestion was to plot a confusion matrix via tensorboard instead of logging all N x N values.

DimitrisMantas:
I think perhaps implementing a plotting callback would be nice to do first and it would also help remove a bunch of boilerplate from the trainers...

adamjstewart:
I actually love this idea, and I think it would also help a lot with @robmarkcole's other request: #2138. @DimitrisMantas do you want to try to take a stab at this? We would add a new `torchgeo/callbacks` directory similar to @hkristen's #3214.

DimitrisMantas:
Sure, I can have a got at it. I'm just not sure what the scope of such a class should be. Supporting all possible loggers and plotable objects seems like too big a step right now.

Perhaps we could start with a metric plotter that accepts or assumes certain trainer fields, like \<stage\>\_confmat, and plots them on on TensorBoard at the right time. This way, the metric mixin can handle updates, and we won't have to change the on\_\<stage\>\_step return signatures, since if we were to plot predictions, like the segmentation trainer does, we would have to return whatever the callback would need to access.

adamjstewart:
> I'm just not sure what the scope of such a class should be. Supporting all possible loggers and plotable objects seems like too big a step right now.

Maybe a base class for most shared code and a subclass for each logger? Just to be clear, I'm not asking you to implement a subclass for all loggers, just one or two that you care about (preferably TensorBoard to maintain easy backwards-compatibility).

> Perhaps we could start with a metric plotter that accepts or assumes certain trainer fields, like `<stage>_confmat`, and plots them on on TensorBoard at the right time. This way, the metric mixin can handle updates, and we won't have to change the `on_<stage>_step` return signatures, since if we were to plot predictions, like the segmentation trainer does, we would have to return whatever the callback would need to access.

Oof, I did not think about this. I don't want to do anything too custom. What do the Lightning docs suggest for callbacks? Should they have access to arbitrary instance attributes, or is it better to return all required values? For training, we must return loss, but for val (where we currently do plotting), we aren't returning anything, so I'm fine with changing this to whatever we need (batch containing images, ground truth, and predictions).

DimitrisMantas:
> Maybe a base class for most shared code and a subclass for each logger?

[...]

Issue #2138, opened 2024-06-27
Support logging figures to multiple loggers


robmarkcole:
### Summary

Currently logging of figures (e.g. in Classification and Segmentation trainers) assumes Tensorboard logger which has the method `add_figure`. However other loggers have similar methods but with different names and are currently not supported. e.g. MLflow uses [log_figure](https://mlflow.org/docs/latest/python_api/mlflow.html#mlflow.log_figure). This FR is to support these loggers and document which are supported

### Rationale

Many orgs use a service for logging such as mlflow or wandb etc, and it would be nice for these to just work.

### Implementation

Use of try: except for known loggers?

### Alternatives

None

adamjstewart:
Another option would be to remove any figure logging from TorchGeo entirely. If users want it, they can subclass the trainer and add it. This feature has historically been a nightmare to maintain/test and has been riddled with bugs. The proposal here is to make it even more complicated.

More positively, can you start a discussion with the Lightning folks to see if there is an easier way to do this automatically? Or a way that can unify the interface somehow? This just seems like too much work to support out of the box.

robmarkcole:
RE lightning I see a lot of work was done in https://github.com/Lightning-AI/pytorch-lightning/pull/6227 but ultimately they decided this was too complex to maintain, and closed the PR.

Perhaps as you say, logging figures should be left to users to subclass the trainers. I think basic logging of samples on the dataset is a necessity however

isaaccorley:
I've actually been moving away from adding the plotting directly inside the LightningModule and using a custom Callback instead. There are methods like `on_validation_batch_end` which have access to the current `batch`, `module`, `batch_idx`, etc. which can essentially replicate what the current figure logging was doing,

robmarkcole:
I agree that using the callbacks will be the most intuitive approach. Do we think it's best to utilise them or just have a 'best practice' guide perhaps?

isaaccorley:
Probably better to have in a tutorial or just reference the Lightning docs so we don't have to maintain them.

robmarkcole:
Been reading on Rastervision, perhaps we can utilise its plotting functionality with lightning callbacks - a simple tutorial could suffice https://docs.rastervision.io/en/stable/usage/tutorials/visualize_data_samples.html

adamjstewart:
[...]

PR #3705, opened 2026-05-15
Custom IOBenchProfiler

evgeniialappo:
This PR creates a custom profiler for IO Benchmarking to simplify the output and calculate sample rate. It is related to Issue #3684. 

### Output example:

Profile Summary

| Action              |      Time (sec) |   Rate (patches/sec) |
| ------------------- | --------------: | -------------------: |
| Training (random)   |        10.19650 |             97.28829 |
| Validation (grid)   |         5.19794 |            190.84471 |

Total duration: 16.24629 seconds

## AI Usage Disclosure

### DIFF
--- a/torchgeo/profilers/io.py
+++ b/torchgeo/profilers/io.py
@@ -50,7 +50,10 @@ class IOBenchProfiler(Profiler):
                 total_time = self._action_total[action_name]
                 count = self._action_count[action_name]
                 total_patches = count * self.batch_size
-                rate = total_patches / total_time
+                if total_time == 0:
+                    rate = 0
+                else:
+                    rate = total_patches / total_time
                 label = 'Training (random)' if train else 'Validation (grid)'
                 res += f'| {label:<19} | {total_time:>15.5f} | {rate:>20.5f} |\n'
         res += f'\nTotal duration: {duration:.5f} seconds\n'
--- a/torchgeo/profilers/io.py
+++ b/torchgeo/profilers/io.py
@@ -9,7 +9,10 @@ from lightning.pytorch.profilers import Profiler
 
 
 class IOBenchProfiler(Profiler):
-    """A custom profiler for I/O benchmark."""
+    """A custom profiler for I/O benchmark.
+
+    .. versionadded:: 0.10
+    """
 
     def __init__(
         self,
--- a/torchgeo/profilers/io.py
+++ b/torchgeo/profilers/io.py
@@ -1,0 +1,64 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+"""Profiler for I/O benchmark."""
+
+import time
+from collections import defaultdict
+
+from lightning.pytorch.profilers import Profiler
+
+
+class IOBenchProfiler(Profiler):
+    """A custom profiler for I/O benchmark."""
+
+    def __init__(
+        self,
+        dirpath: str | None = None,
+        filename: str | None = None,
+        batch_size: int = 32,
+    ) -> None:
+        """Initialise the profiler."""
+        super().__init__(dirpath=dirpath, filename=filename)
+        self._start = time.monotonic()
+        self._action_count: defaultdict[str, int] = defaultdict(int)
+        self._action_start: dict[str, float] = {}
+        self._action_total: defaultdict[str, float] = defaultdict(float)
+        self.batch_size = batch_size
+
+    def start(self, action_name: str) -> None:
+        """Start measuring timing for actions."""
+        self._action_start[action_name] = time.monotonic()
+
+    def stop(self, action_name: str) -> None:
+        """Stop measuring timing for actions."""
+        end_time = time.monotonic()
+        start_time = self._action_start.get(action_name, end_time)
+        self._action_start.pop(action_name, None)
+        self._action_total[action_name] += end_time - start_time
+        self._action_count[action_name] += 1
+
+    def summary(self) -> str:
+        """Print summary of the measurements as a table."""
+        duration = time.monotonic() - self._start
+        res = '\nProfile Summary\n\n'
+        res += f'| {"Action":<19} | {"Time (sec)":>15} | {"Rate (patches/sec)":>20} |\n'
+        res += f'| {"-" * 19} | {"-" * 14}: | {"-" * 19}: |\n'
+        for action_name in self._action_count:
+            train = 'train_dataloader_next' in action_name
+            val = 'val_next' in action_name
+            if train or val:
+                total_time = self._action_total[action_name]
+                count = self._action_count[action_name]
+                total_patches = count * self.batch_size
+                rate = total_patches / total_time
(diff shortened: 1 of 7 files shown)


calebrob6:
I do not like the idea of a `torchgeo.profiles` as this is completely unrelated to geospatial ml and a side effect of us wanting to benchmark using pytorch lightning. In my opinion this should be an upstream contribution and/or put in a project specific repository.

adamjstewart:
For the record, there are also a couple proposed issues or in-progress PRs to add a new `torchgeo.callbacks` namespace for tensorboard/wand/etc. plotting (#2138, #3376, not really GeoML-specific) and stitching (#3214, #3666, #3667, more GeoML-specific). So we will likely add more namespaces regardless of this specific PR.

I agree that a plotting callback and an I/O profiler are not really _GeoML_-specific. I guess I would argue that they are _TorchGeo_-specific though. Plotting depends on details of what the step returns, that the dataset has a plot method, that the datamodule lists mean/std, etc. I/O is a bit more generic, but tying train to random and test to grid is definitely TorchGeo-specific. I think that was also one of your concerns (how am I supposed to know that train_epoch corresponds to RandomPatchSampler).

So if your concern is maintaining more stuff that isn't absolutely required, I'm happy to maintain this namespace. I'm also planning on splitting the API docs into separate sections for the PyTorch API and the Lightning API so that's more clear.

But if your concern is avoiding adding more code that isn't GeoML-specific, we can definitely try to upstream this first and see what the Lightning folks think. They haven't been too responsive to PRs lately though...

calebrob6:
Yeah sorry let me be a little more specific -- this seems like it makes it easier to parse the output of our existing IOBench, that's a :+1:. Bigger picture, I'm thinking about this in the following way:
- If lightning's Profiler changes, then this code breaks. This increases the surface area of features in torchgeo that directly depend on an external library being stable (this seems to historically not be the case; maybe you could simulate this code over the past year or two of releases and see if it would have been historically stable and use that to compute some sort of network maintainability burden). If we are concerned about maintainability, then things like this should be weighed.
- The tests don't verify that it is working correctly, just that it runs
- The whole reason why we're adding another 100 lines of code is so that the existing complicated way of benchmarking becomes _less_ complicated to view. Maybe tangential, but is it possible to run the IOBench routine as a method in a larger script and not have to copy/paste each result individually? What if I want to benchmark each datamodule?
- As a user new to torchgeo, what would I expect in `torchgeo.profilers`? Maybe `torchgeo._profilers` is more appropriate until there's something we actually want to expose to users?
- All of the following could be replaced by a smaller standalone script that doesn't depend on lightning, could be run within other code, etc. (or if you are married to the `torchgeo verb` syntax could be integrated into that somehow)

adamjstewart:
> maybe you could simulate this code over the past year or two of releases and see if it would have been historically stable

It passes our minimum tests, which go back to lightning 2.4 (Aug 2024). We could test it farther back if you want, but I don't expect that this part of lightning is changed very often.

> The tests don't verify that it is working correctly, just that it runs

How would you define "working correctly"? It's basically impossible to assert some specific timings, although we could probably assert that time > 0.

> As a user new to torchgeo, what would I expect in `torchgeo.profilers`? Maybe `torchgeo._profilers` is more appropriate until there's something we actually want to expose to users?

I would expect similar features as Lightning's public profilers. In this case, users may reasonably want to play around with both Lightning's simple profiler and our I/O profiler to optimize their workflows (e.g., play around with COG vs. Zarr and see how timing changes). Note that nothing here is specific to the I/O Bench dataset, it is for I/O benchmarking in general.

I would be fine putting this in `torchgeo/_profiler/io.py` or even in `torchgeo/trainers/iobench.py` if we don't yet want to make it public though.

> All of the following could be replaced by a smaller standalone script that doesn't depend on lightning, could be run within other code, etc.

By the time we add half of the features of LightningCLI and get it working with the `torchgeo` CLI with the same syntax and make it installable, I would argue that we would end up with significantly more code. Part of the goal of LightningCLI was to unify the existing scripts. One more I/O script doesn't seem like much, but several more I/O, training, testing, and inference scripts add up very quickly.

FYI, this is also true for torchgeo-bench. I kinda wish we had extended our LightningCLI instead of inventing yet another CLI. Half of the harness and all of the CLI built there could be deleted if we just used Lightning. Reuse is very important for maintainability, and following existing interfaces is very important for user adoption.

evgeniialappo:
@calebrob6, I've updated the PR. Could you please take another look? It is important for my master's thesis. Thank you!

calebrob6:
Can you post some example output from running it on a few datasets

evgeniialappo:
Sure.
**Chesapeake CVPR Delaware split (batch size 64):** 
Profile Summary

| Action              |      Time (sec) |   Rate (patches/sec) |
| ------------------- | --------------: | -------------------: |
| Training (random)   |       206.35636 |             34.42588 |
| Validation (grid)   |        11.94270 |             48.23028 |

Total duration: 222.46993 seconds

**I/O raw (batch size 32):**

Profile Summary

| Action              |      Time (sec) |   Rate (patches/sec) |
| ------------------- | --------------: | -------------------: |
| Training (random)   |        20.76752 |             40.06256 |
| Validation (grid)   |         8.63288 |            107.49601 |

Total duration: 30.16834 seconds

**I/O preprocessed (batch size 32):**

Profile Summary

| Action              |      Time (sec) |   Rate (patches/sec) |
| ------------------- | --------------: | -------------------: |
| Training (random)   |         8.53098 |            108.78003 |
| Validation (grid)   |         5.33139 |            186.06768 |

Total duration: 14.53605 seconds

**Eurosat100 (batch size 1):**

Profile Summary

| Action              |      Time (sec) |   Rate (patches/sec) |
| ------------------- | --------------: | -------------------: |
| Training (random)   |         0.15056 |            398.51168 |
| Validation (grid)   |         0.02545 |            785.70149 |

Total duration: 0.39800 seconds

torchgeo/profilers/io.py
```diff
@@ -0,0 +1,67 @@
...
+"""Profiler for I/O benchmark."""
+
+import time
+from collections import defaultdict
+
+from lightning.pytorch.profilers import Profiler
+
+
+class IOBenchProfiler(Profiler):
+    """A custom profiler for I/O benchmark."""
```
calebrob6: Needs version added tag

calebrob6 (commented):
Output looks nice!

Need to consider with others whether this should be a top-level thing (torchgeo.profilers), hidden (torchgeo._profilers?), other thing (torchgeo.trainers.profilers). @adamjstewart @isaaccorley @nilsleh @ashnair1 @robmarkcole any opinions?

Needs to be added to docs.

adamjstewart:
My personal opinion is torchgeo.profilers, and I'm willing to maintain it. We'll also hopefully add torchgeo.callbacks soon too.

calebrob6:
Cool, go for it!

adamjstewart:
Superseded by #3861

PR #3715, opened 2026-05-19
Add utae

### LINKED ISSUE #3714: [FR] UTAE time-series model
### Summary

Src: https://github.com/VSainteuf/utae-paps/tree/main

Issue: MIT licensed

CC @VSainteuf

### Rationale

_No response_

### Implementation

_No response_

### Alternatives

_No response_

### Additional information

_No response_

robmarkcole:
Close https://github.com/torchgeo/torchgeo/issues/3714

The src: https://github.com/VSainteuf/utae-paps

## AI Usage Disclosure


calebrob6:
Did you use https://github.com/VSainteuf/utae-paps as a reference?

calebrob6 (commented):
Have you trained this on any tasks? Relatedly, I'm wondering if we can add some sanity checks on all the models we have -- e.g. if I use torchgeo olmoearth, does it replicate the output from the actual olmoearth package?

robmarkcole:
@calebrob6 yes from there with some changes:

- TorchGeo adds project headers, __all__, type hints, Google-style docstrings, local imports, tuple defaults instead of mutable list defaults, and Ruff-style formatting.
- TorchGeo keeps the same core UTAE architecture and forward flow: input conv, down blocks, LTAE2d bottleneck, temporal skip aggregation, up blocks, final conv.
- TorchGeo removes upstream RecUNet entirely, plus the ConvLSTM / BConvLSTM imports.
- TorchGeo renames Temporal_Aggregator to TemporalAggregator and moves it earlier in the file.
- TorchGeo removes unused self.enc_dim and self.stack_dim

I've not tested it yet, wiring to the trainer would be a separate PR - but could be handled in this one. 

Ideally we could get a review from @VSainteuf as I don't appreciate the implications of the changes copilot is suggested

RE consistency checks with other codebases, it sounds like a good idea but also potentially complicated where there are multiple versions of a model, e.g. Olmo v1, v1_1

torchgeo/models/utae.py
```diff
@@ -0,0 +1,519 @@
...
+                assert self.out_shape is not None
+                temp = (
+                    torch.ones(self.out_shape, device=x.device, requires_grad=False)
+                    * self.pad_value
+                )
+                temp[~pad_mask] = self.forward(out[~pad_mask])
+                out = temp
+            else:
+                out = self.forward(out)
+
```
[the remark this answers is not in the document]
robmarkcole: TemporallySharedBlock.smart_forward now:

- handles 4-D inputs by directly calling self.forward
- correctly calls self.forward for 5-D inputs when pad_value is None

torchgeo/models/utae.py
```diff
@@ -0,0 +1,519 @@
...
+                    kernel_size=k,
+                    padding=p,
+                    stride=s,
+                    padding_mode=padding_mode,
+                )
+            )
+            if nl is not None:
+                layers.append(nl(nkernels[i + 1]))
+            if last_relu:
+                layers.append(nn.ReLU())
```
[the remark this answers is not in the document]
robmarkcole: ConvLayer.last_relu now adds ReLU after intermediate convolutions and omits only the final ReLU when last_relu=False, matching the docstring and upstream behavior.

torchgeo/models/utae.py
```diff
@@ -0,0 +1,519 @@
...
+                assert attn_mask is not None
+                attn = attn_mask.mean(dim=0)
+                attn = nn.Upsample(
+                    size=x.shape[-2:], mode='bilinear', align_corners=False
+                )(attn)
+                attn = attn * (~pad_mask).float()[:, :, None, None]
+                return (x * attn[:, :, None, :, :]).sum(dim=1)
+            else:
+                out = x * (~pad_mask).float()[:, :, None, None, None]
+                return out.sum(dim=1) / (~pad_mask).sum(dim=1)[:, None, None, None]
```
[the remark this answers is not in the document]
robmarkcole: Fixed the masked mean aggregation divide-by-zero - The valid timestep count is now clamped to at least 1, so an all-padded sequence returns zeros instead of NaN/inf.

torchgeo/models/utae.py
```diff
@@ -0,0 +1,519 @@
...
+                assert attn_mask is not None
+                n_heads, b, t, h, w = attn_mask.shape
+                attn = attn_mask.view(n_heads * b, t, h, w)
+                if x.shape[-2] > w:
+                    attn = nn.Upsample(
+                        size=x.shape[-2:], mode='bilinear', align_corners=False
+                    )(attn)
+                attn = attn.view(n_heads, b, t, *x.shape[-2:])
+                out = torch.stack(x.chunk(n_heads, dim=2))
+                out = attn[:, :, :, None, :, :] * out
```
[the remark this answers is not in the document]
robmarkcole: Added missing explicit divisibility check in the no-padding att_group path. Misconfigured C % n_heads != 0 now raises the same clear ValueError before x.chunk(...).

torchgeo/models/utae.py
```diff
@@ -0,0 +1,519 @@
...
+            decoder_widths: Number of channels at each decoder stage. If
+                ``None``, mirrors the encoder widths.
+            out_conv: Channel widths for the final output convolutions.
+            str_conv_k: Kernel size for strided up/down convolutions.
+            str_conv_s: Stride for strided up/down convolutions.
+            str_conv_p: Padding for strided up/down convolutions.
+            agg_mode: Skip-connection aggregation mode. One of
+                ``'att_group'``, ``'att_mean'``, or ``'mean'``.
+            encoder_norm: Normalisation layer for the encoder. One of
+                ``'group'``, ``'batch'``, or ``'instance'``.
```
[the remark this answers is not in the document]
robmarkcole: encoder_norm now documents 'none' as supported.

torchgeo/models/ltae.py
```diff
@@ -160,3 +163,289 @@ def forward(self, x: Tensor) -> Tensor:
...
+        if positional_encoding:
+            self.positional_encoder = _PositionalEncoder(
+                self.d_model // n_head, T=T, repeat=n_head
+            )
+
+        self.attention_heads = _MultiHeadAttention(
+            n_head=n_head, d_k=d_k, d_in=self.d_model
+        )
+        self.in_norm = nn.GroupNorm(num_groups=n_head, num_channels=self.in_channels)
+        self.out_norm = nn.GroupNorm(num_groups=n_head, num_channels=n_neurons[-1])
```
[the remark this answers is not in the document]
robmarkcole: Added constructor-time validation in LTAE2d:
n_head must be positive
in_channels % n_head == 0
effective d_model % n_head == 0
mlp[-1] % n_head == 0

torchgeo/models/ltae.py
```diff
@@ -160,3 +163,289 @@ def forward(self, x: Tensor) -> Tensor:
...
+        batch_positions: Tensor | None = None,
+        pad_mask: Tensor | None = None,
+    ) -> Tensor | tuple[Tensor, Tensor]:
+        """Forward pass.
+
+        Args:
+            x: Input tensor of shape ``(B, T, C, H, W)``.
+            batch_positions: Acquisition dates of shape ``(B, T)`` used for
+                positional encoding. Required when ``positional_encoding=True``.
+            pad_mask: Boolean padding mask of shape ``(B, T)`` where ``True``
```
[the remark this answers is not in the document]
robmarkcole: Updated the LTAE2d.forward docstring to match current behavior: batch_positions is optional, and positional encoding is applied only when positions are provided and positional encoding is enabled.

torchgeo/models/utae.py
```diff
@@ -0,0 +1,519 @@
...
+        self.decoder_widths = (
+            list(decoder_widths) if decoder_widths is not None else list(encoder_widths)
+        )
+        self.pad_value = pad_value
+        self.encoder = encoder
+        if encoder:
+            self.return_maps = True
+
+        assert len(self.encoder_widths) == len(self.decoder_widths)
+        assert self.encoder_widths[-1] == self.decoder_widths[-1]
```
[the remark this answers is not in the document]
robmarkcole: Replace the bare asserts in torchgeo/models/utae.py with explicit ValueErrors for:
mismatched encoder_widths / decoder_widths lengths
mismatched final bottleneck widths

robmarkcole:
Results on Pastis100 after 10 epochs, looks better than the ConvLSTM result @calebrob6 

```python
datamodule = PASTIS100DataModule(
    root=root,
    batch_size=4,
    num_workers=7,
    val_split_pct=0.2,
    test_split_pct=0.2,
    padding_length=61,
    download=True,
)

task = SpatioTemporalSegmentationTask(
    model='utae',
    in_channels=10,
    task='multiclass',
    num_classes=20,
    labels=list(PASTIS100.classes),
    loss='ce',
    encoder_widths=(32, 32, 64, 128),
    decoder_widths=(32, 32, 64, 128),
    n_head=8,
    d_model=128,
    d_k=8,
    lr=1e-3,
)
```

image

image

calebrob6 (dismissed):
Sick

tests/trainers/test_spatiotemporal_segmentation.py
```diff
@@ -12,6 +13,32 @@
...
+            ),
+        ]
+    )
+    def model_config(
+        self, request: pytest.FixtureRequest
+    ) -> tuple[Literal['convlstm', 'utae'], dict[str, Any]]:
+        """Return spatiotemporal segmentation model configs."""
+        return cast(tuple[Literal['convlstm', 'utae'], dict[str, Any]], request.param)
+
     @pytest.mark.parametrize('name', ['pastis', 'pastis_focal', 'pastis_jaccard'])
```
adamjstewart: If we change one line in `pastis.yaml` or `pastis_focal.yaml`, we can delete all other changes to this file and keep the tests simple.

torchgeo/models/ltae.py
```diff
@@ -7,11 +7,14 @@
 
 import math
 from collections.abc import Sequence
+from typing import cast
 
 import torch
 import torch.nn as nn
 from torch import Tensor
 
+__all__ = ['LTAE', 'LTAE2d']
```
adamjstewart: Why?

torchgeo/models/ltae.py
```diff
@@ -160,3 +163,299 @@ def forward(self, x: Tensor) -> Tensor:
         pe = self.pe[:, : x.size(1)]  # ty: ignore[not-subscriptable]
         output: Tensor = self.dropout(x + pe)
         return output
+
+
+class _PositionalEncoder(nn.Module):
```
adamjstewart: Quite confusing that we have PositionalEncoding and _PositionalEncoder classes, can we rename these to clearly differentiate between 1D and 2D?

torchgeo/models/ltae.py
```diff
@@ -160,3 +163,299 @@ def forward(self, x: Tensor) -> Tensor:
...
+        """
+        attn = torch.matmul(q.unsqueeze(1), k.transpose(1, 2)) / self.temperature
+        if pad_mask is not None:
+            attn = attn.masked_fill(pad_mask.unsqueeze(1), -1e3)
+        attn = self.dropout(self.softmax(attn))
+        output = torch.matmul(attn, v)
+        return output, attn
+
+
+class _MultiHeadAttention(nn.Module):
```
adamjstewart: Isn't this built into PyTorch now?

torchgeo/models/utae.py
```diff
@@ -0,0 +1,577 @@
...
+from collections.abc import Callable, Sequence
+from typing import Literal, cast
+
+import torch
+import torch.nn as nn
+from torch import Tensor
+
+from .ltae import LTAE2d
+
+__all__ = ['UTAE']
```
adamjstewart: Why?

torchgeo/models/utae.py
```diff
@@ -0,0 +1,577 @@
...
+        self.decoder_widths = (
+            list(decoder_widths) if decoder_widths is not None else list(encoder_widths)
+        )
+        self.pad_value = pad_value
+        self.encoder = encoder
+        if encoder:
+            self.return_maps = True
+
+        if len(self.encoder_widths) != len(self.decoder_widths):
+            raise ValueError(
```
adamjstewart: Should we document these Raises?

torchgeo/models/utae.py
```diff
@@ -0,0 +1,577 @@
...
+                else None,
+            )
+            for module in batch_norms
+        ]
+
+        try:
+            with torch.no_grad():
+                return self.forward(x).shape
+        finally:
+            for (
```
adamjstewart: What in the world is going on here? Some comments or simplification would be greatly appreciated.

torchgeo/models/utae.py
```diff
@@ -0,0 +1,577 @@
...
+                module.running_mean.clone()
+                if module.running_mean is not None
+                else None,
+                module.running_var.clone() if module.running_var is not None else None,
+                module.num_batches_tracked.clone()
+                if module.num_batches_tracked is not None
+                else None,
+            )
+            for module in batch_norms
+        ]
```
adamjstewart: Cool that this can technically be done on one line, but the result is that it's significantly more lines and impossible to understand.

torchgeo/models/utae.py
```diff
@@ -0,0 +1,577 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+
+# Copyright (c) 2021 VSainteuf (Vivien Sainte Fare Garnot)
```
adamjstewart, line 4: Where can I find the reference implementation this came from? Would be good to document this for both models.

torchgeo/models/utae.py
```diff
@@ -0,0 +1,577 @@
...
+            x: Feature maps of shape ``(B, T, C, H, W)``.
+            pad_mask: Boolean mask of shape ``(B, T)``.
+            attn_mask: Attention weights of shape
+                ``(n_head, B, T, H_att, W_att)`` for ``'att_group'`` /
+                ``'att_mean'`` modes.
+
+        Returns:
+            Aggregated feature map of shape ``(B, C, H, W)``.
+        """
+        if pad_mask is not None and pad_mask.any():
```
adamjstewart, line 242: Ton of complex code here, don't understand any of it, would be better with comments.

torchgeo/models/ltae.py
```diff
@@ -160,3 +163,299 @@ def forward(self, x: Tensor) -> Tensor:
...
+            positional_encoding: If False, no positional encoding is applied.
+        """
+        super().__init__()
+        self.in_channels = in_channels
+        self.return_att = return_att
+        self.n_head = n_head
+        self.d_model = d_model if d_model is not None else in_channels
+
+        if n_head <= 0:
+            raise ValueError('n_head must be positive')
```
adamjstewart: Need to document Raises

adamjstewart (changes requested):
Missing a ton of stuff from https://docs.torchgeo.org/en/latest/user/contributing.html#model-architectures

Very complex, hard to review, no idea whether the implementation is identical to the original or if AI made changes for fun.

Would also like @isaaccorley to review

torchgeo/models/utae.py
robmarkcole: From https://github.com/VSainteuf/utae-paps/blob/987874e27a98bb399277f2d55eb010744d26f881/src/backbones/utae.py#L361

torchgeo/models/utae.py
robmarkcole: see top of pr

adamjstewart: Would be good to document this in the code too.

torchgeo/models/utae.py
adamjstewart: @VSainteuf may be able to further elaborate?

torchgeo/models/utae.py
robmarkcole: simplified and added comment

torchgeo/models/utae.py
robmarkcole: Ai overreach to support efficient `from torchgeo.models.utae import *` but removed

torchgeo/models/ltae.py
robmarkcole: done

torchgeo/models/utae.py
robmarkcole: done

torchgeo/models/ltae.py
robmarkcole: this implementation is customised to support a learned master query - it is **not** the same as nn.MultiheadAttention

torchgeo/models/ltae.py
robmarkcole: Done. PositionalEncoding → _IndexPositionalEncoding (used in LTAE, encodes by sequence index), _PositionalEncoder → _DatePositionalEncoding (used in LTAE2d, encodes by actual acquisition date).

tests/trainers/test_spatiotemporal_segmentation.py
robmarkcole: test in pastis_focal.yaml

robmarkcole:
Vs https://github.com/VSainteuf/utae-paps/blob/main/src/backbones/utae.py implementation: 

**One notable omission:** RecUNet (the recurrent alternative architecture from the original) was not ported over


Component | Change
-- | --
TemporalAggregator | Removed AvgPool2d downsampling; added .clamp(min=1) for safe division; refactored to remove duplicated pad_mask/no-pad branches; extracted _check_att_group_channels()
TemporallySharedBlock.smart_forward | Full rewrite — original used a dummy zeros forward to cache out_shape on every call (corrupting BatchNorm); TorchGeo only runs the dummy forward in the all-padding edge case with BN state save/restore
UTAE | assert → ValueError for width mismatches

Everything else (type hints, docstrings, renaming Temporal_Aggregator → TemporalAggregator, snake_case cleanup) is cosmetic.

robmarkcole:
@adamjstewart same read the docs error as https://github.com/torchgeo/torchgeo/pull/3591 dont get an error locally

adamjstewart:
RtD is down for me so can't help investigate that. Focus on the unit tests and hopefully that will fix the RtD tests too?

torchgeo/models/utae.py
```diff
@@ -0,0 +1,568 @@
...
+
+from .ltae import LTAE2d
+
+
+class UTAE(nn.Module):
+    """U-Net with Lightweight Temporal Attention Encoder (U-TAE).
+
+    Spatio-temporal encoder for satellite image time series. A shared L-TAE
+    is applied at the U-Net bottleneck; per-head attention masks are
+    propagated to the skip connections via a :class:`TemporalAggregator`.
```
adamjstewart: Okay, now I see it. Relevant error is:
```
/home/docs/checkouts/readthedocs.org/user_builds/torchgeo/checkouts/3715/torchgeo/models/utae.py:docstring of torchgeo.models.utae.UTAE:3: WARNING: py:class reference target not found: TemporalAggregator [ref.class]
```
You reference TemporalAggregator here in the docs, but this class isn't documented, so Sphinx doesn't know what to do. Either remove the cross reference here:
```suggestion
    propagated to the skip connections via a *TemporalAggregator*.
```
or export it and add it to the documentation if it's important.

torchgeo/models/utae.py
robmarkcole: @adamjstewart is it a blocker if we don't hear back? Results on Pastis100 were very encouraging

adamjstewart: Definitely not a blocker.

adamjstewart:
Will wait for @isaaccorley to review before reviewing again.

torchgeo/models/utae.py
VSainteuf: Hi @robmarkcole @adamjstewart, sorry for the late reply.  

The Temporal_Aggregator class is used to produce the skip connection in the U-Net structure of the U-TAE. It offers different ways to do that: simple temporal mean ("mean"), or a temporally weighted sum using the attention mask averaged across attention heads ("att_mean"), or a temporally weighted sum using the attention masks per channel group ("att_group"). The chosen way in the U-TAE is "att_group" and the other two variants were used in the ablation study. 

Additionally, the first block of the forward method (under the `if pad_mask is not None` condition) makes sure to not use padded input images in the skip connection (in order to constitute a batch of constant shape, some image time series are padded on the temporal dimension to a constant length, and we don't want these in the skip connection).

I hope this helps clarify things, and let me know how I could help.

robmarkcole: @VSainteuf do you think we should expose mean/att_mean/att_group or just hard-code att_group? Not sure if this is a parameter that people will explore in practice

VSainteuf: Good point, I think it does make sense to hard code att_group and leave the rest. 
The other two configs are not just a hyperparameter change but a slightly different architecture which we showed works worst.

robmarkcole: @adamjstewart now hard code `att_group` and improved the docstrings

adamjstewart:
Can you rebase now that we renamed things?

robmarkcole:
@adamjstewart rebased

gatienc:
@robmarkcole It seems in the trained model output that you show that the model is predicting some of the white class, If I'm correct and you use the same cmap as in the original repo, this class is the void class that should be ignored from the training so the model doesn't predict it. 
It is basically parcel for which the class hasn't been given.

gatienc:
Also, I think that for crop segmentation, having the dates in the temporal encoding is important. As the model probably use the crop spectral reflectance over time. 

Pastis dataset currently doesn't provide the acquisition dates. Therefore the Trainer trains with batch_positions to None.

I'm interested in getting that implemented in PASTIS and #3181 (FLAIR-HUB). 
I guess it can probably be done in a subsequent PR once both will be merged

robmarkcole:
@adamjstewart please see comment above about viz of nodata areas - I assume we should standardise on a color for this, i.e. black

@gatienc re support for temporal labels, if FLAIR-HUB adds these I agree we can wait for that contribution then update the trainers/models after. Again @adamjstewart there is intersection with your recent comments about how to handle datetimes

@gatienc also thanks for your comment on the `void` class: `the white class is label 19, the PASTIS void label. This was a configuration oversight rather than a colormap issue. I’ve added ignore_index: 19 to all PASTIS test configurations while retaining 20 output classes, matching the reference implementation. Void pixels are now excluded from the loss and metrics. Predictions on ignored pixels remain unconstrained, but the model will no longer learn void as a supervised semantic class.`

adamjstewart:
Going to let @isaaccorley review this one, this is a bit outside my area of expertise.

torchgeo/models/ltae.py
```diff
@@ -125,7 +126,7 @@ def forward(self, x: Tensor) -> Tensor:
         return output
 
 
-class PositionalEncoding(nn.Module):
+class _IndexPositionalEncoding(nn.Module):
```
isaaccorley: don't make this a private class

torchgeo/models/ltae.py
```diff
@@ -159,3 +160,299 @@ def forward(self, x: Tensor) -> Tensor:
         pe = self.pe[:, : x.size(1)]  # ty: ignore[not-subscriptable]
         output: Tensor = self.dropout(x + pe)
         return output
+
+
+class _DatePositionalEncoding(nn.Module):
```
isaaccorley: don't make this private

torchgeo/models/ltae.py
```diff
@@ -159,3 +160,299 @@ def forward(self, x: Tensor) -> Tensor:
...
+
+        Args:
+            batch_positions: Integer positions of shape ``(B, T)``.
+
+        Returns:
+            Positional encoding of shape ``(B, T, d)`` or ``(B, T, d * repeat)``.
+        """
+        denom = cast(Tensor, self.denom)
+        pe = batch_positions.unsqueeze(-1).float() * denom.unsqueeze(0).unsqueeze(0)
+        pe[..., 0::2] = torch.sin(pe[..., 0::2])
```
isaaccorley: since you're overriding the data directly should just use `torch.sin_` so it's inplace

torchgeo/models/ltae.py
```diff
@@ -159,3 +160,299 @@ def forward(self, x: Tensor) -> Tensor:
...
+        """Encode batch positions.
+
+        Args:
+            batch_positions: Integer positions of shape ``(B, T)``.
+
+        Returns:
+            Positional encoding of shape ``(B, T, d)`` or ``(B, T, d * repeat)``.
+        """
+        denom = cast(Tensor, self.denom)
+        pe = batch_positions.unsqueeze(-1).float() * denom.unsqueeze(0).unsqueeze(0)
```
isaaccorley: can you use einops for this? Too many chained unsqueezes

torchgeo/models/ltae.py
```diff
@@ -159,3 +160,299 @@ def forward(self, x: Tensor) -> Tensor:
...
+        pe[..., 0::2] = torch.sin(pe[..., 0::2])
+        pe[..., 1::2] = torch.cos(pe[..., 1::2])
```
isaaccorley: since you're overriding the data directly should just use `torch.cos_` so it's inplace

torchgeo/models/ltae.py
```diff
@@ -159,3 +160,299 @@ def forward(self, x: Tensor) -> Tensor:
...
+        if self.repeat is not None:
+            pe = pe.repeat(1, 1, self.repeat)
```
isaaccorley: can you use einops for this so its more readable

torchgeo/models/ltae.py
```diff
@@ -159,3 +160,299 @@ def forward(self, x: Tensor) -> Tensor:
...
+        return pe
+
+
+class _ScaledDotProductAttention(nn.Module):
```
isaaccorley: Let's use https://docs.pytorch.org/docs/2.13/generated/torch.nn.functional.scaled_dot_product_attention.html instead because this will be quite slow compared to torch's official implementation

torchgeo/models/ltae.py
```diff
@@ -159,3 +160,299 @@ def forward(self, x: Tensor) -> Tensor:
...
+        """
+        attn = torch.matmul(q.unsqueeze(1), k.transpose(1, 2)) / self.temperature
+        if pad_mask is not None:
+            attn = attn.masked_fill(pad_mask.unsqueeze(1), -1e3)
+        attn = self.dropout(self.softmax(attn))
+        output = torch.matmul(attn, v)
+        return output, attn
+
+
+class _MultiHeadAttention(nn.Module):
```
isaaccorley: I'd prefer that we use torch's MHA rather than roll our own. I saw your comment about a learned query and this is quite easy to do with torch's MHA 

```python
import torch
import torch.nn as nn

class LearnedQueryAttention(nn.Module):
    def __init__(self, num_queries: int, embed_dim: int, num_heads: int, dropout: float = 0.0):
        super().__init__()
        # 1. Initialize standard multihead attention
        self.mha = nn.MultiheadAttention(
            embed_dim=embed_dim, 
            num_heads=num_heads, 
            dropout=dropout, 
            batch_first=True
        )
        
        # 2. Define the learned query as a trainable Parameter
        # Shape: (num_queries, embed_dim)
        self.learned_query = nn.Parameter(torch.empty(num_queries, embed_dim))
        
        # 3. Initialize the parameter weights to prevent exploding/vanishing gradients
        nn.init.normal_(self.learned_query, std=0.02)

    def forward(self, x: torch.Tensor) -> torch.Tensor:
        """
        Args:
            x: Input tensor (Key/Value) of shape (batch_size, seq_len, embed_dim)
        Returns:
            Output tensor of shape (batch_size, num_queries, embed_dim)
        """
        batch_size = x.size(0)
        
        # Expand the learned query to match the batch size
        # Shape changes from (num_queries, embed_dim) -> (batch_size, num_queries, embed_dim)
        q = self.learned_query.unsqueeze(0).expand(batch_size, -1, -1)
        
        # Cross-attend: Query comes from our parameter, Key & Value come from the input
        # mha(query, key, value)
        attn_output, _ = self.mha(query=q, key=x, value=x)
        
        return attn_output
```

torchgeo/models/ltae.py
```diff
@@ -159,3 +160,299 @@ def forward(self, x: Tensor) -> Tensor:
...
+        """
+        super().__init__()
+        self.in_channels = in_channels
+        self.return_att = return_att
+        self.n_head = n_head
+        self.d_model = d_model if d_model is not None else in_channels
+
+        if n_head <= 0:
+            raise ValueError('n_head must be positive')
+        if in_channels % n_head != 0:
```
isaaccorley: change to an assert

torchgeo/models/ltae.py
```diff
@@ -159,3 +160,299 @@ def forward(self, x: Tensor) -> Tensor:
...
+            raise ValueError('in_channels must be divisible by n_head')
+        if self.d_model % n_head != 0:
```
isaaccorley: change to an assert

torchgeo/models/utae.py
```diff
@@ -0,0 +1,528 @@
...
+        self.conv = ConvLayer(
+            nkernels=nkernels, norm=norm, last_relu=last_relu, padding_mode=padding_mode
+        )
+
+    def forward(self, x: Tensor) -> Tensor:
+        """Forward pass."""
+        return self.conv(x)
+
+
+class DownConvBlock(TemporallySharedBlock):
```
isaaccorley, line 427: Can we reuse code from smp? Seems like we are just reimplementing unet here

torchgeo/models/utae.py
```diff
@@ -0,0 +1,528 @@
...
+        )
+
+    def forward(self, x: Tensor) -> Tensor:
+        """Forward pass."""
+        out = self.down(x)
+        out = self.conv1(out)
+        return out + self.conv2(out)
+
+
+class UpConvBlock(nn.Module):
```
isaaccorley, line 478: Can we reuse code from smp? Seems like we are just reimplementing unet here

torchgeo/models/utae.py
```diff
@@ -0,0 +1,528 @@
...
+            if last_relu or i < len(nkernels) - 2:
+                layers.append(nn.ReLU())
+        self.conv = nn.Sequential(*layers)
+
+    def forward(self, x: Tensor) -> Tensor:
+        """Forward pass."""
+        return self.conv(x)
+
+
+class ConvBlock(TemporallySharedBlock):
```
isaaccorley, line 390: Can we reuse code from smp? Seems like we are just reimplementing unet here

torchgeo/models/utae.py
```diff
@@ -0,0 +1,528 @@
...
+        ]
+        try:
+            with torch.no_grad():
+                return self.forward(x).shape
+        finally:
+            for m, state in zip(batch_norms, saved):
+                m.load_state_dict(state)
+
+
+class ConvLayer(nn.Module):
```
isaaccorley, line 322: Can we reuse code from smp? Seems like we are just reimplementing unet here

torchgeo/models/utae.py
```diff
@@ -0,0 +1,528 @@
...
+                else:
+                    out_shape = self._infer_output_shape(out)
+                    out = out.new_full(out_shape, self.pad_value)
+            else:
+                out = self.forward(out)
+
+        _, c_out, h_out, w_out = out.shape
+        return out.view(b, t, c_out, h_out, w_out)
+
+    def _infer_output_shape(self, x: Tensor) -> torch.Size:
```
isaaccorley: "Called only when every frame in a batch is padding" is this actually a real edge case? Seems like overengineering imo

torchgeo/models/utae.py
```diff
@@ -0,0 +1,528 @@
...
+        """Initialize TemporallySharedBlock.
+
+        Args:
+            pad_value: If given, padded frames (all channels equal to this
+                value) are skipped and their outputs are filled with this value.
+        """
+        super().__init__()
+        self.pad_value = pad_value
+
+    def smart_forward(self, x: Tensor) -> Tensor:
```
isaaccorley, line 290: this could be made more readable using einops and dropping some things

```python
from einops import rearrange

def smart_forward(self, x: Tensor) -> Tensor:
    if x.ndim != 5:
        raise ValueError("expected (B, T, C, H, W)")

    b, t = x.shape[:2]

    # (B, T, C, H, W) -> (B*T, C, H, W)
    x = rearrange(x, "b t c h w -> (b t) c h w")

    if self.pad_value is None:
        out = self(x)
    else:
        valid = ~(x == self.pad_value).flatten(1).all(1)

        if not valid.any():
            raise ValueError("batch contains no valid frames")

        valid_out = self(x[valid])

        out = valid_out.new_full(
            (b * t, *valid_out.shape[1:]),
            self.pad_value,
        )
        out[valid] = valid_out

    # (B*T, C, H, W) -> (B, T, C, H, W)
    return rearrange(out, "(b t) c h w -> b t c h w", b=b, t=t)
```

isaaccorley (changes requested):

torchgeo/models/ltae.py
robmarkcole: done

torchgeo/models/ltae.py
robmarkcole: done

torchgeo/models/ltae.py
robmarkcole: Updated both operations to use sin_() and cos_() in place.

torchgeo/models/ltae.py
robmarkcole: done

torchgeo/models/ltae.py
robmarkcole: Replaced the chained unsqueeze calls with explicit einops.rearrange patterns.

torchgeo/models/ltae.py
robmarkcole: replaced with `repeat(pe, 'b t d -> b t (repeat d)', repeat=self.repeat)`

torchgeo/models/ltae.py
robmarkcole: Can’t use scaled_dot_product_attention because U-TAE requires the attention weights for temporal skip-connection aggregation. PyTorch SDPA returns only the attended values, not the weights, so explicit attention computation remains necessary. 

Instead I removed the custom attention class and moved its explicit matrix multiplication, masking, softmax, dropout, and value aggregation directly into _MultiHeadAttention. LTAE2d now always returns both outputs and attention weights.

torchgeo/models/ltae.py
robmarkcole: Replaced the hand-rolled attention with public LearnedQueryAttention, backed by nn.MultiheadAttention.
It uses:

- One learned query expanded across the batch.
- n_head * d_k as the MHA embedding width.
- Per-head attention weights for U-TAE aggregation.
- An output projection back to d_in.

Also fixed non-contiguous MHA weights in TemporalAggregator using einops.rearrange. All 32 LTAE/U-TAE tests pass.

torchgeo/models/utae.py
robmarkcole: Not without changing U-TAE’s architecture:

- SMP upsamples by interpolation; `UpConvBlock` uses transposed convolution.
- U-TAE projects skips and uses residual convolution blocks.
- `ConvBlock` handles 5D temporal inputs and padded frames.
- `ConvLayer` supports reflect padding, several normalization modes, and optional final activation.
- SMP’s decoder cannot directly accommodate attention-weighted temporal skips.

Using SMP would require substantial adapters and break checkpoint compatibility, so retaining these blocks is cleaner.

torchgeo/models/utae.py
robmarkcole: No, as comment above

torchgeo/models/utae.py
robmarkcole: No, as comment above

torchgeo/models/utae.py
robmarkcole: Agreed—it was overengineered. No production path creates an entirely padded batch.
Removed _infer_output_shape and its BatchNorm snapshot logic. Fully padded batches now raise a clear ValueError. U-TAE and task tests pass.

torchgeo/models/utae.py
robmarkcole: no, comment above

torchgeo/models/ltae.py
robmarkcole: done

torchgeo/models/ltae.py
robmarkcole: done

torchgeo/models/utae.py
robmarkcole: Simplified smart_forward using einops.rearrange and flattened validity detection as suggested. Removed the nested padding branches and manual shape handling.

robmarkcole:
Results on the PASTIS 100 after the latest changes, look consistent (additionally using ignore_index @gatienc):

image
image
image
image

torchgeo/models/ltae.py
```diff
@@ -159,3 +161,251 @@ def forward(self, x: Tensor) -> Tensor:
...
+                [
+                    nn.Linear(n_neurons[i], n_neurons[i + 1]),
+                    nn.BatchNorm1d(n_neurons[i + 1]),
+                    nn.ReLU(),
+                ]
+            )
+        self.mlp = nn.Sequential(*layers)
+        self.dropout = nn.Dropout(dropout)
+
+    def forward(
```
isaaccorley, line 354: sorry I must have missed this, can you revise this entire forward method using einops? it's very hard to read

torchgeo/models/utae.py
isaaccorley: Conv2dTranspose has been shown to be worse than Upsample+Conv2d and produces checkerboard artifacts https://distill.pub/2016/deconv-checkerboard/ but I'm okay with this. Just trying to look for ways to reduce reimplementing a conv down or upsample block.

isaaccorley (changes requested):

torchgeo/models/ltae.py
robmarkcole: Rewrote the entire LTAE2d.forward method using named einops.rearrange and repeat operations.

robmarkcole:
@isaaccorley predictions still good after latest changes

image

torchgeo/models/utae.py
adamjstewart: This website looks stupidly good, can't believe this is from 2016.

robmarkcole:
@adamjstewart I've addressed all Isaacs comments, it is ready

isaaccorley (dismissed):

torchgeo/models/utae.py
robmarkcole: added with a note about the differences

robmarkcole:
@isaaccorley there was one outstanding comment I addressed, needs review again, sorry

isaaccorley (dismissed):

docs/api/models/l-tae.rst
```diff
@@ -3,3 +3,7 @@ L-TAE
 
 .. currentmodule:: torchgeo.models
 .. autoclass:: LTAE
+.. autoclass:: LTAE2d
+.. autoclass:: IndexPositionalEncoding
+.. autoclass:: DatePositionalEncoding
+.. autoclass:: LearnedQueryAttention
```
adamjstewart: Do we really want to make these 3 public?

tests/models/test_ltae.py
```diff
@@ -48,3 +48,98 @@ def test_invalid_input(self) -> None:
             model = LTAE(in_channels=in_channels)
             x = torch.randn(batch_size, seq_len, wrong_channels)
             model(x)
+
+
+class TestLTAE2d:
```
adamjstewart, line 53: LTAE tests only required 50 lines, why does LTAE2d require 100 lines?

tests/models/test_utae.py
adamjstewart, line 1: LTAE tests only required 50 lines, why does UTAE require 250 lines?

tests/tasks/test_spatiotemporal_segmentation.py
```diff
@@ -42,7 +43,6 @@ def test_trainer(self, name: str, fast_dev_run: bool) -> None:
         except MisconfigurationException:
             pass
 
-    @pytest.mark.filterwarnings(r'ignore:You are trying to `self.log\(\)`')
```
adamjstewart, line 45: Why remove this, bad rebase? This results in a warning in CI

tests/tasks/test_spatiotemporal_segmentation.py
```diff
@@ -72,3 +72,47 @@ def test_multilabel_predict_step(self) -> None:
         assert probabilities.shape == (2, 4, 16, 16)
         assert torch.all(probabilities >= 0)
         assert torch.all(probabilities <= 1)
+
+    def test_utae_predict_step_forwards_batch_positions(self) -> None:
```
adamjstewart: Could we replace these 50 lines with:
```python
@pytest.mark.parametrize('model', ['convlstm', 'utae'])
```
on line 64?

torchgeo/models/__init__.py
```diff
@@ -24,7 +24,13 @@
 from .farseg import FarSeg
 from .fcn import FCN
 from .fcsiam import FCSiamConc, FCSiamDiff
-from .ltae import LTAE
+from .ltae import (
+    LTAE,
+    DatePositionalEncoding,
+    IndexPositionalEncoding,
+    LearnedQueryAttention,
```
adamjstewart: Let's not make these public in case we change the implementation later

torchgeo/models/ltae.py
```diff
@@ -159,3 +163,254 @@ def forward(self, x: Tensor) -> Tensor:
...
+            T: Period for sinusoidal positional encoding.
+            positional_encoding: If False, no positional encoding is applied.
+        """
+        super().__init__()
+        self.in_channels = in_channels
+        self.n_head = n_head
+        self.d_model = d_model if d_model is not None else in_channels
+
+        if n_head <= 0:
+            raise ValueError('n_head must be positive')
```
adamjstewart:
```suggestion
        assert n_head > 0, 'n_head must be positive'
```
Allows you to remove yet another test

torchgeo/models/ltae.py
```diff
@@ -159,3 +163,254 @@ def forward(self, x: Tensor) -> Tensor:
...
+
+        Returns:
+            Output feature map of shape ``(B, mlp[-1], H, W)`` and attention
+            weights of shape ``(n_head, B, T, H, W)``.
+        """
+        batch_size, _, _, height, width = x.shape
+
+        if pad_mask is not None:
+            pad_mask = repeat(
+                pad_mask, 'b t -> (b height width) t', height=height, width=width
```
adamjstewart: If you use `h` and `w` instead of `height` and `width` most of these expressions could fit on a single line

torchgeo/models/utae.py
```diff
@@ -0,0 +1,510 @@
...
+            self.return_maps = True
+
+        if len(self.encoder_widths) != len(self.decoder_widths):
+            raise ValueError(
+                'encoder_widths and decoder_widths must have the same length'
+            )
+        if self.encoder_widths[-1] != self.decoder_widths[-1]:
+            raise ValueError(
+                'encoder_widths and decoder_widths must have the same final width'
+            )
```
adamjstewart: Could be asserts

torchgeo/models/utae.py
```diff
@@ -0,0 +1,510 @@
...
+        super().__init__()
+        nl: Callable[[int], nn.Module] | None
+        if norm == 'batch':
+            nl = nn.BatchNorm2d
+        elif norm == 'instance':
+            nl = nn.InstanceNorm2d
+        elif norm == 'group':
+
+            def nl(num_feats: int) -> nn.GroupNorm:
+                return nn.GroupNorm(num_channels=num_feats, num_groups=n_groups)
```
adamjstewart: Could use `functools.partial` although this makes type hints trickier

torchgeo/models/utae.py
```diff
@@ -0,0 +1,510 @@
...
+                )
+            )
+            if nl is not None:
+                layers.append(nl(nkernels[i + 1]))
+            if last_relu or i < len(nkernels) - 2:
+                layers.append(nn.ReLU())
+        self.conv = nn.Sequential(*layers)
+
+    def forward(self, x: Tensor) -> Tensor:
+        """Forward pass."""
```
adamjstewart: A lot of functions are missing docs of inputs/outputs

torchgeo/tasks/spatiotemporal_segmentation.py
```diff
@@ -78,24 +83,82 @@ def configure_models(self) -> None:
...
+            'str_conv_p',
+            'encoder_norm',
+            'n_head',
+            'd_model',
+            'd_k',
+            'encoder',
+            'return_maps',
+            'pad_value',
+            'padding_mode',
+        }
```
adamjstewart: Would really like to find a better workaround that doesn't require this hack. How does jsonargparse even know which kwargs are valid?

adamjstewart (dismissed):
I don't really want to maintain another 1.3k lines of code (per day), we need to be more strict about maintainability or TorchGeo will be abandoned by its own maintainers and Claude will be the only maintainer left.

tests/models/test_ltae.py
robmarkcole: there was duplication, reduced LTAE2d tests from ~100 to ~50 lines.

tests/models/test_utae.py
robmarkcole: There was duplication - dropped from 247 to 173 lines; direct UTAE model tests are now ~55 lines, with the remainder covering its reusable building blocks.

tests/tasks/test_spatiotemporal_segmentation.py
robmarkcole: done - 43 lines removed and 15 added.

tests/tasks/test_spatiotemporal_segmentation.py
robmarkcole: yes rebase error, restored

torchgeo/models/ltae.py
robmarkcole: done

torchgeo/models/__init__.py
robmarkcole: done

torchgeo/models/ltae.py
robmarkcole: done

torchgeo/models/utae.py
robmarkcole: done

torchgeo/models/utae.py
robmarkcole: done

torchgeo/models/utae.py
robmarkcole: Documented missing inputs and outputs for:

- UTAE.forward
- TemporalAggregator._check_att_group_channels
- ConvLayer.forward
- ConvBlock.forward
- DownConvBlock.forward

gatienc:
> I don't really want to maintain another 1.3k lines of code (per day), we need to be more strict about maintainability or TorchGeo will be abandoned by its own maintainers and Claude will be the only maintainer left.

Satellite image time series adds a lot of new parameters that are not yet usual for Torchgeo. I think it would be possible to create new classes, types and base methods that could be re-used across multi-temporal datasets/models. That would require some work as most of the parameter names are differents in each implementation and can have a wide variety of formats. However, once implemented it could greatly reduce "cost of entry" as we could have trainers that could work nicely across models and configuration variables that would help navigate between the different input formats.

I had to implement that on my current project as I am benchmarking different SITS models. Let me know if you think we should do this and I could work on a draft that we can discuss.

docs/api/models/l-tae.rst
robmarkcole: removed

torchgeo/tasks/spatiotemporal_segmentation.py
robmarkcole: Addressed, Claudes explanation:

jsonargparse performs static signature analysis on constructor code. It saw `**self.kwargs` passed directly into `ConvLSTM`, followed that call, and inferred ConvLSTM’s accepted keyword arguments and defaults. Consequently, defaults such as `hidden_dim`, `kernel_size`, and `num_layers` were injected even when `model='utae'`, causing UTAE to receive invalid arguments.

Copying the dictionary first creates a boundary that prevents this static inference:

```python
kwargs = self.kwargs.copy()
self.model = ConvLSTM(..., **kwargs)
```

jsonargparse now treats `dict_kwargs` as explicitly supplied arbitrary arguments rather than expanding one model’s constructor signature across every model variant. Each YAML configuration therefore passes only its own settings:

- ConvLSTM receives ConvLSTM arguments.
- Conv3dLSTM receives Conv3dLSTM arguments.
- UTAE receives UTAE arguments.

This removes the need to maintain three overlapping parameter allowlists, which would otherwise need updating whenever any constructor changes. All three LightningCLI configurations pass, as do formatting, linting, and type checks. Net reduction: 35 lines.

gatienc:
To give a further explanation, for example : the positional encoding of date will be seen again in a lot of SITS model. UTAE on pastis is made to be sinusoidal positional encoding on the difference in day between acquisition and the reference date.

However different models don't use exactly the same approach.
It would be possible to define a temporal encoding class that would be injected with the different configs, for example if we want to encode with day of year or difference with a reference date. Choosing the encoding : (Sinusoidal, RoPe ... etc) 

It would indeed be a lot of LoC initially but would simplify the implementation of new datasets and make the total LoC after the implementation of multiple datasets smaller.

robmarkcole:
@gatienc could you create a FR/issue with the outline? Happy to support this effort

gatienc:
> @gatienc could you create a FR/issue with the outline? Happy to support this effort

I opened #3987

robmarkcole:
@adamjstewart what's the verdict on bringing this in? IMO U-TAE is a canonical time-series model, and this implementation is well tested compared with the other public implementations, which makes it a valuable contribution.

adamjstewart:
I'm taking a break from reviewing AI-assisted code, but other maintainers may be willing to review and merge.

robmarkcole:
@isaaccorley previously approved, perhaps @calebrob6 can take a look?

isaaccorley (approved):
I reviewed again and I think this is a sufficient first step towards time-series segmentation and is valuable addition to TorchGeo. I'm happy to assist in maintaining this.

PR #3716, opened 2026-05-19
DLRSD: fix empty-split UserWarnings in trainer tests

calebrob6:
Closes the `UserWarning: Length of split at index N is 0` warnings emitted by the DLRSD trainer tests after #3498 merged.

## Problem

The fake test data in `tests/data/dlrsd/` only contains 4 images. With the datamodule defaults (`val_split_pct=0.1`, `test_split_pct=0.2`) and the no-val variant (`val_split_pct=0`, `test_split_pct=0.2`), `torch.utils.data.random_split` floors fractions and ends up with an empty split:

```
[0.7, 0.1, 0.2] * 4  ->  [3, 1, 0]
[0.8, 0.2]      * 4  ->  [4, 0]
```

That triggers Lightning/PyTorch's `UserWarning: Length of split at index N is 0`, visible in CI runs for:

- `tests/trainers/test_segmentation.py::TestSemanticSegmentationTask::test_trainer[True-dlrsd]`
- `tests/trainers/test_segmentation.py::TestSemanticSegmentationTask::test_trainer[True-dlrsd_no_val]`
- `tests/trainers/test_classification.py::TestClassificationTask::test_trainer[True-dlrsd_multilabel]`
- `tests/trainers/test_classification.py::TestClassificationTask::test_trainer[True-dlrsd_multilabel_no_val]`

## Fix

Override `val_split_pct` / `test_split_pct` in the four trainer conf YAMLs (matching the `spacenet1.yaml` 0.34/0.34 precedent for 3-way splits with small fake data) so every split has at least one sample:

```
[0.32, 0.34, 0.34] * 4  ->  [2, 1, 1]
[0.66, 0.34]       * 4  ->  [3, 1]
```

Verified locally by re-running the datamodule `setup('fit')` under `python -W error::UserWarning` — no warnings emitted, all three splits non-empty.

## AI Usage Disclosure

### DIFF
--- a/tests/conf/dlrsd.yaml
+++ b/tests/conf/dlrsd.yaml
@@ -10,5 +10,7 @@ data:
   class_path: DLRSDDataModule
   init_args:
     batch_size: 1
+    val_split_pct: 0.34
+    test_split_pct: 0.34
   dict_kwargs:
     root: 'tests/data/dlrsd'
--- a/tests/conf/dlrsd.yaml
+++ b/tests/conf/dlrsd.yaml
@@ -10,7 +10,7 @@ data:
   class_path: DLRSDDataModule
   init_args:
     batch_size: 1
-    val_split_pct: 0.34
-    test_split_pct: 0.34
+    val_split_pct: 0.3
+    test_split_pct: 0.3
   dict_kwargs:
     root: 'tests/data/dlrsd'
--- a/tests/conf/dlrsd_multilabel.yaml
+++ b/tests/conf/dlrsd_multilabel.yaml
@@ -10,7 +10,7 @@ data:
   class_path: DLRSDMultilabelDataModule
   init_args:
     batch_size: 1
-    val_split_pct: 0.34
-    test_split_pct: 0.34
+    val_split_pct: 0.3
+    test_split_pct: 0.3
   dict_kwargs:
     root: 'tests/data/dlrsd'
--- a/tests/conf/dlrsd_multilabel.yaml
+++ b/tests/conf/dlrsd_multilabel.yaml
@@ -10,5 +10,7 @@ data:
   class_path: DLRSDMultilabelDataModule
   init_args:
     batch_size: 1
+    val_split_pct: 0.34
+    test_split_pct: 0.34
   dict_kwargs:
     root: 'tests/data/dlrsd'
--- a/tests/conf/dlrsd_multilabel_no_val.yaml
+++ b/tests/conf/dlrsd_multilabel_no_val.yaml
@@ -11,5 +11,6 @@ data:
   init_args:
     batch_size: 1
     val_split_pct: 0
+    test_split_pct: 0.34
   dict_kwargs:
     root: 'tests/data/dlrsd'
--- a/tests/conf/dlrsd_multilabel_no_val.yaml
+++ b/tests/conf/dlrsd_multilabel_no_val.yaml
@@ -11,6 +11,6 @@ data:
   init_args:
     batch_size: 1
     val_split_pct: 0
-    test_split_pct: 0.34
+    test_split_pct: 0.3
   dict_kwargs:
     root: 'tests/data/dlrsd'
--- a/tests/conf/dlrsd_no_val.yaml
+++ b/tests/conf/dlrsd_no_val.yaml
@@ -11,5 +11,6 @@ data:
   init_args:
     batch_size: 1
     val_split_pct: 0
+    test_split_pct: 0.34
   dict_kwargs:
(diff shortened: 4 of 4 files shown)


calebrob6:
@adamjstewart this fixes the warnings you observed. I verified that the latest 3.14 ubuntu test on #3498 has the warnings and this one doesn't.

adamjstewart:
According to the AI policy (still in draft) this kind of PR would be rejected...

calebrob6:
@adamjstewart thanks for the heads up! This is a 6-line change to fix CI warnings, I confidently understand every line (also, the PR summary does an excellent job of explaining it to someone who doesn't), and it fixes the issue. Maybe we should reconsider the part of the draft AI policy to account for this?

adamjstewart:
We thought about adding a qualifier for short PRs (~1–2 lines) but decided against it because humans can easily copy that AI-generated code and open up a PR where AI is not the author and then it's magically "okay" in copyright land. Perhaps the distinction should be less on whether AI wrote 100% of the lines vs. you understand 100% and a human is the author? A human understanding 100% solves the review problem and a human author solves the copyright problem.

isaaccorley:
I don't think this kind of PR should be rejected solely on the basis that it's fully AI generated if the user is aware and it solves an actual problem.

adamjstewart:
If Copilot was the only author, I would still reject it for legal reasons (and we have a few old PRs we need to revert and reapply to fix this). But yes, the code itself is perfect and as long as there are no legal reasons to reject we should probably merge. I'll think about rewording the disclosure form. Perhaps `AI-assisted` vs. `AI-automated`?

I'm still pro-banning AI-generated PR descriptions/comments though. If a contributor doesn't have time to write a description, I don't have time to read it. Although the same is true for code too lol.

adamjstewart (approved):

adamjstewart:
Actually, I read up on this. The decision in Thaler v. Perlmutter and especially in Allen v. Perlmutter is that no matter how hard you work to generate prompts and carefully select results, anything that was created 100% by AI cannot be copyrighted, and all code must be copyrightable. I don't think that applies to this PR simply because it's so small, but would definitely apply to other PRs where AI did everything. Let's keep the wording the same and decide on a case-by-case basis.

adamjstewart (changes requested):
A bit nervous, it may be faster to manually press your fingers on the keyboard to generate the same output and create a new PR...

calebrob6:
Lol I'm absolutely not going to manually type these changes in a new PR. Noone has copyright on `+val_split_pct: 0.34`.

Also, I think this is a distinction without a difference. If I traced through the warnings, then I would have arrived at the same fix or would have added several more test samples to sidestep the issue.  This likely would have taken me like 10 minutes. Instead, I copy pasted the warnings into the copilot instance that had context from #3498 and told it to fix it and open a PR with the fix (as I was also confident that it would be able to do that no problem). This took like 10 seconds, another 30 seconds to read the PR description/proposed changes when it was done (and edit to add the AI policy thing lol), and another minute to compare the CI warnings for ubuntu 3.14 and type the message to you.

I really think you are over-thinking this. If this was like a 500+ line change beast / new feature and I was like "vibe coded it, didn't read it / works on my laptop" then yeah, I agree, totallllly different story. But spending more than 30 seconds on something like this is really annoying.

adamjstewart:
> Noone has copyright on `+val_split_pct: 0.34`.

I agree that this diff is likely too small to be copyrightable anyway. I unfortunately couldn't find any specific definitions of minimum length requirement here.

> I think this is a distinction without a difference. If I traced through the warnings, then I would have arrived at the same fix or would have added several more test samples to sidestep the issue.

100% agree, although the courts unfortunately disagree. _If_ something is copyrightable and it is written by AI, AI cannot hold copyright, and therefore can't transfer copyright to you as "work-for-hire", and therefore you can't transfer copyright to us (TorchGeo). Thus some parts of TorchGeo would be MIT licensed and other parts would be public domain. So _if_ something could be copyrighted, we unfortunately can't accept contributions where it is AI-generated, regardless of how much work went into prompting.

> I really think you are over-thinking this.

Oh I'm definitely overthinking this, no disagreement there.

> But spending more than 30 seconds on something like this is really annoying.

Maintenance can be a fun and enjoyable break from managing humans and robots all day 🙂

adamjstewart (approved):
Pushed a new commit. All lines have now been manually edited by a human, and I can verify that I understand all lines. Now there are no concerns from my side. Still needs a second approval from a human though.

ashnair1 (approved):

PR #3734, opened 2026-05-23
Add branch coverage for GlobalBuildingMap(GBM) plot

crowmoed:
Adds the missing branch coverage for the GlobalBuildingMap.plot method.

Right now the test only hits the default path, so a few branches never get exercised:
- show_titles=False 
- a sample with no 'prediction' key
- suptitle=None

Part of #2501

## AI Usage Disclosure

### DIFF
--- a/tests/datasets/test_gbm.py
+++ b/tests/datasets/test_gbm.py
@@ -44,11 +44,10 @@ class TestGlobalBuildingMap:
         sample['prediction'] = sample['mask']
         dataset.plot(sample, suptitle='Test')
         plt.close()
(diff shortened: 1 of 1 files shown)


adamjstewart (approved):
Confirmed that this now has 100% branch coverage, thanks!

Honestly I wouldn't worry too much about plotting branches right now. I'm trying to find time to work on a generic `PlottingMixin` that will let us delete most duplicated plot methods. That will also be much easier to get branch coverage of. The non-plotting branch coverage is higher priority.

PR #3737, opened 2026-05-23
Fix inplace operations in LTAE causing backpropagation failure

### LINKED ISSUE #3735: Bug in LTAE: inplace operation breaks backpropagation
### Description

While training the LTAE model, backpropagation fails with an in-place modification error during the backward pass.

#### Error

```
RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation: [torch.FloatTensor [128, 32]], which is output 0 of ReluBackward0, is at version 2; expected version 1 instead.
```

The error occurs during `loss.backward()` inside a standard Lightning `Trainer.fit()` run.

The issue appears to come from **in-place operations inside `nn.Dropout` layers used in LTAE**, which interfere with gradient computation in the residual / attention blocks.

#### Temporary workaround

Disabling in-place dropout fixes the issue:

```python
_orig = LTAE.__init__

def patched(self, *args, **kwargs):
    _orig(self, *args, **kwargs)
    for m in self.modules():
        if isinstance(m, nn.Dropout):
            m.inplace = False

LTAE.__init__ = patched
```

### Steps to reproduce
```
import torch
from torchgeo.models import LTAE

model = LTAE(
    in_channels=16,
    d_model=16,
    n_head=4,
    n_neurons=(16, 32),
    len_max_seq=5,
)

x = torch.randn(2, 5, 16)

y = model(x)
loss = y.mean()
loss.backward()
```
### Version

'0.10.0.dev0'

f-schi:
This PR fixes a runtime error during backpropagation in `LTAE` caused by in-place operations interfering with autograd versioning.

### Fix
 Removed `inplace=True` from `nn.Dropout` layers in `LTAE`

Fixes #3735

## AI Usage Disclosure

- [X] 🟢 **No AI usage**: written by humans, for humans


adamjstewart (commented):
Did you check whether or not the reference implementation of L-TAE uses in-place operations? Where was the in-place operation introduced, and by whom? Would love to get more eyes on this so we can know if this is more or less correct than the original.

f-schi:
@adamjstewart 

The source implementation does not use inplace: https://github.com/VSainteuf/lightweight-temporal-attention-pytorch/blob/master/models/ltae.py (Line 70 and 80)

isaaccorley:
This was a recommendation from me to reduce mem usage but if it's needed for backprop then yes we should merge this

isaaccorley (approved):

adamjstewart (approved):

PR #3668, opened 2026-05-06
Add Temporal Regression Trainer

f-schi:
Adds a **TemporalRegressionTask** for sequence-to-value or sequence-to-sequence regression, enabling models to predict future continuous targets from temporal input windows.

Closes #2639

## AI Usage Disclosure

- [ ] 🟢 **No AI usage**: written by humans, for humans
- [x] 🟡 **AI-assisted**: AI helped with the coding, but I understand every line
- [ ] 🔴 **AI-generated**: AI did everything, review with caution

PR #3328, opened 2026-01-25
Trainers: add ClassificationMixin

adamjstewart:
This PR introduces a `ClassificationMixin` that can be shared across all trainers involving patch or pixel classification. In general, I think we should start using mixins more often in our samplers and trainers to avoid code duplication and better support time series variants.

### Pros

* Reduced code duplication
* Ensures consistent feature support across trainers

### New Features

- [x] ClassificationTask: add labels, pos_weight, and ignore_index parameters
- [x] ClassificationTask: add dice loss support
- [x] ClassificationTask: add precision and recall metrics
- [x] SemanticSegmentationTask: add labels and pos_weight parameters
- [x] SemanticSegmentationTask: add dice loss support
- [x] SemanticSegmentationTask: add precision, recall, and F1-score metrics
- [x] ChangeDetectionTask: add labels parameter
- [x] ChangeDetectionTask: add precision and recall metrics
- [x] All three: add overall, average, and class-wise metrics

All while reducing the total number of lines of code 😎 

These could be split into separate PRs if needed.

Closes #2800 
Closes #2130
Closes #2121

@robmarkcole @esmat-fahim47

PR #3738, opened 2026-05-25
RegressionTask: use RegressionMixin

adamjstewart:
#3668 introduced a new `RegressionMixin`, analogous to the `ClassificationMixin` added in #3328. This is already used in `TemporalRegressionTask`, this PR now adds it to `RegressionTask`.

This also fixes a bug in `RegressionMixin` that broke when `num_outputs == 1`. `ClasswiseWrapper` can only be used when `num_outputs > 1`.

## AI Usage Disclosure

### DIFF
--- a/torchgeo/trainers/mixins.py
+++ b/torchgeo/trainers/mixins.py
@@ -149,25 +149,34 @@ class RegressionMixin(LightningModule):
         """
         num_outputs = self.hparams['num_outputs']
         labels = self.hparams['labels']
-        metrics = MetricCollection(
-            {
-                'RMSE': ClasswiseWrapper(
-                    MeanSquaredError(squared=False, num_outputs=num_outputs),
-                    labels=labels,
-                    prefix='RMSE_',
-                ),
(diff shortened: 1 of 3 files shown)


torchgeo/trainers/mixins.py
```diff
@@ -149,25 +149,34 @@ def configure_metrics(self) -> None:
...
-                    prefix='MSE_',
-                ),
-                'MAE': ClasswiseWrapper(
-                    MeanAbsoluteError(num_outputs=num_outputs),
-                    labels=labels,
-                    prefix='MAE_',
-                ),
-            }
-        )
+        if num_outputs > 1:
```
robmarkcole, line 152: Assume this will always be true? Might be better to flag config of num_outputs == 1 as an error?

robmarkcole (approved):
LGTM

torchgeo/trainers/mixins.py
adamjstewart: It is often true for TemporalRegressionTask and almost never true for RegressionTask.

PR #3742, opened 2026-05-26
[Bugfix ]Update jsonargparse version to 4.39 for compatibility with lightning

### LINKED ISSUE #3741: Lightning CLI: missing 1 required keyword-only argument: applied_instantiation_links
### Description

The lightning extras [require](https://github.com/Lightning-AI/pytorch-lightning/blob/master/requirements/pytorch/extra.txt#L8) `jsonargparse >= 4.39` but this is not mirrored in torchgeo, resulting in errors:
```
_InstantiatorFn.__call__() missing 1 required keyword-only argument:
'applied_instantiation_links'
```
Fix: require jsonargparse >= 4.39

### Steps to reproduce

Encountered in https://github.com/torchgeo/torchgeo/pull/3625

### Version

main

robmarkcole:
Close https://github.com/torchgeo/torchgeo/issues/3741

## AI Usage Disclosure


adamjstewart (commented):
Technically this is only correct for the latest version of lightning, for the oldest version of lightning we support, `jsonargparse>=4.27.7` is correct.

adamjstewart (changes requested):
Confirmed that this _does not_ fix #3741

adamjstewart (dismissed):
Reran all of my experiments and it turns out I was wrong! While this lower bound is more strict than necessary for older lightning versions, it is indeed correct for newer lightning versions. Since we can't directly depend on `lightning[pytorch-extra]` because we want to avoid a bunch of other dependencies, there really isn't a better way to document this other than being more strict than necessary.

adamjstewart:
Closing and reopening to restart the failed CI tests...

pyproject.toml
```diff
@@ -41,8 +41,8 @@ dependencies = [
     "einops>=0.3",
     # geopandas 1.0+ required to use pyogrio backend instead of Fiona (unmaintained)
     "geopandas>=1.0",
-    # jsonargparse 4.35+ required for Python 3.12.8+ support
-    "jsonargparse[signatures]>=4.35",
+    # jsonargparse 4.39+ required to match lightning's own jsonargparse dependency
+    "jsonargparse[signatures]>=4.39",
```
```suggestion
    "jsonargparse[jsonnet,signatures]>=4.39",
```
New extra required for lightning 2.5.2+, not required for older but no other way to document this.

uv.lock
```diff
@@ -3805,7 +3805,7 @@ requires-dist = [
     { name = "geopandas", specifier = ">=1.0" },
     { name = "h5py", marker = "python_full_version < '3.14' and extra == 'datasets'", specifier = ">=3.10,<3.15" },
     { name = "ipywidgets", marker = "extra == 'docs'", specifier = ">=7" },
-    { name = "jsonargparse", extras = ["signatures"], specifier = ">=4.35" },
+    { name = "jsonargparse", extras = ["signatures"], specifier = ">=4.39" },
```
adamjstewart:
```suggestion
    { name = "jsonargparse", extras = ["jsonnet", "signatures"], specifier = ">=4.39" },
```

adamjstewart (approved):
@robmarkcole can you confirm that adding a dep on jsonnet doesn't cause issues for you? I know we needed to remove all of pytorch-extra because there were installation issues. I'm not 100% sure it's required since our tests pass without it, but it's recommended by pytorch-extra.

robmarkcole:
You mean issues on my rf-detr branch? I'm AFK until Monday so go ahead and try it out if this is a priority

adamjstewart:
Nah, I mean the issue that required us to drop `lightning[pytorch-extra]` a year ago. Something like bitsandbytes wouldn't install on aarch64 or something like that. Assuming jsonnet doesn't have this same installation issue, we should be good to go.

adamjstewart:
Let's merge just to squash +1 issues, we can always revert later. I double checked and these newer versions of jsonargparse and jsonnet are both on conda-forge, just barely.

PR #3744, opened 2026-05-26
Silence improper Lightning usage warning

adamjstewart:
Silences the following warning in CI:
```
tests/trainers/test_spatiotemporal_segmentation.py::TestSpatioTemporalSegmentationTask::test_binary_task
  /Users/Adam/torchgeo/.venv/lib/python3.14/site-packages/lightning/pytorch/core/module.py:451: You are trying to `self.log()` but the `self.trainer` reference is not registered on the model yet. This is most likely because the model hasn't been passed to the `Trainer`
```
The _correct_ way to do this would be to create a dataset/datamodule and pass it to a trainer. We currently don't have one for binary/multilabel spatiotemporal semantic segmentation, so I opted to ignore the warning for now.

## AI Usage Disclosure

### DIFF
--- a/tests/trainers/test_spatiotemporal_segmentation.py
+++ b/tests/trainers/test_spatiotemporal_segmentation.py
@@ -39,6 +39,7 @@ class TestSpatioTemporalSegmentationTask:
         except MisconfigurationException:
             pass
 
+    @pytest.mark.filterwarnings(r'ignore:You are trying to `self.log\(\)`')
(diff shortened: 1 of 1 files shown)


robmarkcole (approved):
I assume there is nothing to verify

adamjstewart:
You can verify that there is indeed this warning on main and that the warning went away in the CI logs for this PR, but yeah, this PR is pretty innocuous.

PR #3749, opened 2026-05-28
Trainers: organize tasks by inputs/outputs in docs

adamjstewart:
We have a lot more trainers/modules/tasks/models than we used to! The total number is expected to double thanks to our time series efforts with many more still to come: #2382. 

This PR overhauls the documentation to better organize these, making it easier for users to find the correct one. The easiest way I've found to organize them is by their input/output dimensions and whether they are for classification or regression: [Google Sheets](https://docs.google.com/spreadsheets/d/11rENiUywoGp93FSyEaYYIqTd12CCRNDEaWrfqbSZBlI/edit?gid=820154425#gid=820154425). This works well for supervised learning tasks, although may be overkill for the rest.

@robmarkcole @DarthReca FYI

Note that these docs also suffer from #996, no one knows what to call these objects so I called them all of the above as usual...

## AI Usage Disclosure


adamjstewart:
Let me know if anyone has a preference for the format of one of these rows:

| Input  | Output | Task | Trainer |
| ------------- | ------------- | --- | --- |
| $$\mathbb{R}^{C \times H \times W}$$ | $$\mathbb{R}$$ | N/A | RegressionTask
| $$C \times H \times W$$ | scalar | regression | RegressionTask
| C x H x W | scalar | regression | RegressionTask

robmarkcole:
Re format preference, the last looks best to me

docs/api/trainers.rst
```diff
@@ -1,5 +1,111 @@
...
+   :glob:
+
+   trainers/*
+
+TorchGeo provides `LightningModules <https://lightning.ai/docs/pytorch/stable/common/lightning_module.html>`__ for a number of common tasks in geospatial and geotemporal deep learning.
+
+Supervised Learning
+-------------------
+
+Supervised learning tasks have both inputs and labeled outputs.
```
robmarkcole, line 18: `labeled` is American english, thats standard for the repo? Perhaps we doc that

docs/api/trainers.rst
```diff
@@ -1,5 +1,111 @@
...
+     - :ref:`MAETask`
+   * - :math:`\mathbb{R}^{C \times H \times W}` or :math:`\mathbb{R}^{T \times C \times H \times W}`
+     - :ref:`MoCoTask`
+   * - :math:`\mathbb{R}^{C \times H \times W}` or :math:`\mathbb{R}^{T \times C \times H \times W}`
+     - :ref:`SimCLRTask`
+
+Non-Learning Tasks
+------------------
+
+Tasks that do not relate to learning.
```
robmarkcole, line 77: Can add an example?

robmarkcole (approved):
Minor comments

docs/api/trainers.rst
adamjstewart: I guess we don't have any official standards, but most contributors happen to be from America. For once, I'm not sure if I care enough to enforce a standard and change people's English.

docs/api/trainers.rst
adamjstewart: The example is I/O Bench. Couldn't come up with a better name for this section. Could also call it Profiling or Miscellaneous or Other.

docs/api/trainers.rst
robmarkcole: Its just a 'typo' I might pick up on as a Brit, but if we standardised the expectation that could avoid confusion

docs/api/trainers.rst
robmarkcole: Profiling sounds clearer to me

docs/api/trainers.rst
adamjstewart: If we _standardize_ on American English, British readers will be confused. If we _standardise_ on British English, American readers will be confused. I wouldn't bother standardizing on anything unless we have something like codespell to enforce it.

robmarkcole: This would be another PR but under https://docs.torchgeo.org/en/stable/user/contributing.html we could add 

```markdown
## Language conventions

Use American English spelling in:

- code comments
- documentation
- user-facing text
- variable and function names where practical

Examples:
- color, not colour
- initialize, not initialise
- center, not centre
```

codespell looks useful for typos, but not for enforcing style preferences like US/British spelling

docs/api/trainers.rst
adamjstewart: Technically it's more benchmarking than profiling, but technically everything is benchmarking, so that's why I changed again to non-learning.

docs/api/trainers.rst
calebrob6: I don't think the difference is _that_ confusing -- I would not worry about this for any significant length of time

calebrob6:
Why do we need a new .rst file for every trainer now?

adamjstewart:
Technically not required, we could simply add the table to the top of the current unorganized trainers page. 

But we already did this for [models](https://github.com/torchgeo/torchgeo/pull/2892), [datasets](https://github.com/torchgeo/torchgeo/pull/3385), and [samplers](https://github.com/torchgeo/torchgeo/pull/3552), so why not also for trainers? I would maybe also consider doing this for datamodules, as that page is quite large and takes a while to load.

Note that the initial design (everything on one page) was done to mimic torchvision, but torchvision later changed their own docs to have a separate page for each dataset/model/etc. and a single high-level overview that links to everything.

ashnair1 (approved):

adamjstewart:
Since we have the approvals, lets merge before anyone changes their minds. I'm definitely okay with changing from $$\mathbb{R}^{C \times H \times W}$$ to $$C \times H \times W$$ in a follow-up PR if there is consensus for that though. It would be more consistent with our model API docs. I think the organization is the important part, and that's pretty nice.

Issue #1467, opened 2023-07-06
Add legends to semantic segmentation plots


adamjstewart:
### Summary

We should add legends to our semantic segmentation mask plots so we can easily visualize what each class is.

### Rationale

Datasets like CDL have 134 output classes. I have 2 or 3 of these memorized, but certainly not all 134.

calebrob6:
I think we have to be careful here as torchgeo isn't a viz library and questions like "how to make a legend support 134 classes" is in viz territory

alhridoy:
Hi @adamjstewart 

I completely agree that having legends would greatly enhance our understanding and visualization of each class.

I would like to work on this enhancement. Here's a brief plan:

1. Identify where and how the semantic segmentation mask plots are currently generated.
2. Design an intuitive and clear legend that can handle a large number of classes without overwhelming the visualization.
3. Implement the legend into the plotting function, ensuring it adapts to various numbers of classes.
4. Test with datasets of varying class numbers to ensure legibility and accuracy.
5. Update documentation or user guides, if needed, to explain the new feature.

PR #3755, opened 2026-05-31
PlottingMixin: add mixin for all dataset plotting

### LINKED ISSUE #1971: Add plot method to IntersectionDataset
### Summary

Add a plot method to IntersectionDataset that calls the plot method of all GeoDatasets and merges their output into a single figure.

### Rationale

TorchGeo currently relies on the following fundamental design principle:

* Each dataset has a `plot` method that returns a single `Figure`
* Each data module has a `plot` method that calls `dataset.plot` and returns a single `Figure`
* Each trainer calls this `plot` method during validation and logs a single `Figure` to tensorboard

This works great for most `NonGeoDatasets` and `GeoDatasets`, but falls apart as soon as you use an `IntersectionDataset`. Merging two `Figures` together is non-trivial, and so far we've simply ignored this use case by plotting the mask/prediction but not the image. This prevents proper visualization of training and makes `IntersectionDatasets` different from other `GeoDatasets` in terms of features.

### Implementation

This idea can be implemented in multiple stages (possibly by multiple people).

First, we'll need to add an abstract `plot` method to `GeoDataset`. There are a few point datasets that do not yet have `plot` methods (`EDDMapS`, `GBIF`, `INaturalist`), so we'll have to add one. Might as well add one to `NonGeoDataset` as well to enforce consistency. Regardless, this is likely a good idea to do, it's just a prerequisite for what follows.

Second, we'll need to add a new parameter to all `GeoDataset.plot` methods containing an optional pointer. The idea is that normally, `GeoDataset.plot` will create its own `Figure` and returns it, but `IntersectionDataset.plot` will instead create a `Figure` and call each `GeoDataset.plot` method with the same `sample` and this additional pointer telling it which subplot to populate. Or maybe pass it an existing figure and have it select the next empty subplot to use? The number of subplots to create in `IntersectionDataset.plot` could either be based on the number of datasets (would exclude plotting prediction maps) or on the number of image/mask/prediction elements in the `sample` dict (relies on #985 if we want any guarantees). See Additional information for caveats.

### Alternatives

We could instead change `GeoDataset.plot` and `GeoDataModule.plot` to return one _or more_ `Figures` and our trainers to plot a list of figures. This may actually be significantly simpler to implement. The only downside is that not all datasets will be consistent, and it would be nice to be able to display a single `Figure` with all information.

### Additional information

Some remaining TODOs to figure out:

* How will this work for `UnionDataset`?
* How will this work for `IntersectionDataset` with multiple images (e.g., Landsat & Sentinel)?

This is related to #1263 and could likely be done in parallel.

### LINKED ISSUE #2377: Ensure all datasets have a plot method

adamjstewart:
### Motivation

One of the largest sources of code duplication in TorchGeo is in our dataset plotting. There is currently limited consistency in how datasets get plotted, but 90% of datasets involve the following steps:

1. If 'image' key:
  a. Select `rgb_bands` out of `all_bands`
  b. Raise an error if RGB bands not found
  c. Apply normalization (either division or quantile normalization)
  d. Use imshow to plot the image
2. If 'mask' key:
  a. Use imshow to plot the mask (with cmap, vmin, vmax)
  b. Same for predicted mask if available
3. If 'label' key:
  a. Becomes image title
  b. Same for predicted label if available
4. Always:
  a. Remove axes labels
  b. Apply suptitle if requested

As our Sample keys become more consistent, so to can our plotting. I also have a lot of clever ideas for better semantic segmentation plots (#1467) and time-series plots (an animated GIF). These are quite challenging at the moment without repeating the code in dozens of places, but could be made more consistent using a mixin.

### Implementation

I propose a shared [mixin](https://en.wikipedia.org/wiki/Mixin) for all things related to plotting. This will be shared for both NonGeoDataset and GeoDataset to avoid even more duplication. This mixin will define all dataset attributes (e.g., `all_bands`, `rgb_bands`, `cmap`), helper functions (e.g., `quantile_normalization`), and public functions (e.g., `plot`) related to plotting. We can start small and port a few datasets at a time, but the eventual goal is to remove the `plot` method from 90% of datasets. About 10% of datasets will still require a custom `plot` method to deal with intricacies of the data (e.g., multi-modal datasets), but that's okay.

### Reviewer Instructions

I would recommend first reviewing `torchgeo/datasets/mixins.py` to see the standards being proposed. All other file changes are unfortunately backwards-incompatible changes required to align existing datasets with this proposed standard.

### AI Usage Disclosure

Closes #2377 
Closes #1971


adamjstewart:
Closing in favor of #3775, let's go one step at a time: #3774.

PR #3756, opened 2026-06-01
docs: convert tutorial notebooks from pip to uv

ArsalaanAhmad:
Converts all tutorial notebooks from `%pip install` to `!uv pip install`, 
as uv is significantly faster and is already pre-installed on both Google 
Colab and Lightning Studios.

15 notebooks updated. `contribute_datamodule.ipynb` and `geospatial.ipynb` 
were untouched as they contain no install cells.

Addresses part of #2905 (partially since README and other files tracked separately).

## AI Usage Disclosure

### DIFF
--- a/docs/tutorials/change_detection.ipynb
+++ b/docs/tutorials/change_detection.ipynb
@@ -33,7 +33,7 @@
    "id": "3",
    "metadata": {},
    "outputs": [],
-   "source": "%pip install torchgeo tensorboard"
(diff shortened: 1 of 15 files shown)


adamjstewart:
How much faster is it? Is it less verbose?

adamjstewart (approved):
Confirmed that this is much faster and quieter and works on Colab.

calebrob6:
Cool, thanks @ArsalaanAhmad!

PR #3757, opened 2026-06-02
feat: MetaCHM dataset

isaaccorley:
[Meta CHMv2 (DINOv3) global canopy height map](https://registry.opendata.aws/dataforgood-fb-forestsv2/) as a torchgeo `RasterDataset`

It reads the tile index here `https://data.source.coop/tge-labs/meta-chm-v2/tiles.parquet` which contains the `cog_url`, `datetime` of each COG in the official AWS Open Data s3 bucket. 

meta_chm_seattle

## AI Usage Disclosure

I used AI to help create test data and tests and to make sure it works with a `RandomGeoSampler` and plot correctly.


torchgeo/datasets/meta_chm.py
```diff
@@ -0,0 +1,137 @@
...
+
+    .. versionadded:: 0.10
+    """
+
+    url = 'https://data.source.coop/tge-labs/meta-chm-v2/stac/items.parquet'
+
+    is_image = False
+    dtype = torch.float32
+    all_bands = ('chm',)
+    _res = (1.1943285669558463, 1.1943285669558463)
```
calebrob6, line 52: Why is this so precise?

calebrob6 (approved):

calebrob6:
How does this rasterdataset know its CRS?

Also, I confirmed this works and is sick:

marymoor_chm

isaaccorley:
> How does this rasterdataset know its CRS?
> 
> 
> 
> Also, I confirmed this works and is sick:
> 
> 
> 
> marymoor_chm
> 
> 

They are all 3857 but we can be explicit and set the CRS attr

torchgeo/datasets/meta_chm.py
```diff
@@ -0,0 +1,137 @@
...
+      height in meters (0 = no canopy or no data)
+
+    If you use this dataset in your research, please cite the following paper:
+
+    * https://arxiv.org/abs/2603.06382
+
+    .. versionadded:: 0.10
+    """
+
+    url = 'https://data.source.coop/tge-labs/meta-chm-v2/stac/items.parquet'
```
nilsleh, line 47: would there be any reason to also support V1 and make this work for both versions?

calebrob6:
Ahh okay and that's the default for RasterDataset

torchgeo/datasets/meta_chm.py
isaaccorley: I don't believe so. V1 is inferior to V2 and was hosted on GEE so I'm not going to rehost that on Source Coop.

docs/api/datasets/masks.csv
```diff
@@ -11,6 +11,7 @@ Dataset,Type,Source,Size (px),Resolution (m),License
 :ref:`GlobBiomass`,Masks,Landsat,"45,000x45,000",100,CC-BY-4.0
 :ref:`Global Mangrove Distribution`,Masks,"Remote Sensing, In Situ Measurements",-,3,public domain
 :ref:`GlobalBuildingMap`,Masks,PlanetScope,180K,3,CC-BY-4.0
+:ref:`Meta CHM`,Masks,Maxar,"32,768x32,768",1.19,CC-BY-4.0
```
adamjstewart, line 15: Where did you find CC-BY-4.0? According to the paper, the dataset is released under the DINOv3 license.

torchgeo/datasets/meta_chm.py
```diff
@@ -0,0 +1,137 @@
...
+        self.cache = cache
+        self.bands = self.all_bands
+        self.band_indexes = None
+        self.time_series = False
+
+        request = urllib.request.Request(self.url, headers={'User-Agent': 'torchgeo'})
+        with urllib.request.urlopen(request) as response:
+            buffer = io.BytesIO(response.read())
+        columns = ['geometry', 'assets', 'datetime']
+        gdf = gpd.read_parquet(buffer, columns=columns)  # ty: ignore[invalid-argument-type]
```
adamjstewart: What was the cause of the invalid type here? Is this a bug in geopandas' type hints?

adamjstewart (commented):
Quite different from our other datasets, but maybe that's a good thing. I wonder if this could someday be a subclass of STACDataset. Pretty cool that `__getitem__` works without any changes.

docs/api/datasets/masks.csv
isaaccorley: https://registry.opendata.aws/dataforgood-fb-forestsv2/

torchgeo/datasets/meta_chm.py
isaaccorley: Yeah it was having a problem with buffer but it's been awhile so let me remove and see if it's since been fixed

torchgeo/datasets/meta_chm.py
isaaccorley: LOL I have no idea I just copied from one of the tifs

calebrob6:
This lgtm

calebrob6 (dismissed):

docs/api/datasets/masks.csv
adamjstewart: Interesting, not sure what to do in case of discrepancy. Could we ping the original authors? Doesn't have to hold up the PR, but would be nice to get confirmation on who is wrong.

torchgeo/datasets/meta_chm.py
adamjstewart: Problem is that `buffer` is `bytes` and `read_parquet` expects `str`, easy to fix with a cast.

torchgeo/datasets/meta_chm.py
```diff
@@ -0,0 +1,137 @@
...
+
+        request = urllib.request.Request(self.url, headers={'User-Agent': 'torchgeo'})
+        with urllib.request.urlopen(request) as response:
+            buffer = io.BytesIO(response.read())
+        columns = ['geometry', 'assets', 'datetime']
+        gdf = gpd.read_parquet(str(buffer), columns=columns)
+        gdf.to_crs('EPSG:3857', inplace=True)
+        filepaths = (
+            gdf['assets']
+            .map(lambda asset: asset['chm']['href'])
```
adamjstewart: Don't really know what's going on with the ty error here, it seems to think that `asset['chm']` is `NAType`, which doesn't make sense. Easy to ignore, but would be nice to understand it better first...

adamjstewart (dismissed):

calebrob6 (approved):

PR #3759, opened 2026-06-02
test: reduce change-detection model test memory

isaaccorley:
Use `torch.inference_mode()` instead of `no_grad()` and shrink the FarSeg/ChangeStar test inputs from 128² to 64².

Profiling the affected tests (Ubuntu, 2 CPU): peak RAM 1256→1187 MB, runtime 10.8s→9.9s.


isaaccorley:
Closing — negligible peak-RAM impact (1256→1187 MB), not worth a standalone PR.

PR #3765, opened 2026-06-03
Isolate the Inria checksum test

isaaccorley:
Use a temporary directory for the Inria checksum test so it can run safely with pytest-xdist.

## AI Usage Disclosure

### DIFF
--- a/tests/datasets/test_inria.py
+++ b/tests/datasets/test_inria.py
@@ -2,13 +2,12 @@
 # Licensed under the MIT License.
 
 import os
-import shutil
+from pathlib import Path
 
 import matplotlib.pyplot as plt
 import pytest
 import torch
 from _pytest.fixtures import SubRequest
-from pytest import MonkeyPatch
 from torch import nn
 
 from torchgeo.datasets import DatasetNotFoundError, InriaAerialImageLabeling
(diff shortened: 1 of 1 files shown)


tests/datasets/test_inria.py
adamjstewart: This is the real problem. Instead of doing a full copy to tmp_path for all 16 tests, why not use `path.touch()` to create an empty zip file with incorrect checksum? We could also redesign `_verify` to skip this path and remove this test.

torchgeo/datasets/utils.py
```diff
@@ -364,18 +365,25 @@ def extract_archive(
...
-            z.extractall(to_path)
-    elif suffixes[-1] == '.bz2' and '.tar' not in suffixes:
-        stem = pathlib.Path(from_path).stem
-        to_path = os.path.join(to_path, stem)
-        with bz2.open(from_path, 'rb') as src, open(to_path, 'wb') as dst:
-            dst.write(src.read())
-    else:
-        with tarfile.open(from_path, 'r') as t:
-            t.extractall(to_path, filter='data')
+    os.makedirs(to_path, exist_ok=True)
```
adamjstewart: I was unable to reproduce any issues with `extract_archive` under parallel testing. Can you clarify what issue this solves?

adamjstewart (changes requested):

tests/datasets/test_inria.py
isaaccorley: Updated. The checksum test now creates an empty archive in `tmp_path`; the other tests keep using the shared fixture.

torchgeo/datasets/utils.py
isaaccorley: Removed. This PR no longer changes `extract_archive`.

adamjstewart (approved):

PR #3770, opened 2026-06-05
WesternUSALiveFuelMoisture: add time series support

madhavcodez:
## AI Usage Disclosure

---

Part of #3042. Converts `WesternUSALiveFuelMoisture` to return its features as a time series.

### What changed

- `__getitem__` now returns `sample['input']` as a `(T, C)` tensor of shape `(4, 34)` (4 time steps `t, t-1, t-2, t-3` down the rows, 34 variables across the columns) instead of a flat 138-length vector.
- The static point coordinates are returned under their own `sample['lat']` and `sample['lon']` keys (matching the convention already used by `CopernicusBench`, `MMEarth`, and `USAVars`), rather than mixed into the feature vector.
- `input_features` now selects base variable names (e.g. `'ndvi'`) without the time-step suffix; each selected variable is returned across all `time_steps`. The old 138-entry `all_variable_names` tuple is replaced by `variable_names` (the 34 base names) plus `time_steps`.
- `plot()` reads the new `(T, C)` layout and the `lat`/`lon` keys, and raises a clear `ValueError` if none of the requested `variables_to_plot` are present in `input_features`.

This is a backwards-incompatible change to the sample format, consistent with the goal of #3042.

### Tests

- `input` shape `(4, 34)`, dtype, and time-step ordering (asserted against the fixture's distinct per-step `vv` values).
- `lat`/`lon`/`label` values match the fixture.
- A custom `input_features=['vv', 'ndvi']` subset yields `(4, 2)`.
- Invalid `input_features` raises, and `plot()` filters unknown variables / raises when none are valid.

### Notes

- Scoped to the dataset itself; a `LightningDataModule` can follow once a sequence-regression trainer is wired, as discussed on #3042.
- `ruff check` and `ruff format` pass.

### DIFF
--- a/tests/datasets/test_western_usa_live_fuel_moisture.py
+++ b/tests/datasets/test_western_usa_live_fuel_moisture.py
@@ -31,7 +31,32 @@ class TestWesternUSALiveFuelMoisture:
         x = dataset[0]
         assert isinstance(x, dict)
         assert isinstance(x['input'], torch.Tensor)
+        assert x['input'].shape == (4, 34)
+        assert x['input'].dtype == torch.float32
+        assert isinstance(x['lon'], torch.Tensor)
+        assert isinstance(x['lat'], torch.Tensor)
         assert isinstance(x['label'], torch.Tensor)
+        # time steps run t, t-1, t-2, t-3 down the rows
+        vv = dataset.input_features.index('vv')
+        assert x['input'][0, vv] == pytest.approx(-12.80108143, rel=1e-5)
+        assert x['input'][3, vv] == pytest.approx(-12.35794964, rel=1e-5)
+        assert x['lon'] == pytest.approx(-115.8855556, rel=1e-5)
+        assert x['lat'] == pytest.approx(42.44111111, rel=1e-5)
+        assert x['label'] == pytest.approx(132.6666667, rel=1e-5)
+
+    def test_input_features_subset(
+        self, azcopy: Executable, monkeypatch: MonkeyPatch, tmp_path: Path
+    ) -> None:
+        url = os.path.join('tests', 'data', 'western_usa_live_fuel_moisture')
+        monkeypatch.setattr(WesternUSALiveFuelMoisture, 'url', url)
+        dataset = WesternUSALiveFuelMoisture(
+            tmp_path, input_features=['vv', 'ndvi'], download=True
+        )
+        assert dataset[0]['input'].shape == (4, 2)
+
+    def test_invalid_input_features(self, tmp_path: Path) -> None:
+        with pytest.raises(AssertionError):
+            WesternUSALiveFuelMoisture(tmp_path, input_features=['not_a_variable'])
 
     def test_len(self, dataset: WesternUSALiveFuelMoisture) -> None:
         assert len(dataset) == 3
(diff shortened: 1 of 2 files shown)


tests/datasets/test_western_usa_live_fuel_moisture.py
```diff
@@ -1,57 +1,91 @@
...
+
+    def test_not_downloaded(self, tmp_path: Path) -> None:
+        with pytest.raises(DatasetNotFoundError, match='Dataset not found'):
+            WesternUSALiveFuelMoisture(tmp_path)
+
+    def test_plot(self, dataset: WesternUSALiveFuelMoisture) -> None:
+        sample = dataset[0]
+
+        # Test with a single variable - likely one of the missing lines
+        fig = dataset.plot(sample, variables_to_plot=['vv'])
```
robmarkcole, line 75: How do these look?

torchgeo/datasets/western_usa_live_fuel_moisture.py
```diff
@@ -23,8 +23,8 @@ class WesternUSALiveFuelMoisture(NonGeoDataset):
 
     This tabular style dataset contains fuel moisture
     (mass of water in vegetation) and remotely sensed variables
-    in the western United States. It contains 2615 datapoints and 138
```
robmarkcole, line 26: 34 x 4 = 136 - what were the remaining 2, (lat,lon)?

madhavcodez: Yep, exactly — lat and lon. They were the last two entries in the old `all_variable_names` (136 variable×timestep columns, then `lat` and `lon`), which is why the old `plot()` pulled them off as `input_data[-2]`/`[-1]`. This PR just splits them into their own `sample['lat']`/`['lon']` keys so the input reshapes cleanly to 34×4.

tests/datasets/test_western_usa_live_fuel_moisture.py
madhavcodez: Good call to render them first. Multi-variable looks fine, but the single-variable one had the caption overlapping the line — it was pinned at `y=-0.7` of the last axes and `tight_layout` doesn't reserve space for that. Moved it to `fig.supxlabel` (plus a plain `fig.suptitle`) so it lays out cleanly for any number of panels — pushed in 6743458. Before/after for `['vv']` below, with a multi-variable view for reference. Left the x-axis as t→t-1→t-2→t-3 (most recent on the left); happy to flip to chronological if you'd rather.

robmarkcole:
@madhavcodez can you test with one of the trainers, is it pixel or image level regression?

madhavcodez:
It's neither in the imagery sense — this is tabular point regression: each sample is a `(T=4, C=34)` sequence of per-site variables with a single scalar LFMC target, so the closest analog is image-level (one value per sample). There are no spatial dims to do pixel-level regression on.

The matching trainer is `TemporalRegressionTask`, which consumes exactly the `(B, T, C)` input this PR emits (`RegressionTask`/`PixelwiseRegressionTask` build timm/smp image backbones, so they don't apply). I tested it end-to-end with `TemporalRegressionTask` + LTAE through a datamodule mirroring `AirQualityDataModule` — random 60/20/20 split, train-split normalization, `target_mean`/`target_std` exposed for denormalized metrics, and `label` reshaped to a `(1, 1)` `target`:

1. **Wiring (CI fixture):** `fit` / `validate` / `test` / `predict` all run; predictions come out `(B, 1, 1)`.
2. **Real data (all 2,615 samples, 60 epochs, CPU):** trained LTAE reaches test RMSE 24.3 / MAE 17.2 (LFMC %), vs 37.1 / 28.7 for a predict-the-train-mean baseline. For reference, Rao et al. 2020 report RMSE ≈ 25 with their physics-assisted RNN — not directly comparable (random split vs. their site-based CV), but it shows the `(T, C)` layout trains properly rather than just running.
3. **`input_features` subset (`['vv','vh','ndvi','ndwi','red','nir']`, `in_channels=6`):** trains end-to-end (test RMSE 30.3), so the feature-selection semantics flow through training too.

Happy to add the `WesternUSALiveFuelMoistureDataModule` + a trainer test config (mirroring `air_quality_mse.yaml`) either in this PR or as a follow-up — whichever you prefer.

robmarkcole:
@MathiasBaumgartinger the `WesternUSALiveFuelMoistureDataModule` + trainer test config could be added to this PR IMO

tests/datasets/test_western_usa_live_fuel_moisture.py
adamjstewart, line 1: Usually when every line of code changes, it's something to do with line endings. Can you double check this? Your version has a lot of additional tests that are not necessary, you only need to update the expected dimensions.

adamjstewart (commented):
Just need to make this minimal and easy to review now.

PR #3778, opened 2026-06-09
Run ty in uv env during pre-commit

adriantre:
Last time I bumped ty in pre-commit I could not find a way to automatically keep it in sync with pyproject.toml. Did some research and found out it is possible to run it from the uv environment instead of duplicating the dependencies here.

## AI Usage Disclosure

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,25 +16,12 @@ repos:
     hooks:
       - id: ty
         name: ty check
-        entry: ty check
-        language: python
+        # Run in the locked uv environment so the ty version and type stubs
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -16,25 +16,12 @@ repos:
     hooks:
       - id: ty
         name: ty check
-        entry: ty check
-        language: python
+        # Run in the locked uv environment so the ty version and type stubs
```
isaaccorley: Can you remove the comments

isaaccorley (changes requested):

isaaccorley (approved):

PR #3779, opened 2026-06-09
Make GeoDatasets list files in archives automatically (opt-in)

adriantre:
~NB: Builds off of #1399. Will rebase on main once it's merged. Feel free to wait with the review until that happens.~

Adds a class argument to GeoDatasets that lets the user choose to have `files` property list files inside archives.
This was possible already, but they would have to point directly to the archive and use the correct prefix `/vsizip/`. With this PR, it will automatically add the prefix to all archives found in the root.

Any subclass — RasterDataset, VectorDataset, or a user's dataset — can opt in with a one-liner:
```python
  class MyDataset(RasterDataset):
      descend_into_archives = True
```
The default False means zero behavior change for every existing dataset.

## AI Usage Disclosure


adamjstewart:
I actually think this should be the default behavior. `files` should list files, if someone wants the old behavior we could add `paths` to list files and directories. Technically backwards-incompatible, but I would be surprised if people are relying on the current behavior.

adriantre:
> I actually think this should be the default behavior. `files` should list files, if someone wants the old behavior we could add `paths` to list files and directories. Technically backwards-incompatible, but I would be surprised if people are relying on the current behavior.

I agree, but there is one caveat we would have to handle. If people have unpacked the archives and kept the archive alongside the unpacked folder, `files` will return duplicates. Since they both contain the same files, but have different path.

adriantre:
I guess that an archive is technically a file. There are specific cases where single-tif archives is directly open'able `rasterio.open('/vsizip/filename_matching_glob.zip')`. Not that I know how common this useage is in geospatial applications.

What do you think about keeping false by default for now?

adamjstewart:
Another thing we haven't thought about are Zarr/Parquet stores that are typically presented as a directory, but Xarray/Pandas can open the entire directory at once. I agree there are probably cases for and against recursive listing by default. We already have enough minor inconsistencies where datasets claim to accept a list of paths but actually only work properly if one directory is given. I'm ~not~ _now_ leaning towards opt-in, or even making this a private attribute. Could also make this a utility that users can run themselves before passing a list of paths to the dataset.

adriantre:
> I'm not leaning towards opt-in, or even making this a private attribute. Could also make this a utility that users can run themselves before passing a list of paths to the dataset.

Did you mean you _are_ leaning towards opt-in?

adamjstewart:
Yes, typo, fixed 😅

adriantre:
The clearest example btw. is Sentinel SAFE-product archives (zipped). Rasterio/gdal understands how to read from the zip-files directly.

adamjstewart:
Honestly not sure how I feel about this PR. How critical is this feature for you? Would like to see a bit more justification before adding yet another class attribute/helper function and making the dataset even larger than it already is.

adriantre:
With this, users can provide a single root path without unpacking archives within the tree. Without it, they would have to pass the full list of all archives that I want to use. So it is not critical for me, as there exist workarounds.

I would love to find a natural way to have this functionality in torchgeo, though.

adamjstewart:
I thought we did that already somewhere in the code?

adriantre:
> I thought we did that already somewhere in the code?

It does if you pass a vsi-prefixed path like `/vsiaz/...` or `/vsizip/...`. This PR is letting users specify a local `paths=data`, and all files in all archives are automatically listed

adriantre:
All right, so I wrote a version of this that descends into archives automatically, but prefers already-unpacked files (no duplicates).

https://github.com/torchgeo/torchgeo/pull/3984/changes

adriantre:
In case the daft #3984 was TL;DR.

This is my suggestion of how to reduce complexity.

```python
def find_files(path: Path, filename_glob: str = '*') -> list[str]:
    """Return all files under *path* that match *filename_glob*.

    Dispatch is purely by location kind. Each helper knows the one "container"
    native to its branch — archives for local files, Zarr stores for
    directories / VSI paths / URLs — and handles that root itself.
    """
    if os.path.isfile(path):
        files = _files_under_file(path, filename_glob)   # plain file, or archive → expand
    elif os.path.isdir(path):
        files = _files_under_dir(path, filename_glob)    # walk
    elif str(path).startswith('/vsi'):
        files = _files_under_vsi(path, filename_glob)    # GDAL/pyogrio (/vsi prefix) for Raster
    elif '://' in str(path):
        files = _files_under_url(path, filename_glob)    # fsspec (gs:// prefix) for Xarray
    else:
        files = set()
    return sorted(files)
```

The goal is that input can be a root path, or direct filepath, regardless whether it is .zarr, .zip, vsi, or local. Homogeneous behavior for all.

PR #3782, opened 2026-06-11
Add global mangrove watch

### LINKED ISSUE #3605: Add dataset: Global Mangrove Watch Data
Paper: [Global Mangrove Extent Change 1996–2020: Global Mangrove Watch Version 3.0](https://www.mdpi.com/2072-4292/14/15/3657)
Source: [zenodo](https://doi.org/10.5281/zenodo.6894273)
License: https://creativecommons.org/licenses/by-nc/4.0/
Format: geotiff and shp
Additional reference: https://gee-community-catalog.org/projects/mangrove/

akhellad:
Closes #3605

Adds the Global Mangrove Watch (GMW) v3.0 dataset as a RasterDataset, following the pattern of CDL and NLCD as suggested in the issue.

The dataset provides binary mangrove extent maps at 25m resolution for 11 epochs between 1996 and 2020, served as 1° x 1° GeoTIFF tiles from Zenodo (CC-BY-4.0).

- Implemented `GlobalMangroveWatch(RasterDataset)` with year selection, download, extraction, and plotting
- Added fake test data generator and tests (100% coverage)
- Added documentation and metadata CSV entry

## Example output

Mangrove extent mask for 2020 over the Sundarbans region, confirming correct data loading:

gmw_example

## AI Usage Disclosure


torchgeo/datasets/globalmangrovewatch.py
```diff
@@ -0,0 +1,226 @@
...
+
+    0. Non-mangrove
+    #. Mangrove
+
+    If you use this dataset in your research, please cite it using the
+    following format:
+
+    * https://doi.org/10.5281/zenodo.6894273
+
+    .. versionadded:: 0.9
```
nilsleh: 0.9 is the active release, this should be 0.10

torchgeo/datasets/globalmangrovewatch.py
```diff
@@ -0,0 +1,226 @@
...
+
+    * Binary mangrove extent maps at 25m resolution
+    * 11 epochs: 1996, 2007-2010, 2015-2020
+    * Global coverage in WGS84 geographic coordinates
+    * Tiled as 1° x 1° GeoTIFF files
+
+    Dataset classes:
+
+    0. Non-mangrove
+    #. Mangrove
```
nilsleh: why is the class for mangrove a "#"?

torchgeo/datasets/globalmangrovewatch.py
```diff
@@ -0,0 +1,226 @@
...
+
+from .errors import DatasetNotFoundError
+from .geo import RasterDataset
+from .utils import Path, Sample, download_url, extract_archive
+
+
+class GlobalMangroveWatch(RasterDataset):
+    """Global Mangrove Watch (GMW) dataset.
+
+    The `Global Mangrove Watch
```
nilsleh, line 25: could add a comment that this is Version 3 of the dataset

nilsleh (commented):
Thank you for the contribution. Can you show some example plots to the PR description to check that the data loading works correctly and the shown masks make sense.

torchgeo/datasets/globalmangrovewatch.py
adamjstewart: This is a rST thing, #. generates a numbered list where you don't have to care about numbers, but it starts at 1 unless you explicitly start at 0.

akhellad: Switched to explicit numbering (0. / 1.) since the classes are 0-indexed, avoids mixing manual and auto-numbered list items.

akhellad:
Thanks for the review! I've addressed all points. I also fixed the filename regex and download URL after testing against the real Zenodo data, the initial pattern was based on an assumption that didn't match the actual file naming (GMW_N00E008_2020_v3.tif).
Here's an example plot showing the 2020 mangrove mask over the Sundarbans, confirming the data loads correctly:

gmw_example

nilsleh:
Can you also rebase the PR with the latest changes from main, then we can do another review.

akhellad:
Done, rebased onto the latest main !

torchgeo/datasets/globalmangrovewatch.py
```diff
@@ -0,0 +1,226 @@
...
+
+        if not self.download:
+            raise DatasetNotFoundError(self)
+
+        self._download()
+        self._extract()
+
+    def _download(self) -> None:
+        """Download the dataset."""
+        assert isinstance(self.paths, str | os.PathLike)
```
nilsleh: Both in the _download and _extract function you are checking the self.paths argument. Perhaps this can be moved once into the __init__ for a single check and also document the assertion error in that docstring, then these sub functions are cleaner

nilsleh: similarly the paths can be casted once I think.

torchgeo/datasets/globalmangrovewatch.py
```diff
@@ -0,0 +1,226 @@
...
+            return
+
+        exists = []
+        assert isinstance(self.paths, str | os.PathLike)
+        paths = cast(Path, self.paths)
+        for year in self.years:
+            pathname = os.path.join(paths, self.zipfile_glob.replace('*', str(year)))
+            if os.path.exists(pathname):
+                exists.append(True)
+                self._extract()
```
nilsleh: you are repeatedly calling this function per year, and then again the extract function is also looping again over the years every time. Maybe there could be a single _extract(year) function with a year argument and another function _extract_all() which you can call at the end of this _verify() function to extract all. the all extract function can then also call the single extract with each year.

akhellad:
Addressed all three points: moved the path assertion and cast once into __init__ (documented in the docstring), and split extraction into _extract(year) and _extract_all().

nilsleh (dismissed):

akhellad:
Gentle ping @adamjstewart, happy to address any feedback if needed!

adamjstewart:
Thanks for the reminder, will try to review on Thursday after I survive the next several deadlines.

akhellad:
No worries, thanks and good luck with the deadlines!

isaaccorley (dismissed):
This initially looked good to me but Codex found some suggestions and bugs that should be fixed when I tried using it locally.

1. Year filtering is broken. years=[1996] can load other years already under the root.
2. Time-series plotting crashes. Multi-year samples produce an unsupported 4D image array.
3. Multi-path input is rejected. The signature permits multiple paths, but the constructor requires one root.
4. years=[2020] is a mutable default. Should use tuples that can't be modified.
5. Existing ZIPs bypass checksum validation. Validate them before extraction when requested.
6. Dataset metadata is inaccurate. Real tiles are 4,500×4,500; the fake raster also differs in nodata and transform.

akhellad:
Thanks for taking the time to test this locally, all six points are fixed now.

- **Year filtering.** `files` is now overridden to keep only the tiles whose `date` group is in `years`, so `years=[1996]` no longer picks up other epochs sitting under the same root.
- **Time-series plotting.** `plot()` reshapes both `[H, W]` and `[T, H, W]` masks into a stack of frames and renders one column per time step (one row for the mask, one for the prediction), instead of handing a 4D array to `imshow`.
- **Multi-path input.** The single-root assertion moved out of `__init__` into a `_root` property that is only used for download/extraction, so multiple paths work as the signature advertises and still fail with a clear message when a download is actually needed.
- **Mutable default.** `years` now defaults to `(2020,)` and `all_years` is a tuple.
- **Checksum of existing ZIPs.** Pre-existing archives are validated with `check_integrity` before extraction and raise `RuntimeError('Dataset found, but corrupted.')`. Freshly downloaded archives aren't re-hashed, since `download_url` already validated them.
- **Metadata.** Corrected in the docstring *and* in `docs/api/datasets/masks.csv`, which claimed 3,600x3,600. The fake raster now uses the real nodata value and transform.

While going through it I found three more issues:

1. `cmap` was declared as a `ClassVar[dict]`, which is an invalid override of `PlottingMixin.cmap` — `ty check` was already failing on this branch. It's now a `ListedColormap`, following the NLCD convention, and matching the palette embedded in the source files.
2. Partial extraction was broken: `if self.files: return` meant that if one requested year was already extracted, the remaining years were never downloaded. Verification is now done year by year. Note that `NLCD` still has this bug, if you want me to open a separate issue.
3. `masks.csv` had the same wrong tile size as the docstring.

I verified the metadata against the real data rather than guessing: the 2020 archive from Zenodo contains 4,500x4,500 uint8 tiles with `nodata=0` and a resolution of `0.000222222222222`, laid out as `gmw_v3_2020/GMW_N00E008_2020_v3.tif`, which matches the glob and regex. I also re-hashed all eleven archives and every MD5 in the table is correct.

Each fix has a regression test; I checked that all six new tests fail against the previous version of the file. Coverage on the dataset is still 100%.

tests/datasets/test_globalmangrovewatch.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+class TestGlobalMangroveWatch:
+    @pytest.fixture
+    def dataset(self, monkeypatch: MonkeyPatch, tmp_path: Path) -> GlobalMangroveWatch:
+        url = os.path.join(
+            'tests', 'data', 'globalmangrovewatch', 'gmw_v3_{}_gtiff.zip'
+        )
+        monkeypatch.setattr(GlobalMangroveWatch, 'url', url)
+        monkeypatch.setattr(plt, 'show', lambda *args: None)
```
adamjstewart: Is this needed? I don't think you use plt.show anywhere

tests/datasets/test_globalmangrovewatch.py
adamjstewart, line 1: Tests are quite a bit longer than most of our other datasets, can you remove the tests that aren't needed to achieve coverage?

docs/api/datasets/masks.csv
```diff
@@ -11,6 +11,7 @@ Dataset,Type,Source,Size (px),Resolution (m),License
 :ref:`GlobBiomass`,Masks,Landsat,"45,000x45,000",100,CC-BY-4.0
 :ref:`Global Mangrove Distribution`,Masks,"Remote Sensing, In Situ Measurements",-,3,public domain
 :ref:`GlobalBuildingMap`,Masks,PlanetScope,180K,3,CC-BY-4.0
+:ref:`GlobalMangroveWatch`,Masks,"Landsat, ALOS PALSAR","4,500x4,500",25,CC-BY-4.0
```
adamjstewart: Confirmed the license from https://zenodo.org/records/6894273

docs/api/datasets/global-mangrove-watch.rst
```diff
@@ -0,0 +1,8 @@
+.. _GlobalMangroveWatch:
+
+GlobalMangroveWatch
+===================
```
adamjstewart:
```suggestion
.. _Global Mangrove Watch:

Global Mangrove Watch
=====================
```

docs/api/datasets/masks.csv
adamjstewart:
```suggestion
:ref:`Global Mangrove Watch`,Masks,"Landsat, ALOS PALSAR","4,500x4,500",25,CC-BY-4.0
```

torchgeo/datasets/globalmangrovewatch.py
```diff
@@ -0,0 +1,289 @@
...
+from .errors import DatasetNotFoundError
+from .geo import RasterDataset
+from .utils import Path, Sample, check_integrity, download_url, extract_archive
+
+
+class GlobalMangroveWatch(RasterDataset):
+    """Global Mangrove Watch (GMW) dataset.
+
+    The `Global Mangrove Watch
+    <https://www.globalmangrovewatch.org>`_ Version 3.0 dataset provides maps of global
```
adamjstewart:
```suggestion
    <https://www.globalmangrovewatch.org>`__ Version 3.0 dataset provides maps of global
```

torchgeo/datasets/globalmangrovewatch.py
```diff
@@ -0,0 +1,289 @@
...
+    * Tiled as 1° x 1° GeoTIFF files of 4,500 x 4,500 pixels
+
+    Dataset format:
+
+    * single-channel uint8 GeoTIFFs, one ZIP archive per epoch
+    * 0 is used as the nodata value
+
+    Dataset classes:
+
+    0. Non-mangrove
```
adamjstewart: Is 0 nodata or non-mangrove? It can't be both

torchgeo/datasets/globalmangrovewatch.py
```diff
@@ -0,0 +1,289 @@
...
+        2010,
+        2015,
+        2016,
+        2017,
+        2018,
+        2019,
+        2020,
+    )
+
+    md5s: ClassVar[dict[int, str]] = {
```
adamjstewart: If you have these downloaded locally, can you convert from MD5 to SHA256? MD5 isn't cryptographically secure, so we're trying to update our datasets.

torchgeo/datasets/globalmangrovewatch.py
```diff
@@ -0,0 +1,289 @@
...
+        self.checksum = checksum
+
+        self._verify()
+
+        super().__init__(
+            paths, crs, res, transforms=transforms, cache=cache, time_series=time_series
+        )
+
+    @property
+    def _root(self) -> Path:
```
adamjstewart: Our other datasets don't have all of these private helper functions. For 1-line helper functions, it's actually significantly less code to just copy that 1 line everywhere. Can we reduce these somehow?

adamjstewart (commented):
Good start, just need to clean up a few things and make it simpler and it should be ready to go.

tests/datasets/test_globalmangrovewatch.py
akhellad: You're right, removed, plt.show is never called in these tests. Kept the matplotlib.pyplot import since it's still used for plt.close()

tests/datasets/test_globalmangrovewatch.py
akhellad: Trimmed from 19 tests down to 14. I looked at the other dataset tests to calibrate, the median across tests/datasets/ is 7, and the closest comparable datasets (test_cdl.py, test_nlcd.py) are both at 14, so I aimed for that.

Removed test_multiple_paths, test_partially_extracted and test_time_series_prediction (redundant with existing tests), and rather than keeping dedicated tests for the two GMW-specific behaviors, I folded the year-filtering check into test_already_extracted, i can  and the time-series check into test_plot. Still at 100% coverage for the module.

I can rename test_already_extracted if you'd prefer, it now covers year filtering too, so something like test_years might be a better fit.

torchgeo/datasets/globalmangrovewatch.py
akhellad: Applied all three. Also bumped the title underline to match the longer title, and moved the masks.csv row up one line, the file is sorted in strict ASCII order, and since a space sorts before letters, Global Mangrove Watch now belongs before GlobalBuildingMap.

torchgeo/datasets/globalmangrovewatch.py
akhellad: It's non-mangrove. The source GeoTIFFs do tag 0 as nodata, but GMW only distributes the mangrove extent, so everything else, background and gaps between tiles, is 0, and merge_datasets fills gaps with 0 too. So from the user's point of view there's only one meaning, and mentioning nodata just muddied it. Dropped that line (and reworded the cmap comment, which repeated the same confusion).

torchgeo/datasets/globalmangrovewatch.py
akhellad: Done. I re-downloaded all 11 archives, confirmed their MD5s matched what was already in the file, then computed SHA256 for each. Switched md5s → sha256s and updated the check_integrity / download_url calls plus the checksum docstring.

torchgeo/datasets/globalmangrovewatch.py
akhellad: Down to 2 private helpers (_verify, _year), nlcd, cdl, esri2020 and chesapeake all have 3 (_verify, _download, _extract), so this is now leaner than the datasets I modelled it on. I kept _year because it's called from both files and _verify and wraps a 2-line regex, so inlining it would add code rather than remove it.

adamjstewart (dismissed):
Looks great now, thanks!

adamjstewart (approved):

PR #3485, opened 2026-03-05
Sample: only support Tensor values

adamjstewart:
See #985 for motivation. With this, we're finally able to check the type of all Sample dictionaries. Only Tensors are supported, ensuring that PyTorch's default collate_fn and Lightning's default transfer_to_device functions work without changes.

Note that `dict[str, Tensor]` is still very open-ended, allowing for keys with any name, dtype, etc. We should really switch to TypedDict in the future and standardize key names as well.

Depends on #3484, #3483, #3481, #3458, #3788

Closes #3783

PR #3783, opened 2026-06-11
Sample: dict[str, Any] -> TypedDict

adamjstewart:
Latest ty release finally supports TypedDict `extra_items` coming in Python 3.15. This allows us to narrow our Sample type from `dict[str, Any]` to (almost) `dict[str, Tensor]` (only exception being caption strings). This is still looser than I would like, but we can try to standardize dict keys later, that isn't backwards-incompatible. See #985 for motivation.

Note that there are several instances of dynamic typing that ty isn't able to resolve, mostly when iterating over all key-value pairs or using dynamic string formatting. I'm not sure of a great way to handle this other than casting or ignoring. We could open an issue with ty to relax this check, but it's true that it isn't checkable.

Alternative is to drop support for str captions and first encode them to Tensors. That would make this PR ~5 lines of code like #3485. Of course, if we wanted to go from `dict[str, Tensor]` to a narrower type hint, this PR would become necessary once again.

Closes #3485
Related to #3278 (dropping one more deprecated TypeAlias)
Related to #3450 (documentation changed from `dict[str, Tensor]` to `Sample`)

## AI Usage Disclosure


adamjstewart:
> Alternative is to drop support for str captions and first encode them to Tensors.

I think this would use something like sentencepiece? I'm new to NLP.

calebrob6:
This is marked as draft, do you still need a review?

adamjstewart:
Nope, it's superseded by #3485 which is just waiting on #3788

calebrob6:
Cool, will close

PR #3595, opened 2026-04-07
Forest-Change dataset: Bi-temporal segmentation and captioning of deforestation scenes

JamesBrockUoB:
## Summary

This PR adds support for the Forest-Change image change interpretation dataset.

Published as part of the paper: https://www.sciencedirect.com/science/article/pii/S1574954126001470

Available at HF: https://huggingface.co/datasets/JimmyBrocko/Forest-Change

### Included
- `ForestChange` dataset class in `torchgeo/datasets/forestchange.py`
- import/export in `torchgeo/datasets/__init__.py`
- fake HDF5 test data generator in 'torchgeo/tests/data/forestchange/'
- zip folder and extracted example files in 'torchgeo/tests/data/forestchange/'
- dataset unit tests in 'torchgeo/tests/datasets/test_forestchange.py'
- docs stub for autodoc

## Dataset structure

This implementation follows the public ForestChange structure:

Train, validation, and test splits, each containing A/B image pairs with corresponding binary change masks. Each sample is provided with five captions describing observable changes.

## Validation

- added unit tests for split handling, sample shapes, mask presence/absence, caption token processing, preprocessing of original captions json file
- all introduced tests for forestchange pass locally running under pytest

PR #3788, opened 2026-06-12
SkyScript: tokenize caption

adamjstewart:
This is the only remaining dataset in TorchGeo that doesn't return `dict[str, Tensor]`. Str can't be moved to the GPU, so tokenization is needed at some point. This could probably be done in a datamodule, but we don't have a trainer for image captioning, so I put it in the dataset for now.

> [!WARNING]
> I'm an NLP novice, so please tell me if I'm doing anything dumb.

### Example

Caption: "a satellite image of roof shape of gabled, building part, building material of stone"
Encoding: [2468, 24129]
Successful decoding in the plot method:

skyscript

@JamesBrockUoB can you help review? We can use the same logic in #3595 to make it simpler.

## AI Usage Disclosure

### DIFF
--- a/torchgeo/datasets/skyscript.py
+++ b/torchgeo/datasets/skyscript.py
@@ -6,7 +6,7 @@
 import pathlib
 import textwrap
 from collections.abc import Callable
-from typing import ClassVar, Literal
+from typing import TYPE_CHECKING, ClassVar, Literal
 
 import numpy as np
 import pandas as pd
@@ -26,6 +26,9 @@ from .utils import (
     extract_archive,
     lazy_import,
 )
+
+if TYPE_CHECKING:
+    import tokenizers
 
 
 class SkyScript(NonGeoDataset):
@@ -79,6 +82,7 @@ class SkyScript(NonGeoDataset):
         transforms: Callable[[Sample], Sample] | None = None,
         download: bool = False,
         checksum: bool = False,
+        tokenizer: 'tokenizers.models.Model | None' = None,
     ) -> None:
         """Initialize a new SkyScript instance.
 
@@ -89,14 +93,18 @@ class SkyScript(NonGeoDataset):
                 entry and returns a transformed version.
             download: If True, download dataset and store it in the root directory.
             checksum: If True, check the MD5 of the downloaded files (may be slow).
+            tokenizer: A pre-trained tokenizer
+                (defaults to :class:`~tokenizers.models.BPE`).
 
         Raises:
             AssertionError: If *split* is invalid.
             DatasetNotFoundError: If dataset is not found and *download* is False.
             DependencyNotFoundError: If tokenizers is not installed.
+
+        .. versionadded:: 0.10
+           The *tokenizer* parameter.
         """
         assert split in self.caption_files
-        tokenizers = lazy_import('tokenizers')
 
         self.root = pathlib.Path(root)
         self.split = split
@@ -107,11 +115,15 @@ class SkyScript(NonGeoDataset):
         self._verify()
 
         self.captions = pd.read_csv(self.root / self.caption_files[split])
-        train_captions = pd.read_csv(self.root / self.caption_files['train'])
 
-        self.tokenizer = tokenizers.Tokenizer(tokenizers.models.BPE())
-        trainer = tokenizers.trainers.BpeTrainer()
-        self.tokenizer.train_from_iterator(train_captions['title'], trainer)
+        if tokenizer is None:
+            tokenizers = lazy_import('tokenizers')
+            self.tokenizer = tokenizers.Tokenizer(tokenizers.models.BPE())
+            trainer = tokenizers.trainers.BpeTrainer()
+            train_captions = pd.read_csv(self.root / self.caption_files['train'])
+            self.tokenizer.train_from_iterator(train_captions['title'], trainer)
+        else:
+            self.tokenizer = tokenizer
 
     def __len__(self) -> int:
         """Return the number of images in the dataset.
--- a/torchgeo/datasets/skyscript.py
+++ b/torchgeo/datasets/skyscript.py
@@ -3,9 +3,9 @@
 
 """SkyScript dataset."""
 
-import os
+import pathlib
 from collections.abc import Callable
-from typing import ClassVar, Literal, NotRequired, TypedDict
+from typing import ClassVar, Literal
 
 import numpy as np
 import pandas as pd
@@ -14,19 +14,17 @@ from einops import rearrange
 from matplotlib import pyplot as plt
 from matplotlib.figure import Figure
 from PIL import Image
-from torch import Tensor
 
 from .errors import DatasetNotFoundError
 from .geo import NonGeoDataset
-from .utils import Path, download_and_extract_archive, download_url, extract_archive
-
-
-class CaptionSample(TypedDict):
-    """Sample for image captioning."""
-
-    image: Tensor
-    caption: str
-    prediction: NotRequired[str]
+from .utils import (
+    Path,
+    Sample,
+    download_and_extract_archive,
+    download_url,
+    extract_archive,
+    lazy_import,
+)
 
 
 class SkyScript(NonGeoDataset):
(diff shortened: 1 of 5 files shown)


torchgeo/datasets/skyscript.py
```diff
@@ -89,18 +92,25 @@ def __init__(
...
         self.download = download
         self.checksum = checksum
 
         self._verify()
 
-        self.captions = pd.read_csv(os.path.join(self.root, self.caption_files[split]))
+        self.captions = pd.read_csv(self.root / self.caption_files[split])
+        train_captions = pd.read_csv(self.root / self.caption_files['train'])
+
+        self.tokenizer = tokenizers.Tokenizer(tokenizers.models.BPE())
```
adamjstewart: The original SkyScript paper also used BPE: https://github.com/wangzhecheng/SkyScript/blob/main/src/open_clip/tokenizer.py

Not sure if we need any other custom normalization/pre-tokenization.

adamjstewart:
> Caption: "a satellite image of roof shape of gabled, building part, building material of stone"
> Encoding: [2468, 24129]

Is it normal that the encoded values are so large and the length is so short? Do we need to force it to break more often or use a smaller vocabulary size? I think the captions are auto-generated, so likely not very diverse and always follow the same structure.

isaaccorley:
> > Caption: "a satellite image of roof shape of gabled, building part, building material of stone"
> > Encoding: [2468, 24129]
> 
> Is it normal that the encoded values are so large and the length is so short? Do we need to force it to break more often or use a smaller vocabulary size? I think the captions are auto-generated, so likely not very diverse and always follow the same structure.

Don't change the dictionary size. It typically gets fed to a pretrained text encoder that is able to interpret the values via an `nn.Embedding` layer. If you change it then the values won't map.

Regarding the length, we will need to make the tokenizer padding or truncation params configurable which matters these auto-generated captions are short so 1) you want less tokens to reduce compute and 2) you want tokens padded to the same length so the stack into a batched tensor.

adamjstewart:
We do have `pad_across_batches`, is that sufficient?

DarthReca:
I'd suggest exposing a tokenizer parameter in __init__ (defaulting to None, returning the raw caption string). Users can then pass any pretrained tokenizer that fits their model (e.g., CLIPTokenizer for CLIP-based models, or the Gemma tokenizer for SigLIP-based, etc). This keeps the dataset model-agnostic to a specific backbone.

adamjstewart:
> I'd suggest exposing a tokenizer parameter in init (defaulting to None, returning the raw caption string).

Happy to add a tokenizer parameter, but definitely can't default to returning the raw caption string, as that's exactly the thing I'm trying to remove here.

DarthReca:
> > I'd suggest exposing a tokenizer parameter in init (defaulting to None, returning the raw caption string).
> 
> Happy to add a tokenizer parameter, but definitely can't default to returning the raw caption string, as that's exactly the thing I'm trying to remove here.

So I think it is necessary to force the user to provide their own preprocessor. We can even include a default one, but I am not sure it is a real gain actually. For example, if you look at [CLIP tokenizer](https://github.com/openai/CLIP/blob/main/clip/simple_tokenizer.py) they do some cleaning before tokenization.

adamjstewart:
I would like to default to the closest thing we can get to the CLIP tokenizer so we can match the SkyScript paper. I attempted to do that in this PR, although I couldn't find all of the cleaning functions. Your proposal is just to make this configurable, which I'm okay with.

adamjstewart:
Struggling a bit with this. I tried something like
```diff
diff --git a/torchgeo/datasets/skyscript.py b/torchgeo/datasets/skyscript.py
index 916e070c6..a6772a315 100644
--- a/torchgeo/datasets/skyscript.py
+++ b/torchgeo/datasets/skyscript.py
@@ -79,6 +79,7 @@ class SkyScript(NonGeoDataset):
         transforms: Callable[[Sample], Sample] | None = None,
         download: bool = False,
         checksum: bool = False,
+        tokenizer: 'tokenizers.models.Model | None' = None,
     ) -> None:
         """Initialize a new SkyScript instance.
 
@@ -89,6 +90,7 @@ class SkyScript(NonGeoDataset):
                 entry and returns a transformed version.
             download: If True, download dataset and store it in the root directory.
             checksum: If True, check the MD5 of the downloaded files (may be slow).
+            tokenizer: A pre-trained tokenizer (defaults to BPE).
 
         Raises:
             AssertionError: If *split* is invalid.
@@ -96,7 +98,6 @@ class SkyScript(NonGeoDataset):
             DependencyNotFoundError: If tokenizers is not installed.
         """
         assert split in self.caption_files
-        tokenizers = lazy_import('tokenizers')
 
         self.root = pathlib.Path(root)
         self.split = split
@@ -107,11 +108,15 @@ class SkyScript(NonGeoDataset):
         self._verify()
 
         self.captions = pd.read_csv(self.root / self.caption_files[split])
-        train_captions = pd.read_csv(self.root / self.caption_files['train'])
 
-        self.tokenizer = tokenizers.Tokenizer(tokenizers.models.BPE())
-        trainer = tokenizers.trainers.BpeTrainer()
-        self.tokenizer.train_from_iterator(train_captions['title'], trainer)
+        if tokenizer is None:
+            tokenizers = lazy_import('tokenizers')
+            self.tokenizer = tokenizers.Tokenizer(tokenizers.models.BPE())
+            trainer = tokenizers.trainers.BpeTrainer()
+            train_captions = pd.read_csv(self.root / self.caption_files['train'])
+            self.tokenizer.train_from_iterator(train_captions['title'], trainer)
+        else:
+            self.tokenizer = tokenizer
 
     def __len__(self) -> int:
         """Return the number of images in the dataset.
```
The problem is that there's no way to add proper type hints for `tokenizer` because tokenizers isn't a required dependency. Best workaround I can think of is to ignore ruff/ty complaints, but that doesn't sound like a good idea. I would rather stick to hard-coding BPE if necessary. Any other options I'm missing?

adamjstewart:
Technically a user could do:
```python
ds = SkyScript(...)
ds.tokenizer = MyPretrainedTokenizer
ds[0]
```
Is that a sufficient amount of customizability?

DarthReca:
What about simply assuming the tokenizer is a callable[str, np.ndarray] (or another output you like)? So you do not need any extra deps, and the user can provide any implementation he prefers.

adamjstewart:
Technically it would have to be a Protocol with encode and decode methods, but yeah that's one workaround. Let me also check if a try-except import would help.

nilsleh:
What about just making the tokenizer arg an nn.Module?

adamjstewart:
That would be quite underspecified, as most `nn.Module`s don't have `.encode()` and `.decode()` methods.  We could use [typing.Protocol](https://docs.python.org/3/library/typing.html#typing.Protocol), but I can't imagine anyone not wanting to use tokenizers.

nilsleh:
what if you make a base class nn.Module with encode and decode that would also satisfy tokenizer type hints. I guess this is the only blocking question, that you want to make type hints happy and avoiding the dependency?

adamjstewart:
I think we would just end up reinventing the wheel if we try to make our own tokenizers. If we only make an ABC, then it would end up looking quite similar to a Protocol. If we make subclasses that actually work, we would end up with tokenizers. Since it's also going to be used in #3595, I don't mind adding yet another optional dependency.

torchgeo/datasets/skyscript.py
```diff
@@ -120,13 +143,14 @@ def __getitem__(self, index: int) -> CaptionSample:  # ty: ignore[invalid-method
             A dict containing image and caption at index.
```
nilsleh: should the doc string say that caption is already tokenized?

nilsleh (commented):
Otherwise this makes sense to me. Depending on how many vision language dataset there will be there could eventually be a subclass or so that inherits to deduplicate the tokenizer logic, but certainly out of scope for this.

isaaccorley (approved):
This lgtm although you may want to test it in a torch DataLoader because if the tokenizer doesn't pad or truncate then each sample token will be different lengths and non-stackable to a batch tensor.

adamjstewart:
We have a padding utility for that. We can revisit once someone adds a dm/task.

Issue #3774, opened 2026-06-06
Plotting Mixin


adamjstewart:
### Summary

I propose introducing a shared mixin for everything related to dataset plotting. This includes both metadata attributes and the actual `plot()` method.

### Rationale

One of the largest sources of code duplication in TorchGeo is in our dataset plotting. There is currently limited consistency in how datasets get plotted, but 90% of datasets involve the following steps:

1. If 'image' key:
  a. Select `rgb_bands` out of `all_bands`
  b. Raise an error if RGB bands not found
  c. Apply normalization (either division or quantile normalization)
  d. Use imshow to plot the image
2. If 'mask' key:
  a. Use imshow to plot the mask (with cmap, vmin, vmax)
  b. Same for predicted mask if available
3. If 'label' key:
  a. Becomes image title
  b. Same for predicted label if available
4. Always:
  a. Remove axes labels
  b. Apply suptitle if requested

As our Sample keys become more consistent, so to can our plotting. I also have a lot of clever ideas for better semantic segmentation plots (#1467) and time-series plots (an animated GIF). These are quite challenging at the moment without repeating the code in dozens of places, but could be made more consistent using a mixin.

### Implementation

I propose a shared [mixin](https://en.wikipedia.org/wiki/Mixin) for all things related to visualization. This will be shared by both NonGeoDataset and GeoDataset to avoid additional code duplication. This mixin will define all dataset attributes (e.g., `all_bands`, `rgb_bands`, `cmap`), helper functions (e.g., `quantile_normalization`), and public functions (e.g., `plot`) related to plotting. We can start small and port a few datasets at a time, but the eventual goal is to remove the `plot` method from 90% of datasets. About 10% of datasets will still require a custom `plot` method to deal with intricacies of the data (e.g., multi-modal datasets), but that's okay.

In order to make things easier to review, I suggest we proceed in several smaller PRs:

- [x] Add a PlottingMixin class: #3775
- [x] Migrate `rgb_bands` and `all_bands: tuple[str, ...] = ()`: #3789
- [x] Migrate `cmap: str | Colormap | None = None`: #3831
- [ ] Migrate `quantile_normalization`
- [ ] Add `rgb_min`, `rgb_max`, `rgb_mean`, and `rgb_std` for normalization?
- [ ] Add basic `plot` method: will close #1971, #2377
- [ ] Convert most classification datasets: image/label
- [ ] Convert most semantic segmentation datasets: mask
[...]

PR #3789, opened 2026-06-12
PlottingMixin: migrate bands attributes

adamjstewart:
Second step of #3774

Only one dataset didn't align with this, so easy change.

## AI Usage Disclosure

### DIFF
--- a/tests/datasets/test_copernicus.py
+++ b/tests/datasets/test_copernicus.py
@@ -110,8 +110,8 @@ class TestCopernicusBench:
         plt.close()
 
     def test_not_rgb(self, dataset: CopernicusBench) -> None:
-        all_bands = list(dataset.all_bands)
-        rgb_bands = list(dataset.rgb_bands)
+        all_bands = list(dataset.dataset.all_bands)
+        rgb_bands = list(dataset.dataset.rgb_bands)
         for band in rgb_bands:
             all_bands.remove(band)
 
--- a/torchgeo/datasets/treesatai.py
+++ b/torchgeo/datasets/treesatai.py
@@ -99,7 +99,7 @@ class TreeSatAI(NonGeoDataset):
 
     # https://zenodo.org/records/6780578/files/220629_doc_TreeSatAI_benchmark_archive.pdf
     all_sensors: tuple[Literal['aerial', 's1', 's2'], ...] = ('aerial', 's1', 's2')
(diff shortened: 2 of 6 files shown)


tests/datasets/test_copernicus.py
```diff
@@ -110,8 +110,8 @@ def test_plot(self, dataset: CopernicusBench) -> None:
         plt.close()
 
     def test_not_rgb(self, dataset: CopernicusBench) -> None:
-        all_bands = list(dataset.all_bands)
-        rgb_bands = list(dataset.rgb_bands)
+        all_bands = list(dataset.dataset.all_bands)
```
robmarkcole, line 113: This always catches me out..

adamjstewart: Yeah, these dataset wrappers are kinda weird. FWIW, I did propose attribute forwarding for Subset but they didn't like it: https://github.com/pytorch/pytorch/pull/149272

torchgeo/datasets/treesatai.py
```diff
@@ -117,7 +117,7 @@ class TreeSatAI(NonGeoDataset):
             'B09',
         ],
     }
-    rgb_bands: ClassVar[dict[str, list[str]]] = {
+    rgb_bands_dict: ClassVar[dict[str, list[str]]] = {
```
robmarkcole: Why do we use list instead of tuple?

adamjstewart: No reason. Only difference is mutability. I think everything was originally a list, but ruff doesn't like that as modifying one instance modifies all instances, so I switched most to tuple and used ClassVar for the rest because there was no frozendict option. Happy to change it to tuple to make it more consistent.

robmarkcole: I vote for adopting tuples - consistency benefits + bands shouldn't change

adamjstewart: I was able to do this for SSL4EO and TreeSatAI. Couldn't do this for MMEarth due to Python dict being [invariant](https://mypy.readthedocs.io/en/latest/common_issues.html#invariance-vs-covariance), it's a lot more complicated. There are a lot more `ClassVar[dict[str, list]]` in TorchGeo, but they aren't related to bands, so didn't include them. Trying to keep this PR simple and easy to review.

robmarkcole (approved):

PR #3795, opened 2026-06-13
AgriFieldNet: RasterDataset -> IntersectionDataset

adamjstewart:
Greatly simplifies the code and puts it more inline with other GeoDatasets containing both images and masks.

Plotting still works as expected:

agrifieldnet

## AI Usage Disclosure

### DIFF
--- a/tests/datasets/test_agrifieldnet.py
+++ b/tests/datasets/test_agrifieldnet.py
@@ -38,7 +38,7 @@ class TestAgriFieldNet:
         assert isinstance(x['mask'], torch.Tensor)
 
     def test_len(self, dataset: AgriFieldNet) -> None:
-        assert len(dataset) == 50
+        assert len(dataset) == 10
 
     def test_and(self, dataset: AgriFieldNet) -> None:
         ds = dataset & dataset
--- a/tests/datasets/test_agrifieldnet.py
+++ b/tests/datasets/test_agrifieldnet.py
@@ -38,7 +38,7 @@ class TestAgriFieldNet:
         assert isinstance(x['mask'], torch.Tensor)
 
     def test_len(self, dataset: AgriFieldNet) -> None:
-        assert len(dataset) == 10
+        assert len(dataset) == 50
 
     def test_and(self, dataset: AgriFieldNet) -> None:
         ds = dataset & dataset
--- a/torchgeo/datasets/agrifieldnet.py
+++ b/torchgeo/datasets/agrifieldnet.py
@@ -321,7 +321,7 @@ class AgriFieldNet(IntersectionDataset):
 
         fig, axs = plt.subplots(nrows=1, ncols=ncols, figsize=(ncols * 4, 4))
         kwargs = {
-            'cmap': self.cmap,
+            'cmap': self.mask.cmap,
             'vmin': self.mask.valid_classes[0],
             'vmax': self.mask.valid_classes[-1],
             'interpolation': 'none',
--- a/torchgeo/datasets/agrifieldnet.py
+++ b/torchgeo/datasets/agrifieldnet.py
@@ -238,6 +238,9 @@ class AgriFieldNet(IntersectionDataset):
 
         super().__init__(self.image, self.mask)
 
+        # Ignore unintentional partial overlap
+        #self.index = self.image.index
+
         # Map chosen classes to ordinal numbers, all others mapped to background class
         self.ordinal_map = torch.zeros(self.mask.valid_classes[-1] + 1, dtype=self.mask.dtype)
         self.inverse_map = torch.zeros(len(classes), dtype=self.mask.dtype)
--- a/torchgeo/datasets/agrifieldnet.py
+++ b/torchgeo/datasets/agrifieldnet.py
@@ -239,7 +239,7 @@ class AgriFieldNet(IntersectionDataset):
         super().__init__(self.image, self.mask)
 
         # Ignore unintentional partial overlap
-        #self.index = self.image.index
+        self.index = self.image.index
 
         # Map chosen classes to ordinal numbers, all others mapped to background class
         self.ordinal_map = torch.zeros(self.mask.valid_classes[-1] + 1, dtype=self.mask.dtype)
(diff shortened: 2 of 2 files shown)


torchgeo/datasets/agrifieldnet.py
```diff
@@ -196,59 +225,8 @@ def __getitem__(self, index: GeoSlice) -> Sample:
...
-        image = torch.cat(data_list, dim=-3)
-
-        mask_filepaths = []
-        for root, dirs, files in os.walk(os.path.join(paths, 'train_labels')):
-            for file in files:
-                if not file.endswith('_field_ids.tif') and file.endswith('.tif'):
-                    file_path = os.path.join(root, file)
-                    mask_filepaths.append(file_path)
-
-        mask = self._merge_or_stack(mask_filepaths, index)
```
adamjstewart, line 270: @yichiac this code wasn't working for me, and I'm not sure if it ever worked. We're basically loading and merging every single mask file at once, which leads to:
```
  File "rasterio/_base.pyx", line 320, in rasterio._base.DatasetBase.__init__
  File "rasterio/_base.pyx", line 232, in rasterio._base.open_dataset
  File "rasterio/_err.pyx", line 359, in rasterio._err.exc_wrap_pointer
rasterio._err.CPLE_OpenFailedError: data/agrifieldnet/train_labels/ref_agrifieldnet_competition_v1_labels_train_38b06.tif: Too many open files
```

tests/datasets/test_agrifieldnet.py
```diff
@@ -38,7 +38,7 @@ def test_getitem(self, dataset: AgriFieldNet) -> None:
         assert isinstance(x['mask'], torch.Tensor)
 
     def test_len(self, dataset: AgriFieldNet) -> None:
-        assert len(dataset) == 10
+        assert len(dataset) == 50
```
calebrob6: So I understand, is this now reporting the number of test files vs. previous patches? (kinda spooking seeing a test assert change and passing the tests but the underlying data its operating on not changing)

calebrob6:
I'd be more comfortable if @yichiac can confirm that this gives identical training results?

tests/datasets/test_agrifieldnet.py
adamjstewart: We were previously reporting the length of `images`, we are now reporting the length of `images & masks`. Because our _test_ data happens to have all images and masks in the same location, the length has increased.

More importantly, the total length of the _real_ dataset has also changed:

* `len(self.image)`: 1217 (previous dataset length)
* `len(self.mask)`: 1165
* `len(self)`: 3677 (new dataset length)

I don't know why there are more images than masks. Typically I resolve this problem by copying `self.image.index` to `self.index`, but I don't know if I should grab the index from `self.image` or `self.mask`. I guess it depends on if any images are missing masks and/or vice versa.

adamjstewart: It seems that every image has a mask, so there must be a few images with the same location but different timestamp and same mask. I grabbed the index from self.image to preserve backwards compatibility. Behavior should now be identical to main, without the bug.

adamjstewart:
@yichiac is defending his thesis on Monday, so definitely won't see this before then.

yichiac:
Thanks @adamjstewart and @calebrob6. Sorry for the delay. 

Regarding the error:
`rasterio._err.CPLE_OpenFailedError: data/agrifieldnet/train_labels/ref_agrifieldnet_competition_v1_labels_train_38b06.tif: Too many open files` 
I will re-run the exp to see if I had the same issue. We did run lots of experiments with Agrifieldnet without issues then using high-memory (128GB). 

My plan is: 
1. Examine sample-level images and masks to verify they're identical between the old and new datasets.
2. The image and mask datasets having different lengths seems odd. I'll look into that as well.
3. I will re-run exp to see whether we have identical training results.

Agrifieldnet's labels are very sparse (shown in Adam's demo picture in this PR). Labeled pixels are a tiny fraction of the total image pixels, so training is noisy.

nilsleh (dismissed):

yichiac:
> Regarding the error: `rasterio._err.CPLE_OpenFailedError: data/agrifieldnet/train_labels/ref_agrifieldnet_competition_v1_labels_train_38b06.tif: Too many open files` I will re-run the exp to see if I had the same issue. We did run lots of experiments with Agrifieldnet without issues then using high-memory (128GB).
> 
The error was caused by a lower ulimit. On macOS, I can reproduce the error, but the compute node with a higher ulimit can avoid it. Anyway, this PR has fixed this error.

yichiac:
> 1. Examine sample-level images and masks to verify they're identical between the old and new datasets.

Images are bit-identical; masks are identical except at tile seams. This PR is more correct: AgriFieldNetMask is is_image=False, so labels are warped with nearest resampling, whereas main loaded labels through the float image dataset and bilinearly interpolated class IDs at seams.

tests/datasets/test_agrifieldnet.py
yichiac: > It seems that every image has a mask, so there must be a few images with the same location but different timestamp and same mask. 

The extra length isn't due to timestamps (all files share the same time interval). The 52 "missing" masks are chips containing only test fields (test labels were withheld; test_labels/contains only field IDs). The AgriFieldNet competition didn't release the actual test labels.

yichiac: So there are 52 images without their own masks.

yichiac:
> 2. The image and mask datasets having different lengths seems odd. I'll look into that as well.

> len(self.image): 1217 (previous dataset length)
> len(self.mask): 1165
> len(self): 3677 (new dataset length)

Can confirm the lengths are correct. Just a small thing: the PR sets self.index = self.image.index (1,217 tiles), so geo-samplers will also draw patches inside those 52 label-less (test) chips. But that’s not an issue because the test labels are not publicly available, so there’s no label leakage during training.

yichiac:
> 3. I will re-run exp to see whether we have identical training results.

I went back to check our previous experiments. We used self-prepared Sentinel-2 images instead of the original images, so the results are not comparable to those of the main branch.

calebrob6 (approved):
I just verified that this works. Given @yichiac 's confirmation also, I think this is good to go.

PR #3804, opened 2026-06-14
Decouple index CRS and read crs (out_crs)

adriantre:
GeoDatasets currently enforce a single CRS across the dataset.

The motivation for considering other options is that reprojection can be expensive, lossy, and conflict with inference-time use, where the native CRS is preferable.

This PR is the first step, as it decouples the two roles `self.crs` currently has 
- the index CRS, and 
- the output (warp/reproject) CRS

by making the output CRS a per-read parameter `out_crs`, clarifying the distinct scoped role from `self.crs`. This makes it easier to experiment with per-sample CRS policies later.

No behaviour changes. `out_crs` defaults to `self.crs` in both RasterDataset and XarrayDataset. `_cached_load_warp_file` now keys on out_crs as well as the filepath, so a varying warp target can't return a VRT cached in the wrong CRS.

## AI Usage Disclosure

While the code is simple, I discussed with AI to make the PR as atomic as possible, due to my history of long-running PRs.


adamjstewart:
@evgeniialappo is also working on this right now.

adriantre:
@evgeniialappo feel free to tag me in your efforts if you want to discuss. Would you consider this worth a merge in the meantime?

adamjstewart:
We definitely don't need to wait for @evgeniialappo, but let's discuss long-term design first.

There are (at least) 3 different kinds of "CRS":

1. src_crs: the original CRS of each file
2. index_crs: the CRS of our spatiotemporal `self.index`
3. out_crs: the CRS of output samples

Currently, 2 = 3 = `self.crs`, which defaults to the first 1 found. This PR allows 2 and 3 to be different.

What @evgeniialappo is working on for her MS thesis is approximately as follows:

1. src_crs: no changes
2. index_crs: (always?) default to EPSG:4326
3. out_crs: allow each sample to have a different CRS, default to statistical mode of each sample bbox

This allows us to avoid unnecessary reprojection: #409.

My concern is that this PR would introduce a new dataset-level `out_crs`, but @evgeniialappo's PR would introduce a new getitem-level `out_crs`. This would make the dataset-level `out_crs` obsolete, and we wouldn't be able to remove it without breaking backwards compatibility. So I'm hesitant to add a new dataset-level parameter until we finalize the long-term design.

adriantre:
What you describe is exactly what this change sets up. The concept of out_crs is a getitem-only concept already in this PR. It clarifies the responsibility/scope Without breaking existing behaviour.

Next step is that getitem gets out_crs as input :)

adamjstewart:
Ah you're right, I misread this. I thought it was going in `__init__`. Seems like GitHub dropped the function name from the diff minimization. So then yes, this is probably a step in the right direction. Will wait for @nilsleh and @evgeniialappo and others to review before taking a closer look.

evgeniialappo:
I think it can make the implementation of further changes easier, including the majority voting idea

adriantre:
That's great to hear!

torchgeo/datasets/geo.py
```diff
@@ -1128,20 +1147,22 @@ def __getitem__(self, index: GeoSlice) -> Sample:
...
         shapes = []
         for filepath in df.filepath:
             if pathlib.Path(filepath).suffix.lower() == '.parquet':
                 src = gpd.read_parquet(filepath)
             else:
                 src = gpd.read_file(filepath, layer=self.layer)
 
-            # We need to know the bounding box of the index in the source CRS
-            transformer = pyproj.Transformer.from_crs(self.crs, src.crs, always_xy=True)
+            # We need to know the bounding box of the query in the source CRS
```
adriantre, line 1167: index = how we found what to sample.
query = what to sample. Will differ from index once it reads from native crs, where the box is aligned with a different grid than index.

adamjstewart: I agree we should more carefully distinguish these, although there is also "spectral index" to add to the list of conflated terms. Also note that we are forced to use `__getitem__(index):` due to type hint reasons, so we can't change that name to something less ambiguous. Want to add something to https://docs.torchgeo.org/en/stable/user/glossary.html in a separate PR?

adriantre: Can do! Leaving this unresolved as a reminder to myself.

adriantre: Done, PR on its way

nilsleh (commented):
This makes sense, would be interesting to see what kind of speed ups are enabled with this.

adriantre:
> This makes sense, would be interesting to see what kind of speed ups are enabled with this.

One of the PRs will be a benchmark on a synthetic global dataset with overlapping scenes of neighbouring UTM-zones. This is where the big speedup is obvious, as single-CRS datasets will be unaffected.

adriantre:
@nilsleh, for a global dataset where the main branch has to warp everything to the same CRS, my benchmark shows:

Case 1: No overlapping tiles:
55% faster data loading than on the main branch.
Average number of warps goes from 1.39 (main) → 0.00 (my branch)

Case 2: All tiles overlap 50% with another tile of neighbouring CRS:
7-13% faster dataloading than on main branch on different runs and latitudes.
Average number of warps goes from 1.40 (main) → 0.40 (my branch)
(Average, because not all samples overlap with multiple files.) 

In practice, Case 1 should occur more often than samples being merged from multiple files.

(This is on a branch with all required changes, not this one ofc)

adriantre:
If we merge this tomorrow, I will open the next PR towards native-crs reads before I leave for the holiday 🙌

adamjstewart (dismissed):
I think this looks good to me, can you resolve the merge conflicts?

This is currently only done for base classes. There are many subclasses with `__getitem__` as well. They may need the same logic.

adriantre:
`SouthAfricaCropType`, `LandCoverAIGeo` and `GlobBiomass`.

These three datasets call `_merge_or_stack` in their `__getitem__`.
Until now, I made the `out_crs` param of `_merge_or_stack` optional, defaulting to self.crs inside of it (too).

I _want_ to make `out_crs` non-optional, but then I would have to change the order of the params. If `band_indexes` were keyword-only, it would be fine, but it is not. 

What do you think? If we keep it optional, I don't think these datasets need to be updated.

adamjstewart:
Let's keep it as is for now and break things if/when we add `crs` to the `__getitem__` signature.

adamjstewart (approved):

adriantre:
Agreed! On it

PR #3807, opened 2026-06-15
ChesapeakeCVPR as a pure raster dataset

evgeniialappo:
This PR makes ChesapeakeCVPR a pure raster dataset. 
Previously, the dataset had a hybrid logic. 
Here it acts as a raster dataset and should perform reprojection if necessary.  

It is related to issue #3684

## AI Usage Disclosure


calebrob6:
Hi @evgeniialappo -- this is a big change! While you are developing on this can you verify that training with the real DE and MD data gets through a single epoch of training with the SemanticSegmentationTrainer please? I want to ensure that the changes here still let people train with the actual dataset.

Also, can you use your new IOBench to benchmark a training dataloader pass before and after this PR?

evgeniialappo:
Hello @calebrob6, I just checked that real data works with the SemanticSegmentationTrainer. And these are the benchmark results. I think they look interesting.

**Pure Raster Version:** 
Profile Summary

| Action              |      Time (sec) |   Rate (patches/sec) |
| ------------------- | --------------: | -------------------: |
| Training (random)   |       428.33440 |             22.56181 |
| Validation (grid)   |         9.81376 |             71.73599 |

Total duration: 443.18733 seconds
**Hybrid Version:**
Profile Summary

| Action              |      Time (sec) |   Rate (patches/sec) |
| ------------------- | --------------: | -------------------: |
| Training (random)   |       454.17715 |             35.22854 |
| Validation (grid)   |        24.77674 |             43.91214 |

Total duration: 485.22811 seconds

calebrob6:
Is this still draft?

>  I think they look interesting.

I think they look great!

Can I bother you to paste the output of `.plot()` here too? Basically I just want to ensure that training works with real data before we merge.

adamjstewart:
Also need to resolve merge conflicts, I changed some of the plotting cmap logic. Let me know if you need help with that, rebase can be less than fun if it's your first time.

evgeniialappo:
Here is the output of `.plot()`. It is done after my rebase attempt. 
plot_after_rebase

calebrob6:
LGTM!

tests/datasets/test_chesapeake.py
```diff
@@ -221,3 +212,34 @@ def test_partially_out_of_raster_query(self, dataset: ChesapeakeCVPR) -> None:
         # least one column on the right should be zero-filled.
         assert torch.all(sample['image'][..., -1] == 0)
         assert torch.all(sample['mask'][..., -1] == 0)
+
+    def test_dataset_splits(self, monkeypatch: MonkeyPatch, tmp_path: Path) -> None:
```
adamjstewart: What is this test needed for? Can we at least move the monkeypatching to the fixture so it isn't duplicated? Or remove it if it doesn't add much.

torchgeo/datasets/chesapeake.py
```diff
@@ -531,20 +523,10 @@ class ChesapeakeCVPR(GeoDataset):
...
-        'epsg:26918': pyproj.Transformer.from_crs(
-            p_src_crs, pyproj.CRS('epsg:26918'), always_xy=True
-        ),
-    }
-
     def __init__(
         self,
         root: Path = 'data',
-        splits: Sequence[str] = ['de-train'],
+        splits: Sequence[str] = ['de-test'],
```
adamjstewart: Why change the default?

torchgeo/datasets/chesapeake.py
```diff
@@ -817,3 +739,27 @@ def plot(
         if suptitle is not None:
             plt.suptitle(suptitle)
         return fig
+
+
+class ChesapeakeCVPRHelperClass(RasterDataset):
+    """This is a helper class for the ChesapeakeCVPR dataset."""
+
+    def __init__(self, paths: str, layer: str, **kwargs: Any) -> None:
+        """Initialise helper class."""
```
adamjstewart: Need to document parameters. I would also forward *args, not just **kwargs.

torchgeo/datasets/chesapeake.py
```diff
@@ -817,3 +739,27 @@ def plot(
...
+
+        self.is_image = layer in [
+            'naip-new',
+            'naip-old',
+            'landsat-leaf-on',
+            'landsat-leaf-off',
+        ]
+        super().__init__(paths, **kwargs)
+
+    def __getitem__(self, index: GeoSlice) -> dict:
```
adamjstewart: Why is this needed?

torchgeo/datasets/chesapeake.py
```diff
@@ -583,21 +565,28 @@ def __init__(
...
+                split_dataset.index = split_dataset.index[['geometry']]
+            split_datasets.append(split_dataset)
+        self.datasets = split_datasets[0]
+        for dataset in split_datasets[1:]:
+            self.datasets |= dataset
+        self.index = self.datasets.index
+        self.crs = self.datasets.crs
+        self.res = self.datasets.res
 
     def __getitem__(self, index: GeoSlice) -> Sample:
```
adamjstewart: Wonder if we can delete this and use the RasterDataset getitem only

torchgeo/datasets/chesapeake.py
```diff
@@ -817,3 +739,27 @@ def plot(
         if suptitle is not None:
             plt.suptitle(suptitle)
         return fig
+
+
+class ChesapeakeCVPRHelperClass(RasterDataset):
```
adamjstewart: Maybe remove "Class" from the name and move this above CheseapeakeCVPR so that as you read from top to bottom you see this beforehand?

torchgeo/datasets/chesapeake.py
```diff
@@ -583,21 +565,28 @@ def __init__(
...
+            for layer in self.layers[1:]:
+                new_part = ChesapeakeCVPRHelperClass(
+                    paths=state_dir, layer=layer, cache=self.cache
+                )
+                split_dataset &= new_part
+                split_dataset.index = split_dataset.index[['geometry']]
+            split_datasets.append(split_dataset)
+        self.datasets = split_datasets[0]
+        for dataset in split_datasets[1:]:
+            self.datasets |= dataset
```
adamjstewart: For these kinds of multi-part datasets, I like to explicitly subclass IntersectionDataset or UnionDataset. For example, see AgriFieldNet, IOBench, L7Irish, L8Biome, and MMFlood. Not a huge difference, but let's you skip the 3 lines below.

adamjstewart (changes requested):
Definitely a huge improvement to speed, compatibility, and maintainability! Just a few minor suggestions to make things even simpler.

tests/datasets/test_chesapeake.py
adamjstewart, line 91: Let's expand our parameterization. We should test the cross product of:

1. state
2. states

and:

1. split
2. splits

and:

0. images
1. image
2. images

and:

0. masks
1. mask
2. masks

My concern is that the code works fine for 0 or 2+ images/masks but doesn't work for exactly 1 image/mask.

You can use `itertools.product` to get a cross product. You may want to exclude the product of 0 images and 0 masks as that doesn't make sense. `request.param` will become a tuple of strings instead of a single string. You may need to update `tests/data/chesapeake/cvpr/` to add more fake test data. You can write a `data.py` script to generate this, see `tests/data/*/data.py` for examples. The goal is to mimic the real data, so if the real data has different projections, resolutions, etc. you should copy that.

calebrob6:
I think it will be necessary to update the datamodule here too as the current datamodule will break immediately after this is merged.

adamjstewart:
Oh yeah, thanks for the reminder. We just need to drop all the resize/crop logic I think?

calebrob6:
Is this ready to merge (I'm auditing PRs)

evgeniialappo:
Almost, it should pass the tests after the UnionDataset PR is merged

PR #3828, opened 2026-06-19
Add tessera-cdl datamodule

enomis-dev:
## AI Usage Disclosure

Split of the original PR #3499
It contains a datamodule implementation that combines CDL and Tessera embeddings. The idea is to create a specific datamodule that could be re-used in an exemple notebook in #3499.


adamjstewart:
Please fill out the AI usage disclosure form, see https://github.com/torchgeo/governance/blob/main/AI-POLICY.md#disclosure for details.

calebrob6:
@enomis-dev @adamjstewart is this something that you all have discussed splitting off in the context of #3499 ? My hot take is that a tessera / CDL datamodule is way too specific but I don't want to throw cold water on an existing discussion.

Regardless, @enomis-dev can you update to include a short description of the PR in the description field and a descriptive title so people who aren't familiar with #3499 can contribute. E.g. "WIP: add datamodule" doesn't tell me (a maintainer) what this PR is about. Just helpful tip for contributing to bigger projects :)

enomis-dev:
> @enomis-dev @adamjstewart is this something that you all have discussed splitting off in the context of #3499 ? My hot take is that a tessera / CDL datamodule is way too specific but I don't want to throw cold water on an existing discussion.
> 
> Regardless, @enomis-dev can you update to include a short description of the PR in the description field and a descriptive title so people who aren't familiar with #3499 can contribute. E.g. "WIP: add datamodule" doesn't tell me (a maintainer) what this PR is about. Just helpful tip for contributing to bigger projects :)

Hi @calebrob6  thanks for your comment :). I also share your point of view, when I started the PR the purpose was to leave everything in an isolated notebook but the suggestion was to create datamodule and/or dataset.  I think the created datamodule is not particularly reusable and would add just an extra piece of code to maintain to this lib.
The MR was more an attempt to see how embeddings could be used within the torchgeo framework. I also believe that NN performances are limited on these embeddings, a linear baseline with scikit-learn could perform even better (the original implementation has shown the difference between these 2 approaches but we decided to drop the scikit-learn part since outside of this lib). 
If you believe that the contribution would not add anything relevant I could simply close it :).

PS. for the naming of the PR, apologies, I contribute to large codebase even if not soo much on github and totally agree with you, the title should be understandable for maintainers. In this case I forgot to change it and it was simply the name of my commit :)

adamjstewart:
I guess I don't have strong opinions as long as it's < 5 lines of code in the final tutorial. So either:
```python
dm = TesseraCDLDataModule(...)
```
or:
```python
tessera = Tessera(...)
cdl = CDL(...)
dataset = tessera & cdl
sampler = RandomPatchSampler(dataset, ...)
dataloader = DataLoader(dataset, sampler=sampler)
```
I'm personally fine with more GeoDataModules, but understand @calebrob6's point about them quickly growing cumbersome if we have all possible combinations.

tests/data/tessera_cdl/data.py
adamjstewart, line 1: Not needed, we already have fake tessera and CDL data you can use during testing. May need to modify their data.py scripts to ensure they overlap.

tests/datamodules/test_tessera_cdl_datamodule.py
adamjstewart, line 1: See https://github.com/torchgeo/torchgeo/blob/main/tests/datamodules/README.md

torchgeo/datamodules/tessera_cdl_datamodule.py
```diff
@@ -0,0 +1,159 @@
...
+import torch
+from torch.utils.data import DataLoader
+
+from torchgeo.datamodules import GeoDataModule
+from torchgeo.datasets import CDL, GeoDataset, TesseraEmbeddings
+from torchgeo.datasets.utils import Sample
+from torchgeo.samplers import GriddedPatchSampler, RandomPatchSampler
+
+
+def collate_fn_embeddings(batch: list[dict[str, Any]]) -> dict[str, torch.Tensor]:
```
adamjstewart, line 17: Why does the default collation method not suffice?

torchgeo/datamodules/tessera_cdl_datamodule.py
```diff
@@ -0,0 +1,159 @@
...
+            batch_size=self.batch_size,
+            sampler=sampler,
+            collate_fn=collate_fn_embeddings,
+            num_workers=self.num_workers,
+            pin_memory=torch.cuda.is_available(),
+        )
+
+    def on_after_batch_transfer(self, batch: Sample, dataloader_idx: int) -> Sample:
+        """Skip augmentations for flattened pixel data."""
+        return batch
```
adamjstewart: Not needed, these are present in the base class

adamjstewart (changes requested):
If we do decide to implement this, it should be much simpler, see https://github.com/torchgeo/torchgeo/blob/main/torchgeo/datamodules/sentinel2_cdl.py for a good example.

torchgeo/datamodules/tessera_cdl_datamodule.py
enomis-dev: The extra function has the purpose to flatten the embeddings

adamjstewart: Can we do that in the dataset or trainer instead?

adamjstewart:
Currently violates our AI policy in terms of conciseness: more than double the amount of code it needs to be.

enomis-dev:
> Currently violates our AI policy in terms of conciseness: more than double the amount of code it needs to be.

I'm sorry but I have to disagree with this @adamjstewart. This is human generated code, this is not an AI generated MR. AI has been used as support.

How would you simplify it?

enomis-dev:
Here a summary of the changes and why I added some extra code

- I added TesseraCDLDataModule, which pairs Tessera embeddings with CDL crop labels for per-pixel classification.
-  Since a sample here is a single pixel's embedding or groups of embeddings rather than an image, a custom collate function flattens the usual image/mask batches into embeddings/labels pairs, and batch-level Kornia augmentation is skipped since it doesn't apply to flattened data.
- I also tried to re-use existing cdl data test and did a small refactoring.
- I've finally reduced the code changes by about the half by removing some defensive checks that I thought was a good idea to have.

Finally available to do any change you think is the best to match the style of this repo.

adamjstewart:
Apologies for the false accusation, I saw "AI-assisted" and just assumed. 

I think you could remove all changes except the following files:

* `docs/api/datamodules.rst`
* `torchgeo/datamodules/__init__.py`
* `torchgeo/datamodules/tessera_cdl.py`

Everything else isn't required. We have several other IntersectionDataset datamodules like this that only require those files.

enomis-dev:
> Apologies for the false accusation, I saw "AI-assisted" and just assumed.
> 
> I think you could remove all changes except the following files:
> 
> * `docs/api/datamodules.rst`
> * `torchgeo/datamodules/__init__.py`
> * `torchgeo/datamodules/tessera_cdl.py`
> 
> Everything else isn't required. We have several other IntersectionDataset datamodules like this that only require those files.

Thanks for the comment. Fair to assume, and no worries :)

The reason this datamodule needs its own tests: it's a different kind of IntersectionDataset datamodule. It produces embeddings/labels batches (for per-pixel classification), not image/mask. Every other IntersectionDataset datamodule in the repo gets full coverage for free through SemanticSegmentationTask's generic trainer test, because they all produce image/mask. This one can't plug into that, since no Task in torchgeo.tasks consumes flattened per-pixel embeddings.

You can see this concretely in my last commit, where I removed the dedicated test file: coverage "dropped" to 30%, (BTW that number is misleading because it's just import-time coverage of class/def signatures, not real exercise of any logic, in reality it is 0%).

I see the following options:

- Add a new Task for per-pixel classification, so this (and similar future datamodules) get tested the same way as everything else.
- Keep dedicated tests for this one, since it can't ride along on the existing mechanism. Understood if that's inconsistent with how the repo prefers to keep datamodules test-light.
- Drop the MR if this turns out to be irreconcilable.

Open to whichever direction you'd prefer :)

PR #3829, opened 2026-06-20
Glossary: distinguish terms related to 'index'

adriantre:
## AI Usage Disclosure

Written by me, AI helped with proofreading and formatting.


calebrob6 (dismissed):

calebrob6:
Nice, I think it is a good idea to disambiguate these!

adriantre:
Good! I don't seem to have the option to merge it into main, is that correct? My button said "Update Branch" even though I had already done so 😅 I clicked it just in case, though

docs/user/glossary.rst
```diff
@@ -23,6 +23,9 @@ A collection of common terms used in :mod:`torchgeo` that may be unfamiliar to e
    coordinate reference system (CRS)
        Synonym for :term:`spatial reference system (SRS)`. A system that defines how to locate geographic entities on a :term:`projected <projection>` surface.
 
+   index
+       The lookup table of a geospatial dataset (``GeoDataset.index``): a mapping from each file's spatiotemporal footprint to its path, used to find which files a :term:`query` overlaps. It is bookkeeping and sampling space, not pixel data. Unrelated to a :term:`spectral index`; and note that the ``index`` argument of ``__getitem__`` is, despite its name, a :term:`query`.
```
adamjstewart:
```suggestion
       The lookup table of a geospatial dataset (``GeoDataset.index``): a mapping from each file's spatiotemporal footprint to its path, used to find which files a :term:`query` overlaps. It is metadata, not pixel data. Unrelated to a :term:`spectral index`; and note that the ``index`` argument of ``__getitem__`` is, despite its name, a :term:`query`.
```

docs/user/glossary.rst
```diff
@@ -35,6 +38,9 @@ A collection of common terms used in :mod:`torchgeo` that may be unfamiliar to e
    projection
        A geometric transformation for portraying the surface of a 3D Earth onto a 2D planar image.
 
+   query
+       The spatiotemporal bounding box passed to a geospatial dataset's ``__getitem__`` (a ``GeoSlice``), describing *what to sample*. Distinct from the dataset's :term:`index`, which is *how* files are found. When a query is read in a file's native :term:`coordinate reference system (CRS)`, its box may align to a different grid than the index.
```
adamjstewart:
```suggestion
       The spatiotemporal bounding box passed to a geospatial dataset's ``__getitem__`` (a ``GeoSlice``), describing *where and when to sample*. Distinct from the dataset's :term:`index`, which is *how* files are found. When a query is read in a file's native :term:`coordinate reference system (CRS)`, its box may align to a different grid than the index.
```

adamjstewart (commented):
A bit verbose, but definitely good to add these to the docs!

docs/user/glossary.rst
adriantre: I like this

adamjstewart (approved):

adamjstewart:
If you like these kind of boring semantics, you could try to find a solution to #996, but I've kinda given up on that for 1.0.

PR #3832, opened 2026-06-22
Relieve files property of download responsibility

adriantre:
`GeoDataset.files` now raises `DatasetNotFoundError` when no files are found, and no longer cares about downloadable datasets. Download destination is centralized in a new `_download_root_path` property, and datasets check existence via `find_files` instead of the `files` property.

Removed `assert` in runtime code.

## AI Usage Disclosure


adamjstewart:
Can you explain the motivation for this? Perhaps these kind of refactors should start with an issue/discussion to decide whether or not they are needed and what the long-term vision is. I feel like I'm trying to simplify GeoDataset and you are trying to make it more complex (oftentimes for good reason) and want to make sure we're more aligned on this.

adriantre:
Sure thing! I have been deep into the details here, but I fully agree that this would suit a discussion.

I made an issue for it #3944 where I explain my thoughts. I hope it can show that the decoupling is meant to lower complexity rather than increase it. It is actually supposed to make GeoDataset simpler.

PR #3833, opened 2026-06-22
README: fix sponsorship logo

adamjstewart:
https://github.com/simple-icons/simple-icons/pull/14800 is now merged and released. https://simpleicons.org/?q=radiant+earth shows what the logo looks like. Turns out that spaces are removed instead of replaced with a dash. So far it still doesn't seem to load, so we may have to wait for img.shields.io to update their own simple icons version.

## AI Usage Disclosure


robmarkcole (approved):
lgtm

Issue #1078, opened 2023-01-31
`RasterDataset` support for nodata masks.


pmaldonado:
### Summary

We suggest adding support for nodata masking in `RasterDataset`, allowing users to optionally include a `nodata` mask with each sample returned by `__getitem__`.

### Rationale

In one of our applications, we take NAIP quarter quadrangles and sample the RasterDataset with a spatial sampler. We noticed that our labels will occasionally appear at the edge of a raster where there is no data present. We'd like to mask those regions during training and evaluation to prevent those labels from being included in the loss.

### Implementation

To produce optional "nodata" masks, we simply add a field to `RasterDataset` defining whether to return said masks with each sample, and when reading bands during a query we additionally read the raster nodata masks and return them as a tensor.

We previously implemented this feature on a forked version of `RasterDataset` and would be happy to submit a PR with the implementation.

adamjstewart:
I agree that we should have a way to access a nodata mask. However, I'm not sure the best way to do this. Your approach would work, but if you then want to combine NAIP with another mask dataset, the mask will go from B x H x W to B x C x H x W. Our current semantic segmentation trainers can't handle this, and I've been trying to standardize the output of our datasets more: #985

Reading through https://rasterio.readthedocs.io/en/latest/topics/masks.html, it seems like the nodata mask is simply the locations in the image where the value equals the nodata value. So if nodata=0, then all pixels in the image equal to zero are nodata pixels.

One option would be to manually parse this information from the image yourself during training instead of creating a separate mask returned by the data loader. The problem with this is you have to be careful to track how this value changes if you use image normalization.

Another option would be to use what you described and either forbid this kind of dataset being combined with a mask, or to make all of our masks B x C x H x W like Kornia expects. What do you think?

pmaldonado:
Parsing the nodata mask from the image during training strikes me as breaking the abstractions of the dataloader and the Trainer. I would expect the dataloader to do all necessary disk access, particularly given that `GeoDataset`s provide additional abstraction above raw imagery files.

One potential problem with pulling the `nodata` value from the imagery is that `nodatavals` is optional and sometimes returns `None`. For instance, this happens for all of our NAIP imagery. Nevertheless, using `.read([band], masked=True)` still returns the appropriate nodata mask.

Rasterio stores masks per-band, so the dataset could return a mask of shape C x H x W to match the `input` key and users can decide how to apply the mask downstream.

[...]

PR #3847, opened 2026-06-28
RasterDatasets: Let user specify nodata value of rasters

### LINKED ISSUE #2466: torchgeo reads dataset as all zeros while rasterio can read the contents
### Description

Hi, I'm not exactly sure what's causing this issue, though I have a suspicion it might be related to compression of geotiffs.

When data is downloaded from the UK environment agency's LIDAR data programme, it shows in torchgeo as all zeros.

This can be rectified by running the file through `gdal_merge.py` - I'm unsure why.

This is not an issue when reading directly with rasterio.

### Steps to reproduce

Run the following file. It downloads the relevant data,
Checking for a dataset reading all zeros, I get:
```
Rasterio compressed all zero: False
Rasterio uncompressed all zero: False
torchgeo compressed all zero: True
torchgeo uncompressed all zero: False
```

```python
import subprocess
from pathlib import Path

import rasterio

subprocess.run(["wget", "https://api.agrimetrics.co.uk/tiles/collections/survey/national_lidar_programme_dsm/2023/1/SP7015?subscription-key=public", "-O", "test_file.zip"])
subprocess.run(["unzip", "test_file.zip"])

test_file = Path("DSM_SP7015_P_12740_20230402_20230404.tif")
uncompressed_file = test_file.with_suffix(".uncompressed.tif")

with rasterio.open(test_file) as src:
    rasterio_compressed = src.read()
    rasterio_compressed_all_zero = (rasterio_compressed == 0).all()

torchgeo_compressed_dataset = RasterDataset(test_file)
torchgeo_compressed = torchgeo_compressed_dataset[torchgeo_compressed_dataset.bounds]["image"]
torchgeo_compressed_all_zero = (torchgeo_compressed == 0).all()

uncompressed_file.unlink(missing_ok=True)
subprocess.run(["gdal_merge.py", "-ot", "Float32", "-of", "GTiff", "-o", str(uncompressed_file), str(test_file)])

with rasterio.open(uncompressed_file) as src:
    rasterio_uncompressed = src.read()
    rasterio_uncompressed_all_zero = (rasterio_uncompressed == 0).all()

torchgeo_uncompressed_dataset = RasterDataset(uncompressed_file)
torchgeo_uncompressed = torchgeo_uncompressed_dataset[torchgeo_uncompressed_dataset.bounds]["image"]
torchgeo_uncompressed_all_zero = (torchgeo_uncompressed == 0).all()

print(f"Rasterio compressed all zero: {rasterio_compressed_all_zero}")
print(f"Rasterio uncompressed all zero: {rasterio_uncompressed_all_zero}")
print(f"torchgeo compressed all zero: {torchgeo_compressed_all_zero}")
print(f"torchgeo uncompressed all zero: {torchgeo_uncompressed_all_zero}")
```

### Version

ff3d087f76cd3df2ee01c3b7999f82459785a0a6 (latest main)

adriantre:
### Summary

- Additions
  - RasterDataset.nodata_value — overrides the source nodata via WarpedVRT(src_nodata=...), applied with and without reprojection.
  - XarrayDataset now fills no-data with the source's own nodata (NaN for typical xarray data) instead of a hard-coded 0.
- Motivations
  - Some rasters declare no nodata, or the wrong one (e.g. Sentinel-2 edge tiles), so derived valid-data/nodata masks are incorrect. nodata_value lets users set the correct value. Important for #2903, where valid-data masks are computed from nodata masks.
- Not relevant for XarrayDataset as xarray already carry nodata natively (decoded to NaN), so an override isn't needed.

Fixes #2466

### Checklist


isaaccorley (dismissed):

adamjstewart:
Long shot, but does this fix #2466? I'm pretty sure it doesn't fix #1078, but that's another request users have had.

torchgeo/datasets/geo.py
```diff
@@ -729,7 +726,7 @@ def _load_warp_file(self, filepath: Path, crs: CRS | None = None) -> DatasetRead
                 width=dst_width,
                 src_crs=src_crs,
                 src_transform=src_transform,
-                **nodata,
+                src_nodata=self.nodata_value,
```
adriantre: @isaaccorley this commit/change will erase the datasets existing nodata-value when self.nodata_value is None. My previous implementation passed it only when the user set it. I will revert it, ok?

adriantre:
> Long shot, but does this fix #2466? I'm pretty sure it doesn't fix #1078, but that's another request users have had.

#2466 was fixed by this PRm but only when `nodata_value` is overridden. My last commit now fixes it always, and I added a test case for it. The problem is this [line in rasterio](https://github.com/rasterio/rasterio/blob/5044f0c8cd08655c63c8b2e452618254c15d8f29/rasterio/merge.py#L363-L364) that has [a boundary case explained here](https://github.com/rasterio/rasterio/pull/3075). Fixed it here by explicitly casting to np.dtype during merge. 

For #1078 this PR enables it (but does not solve it). Now it is possible to get the valid mask like this, in `_merge_or_stack`:
```python
dest = rasterio.merge.merge(vrt_fhs, **kwargs, masked=True)[0]
valid_mask = torch.from_numpy(~np.ma.getmaskarray(dest)).any(dim=-3)  # [H, W]
```

calebrob6:
Other than a rebase, is this missing anything substantial @adamjstewart ?

torchgeo/datasets/geo.py
```diff
@@ -375,6 +375,9 @@ class RasterDataset(GeoDataset):
     #: True if data is stored in a separate file for each band, else False.
     separate_files = False
 
+    #: Nodata value for the dataset. If None, the source files' nodata value is used.
+    nodata_value: float | None = None
```
adamjstewart: Do we want to call this `nodata_value` or simply `nodata`? Also, can you add a sentence to `docs/tutorials/custom_raster_dataset.ipynb` describing this new attribute?

torchgeo/datasets/geo.py
```diff
@@ -644,11 +647,21 @@ def _merge_or_stack(
...
 
         x, y, _ = self._disambiguate_slice(index)
+
+        # rasterio.merge rejects a nodata equal to the dtype's own min because it
+        # runs np.min_scalar_type() on a float64 Python float internally.
+        # We cast it to np.dtype as a workaround.
+        fh = vrt_fhs[0]
+        nodata = self.nodata_value if self.nodata_value is not None else fh.nodata
+        if nodata is not None:
+            nodata = np.dtype(fh.dtypes[0]).type(nodata)
```
adamjstewart: I'll be honest, I don't understand this. By min you mean `np.finfo(dtype).min`, like a very small negative number? Or do you mean 0 for uint16, which is likely much more common?

torchgeo/datasets/geo.py
```diff
@@ -714,7 +727,11 @@ def _load_warp_file(self, filepath: Path, crs: CRS | None = None) -> DatasetRead
             )
         )
 
-        if needs_warp:
+        if needs_warp or self.nodata_value is not None:
+            # Only override the source nodata when explicitly set, else uses src.nodata.
+            override: dict[str, float] = {}
+            if self.nodata_value is not None:
+                override['src_nodata'] = self.nodata_value
```
adamjstewart: Are these changes actually needed if we always pass this through `_merge_or_stack`?

torchgeo/datasets/geo.py
adriantre: Can do! I'm not sure on the naming. I feel like `nodata` isn't descriptive enough, but there is probably a reason rasterio uses it. What do you think?

torchgeo/datasets/geo.py
adriantre: Yes the dtype min like this comment describes https://github.com/torchgeo/torchgeo/issues/2466#issuecomment-3725802208
I verified the bug and reported it [here](https://github.com/rasterio/rasterio/pull/3075)

torchgeo/datasets/geo.py
adriantre: I had the same thought, but `merge(nodata=…)` only affects the empty space filling between merged coverages (and the bugfix), and `src_nodata` on the VRT is responsible for the nodata inside the raster are masked, as well as being used during resampling and eventual calls to `dataset_masks()`.

torchgeo/datasets/geo.py
adamjstewart: I like to follow standards set by other people, then I don't have to convince myself why a different name is better. Let's go with `nodata`.

torchgeo/datasets/geo.py
adamjstewart: Oh weird. Not super excited about adding workarounds for rare corner cases. Maybe we can write this off as a bug in rasterio?

adriantre: I feel ya. Do I try to warn the user about it runtime or no?

adamjstewart: Nah, let's keep it simple. As long as users who encounter this obscure rasterio bug (and weird GeoTIFF setting) have a way to workaround this by manually setting `nodata=0` in their dataset then we're good.

adriantre: Removed it. Files with it nodata value specified as its dtypes extreme (e.g. min) will not be fixed automatically, nor through the override (only possible through casting input to merge).

torchgeo/datasets/geo.py
```diff
@@ -1010,8 +1018,12 @@ def _merge_files(
 
                 datasets.append(src)
 
+            nodata = datasets[0][self.data_vars[0]].rio.nodata
+            if nodata is None:
+                nodata = 0
```
adamjstewart: Should XarrayDataset also have a `nodata` attr? Why does XarrayDataset default to 0 and RasterDataset default to None?

adriantre: Can do, but it might be a bit cumbersome to test. I'll report back in a sec.

adriantre: Done. Tested it by shifting the query window east

adamjstewart: Still feels a bit inconsistent that we fallback to 0 for one but not the other, but maybe that's okay. Wonder if we shouldn't do something like:
```python
nodata = self.nodata or src.nodata or 0
```
in more locations and always explicitly set some value.

adriantre: So there is a subtle difference in 
1. overriding the source, and 
2. passing nodata to merge

both here and in RasterDataset.

For merge we chose a value for the empty spaces between mosaiced rasters, and for 1. we say "we know better about the internals of you". 

Currently: When nodata is not overridden and files does not have it:
- We trust `rasterio.merges`' [choise of nodata ](https://rasterio.readthedocs.io/en/stable/api/rasterio.merge.html#rasterio.merge.merge)in the "new" spaces (is `0`).
- We do not trust `xarray.merge_dataset`'s [choise of nodata](https://corteva.github.io/rioxarray/stable/rioxarray.html#rioxarray.merge.merge_datasets)  (also 0)

For symmetry we could explicitly fallback to 0 for rasterio too, or we could remove the fallback of 0 for xarray. I kept it explicit here, as it already was specified in the code.

But I'm hesitant to fallback to 0 for 1. since we don't want to guess on the internals of the rasters.

What do you think?

adriantre: Either way, we can not do

```python
nodata = self.nodata or src.nodata or 0
```
because if `self.nodata` is set to 0, which is falsy, it would take no effect. Must check None

adriantre: I suggest we remove the explicit nodata input to merge_datasets here, symmetric to RasterDataset

adriantre: Ugh, xarray file nodata override does not survive auto-detect of nodata in merge_datasets.

adriantre: I cleaned it up now. We have to explicitly pass `self.nodata` to `merge_datasets` because it does not see our overridden value on the source. And passing self.nodata when it is None (not overridden) makes it fallback to either the files nodata specification or 0. I'm happy

adamjstewart (approved):
Like the symmetry. I feel like there was a reason I used 0 instead of None (maybe a warning?) but we can investigate that later. XarrayDataset has always been and still is experimental, so I'm less worried about breaking things there than I am in RasterDataset.

PR #3249, opened 2026-01-06
RasterDataset: fix support for flipped or upside down rasters

adamjstewart:
Context: https://github.com/rasterio/rasterio/issues/3094
Fixes #1746 
Required by #3244

PR #3849, opened 2026-06-29
XarrayDataset: normalize y-axis orientation before merge

adriantre:
### Summary

XarrayDataset returned all-nodata for ascending-latitude data (and crashed on north-up) because _merge_files passed the source's signed resolution to merge_datasets.
Fixed by flipping sources to north-up and passing a positive resolution.

### Rationale

I have never used XarrayDataset myself, but I found this while implementing a feature for RasterDataset which I wanted to harmonise in XarrayDataset. Seems like the tests for XarrayDataset use a dataset that it does not support reading.

The added test confirms the problem when run on the main branch.

### Checklist

### DIFF
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -597,6 +597,12 @@ class TestXarrayDataset:
         ):
             dataset[0:0, 0:0, pd.Timestamp.min : pd.Timestamp.min]
 
+    def test_getitem_returns_source_data(self) -> None:
+        pytest.importorskip('h5py', minversion='3.10')
+        ds = XarrayDataset(os.path.join('tests', 'data', 'hdf5'))
+        image = ds[ds.bounds]['image']
+        assert (image > 0).any()
+
     def test_no_data(self, tmp_path: Path) -> None:
         with pytest.raises(DatasetNotFoundError, match='Dataset not found'):
             XarrayDataset(tmp_path)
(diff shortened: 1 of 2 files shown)


adamjstewart:
Good catch, we just fixed the same thing in RasterDataset: #3249.

adriantre:
Yeah, I remembered from when I refactored those lines. That's why I started looking at the orientation, after verifying the test-data was correct 😅 hard stuff to debug

tests/datasets/test_geo.py
```diff
@@ -591,6 +591,12 @@ def test_invalid_index(self, dataset: XarrayDataset) -> None:
         ):
             dataset[0:0, 0:0, pd.Timestamp.min : pd.Timestamp.min]
 
+    def test_getitem_returns_source_data(self) -> None:
+        pytest.importorskip('h5py', minversion='3.10')
+        ds = XarrayDataset(os.path.join('tests', 'data', 'hdf5'))
```
adriantre: Change this to use the fixture

nilsleh (dismissed):

adamjstewart:
Does XarrayDataset already work for negative x-res? That also required special handling in RasterDataset.

adriantre:
I've never heard about East-West oriented rasters, is that an actual occurrence?

I would think it would be documented alongside this
https://gdal.org/en/stable/tutorials/geotransforms_tut.html

isaaccorley (approved):

adamjstewart:
> I've never heard about East-West oriented rasters, is that an actual occurrence?

Never seen one before, but if NetCDF allows it, people will do it lol.

PR #3861, opened 2026-07-07
Add  IO profiler

evgeniialappo:
### Summary

This is a PR for the profiler to measure I/O speed. It simplifies the output and calculates the sampling rate. 

Sample of output (IO preprocessed data):
```
Profile Summary

| Split      | Strategy   | Samples    | Time (s)   | Rate (samples/s) |
| :--------- | :--------- | ---------: | ---------: | ---------------: |
| Train      | Random     |        928 |      7.649 |          121.319 |
| Validation | Grid       |        992 |      4.489 |          221.005 |
```
When copied into GitHub, this renders as:

| Split      | Strategy   | Samples    | Time (s)   | Rate (samples/s) |
| :--------- | :--------- | ---------: | ---------: | ---------------: |
| Train      | Random     |        928 |      7.649 |          121.319 |
| Validation | Grid       |        992 |      4.489 |          221.005 |

### Rationale

This PR is related to issue #3684.

### Checklist


torchgeo/profilers/io.py
```diff
@@ -0,0 +1,70 @@
...
+        super().__init__(dirpath=dirpath, filename=filename)
+        self._start: dict[str, float] = {}
+        self._action_count: defaultdict[str, int] = defaultdict(int)
+        self._batch_size = batch_size
+        self._end: dict[str, float] = {}
+        self._action_total_time: defaultdict[str, float] = defaultdict(float)
+
+    def start(self, action_name: str) -> None:
+        """Start recording."""
+        self._start[action_name] = time.monotonic()
```
adriantre: I'm really no expert here, but why not use `time.perf_counter()` here?

torchgeo/profilers/io.py
```diff
@@ -0,0 +1,70 @@
...
+    ) -> None:
+        """Initialise profiler."""
+        super().__init__(dirpath=dirpath, filename=filename)
+        self._start: dict[str, float] = {}
+        self._action_count: defaultdict[str, int] = defaultdict(int)
+        self._batch_size = batch_size
+        self._end: dict[str, float] = {}
+        self._action_total_time: defaultdict[str, float] = defaultdict(float)
+
+    def start(self, action_name: str) -> None:
```
adriantre, line 39: In [SimpleProfiler](https://lightning.ai/docs/pytorch/stable/_modules/lightning/pytorch/profilers/simple.html#SimpleProfiler), I see that they, in `start()` and `stop()`, check whether `action_name` has already started / not started. I'm not sure why/when it is necessary to safeguard against this, but should we do this too?

They then `.pop` from `_start` in `stop()` to ensure the value is not reused.

adriantre, line 39: Could add the `@override` decorator just in case the base Profiler from lightning changes signature for the method. The decorator would then trigger an exception.

torchgeo/profilers/io.py
```diff
@@ -0,0 +1,70 @@
...
+        self._end[action_name] = time.monotonic()
+        self._action_count[action_name] += 1
+        self._action_total_time[action_name] += (
+            self._end[action_name] - self._start[action_name]
+        )
+
+    def summary(self) -> str:
+        """Print summary."""
+        res = '\nProfile Summary: \n'
+        res += '\n| Action:| Time(sec):| Rate (patches/sec): |'
```
adriantre: I like the addition of throughput. `samples/sec` seems more fitting to me, as this the return-type (`Sample`) of the dataloader / `__getitem__`.

torchgeo/profilers/io.py
```diff
@@ -0,0 +1,70 @@
...
+        res += '\n| ------------- | ----------- | ----------- |'
+        for action_name in self._action_count:
+            train = 'train_dataloader_next' in action_name
+            val = 'val_next' in action_name
```
adriantre, line 72: Could add these to a class attribute "targeted_actions" for easy override/extending. Then use the raw action name in the print. I could see myself extending this with e.g. `predict_next`.

adriantre (commented):
Looks good to me apart from the questions that came while comparing with SimpleProfiler. Could possibly add a test that verify compatibility with the Torcgeo trainers too, for good measure.

```python
profiler = IOBenchProfiler()
trainer = ClassificationTask(..., profiler=profiler)
```

adamjstewart (changes requested):
A few other places that need to be updated:

- [x] Docs: `docs/api/index.rst`, `docs/api/profilers.rst`, `docs/user/contributing.rst`
- [x] Codeowners: `.github/CODEOWNERS`, add me for now
- [x] Labelers: `.github/labeler.yml`, `.github/release.yml`

calebrob6:
Why the new PR?

adamjstewart:
No AI is faster and needs fewer approvals.

nilsleh:
Regarding No AI, what is the difference then to the closed PR when the code is almost the same? Or is copying from AI-Assisted PR to a new PR then a No AI usage?

adamjstewart:
> Or is copying from AI-Assisted PR to a new PR then a No AI usage?

No, I would consider this AI-assisted still. 

@evgeniialappo can you comment on whether this new PR was written from scratch or simply copied from the previous PR?

evgeniialappo:
I wrote it from scratch as I was told to. I did not copy the previous PR, and I can't forget my old work I spent about 2 months on.

calebrob6 (dismissed):

calebrob6:
FYI I was totally fine with approving the old PR, and I'm guessing you would be too @adamjstewart? Unless I'm missing something, this seems like a bunch of extra work for no reason other than to satisfy a no AI purity test. Can we focus on the important things?

adamjstewart:
The main problem was that the old PR sat for 2 months and no one was willing to review it. At least if it's no-AI, then I can review it myself and don't need to wait for anyone else to review.

There are dozens of PRs that have been sitting for months without review because everyone wants to use AI (to save themselves time), but no one is willing to review PRs written by AI (to save themselves time). When I say "review", I don't mean "looks good to me", I mean an actual thorough review that involves looking at every line of code, understanding it, and agreeing to maintain it in perpetuity. Most of the AI-assisted PRs I've come into as the second approving reviewer have been miles away from being in a mergeable state, yet they have a ✅ because no one wants to take the time to review properly.

Let's discuss at the next TSC meeting. I don't know a great way to solve the current PR bottleneck if people aren't willing to review PRs. I also don't want contributors like @evgeniialappo to get stuck due to issues with our current AI policy. Hence the AI-free workaround.

docs/api/index.rst
adamjstewart, line 1: For the first 4 files, can you sort these alphabetically?

docs/user/contributing.rst
```diff
@@ -229,13 +229,13 @@ For PRs that may affect GeoDataset sampling speed, you can test the performance
...
+This code will download a small (1 GB) dataset consisting of a single Landsat 9 scene and CDL file. It will then profile the speed at which various samplers work for both raw data (original downloaded files) and preprocessed data (same CRS, res, TAP, COG). The important output to look out for is the sampling rate for ``Train (random)`` (RandomPatchSampler) and ``Validation (grid)`` (GriddedPatchSampler). With this, you can create a table on your PR like:
 
 ======  ============  ==========  =====================  ===================
  state  raw (random)  raw (grid)  preprocessed (random)  preprocessed (grid)
 ======  ============  ==========  =====================  ===================
-before        17.223      10.974                 15.685               4.6075
- after        17.360      11.032                  9.613               4.6673
+before      43.32153   112.87498              106.87363            197.57038
+ after      43.22781   111.97093              174.89302            197.48845
 ======  ============  ==========  =====================  ===================
```
adamjstewart: Let's simply include an example table from the output. The PR author can simply add two tables to the PR for before and after instead of manually reformatting everything and creating a new table.

tests/conf/io_raw.yaml
adamjstewart, line 1: Also add this to `tests/conf/iobench.yaml`

tests/profilers/__init__.py
adamjstewart, line 1: Delete this directory, it makes more sense to test these with an actual trainer.

torchgeo/profilers/io.py
```diff
@@ -0,0 +1,70 @@
...
+# Licensed under the MIT License.
+"""A custom I/O Profiler."""
+
+import time
+from collections import defaultdict
+
+from lightning.pytorch.profilers import Profiler
+
+
+class IOBenchProfiler(Profiler):
```
adamjstewart: Let's call this IOProfiler instead. It can be used with any dataset, not just IOBench.

torchgeo/profilers/io.py
adamjstewart: Remove trailing colon to match builtin profilers

tests/conf/io_preprocessed.yaml
```diff
@@ -11,4 +11,5 @@ trainer:
   accelerator: 'cpu'
   max_epochs: 1
   num_sanity_val_steps: 0
-  profiler: 'simple'
+  profiler:
+    class_path: IOBenchProfiler
```
adamjstewart: Any idea how Lightning goes from 'simple' to SimpleProfiler? Would be nice if we could also simply use 'io' instead of IOProfiler.

torchgeo/profilers/io.py
```diff
@@ -0,0 +1,70 @@
...
+    """A custom I/O Benchmarking Profiler.
+
+    .. versionadded:: 0.10
+    """
+
+    def __init__(
+        self,
+        dirpath: str | None = None,
+        filename: str | None = None,
+        batch_size: int = 32,
```
adamjstewart, line 22: This looks like it will cause a lot of confusion, as users will forget to change batch_size in two locations to match. Is it possible to grab this directly from the dataloader?

torchgeo/profilers/io.py
```diff
@@ -0,0 +1,70 @@
...
+    ) -> None:
+        """Initialise profiler."""
```
adamjstewart: Needs descriptions of all parameters

torchgeo/profilers/io.py
```diff
@@ -0,0 +1,70 @@
...
+    def stop(self, action_name: str) -> None:
+        """End recording."""
+        self._end[action_name] = time.monotonic()
+        self._action_count[action_name] += 1
+        self._action_total_time[action_name] += (
+            self._end[action_name] - self._start[action_name]
+        )
+
+    def summary(self) -> str:
+        """Print summary."""
```
adamjstewart: All methods need descriptions of parameters and return values (if any).

torchgeo/profilers/io.py
```diff
@@ -0,0 +1,70 @@
...
+
+class IOBenchProfiler(Profiler):
+    """A custom I/O Benchmarking Profiler.
+
+    .. versionadded:: 0.10
+    """
+
+    def __init__(
+        self,
+        dirpath: str | None = None,
```
adamjstewart, line 20: pathlib.Path is also supported, we should use `torchgeo.datasets.utils.Path` instead of only str

torchgeo/profilers/io.py
```diff
@@ -0,0 +1,70 @@
...
+        filename: str | None = None,
+        batch_size: int = 32,
+    ) -> None:
+        """Initialise profiler."""
+        super().__init__(dirpath=dirpath, filename=filename)
+        self._start: dict[str, float] = {}
+        self._action_count: defaultdict[str, int] = defaultdict(int)
+        self._batch_size = batch_size
+        self._end: dict[str, float] = {}
+        self._action_total_time: defaultdict[str, float] = defaultdict(float)
```
adamjstewart: No need for private attributes, none of the builtin profilers prefix with an underscore.

adamjstewart (changes requested):

torchgeo/profilers/io.py
adamjstewart: Builtin profilers use perf_counter, let's use that too

torchgeo/profilers/io.py
adamjstewart: I'm not sure why either. Feels like defensive coding. Let's keep it simple until we find a need for it.

torchgeo/profilers/io.py
adamjstewart: I would use `samples/s`

adamjstewart: Let's add one more column for samples, that way we can also compare epoch length when it changes

tests/conf/io_preprocessed.yaml
evgeniialappo: I found definitions in `trainer/setup.py` particularly in `_init_profiler`. I do not think it is designed to add custom names. Maybe there is a way, but I do not see it rn

torchgeo/profilers/io.py
evgeniialappo: I agree, but unfortunately, i did not find a way to make it work

docs/user/contributing.rst
```diff
@@ -229,16 +229,16 @@ For PRs that may affect GeoDataset sampling speed, you can test the performance
    $ python -m torchgeo fit --config tests/conf/io_raw.yaml
    $ python -m torchgeo fit --config tests/conf/io_preprocessed.yaml
 
-This code will download a small (1 GB) dataset consisting of a single Landsat 9 scene and CDL file. It will then profile the speed at which various samplers work for both raw data (original downloaded files) and preprocessed data (same CRS, res, TAP, COG). The important output to look out for is the total time taken by ``train_dataloader_next`` (RandomPatchSampler) and ``val_next`` (GriddedPatchSampler). With this, you can create a table on your PR like:
+This code will download a small (1 GB) dataset consisting of a single Landsat 9 scene and CDL file. It will then profile the speed at which various samplers work for both raw data (original downloaded files) and preprocessed data (same CRS, res, TAP, COG). The important output to look out for is the sampling rate for ``Train (random)`` (RandomPatchSampler) and ``Validation (grid)`` (GriddedPatchSampler). With this, you can add two tables before and after on your PR:
```
adamjstewart:
```suggestion
This code will download a small (1 GB) dataset consisting of a single Landsat 9 scene and CDL file. It will then profile the speed at which various samplers work for both raw data (original downloaded files) and preprocessed data (same CRS, res, TAP, COG). Each run will output a table like below. You can add these two tables to your PR to describe I/O performance before and after your change.
```

docs/user/contributing.rst
```diff
@@ -229,16 +229,16 @@ For PRs that may affect GeoDataset sampling speed, you can test the performance
...
-======  ============  ==========  =====================  ===================
+===================  =========  =========  ==================
+ Action               Samples    Time(s)    Rate (samples/s)
+===================  =========  =========  ==================
+Train (random)        928        8.35074    111.12790
+Validation (grid)     992        4.97084    199.56398
+===================  =========  =========  ==================
 
-In this example, we see a 60% speed-up for RandomPatchSampler on preprocessed data. All other numbers are more or less the same across multiple runs.
+All other numbers are more or less the same across multiple runs.
```
adamjstewart: Delete this, we are no longer comparing anything. Or add two tables and mention what changed.

torchgeo/profilers/io.py
```diff
@@ -0,0 +1,94 @@
...
+            dirpath: root directory to save profiler's results
+            filename: name of the file where the profiler's results will be saved
+            batch_size: batch size used by data loader
+        """
+        super().__init__(dirpath=dirpath, filename=filename)
+        self.start_time: dict[str, float] = {}
+        self.action_count: defaultdict[str, int] = defaultdict(int)
+        self.batch_size = batch_size
+        self.end_time: dict[str, float] = {}
+        self.action_total_time: defaultdict[str, float] = defaultdict(float)
```
adamjstewart: Are these type hints required or can ty infer them?

torchgeo/profilers/io.py
```diff
@@ -0,0 +1,94 @@
...
+                action_count = self.action_count[action_name]
+                samples = action_count * self.batch_size
+                rate = 0.0 if total_time == 0 else samples / total_time
+                label = 'Train (random)' if train else 'Validation (grid)'
+                res += f'\n| {label} | {samples} | {total_time:.5f} | {rate:.5f} |'
+
+        return res
+
+    @override
+    def teardown(self, stage: str | None) -> None:
```
adamjstewart: Do we actually need any teardown?

torchgeo/profilers/io.py
```diff
@@ -0,0 +1,94 @@
...
+            action_name: name of the action that is being profiled
+        """
+        self.end_time[action_name] = time.perf_counter()
+        self.action_count[action_name] += 1
+        self.action_total_time[action_name] += (
+            self.end_time[action_name] - self.start_time[action_name]
+        )
+
+    @override
+    def summary(self) -> str:
```
adamjstewart, line 61: Main thing left is to finalize the table format. A couple things worth thinking about:

1. I think we should split Action into 2 columns, Split (Train or Validation) and Strategy (Random or Grid). Technically these are specific to I/O Bench DataModule, you don't _have_ to use random for train and grid for val. We could consider dropping Strategy, but then it's less clear.
2. We should left-pad each cell with spaces so that each column lines up in the terminal, see the simple profiler source code for how
3. We should make numeric columns right-aligned, see https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-tables#formatting-content-within-your-table
4. I'm fine with 5 significant digits after the decimal point, although we probably don't need so many.

adamjstewart (commented):
Pretty close now!

torchgeo/profilers/io.py
adamjstewart: Let's keep it simple for now, we can always extend it later

adamjstewart (dismissed):
LGTM now! Would still like to get 1 more maintainer approval as this introduces an entirely new subpackage. But I think it makes perfect sense in the context of the torchvision and lightning APIs.

isaaccorley (dismissed):
this lgtm just needs to fix the merge conflicts

isaaccorley (approved):

PR #1453, opened 2023-06-29
Move USAVars to Huggingface

nilsleh:
This PR moves the USAVars dataset to Huggingface @calebrob6

PR #3865, opened 2026-07-09
Fix USAVars

calebrob6:
USAVars download seems to have not worked since after it was originally refactored? If you refactor something pleaseeeee make sure it actually works in real life with real bits 🙏 

Related https://github.com/torchgeo/torchgeo/pull/3777

## AI Usage Disclosure


adamjstewart (approved):
Looks like this broke way back in #1453 3 years ago and no one noticed until now. Thanks for the fix!

PR #3866, opened 2026-07-09
Add xBDDistShift dataset for disaster-based domain shift splits

calebrob6:
New version of https://github.com/torchgeo/torchgeo/pull/2416 as rebasing was insane. Kept @burakekim's contribution I think
 
xBDDistShift subclasses xBD to reformulate the train/test splits by disaster, enabling domain adaptation and out-of-distribution detection experiments (similar to EuroSATSpatial). One disaster is used as the in-distribution training set and another as the out-of-distribution test set, and the multiclass damage masks are reformulated as a binary building segmentation task.


torchgeo/datasets/xbd.py
```diff
@@ -294,3 +294,175 @@ def plot(
...
+
+        self._verify()
+
+        files = self._gather_files(root)
+        self.split_files = {
+            'train': self._filter_files(files, id_disaster, id_pre_post),
+            'test': self._filter_files(files, ood_disaster, ood_pre_post),
+        }
+
+    def __getitem__(self, index: int) -> Sample:
```
calebrob6: The `plot` method inherited from `xBD` crashes on samples from this class: base `plot` expects `sample['image']` to be T x C x H x W and indexes `sample['image'][0]` / `[1]` for the pre/post panels, but `__getitem__` here returns a single C x H x W image, so those expressions slice off 2D channel planes and `draw_semantic_segmentation_masks` raises `ValueError: Pass individual images, not batches`.

Repro:
```python
ds = xBDDistShift(root='tests/data/xbd', id_disaster='hurricane-harvey', ood_disaster='hurricane-michael')
ds.plot(ds[0])  # ValueError: Pass individual images, not batches
```

Needs a `plot` override (single image panel + binary overlay, passing a one-color list since the inherited 4-color `colormap` matches the 5-class scheme), plus a `test_plot` in `TestxBDDistShift` to cover it.

torchgeo/datasets/xbd.py
```diff
@@ -294,3 +294,175 @@ def plot(
...
+    :class:`EuroSATSpatial`. One disaster is used as the in-distribution (ID) training
+    set and a different disaster as the out-of-distribution (OOD) test set, so that a
+    model can be trained on one disaster type and evaluated on another. This tests the
+    ability of models trained on one disaster to generalize to a different one.
+
+    The multiclass damage masks are reformulated as a binary building segmentation task,
+    where minor-damage and no-damage buildings are mapped to the *building* class and
+    major-damage and destroyed buildings are mapped to the *background* class.
+
+    .. versionadded:: 0.10
```
calebrob6, line 320: Missing the citation block. Convention for variant classes (see the DLRSD review) is to duplicate the short citation section even though the base class has it:

```rst

* https://arxiv.org/abs/1911.09296
```

If the disaster-based split protocol comes from a paper, worth citing that here too.

torchgeo/datasets/xbd.py
```diff
@@ -294,3 +294,175 @@ def plot(
...
+    """
+
+    binary_classes = ('background', 'building')
```
calebrob6: `binary_classes` is never used. Since this class emits binary masks, the inherited `classes` attribute (5 damage classes) is now misleading for anyone introspecting the dataset — consider overriding `classes = ('background', 'building')` instead of introducing a new attribute name, and using it in the `plot` override.

torchgeo/datasets/xbd.py
```diff
@@ -294,3 +294,175 @@ def plot(
...
+            checksum: If True, check the MD5 of the downloaded files (may be slow).
+
+        Raises:
+            AssertionError: If *split* is invalid.
+            ValueError: If *id_disaster* or *ood_disaster* is not one of
+                *valid_disasters*.
+            DatasetNotFoundError: If dataset is not found.
+        """
+        assert split in ('train', 'test')
+        for disaster in (id_disaster, ood_disaster):
```
calebrob6: Optional: nothing rejects `id_disaster == ood_disaster`, which silently makes the train and test splits identical. A one-line `ValueError` here would prevent that (or a docstring note if it's intentionally allowed).

calebrob6 (commented):
(Note the following is me trying out a dataset review skill)

Self-review notes from running the branch locally: 17/17 tests pass, 100% line coverage on `torchgeo/datasets/xbd.py`, ruff format/check clean. One real bug (inherited `plot` crashes on the new sample shape) plus a few smaller items inline. Checked and fine: `_verify` correctly covers both split directories that `_gather_files` walks; `valid_disasters` matches the 10 tier-1 challenge disasters; no CSV row needed (EuroSATSpatial precedent); mask values confirmed binary at runtime.

adamjstewart:
> (Note the following is me trying out a dataset review skill)

Probably violates the AI policy, as all communication should be written by you, not AI. Ignoring, but feel free to self-review before opening a PR.

torchgeo/datasets/xbd.py
```diff
@@ -294,3 +294,175 @@ def plot(
 @deprecated('Use torchgeo.datasets.xBD instead')
 class XView2(xBD):
     """Deprecated alias for the xBD dataset."""
+
+
+class xBDDistShift(xBD):
```
adamjstewart, line 311: How similar is the new dataset? If we keep the structure the same, the subclass should only change `metadata` and nothing else.

tests/datasets/test_xbd.py
```diff
@@ -82,3 +82,60 @@ def test_plot(self, dataset: xBD) -> None:
         x['prediction'] = x['mask']
         dataset.plot(x)
         plt.close()
+
+
+class TestxBDDistShift:
```
adamjstewart: How similar is the new dataset? Can we parameterize the original tests over 2 classes like we do for EuroSAT? Then we only need to maintain 1 line of code instead of 60.

adamjstewart (changes requested):

torchgeo/datasets/xbd.py
calebrob6: This is the same fundamental dataset but with capabilities for training testing on different splits. We could pull these features into the original dataset

calebrob6:
Tagging @burakekim in case he wants to chime in here too (on why this would be good to have in torchgeo, the point of the dataset, etc.)

burakekim:
Distribution shift and generalization work in this space typically builds its own splits and this makes results hard to compare. `xBDDistShift` fixes this (by treating disaster event as the shift axis) -- the same way `EuroSATSpatial` fixes geography for `EuroSAT`

@adamjstewart  @calebrob6

torchgeo/datasets/xbd.py
burakekim: split-by-disaster and binary mask logic are not just metadata, so a metadata subclass would not work as is. We can pull the disaster-split and binary-remap logic into the main xBD class as optional arguments (then xBDDistShift could be like EuroSATSpatial) -- a few lines of code to maintain on the base class but makes this class an easy addition

tests/datasets/test_xbd.py
burakekim: agreed this should be parameterizable the same way EuroSAT shift class is -- the only real difference is indeed the split/class config

adamjstewart:
Definitely agree with the motivation to add this! @calebrob6 you decide how much work is required to implement this new split without code duplication. If it's not easy, we can always merge as is and simplify it later. Trying to promote good design/less code to maintain without being too restrictive.

torchgeo/datasets/xbd.py
```diff
@@ -172,6 +159,27 @@ def _load_files(
...
+        image1 = self._load_image(files['image1'])
+        image2 = self._load_image(files['image2'])
+        mask1 = self._load_target(files['mask1'])
+        mask2 = self._load_target(files['mask2'])
+
+        image = torch.stack(tensors=[image1, image2], dim=0)
+        # Dataset consists of semantic segmentation masks before and after event
+        # Convert to change detection by subtracting damage before from damage after
+        # Clamp to avoid potential negative numbers
+        mask = torch.clamp(mask2 - mask1, 0, 4)
```
calebrob6, line 180: This seems wrong. Anyone know why it is like this?

calebrob6:
Okay see what you guys think about this lightly refactored one @burakekim @adamjstewart. Note I made the citation for this the TARDIS paper where we used it https://arxiv.org/abs/2412.13394

torchgeo/datasets/xbd.py
adamjstewart: I wrote this bit. We could either make this the only semantic segmentation dataset with time-series for both inputs and outputs, or make it a traditional change detection dataset. I was chatting with @RitwikGupta and we opted for the latter. So instead of predicting damage level before and after we now predict change in damage level. The clamp is for the off chance that a building goes from damaged to undamaged.

torchgeo/datasets/xbd.py
```diff
@@ -298,3 +306,197 @@ def plot(
...
+
+        Returns:
+            image and binary building mask
+        """
+        image = self._load_image(files['image'])
+        mask = self._load_target(files['mask'])
+        mask = ((mask == 1) | (mask == 2)).long()
+        return {'image': image, 'mask': mask}
+
+    def _validate_disaster_shift(
```
adamjstewart: We only use this helper function in one place. Could we replace this with 3 assert statements?
```python
assert set(id_disaster, ood_disaster) <= set(self.valid_disasters)
assert id_disaster != ood_disaster
assert set(id_pre_post, ood_pre_post) <= {'pre', 'post', 'both'}
```
Then we could also drop another 30 lines of unit tests

adamjstewart (commented):
Probably possible to shave off another 100 lines if the tests and plots are made more similar. Not too worried about the plots since I have a student working on one plot to rule them all.

torchgeo/datasets/xbd.py
calebrob6: Okay sounds good, thanks!

torchgeo/datasets/xbd.py
calebrob6: Fair!

calebrob6:
Tests cut using asserts, leaving the plot alone to give someone the pure endorphin rush of deleting it later

adamjstewart (approved):

burakekim:
Yey! Crazy how I ghosted this task for n years. @calebrob6 the great is leading by example!!!

PR #3877, opened 2026-07-11
Add progress indicator to download_url via requests+tqdm

### LINKED ISSUE #3797: Better download indicators
### Summary

#3338 replaced our previous download logic with a simple one-liner using Python's built-in urllib. Unfortunately, it also removed all user-facing messages. I would like to add more detailed information, including:

* what is being downloaded: from where, to where, and
* a progress bar: preferably also with download time so far, data transfer rate, and expected time to finish.

### Rationale

Downloads can run for minutes to hours to days. **All** slow-running processes should have some indication of what is happening and how long it might take. This is needed so users aren't left wondering whether or not the program has hung, or how much time/storage will be required so they can decide whether or not to kill the program.

P.S. If there are other slow-running processes other than downloading that lack a message (maybe checksumming?), please open another issue to track those.

### Implementation

I would suggest first looking at existing CLIs to know what good designs look like. For example, curl, wget, pip, uv, azcopy, s3, etc. I'm completely fine with adding required dependencies on requests/tqdm if needed, but would like to minimize other more heavy required dependencies that aren't already in the dependency tree. We may even be able to make things faster by parallelizing downloads, but that could be a separate PR to keep this one focused.

> [!WARNING]
> Please complete this literature review first before asking to take this issue or opening a PR, as it may take a lot of back-and-forth before we decide on an implementation.

### Alternatives

#2957 (fsspec, obstore), boto3, and other libraries may help, especially for cloud storage. But if there are many dependencies, I would prefer that they remain optional.

### Additional information

_No response_

Haris-bin-shakeel:
Closes #3797

## Summary

This PR adds a progress bar to `download_url()`. Since `download_and_extract_archive()` uses it internally, downloads there also show progress now. I also included the source URL and destination path in the output so it's easier to see what is being downloaded and where it is being saved.

## Why

At the moment, there is no feedback while a file is downloading. If the file is large or the connection is slow, it can look like nothing is happening. A progress bar makes it clear that the download is still running and gives a rough idea of how much is left.

## What I changed

For HTTP and HTTPS downloads, I used `requests` with streaming and wrapped it with `tqdm` so progress can be shown while the file is being received.

For other URL types, especially `file://` which is used in the tests, I kept the existing `urllib` code and added the same progress display there.

I didn't change the `download_url()` function signature, so existing code should continue to work the same way. The only new dependency is `tqdm`.

As discussed in the issue, this PR is only about showing download progress. I didn't make any changes related to parallel downloads or performance.

## Checklist

Before writing the code, I looked at how tools like `curl`, `wget`, `pip`, and `uv` handle download progress. I decided to keep the existing behavior for non-HTTP URLs and only use `requests` for streamed HTTP downloads. AI helped me think through a few implementation details and edge cases, but I made the final implementation decisions, fixed the CI issues, and I understand every part of the code.

### DIFF
--- a/torchgeo/datasets/utils.py
+++ b/torchgeo/datasets/utils.py
@@ -24,11 +24,13 @@ from collections.abc import Iterable, Iterator, Mapping, Sequence
 from dataclasses import dataclass
 from datetime import datetime, timedelta
 from typing import Any, TypeAlias, cast, overload
+from urllib.parse import urlparse
 
 import numpy as np
 import pandas as pd
 import pyogrio
 import rasterio
+import requests
 import shapely.affinity
 import torch
 from numpy.typing import NDArray
@@ -37,6 +39,7 @@ from rasterio import Affine
 from shapely import Geometry
 from torch import Tensor
 from torchvision.utils import draw_segmentation_masks
+from tqdm import tqdm
 from typing_extensions import deprecated
 
 from .errors import DependencyNotFoundError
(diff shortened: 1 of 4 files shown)


nilsleh (changes requested):
Please indicate the use of AI tools with the dedicated traffic light indicator from the PR template.

Haris-bin-shakeel:
Hi @nilsleh, I've added the 🟡 AI-assisted disclosure to the PR description and resolved the uv.lock conflict with upstream main. Could you take another look when you get a chance? Thanks!

torchgeo/datasets/utils.py
```diff
@@ -423,15 +426,54 @@ def download_url(
...
+                        unit_divisor=1024,
+                    ) as bar,
+                ):
+                    for chunk in response.iter_content(chunk_size=8192):
+                        if chunk:
+                            f.write(chunk)
+                            bar.update(len(chunk))
+            else:
+                # file:// and other non-HTTP schemes: urllib handles these
+                # natively, requests does not.
```
adamjstewart, line 458: Why not always use urllib then?

adamjstewart (changes requested):
If we can get this PR down to ~5 lines of code I would be much happier.

adamjstewart:
Superseded by #3961

Issue #2839, opened 2025-06-16
Linting PyTorch Errors with TorchFix


isaaccorley:
### Summary

There are some torch specific syntax that are recommended over others. We can use a linter like [TorchFix](https://github.com/pytorch-labs/torchfix) to double check that we aren't introducing any silent bugs.

Running the following in the root of the torchgeo repo yields the below errors (looks like they can be autofixed by passing the `--fix` arg):

```bash
pip install torchfix

torchfix --select=ALL .

```

```bash.
experiments/ssl4eo/plot_example_predictions.py:20:14: TOR102 [*] `torch.load` without `weights_only` parameter is unsafe. Explicitly set `weights_only` to False only if you trust the data you load and full pickle functionality is needed, otherwise set `weights_only=True`.
tests/conftest.py:15:12: TOR102 [*] `torch.load` without `weights_only` parameter is unsafe. Explicitly set `weights_only` to False only if you trust the data you load and full pickle functionality is needed, otherwise set `weights_only=True`.
--- /home/ubuntu/github/torchgeo/experiments/ssl4eo/plot_example_predictions.py
+++ /home/ubuntu/github/torchgeo/experiments/ssl4eo/plot_example_predictions.py
@@ -15,11 +15,11 @@
 
 device = torch.device('cpu')
 
 # Load weights
 path = 'data/l7irish/checkpoint-epoch=26-val_loss=0.68.ckpt'
-state_dict = torch.load(path, map_location=device)['state_dict']
+state_dict = torch.load(path, map_location=device, weights_only=True)['state_dict']
 state_dict = {key.replace('model.', ''): value for key, value in state_dict.items()}
 
 # Initialize model
 model = smp.Unet(encoder_name='resnet18', in_channels=9, classes=5)
 model.to(device)
--- /home/ubuntu/github/torchgeo/tests/conftest.py
+++ /home/ubuntu/github/torchgeo/tests/conftest.py
@@ -10,11 +10,11 @@
 import torchvision
 from pytest import MonkeyPatch
 
 
 def load(*args: Any, progress: bool = False, **kwargs: Any) -> Any:
-    return torch.load(*args, **kwargs)
+    return torch.load(*args, **kwargs, weights_only=True)
 
 
 @pytest.fixture
 def load_state_dict_from_url(monkeypatch: MonkeyPatch) -> None:
     monkeypatch.setattr(torchvision.models._api, 'load_state_dict_from_url', load)
[...]

PR #3893, opened 2026-07-23
Security hardening: torch.load(weights_only=True)

adamjstewart:
### Summary

Use `weights_only=True` for all uses of `torch.load()` or `torchvision.models.WeightsEnum.get_state_dict()`, which calls `torch.hub.load_state_dict_from_url()`, which calls `torch.load()`.

### Rationale

Security vulnerability reported by @tinyb0y

Also suggested in #2839

### Checklist


robmarkcole (approved):

isaaccorley (approved):

isaaccorley:
Might want to run the integration test for loading all the real checkpoints because I _think_ this will break if loading from a pytorch lightning ckpt

adamjstewart:
All tests passed locally, I think this is safe to merge.

PR #3895, opened 2026-07-23
OlmoEarth: force model to use v1

adamjstewart:
### Summary

Previously, the model would use whatever the latest version is. Now, we force the model to use version 1.0.

### Rationale

The weights themselves are for v1.0, so the model architecture silently changing underneath is not a good idea. I also hit the following test failure:
```console
> pytest -m slow --force-enable-socket tests/models/test_olmoearth.py::TestOlmoEarthV1::test_olmoearth_v1_download[OlmoEarthV1_Weights.LARGE]
...
ValueError: model_size 'large' is not available for v1.2. Must be one of ['nano', 'small', 'tiny', 'base']
```

@piperwolters @favyen2 

### Checklist

### DIFF
--- a/torchgeo/models/olmoearth.py
+++ b/torchgeo/models/olmoearth.py
@@ -90,7 +90,9 @@ def olmoearth_v1(
     model_size = kwargs.pop('model_size', 'nano')
     if weights is not None:
         model_size = weights.meta.get('model_size', model_size)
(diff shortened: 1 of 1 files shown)


favyen2 (approved):
Looks good to me, I didn't realize there was this path in our `olmoearth_pretrain_minimal` code since we mostly only use the `load_model_from_id` path. Thanks for the fix for using OlmoEarth in torchgeo!

isaaccorley (approved):

PR #3894, opened 2026-07-23
HF: md5 -> sha256

adamjstewart:
### Summary

Converts all Hugging Face datasets from MD5 to SHA256 checksums.

### Rationale

MD5 is not cryptographically secure, SHA256 is a better choice. I started with HF since it directly reports SHA256, making it easy to convert without actually downloading anything. Will work on other datasets in follow-up PRs, but we should advise people to convert WIP PRs as well.

### Checklist

- [x] I have read the [Contributing docs](https://docs.torchgeo.org/en/stable/user/contributing.html)
- [x] I have read the [AI policy](https://github.com/torchgeo/governance/blob/main/AI-POLICY.md)

- [x] 🟢 **No AI usage**: written by humans, for humans
- [ ] 🟡 **AI-assisted**: AI helped with the coding, but I understand every line
- [ ] 🔴 **AI-generated**: AI did everything, review with caution

PR #3899, opened 2026-07-24
xBD: MD5 -> SHA256

adamjstewart:
### Summary

Same as #3894 but for xBD

### Rationale

Same as #3894 

### Checklist

### DIFF
--- a/torchgeo/datasets/xbd.py
+++ b/torchgeo/datasets/xbd.py
@@ -60,12 +60,12 @@ class xBD(NonGeoDataset):
     metadata: ClassVar[dict[str, dict[str, str]]] = {
         'train': {
             'filename': 'train_images_labels_targets.tar.gz',
-            'md5': 'a20ebbfb7eb3452785b63ad02ffd1e16',
+            'sha1': 'b37a4ef4ee9c909e2b19d046e49d42ee3965714b',
             'directory': 'train',
         },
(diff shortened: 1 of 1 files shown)


RitwikGupta:
These are the official hashes from the xView2 download page, lgtm!

adamjstewart:
Hmm, apparently SHA1 can also be broken, maybe we should calculate SHA256 instead? From https://en.wikipedia.org/wiki/Cryptographic_hash_function#SHA-1:

> Collisions against the full SHA-1 algorithm can be produced using the shattered attack and the hash function should be considered broken.

calebrob6:
I'm not really worried about this, but here's the sha256 if its helpful:
```
0fcdbfe3ee7d0842729dd2230217e74b2f12be35546ff666df4dae5388e2541c  test_images_labels_targets.tar.gz
a5941b7a3e523eafc4aeaa740a1c83f1af6a18c894e7e8c62dd830a76921ecd4  train_images_labels_targets.tar.gz
```

"Proof" that these match the expected md5sum/sha1sums from this PR:

image

adamjstewart:
Ah, I remember the days when images constituted proof. But I trust you. Let me convert these.

calebrob6:
Trust, but verify!

robmarkcole (approved):

PR #3906, opened 2026-07-25
Add rasterio type hints

adamjstewart:
### Summary

Trying out the new typeshed stubs for rasterio (thanks @thomas-maschler!). There seem to be a few bugs in both the type hints and in TorchGeo, will work on fixing those one-by-one.

### Rationale

The type hints caught bugs in TorchGeo (mostly pyproj CRS vs. rasterio CRS not guaranteed to be compatible).

### Checklist


adamjstewart:
Split up into separate PRs for each new rule, please review those instead, will recreate this once those are merged.

thomas-maschler:
@adamjstewart great that the type hints can be of use. Let me know if there are bugs (or create PR against typeshet). This was the first big pass and it's possible that I missed some details.

adamjstewart:
@adriantre can you take a look at the CRS stuff in the last commit to make sure I didn't mess anything up? It gets pretty complicated, especially in your helper functions.

adamjstewart:
> ```
> Error: torchgeo/datasets/openaerialmap.py:457:17: invalid-assignment: Invalid assignment to data descriptor attribute `crs` on type `DatasetWriter` with custom `__set__` method
> ```

@kshitijrajsharma can you take a look at this error? I have no idea what you're trying to do in this code, but it doesn't seem to be valid?

adamjstewart:
@thomas-maschler first possible bug/improvement: we use DatasetReader and WarpedVRT interchangeably in many places in TorchGeo. This seems to work, although I don't know if this is by plan or by coincidence. Should we consider supporting both in functions like `rasterio.merge.merge` and `rasterio.stack.stack`? Should one be a subclass of another?

adamjstewart:
Second bug: `rasterio.features.rasterize` should accept any Geometry shape (e.g., Polygon), not only a GeoJSON-like Mapping.

thomas-maschler:
Thanks for the feedback. I included the shared base class `DatasetReaderBase` which you can use when both DatasetReader and WarpedVRT are accepted. 

Please note that while in most cases these two classes can be use interchangeably, you cannot use WarpedVRT for `plot.show` / `plot.show_hist`. The functions perform a an `isinstance(source, DatasetReader)` check at runtime.

I also refactored the Geometry type to `Geometry: TypeAlias = Mapping[str, Any] | _SupportsGeoInterface`

adriantre:
> @adriantre can you take a look at the CRS stuff in the last commit

I'm back from holiday next week, and will take a look 👌🏻

torchgeo/datasets/mapinwild.py
```diff
@@ -222,8 +222,7 @@ def _load_raster(self, filename: int, source: Path) -> Tensor:
             the raster image or target
         """
         with rasterio.open(os.path.join(self.root, source, f'{filename}.tif')) as f:
-            raw_array = f.read()
-            array: np.typing.NDArray[np.int_] = np.stack(raw_array, axis=0)
```
adamjstewart, line 228: @burakekim do you remember why this was added? It just seems to stack a single array into itself.

torchgeo/datasets/openaerialmap.py
```diff
@@ -454,7 +454,6 @@ def _georeference_tile(self, filepath: str, tile: TileUtils.Tile) -> None:
                     dataset.width,
                     dataset.height,
                 )
-                dataset.crs = CRS.from_epsg(4326)
```
adamjstewart, line 457: `dataset.crs` is a read-only attribute, I don't think this is valid.

kshitijrajsharma:
> > ```
> > Error: torchgeo/datasets/openaerialmap.py:457:17: invalid-assignment: Invalid assignment to data descriptor attribute `crs` on type `DatasetWriter` with custom `__set__` method
> > ```
> 
> @kshitijrajsharma can you take a look at this error? I have no idea what you're trying to do in this code, but it doesn't seem to be valid?

Thanks  , am also in summer break will look into this upon return !

adamjstewart:
@kshitijrajsharma when does your summer break end? Would love to sneak this PR into the next release.

kshitijrajsharma:
@adamjstewart  I have send you the PR on your fork branch to cleanly import CRS from rasterio !

adamjstewart:
@kshitijrajsharma this line is not valid:
```python
dataset.crs = RIO_CRS.from_epsg(4326)
```
Can you suggest a replacement?

kshitijrajsharma:
> @kshitijrajsharma this line is not valid:
> 
> ```python
> dataset.crs = RIO_CRS.from_epsg(4326)
> ```
> 
> Can you suggest a replacement?

Because dataset crs being readonly ? How do we set the CRS of dataset in that case ?  
Here dataset is being open in `r+` which should give crs override option for rasterio ?

adamjstewart:
Correct, I have no idea. You tested this dataset with real data and it worked, right?

kshitijrajsharma:
> Correct, I have no idea. You tested this dataset with real data and it worked, right?
@adamjstewart  

Check this tiny python script : 
```python
import numpy as np
import rasterio
from rasterio.crs import CRS

with rasterio.open('/tmp/t.tif', 'w', width=1, height=1, count=1, dtype='uint8') as f:
    f.write(np.zeros((1, 1, 1), dtype='uint8'))

with rasterio.open('/tmp/t.tif', 'r+') as f:
    f.crs = CRS.from_epsg(4326)
    print(f'r+  {type(f).__name__:14} -> crs = {f.crs}')

with rasterio.open('/tmp/t.tif', 'r') as f:
    try:
        f.crs = CRS.from_epsg(3857)
    except rasterio.errors.DatasetAttributeError as error:
        print(f'r   {type(f).__name__:14} -> {error}')

print(f'rasterio {rasterio.__version__}')
``` 
it should allow to write crs in r+ mode 
```
r+  DatasetWriter  -> crs = EPSG:4326
r   DatasetReader  -> read-only attribute
rasterio 1.5.0
```

kshitijrajsharma:
Tested and verified the dataset @adamjstewart  Attaching tiny real API call hereby : 

```python
import glob
import tempfile

import rasterio
import requests
from pyproj import CRS

from torchgeo.datasets import OpenAerialMap

url = 'https://api.imagery.hotosm.org/stac/search'
bbox = requests.post(url, json={'limit': 1}, timeout=30).json()['features'][0]['bbox']

with tempfile.TemporaryDirectory() as root:
    OpenAerialMap(paths=root, bbox=tuple(bbox), zoom=16, download=True)
    for path in sorted(glob.glob(f'{root}/OAM-*.tif')):
        with rasterio.open(path) as tile:
            print(f'{path.split("/")[-1]:26} {tile.crs}')

    dataset = OpenAerialMap(paths=root, crs=CRS.from_epsg(3857))
    print(f'reprojected to EPSG:3857 -> {dataset.res[0]:.4f} m/px')
```

adamjstewart:
@thomas-maschler perhaps this is another bug in the type hints?

adamjstewart:
Hmm, seems to be working now, perhaps this was fixed in the last release too.

robmarkcole (dismissed):
Approve, pending response from @burakekim

burakekim:
@adamjstewart @robmarkcole  

tldr: Fine to drop it and use raw_array directly (while keeping the dtype casting)

No good reason that I can defend today. It seems f.read() already returns (C, H, W), so np.stack(raw_array, axis=0) is a useless copy. I am guessing it is most likely a leftover from an earlier version where I read bands individually

robmarkcole (approved):

PR #3907, opened 2026-07-25
TRY: ruff 0.16 fixes for tryceratops

adamjstewart:
### Summary

Fixes all TRY002 and TRY004 errors in TorchGeo.

### Rationale

Ruff 0.16 now enables these rules by default. Since these are backwards-incompatible, I figured it's worth separating them out from the larger ruff PR for detailed review.

@ashnair1 some of the code in OSM is the definition of defensive coding/AI slop. Would love to improve this and simplify the class.

### Checklist

### DIFF
--- a/tests/datasets/test_geo.py
+++ b/tests/datasets/test_geo.py
@@ -199,7 +199,7 @@ class TestGeoDataset:
     def test_and_nongeo(self, dataset: GeoDataset) -> None:
         ds2 = CustomNonGeoDataset()
         with pytest.raises(
-            ValueError, match='IntersectionDataset only supports GeoDatasets'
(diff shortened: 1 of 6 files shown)


ashnair1:
@adamjstewart Could you clarify which sections you consider problematic? If you mean _validate_classes which is modified in this PR, I can see it might be too explicit and could be trimmed to just check whether its empty or has the name and selector keys. Can look into simplifying where it makes sense.

adamjstewart:
Yes, that's the main part I'm referring to, although I can't say I understand most of the dataset or tests. I realize OSM is complex, I just wish the dataset class was easier to parse. Some stuff can be handled better by type checking anyway.

ashnair1 (approved):

PR #3930, opened 2026-07-30
Security hardening: checksum datasets by default

adamjstewart:
### Summary

Changes all datasets to validate checksums by default.

### Rationale

Checksum verification adds security and reproducibility to our datasets. We should probably do this by default or even make it required (like torchvision).

### Checklist


calebrob6 (approved):
Note I didn't actually check to see if you did this for every dataset

adamjstewart:
I just used sed, it should be done consistently.

PR #3887, opened 2026-07-18
Rename trainers to tasks

adamjstewart:
### Summary

This PR renames the `torchgeo.trainers` namespace to `torchgeo.tasks` and renames each task to drop the `Task` suffix. For example:

* `torchgeo.trainers.SpatioTemporalSegmentationTask` -> `torchgeo.tasks.SpatioTemporalSegmentation`

Deprecated aliases are added to maintain backwards compatibility, but these will be removed in the 1.0 release.

### Rationale

`torchgeo.trainers` and `pl.Trainer` are unrelated things, we should more clearly distinguish between the two. `Task` is redundant, as we are importing from `torchgeo.tasks` and there are no conflicts with other imports.

Closes #996

### Checklist

- [x] I have read the [Contributing docs](https://docs.torchgeo.org/en/stable/user/contributing.html)
- [x] I have read the [AI policy](https://github.com/torchgeo/governance/blob/main/AI-POLICY.md)

- [x] 🟢 **No AI usage**: written by humans, for humans
- [ ] 🟡 **AI-assisted**: AI helped with the coding, but I understand every line
- [ ] 🔴 **AI-generated**: AI did everything, review with caution

PR #3935, opened 2026-08-03
Fix crashes from tensor truthiness and unhandled branches

isaaccorley:
Running [pyrefly](https://pyrefly.org/) over the codebase as a one-off audit turned up a set of latent runtime errors. Each one below is reachable from public API and now has a regression test. Pyrefly is **not** added as a dependency and is not wired into CI — this PR is just the bug fixes it found.

Annotation-only cleanups from the same audit are split out into #3936 to keep this one reviewable.

> Rebased onto `main` after #3887 (trainers → tasks). The affected fixes now live in `torchgeo/tasks/` and use the new `MoCo`/`SimCLR` class names.

### Tensor truthiness

`MoCo` and `SimCLR` both did:

```python
grayscale_weights = grayscale_weights or torch.ones(in_channels)
```

`Tensor.__bool__` raises for anything with more than one element, so **any caller passing `grayscale_weights` explicitly hit `RuntimeError: Boolean value of Tensor with more than one value is ambiguous`**. No test passed the argument, so it went unnoticed.

The same `or` idiom on `augmentation1`/`augmentation2` was subtler: `nn.Sequential` defines `__len__`, so an empty one is falsy and got silently replaced by the default augmentation.

### Wrong attribute name

`IOBench._download` read `self.sha256`, which doesn't exist — the class defines `sha256s`. Only reachable with `checksum=True`, because the ternary short-circuits otherwise, which is exactly why no test caught it.

### Unbound locals

* `BaseDataModule.on_after_batch_transfer` — `split` was never assigned when a trainer was attached but in none of the train/val/test/predict states, giving `UnboundLocalError`. Now returns the batch unaugmented.
* `CROMA.forward` — a configured modality with a `None` input skipped its block and then hit `NameError` on `sar_encodings`/`optical_encodings` at the joint encoder. Now raises `ValueError` naming the missing argument.
* `ChnEmb.forward` — `mus` was unbound for input that was neither 2D nor 3D.
* `PASTIS.__getitem__` — `sample` was unbound if the `assert` was stripped under `python -O`. `mode` is now `Literal['semantic', 'instance']` and the chain ends in `else`.
* `MoCo.training_step` and `MMEarth._preprocess_modality` — both chains are exhaustive over validated inputs, so the last branch is now `else` rather than a condition that a checker can't prove is total.
* `_get_input_layer_name_and_module` — raises on a model with no child modules instead of `UnboundLocalError`.

### Attribute contract

`Substation.rgb_bands` held channel *indices* `(3, 2, 1)` while `PlottingMixin.rgb_bands` is documented as band *names* and typed `tuple[str, ...]`. Substation's `bands` are integer indices throughout, so it can't use the mixin attribute — renamed to a private `_rgb_indices`. It's referenced only inside `Substation.plot`, nowhere in tests or docs.

### Verification

`pre-commit run --all-files` passes (ruff, ruff-format, ty, prettier). All touched modules under `torchgeo/` are at 100% coverage.

### DIFF
--- a/tests/datasets/test_iobench.py
+++ b/tests/datasets/test_iobench.py
@@ -63,20 +63,6 @@ class TestIOBench:
             shutil.copy(tarfile, root)
         IOBench(root)
 
-    def test_download_checksum(
-        self, dataset: IOBench, monkeypatch: MonkeyPatch
-    ) -> None:
-        checksum = ''
-
-        def download_url(url: str, root: Path, sha256: str | None = None) -> None:
-            nonlocal checksum
-            checksum = sha256 or ''
-
-        monkeypatch.setattr('torchgeo.datasets.iobench.download_url', download_url)
-        dataset.checksum = True
-        dataset._download()
-        assert checksum == dataset.sha256s[dataset.split]
-
     def test_not_downloaded(self, tmp_path: Path) -> None:
         with pytest.raises(DatasetNotFoundError, match='Dataset not found'):
             IOBench(tmp_path)
--- a/tests/datasets/test_iobench.py
+++ b/tests/datasets/test_iobench.py
@@ -63,6 +63,20 @@ class TestIOBench:
             shutil.copy(tarfile, root)
         IOBench(root)
 
+    def test_download_checksum(
+        self, dataset: IOBench, monkeypatch: MonkeyPatch
+    ) -> None:
+        checksum = ''
+
+        def download_url(url: str, root: Path, sha256: str | None = None) -> None:
+            nonlocal checksum
+            checksum = sha256 or ''
+
+        monkeypatch.setattr('torchgeo.datasets.iobench.download_url', download_url)
+        dataset.checksum = True
+        dataset._download()
+        assert checksum == dataset.sha256s[dataset.split]
+
     def test_not_downloaded(self, tmp_path: Path) -> None:
         with pytest.raises(DatasetNotFoundError, match='Dataset not found'):
             IOBench(tmp_path)
--- a/tests/datasets/test_pastis.py
+++ b/tests/datasets/test_pastis.py
@@ -104,10 +104,6 @@ class TestPASTIS:
         with pytest.raises(AssertionError):
             PASTIS(folds=(0,))
 
-    def test_invalid_mode(self) -> None:
-        with pytest.raises(AssertionError):
-            PASTIS(mode='invalid')  # ty: ignore[invalid-argument-type]
-
     def test_invalid_bands(self) -> None:
         with pytest.raises(ValueError, match='bands must be a subset of'):
             PASTIS(bands=('B01',))
--- a/tests/datasets/test_pastis.py
+++ b/tests/datasets/test_pastis.py
@@ -5,6 +5,7 @@ import os
 import shutil
 from itertools import product
 from pathlib import Path
+from typing import Literal, cast
 
 import matplotlib.pyplot as plt
 import pytest
@@ -37,8 +38,7 @@ class TestPASTIS:
 
         root = tmp_path
         bands = params['bands']
-        mode = params['mode']
-        assert isinstance(mode, str)
+        mode = cast(Literal['semantic', 'instance'], params['mode'])
         transforms = nn.Identity()
 
         url = os.path.join('tests', 'data', 'pastis', 'PASTIS-R.zip')
@@ -106,7 +106,7 @@ class TestPASTIS:
 
     def test_invalid_mode(self) -> None:
         with pytest.raises(AssertionError):
-            PASTIS(mode='invalid')
+            PASTIS(mode='invalid')  # ty: ignore[invalid-argument-type]
 
     def test_invalid_bands(self) -> None:
         with pytest.raises(ValueError, match='bands must be a subset of'):
--- a/torchgeo/datamodules/geo.py
+++ b/torchgeo/datamodules/geo.py
@@ -134,8 +134,6 @@ class BaseDataModule(LightningDataModule):
                 split = 'test'
             elif self.trainer.predicting:
                 split = 'predict'
-            else:
-                return batch
 
             aug = self._valid_attribute(f'{split}_aug', 'aug')
             batch = aug(batch)
--- a/torchgeo/datamodules/geo.py
+++ b/torchgeo/datamodules/geo.py
@@ -134,6 +134,8 @@ class BaseDataModule(LightningDataModule):
                 split = 'test'
             elif self.trainer.predicting:
                 split = 'predict'
+            else:
+                return batch
 
             aug = self._valid_attribute(f'{split}_aug', 'aug')
             batch = aug(batch)
--- a/torchgeo/datasets/mmearth.py
+++ b/torchgeo/datasets/mmearth.py
@@ -568,7 +568,11 @@ class MMEarth(NonGeoDataset):
             data = data.astype(np.int32)
             # no data value also 255 for biome and 65535 for eco_region
             tensor = torch.from_numpy(data).long()
-        else:  # sentinel2_cloudmask, sentinel2_cloudprod, sentinel2_scl
+        elif modality in [
+            'sentinel2_cloudmask',
+            'sentinel2_cloudprod',
+            'sentinel2_scl',
+        ]:
             tensor = torch.from_numpy(data.astype(np.int32)).long()
 
         # TODO: tensor might still contain nans, how to handle this?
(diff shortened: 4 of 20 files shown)


robmarkcole (dismissed):
lgtm

tests/datasets/test_iobench.py
```diff
@@ -63,6 +63,20 @@ def test_already_downloaded(self, tmp_path: Path) -> None:
             shutil.copy(tarfile, root)
         IOBench(root)
 
+    def test_download_checksum(
```
adamjstewart: Feels strange to test this for iobench but for no other datasets. We should generally try to improve if-statement branch coverage in the future. I don't expect it to get better unless we enable it and take the coverage cut, then enforce it for new PRs. Let's remove this for now.

torchgeo/datamodules/geo.py
```diff
@@ -134,6 +134,8 @@ def on_after_batch_transfer(self, batch: Sample, dataloader_idx: int) -> Sample:
                 split = 'test'
             elif self.trainer.predicting:
                 split = 'predict'
+            else:
+                return batch
```
adamjstewart: I know why the type checker is flagging this, but I also feel like this is overkill.

torchgeo/datasets/mmearth.py
```diff
@@ -568,11 +568,7 @@ def _preprocess_modality(
             data = data.astype(np.int32)
             # no data value also 255 for biome and 65535 for eco_region
             tensor = torch.from_numpy(data).long()
-        elif modality in [
-            'sentinel2_cloudmask',
-            'sentinel2_cloudprod',
-            'sentinel2_scl',
-        ]:
+        else:  # sentinel2_cloudmask, sentinel2_cloudprod, sentinel2_scl
```
adamjstewart: Not sure how I feel about these changes. I frequently use match-case statements with no fallbacks because other values are not allowed. If we change `modality` from `str` to `Literal[...]`, is this still needed to make pyrefly happy?

torchgeo/datasets/pastis.py
```diff
@@ -226,7 +226,7 @@ def __getitem__(self, index: int) -> Sample:
         if self.mode == 'semantic':
             mask = self._load_semantic_targets(index)
             sample = {'image': image, 'mask': mask}
-        elif self.mode == 'instance':
+        else:  # instance
```
adamjstewart: Well this answers my above question. This feels like a bug in pyrefly.

adamjstewart (commented):
Bug fixes look great and could be merged right away if they are in a separate PR.

Remaining type changes feel overly cautious. Some of these feel like bugs in pyrefly instead. We should decide how much "weirdness" we want to put up with in order to make type checkers happy (and find real bugs!) while also keeping the code style nice to read. We should also strongly consider making branch coverage required, not optional.

isaaccorley:
> Bug fixes look great and could be merged right away if they are in a separate PR.
> 
> Remaining type changes feel overly cautious. Some of these feel like bugs in pyrefly instead. We should decide how much "weirdness" we want to put up with in order to make type checkers happy (and find real bugs!) while also keeping the code style nice to read. We should also strongly consider making branch coverage required, not optional.

I'll kill the type changes. It seems to be mostly rules that pyrefly enforces that ty doesn't like requiring that there be an `else` statement in the if/else branch.

Edit: pyrefly seems to be happy with just Literals now rather than the else statement

torchgeo/datasets/mmearth.py
isaaccorley: yeah pyrefly has a rule that requires an `else` to exist. We can remove though.

tests/datasets/test_iobench.py
isaaccorley: Removed in c588d9b1 — the `sha256s` fix stays, the test is gone. Agreed on branch coverage; happy to look at enabling it in a separate PR.

torchgeo/datamodules/geo.py
isaaccorley: Reverted in c588d9b1.

torchgeo/datasets/mmearth.py
isaaccorley: Reverted in c588d9b1, so this is moot for now — but to answer the question properly, since I gave you bad information earlier: **yes, `Literal[...]` alone is enough. Pyrefly does not require an `else`.**

I tested a matrix of cases — 2 vs 3 literals, local parameter vs instance attribute, inside a `for` loop — and `if`/`elif` with no `else` is clean in every one. Only a plain `str` chain gets flagged.

What actually broke it here is a separate thing:

```python
assert target in self.targets   # targets: tuple[str, ...]
```

Comparing against a `tuple[str, ...]` (or a `list[str]`) **widens the parameter from `Literal[...]` back to `str`**, which is what defeats exhaustiveness downstream. I confirmed by reproducing FTW in isolation and deleting only that line: 1 error, then 0.

So if we ever wanted `_preprocess_modality` clean, it would be `modality: Literal[...]` plus making sure nothing widens it — no `else` needed.

torchgeo/datasets/pastis.py
isaaccorley: It is not a pyrefly bug — my `else` here was unnecessary and led you to the wrong conclusion. Sorry about that.

PASTIS is back to `if`/`elif` with no `else` in c588d9b1, and pyrefly is clean on `__getitem__`. Two things make that work:

```python
class PASTIS(NonGeoDataset):
    mode: Literal["semantic", "instance"]   # <-- the part that matters
```

The parameter annotation alone was not enough, because `assert mode in ["semantic", "instance"]` compares against a `list[str]` and widens `mode` back to `str`. Declaring the attribute on the class pins it, and the exhaustive chain then needs no fallback.

Kept the `Literal` since it seemed like the style you were after, but happy to drop it entirely if you would rather this PR be strictly bug fixes.

tests/datasets/test_pastis.py
```diff
@@ -37,8 +38,7 @@ def dataset(
 
         root = tmp_path
         bands = params['bands']
-        mode = params['mode']
-        assert isinstance(mode, str)
+        mode = cast(Literal['semantic', 'instance'], params['mode'])
```
adamjstewart, line 41: Does this work?
```python
model: Literal['semantic', 'instance'] = params['mode']
```

tests/datasets/test_pastis.py
```diff
@@ -106,7 +106,7 @@ def test_invalid_fold(self) -> None:
 
     def test_invalid_mode(self) -> None:
         with pytest.raises(AssertionError):
-            PASTIS(mode='invalid')
+            PASTIS(mode='invalid')  # ty: ignore[invalid-argument-type]
```
adamjstewart: Can we just delete this test?

torchgeo/datasets/substation.py
```diff
@@ -48,8 +48,10 @@ class Substation(NonGeoDataset):
     * https://doi.org/10.48550/arXiv.2409.17363
     """
 
-    # Sentinel-2 true color: B04 (Red), B03 (Green), B02 (Blue) = indices 3, 2, 1
-    rgb_bands = (3, 2, 1)
+    # Sentinel-2 true color: B04 (Red), B03 (Green), B02 (Blue).
+    # Unlike most datasets, *bands* are channel indices rather than names, so this
+    # cannot use PlottingMixin.rgb_bands, which holds band names.
+    _rgb_indices: tuple[int, int, int] = (3, 2, 1)
```
adamjstewart: How hard would it be to change this to strings?

torchgeo/models/croma.py
```diff
@@ -141,8 +150,8 @@ def forward(
 
         if set(self.modalities) == {'sar', 'optical'}:
             joint_encodings = self.joint_encoder(
-                x=sar_encodings,
-                context=optical_encodings,
+                x=return_dict['sar_encodings'],
+                context=return_dict['optical_encodings'],
```
adamjstewart, line 154: What's the reason for this change?

torchgeo/tasks/simclr.py
```diff
@@ -143,9 +143,12 @@ def __init__(
         self.weights = weights
         super().__init__()
 
-        grayscale_weights = grayscale_weights or torch.ones(in_channels)
-        self.augmentations = augmentations or simclr_augmentations(
-            size, grayscale_weights
+        if grayscale_weights is None:
+            grayscale_weights = torch.ones(in_channels)
+        self.augmentations = (
```
adamjstewart, line 148: Optional: could this also be an if-statement? I dislike single lines that stretch onto multiple, harder to read

adamjstewart:
Also missing AI usage template, I hope the PR description was written by you 😉

tests/datasets/test_pastis.py
adamjstewart: Nope.

adamjstewart (approved):

PR #3936, opened 2026-08-03
Correct type annotations found by pyrefly

isaaccorley:
## Summary

Annotation-only cleanups found by [pyrefly](https://pyrefly.org/)

### Wrong types

* **`get_model_weights` returns `type[WeightsEnum]`, not `WeightsEnum`.** The values of `_model_weights` are enum *classes* (`Aurora_Weights`, etc.), not members — which is why `get_weight` has to iterate them. The docstring already said "the weights enum class". This drops two suppressions, one of which (`# ty :ignore[invalid-assignment]`) was inert anyway thanks to a typo — a space before `:ignore`.
* **`OpenBuildings._filter_geometries` returns `list[BaseGeometry]`.** It builds its result from `gdf.geometry.tolist()`, so `list[dict[str, Any]]` was never right.
* **`rasterio_loader` returns `NDArray[np.int32]`.** The body does `.astype(np.int32)`; `np.int_` is int64 on Linux and macOS.
* **`TreeSatAI._multilabel_to_string`** collects into `list[tuple[str, float]]` but was appending 0-d tensors — now calls `.item()`.

### Matplotlib compatibility

`SpatioTemporalSampler`'s animation callbacks are annotated `-> Iterable[Artist]` but returned `ax.add_collection3d(poly)` directly. That only started returning the collection in **matplotlib 3.10**; torchgeo supports `>=3.7.3`, where it returns `None`. Harmless today because `FuncAnimation` is constructed with the default `blit=False` and discards the value, but the annotation was wrong on most supported versions. Now calls `add_collection3d` and returns `(poly,)`, which is correct everywhere.

### Untypeable indirection

`PreChippedGeoSampler.__iter__` routed through a `Callable[[int], Iterable[int]]` alias assigned either `range` or `partial(torch.randperm, ...)` — no single signature describes both. Building the index sequence directly in the branch is equivalent and drops the `functools.partial` import.

### Missing declarations

`DLRSDBase` declares `image_fns`, and `_list_dict_to_dict_list` / `stack_samples` annotate their accumulator dicts.

### Checklist

### DIFF
--- a/tests/models/test_api.py
+++ b/tests/models/test_api.py
@@ -161,8 +161,8 @@ def test_get_model_weights(builder: Callable[..., nn.Module]) -> None:
 
 
 @pytest.mark.parametrize('enum', enums)
-def test_get_weight(enum: WeightsEnum) -> None:
-    for weight in enum:  # ty: ignore[not-iterable]
+def test_get_weight(enum: type[WeightsEnum]) -> None:
+    for weight in enum:
         assert weight == get_weight(str(weight))
 
 
--- a/torchgeo/datasets/dlrsd.py
+++ b/torchgeo/datasets/dlrsd.py
@@ -75,6 +75,8 @@ class DLRSDBase(NonGeoDataset):
 
     directory = 'DLRSD'
 
+    image_fns: list[str]
+
     classes = (
         'airplane',
         'bare soil',
--- a/torchgeo/datasets/openbuildings.py
+++ b/torchgeo/datasets/openbuildings.py
@@ -6,7 +6,7 @@
 import glob
 import os
 from collections.abc import Callable, Iterable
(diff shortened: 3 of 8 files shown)


robmarkcole (approved):

adamjstewart:
How fast is pyrefly compared to ty? Does it do all of the type checking things ty does plus Tensor shapes? Do we think pyrefly is worth adding to CI, either as a supplement or replacement? I am very interested in having shape types for all of TorchGeo someday, it would help greatly with https://github.com/torchgeo/torchgeo/discussions/985

isaaccorley:
It was actually quite fast maybe like a second or two I think? It wouldn't be much overhead on CI to add it. Yes it's basically ty + tensor shapes.

PR #3940, opened 2026-08-04
Fix rasterio typing errors

isaaccorley:
## Summary

- import `Affine` from its canonical package so `ty` recognizes it as callable
- reopen the in-memory raster with `rasterio.open` so its read-only type is explicit

### Checklist

### DIFF
--- a/tests/datasets/test_utils.py
+++ b/tests/datasets/test_utils.py
@@ -14,9 +14,10 @@ import pandas as pd
 import pytest
 import rasterio
 import torch
+from affine import Affine
 from numpy.typing import NDArray
 from pytest import MonkeyPatch
-from rasterio import Affine, MemoryFile
+from rasterio import MemoryFile
 from rasterio.transform import from_origin
 from rasterio.vrt import WarpedVRT
 from shapely import MultiPolygon, Polygon, box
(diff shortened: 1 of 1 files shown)


calebrob6 (approved):

calebrob6:
lgtm and the g is for great

adamjstewart:
We need to add a dependency on affine if we import it this way, that's why we were importing it from rasterio instead.

adamjstewart:
@thomas-maschler are there plans to add type hints for `rasterio.Affine`? I know it's simply rasterio forwarding the affine dependency, but we've secretly been using it directly to avoid yet another direct dependency to track.

thomas-maschler:
I was back and forth on that and decided not to. Affine v3 will ship with type hints and is already in  v3.0rc3. I'd rather give maintainers a nudge to make the final release.

adamjstewart:
Affine v3 is now out, but the type hints don't seem to propagate through to rasterio. I tried undoing this in https://github.com/torchgeo/torchgeo/pull/3953 but it still complains with:
```console
> ty check
error[call-non-callable]: Object of type `<special-form 'typing.Any'>` is not callable
   --> tests/datasets/test_utils.py:711:17
    |
711 |     transform = Affine(res, 0.0, 0.0, 0.0, y_res, 0.0)
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Found 1 diagnostic
```
It looks like we need to update or remove this? https://github.com/python/typeshed/blob/main/stubs/rasterio/rasterio/_affine_types.pyi

PR #3942, opened 2026-08-05
docs: clarify DOFA Sentinel-1 wavelength convention

### LINKED ISSUE #3931: Clarify Sentinel-1 wavelength convention for DOFA checkpoints
### Summary

The TorchGeo DOFA API describes wavelengths in µm, but the upstream Sentinel-1 example uses `[5.405, 5.405]`. This appears to be the C-band frequency in GHz rather than a physical wavelength. The physical wavelength would be approximately 55,500 µm.

I have also been advised that the released DOFA checkpoint may instead expect `[3.75, 3.75]` as a placeholder because the physical value is far outside the wavelength distribution seen during pretraining.

Could the authors please clarify:

1. What values should be used for Sentinel-1 VV/VH with the DOFA MAE checkpoints currently exposed by TorchGeo?
2. Are these values intended as modality placeholders rather than physical wavelengths?
3. Does `[5.405, 5.405]` correspond to a different or newer DOFA checkpoint/version?

This matters because the dynamic patch embedding is wavelength-conditioned, so using a different convention can substantially change the extracted representations.

References:

- [TorchGeo DOFA implementation](https://docs.torchgeo.org/en/stable/_modules/torchgeo/models/dofa.html#DOFA.forward_features)
- [Upstream DOFA Sentinel-1 example](https://github.com/zhu-xlab/DOFA?tab=readme-ov-file#load-sentinel-1-data-with-2-channels)

### Fix

_No response_

hfangcat:
### Summary

Updates the `wavelengths` parameter descriptions in `DOFA.forward()` and `DOFA.forward_features()` to document the Sentinel-1 input conventions for DOFA v1 and v2. The current TorchGeo weights use the DOFA v1 convention, `[3.75, 3.75]`, while DOFA v2 uses `[5.405, 5.405]`.

### Rationale

The previous descriptions state that `wavelengths` contains spectral band wavelengths in micrometers. However, for Sentinel-1, these inputs are modality placeholders rather than physical wavelengths. The Sentinel-1 C-SAR instrument has a center frequency of 5.405 GHz, corresponding to a physical wavelength of approximately 5.55 cm (55,500 μm), so the previous wording could be misleading.

Reference: [Sentinel-1 Mission](https://sentiwiki.copernicus.eu/web/s1-mission)

Closes #3931

### Checklist

### DIFF
--- a/torchgeo/models/dofa.py
+++ b/torchgeo/models/dofa.py
@@ -328,10 +328,9 @@ class DOFA(nn.Module):
         Args:
             x: Input mini-batch.
             wavelengths: Wavelengths of each spectral band (μm). For Sentinel-1
-                VV/VH, the DOFA v1 weights currently provided by TorchGeo use
+                VV/VH, the DOFA v1 weights use
                 ``[3.75, 3.75]`` as modality placeholders rather than physical
-                wavelengths. DOFA v2 instead uses ``[5.405, 5.405]``; its weights
-                are not currently provided by TorchGeo.    
+                wavelengths.
 
         Returns:
             Output mini-batch.
@@ -380,11 +379,10 @@ class DOFA(nn.Module):
         Args:
             x: Input mini-batch.
             wavelengths: Wavelengths of each spectral band (μm). For Sentinel-1
-                VV/VH, the DOFA v1 weights currently provided by TorchGeo use
+                VV/VH, the DOFA v1 weights use
                 ``[3.75, 3.75]`` as modality placeholders rather than physical
-                wavelengths. DOFA v2 instead uses ``[5.405, 5.405]``; its weights
-                are not currently provided by TorchGeo.
-        
+                wavelengths.
+
         Returns:
             Output mini-batch.
         """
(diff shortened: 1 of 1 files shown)


torchgeo/models/dofa.py
```diff
@@ -327,7 +327,11 @@ def forward_features(self, x: Tensor, wavelengths: list[float]) -> Tensor:
 
         Args:
             x: Input mini-batch.
-            wavelengths: Wavelengths of each spectral band (μm).
+            wavelengths: Wavelengths of each spectral band (μm). For Sentinel-1
+                VV/VH, the DOFA v1 weights currently provided by TorchGeo use
+                ``[3.75, 3.75]`` as modality placeholders rather than physical
+                wavelengths. DOFA v2 instead uses ``[5.405, 5.405]``; its weights
+                are not currently provided by TorchGeo.
```
adamjstewart:
```suggestion
            wavelengths: Wavelengths of each spectral band (μm). For Sentinel-1
                VV/VH, the DOFA v1 weights use
                ``[3.75, 3.75]`` as modality placeholders rather than physical
                wavelengths. 
```
Let's keep it simple, it's obvious we are talking about TorchGeo and no need to mention a model that isn't yet in TorchGeo.

adamjstewart (commented):
Make it simpler and make ruff happy

adamjstewart (dismissed):

adamjstewart (approved):

PR #3954, opened 2026-08-09
Fix BYOL trainer

calebrob6:
Fixing two bugs in the BYOL trainer:
- We used to use the same model instance for both the target and the backbone. This is not how BYOL is suppossed to work.
- We used to create the backbone model without `num_classes=0`. By default this creates an 1000-class fc head that is never used or updated. (**Edit:** fun fact, with a resnet50 sized feature representation this adds 2M extra parameters 🥲)

Note that any BYOL checkpoints written before this change will contain `fc.*` keys (and target weights identical to online weights). This will give you unexpected `fc.*` keys under strict loading.

### Checklist

### DIFF
--- a/torchgeo/tasks/byol.py
+++ b/torchgeo/tasks/byol.py
@@ -3,6 +3,7 @@
 
 """BYOL task for self-supervised learning (SSL)."""
 
+import copy
 import os
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


torchgeo/tasks/byol.py
```diff
@@ -333,7 +348,10 @@ def configure_models(self) -> None:
 
         # Create backbone
         backbone = timm.create_model(
-            self.hparams['model'], in_chans=in_channels, pretrained=weights is True
+            self.hparams['model'],
+            in_chans=in_channels,
+            num_classes=0,
```
calebrob6, line 353: If we don't do this then the backbone has an additional massive imagenet sized fc head that gets run all the time for no reason.

isaaccorley (approved):

calebrob6:
I'm getting a ~93.5% test set score on EuroSat 13-band, 224px pretraining of a resnet50 with this fixed version of BYOL

PR #3958, opened 2026-08-09
Remove webdataset dependency

calebrob6:
webdataset's only consumer in torchgeo was `CopernicusPretrain`. This PR keeps the dataset but replaces webdataset with a minimal built-in tar-shard streamer, dropping `webdataset` and its `braceexpand` dependency from the `datasets` extra.

### What the built-in loader supports

* Numeric brace notation (`example-{000000..000009}.tar`) for shard lists.
* Streaming reads of local shards and remote http(s) shards (via `requests`, already a required dependency) — shards are never fully loaded into memory.
* Sample grouping by key prefix with `.pth` entries decoded via `torch.load(..., weights_only=True)`; `.json` metadata is skipped (matching the existing behavior of dropping non-Tensor values).
* A fixed-size sample shuffle buffer (`shuffle_buffer`, default 10, matching the previous `.shuffle(10)`).
* Automatic shard splitting across distributed ranks and DataLoader workers, and `resampled=True` for an infinite stream sampling shards with replacement.

### API changes

* The constructor now takes explicit `urls`, `shardshuffle`, `resampled`, and `shuffle_buffer` arguments instead of forwarding `*args/**kwargs` to `webdataset.WebDataset` (documented with `versionchanged:: 0.10`).
* The internal `dataset.dataset` attribute (the wds pipeline) is gone, so the documented `webdataset.WebLoader` / `.batched()` usage patterns no longer apply — the docstring example now shows plain `DataLoader` usage. Sample dicts and their keys (`s1_grd.pth`, …) are unchanged.

### Testing

* Existing `test_getitem`/`test_plot` pass unchanged against the real fixture shard.
* New tests cover brace expansion, multi-sample shards, remote (http) streaming via a stubbed `requests.get`, worker and distributed-rank splitting, `shardshuffle`, `resampled`, the shuffle buffer, and samples with missing modalities.
* Full `tests/datasets/test_copernicus.py` run with a stub `webdataset` package that raises `ImportError` on import: 165 passed, `pretrain.py` at 100% line coverage. `ruff` 0.16.1 and `ty` 0.0.66 clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,8 +109,6 @@ datasets = [
     # tokenizers 0.14+ required for Python 3.12 wheels
     # tokenizers does not yet provide wheels for Python 3.14t
     "tokenizers>=0.14; python_version<'3.14'",
-    # webdataset 0.2.4+ required for PyTorch tuple seed support
-    "webdataset>=0.2.4",
     # xarray 0.17+ required by rioxarray
     "xarray>=0.17",
 ]
--- a/tests/datasets/test_copernicus.py
+++ b/tests/datasets/test_copernicus.py
@@ -1,13 +1,18 @@
 # Copyright (c) TorchGeo Contributors. All rights reserved.
 # Licensed under the MIT License.
 
+import io
 import os
 import shutil
+import tarfile
 from pathlib import Path
+from typing import NamedTuple
 
 import pandas as pd
 import pytest
+import requests
 import torch
+import torch.utils.data
 from _pytest.fixtures import SubRequest
 from matplotlib import pyplot as plt
 from pytest import MonkeyPatch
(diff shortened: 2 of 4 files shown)


adamjstewart:
Violates our AI policy, which forbids fully AI-generated code for copyright reasons.

calebrob6:
@adamjstewart, I marked this as a draft for exactly this reason, it isn't ready to review by any stretch (I pretty clearly haven't written a human pr description either). There are quite a few things that I'd like to test on multiple machines as I'm looking into replacing the webdataset functionality, and having a PR is easiest way for me to do that. E.g.s:

- what are the implications of all this? Fable seems to think that we will need to handle distributed training natively in order to mirror the webdataset functionality.
```
self.dataset = (
    wds.WebDataset(*args, **kwargs)
    .shuffle(10)  # shuffle individual samples before batching
    .decode()  # decode binary data
    .map(self._drop_metadata)  # remove non-Tensor metadata
    .select(self._has_all_modalities)  # select samples with all modalities
    .map(self._sample_one_local_patch)  # sample one local patch for S1 and S2
    .map(self._sample_one_time_stamp)  # sample one timestamp for all modalities
)
```
- what is the simplest test script that I can use to assert the functionality of the old dataset is the same as the new?
- does our current dataset object work correctly?

But, this was a good reminder that I have much higher priority things to be doing lol, thanks!

adamjstewart:
Our AI policy doesn't distinguish between draft and ready-for-review. It _could_, but it doesn't yet. You can always open a PR against your own fork if you want to experiment.

PR #3961, opened 2026-08-09
Add progress bar to download_url via tqdm.wrapattr

### LINKED ISSUE #3797: Better download indicators

calebrob6:
Closes #3797. A minimal alternative to #3877 (credit to @Haris-bin-shakeel for the original approach there).

Test with something like:
```
import tempfile
from torchgeo.datasets.utils import download_url

url = 'https://hf.co/datasets/torchgeo/eurosat/resolve/1ce6f1bfb56db63fd91b6ecc466ea67f2509774c/EuroSAT100.zip'
with tempfile.TemporaryDirectory() as root:
    download_url(url, root, filename='EuroSAT100.zip')
    print('downloaded')

    download_url(url, root, filename='EuroSAT100.zip')  # shouldn't do anything
    print('done')
```

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,6 +86,8 @@ dependencies = [
     "torchmetrics[detection]>=1.5",
     # torchvision 0.17+ required for Python 3.12 wheels
     "torchvision>=0.17",
+    # tqdm 4.57+ required to match lightning's own tqdm dependency
+    "tqdm>=4.57",
     # typing-extensions 4.8+ required by torch 2.2
     # can be removed once Python 3.13 is minimum supported version
     "typing-extensions>=4.8",
--- a/torchgeo/datasets/utils.py
+++ b/torchgeo/datasets/utils.py
@@ -39,6 +39,7 @@ from rasterio.vrt import WarpedVRT
 from shapely import Geometry, MultiPolygon, Polygon, box
 from torch import Tensor
 from torchvision.utils import draw_segmentation_masks
+from tqdm import tqdm
 from typing_extensions import deprecated
 
 from .errors import DependencyNotFoundError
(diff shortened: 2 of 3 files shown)


calebrob6:
I tested and indeed see a tqdm progress bar

calebrob6:
If we didn't have the pesky formatting rules, this would indeed be @adamjstewart's 5 line request.

Speaking of @adamjstewart, this is where I got the code from https://stackoverflow.com/a/63831344 but Mike might be an AI so I clicked partial AI.

adamjstewart:
I don't think Mike from 6 years ago was AI lol. Technically everything on StackOverflow is CC-BY-SA-4.0, but this is only a few lines of code, so probably not copyrightable anyway. I think it's safe to call this human-made.

Tested on PASTIS100 and output looks great!

adamjstewart (approved):

PR #3881, opened 2026-07-15
PASTIS: normalize each timestep independently in plot

yichiac:
### Summary
This PR improves quantile normalization in PASTIS's plot function to normalize each timestep independently. The quantiles are computed over the entire TxCxHxW in main, and this PR's quantiles are computed per CxHxW frame. Each frame is stretched to its own contrast range per time step.

### Rationale
Previously, a single quantile range was shared across all time steps. Because some timesteps contain extremely bright imagery (heavy cloud cover or high exposure), the upper quantile becomes too high for typical clear-sky frames, thereby dimming them. Normalizing per frame makes each displayed frame's contrast independent of outliers at other timesteps in the series.

Here are examples of before and after the quantile normalization change:
pastis_compare_49
pastis_compare_96
pastis_compare_61

### Checklist

- [x] I have read the [Contributing docs](https://docs.torchgeo.org/en/stable/user/contributing.html)
- [x] I have read the [AI policy](https://github.com/torchgeo/governance/blob/main/AI-POLICY.md)

- [ ] 🟢 **No AI usage**: written by humans, for humans
- [x] 🟡 **AI-assisted**: AI helped with the coding, but I understand every line
- [ ] 🔴 **AI-generated**: AI did everything, review with caution

@adamjstewart @robmarkcole

PR #3963, opened 2026-08-10
PASTIS: animated time series GIF

adamjstewart:
### Summary

Includes the following changes:

- [x] Instead of plotting the first two timestamps, plot an animated GIF
- [x] Don't cast to numpy, keep everything as a Tensor

Could be separated into two PRs if desired to make it easier to review.

#### Before

before

#### After

after

Note that this is a backwards-incompatible change, as we now return a FuncAnimation instead of a Figure. However, `plt.show()` continues to work as expected, and saving the plot saves it as a static PNG of the current image, so it shouldn't be too different. Unfortunately only returning `fig` doesn't work, as the animation is closed immediately. `ds.plot()` also doesn't work, you have to use `ani = ds.plot()`.

Curious what @calebrob6 thinks of this. I have a student working on TorchGeo plotting and want to make more of our plots interactive. This PR is a proof of concept, but the plan is to eventually do this for all time-series datasets.

### Rationale

We used the previous two-timestamp plot in a manuscript and the reviewers complained that we should either plot the first timestamp or plot all timestamps. This PR is the best of both worlds, as you can directly visualize the animation and save it as either a GIF (dynamic) or PNG (static).

P.S. I'm considering reverting #3881, as the animation is more clear if all frames use the same normalization. We can change the threshold to account for frequent cloud cover.

@yichiac @robmarkcole 

### Checklist

### DIFF
--- a/torchgeo/datasets/pastis.py
+++ b/torchgeo/datasets/pastis.py
@@ -7,12 +7,13 @@ import os
 from collections.abc import Callable, Sequence
 from typing import ClassVar
 
+import einops
 import geopandas as gpd
 import matplotlib.pyplot as plt
 import numpy as np
 import torch
+from matplotlib.animation import FuncAnimation
 from matplotlib.colors import ListedColormap
-from matplotlib.figure import Figure
 from torch import Tensor
 
 from .errors import DatasetNotFoundError
@@ -377,7 +378,7 @@ class PASTIS(NonGeoDataset):
 
     def plot(
         self, sample: Sample, show_titles: bool = True, suptitle: str | None = None
-    ) -> Figure:
+    ) -> FuncAnimation:
         """Plot a sample from the dataset.
 
         Args:
@@ -386,53 +387,49 @@ class PASTIS(NonGeoDataset):
             suptitle: optional string to use as a suptitle
 
         Returns:
-            a matplotlib Figure with the rendered sample
+            a matplotlib animation with the rendered sample
         """
-        # Keep the RGB bands and quantile-normalize each frame independently.
-        rgb_frames = sample['image'][:, [2, 1, 0], :, :]
-        rgb_frames = torch.stack(
-            [quantile_normalization(frame) for frame in rgb_frames]
-        )
-        images = rgb_frames.numpy().transpose(0, 2, 3, 1)
-        mask = sample['mask'].numpy()
+        frames = sample['image'][:, [2, 1, 0], :, :]
+        frames = torch.stack([quantile_normalization(frame) for frame in frames])
+        frames = einops.rearrange(frames, 't c h w -> t h w c')
 
+        mask = sample['mask']
         if self.mode == 'instance':
             label = sample['label']
-            mask = label[mask.argmax(axis=0)].numpy()
+            mask = label[mask.argmax(axis=0)]
 
-        num_panels = 3
+        num_panels = 2
         showing_predictions = 'prediction' in sample
         if showing_predictions:
-            predictions = sample['prediction'].numpy()
+            predictions = sample['prediction']
             num_panels += 1
             if self.mode == 'instance':
                 predictions = predictions.argmax(axis=0)
                 label = sample['prediction_labels']
-                predictions = label[predictions].numpy()
+                predictions = label[predictions]
 
         fig, axs = plt.subplots(1, num_panels, figsize=(num_panels * 4, 4))
-        axs[0].imshow(images[0])
-        axs[1].imshow(images[1])
-        axs[2].imshow(mask, vmin=0, vmax=19, cmap=self.cmap, interpolation='none')
+        ani = FuncAnimation(fig, func=axs[0].imshow, frames=frames)
         axs[0].axis('off')
+
+        axs[1].imshow(mask, vmin=0, vmax=19, cmap=self.cmap, interpolation='none')
         axs[1].axis('off')
-        axs[2].axis('off')
         if showing_predictions:
-            axs[3].imshow(
+            axs[2].imshow(
                 predictions, vmin=0, vmax=19, cmap=self.cmap, interpolation='none'
             )
-            axs[3].axis('off')
+            axs[2].axis('off')
 
         if show_titles:
-            axs[0].set_title('Image 0')
-            axs[1].set_title('Image 1')
-            axs[2].set_title('Mask')
+            axs[0].set_title('Images')
+            axs[1].set_title('Mask')
             if showing_predictions:
-                axs[3].set_title('Prediction')
+                axs[2].set_title('Prediction')
 
         if suptitle is not None:
-            plt.suptitle(suptitle)
-        return fig
+            fig.suptitle(suptitle)
+
+        return ani
 
 
 class PASTIS100(PASTIS):


robmarkcole:
Plotting of long timeseries is quite challenging - samples with many ts typically end up looking crowded. What I've done is practice is passed a **number** of timestamps to plot, and that typically depends on how dense the scene is, what changes need to be shown over time etc. 

RE gifs I have generated them for some projects, but I often end up playing with the time between each change, and you also want to show the timestamp in the frame. Overall it can get quite custom and project specific, and I'm not sure we want to go too deep on plotting in torchgeo?

adamjstewart:
I like the idea of adding the timestamp. The goal of #3774 is to have a single `plot()` method for _most_ datasets. This allows us to go deep into plotting without the inconsistencies and maintenance burden that result from code duplication. We could easily decide on a fixed timestamp sample key and use that consistently.

gatienc:
I'm working a lot with satellite image timeseries and I'm most of the time ending up with implementing a slider for the timestamp in the timeseries, I usually use the notebook capability for it but matplotlib might have some interactive part to do it

adamjstewart:
I like the slider idea too, will let me student explore that.

calebrob6:
Some thoughts as requested:
- GIF is going wayy too fast to see anything
  - That said, it looks like there is some jitter/misalignment between time steps at the 1-2 pixel scale
- I wouldn't say raw torchgeo plot is often (ever?) appropriate for paper figures and that shouldn't be the goal
- +1 to adding time steps or at least frame numbers
- The most useful thing to me here would likely be plotting a specific index (or set of indices) from the time series vs. seeing all of them

Zooming out, I think the most useful potential features in torchgeo related to plotting would be:
- plotting the same sample before and after augmentation step without having to stand up/care about a ton of lightning details
- option of plotting masks on top of imagery instead of side-by-side

nilsleh:
If the rational for adding GIF functionality is a reviewer complaint, then I am not sure how this is addressing the complaint effectively, because you cannot include the GIF in a paper anyway? And I also see the torchgeo data plot as just a fast convenient check for dataset loading, but not at all for figures I would include in a paper or use as data exploration to get an understanding of the dataset.

adamjstewart:
> GIF is going wayy too fast to see anything

This can be changed.

> I wouldn't say raw torchgeo plot is often (ever?) appropriate for paper figures and that shouldn't be the goal

I use it for this, I thought that was also your motivation for adding `show_titles` to most `plot` methods?

> plotting the same sample before and after augmentation step without having to stand up/care about a ton of lightning details

I think this is already the case for all plot methods that use `quantile_normalization` instead of hard-coded division.

> option of plotting masks on top of imagery instead of side-by-side

Completely agree, see #1467

> I am not sure how this is addressing the complaint effectively, because you cannot include the GIF in a paper anyway?

In particular, this is for a tutorial notebook. So the notebook can display the animation, and the paper can include a single static image instead of the first 2 (feels very random).

P.S. I just realized there's a chance that returning a `FuncAnimation` instead of a `Figure` will break our DataModule add_figure logic, so let's hold off on this until @yichiac has time to review this in combination with the tutorial.

calebrob6:
> I think this is already the case for all plot methods that use quantile_normalization instead of hard-coded division.

Sorry, maybe I wasn't clear. A common pre-training / training debugging step is looking at your data _right_ before it gets pushed through the model, i.e. after it has been passed through any augmentation. Augmentations include things like rotation, resize/crop, brightness, gamma, etc. It is important to do this to make sure your augmentations aren't washing out the signal that you are trying to learn (e.g. saturating your imagery). This is not simply a what you are dividing by to get visible image issue. As far as I know, this is not addressed by torchgeo, and further made complicated by the fact that datamodules only apply augmentation when used with a trainer.

adamjstewart:
> looking at your data _right_ before it gets pushed through the model

We plot the image, ground truth, and predictions _after_ being pushed through the model. Assuming the model doesn't overwrite the image, this should be the same. Note that we don't plot _before_ data augmentation, although one could do this.

> further made complicated by the fact that datamodules only apply augmentation when used with a trainer.

I think it's really up to the user. If you want to use a dataset/model manually, you have always needed to apply your own data augmentation and decide if/when to plot. If you want to use a datamodule without a trainer, you need to manually prepare_data, setup, on_batch_to_device, etc. If you want to use a datamodule with a trainer, you don't need to do anything. We can't really change the design of Lightning, we can just follow best practice and keep things standardized.

Anyway, let's keep this specific PR focused on what the PASTIS plot should look like and leave general plotting discussions to #3774 or #1467 or a new issue/discussion. This is a short-term fix/suggestion, but at some point all of this will be moved to PlottingMixin and shared by most datasets.

yichiac:
I am thinking of introducing date information to the plot. Instead of showing "Images", I want to print dates for the corresponding images as titles. To do so, we have two options: 
1. Printing the dynamic time indice for each image, so the date title changes with the image in the GIFs.
2. Printing the range of time, say 2018-09-24 - 2019-12-31, and stay the same in the GIFs (mentioned by @calebrob6).

To add time to plot titles, we need to introduce date information from `metadata.json` `to _load_files()` in `datasets/pastis.py`. @adamjstewart do you think we can integrate this new feature to this PR or should be a separate PR? This new feature may need to change tests and test data since the current test data don't have dates info. 

I also think that @gatienc's slider idea is cool.

adamjstewart:
I think 1 is quite doable, let me do this in this PR. But can you check whether or not this breaks tensorboard support? Not sure if we can log animations instead of figures.

yichiac:
@adamjstewart FuncAnimation breaks TensorBoard logging. I think we need a separate function like `plot_animation()` to support time-series animation. 

```python
from torch.utils.tensorboard import SummaryWriter
from torchgeo.datasets import PASTIS100

ds = PASTIS100(root='data/pastis100')
ani = ds.plot(ds[0])
SummaryWriter('runs').add_figure('image/0', ani, global_step=0)
```
```python
  File "<python-input-0>", line 6, in <module>
    SummaryWriter('runs').add_figure('image/0', ani, global_step=0)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniforge/base/envs/dali/lib/python3.14/site-packages/torch/utils/tensorboard/writer.py", line 761, in add_figure
    figure_to_image(figure, close),
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniforge/base/envs/dali/lib/python3.14/site-packages/torch/utils/tensorboard/_utils.py", line 37, in figure_to_image
    image = render_to_rgb(figures)
  File "/opt/homebrew/Caskroom/miniforge/base/envs/dali/lib/python3.14/site-packages/torch/utils/tensorboard/_utils.py", line 23, in render_to_rgb
    canvas = plt_backend_agg.FigureCanvasAgg(figure)
  File "/opt/homebrew/Caskroom/miniforge/base/envs/dali/lib/python3.14/site-packages/matplotlib/backend_bases.py", line 1718, in __init__
    figure.set_canvas(self)
    ^^^^^^^^^^^^^^^^^
AttributeError: 'FuncAnimation' object has no attribute 'set_canvas'
```

adamjstewart:
Looks like there is no option to log animations in tensorboard: https://docs.pytorch.org/docs/2.13/tensorboard.html. I guess I could open a PR to add that, but we would have to wait a while before we could use it.

PR #3969, opened 2026-08-11
Restore uv cache pruning

isaaccorley:
Restore uv cache pruning after the `setup-uv` v9 default changed.

See https://github.com/astral-sh/setup-uv/releases/tag/v9.0.0

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -22,6 +22,7 @@ jobs:
         with:
           python-version: '3.14'
           enable-cache: true
+          prune-cache: true
           activate-environment: true
           cache-dependency-glob: |
             pyproject.toml
@@ -50,6 +51,7 @@ jobs:
         with:
           python-version: '3.14'
           enable-cache: true
+          prune-cache: true
           activate-environment: true
           cache-dependency-glob: |
             pyproject.toml
--- a/.github/workflows/style.yaml
+++ b/.github/workflows/style.yaml
@@ -24,6 +24,7 @@ jobs:
         with:
           python-version: '3.14'
           enable-cache: true
+          prune-cache: true
           activate-environment: true
           cache-dependency-glob: |
             pyproject.toml
@@ -48,6 +49,7 @@ jobs:
         with:
           python-version: '3.14'
           enable-cache: true
+          prune-cache: true
           activate-environment: true
           cache-dependency-glob: |
             pyproject.toml
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -27,6 +27,7 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           enable-cache: true
+          prune-cache: true
           activate-environment: true
           cache-dependency-glob: |
             pyproject.toml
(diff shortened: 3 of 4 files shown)


calebrob6 (approved):

adamjstewart:
I think we've gone back and forth on this a couple times. I think at one point we added setup-python caching, but it was broken, so we switched to GitHub Actions caching, but the caches were massive and almost immediately got evicted? I'm not even sure I was aware of the change back to no-caching when we switched to setup-uv.

Looking at https://github.com/torchgeo/torchgeo/actions/caches, it seems like we are recreating the cache constantly, so not sure how often the cache is ever reused anyway. Only way to improve reuse would be to slow down or group dependabot updates, but some caches are 2.5 GB (1/4 of our entire cache budget).

Looking at the most recent commit on main, I see:

1. [cache hit](https://github.com/torchgeo/torchgeo/actions/runs/31457874798/job/93675319329):
   * set up uv: 1m 36s
   * install deps: 3s
   * post set up uv: 0s
2. [no cache hit, failed to cache](https://github.com/torchgeo/torchgeo/actions/runs/31457874798/job/93675319496):
   * set up uv: 3s
   * install deps: 22s
   * post set up uv: 15s
3. [no cache hit, successful cache](https://github.com/torchgeo/torchgeo/actions/runs/31457874798/job/93675319453)
   * set up uv: 1s
   * install deps: 22s
   * post set up uv: 24s

Looking at this PR, I see:

4. [no cache hit, no cache](https://github.com/torchgeo/torchgeo/actions/runs/31445979291/job/93640187662?pr=3969)
   * set up uv: 2s
   * install deps: 21s
   * post set up uv: 1s

So in conclusion:

1. We only have a 10 GB cache budget
2. Each unit test dependency cache can be up to 2.5 GB
3. This is way too small for all of our Python version x OS matrix combinations (at least 9)
4. As a result, we rarely use the cache anyway
5. Even if we do hit the cache, it's significantly slower than redownloading all deps

I feel bad for PyPI, but until the cache gets smaller (it won't) or faster (it might), I agree that this is the right approach to save CI times.

adamjstewart:
If I understand correctly from reading https://github.com/astral-sh/setup-uv/issues/745, `enable-cache: true` + `prune-cache: true` means we cache locally-built wheels and sdists but not pre-built wheels? I'm curious how fast things would be if we don't cache anything, we shouldn't have many deps without wheels anyway. Let me benchmark this in a separate PR.

adamjstewart:
Okay, deps that do not yet have wheels:

1. antlr4-python3-runtime
   * When: all versions/OSes
   * Why: old version pulled in by omegaconf/hydra-core, will be updated in next releases
2. tornado
   * When: 3.14t
   * Why: pulled in by jupyter, no 3.14t wheels yet: https://github.com/tornadoweb/tornado/issues/3501
3. safetensors
   * When: 3.14t
   * Why: pulled in by transformers, no 3.14t wheels yet: https://github.com/safetensors/safetensors/issues/698

That's actually everything! I'm sure there will be more once we early adopt 3.15. But the ones that do require building seem to take quite some time, so I think this PR is better than #3970. Let's merge.

adamjstewart (approved):

PR #3971, opened 2026-08-11
Docs: fix linkcheck errors

adamjstewart:
### Summary

- [x] Fix most broken links discovered by `sphinx-build linkcheck`
- [x] Silence expected warnings about redirects and missing anchors
- [x] Replace all instances of http:// with https:// 

### Rationale

Part of the release process, broken links are bad, raw HTTP is unsafe.

We could consider adding this check to CI someday, but websites like spacenet.ai are down more often than they are up, so we would end up with a lot of flaky tests. Even gdal.org is down at the time of this PR.

Reference: https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-the-linkcheck-builder

### Checklist


torchgeo/datasets/agrifieldnet.py
```diff
@@ -116,7 +116,7 @@ class AgriFieldNet(IntersectionDataset):
     """AgriFieldNet India Challenge dataset.
 
     The `AgriFieldNet India Challenge
-    <https://zindi.africa/competitions/agrifieldnet-india-challenge>`__ dataset
+    <https://zindi.world/competitions/agrifieldnet-india-challenge>`__ dataset
```
robmarkcole, line 119: Did you update date these links manually?

robmarkcole (approved):
Not sure how to review this without checking links manually, so I am going to trust that you did.!

torchgeo/datasets/agrifieldnet.py
adamjstewart: Yes

adamjstewart: `make linkcheck` also prints out before and after in the case of redirect like this one

adamjstewart:
You can run:
```console
> cd docs
> make linkcheck
```
to check all links automatically.

isaaccorley (approved):

PR #3076, opened 2025-11-05
CI: run all PyTorch tests in inference mode

adamjstewart:
Curious to see if this has any effect on our test times. It didn't locally, but it might in CI where RAM is limited.

PR #3972, opened 2026-08-11
Test model forwards in inference mode

isaaccorley:
### Summary

Use inference mode for model forward tests.

### Rationale

Avoid autograd overhead in inference-only tests.

### Checklist

### DIFF
--- a/tests/models/test_changestar.py
+++ b/tests/models/test_changestar.py
@@ -17,7 +17,7 @@ SF = [4, 8, 1]
 
 
 class TestChangeStar:
-    @torch.no_grad()
+    @torch.inference_mode()
     def test_changestar_farseg_classes(self) -> None:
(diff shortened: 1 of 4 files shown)


nilsleh (approved):

adamjstewart:
What's the performance difference?

adamjstewart:
Note that we tried something similar in https://github.com/torchgeo/torchgeo/pull/3076 and found that it made CI slower, not faster. This PR tried both `inference_mode` and `no_grad`, but the former doesn't work for every model for some reason. @nilsleh @isaaccorley how was this PR tested/reviewed?

isaaccorley:
It reduces memory usage which is the goal of this PR. If we can't reduce mem pressure then we can't use pytest-xdist to parallelize.

It would be good to pick one or the other for CI.

PR #3981, opened 2026-08-12
Add ELECTS EarlyRewardLoss 

robmarkcole:
### Summary

Add loss from https://github.com/MarcCoru/elects

CC @MarcCoru

Note: Unlike the reference ELECTS implementation by Marc, this version fixes an off-by-one error in stopping probabilities (TBC) and supports both sequence-level `(B,)` and spatial `(B, H, W)` targets **without repeating labels across time**. This makes the loss compatible with standard TorchGeo classification and segmentation data formats while preserving the reference implementation support for repeated targets.

### Rationale

Long term objective: Reproducing elects paper with torchgeo and use of ELECTS with other models. 

Note ELECTS a training method, not a fixed architecture. Any causal sequence encoder could be used if it produces features at every timestep and has:

- A classification head producing (B, T, classes).
- A sigmoid stopping head producing (B, T).
- No access to future observations when predicting timestep t.

Thus, an LSTM with those two heads is sufficient -> a follow up PR can address this with a PASTIS-compatible ConvLSTM variant

### Checklist

- [ ] I have read the [Contributing docs](https://docs.torchgeo.org/en/stable/user/contributing.html)

### DIFF
--- a/torchgeo/losses/elects.py
+++ b/torchgeo/losses/elects.py
@@ -26,7 +26,7 @@ class EarlyRewardLoss(Module):
     <https://doi.org/10.1016/j.isprsjprs.2022.12.016>`_. It balances
     classification accuracy against making correct predictions early in a sequence.
 
-    .. versionadded:: 0.10
+    .. versionadded:: 0.11
     """
 
     def __init__(
--- a/torchgeo/losses/elects.py
+++ b/torchgeo/losses/elects.py
@@ -1,0 +1,112 @@
+# Copyright (c) TorchGeo Contributors. All rights reserved.
+# Licensed under the MIT License.
+
+"""Loss function for early classification of time series."""
+
+import torch
+from torch import Tensor
+from torch.nn import NLLLoss
+from torch.nn.modules import Module
+
+
+def _decision_probability(probability_stopping: Tensor) -> Tensor:
+    """Compute the probability of making a decision at each time step."""
+    remaining = torch.cumprod(1 - probability_stopping[:, :-1], dim=1)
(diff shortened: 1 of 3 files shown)


torchgeo/losses/elects.py
```diff
@@ -0,0 +1,141 @@
...
+
+class EarlyRewardLoss(Module):
+    """ELECTS loss for early classification of time series.
+
+    This loss is defined in `'End-to-end learned early classification of time
+    series for in-season crop type mapping'
+    <https://doi.org/10.1016/j.isprsjprs.2022.12.016>`_. It balances
+    classification accuracy against making correct predictions early in a sequence.
+
+    .. versionadded:: 0.10
```
calebrob6:
```suggestion
    .. versionadded:: 0.11
```

calebrob6 (commented):
This is a really solid addition to our lonely torchgeo.losses! I checked and it seems to follow the same pattern as the QR and RQ losses.

~Probably need to add to docs?~ Nope!

calebrob6:
Is this ready for review @robmarkcole? it LGTM!

robmarkcole:
@calebrob6 yes ready, but I'm holidaying for a couple
Of weeks so not urgent!

calebrob6 (approved):

calebrob6:
I tested this on CropHarvest and it works great

isaaccorley (approved):

robmarkcole:
> I tested this on CropHarvest and it works great

Can you publish the gist?

PR #3984, opened 2026-08-13
Automatically list archives that are not unpacked

adriantre:
### Summary

`find_files` now automatically descends into archives (zip, tar, …), so a GeoDataset can be pointed at a single root directory and still find files inside any archives under it. Archives already extracted alongside themselves are skipped, so the same data is never listed twice.

Given a root directory that mixes loose files with zip archives — some already extracted, some not:

```
data/
├── scene_a.tif        # loose raster
├── readme.txt         # non-matching file
├── scene_b.zip        # archive, NOT extracted
├── scene_c.zip        # archive, already extracted (into scene_c/ below)
└── scene_c/           # extracted contents of scene_c.zip
    ├── B02.tif
    └── B03.tif
```

find_files descends into archives automatically and returns files only:

```python
>>> find_files('data', '*.tif')
[
    'data/scene_a.tif',                          # loose file
    'data/scene_c/B02.tif',                      # extracted → archive skipped, no duplicate
    'data/scene_c/B03.tif',
    '/vsizip/data/scene_b.zip/scene_b/B02.tif',  # not extracted → read straight from the archive
    '/vsizip/data/scene_b.zip/scene_b/B03.tif',
]
```
(Paths shown relative to the root for readability; in practice they are absolute.)

### Rationale

On main, `files` inside an archive are only listed if the user passes an explicit `/vsizip/`-prefixed path for each archive. So you can't just point a dataset's paths at one root and have it find anything. This makes descent automatic, with no new API or per-dataset configuration.

### Implementation Details
I refactored `find_files` to overcome increased complexity. It now acts like a singledispatch for separate handling of the three different path types.
```python
def find_files(path: Path, filename_glob: str = '*') -> list[str]:
    if os.path.isfile(path):
        files = _files_under_file(path, filename_glob)
    elif os.path.isdir(path):
        files = _files_under_dir(path, filename_glob)
    elif str(path).startswith('/vsi'):
        files = _files_under_vsi(path, filename_glob)
    else:
        files = set()
    return sorted(files)
```

| Group | Functions |
|-------|-----------|
| Archives (pure, no I/O) | `_ARCHIVE_VSI_PREFIXES`, `_Archive` (NamedTuple), `_as_archive` |
| Enumeration (I/O) | `_walk_local`, `_walk_vsi` (paired "list every file under root") |
| Expansion + dedup | `_is_extracted`, `_archive_members` |
| Matching | `_match_basename` |
| Entry point | `_files_under_file` / `_files_under_dir` / `_files_under_vsi`, then `find_files` |

Keeping the PR in draft until I get a go.

### Checklist


calebrob6:
Hey @adriantre is this something that you usually run into?

adriantre:
I use this all the time, I like not needing the extra step of unpacking, and it gives me the flexibility of running on any directory without altering the file structure. I currently have this functionality as a wrapper around lightning cli, listing the files archives and prepending the `/vsizip/` prefix to all before passing into torchgeo. So consider it a nice-to-have, but maybe this is useful to others?

PR #3985, opened 2026-08-13
0.10.0 release

adamjstewart:
### Summary

Closes https://github.com/torchgeo/torchgeo/milestone/31
Closes https://github.com/torchgeo/torchgeo/discussions/3945

### Rationale

Get new features to our users!

### Checklist

### DIFF
--- a/torchgeo/__init__.py
+++ b/torchgeo/__init__.py
@@ -10,5 +10,7 @@ The :mod:`torchgeo` package consists of popular datasets, model architectures, a
 common image transformations for geospatial data.
 """
 
+from importlib.metadata import version
+
 __author__ = 'Adam J. Stewart'
-__version__ = '0.10.0.dev0'
+__version__ = version(__name__)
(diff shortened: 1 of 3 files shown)


torchgeo/__init__.py
```diff
@@ -10,5 +10,7 @@
 common image transformations for geospatial data.
 """
 
+from importlib.metadata import version
+
 __author__ = 'Adam J. Stewart'
-__version__ = '0.10.0.dev0'
+__version__ = version(__name__)
```
adamjstewart, line 16: I don't love this solution as it doesn't seem to work reliably, but I also don't want to duplicate our version in even more places. `uv-build` does not support dynamic versioning like setuptools does, so we can't reference `__version__` in pyproject.toml. Only option is to go the other way. This is also recommended by the uv devs, and there seems to be no intentions of supporting other options: https://github.com/astral-sh/uv/issues/20123

adamjstewart: Alternative is to not have any `__version__` and instead use `importlib.metadata.version` everywhere we need this info.

adamjstewart:
Wrong remote...

PR #3989, opened 2026-08-13
VHR-10: derive length from loaded images

calebrob6:
### Summary

Determine VHR10 split length dynamically.
Check for the case where the VHR10 dataset was downloaded but not extracted.

### Rationale

VHR10 previously hardcoded a length of 650 positive and 150 negative images even when a partial or fake dataset was loaded (https://github.com/torchgeo/torchgeo/pull/3500). This didn't cause problems previously because we also hardcoded those values into the dataset tests, and monkeypatched the instance segmentation tests to make len return 5 instead. It looks like we missed monkeypatching in the predict step (which didn't cause tests to fail as it only iterates for a single batch as fast_dev_run is usually true). When running the integration tests with fast_dev_run=False, the prediction step is looking for all 650 items, which don't exist because the test dataset only has 5.

The simplest fix is to just add `monkeypatch.setattr(VHR10, '__len__', lambda self: 5)` to the test_predict method in the instance segmentation trainer tests, but this seems like a better way to do it overall (as a reader of the instance segmentation trainer tests shouldn't need to figure our why tf we're monkeypatching a dataset length method).

Bonus round: previously,  `_check_integrity()` could succeed when the archive existed even if its extracted image directory was missing. We now check for this too.

### Checklist

### DIFF
--- a/tests/datasets/test_vhr10.py
+++ b/tests/datasets/test_vhr10.py
@@ -39,7 +39,7 @@ class TestVHR10:
 
     def test_len(self, dataset: VHR10) -> None:
         if dataset.split == 'positive':
-            assert len(dataset) == 650
+            assert len(dataset) == 5
         elif dataset.split == 'negative':
             assert len(dataset) == 150
 
--- a/tests/datasets/test_vhr10.py
+++ b/tests/datasets/test_vhr10.py
@@ -38,10 +38,7 @@ class TestVHR10:
             assert isinstance(x['mask'], torch.Tensor)
 
     def test_len(self, dataset: VHR10) -> None:
-        if dataset.split == 'positive':
-            assert len(dataset) == 650
-        elif dataset.split == 'negative':
-            assert len(dataset) == 150
+        assert len(dataset) == 5
 
     def test_already_downloaded(self, dataset: VHR10) -> None:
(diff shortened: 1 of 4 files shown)


calebrob6:
Btw I verified this works with the actual VHR10 dataset

nilsleh (approved):

adamjstewart (approved):
Changes to `__len__` look good and are definitely more in-line with our other datasets. Changes to `__init__` and `__getitem__` aren't necessary and are probably slower, although not noticeably for such a small dataset.

adamjstewart:
How did dependabot end up as a co-author on this PR lol?

calebrob6:
> Changes to __init__ and __getitem__ aren't necessary and are probably slower, although not noticeably for such a small dataset.

@adamjstewart I'm not sure you looked over this closely. To be clear re speed, the changes to __getitem__ just change a hardcoded filename to a dict lookup. The changes to init add a single additional glob.glob over 750 files which we've previously established is such a small diff even over tens of thousands of files that its irrelevant.

The file index is necessary to count the length dynamically. Suggestions on how to do this differently are welcome if you think this is actually unnecessary.

The other changes in __init__ implement the check I described in the PR description.

